PR #26292, opened 2019-03-29
Adding Era dependent  effective area for miniIsolation calculation in muon selectors

bmahakud:
- Adding Era dependent effective area for mini isolation calculation 
- The ticket that concerns this is : https://its.cern.ch/jira/browse/CMSMUONS-206
-  The effective areas that are used to calculate the mini Isolation changes with Era
- Currently we have two sets of optimized effective areas are available: https://github.com/bmahakud/cmssw/blob/EraBasedEffectiveAreaforMiniIsolation/PhysicsTools/PatUtils/src/MiniIsolation.cc#L67

### DIFF
--- a/PhysicsTools/PatUtils/src/MiniIsolation.cc
+++ b/PhysicsTools/PatUtils/src/MiniIsolation.cc
@@ -58,16 +58,35 @@ PFIsolation getMiniPFIsolation(const pat::PackedCandidateCollection *pfcands,
   float muonRelMiniIsoPUCorrected(const PFIsolation& iso,
 				  const math::XYZTLorentzVector& p4,
 				  float dr,
-				  float rho)
+				  float rho,
+                                  std::string era)
   {
     double absEta = fabs(p4.eta());
     double ea = 0;
     //Spring15 version
+       if(era=="run2_miniAOD_80XLegacy"){
     if      (absEta<0.800) ea = 0.0735;
     else if (absEta<1.300) ea = 0.0619;
     else if (absEta<2.000) ea = 0.0465;
     else if (absEta<2.200) ea = 0.0433;
     else if (absEta<2.500) ea = 0.0577;
+    }
+    //Fall17 effective area
+    if(era=="run2_miniAOD_94XFall17"){
+    if      (absEta<0.800) ea = 0.0566;
+    else if (absEta<1.300) ea = 0.0562;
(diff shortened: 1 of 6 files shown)


PhysicsTools/PatUtils/src/MiniIsolation.cc
```diff
@@ -58,16 +58,35 @@ PFIsolation getMiniPFIsolation(const pat::PackedCandidateCollection *pfcands,
...
     //Spring15 version
+       if(era=="run2_miniAOD_80XLegacy"){
     if      (absEta<0.800) ea = 0.0735;
     else if (absEta<1.300) ea = 0.0619;
     else if (absEta<2.000) ea = 0.0465;
     else if (absEta<2.200) ea = 0.0433;
     else if (absEta<2.500) ea = 0.0577;
+    }
+    //Fall17 effective area
+    if(era=="run2_miniAOD_94XFall17"){
```
perrotta: This should be ```else if```, otherwise the "run2_miniAOD_80XLegacy" will never be hit

perrotta (commented):
The whole if block needs to be correctly indented.

But, more importantly: wouldn't it be more convenient to make such "ea" parameters configurable (as a vector of doubles/floats inside PATMuonProducer?), and dependent themselves on the era modifier, instead of having to add this further "era" parameter to this method?

Moreover: the method returns a float and the parameters passed to it are also float. I don't think you need all those double's in it: please move all them to float if you agree with it

PhysicsTools/PatUtils/src/MiniIsolation.cc
```diff
@@ -58,16 +58,35 @@ PFIsolation getMiniPFIsolation(const pat::PackedCandidateCollection *pfcands,
   float muonRelMiniIsoPUCorrected(const PFIsolation& iso,
 				  const math::XYZTLorentzVector& p4,
 				  float dr,
-				  float rho)
+				  float rho,
+                                  std::string era)
   {
     double absEta = fabs(p4.eta());
```
perrotta: ```std::abs```

perrotta:
@bmahakud when do you plan to apply the fixes, and also solve the conflicts?

bmahakud:
@perrotta 
yes, Sorry for bit slow response in this thread. Will do tonight or tomorrow.

bmahakud:
@slava77 
I will just create new PR ASAP

PR #26194, opened 2019-03-15
HGCAL Clue algorithm 106 x

rovere:
#### PR description:

This PR is (re)based on top of #26099. It includes several changes:

* The HGCAL clustering algorithms are now treated as plugins used by the central LayerCluster producer. This required quite some files reshuffling around packages.
* The Validation of the configuration of the algorithm is done using the central Plugin Parameter Validation.
* The default clustering algorithm has been changed. Now we use the CLUE (clustering energy) algorithm. The results of this are much better if compared to the up-to-now default one. Results have been presented in the HGCAL DPG Meeting [See here, slides 42 onward](https://indico.cern.ch/event/803066/contributions/3341958/attachments/1807341/2950233/20190306_HGCAL_DPG_ImagingAlgoModification_MR.pdf)

#### PR validation:

Tested using:

- [x] scram b code-checks
- [x] runTheMatrix.py -l limited -i all

PR #26295, opened 2019-03-29
HGCal TICL Framework

rovere:
#### PR description:

This PR will introduce the TICL framework into CMSSW. It is still not final, but we are pushing it into the release so that others could actively start developing the HGCAL reconstruction using it. Currently, it has no link with the official upgrade reconstruction sequence, but a python snippet with 2 customization functions has been added in the test directory of the newly created `RecoHGCal/TICL` to be able to plug it in. The 2 customisation function will add the TICL-sequence either replacing the full reconstruction sequence or simply adding TICL on top of that. This, we believe, gives enough flexibility to the end users.

This PR is based, once again, on another PR #26194 since TICL needs the CLUE algorithm in order to properly work. I'll rebase once that PR will be integrated, but code review can start in the meantime.

Currently, TICL has 2 iterations: one for reconstructing MIP like object within HGCAL and another for generic clusters. No PID, nor energy regression nor linking has been implemented, yet.

#### PR validation:

Validated using code-checks and the standard runTheMatrix tests, even if those tests will not cover any of the TICL code.


andrius-k:
please test

andrius-k:
+1

RecoHGCal/TICL/interface/TICLConstants.h
```diff
@@ -0,0 +1,12 @@
+#ifndef RecoHGCal_TICL_interface_TICLConstants_h
+#define RecoHGCal_TICL_interface_TICLConstants_h
+
+#include <cstdint>
+
+namespace ticlConstants {
+constexpr float minEta = 1.5f;
+constexpr float maxEta = 3.0f;
+constexpr uint8_t maxNumberOfLayers = 52;
```
kpedro88: as mentioned in numerous other PRs/issues, it is much better to get this info directly from the geometry, because it can/will change

RecoHGCal/TICL/plugins/FilteredLayerClustersProducer.cc
```diff
@@ -0,0 +1,71 @@
...
+FilteredLayerClustersProducer::FilteredLayerClustersProducer(const edm::ParameterSet& ps) {
+  clusters_token_ =
+      consumes<std::vector<reco::CaloCluster>>(ps.getParameter<edm::InputTag>("HGCLayerClusters"));
+  clustersMask_token_ =
+      consumes<std::vector<float>>(ps.getParameter<edm::InputTag>("LayerClustersInputMask"));
+  clusterFilter_ = ps.getParameter<std::string>("ClusterFilter");
+  theFilter_ = ClusterFilterFactory::get()->create(clusterFilter_, ps);
+  iteration_label_ = ps.getParameter<std::string>("iteration_label");
+
+  produces<std::vector<std::pair<unsigned int, float>>>(iteration_label_);
```
kpedro88: maybe worth `typedef`ing this?

RecoHGCal/TICL/plugins/FilteredLayerClustersProducer.cc
```diff
@@ -0,0 +1,71 @@
...
+  evt.getByToken(clusters_token_, clusterHandle);
+  evt.getByToken(clustersMask_token_, inputClustersMaskHandle);
+  const auto& inputClusterMask = *inputClustersMaskHandle;
+
+  const auto& layerClusters = *clusterHandle;
+  auto numLayerClusters = layerClusters.size();
+  availableLayerClusters->reserve(numLayerClusters);
+  for (unsigned int i = 0; i < numLayerClusters; ++i) {
+    if (inputClusterMask[i] > 0.f) {
+      availableLayerClusters->emplace_back(std::make_pair(i, inputClusterMask[i]));
```
kpedro88: i think `make_pair` is unnecessary here

RecoHGCal/TICL/plugins/FilteredLayerClustersProducer.h
```diff
@@ -0,0 +1,22 @@
+#ifndef __RecoHGCal_TICL_FilteredLayerClustersProducer_H__
+#define __RecoHGCal_TICL_FilteredLayerClustersProducer_H__
+#include <string>
+#include "RecoHGCal/TICL/interface/ClusterFilterBase.h"
+
+class FilteredLayerClustersProducer : public edm::stream::EDProducer<> {
```
kpedro88: looks like this could be `global`

RecoHGCal/TICL/plugins/HGCDoublet.h
```diff
@@ -0,0 +1,176 @@
...
+        auto otherDoubletId = innerDoublets[i + j];
+        auto &otherDoublet = allDoublets[otherDoubletId];
+        xi[j] = otherDoublet.getInnerX();
+        yi[j] = otherDoublet.getInnerY();
+        zi[j] = otherDoublet.getInnerZ();
+      }
+      for (int j = 0; j < vs; ++j) {
+        ok[j] = areAligned(xi[j], yi[j], zi[j], xo, yo, zo, minCosTheta, minCosPointing, debug);
+        if (debug) {
+          std::cout << "Are aligned for InnerDoubletId: " << i + j << " is " << ok[j] << std::endl;
```
kpedro88: cout is not allowed, please use MessageLogger

kpedro88: (everywhere in this file)

RecoHGCal/TICL/plugins/HGCDoublet.h
```diff
@@ -0,0 +1,176 @@
...
+    auto dz1 = zo - zi;
+
+    auto dx2 = getInnerX() - xi;
+    auto dy2 = getInnerY() - yi;
+    auto dz2 = getInnerZ() - zi;
+
+    // inner product
+    auto dot = dx1 * dx2 + dy1 * dy2 + dz1 * dz2;
+    // magnitudes
+    auto mag1 = std::sqrt(dx1 * dx1 + dy1 * dy1 + dz1 * dz1);
```
kpedro88: would it make more sense to store the inner and outer points as two 3-vectors that know how to do these operations?

RecoHGCal/TICL/plugins/HGCDoublet.h
```diff
@@ -0,0 +1,176 @@
...
+    // angle between the vectors
+    auto cosTheta = dot / (mag1 * mag2);
+    if (debug) {
+      std::cout << "dot: " << dot << " mag1: " << mag1 << " mag2: " << mag2
+                << " cosTheta: " << cosTheta << " isWithinLimits: " << (cosTheta > minCosTheta)
+                << std::endl;
+    }
+
+    // Now check the compatibility with the pointing origin.
+    // TODO(rovere): pass in also the origin, which is now fixed at (0,0,0)
```
kpedro88: should this todo be implemented before the PR is merged?

RecoHGCal/TICL/plugins/HGCDoublet.h
```diff
@@ -0,0 +1,176 @@
...
+      unsigned int numberOfOuterNeighbors = theOuterNeighbors_.size();
+      for (unsigned int i = 0; i < numberOfOuterNeighbors; ++i) {
+        allDoublets[theOuterNeighbors_[i]].findNtuplets(allDoublets, tmpNtuplet);
+      }
+    }
+  }
+
+ private:
+  const std::vector<reco::CaloCluster> *layerClusters_;
+  std::vector<int> theOuterNeighbors_;
```
kpedro88: don't need both "the" and trailing "_" to denote a member variable (just use the latter) (for all members below as well)

RecoHGCal/TICL/plugins/HGCGraph.cc
```diff
@@ -0,0 +1,97 @@
...
+
+#include "HGCDoublet.h"
+#include "HGCGraph.h"
+
+void HGCGraph::makeAndConnectDoublets(const std::vector<std::vector<std::vector<unsigned int>>> &h,
+                                      int nEtaBins, int nPhiBins,
+                                      const std::vector<reco::CaloCluster> &layerClusters,
+                                      int deltaIEta, int deltaIPhi, float minCosTheta,
+                                      float minCosPointing, int missing_layers) {
+  isOuterClusterOfDoublets_.clear();
```
kpedro88: should check memory performance - might want to use `swap()` or `shrink_to_fit()` to avoid hoarding memory after high-occupancy events

RecoHGCal/TICL/plugins/HGCDoublet.h
```diff
@@ -0,0 +1,176 @@
...
+    loop(lim, nDoublets - lim);
+
+    if (debug) {
+      std::cout << "Found " << theInnerNeighbors_.size() << " compatible doublets out of "
+                << nDoublets << " considered" << std::endl;
+    }
+    return theInnerNeighbors_.empty();
+  }
+
+  int areAligned(double xi, double yi, double zi, double xo, double yo, double zo,
```
kpedro88: could this function be `const`?

RecoHGCal/TICL/plugins/HGCDoublet.h
```diff
@@ -0,0 +1,176 @@
...
+    if (debug) {
+      std::cout << "dot_pointing: " << dot_pointing << " mag_pointing: " << mag_pointing
+                << " mag2: " << mag2 << " cosTheta_pointing: " << cosTheta_pointing
+                << " isWithinLimits: " << (cosTheta_pointing < minCosPointing) << std::endl;
+    }
+
+    return (cosTheta > minCosTheta) && (cosTheta_pointing > minCosPointing);
+  }
+
+  void findNtuplets(std::vector<HGCDoublet> &allDoublets, HGCntuplet &tmpNtuplet) {
```
kpedro88: could this function be `const`?

RecoHGCal/TICL/plugins/HGCGraph.cc
```diff
@@ -0,0 +1,97 @@
...
+  }
+  // #ifdef FP_DEBUG
+  if (verbosity_ > None) {
+    LogDebug("HGCGraph") << "number of Root doublets " << theRootDoublets_.size()
+                         << " over a total number of doublets " << allDoublets_.size() << std::endl;
+  }
+  // #endif
+}
+
+void HGCGraph::findNtuplets(std::vector<HGCDoublet::HGCntuplet> &foundNtuplets,
```
kpedro88: could this function be `const`?

RecoHGCal/TICL/plugins/PatternRecognitionbyCA.h
```diff
@@ -0,0 +1,85 @@
...
+                      const std::vector<std::pair<unsigned int, float> >& mask,
+                      std::vector<Trackster>& result) override;
+
+ private:
+  int getEtaBin(float eta) const {
+    constexpr float etaRange = ticlConstants::maxEta - ticlConstants::minEta;
+    static_assert(etaRange >= 0.f);
+    float r = nEtaBins_ / etaRange;
+    int etaBin = (std::abs(eta) - ticlConstants::minEta) * r;
+    etaBin = std::min(etaBin, nEtaBins_);
```
kpedro88: could use `std::clamp()` for this (new C++17 feature)

RecoHGCal/TICL/plugins/PatternRecognitionbyMultiClusters.h
```diff
@@ -0,0 +1,32 @@
...
+class Event;
+class EventSetup;
+}  // namespace edm
+// namespace reco {class CaloCluster;}
+
+class PatternRecognitionbyMultiClusters final : public PatternRecognitionAlgoBase {
+ public:
+  PatternRecognitionbyMultiClusters(const edm::ParameterSet& conf)
+      : PatternRecognitionAlgoBase(conf) {
+    std::cout << "constructing PatternRecognitionbyMultiClusters from conf" << std::endl;
```
kpedro88: cout is not allowed

RecoHGCal/TICL/plugins/TrackstersProducer.h
```diff
@@ -0,0 +1,22 @@
+#ifndef __RecoHGCal_TICL_TrackstersProducer_H__
+#define __RecoHGCal_TICL_TrackstersProducer_H__
+#include "DataFormats/CaloRecHit/interface/CaloCluster.h"
+#include "RecoHGCal/TICL/interface/PatternRecognitionAlgoBase.h"
+
+class TrackstersProducer : public edm::stream::EDProducer<> {
```
kpedro88: could be `global`

RecoHGCal/TICL/plugins/TrackstersProducer.cc
```diff
@@ -0,0 +1,90 @@
...
+  auto output_mask = std::make_unique<std::vector<float>>();
+
+  edm::Handle<std::vector<reco::CaloCluster>> cluster_h;
+  edm::Handle<std::vector<std::pair<unsigned int, float>>> filtered_layerclusters_mask_h;
+  edm::Handle<std::vector<float>> original_layerclusters_mask_h;
+
+  evt.getByToken(clusters_token_, cluster_h);
+  evt.getByToken(filtered_layerclusters_mask_token_, filtered_layerclusters_mask_h);
+  evt.getByToken(original_layerclusters_mask_token_, original_layerclusters_mask_h);
+  std::cout << "TrackstersProducer::produce" << std::endl;
```
kpedro88: cout is not allowed

RecoHGCal/TICL/plugins/TrackstersProducer.cc
```diff
@@ -0,0 +1,90 @@
...
+
+  // Now update the global mask and put it into the event
+  output_mask->reserve(original_layerclusters_mask_h->size());
+  // Copy over the previous state
+  std::copy(std::begin(*original_layerclusters_mask_h), std::end(*original_layerclusters_mask_h),
+            std::back_inserter(*output_mask));
+  // Mask the used elements, accordingly
+  for (auto const& trackster : *result) {
+    for (auto const v : trackster.vertices) {
+      // TODO(rovere): for the moment we mask the layer cluster completely. In
```
kpedro88: should this todo be implemented before the PR is merged?

RecoHGCal/TICL/plugins/TrackstersToMultiClusterProducer.cc
```diff
@@ -0,0 +1,98 @@
...
+
+#include "DataFormats/ParticleFlowReco/interface/HGCalMultiCluster.h"
+#include "RecoHGCal/TICL/interface/Trackster.h"
+
+#include <algorithm>
+#include <array>
+#include <string>
+#include <vector>
+
+class TrackstersToMultiClusterProducer : public edm::stream::EDProducer<> {
```
kpedro88: could be `global`

RecoHGCal/TICL/plugins/TrackstersToMultiClusterProducer.cc
```diff
@@ -0,0 +1,98 @@
+#ifndef __RecoHGCal_TICL_TracksterToMultiClusterProducer_H__
```
kpedro88: include guards usually aren't used in a .cc plugin file

RecoHGCal/TICL/plugins/TrackstersToMultiClusterProducer.cc
```diff
@@ -0,0 +1,98 @@
...
+  std::string label_;
+  edm::EDGetTokenT<std::vector<reco::CaloCluster>> layer_clusters_token_;
+  edm::EDGetTokenT<std::vector<Trackster>> tracksters_token_;
+};
+
+DEFINE_FWK_MODULE(TrackstersToMultiClusterProducer);
+
+TrackstersToMultiClusterProducer::TrackstersToMultiClusterProducer(const edm::ParameterSet& ps)
+    : label_(ps.getParameter<std::string>("label")) {
+  layer_clusters_token_ =
```
kpedro88: these can be initialized in the constructor initialization list

RecoHGCal/TICL/plugins/TrackstersToMultiClusterProducer.cc
```diff
@@ -0,0 +1,98 @@
...
+  auto const& tracksters = *tracksterHandle;
+  auto const& layerClusters = *layer_clustersHandle;
+
+  edm::PtrVector<reco::BasicCluster> clusterPtrs;
+  for (unsigned i = 0; i < layerClusters.size(); ++i) {
+    edm::Ptr<reco::BasicCluster> ptr(layer_clustersHandle, i);
+    clusterPtrs.push_back(ptr);
+  }
+
+  std::for_each(std::begin(tracksters), std::end(tracksters), [&](auto const& trackster) {
```
kpedro88: why use `for_each` rather than a standard (range-based) loop?

RecoHGCal/TICL/test/ticl_iterations.py
```diff
@@ -0,0 +1,128 @@
...
+  process.Tracksters.algo_verbosity = 0
+  process.Tracksters.missing_layers = 2
+  process.Tracksters.min_clusters_per_ntuplet = 15
+  process.Tracksters.min_cos_theta = 0.94 # ~20 degrees
+  process.Tracksters.min_cos_pointing = 0.7
+  process.TrackstersToMultiCluster = TrackstersToMultiCluster.clone()
+  process.TrackstersToMultiCluster.Tracksters = cms.InputTag("Tracksters", "TrackstersByCA")
+
+
+#  process.hgcalLayerClusters.ecut = 5
```
kpedro88: remove commented-out code

RecoHGCal/TICL/test/ticl_iterations.py
```diff
@@ -0,0 +1,128 @@
...
+  process.Tracksters.min_cos_theta = 0.94 # ~20 degrees
+  process.Tracksters.min_cos_pointing = 0.7
+  process.TrackstersToMultiCluster = TrackstersToMultiCluster.clone()
+  process.TrackstersToMultiCluster.Tracksters = cms.InputTag("Tracksters", "TrackstersByCA")
+
+
+  process.HGCalUncalibRecHit = HGCalUncalibRecHit
+  process.HGCalRecHit = HGCalRecHit
+  process.hgcalLayerClusters = hgcalLayerClusters
+#  process.hgcalLayerClusters.ecut = 5
```
kpedro88: remove commented-out code

RecoHGCal/TICL/interface/ClusterFilterByAlgo.h
```diff
@@ -0,0 +1,35 @@
...
+#include "RecoHGCal/TICL/interface/ClusterFilterBase.h"
+
+#include <memory>
+#include <utility>
+
+// Filter clusters that belong to a specific algorithm
+class ClusterFilterByAlgo final : public ClusterFilterBase {
+ public:
+  ClusterFilterByAlgo(const edm::ParameterSet& ps)
+      : ClusterFilterBase(ps), algo_number_(ps.getParameter<int>("algo_number")) {}
```
kpedro88: will this "algo_number" be centrally enforced somehow? It seems like it could cause confusion if not.

kpedro88 (commented):
Reviewed all the code in the new ` RecoHGCal/TICL` package. Assuming there are not other changes on top of the CLUE PR.

kpedro88:
#26194 is merged, so this PR can be rebased on CMSSW_10_6_X_2019-04-02-1100 or any later IB

slava77:
review status from reco: waiting for a rebase to get rid of the diffs from #26194

perrotta:
please test

smuzaffar:
please test
cms-bot PR https://github.com/cms-sw/cms-bot/pull/1115 should fix this

slava77:
please test

slava77:
@rovere 
please summarize the updates briefly.
I understand that the review was slow and you may have also some recent new developments.

rovere:
@slava77 I mainly fixed the comments made by Pedro. No new development has been squeezed into the review process.
So, remove couts, rename variables, avoid hard-coded numbers are all I pushed.

slava77:
@mrodozov @smuzaffar 
is the check-headers  skipping new packages?
I do not see RecoHGCal/TICL in https://cmssdt.cern.ch/SDT/jenkins-artifacts/pull-request-integration/PR-26295/34217/headers_chks.log

What about the static analyzer?

RecoHGCal/TICL/interface/ClusterFilterBase.h
```diff
@@ -0,0 +1,28 @@
...
+#include <vector>
+
+namespace edm {
+class ParameterSet;
+}
+namespace reco {
+class CaloCluster;
+}
+
+class ClusterFilterBase {
```
slava77: this name is too simple to stay in a global namespace. Please move to ```ticl```

RecoHGCal/TICL/plugins/FilteredLayerClustersProducer.h
```diff
@@ -0,0 +1,22 @@
+#ifndef __RecoHGCal_TICL_FilteredLayerClustersProducer_H__
+#define __RecoHGCal_TICL_FilteredLayerClustersProducer_H__
```
slava77: producers are recommended to stay in one file, without a header file.

RecoHGCal/TICL/plugins/HGCDoublet.h
```diff
@@ -0,0 +1,177 @@
...
+        outerR_((*layerClusters)[outerClusterId].position().r()),
+        innerX_((*layerClusters)[innerClusterId].x()),
+        outerX_((*layerClusters)[outerClusterId].x()),
+        innerY_((*layerClusters)[innerClusterId].y()),
+        outerY_((*layerClusters)[outerClusterId].y()),
+        innerZ_((*layerClusters)[innerClusterId].z()),
+        outerZ_((*layerClusters)[outerClusterId].z()),
+        alreadyVisited_(false) {}
+
+  double getInnerX() const { return innerX_; }
```
slava77: please drop the ```get``` prefix

RecoHGCal/TICL/plugins/HGCDoublet.h
```diff
@@ -0,0 +1,177 @@
...
+
+  void tagAsOuterNeighbor(unsigned int otherDoublet) { outerNeighbors_.push_back(otherDoublet); }
+
+  void tagAsInnerNeighbor(unsigned int otherDoublet) { innerNeighbors_.push_back(otherDoublet); }
+
+  bool checkCompatibilityAndTag(std::vector<HGCDoublet> &allDoublets,
+                                const std::vector<int> &innerDoublets, float minCosTheta,
+                                float minCosPointing = 1., bool debug = false) {
+    int nDoublets = innerDoublets.size();
+    int constexpr VSIZE = 4;
```
slava77: are there good reasons to keep this and other complex code in this file inlined?

RecoHGCal/TICL/plugins/PatternRecognitionbyCA.h
slava77: ```clamp``` , indeed

RecoHGCal/TICL/plugins/TrackstersProducer.cc
```diff
@@ -0,0 +1,89 @@
...
+
+TrackstersProducer::TrackstersProducer(const edm::ParameterSet& ps)
+    : myAlgo_(std::make_unique<PatternRecognitionbyCA>(ps)) {
+  clusters_token_ = consumes<std::vector<reco::CaloCluster>>(
+      ps.getParameter<edm::InputTag>("hgcal_layerclusters"));
+  filtered_layerclusters_mask_token_ = consumes<std::vector<std::pair<unsigned int, float>>>(
+      ps.getParameter<edm::InputTag>("filtered_layerclusters_mask"));
+  original_layerclusters_mask_token_ =
+      consumes<std::vector<float>>(ps.getParameter<edm::InputTag>("original_layerclusters_mask"));
+  produces<std::vector<Trackster>>("TrackstersByCA");
```
slava77: is there a plan to produce multiple collections from a single producer?
If not, please drop the instance specs

RecoHGCal/TICL/plugins/TrackstersProducer.cc
```diff
@@ -0,0 +1,89 @@
...
+#include "FWCore/Framework/interface/stream/EDProducer.h"
+#include "FWCore/MessageLogger/interface/MessageLogger.h"
+#include "FWCore/ParameterSet/interface/ConfigurationDescriptions.h"
+#include "FWCore/ParameterSet/interface/ParameterSetDescription.h"
+
+#include "DataFormats/ParticleFlowReco/interface/PFCluster.h"
+
+#include "RecoHGCal/TICL/interface/PatternRecognitionAlgoBase.h"
+#include "RecoHGCal/TICL/interface/Trackster.h"
+#include "RecoHGCal/TICL/plugins/TrackstersProducer.h"
```
slava77: producers are recommended to stay in one .cc file.
Please merge defs from .h here.

RecoHGCal/TICL/plugins/TrackstersProducer.cc
```diff
@@ -0,0 +1,89 @@
...
+  desc.add<edm::InputTag>("filtered_layerclusters_mask",
+                          edm::InputTag("FilteredLayerClusters", "iterationLabelGoesHere"));
+  desc.add<edm::InputTag>("original_layerclusters_mask",
+                          edm::InputTag("hgcalLayerClusters", "InitialLayerClustersMask"));
+  desc.add<int>("algo_verbosity", 0);
+  desc.add<double>("min_cos_theta", 0.915);
+  desc.add<double>("min_cos_pointing", -1.);
+  desc.add<int>("missing_layers", 0);
+  desc.add<int>("min_clusters_per_ntuplet", 10);
+  descriptions.add("Tracksters", desc);
```
slava77: please start with a lower case

RecoHGCal/TICL/plugins/TrackstersToMultiClusterProducer.cc
```diff
@@ -0,0 +1,92 @@
...
+
+void TrackstersToMultiClusterProducer::fillDescriptions(
+    edm::ConfigurationDescriptions& descriptions) {
+  // hgcalMultiClusters
+  edm::ParameterSetDescription desc;
+  desc.add<edm::InputTag>("Tracksters", edm::InputTag("Tracksters", "TrackstersByCA"));
+  desc.add<edm::InputTag>("LayerClusters", edm::InputTag("hgcalLayerClusters"));
+  desc.add<std::string>("label", "MultiClustersFromTracksterByCA");
+  desc.addUntracked<unsigned int>("verbosity", 3);
+  descriptions.add("TrackstersToMultiCluster", desc);
```
slava77: please start module instance name from a lower case.
Consider to flip the order in the name: ```MultiClustersFromTrackstersProducer``` to be more clear what is made rather than the origin.

RecoHGCal/TICL/interface/ClusterFilterByAlgo.h
```diff
@@ -0,0 +1,35 @@
+// Author: Marco Rovere - marco.rovere@cern.ch
+// Date: 11/2018
+// Copyright CERN
+
+#ifndef RecoHGCal_TICL_ClusterFilterByAlgo_H__
+#define RecoHGCal_TICL_ClusterFilterByAlgo_H__
```
slava77: http://cms-sw.github.io/cms_coding_rules.html#plugins
apparently this and other filters which will be available only via a factory should be moved to the /plugins dir

RecoHGCal/TICL/src/classes.h
```diff
@@ -0,0 +1,11 @@
+#include <vector>
+#include "FWCore/Common/interface/EventBase.h"
+#include "RecoHGCal/TICL/interface/Trackster.h"
+
+namespace RecoHGCal_TICL {
```
slava77: the namespace_dictionary detail is no longer necessary. Only the includes are needed.

RecoHGCal/TICL/src/classes_def.xml
```diff
@@ -0,0 +1,10 @@
+
+<lcgdict>
+  <class name="Trackster" ClassVersion="10">
+   <version ClassVersion="10" checksum="1507284825"/>
```
slava77: please start from version 3. 

In fact, do we even need version tracking for this internal data for the algorithms?
Please consider marking these as transient.
However, if there are plans to keep this saved in any production data files, the class def should be moved to the DataFormats subsystem.

RecoHGCal/TICL/test/ticl_iterations.py
```diff
@@ -0,0 +1,115 @@
...
+def TICL_iterations_withReco(process):
+  process.FEVTDEBUGHLTEventContent.outputCommands.extend(['keep *_TrackstersToMultiCluster*_*_*'])
+
+  process.FilteredLayerClustersMIP = FilteredLayerClusters.clone()
+  process.FilteredLayerClustersMIP.ClusterFilter = "ClusterFilterBySize"
+  process.FilteredLayerClustersMIP.algo_number = 8
+  process.FilteredLayerClustersMIP.max_cluster_size = 2 # inclusive
+  process.FilteredLayerClustersMIP.iteration_label = "MIP"
+  process.TrackstersMIP = Tracksters.clone()
+  process.TrackstersMIP.original_layerclusters_mask = cms.InputTag("hgcalLayerClusters", "InitialLayerClustersMask")
```
slava77: it's a quite a bit less verbose to specify all the module parameters that differ from the origin in the parentheses of the ```clone``` method call.

RecoLocalCalo/Configuration/python/RecoLocalCalo_EventContent_cff.py
```diff
@@ -71,10 +71,12 @@ def _updateOutput( era, outputPSets, commands):
...
         'keep *_HGCalUncalibRecHit_*_*',
         'keep *_hgcalLayerClusters_*_*',
-        'keep *_hgcalMultiClusters_*_*'
+        'keep *_hgcalMultiClusters_*_*',
+        'keep *_iterHGCalMultiClusters_*_*',
+
     ]
 )
-phase2_hgcal.toModify( RecoLocalCaloRECO, outputCommands = RecoLocalCaloRECO.outputCommands + ['keep *_HGCalRecHit_*_*','keep *_hgcalLayerClusters_*_*', 'keep *_hgcalMultiClusters_*_*'] )
+phase2_hgcal.toModify( RecoLocalCaloRECO, outputCommands = RecoLocalCaloRECO.outputCommands + ['keep *_HGCalRecHit_*_*','keep *_hgcalLayerClusters_*_*', 'keep *_hgcalMultiClusters_*_*', 'keep *_iterHGCalMultiClusters_*_*'] )
```
slava77: please add some linebreaks in this long line to improve readability

RecoLocalCalo/HGCalRecProducers/plugins/HGCalLayerClusterProducer.cc
```diff
@@ -86,6 +86,7 @@ HGCalLayerClusterProducer::HGCalLayerClusterProducer(const edm::ParameterSet &ps
   algo.reset(HGCalLayerClusterAlgoFactory::get()->create(pluginPSet.getParameter<std::string>("type"), pluginPSet));
   algo->setAlgoId(algoId);
 
+  produces<std::vector<float> >("InitialLayerClustersMask");
```
slava77: lower case, perhaps

slava77 (commented):
This is rather superficial at this point. Consider as an iteration 1 of the review.

Please place RecoHGCal/TICL classes in a namespace to avoid name collisions with current/future developments. A namespace ```ticl``` seems relevant.

Please also make sure that comments from @kpedro88 are addressed

mrodozov:
@slava77 I have to see the check-headers job for what you are asking, but the logic is all "touched" in the PR headers should be checked ,that (should) include new headers. The static checks has been delayed as we lost the nodes they were running on. I see the checks are available now

slava77:
> @slava77 I have to see the check-headers job for what you are asking, but the logic is all "touched" in the PR headers should be checked ,that (should) include new headers.

If the package is compiled, I would expect some header files from it mentioned in the log file
https://cmssdt.cern.ch/SDT/jenkins-artifacts/pull-request-integration/PR-26295/34217/headers_chks.log
However, they do not appear.

> The static checks has been delayed as we lost the nodes they were running on. I see the checks are available now

The check results were available originally; my question was on the same subject of including the code from the new package in the checks. It may just mean that the code here is all good and passes all the checks and we are fine.

mrodozov:
Can't say about the check-headers, it's in the scram rules which headers are selected. 
I'd search in other PRs to see if this is true for all new headers or this is exception

slava77:
@cmsbuild please test

smuzaffar:
@slava77 , yes looks like header check is not running for header headers. We will update the build rules to check run this tests for new header files too.

kpedro88:
@rovere a number of review comments are still pending. Will these be addressed soon?

rovere:
@kpedro88 
I address all the comments I share and deem useful to implement in 1 go. So, the pending comments are related to topics that I either consider minor or that could be implemented in future as optimization.
Unless there are strong arguments for specific comments that you or the other reviewers what to raise again and consider show-stoppers, I'd ask the inclusion of this PR as is now.

kpedro88:
at minimum I would like responses to the following pending comments:

technical:
https://github.com/cms-sw/cmssw/pull/26295/files#r270893309
https://github.com/cms-sw/cmssw/pull/26295/files#r270899921
https://github.com/cms-sw/cmssw/pull/26295/files#r270900206

todos:
https://github.com/cms-sw/cmssw/pull/26295/files#r270896482
https://github.com/cms-sw/cmssw/pull/26295/files#r270903583

RecoHGCal/TICL/plugins/HGCDoublet.h
rovere: Having imagined a review of 30+ days maybe I should have started.
Other people are currently committed to doing this so, so no.

RecoHGCal/TICL/plugins/TrackstersProducer.cc
rovere: Same as other TODO, but on an even longer timescale.

RecoHGCal/TICL/plugins/FilteredLayerClustersProducer.cc
rovere: Not sure what the added value is here.

RecoHGCal/TICL/plugins/HGCDoublet.h
rovere: No, read the code.

RecoHGCal/TICL/plugins/HGCGraph.cc
rovere: Yes, that could be marked as constant.

RecoHGCal/TICL/plugins/FilteredLayerClustersProducer.cc
kpedro88: avoiding a pointless operation and making correct/optimal use of `emplace_back`

slava77:
@rovere 
sorry for the delay with the review from reco.
Please remind me a configuration to be used with the best way to test your code at run time .
Thank you.

rovere:
@slava77 
the procedure is a little cumbersome but doable, of course.
I'd start from any runTheMatrix workflow for PhaseII, maybe with D41 geometry, to be up to date, and run it.
Then, in order to trigger TICL reconstruction, I'd:

1. Copy the customization file from  RecoHGCal/TICL/test into your local test directory:
```cp RecoHGCal/TICL/test/ticl_iterations.py .```
2. edit step3.py with something along the lines:
```
from ticl_iterations import TICL_iterations
```
or
```
from ticl_iterations import TICL_iterations_withReco
```

and finally, customize the process:

```
process = TICL_iterations(process)
```
or
```
process = TICL_iterations_withReco(process)
```


The former will attach to process (```process.TICL```)  anything you need to run **TICL Only**, and everything it needs. You just need to add it to the schedule, alone.
The latter will attach to process a TICL sequence (```process.TICL```) that you can then add to the final schedule, together with the full Reconstruction. I usually run this when I want to make some comparison between TICL products and what we currently have in release, i.e., multiclusters.

I hope the instructions are clear. Sorry if this requires some extra work to test, but I was not sure how to push TICL in w/o any link with the official reco and yet with something to use to quickly test it. This is what I came up with.

**Comment by rovere** [comment]
Dear all,
10_6_0 is out and I think it's really a pity that this PR did not make it!
Could we please have this integrated? There is development ongoing on top of this branch and I'd like our customers to have this available within the release.
@fabiocos @cseez @malgeri

**Comment by slava77** [comment]
@cmsbuild please test

File: RecoHGCal/TICL/plugins/TrackstersProducer.cc [hunk 39]
```diff
@@ -0,0 +1,103 @@
...
+  desc.add<edm::InputTag>("hgcal_layerclusters", edm::InputTag("hgcalLayerClusters"));
+  desc.add<edm::InputTag>("filtered_layerclusters_mask",
+                          edm::InputTag("FilteredLayerClusters", "iterationLabelGoesHere"));
+  desc.add<edm::InputTag>("original_layerclusters_mask",
+                          edm::InputTag("hgcalLayerClusters", "InitialLayerClustersMask"));
+  desc.add<int>("algo_verbosity", 0);
+  desc.add<double>("min_cos_theta", 0.915);
+  desc.add<double>("min_cos_pointing", -1.);
+  desc.add<int>("missing_layers", 0);
+  desc.add<int>("min_clusters_per_ntuplet", 10);
```
slava77: the algo config parts better be in its own pset

RecoHGCal/TICL/plugins/TrackstersProducer.cc
```diff
@@ -0,0 +1,103 @@
...
+  edm::EDGetTokenT<std::vector<reco::CaloCluster>> clusters_token_;
+  edm::EDGetTokenT<std::vector<std::pair<unsigned int, float>>> filtered_layerclusters_mask_token_;
+  edm::EDGetTokenT<std::vector<float>> original_layerclusters_mask_token_;
+
+  std::unique_ptr<PatternRecognitionAlgoBase> myAlgo_;
+};
+DEFINE_FWK_MODULE(TrackstersProducer);
+
+TrackstersProducer::TrackstersProducer(const edm::ParameterSet& ps)
+    : myAlgo_(std::make_unique<PatternRecognitionbyCA>(ps)) {
```
slava77: is there a plan to make this concrete PR algo selection configurable?
If no, the producer name is inappropriate and better have ```ByCA``` in the name.

RecoHGCal/TICL/plugins/TrackstersProducer.cc
```diff
@@ -0,0 +1,103 @@
...
+  desc.add<edm::InputTag>("filtered_layerclusters_mask",
+                          edm::InputTag("FilteredLayerClusters", "iterationLabelGoesHere"));
+  desc.add<edm::InputTag>("original_layerclusters_mask",
+                          edm::InputTag("hgcalLayerClusters", "InitialLayerClustersMask"));
+  desc.add<int>("algo_verbosity", 0);
+  desc.add<double>("min_cos_theta", 0.915);
+  desc.add<double>("min_cos_pointing", -1.);
+  desc.add<int>("missing_layers", 0);
+  desc.add<int>("min_clusters_per_ntuplet", 10);
+  descriptions.add("tracksters", desc);
```
slava77: if this ```TrackstersProducer``` module is going to be used in multiple iterations, apparently there is not going to be a unique ```tracksters``` useful for all users.
So, it's more appropriate to stay with the default module name ```trackstersProducer```  for the auto-generated cfi and then clone to an appropriate iteration-driven name in the actual workflow

RecoHGCal/TICL/plugins/MultiClustersFromTrackstersProducer.cc
```diff
@@ -0,0 +1,92 @@
...
+
+void MultiClustersFromTrackstersProducer::fillDescriptions(
+    edm::ConfigurationDescriptions& descriptions) {
+  // hgcalMultiClusters
+  edm::ParameterSetDescription desc;
+  desc.add<edm::InputTag>("Tracksters", edm::InputTag("Tracksters", "TrackstersByCA"));
+  desc.add<edm::InputTag>("LayerClusters", edm::InputTag("hgcalLayerClusters"));
+  desc.add<std::string>("label", "MultiClustersFromTracksterByCA");
+  desc.addUntracked<unsigned int>("verbosity", 3);
+  descriptions.add("multiClustersFromTracksters", desc);
```
slava77: Please use the default class name for a cfi ```multiClustersFromTrackstersProducer```
 because there will be multiple clones of this over the iterations

RecoHGCal/TICL/test/ticl_iterations.py
```diff
@@ -0,0 +1,135 @@
...
+      ClusterFilter = "ClusterFilterBySize",
+      algo_number = 8,
+      max_cluster_size = 2, # inclusive
+      iteration_label = "MIP"
+  )
+
+  process.TrackstersMIP = tracksters.clone(
+      original_layerclusters_mask = cms.InputTag("hgcalLayerClusters", "InitialLayerClustersMask"),
+      filtered_layerclusters_mask = cms.InputTag("FilteredLayerClustersMIP", "MIP"),
+      algo_verbosity = 0,
```
slava77: this is already defined in the cloned original, no need to repeat

RecoHGCal/TICL/test/ticl_iterations.py
```diff
@@ -0,0 +1,135 @@
...
+      missing_layers = 3,
+      min_clusters_per_ntuplet = 15,
```
slava77: here and in all instances the value 15  is used, while the default is set to 10 in the fillDescriptions.
Please define a more appropriate default in the fillDescriptions and then there is no need to repeat the value in the clones

RecoHGCal/TICL/test/ticl_iterations.py
```diff
@@ -0,0 +1,135 @@
...
+
+  process.MultiClustersFromTrackstersMIP = multiClustersFromTracksters.clone(
+      label = "MIPMultiClustersFromTracksterByCA",
+      Tracksters = cms.InputTag("TrackstersMIP", "TrackstersByCA")
+  )
+
+  process.FilteredLayerClusters = FilteredLayerClusters.clone(
+      algo_number = 8,
+      iteration_label = "algo8",
+      LayerClustersInputMask = cms.InputTag("TrackstersMIP")
```
slava77: type specs are not needed for existing parameters

RecoLocalCalo/HGCalRecAlgos/interface/HGCal3DClustering.h
```diff
@@ -23,17 +23,30 @@ class HGCal3DClustering
...
+   es(0),
+   zees(2*(maxlayer+1),0.),
+   clusterTools(std::make_unique<hgcal::ClusterTools>(conf,sumes))
+   {
+   }
+
+
+   HGCal3DClustering(const edm::ParameterSet& conf, edm::ConsumesCollector& sumes) :
+    radii(conf.getParameter<std::vector<double>>("multiclusterRadii")),
+    minClusters(conf.getParameter<unsigned>("minClusters")),
```
slava77: this can be replaced by 
```c++
: HGCal3DClustering(conf, sumes, conf.getParameter<std::vector<double>>("multiclusterRadii"),
            conf.getParameter<unsigned>("minClusters") {}
```

slava77 (commented):
thanks to the release of 10_6_0, there was finally time to wrap up with the review here.
I'm hoping that for the next updates in TICL the delay will not exceed a few days.

RecoHGCal/TICL/plugins/ClusterFilterByAlgoOrSize.h
```diff
@@ -0,0 +1,42 @@
...
+        max_cluster_size_(ps.getParameter<int>("max_cluster_size")) {}
+  ~ClusterFilterByAlgoOrSize() override {};
+
+  std::unique_ptr<std::vector<std::pair<unsigned int, float>>> filter(
+      const std::vector<reco::CaloCluster>& layerClusters,
+      const std::vector<std::pair<unsigned int, float>>& availableLayerClusters) const override {
+    auto filteredLayerClusters = std::make_unique<std::vector<std::pair<unsigned int, float>>>();
+    for (auto const& cl : availableLayerClusters) {
+      if (layerClusters[cl.first].algo() == algo_number_ ||
+          layerClusters[cl.first].hitsAndFractions().size() <= max_cluster_size_)
```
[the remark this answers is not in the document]
rovere: Ciao @slava77 ,
I'm not sure I understand your suggestion: do you mean moving the loop in the base-class and call an ad-hoc filter on each CaloCluster?

RecoHGCal/TICL/plugins/PatternRecognitionbyCA.cc
```diff
@@ -0,0 +1,83 @@
...
+#include "PatternRecognitionbyCA.h"
+
+void PatternRecognitionbyCA::fillHistogram(
+    const std::vector<reco::CaloCluster> &layerClusters,
+    const std::vector<std::pair<unsigned int, float>> &mask) {
+  if (algo_verbosity_ > None) {
+    LogDebug("HGCPatterRecoByCA") << "filling eta/phi histogram per Layer" << std::endl;
+  }
+  for (auto &m : mask) {
+    auto lcId = m.first;
```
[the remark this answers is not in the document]
rovere: In the long term, we will introduce a threshold depending on the ID results (probability). At present, if it's in the mask it's used.

RecoHGCal/TICL/plugins/ClusterFilterByAlgoOrSize.h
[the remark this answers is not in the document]
slava77: yes, because the loop and related setup code has to be repeated in all derived variants of the ClusterFilterBase.

RecoHGCal/TICL/test/ticl_iterations.py
rovere: Actually, 10 is a more sound default wrt 15 and recent studies seem to point into this direction. We will update these redefinitions in coming PRs, or eventually set a more appropriate default as soon as we have a better understanding.

RecoHGCal/TICL/plugins/ClusterFilterByAlgoOrSize.h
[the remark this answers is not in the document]
rovere: yes, indeed, but maybe this loop w/o virtuals will vectorize?

rovere:
@slava77 I did not touch any of the PSet business because I wanted to be sure and converge on the ClsuterFilter design. Also, in case, this could be done later on an improvement.

RecoHGCal/TICL/plugins/ClusterFilterByAlgoOrSize.h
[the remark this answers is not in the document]
slava77: I'm not sure what is implied by lack of virtuals.
Simple function calls, even via a vtable can be vectorizable (or is it impossible?); access to layerClusters via a generally random list of indices does not look vectorizable by itself.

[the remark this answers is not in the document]
rovere: @slava77 I'm not sure if there can be vectorization with vtable in between.
In any case, forcing the loop in the base will fix the logic for all derived (i.e. no comparison between 2 elements will ever be possible). I'm not sure this will ever be required, but I think this is good enough for the release.

rovere:
@slava77 I'll be out of reach up to the middle of next week. I tested again TICL on a private sample and the results are still the same even after all the changes. So, from my side, if everyone is happy with the current state, it's good to go.

slava77:
> @slava77 I did not touch any of the PSet business because I wanted to be sure and converge on the ClsuterFilter design. Also, in case, this could be done later as an improvement.

[I didn't check the updates yet]
The PSet business may requires some structural updates. Better start doing it upfront before other algorithms show up within the framework.

slava77:
@cmsbuild please test

slava77:
Looking at https://github.com/cms-sw/cmssw/pull/26295#issuecomment-493138358 and 88 commits in this PR already,
may I suggest a rebase with some squashing, unless you are particularly attached to the history of the developments.

rovere:
@slava77 indeed some squashing would be nice to have.
Sorry, @amartelli I guess you'll have to "rebase" again after I do that.

rovere:
@slava77 I think this is better. I tried to keep some history, at a reasonable level.

slava77:
@cmsbuild please test

rovere:
@slava77: if deemed necessary, I can try and squeeze it even more...

jfernan2:
+1 from DQM point of view

RecoHGCal/TICL/plugins/PatternRecognitionbyCAConstants.h
```diff
@@ -0,0 +1,15 @@
+#ifndef __RecoHGCal_TICL_PRbyCAConstants_H__
+#define __RecoHGCal_TICL_PRbyCAConstants_H__
+
+namespace ticl {
+  namespace patternbyCA {
+    constexpr int nEtaBins = 34;
+    constexpr int nPhiBins = 126;
+    constexpr int nLayers = 104;
+    typedef std::array<std::array<std::vector<unsigned int>, nEtaBins*nPhiBins>, nLayers>
+      tilePatternRecognitionByCA;
```
slava77: typenames with upper case may be better.
Is there a good reason to maintain redundancy in the names? In the currently named namespace, this can be simply ```Tile``` or perhaps more descriptive ```LayerTiles```.

RecoHGCal/TICL/plugins/HGCGraph.h
```diff
@@ -0,0 +1,40 @@
...
+#ifndef __RecoHGCal_TICL_HGCGraph_H__
+#define __RecoHGCal_TICL_HGCGraph_H__
+
+#include <vector>
+#include "PatternRecognitionbyCAConstants.h"
+#include "HGCDoublet.h"
+
+class HGCGraph {
+ public:
+  void makeAndConnectDoublets(const ticl::patternbyCA::tilePatternRecognitionByCA &h,
```
slava77: Is HGCGraph really specific to ```PatternRecognitionbyCA``` class implementation?
Would the other methods deriving from PatternRecognitionAlgoBase have another ```makeAndConnectDoublets``` with a different signature or is the type behind ```ticl::patternbyCA::tilePatternRecognitionByCA``` supposed to be generic?
If it's generic, it is named incorrectly and should go to a more common place.

RecoHGCal/TICL/plugins/PatternRecognitionbyCAConstants.h
```diff
@@ -0,0 +1,15 @@
+#ifndef __RecoHGCal_TICL_PRbyCAConstants_H__
+#define __RecoHGCal_TICL_PRbyCAConstants_H__
+
+namespace ticl {
+  namespace patternbyCA {
```
slava77: just for syntax, this can just be ```namespace ticl::patternbyCA {```

namespaces should be in all lower case.
Will these constants and typedef ever be used without PatternRecognitionByCA.h already included?
If no, perhaps it's more expressive to put them inside the class PatternRecognitionByCA.
OTOH, it seems like the use case covers more than just PatternRecognitionByCA and perhaps a less dual naming convention is better.

RecoHGCal/TICL/plugins/TrackstersProducer.cc
```diff
@@ -0,0 +1,103 @@
...
+ public:
+  TrackstersProducer(const edm::ParameterSet &);
+  ~TrackstersProducer() override {}
+  static void fillDescriptions(edm::ConfigurationDescriptions &descriptions);
+
+  void produce(edm::Event &, const edm::EventSetup &) override;
+
+ private:
+  edm::EDGetTokenT<std::vector<reco::CaloCluster>> clusters_token_;
+  edm::EDGetTokenT<ticl::hgcalClusterFilterMask> filtered_layerclusters_mask_token_;
```
slava77: BTW, ```using namespace ticl``` is OK in .cc. The current choice is fine though.

RecoHGCal/TICL/plugins/TrackstersProducer.cc
```diff
@@ -0,0 +1,103 @@
...
+  edm::EDGetTokenT<std::vector<float>> original_layerclusters_mask_token_;
+
+  std::unique_ptr<PatternRecognitionAlgoBase> myAlgo_;
+};
+DEFINE_FWK_MODULE(TrackstersProducer);
+
+TrackstersProducer::TrackstersProducer(const edm::ParameterSet& ps)
+    : myAlgo_(std::make_unique<PatternRecognitionbyCA>(ps)) {
+  clusters_token_ = consumes<std::vector<reco::CaloCluster>>(
+      ps.getParameter<edm::InputTag>("hgcal_layerclusters"));
```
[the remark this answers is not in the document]
slava77: > I suggest to reduce verbosity
> `clusters`, `filtered_mask`, and `original_mask`

I guess this didn't seem like an improvement

RecoLocalCalo/HGCalRecAlgos/interface/HGCal3DClustering.h
slava77: > this can be replaced by
> 
> ```c++
> : HGCal3DClustering(conf, sumes, conf.getParameter<std::vector<double>>("multiclusterRadii"),
>             conf.getParameter<unsigned>("minClusters") {}
> ```

please use this to reduce code replication

slava77 (commented):
@rovere 
thank you for the quick follow up to the last iteration of the reco review.

I think that some newly introduced constants and typedefs are somewhat dubious, at least at a quick glance. Please check and update if my simplified naming suggestions make sense or unless there is a strong preference to keep somewhat redundant name patterns.

RecoHGCal/TICL/plugins/PatternRecognitionbyCAConstants.h
rovere: Indeed they are used in the HGCGraph.

RecoHGCal/TICL/plugins/HGCGraph.h
rovere: Ciao @slava77 
at the current state of development, I cannot really say if this will be the only "tool" used by the CA and, also, if there will be other PR other than CA.

rovere:
@slava77 I squashed again a bit the history that now should read more "linearly" w/o exposing the trials and errors that inevitably were done in the development phase.
I also took care of most of your comments and, for the ones I did not implement, I commented directly into the code in question.
Thanks for the detailed review: I believe the code reads much better now.

slava77:
@cmsbuild please test

slava77:
+1

for https://github.com/cms-sw/cmssw/pull/26295	67e829d
- code changes are somewhat in line with the PR description and the follow up review.
- jenkins tests pass and comparisons with the baseline show no differences (the new code is not used in the standard workflows)

kpedro88:
+upgrade

RecoHGCal/TICL/interface/Trackster.h
```diff
@@ -0,0 +1,29 @@
+// Author: Felice Pantaleo - felice.pantaleo@cern.ch
+// Date: 09/2018
+// Copyright CERN
```
fabiocos: @tommasoboccali, @klute, @dpiparo although our code is already full of "copyright" statements, I wonder about the compatibility of this lines and https://github.com/cms-sw/cmssw/blob/master/LICENSE

perrotta:
@cmsbuild please test

RecoHGCal/TICL/interface/Trackster.h
tommasoboccali: I have to admit i am not really passionate with CP discussion, but indeed the top level file 

https://github.com/cms-sw/cmssw/blob/master/NOTICE

already says

Copyright 2019 CERN¹ for the benefit of the CMS Collaboration²

so it should be a catch-all for the code

slava77:
+1

for https://github.com/cms-sw/cmssw/pull/26295	60bc3ea 
- changes since the last reco signoff were trivial for the code (removal of "copyright" comments)

kpedro88:
+upgrade

RecoHGCal/TICL/interface/Trackster.h
fabiocos: @tommasoboccali thanks, these discussions are not passionating but I have to check what we store in our repositories, now after @rovere last update there is not reason to further delay the integration of this code

fabiocos:
+1

Issue #25751, opened 2019-01-23
Do MET uncertainties have a bug when they scale and shift down electrons?


perrotta:
During the review of PR #25699 it was noticed that when adding extra electrons which are supposedly not considered by the PF, the MET uncertainties were instead somehow affected, see e.g.
![image](https://user-images.githubusercontent.com/4069749/51628836-5d992480-1f46-11e9-9aec-4f2db56f2b2a.png)

The relevant part of the discussion starts here: https://github.com/cms-sw/cmssw/pull/25699#issuecomment-456166898

@Sam-Harper do you have a clearer idea in which part of the code the bug is located?
@eioannou can you please have a look from the MET side?

perrotta:
assign reconstruction

eioannou:
> During the review of PR #25699 it was noticed that when adding extra electrons which are supposedly not considered by the PF, the MET uncertainties were instead somehow affected, see e.g.
> ![image](https://user-images.githubusercontent.com/4069749/51628836-5d992480-1f46-11e9-9aec-4f2db56f2b2a.png)
> 
> The relevant part of the discussion starts here: [#25699 (comment)](https://github.com/cms-sw/cmssw/pull/25699#issuecomment-456166898)
> 
> @Sam-Harper do you have a clearer idea in which part of the code the bug is located?
> @eioannou can you please have a look from the MET side?

@perrotta : I wiil have a look at it.

Sam-Harper:
The bug is simple and trivial for jetMET to fix. The MET uses as input particle flow electrons. However the met uncertainties are computed by scaling up and down all electrons, including electrons which are not used in the MET calculation. 

GsfElectron::passingPflowPreselection()==true should be enforced in the selector. Unless I miss understand what this is doing.

Sam-Harper:
Also not a super huge deal of course but still technically a bug I believe

eioannou:
I confirm that there is a bug which is related to the computation of the met uncertainties. They are computed by scalling up and down all electrons. A possible solution is to edit the two files below in order to scale up and down only PF electrons.
https://github.com/cms-sw/cmssw/blob/17b8235c793832707138ee5eeb04a427de5fda02/PhysicsTools/PatUtils/plugins/ShiftedParticleProducer.cc#L41-L57
[...]

PR #26302, opened 2019-03-30
Fixing the calculation of MET Uncertainties by using PFMuons, PFElectrons, PFTaus, and PFPhotons

eioannou:
This PR is submitted in order to fix a bug in the calculation of MET Uncertainties. MET Uncertainties must be calculated by using PF Electrons, PF Photons, PF Taus, and PF Muons only.

The additional snippet of the code finds PF objects  and stores them into the collections and then, those collections are used for the calculations of MET uncertainties

This PR is related to the issue #25751

### DIFF
--- a/PhysicsTools/PatUtils/python/tools/runMETCorrectionsAndUncertainties.py
+++ b/PhysicsTools/PatUtils/python/tools/runMETCorrectionsAndUncertainties.py
@@ -829,6 +829,66 @@ class RunMETCorrectionsAndUncertainties(ConfigToolBase):
         #===================================================================================
         # energy shifts
         #===================================================================================
+        # PFMuons, PFElectrons, PFPhotons, and PFTaus will be used 
+        # to calculate MET Uncertainties.
+        #===================================================================================
+        #--------------
+        # PFElectrons :
+        #--------------
+        pfElectrons = cms.EDFilter("CandPtrSelector",
+                                   src = electronCollection,
+                                   cut = cms.string("pt > 5 && isPF && gsfTrack.isAvailable() && gsfTrack.hitPattern().numberOfLostHits(\'MISSING_INNER_HITS\') < 2")
+                                   )
+        addToProcessAndTask("pfElectrons"+postfix, pfElectrons, process, task)
+        metUncSequence += getattr(process, "pfElectrons"+postfix)
+        #--------------------------------------------------------------------
+        # PFTaus :
+        #---------
+        pfTaus = cms.EDFilter("PATTauRefSelector",
+                              src = tauCollection,
+                              cut = cms.string('pt > 18.0 & abs(eta) < 2.6 & tauID("decayModeFinding") > 0.5 & isPFTau')
+                              )
+        addToProcessAndTask("pfTaus"+postfix, pfTaus, process, task)
+        metUncSequence += getattr(process, "pfTaus"+postfix)
+        #---------------------------------------------------------------------
+        # PFMuons :
+        #----------
+        pfMuons = cms.EDFilter("CandPtrSelector",
+                               src = muonCollection,
+                               cut = cms.string("pt > 5.0 && isPFMuon && abs(eta) < 2.4")
+                               )
+        addToProcessAndTask("pfMuons"+postfix, pfMuons, process, task)
+        metUncSequence += getattr(process, "pfMuons"+postfix)
+        #---------------------------------------------------------------------
+        # PFPhotons :
+        #------------
+        if self._parameters["Puppi"].value or not self._parameters["onMiniAOD"].value:
+            cutforpfNoPileUp = cms.string("")
+        else:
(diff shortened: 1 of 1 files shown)


perrotta:
please test
(thank you @eioannou )

perrotta:
@Sam-Harper , FYI

perrotta:
@eioannou 

Here you re-define pfElectrons, pfPhotons, pfMuons, pfTaus, and then you use them to calculate the MET uncertainties.

Isn't it quite error prone doing so? What happen if any of those objects gets re-defined for the MET calculation sometime in the future? Still you will end up with having decoupled the calculation of the MET and the one of its its error, as it was the origin of the issue pointed out in #25751.

The original python configurations are quite convoluted and I can hardly reconnect all pieces... However, I would say that the real fix would be to reuse the very same objects that entered the calculation of the MET for the calculation of the MET uncertainties, which means import'ing for the errors the same configuration which are used (elsewhere) for the MET. Is there any technical hindrance in doing so?

eioannou:
If I am correct, the MET is computed in this config file
https://github.com/cms-sw/cmssw/blob/e2a0510e0d05fe033646399e49596349fc6e8ef9/RecoMET/METProducers/python/PFMET_cfi.py#L1-L11

So, the collection which is used to compute the MET is the "particleFlow".  However, if I try to compute the MET uncertainties by using only the particleFlow, I guess, I have to modify the whole code which calculates the met uncertainties in the MET tool. So,  an idea was, before the calculation of the met uncertainties, to create new collections of PF Objects only.

The current code uses all objects (Muons, Electrons, Taus, Photons) to compute the met uncertainties which it is wrong. So, my idea was to create collections which include only PF objects before the calculation of met uncertainties.

perrotta:
@eioannou this PR modifies in a significant way the errors on MET.
If this is the result of a bug fix, as it should, the modified errors must replace the previous, bugged ones.
Given the significant differences, however, I would like the new errors get endorsed by JetMET POG: please post a presentation that you made (or will make) at the POG which can ceritfy that the new errors have to be preferred.

eioannou:
I will prepare a presentation and will include the link here when it is ready to be presented at MET meeting.

perrotta:
@eioannou 
This PR is ready to get accepted, and enter the UL 10_6 release, so that the amended MET errors can replace the bugged ones.
Please just send us a link to that presentation (just one or a few slides of comparisons can be enough), so that a reference of the improved uncertainities will remain here: then I will sign it for reco.

perrotta:
ping @eioannou

eioannou:
Hi @perrotta, I am trying to prepare the presentation about this PR, by performing some validation tests, and I am going to present the results in the next JETMET meeting. Sorry for the delay.

perrotta:
type bugfix

perrotta:
@eioannou it would be nice to include this bugfix in 10_6. I don't think there will be any JetMET meeting by tomorrow, the deadline set for the the latest open pre of 10_6: if you have any validation ready, and if you are confident enough that it does correspond to what the MET uncertainties must be, please post it here.

eioannou:
> @eioannou it would be nice to include this bugfix in 10_6. I don't think there will be any JetMET meeting by tomorrow, the deadline set for the the latest open pre of 10_6: if you have any validation ready, and if you are confident enough that it does correspond to what the MET uncertainties must be, please post it here.

I will try to post the plots here today.

fabiocos:
@eioannou as @perrotta says this is a fix needed for UL, and we should try to converge asap on it

eioannou:
I have performed validation tests for MET uncertainties of Type1 MET. The results are shown below. The green lines correspond to MET Uncertainties which are calculated by using PF particles only and the red lines stand for the MET uncertainties which are calculated by using all particles (as it is done until now). The plots show that the changes are minor between of two calculations, however, the correct method to compute the MET uncertainties is to use only PF particles in the calculations. 

Validation tests were performed by using a DY sample of 50K events (DYJetsToLL_M-50_TuneCP5_13TeV-amcatnloFXFX-pythia8).

DMET_ElectronEnDown
DMET_ElectronEnUp
DMET_MuonEnDown
DMET_MuonEnUp
DMET_PhotonEnDown
DMET_PhotonEnUp
DMET_TauEnDown
DMET_TauEnUp

perrotta:
+1
- This fixes a bug in the calculation of the MET uncertainties, where those uncertainties were calculated by using all particles, while now only the very same PF particles used for the calculation of the MET itself are considered.
- It solves issue #25751
- Jenkins tests pass and show differences only in the MET uncertainties, as expected: those differences have been validated as reported in https://github.com/cms-sw/cmssw/pull/26302#issuecomment-490574225 and found corresponding to the correct expectations for them

fabiocos:
+1

fabiocos:
merge

santocch:
+1

PR #26306, opened 2019-04-01
speed up Mixins.py

davidlange6:
#### PR description:

As reported by @vininn, start up time for MC jobs in 10_6_0_pre2 is due to a python3 compatibility change. This PR is to improve that change.

### DIFF
--- a/FWCore/ParameterSet/python/Mixins.py
+++ b/FWCore/ParameterSet/python/Mixins.py
@@ -656,9 +656,8 @@ class _ValidatingParameterListBase(_ValidatingListBase,_ParameterTypeBase):
         return (converter(x).value() for x in strings)
 
 def saveOrigin(obj, level):
-    frame = inspect.stack()[level+1]
-    if isinstance(frame,tuple): frame=frame[0] #python3 changes this to a tuple where the first thing is the frame
-    fInfo=inspect.getframeinfo(frame)
+    import sys
+    fInfo = inspect.getframeinfo(sys._getframe(level+1))
     obj._filename = fInfo.filename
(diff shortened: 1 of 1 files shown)


davidlange6:
please test

Dr15Jones:
+1

VinInn:
> hi @Dr15Jones - I was hoping we ran the timing reports in the PR tests, but I guess we do not, so I don't have great numbers - it appears to save 2-3 minutes in the total time for step3 of 10224 for example.

exactly what we expect!

fabiocos:
@davidlange6 we do have the TimeMemorySummary activated in IBs, the problem for CPU is that it depends on the machine and on its load (although the kind of spectacular effect looked here should be anyway appreciable)

fabiocos:
+1

fabiocos:
not the same machine on which I did previous checks but for step3 of wf 10824.0 I now see:
```
TimeReport> Time report complete in 262.775 seconds
 Time Summary: 
 - Min event:   1.60077
 - Max event:   25.6009
 - Avg event:   5.36617
 - Total loop:  86.6168
 - Total init:  163.666
 - Total job:   262.775
 - EventSetup Lock:   0.000497103
 - EventSetup Get:   41.4004
 Event Throughput: 0.115451 ev/s
 CPU Summary: 
 - Total loop:  77.3087
 - Total init:  186.999
 - Total extra: 0
 - Total job:   276.078
 Processing Summary: 
 - Number of Events:  10
 - Number of Global Begin Lumi Calls:  1
 - Number of Global Begin Run Calls: 1
```
as expected from David's checks

PR #26320, opened 2019-04-02
Add Alignment Records for GEM Geometry

watson-ij:
#### PR description:

This PR adds alignment records for the GEM, and adds code to the GEMGeometryBuilder to allow the alignment to be applied. The code was written by following that used for the CSC.

#### PR validation:

A test of the alignment showed that the GEM geometry has left-handed local coordinate systems, which are incompatible with the standard Alignment records, which store the detector rotations as Euler angles from the global coordinate system. I removed the explicit axis flip which was responsible for this, and reran an upgrade sample and checked that the GEM rechits and segments were still being built properly.

By default, the alignment is off (an applyAlignment parameter is introduced in the GEM geometry builder), to avoid picking up currently non-existing alignment records. This PR is a first step to allow alignment to be applied to GEM, and while we plan to build on this by writing GEM alignment code, for now the alignment will not be used. Therefore, the PR is expected to produce no changes in output.

@jshlee @kpedro88


kpedro88:
please test

Geometry/GEMGeometryBuilder/plugins/GEMGeometryESModule.cc
```diff
@@ -20,13 +20,23 @@
...
+
 #include <memory>
 
 using namespace edm;
 
 GEMGeometryESModule::GEMGeometryESModule(const edm::ParameterSet & p)
+  : alignmentsLabel(p.getParameter<std::string>("alignmentsLabel"))
 {
   useDDD = p.getParameter<bool>("useDDD");
+  applyAlignment = p.getParameter<bool>("applyAlignment");
```
kpedro88: all of the variables could be initialized in the constructor initialization list

kpedro88:
@watson-ij there are some differences in muon reco quantities in Phase 2 workflows. It seems likely that these differences are caused by the change in the coordinate system. Can you check it carefully?

watson-ij:
@kpedro88, the differences are mostly from a single workflow, 20034.0, so I ran this on a nightly from 04-01, and with and without my changes it gave the reference distribution. I then tried a 04-07 nightly and with or without my changes, it produced the "incorrect" result. I'm not sure if its possible another commit caused changes from the reference without updating the reference? I also fixed a merge conflict with a recent commit, and implemented the changes you suggested in the constructor. Could we try rerunning the comparison checks?

kpedro88:
please test

cvuosalo:
Geometry is ready to approve if the comparison test differences can be explained.

watson-ij:
Okay, I can see the differences in the workflow now when I take the last nightly with and without my changes. I'll see if I can debug whats going on.

pohsun:
No significant flaw in current commit. Alca is pending for updates.

kpedro88:
@watson-ij what is the status of this PR?

watson-ij:
@kedpro88, I've been drilling into the code to see the changes, and
think I know the reason, but it would be good for a tracking expert to
cross-check.

The plot differences are mostly from a single standalone displaced
muon missing after my PR, but another GEMMuon having the standalone
flag on that previously wasn't.

After tracing through several systems, I found that the PR changes the
results of the kalman update step in the even chambers which have
their z-flipped (in particular, I checked the muon reco systems, such
as the fast detector hit finders, and none of these is changed by the
PR).

For example, first muon passing through GEM in event 10 has these
local parameters after the Kalman update (before the update they are
identical):

Before PR:
local parameters (q/p,v',w',v,w)
     0.355115   -0.0415613     0.241321      8.87442     0.250691

After PR:
local parameters (q/p,v',w',v,w)
    -0.430425    0.0415613    -0.241321      8.87442     0.250691

The v'=dv/du (u being along the local z), w' sign flip makes sense,
since we flipped the z-axis, so only the q/p is different. Looking at
what passes through the KFUpdator step, the measurement parameters are
all the same, the difference is coming from the
BasicTrajectoryState::createLocalErrorFromCurvilinearError() which
uses JacobianCurvilinearToLocal curv2Loc to get the jacobian and then
uses it to translate the curvilinear covariance matrix of the track to
the local parameters.

The paper describing this transformation is here:
https://doi.org/10.1016/j.nima.2006.07.032 and in it, they state
several times that they are defining axes to be right-handed, so I
think that using left-handed coordinates is wrong and only
right-handed coordinates should be used, so this PR (which is changing
all local frame to be right-handed) would fix that issue.

I was trying to follow through the derivation in this particular case
to confirm this, but getting bogged down in the math. So, it would be
good for a tracking expert to weigh in if my understanding here is
correct, and any additional suggestions would be welcome.

kpedro88:
@watson-ij thanks for investigating this. I'll tag some tracking experts here, since I am not one: @VinInn @makortel @JanFSchulte @mtosi, can you check if the above explanation makes sense?

kpedro88:
ping @ tracking experts (@VinInn @makortel @JanFSchulte @mtosi)

kpedro88:
ping @ tracking experts (@VinInn @makortel @JanFSchulte @mtosi)

kpedro88:
+upgrade
seeing no objection from tracking experts, the proposed explanation for observed changes looks reasonable to me

pohsun:
+1

pohsun:
+1

ggovi:
+1

fabiocos:
please test

fabiocos:
@cvuosalo could you please come back to this PR? Your signature is the last missing

cvuosalo:
+1

fabiocos:
+1

PR #26321, opened 2019-04-02
Fix thread safety issue in FEDNumbering

Dr15Jones:
#### PR description:

Removed the thread unsafe initialization of static member data. Change to initialize the data either at compile time (via the use of constexpr) or at shared library load time.

#### PR validation:

The code compiles and a new unit test was added to check the code.


Dr15Jones:
please test

Dr15Jones:
please test

DataFormats/FEDRawData/src/FEDNumbering.cc
```diff
@@ -3,180 +3,357 @@
...
+    return in;
+  }
+
+  constexpr std::array<int, FEDNumbering::MAXFEDID+1> initFromIndex() {
+    std::array<int, FEDNumbering::MAXFEDID+1> fromIndex= {{0}};
+    int i = 0;
+    int index = 0;
+    for(i=0; i< FEDNumbering::lastFEDId(); i++)
+      fromIndex[i] = index; //empty
+    ++index;
```
slava77: can't the ```fromString```  be filled here as well?
... and also all of this code merged together with the definition of ```in ``` in ```initIn```.

slava77 (commented):
The coding pattern looks too error-prone.
Instead of one set covers all in ```init```, this PR now requires 5 places to be edited.

DataFormats/FEDRawData/src/FEDNumbering.cc
Dr15Jones: It can not be made `constexpr`. All the `constexpr` functions are run at compile time with the arrays the created filled at compile time. There is no runtime. One can not fill std::string at compile time.

I added a very extensive test which gets run at each pull request which makes sure the values are correct.

slava77: > I added a very extensive test which gets run at each pull request which makes sure the values are correct.

this only tests the code that exists now; there is no mechanism to require/fail the unit test if a new min/max pair is added.

mommsen: I wonder if we could get rid of the string handling altogether. The method FEDNumbering::fromDet is not used anywhere in CMSSW according to http://cmslxr.fnal.gov/ident?_i=fromDet

slava77: this would be much easier.
Thanks for the proposal.

Dr15Jones: Is the consensus for the string handling to be removed? If so, I will make that change.

slava77: daq and reco are apparently OK with this.

mommsen: IMHO, it can be removed.

Dr15Jones:
please test

slava77:
+1

for https://github.com/cms-sw/cmssw/pull/26321	9073dac

mommsen:
+1
Thanks @Dr15Jones for taking care of improving this code!

fabiocos:
@Dr15Jones @smuzaffar the reported compilation warning looks real, and I can easily it fix in a separate PR. But it looks to me that the fednum variable is unused regardless of this PR, so why is it popping out only now? The change introduced by this PR is that FEDNumbering has no more private data members.

Dr15Jones:
@fabiocos I promise to fix the new compilation warning in another pull request. No good guess as to why it only appears now.

fabiocos:
+1

Issue #26154, opened 2019-03-12
Crash in bParking processing


fabiocos:
A problem with bParking code has been reported by PdmV. This issue has been initially observed during CRAB-based private tests, and were reproducible when running in CRAB but not when running locally.  The problem is:

```
----- Begin Fatal Exception 11-Mar-2019 20:27:43 UTC-----------------------
An exception of category 'FatalRootError' occurred while
[0] Constructing the EventProcessor
[1] Constructing module: class=LowPtGsfElectronIDProducer label='lowPtGsfElectronID'
Additional Info:
[a] Fatal Root Error: @SUB=TXMLEngine::ParseFile
Unexpected end of xml file
----- End Fatal Exception -------------------------------------------------
```

initially reported in https://hypernews.cern.ch/HyperNews/CMS/get/computing-tools/4440.html and https://indico.cern.ch/event/790963/contributions/3306140 .
It looks present in 10_2_X, not in cycles after 10_3_X, starting with merges on top of 10_2_9.

fabiocos:
@prebello could you please point us to a failing workflow/initial settings?

fabiocos:
@mverzett @bainbrid FYI

prebello:
> @prebello could you please point us to a failing workflow/initial settings?

sure for instance
full set of injected relvals are at
https://dmytro.web.cern.ch/dmytro/cmsprodmon/workflows.php?campaign=CMSSW_10_2_12__fullsim_noPU_2018_bParking-1552246543

details for each step can be found at 
https://cmsweb.cern.ch/reqmgr2/fetch?rid=prebello_RVCMSSW_10_2_12SingleElectronPt1000_190310_203726_6599
clicking on the respective step and then "Raw Data" tab

does it help? any further detail?

davidlange6:
previously reported errors were due to special characters in the file (it seems)

The file read here is
RunII_Autumn18_LowPtElectrons_mva_id.xml.gz

is that correct?

mverzett:
@davidlange6 yes, the file is the right one.

> previously reported errors were due to special characters in the file (it seems)

  * Out of curiosity, how can you tell?
  * Why it works locally but not remotely?
  * How do we clean the xml from these characters?

davidlange6:
> On Mar 12, 2019, at 5:20 PM, Mauro Verzetti <notifications@github.com> wrote:
> 
> @davidlange6 yes, the file is the right one.
> 
[...]

PR #26334, opened 2019-04-03
Added missing skipEvents to LHESource

Dr15Jones:
#### PR description:

When the LHESource gained a fillDescriptions validator, that function did not include 'skipEvents' as an allowed parameter.

#### PR validation:

The code was tested by running a configuration which was previously failing do to the validation problem. After the code change, the configuration runs.

### DIFF
--- a/GeneratorInterface/LHEInterface/plugins/LHESource.cc
+++ b/GeneratorInterface/LHEInterface/plugins/LHESource.cc
@@ -205,6 +205,7 @@ LHESource::fillDescriptions(edm::ConfigurationDescriptions& descriptions) {
   edm::ParameterSetDescription desc;
   desc.setComment("A source which reads LHE files.");
   edm::ProducerSourceFromFiles::fillDescription(desc);
+  desc.addUntracked<unsigned int>("skipEvents", 0U)->setComment("Skip the first 'skipEvents' events.");
   descriptions.add("source", desc);
 }
 


Dr15Jones:
please test

fabiocos:
@efeyazgan @qliphy @alberto-sanchez this is the fix addressing the problem find in production, and which would benefit from the addition of a test using and external lhe file as input, could you please have a look at it?

alberto-sanchez:
+1

fabiocos:
+1

prebello:
@Dr15Jones  could you please take a look in https://its.cern.ch/jira/browse/CMSCOMPPR-5297? PdmV would like an update about the issue. Thank you for your help.

fabiocos:
@prebello not sure what is the issue: the fix has been merged and will be part of 10_6_0_pre4...

prebello:
@fabiocos AFAIK the L3 manager has implemented by hand (ad hoc) and resubmit the relval. 
The rsb has failed again with the same error. 
Therefore I was wondering if this would be a fix that would work "ad hoc" or only after merging in the pre4.

Dr15Jones:
@prebello in this context what would 'ad hoc' mean? Does McM have the ability to use code that is from the user and not in a release?

prebello:
> @prebello in this context what would 'ad hoc' mean? Does McM have the ability to use code that is from the user and not in a release?

sorry.. I have meant "by hand". for relvals we don't use McM but runThematrix. We usually hack the source code in Configuration/PyReleaseValidation/python that works.

boudoul:
Hello @prebello - No you cannot have the ability to produce centrally events with runthematrix (or McM it's the same) with a modified cmssw (thankfully! Otherwise all the work you are doing for the  release validation for production would be pointless...) . 
When you are modifying PyReleaseValidation , it is just the python used to prepare the submission, not an actual code from cmssw. 
You already met the same situation with the Bparking ... : https://github.com/cms-sw/cmssw/issues/26154#issuecomment-473872451 - Hope this clarifies - Otherwise I'm sure that @franzoni or  @srimanob can give you more details.  Thanks.

prebello:
@boudoul is perfectly clear. never mind. Thank you.

srimanob:
@prebello 
One option: if you would like to implement a new relval workflows, i.e. pLHE in this case, you can also place the PR now, then do PR test on new additional workflow. That should pick the last code in IB. No need to wait for new release, then add PR again.

PR #26336, opened 2019-04-03
A small fix to dqmMemoryStats.py

andrius-k:
#### PR description:

A small fix to a `dqmMemoryStats.py` tool.

#### PR validation:

The tool was tested locally after applying the fix.

### DIFF
--- a/DQMServices/FileIO/python/DQM.py
+++ b/DQMServices/FileIO/python/DQM.py
@@ -52,7 +52,7 @@ class DQMReader(object):
                 yield (fullname, t_tree.Value, )
 
     def read_objects_root(self):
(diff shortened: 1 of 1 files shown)


andrius-k:
please test

andrius-k:
+1

fabiocos:
@andrius-k for my understanding which observed issue is being addressed by this fix?

schneiml:
@fabiocos the regular expression assumes that all non-histogram `MonitoElement`s (which are parsed using this expression) need to have non-empty values. That is typically true in our DQM output, but it blew up now and then when values are not correctly populated.

fabiocos:
+1

@schneiml ok, thanks for the clarification

PR #26339, opened 2019-04-03
keep hidden valley and dark matter particles in miniAOD

kpedro88:
#### PR description:

Store hidden valley and dark matter particles in the miniAOD pruned gen particles collection. These particles only appear in specific signal models, so the change will have no effect on 99% of MC samples.

Hidden valley PDG IDs are described here: http://home.thep.lu.se/Pythia/pythia82php/HiddenValleyProcesses.php

Dark matter PDG IDs are described here (page 3, "special particles"): http://pdg.lbl.gov/2018/reviews/rpp2018-rev-monte-carlo-numbering.pdf

#### PR validation:

Ran matrix workflow 10024.0, which completed successfully.

Re-ran the workflow, substituting a hidden valley fragment (from https://github.com/cms-sw/genproductions/tree/master/python/ThirteenTeV/SemiVisibleJets), and confirmed that the particles were kept in miniAOD by comparing to the RECO file.

This PR will be backported to active miniAOD releases: 9_4_X and 10_2_X.


kpedro88:
please test

santocch:
+1

perrotta:
+1
- The additions will be only effective for a tiny number of new physocs signal simulations
- Jenkins tests pass

fabiocos:
+1

@kpedro88 for master no problem, for the rest it is important that everybody is aware of the potential consequence (although I assume this will be limited to some specific samples)

fabiocos:
@kpedro88 I am not concerned about reco conveners, but an analysis code written to work with this update will provide different results on past samples. The only argument here is the limited number of users affected by this (and supposedly aware and desiring this change)

PR #26343, opened 2019-04-03
DyT algorithm new thresholds parameterization 

clacaputo:
#### PR description:
This PR introduces a new threshold parameterization for the DyT algorithm. 
Muon POG presentation [here](https://indico.cern.ch/event/797254/contributions/3323400/attachments/1797744/2931260/19_01_25_DYT.pdf).

modified:   RecoMuon/GlobalTrackingTools/interface/DynamicTruncation.h
modified:   RecoMuon/GlobalTrackingTools/src/DynamicTruncation.cc
- new namespace 'dyt_utils' that contains EtaRegion enum
- p_reco and eta_reco private variables
- function to obtain the EtaRegion
- lambda expr for threshold analytic parameterization

modified:   RecoMuon/GlobalTrackingTools/python/GlobalMuonRefitter_cff.py
modified:   RecoMuon/GlobalTrackingTools/python/GlobalTrajectoryBuilderCommon_cff.py
- add bool for selecting dyt version to use
- add PSet for threshold functions parameters

modified:   RecoMuon/GlobalTrackingTools/src/GlobalMuonRefitter.cc
- add few lines in dyt block

The bool for selecting the new DyT is set on `False`

#### PR validation:
Validation performed using 
```
runTheMatrix.py --nThreads=4 -l 10809.0
```
with output:
```
10809.0_SingleMuPt1000+SingleMuPt1000_pythia8_2018_GenSimFull+DigiFull_2018+RecoFull_2018+HARVESTFull_2018+ALCAFull_2018+NanoFull_2018 Step0-PASSED Step1-PASSED Step2-PASSED Step3-PASSED Step4-PASSED Step5-PASSED  - time date Mon Apr  1 17:41:48 2019-date Mon Apr  1 17:17:32 2019; exit: 0 0 0 0 0 0
1 1 1 1 1 1 tests passed, 0 0 0 0 0 0 failed
```

### DIFF
--- a/RecoMuon/GlobalTrackingTools/interface/DynamicTruncation.h
+++ b/RecoMuon/GlobalTrackingTools/interface/DynamicTruncation.h
@@ -107,7 +107,6 @@ class DynamicTruncation {
   void                 useSegment(CSCSegment const &, TrajectoryStateOnSurface const &);
   void                 sort(ConstRecHitContainer&);
   void                 setEtaRegion();
-  static edm::ParameterSetDescription fillDefaultThrsParameters();
 
   ConstRecHitContainer result, prelFitMeas;
   bool useAPE;
--- a/RecoMuon/GlobalTrackingTools/interface/DynamicTruncation.h
+++ b/RecoMuon/GlobalTrackingTools/interface/DynamicTruncation.h
@@ -85,8 +85,6 @@ class DynamicTruncation {
     return dytInfo;
   }
 
-  void                 correctThrByPtAndEta();
-
  private:
 
   void                 compatibleDets(TrajectoryStateOnSurface&, std::map<int, std::vector<DetId> >&);
@@ -100,7 +98,7 @@ class DynamicTruncation {
   void                 updateWithDThits(TrajectoryStateOnSurface&, DTRecSegment4D const &);
   void                 updateWithCSChits(TrajectoryStateOnSurface&, CSCSegment const &);
   void                 getThresholdFromDB(double&, DetId const&);
-  void                 correctThrByPtAndEta(double&);
+  void                 correctThrByPAndEta(double&);
   void                 getThresholdFromCFG(double&, DetId const&);
   void                 testDTstation(TrajectoryStateOnSurface&, std::vector<DTRecSegment4D> const &, double&, DTRecSegment4D&, TrajectoryStateOnSurface&);
   void                 testCSCstation(TrajectoryStateOnSurface&, std::vector<CSCSegment> const &, double&, CSCSegment&, TrajectoryStateOnSurface&);
--- a/RecoMuon/GlobalTrackingTools/interface/DynamicTruncation.h
+++ b/RecoMuon/GlobalTrackingTools/interface/DynamicTruncation.h
@@ -107,6 +107,7 @@ class DynamicTruncation {
   void                 useSegment(CSCSegment const &, TrajectoryStateOnSurface const &);
   void                 sort(ConstRecHitContainer&);
   void                 setEtaRegion();
+  static edm::ParameterSetDescription fillDefaultThrsParameters();
 
   ConstRecHitContainer result, prelFitMeas;
   bool useAPE;
--- a/RecoMuon/GlobalTrackingTools/interface/DynamicTruncation.h
+++ b/RecoMuon/GlobalTrackingTools/interface/DynamicTruncation.h
@@ -66,13 +66,10 @@ class DynamicTruncation {
   void setUpdateState(bool);
   void setUseAPE(bool);
   /*---- DyT v2-----*/
-  void setRecoP(double p) {
-    p_reco = p;
-    useParametrizedThr = true;
-   }
+  void setParThrsMode(bool dytParThrsMode) { useParametrizedThr = dytParThrsMode;}
+  void setRecoP(double p) { p_reco = p; }
   void setRecoEta(double eta) {
     eta_reco = eta;
-    useParametrizedThr = true;
     setEtaRegion();
   }
 
--- a/RecoMuon/GlobalTrackingTools/interface/DynamicTruncation.h
+++ b/RecoMuon/GlobalTrackingTools/interface/DynamicTruncation.h
@@ -46,7 +46,7 @@ namespace dyt_utils{
 };
 
 class DynamicTruncation {
-
+  
  public:
 
   typedef TransientTrackingRecHit::ConstRecHitPointer ConstRecHitPointer;
@@ -56,7 +56,7 @@ class DynamicTruncation {
 
   ~DynamicTruncation();
 
-  void setProd(const edm::Handle<DTRecSegment4DCollection>& DTSegProd,
+  void setProd(const edm::Handle<DTRecSegment4DCollection>& DTSegProd, 
 	       const edm::Handle<CSCSegmentCollection>& CSCSegProd) {
     getSegs->initCSU(DTSegProd, CSCSegProd);
   }
@@ -77,7 +77,7 @@ class DynamicTruncation {
   // Return the vector with the tracker plus the selected muon hits
   TransientTrackingRecHit::ConstRecHitContainer filter(const Trajectory&);
 
-  // Return the DYTInfo object
+  // Return the DYTInfo object 
   reco::DYTInfo getDYTInfo() {
     dytInfo.setNStUsed(nStationsUsed);
     dytInfo.setDYTEstimators(estimatorMap);
--- a/RecoMuon/GlobalTrackingTools/interface/DynamicTruncation.h
+++ b/RecoMuon/GlobalTrackingTools/interface/DynamicTruncation.h
@@ -13,6 +13,7 @@
  *  Authors :
  *  D. Pagano & G. Bruno - UCL Louvain
  *
+ *  \modified by C. Caputo, UCLouvain
  **/
 
 #include <memory>
@@ -40,23 +41,22 @@
 #include "RecoMuon/GlobalTrackingTools/interface/ThrParameters.h"
 #include "RecoMuon/GlobalTrackingTools/interface/ChamberSegmentUtility.h"
 
+namespace dyt_utils{
+  enum class etaRegion {eta0p8, eta1p2, eta2p0, eta2p2, eta2p4};
+};
 
 class DynamicTruncation {
-  
- public:
 
-  /* Variables for v2 */
-  double p_reco;
-  double eta_reco;
+ public:
 
   typedef TransientTrackingRecHit::ConstRecHitPointer ConstRecHitPointer;
   typedef TransientTrackingRecHit::ConstRecHitContainer ConstRecHitContainer;
 
-  DynamicTruncation(const edm::Event&, const MuonServiceProxy&);
+  DynamicTruncation(const edm::Event&, const MuonServiceProxy&, const edm::ParameterSet&);
 
   ~DynamicTruncation();
 
-  void setProd(const edm::Handle<DTRecSegment4DCollection>& DTSegProd, 
+  void setProd(const edm::Handle<DTRecSegment4DCollection>& DTSegProd,
 	       const edm::Handle<CSCSegmentCollection>& CSCSegProd) {
     getSegs->initCSU(DTSegProd, CSCSegProd);
   }
@@ -65,11 +65,21 @@ class DynamicTruncation {
   void setThr(const std::vector<int>&);
   void setUpdateState(bool);
   void setUseAPE(bool);
-  
-  // Return the vector with the tracker plus the selected muon hits
-  TransientTrackingRecHit::ConstRecHitContainer filter(const Trajectory&, double globalMomentum, double globalEta);
+  /*---- DyT v2-----*/
+  void setRecoP(double p) {
+    p_reco = p;
+    useParametrizedThr = true;
+   }
+  void setRecoEta(double eta) {
+    eta_reco = eta;
+    useParametrizedThr = true;
+    setEtaRegion();
+  }
 
-  // Return the DYTInfo object 
+  // Return the vector with the tracker plus the selected muon hits
+  TransientTrackingRecHit::ConstRecHitContainer filter(const Trajectory&);
+
+  // Return the DYTInfo object
   reco::DYTInfo getDYTInfo() {
     dytInfo.setNStUsed(nStationsUsed);
     dytInfo.setDYTEstimators(estimatorMap);
@@ -78,10 +88,12 @@ class DynamicTruncation {
     return dytInfo;
   }
 
+  void                 correctThrByPtAndEta();
+
  private:
 
   void                 compatibleDets(TrajectoryStateOnSurface&, std::map<int, std::vector<DetId> >&);
-  void                 filteringAlgo(double momentum,double eta);
+  void                 filteringAlgo();
   void                 fillSegmentMaps(std::map<int, std::vector<DetId> >&, std::map<int, std::vector<DTRecSegment4D> >&, std::map<int, std::vector<CSCSegment> >&);
   void                 preliminaryFit(std::map<int, std::vector<DetId> >, std::map<int, std::vector<DTRecSegment4D> >, std::map<int, std::vector<CSCSegment> >);
   bool                 chooseLayers(int&, double const &, DTRecSegment4D const &, TrajectoryStateOnSurface const &, double const &, CSCSegment const &, TrajectoryStateOnSurface const &);
@@ -98,7 +110,8 @@ class DynamicTruncation {
   void                 useSegment(DTRecSegment4D const &, TrajectoryStateOnSurface const &);
   void                 useSegment(CSCSegment const &, TrajectoryStateOnSurface const &);
   void                 sort(ConstRecHitContainer&);
-  
+  void                 setEtaRegion();
+
   ConstRecHitContainer result, prelFitMeas;
   bool useAPE;
   std::vector<int> Thrs;
@@ -121,15 +134,19 @@ class DynamicTruncation {
   TrajectoryStateOnSurface prelFitState;
   reco::DYTInfo dytInfo;
   std::map<DTChamberId, GlobalError> dtApeMap;
-  std::map<CSCDetId, GlobalError> cscApeMap; 
+  std::map<CSCDetId, GlobalError> cscApeMap;
   double muonPTest, muonETAest;
   const DYTThrObject* dytThresholds;
   ChamberSegmentUtility* getSegs;
   ThrParameters* thrManager;
   bool useDBforThr;
   bool doUpdateOfKFStates;
+  /* Variables for v2 */
+  double p_reco;
+  double eta_reco;
+  bool useParametrizedThr;
+  dyt_utils::etaRegion region;
+  std::map<dyt_utils::etaRegion, std::vector<double>> parameters;
 };
 
 #endif
-
-
--- a/RecoMuon/GlobalTrackingTools/interface/DynamicTruncation.h
+++ b/RecoMuon/GlobalTrackingTools/interface/DynamicTruncation.h
@@ -52,7 +52,7 @@ class DynamicTruncation {
   typedef TransientTrackingRecHit::ConstRecHitPointer ConstRecHitPointer;
   typedef TransientTrackingRecHit::ConstRecHitContainer ConstRecHitContainer;
 
-  DynamicTruncation(const edm::Event&, const MuonServiceProxy&, const edm::ParameterSet&);
+  DynamicTruncation(const edm::Event&, const MuonServiceProxy&);
 
   ~DynamicTruncation();
 
@@ -66,6 +66,7 @@ class DynamicTruncation {
   void setUpdateState(bool);
   void setUseAPE(bool);
   /*---- DyT v2-----*/
+  void setThrsMap(const edm::ParameterSet&);
   void setParThrsMode(bool dytParThrsMode) { useParametrizedThr = dytParThrsMode;}
   void setRecoP(double p) { p_reco = p; }
   void setRecoEta(double eta) {
--- a/RecoMuon/GlobalTrackingTools/interface/DynamicTruncation.h
+++ b/RecoMuon/GlobalTrackingTools/interface/DynamicTruncation.h
@@ -45,6 +45,10 @@ class DynamicTruncation {
   
  public:
 
+  /* Variables for v2 */
+  double p_reco;
+  double eta_reco;
+
   typedef TransientTrackingRecHit::ConstRecHitPointer ConstRecHitPointer;
   typedef TransientTrackingRecHit::ConstRecHitContainer ConstRecHitContainer;
 
@@ -63,7 +67,7 @@ class DynamicTruncation {
   void setUseAPE(bool);
   
   // Return the vector with the tracker plus the selected muon hits
-  TransientTrackingRecHit::ConstRecHitContainer filter(const Trajectory&);
+  TransientTrackingRecHit::ConstRecHitContainer filter(const Trajectory&, double globalMomentum, double globalEta);
 
   // Return the DYTInfo object 
   reco::DYTInfo getDYTInfo() {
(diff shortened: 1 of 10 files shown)


slava77:
@cmsbuild please test

slava77:
@clacaputo 
please add a link to a presentation in the muon POG to the PR description.

clacaputo:
> @clacaputo
> please add a link to a presentation in the muon POG to the PR description.

@slava77, PR description updated with the link to Muon POG presentation.

slava77:
@cmsbuild please test

RecoMuon/GlobalTrackingTools/src/DynamicTruncation.cc
```diff
@@ -453,17 +470,38 @@ void DynamicTruncation::getThresholdFromDB(double& thr, DetId const& id) {
...
-//===> correctThrByPtAndEta
-void DynamicTruncation::correctThrByPtAndEta(double& thr) {
+//===> correctThrByPAndEta
+void DynamicTruncation::correctThrByPAndEta(double& thr) {
 
-  //////////////////////////////////////
-  // This section will be implemented //
-  //    after the release of APEs     //
-  //////////////////////////////////////
+
```
perrotta: Remove these two empty lines, for better readability

RecoMuon/GlobalTrackingTools/src/DynamicTruncation.cc
```diff
@@ -453,17 +470,38 @@ void DynamicTruncation::getThresholdFromDB(double& thr, DetId const& id) {
...
+
+    auto parametricThreshold = [this]{
+      double thr50 = this->parameters[this->region].at(0);
+      double p0    = this->parameters[this->region].at(1);
+      double p1    = this->parameters[this->region].at(2);
+      return thr50 * ( 1 + p0*p_reco + TMath::Power( this->p_reco, p1));
```
perrotta: There is no need to use TMath methods: use ```std::pow(x,y)``` instead

RecoMuon/GlobalTrackingTools/src/DynamicTruncation.cc
```diff
@@ -453,17 +470,38 @@ void DynamicTruncation::getThresholdFromDB(double& thr, DetId const& id) {
...
+    };
+
+    // printf("thr before = %f \n", thr);
```
perrotta: remove useless commented out lines

RecoMuon/GlobalTrackingTools/src/DynamicTruncation.cc
```diff
@@ -453,17 +470,38 @@ void DynamicTruncation::getThresholdFromDB(double& thr, DetId const& id) {
...
+    // printf("parametricThreshold(1,%f, %f) = %f \n", this->parameters[this->region].at(0), this->parameters[this->region].at(1), parametricThreshold());
+    thr = parametricThreshold();
+}
+
+void DynamicTruncation::setEtaRegion(){
+
+  // Defaul value for muons with abs(eta) > 2.4
+  region = dyt_utils::etaRegion::eta2p4;
+
+  if( TMath::Abs(eta_reco) > 0 && TMath::Abs(eta_reco) <= 0.8 ) region = dyt_utils::etaRegion::eta0p8;
```
perrotta: Use ```std::abs(eta_reco)``` instead
Even better if you define only once
```C++
   absEta = std::abs(eta_reco);
```
and keep using it

RecoMuon/GlobalTrackingTools/src/DynamicTruncation.cc
```diff
@@ -453,17 +470,38 @@ void DynamicTruncation::getThresholdFromDB(double& thr, DetId const& id) {
...
+  else if( TMath::Abs(eta_reco) > 0.8 && TMath::Abs(eta_reco) <= 1.2 ) region = dyt_utils::etaRegion::eta1p2;
```
perrotta: In order to avoid unlikely but not impossible rounding issues, please remove the lower bound and make all regions really exclusive:
```C++
   absEta = std::abs(eta_reco);
   if (absEta <= 0.8)        region = dyt_utils::etaRegion::eta0p8;
   else if (absEta <= 1.2) region = dyt_utils::etaRegion::eta1p2;
   else if ...
```

RecoMuon/GlobalTrackingTools/src/GlobalMuonRefitter.cc
```diff
@@ -118,8 +119,10 @@ GlobalMuonRefitter::GlobalMuonRefitter(const edm::ParameterSet& par,
 
   theDYTthrs     = par.getParameter< std::vector<int> >("DYTthrs");
   theDYTselector = par.existsAs<int>("DYTselector")?par.getParameter<int>("DYTselector"):1;
-  theDYTupdator = par.existsAs<bool>("DYTupdator")?par.getParameter<bool>("DYTupdator"):false;
-  theDYTuseAPE = par.existsAs<bool>("DYTuseAPE")?par.getParameter<bool>("DYTuseAPE"):false;
+  theDYTupdator  = par.existsAs<bool>("DYTupdator")?par.getParameter<bool>("DYTupdator"):false;
```
perrotta: I think you can define all these parameters with a ```fillDescriptions()``` method, and get rid of all these ```par.existsAs``` checks

RecoMuon/GlobalTrackingTools/src/GlobalMuonRefitter.cc
```diff
@@ -256,27 +259,33 @@ vector<Trajectory> GlobalMuonRefitter::refit(const reco::Track& globalTrack,
...
       //
       DynamicTruncation dytRefit(*theEvent,*theService);
       dytRefit.setProd(all4DSegments, CSCSegments);
       dytRefit.setSelector(theDYTselector);
       dytRefit.setThr(theDYTthrs);
       dytRefit.setUpdateState(theDYTupdator);
       dytRefit.setUseAPE(theDYTuseAPE);
+      if(dytParThrsMode) {
+        dytRefit.setParThrsMode(dytParThrsMode);
+				dytRefit.setThrsMap(theDYTthrsParameters);
```
perrotta: Please indent as in the other lines

RecoMuon/GlobalTrackingTools/src/GlobalMuonRefitter.cc
```diff
@@ -817,25 +826,25 @@ vector<Trajectory> GlobalMuonRefitter::transform(const reco::Track& newTrack,
     }
   }
 
-/*  
+/*
```
perrotta: All the commented out lines below could be simply removed, or replaced with some LogDebug/LogTrace

perrotta:
@clacaputo @Fedespring : this PR is going to modify heavily the DYT method for high pt muons.
Could you please confirm that it is endorsed by the muon POG as a replacement of the current TeVMuons?
Is this meant for the UL processing, or for some later run3 development?

Fedespring:
Actually the final validation is going arrive. The code allows to switch between old and new DYT, but at the moment, in order to get new DYT for the UL, the flag is set "True". In case of problems we can easily switch the flag to "False". In any case I would wait the final validation, which should arrive by tomorrow (if no deadline will be missed).

RecoMuon/GlobalTrackingTools/src/GlobalMuonRefitter.cc
clacaputo: Dear @perrotta, could you please point me an example of how `fillDescriptions()` works?

perrotta: See https://twiki.cern.ch/twiki/bin/view/CMSPublic/SWGuideConfigurationValidationAndHelp

You can find concrete examples in the vast majority of classes in CMSSW which have configurable parameters

perrotta:
@Fedespring @clacaputo 

While still waiting for your validation results and fixes, let me point out that it doesn't seems true to me that "In case of problems we can easily switch the flag to False", as this would not restore the past behavior of this algo but it will only avoid correcting for P and eta
https://github.com/cms-sw/cmssw/pull/26343/files#diff-3a264143d32993e2dbcb83965ba78272R473
an option which is not foreseen for the code currently in the IB.
Could you please verify?

clacaputo:
@perrotta 
The core of the algorithm is not modified by this PR.
`correctThrByPAndEta` function change the threshold value used by the refitting procedure using p and eta info about the track. Setting `dytParThrsMode` will avoid the use of `correctThrByPAndEta`, keeping the thresholds as they were in the previous DyT version.

perrotta:
> @perrotta
> The core of the algorithm is not modified by this PR.
> `correctThrByPAndEta` function change the threshold value used by the refitting procedure using p and eta info about the track. Setting `dytParThrsMode` will avoid the use of `correctThrByPAndEta`, keeping the thresholds as they were in the previous DyT version.

Ok, I see...

In the code currently in the IB you call the ```DynamicTruncation::correctThrByPtAndEta(double& thr) ``` method, which is however an empty one. In your new code you call ```DynamicTruncation::correctThrByPAndEta(double& thr) ``` only when ```useParametrizedThr``` is True: when that flag is false you don't call any correction method, which somehow reproduces the old behavior because and empty method was called there...

Then, I agree that while setting ```useParametrizedThr = False``` the current behaviour can be restored

Fedespring:
As discussed, please go ahead with the PR as it is, so that we can easily get the central validation. In parallel further studies will be done, and in pre5 we will either update the parametrization or switch the flag off.

perrotta:
@clacaputo @Fedespring 
Yes, please go ahead and apply the required fixes

kpedro88:
please follow http://cms-sw.github.io/tutorial-resolve-conflicts.html to learn how to update a PR correctly (do not use `cms-merge-topic`)

slava77:
-1

same as https://github.com/cms-sw/cmssw/pull/26343#issuecomment-482769208

slava77:
@cmsbuild please test

slava77:
@smuzaffar , please check if the following indicates a problem in the bot logic
- I gave a "-1" to this on Apr 13 at 12:26
- the bot recognized it and changed the label at 12:27
- There was an update in the PR 10 mins later:  "clacaputo force-pushed the clacaputo:DyT_v2_10_6_X_tmp branch from d60e4b8 to bdf5109 " on Apr 13 at 12:36
- the bot removed a bunch of unnecessary labels at 12:37, but it did not reset the -1 signature from reco.

slava77:
> As discussed, please go ahead with the PR as it is, so that we can easily get the central validation. In parallel further studies will be done, and in pre5 we will either update the parametrization or switch the flag off.

Does this mean that you are proposing to use central default RelVals for debugging the algorithm?
My understanding of the discussion in the ORP was that this PR will introduce the change off by default; there will be a request to validate this change incrementally, and then it will be turned on in pre5/10_6_0 if validation is OK.

Fedespring:
The algorithm is the same, what changes is the parametrization. The new parametrization has been calculated using some dedicated samples, and further debugging is ongoing using these samples. Of course having all the RelVal available, like W' and Z' at very high mass, would make the debug much faster and easier. In pre5/10_6_0 a new PR will be done in any case, either to switch the new parametrization off, or to change the new parametrization keeping it on. If this is not a problem, please merge the PR as it is, otherwise let us know and the new parametrization will be switched off. Thanks.

slava77:
On 4/15/19 2:49 PM, Fedespring wrote:
> In pre5/10_6_0 a new PR will be done in any case, either to switch the
> new parametrization off, or to change the new parametrization keeping it on.

I'm confused.
Does this mean that you know already that some changes have to be made
to this PR
(even if we assume that the new parameterization doesn't need to be
turned off)?
This sounds like a request to keep the pre5 open.

@fabiocos

Fedespring:
I would say yes @slava77, we are already working on a better implementation/validation. I don't think we can keep the parametrization as it is now, unless the ongoing checks at very high mass show that the impact is perfectly under control.

fabiocos:
@slava77 @Fedespring this essentially means that in any case some new development will have to enter after pre4 for muons...

slava77:
based on the discussion in the ORP today, I expect that this PR is updated:
- the default is made to not change the defaults
- a config switch is supplied to enable the new parameterization so that this can be passed to the PDMV for making a few extra samples needed to validate/finalize the DYT updates

fabiocos:
@clacaputo @Fedespring when this new version can become available? If we want to have it in pre4, this should happen quite quickly...

Fedespring:
Dear @slava77 this PR contains: a new parametrization and a flag to switch it on and off. We will simply switch it off, as agreed in the ORP. 
Then to get the dedicated RelVal samples, the PdmV will take care about to create a module to switch the parameter on.  I understood that is the default procedure in this case.

prebello:
@fabiocos please give more details about it. 
Dedicated relvals with special configurations can be requested in JIRA project https://its.cern.ch/jira/projects/PDMVRELVALS

fabiocos:
@prebello I think MUO POG people (@Fedespring  et al.) will communicate to PdmV in JIRA the few special samples needed with the corresponding settings

Fedespring:
@prebello @fabiocos the ticket* has been created, and the PR is going to be updated very soon
(*)https://its.cern.ch/jira/browse/PDMVRELVALS-43

fabiocos:
please test

RecoMuon/GlobalTrackingTools/interface/GlobalMuonRefitter.h
```diff
@@ -171,15 +172,17 @@ class GlobalMuonRefitter {
     int theDYTselector;
     bool theDYTupdator;
     bool theDYTuseAPE;
+    bool dytParThrsMode;
+    edm::ParameterSet theDYTthrsParameters;
```
slava77: please keep ```the``` as a prefix for data members

RecoMuon/GlobalTrackingTools/plugins/GlobalTrackQualityProducer.cc
```diff
@@ -311,5 +311,77 @@ GlobalTrackQualityProducer::trackProbability(Trajectory& track) const {
...
+
+    descGlbMuonRefitter.add<int>("SkipStation", -1);
+    descGlbMuonRefitter.add<int>("TrackerSkipSystem",-1);
+    descGlbMuonRefitter.add<int>("TrackerSkipSection", -1);
+    descGlbMuonRefitter.add<bool>("RefitFlag", true );
+
+    //descRefitter.add<edm::ParameterSetDescription>("GlobalMuonRefitter", descGlbMuonRefitter);
+    desc.add<edm::ParameterSetDescription>("RefitterParameters", descGlbMuonRefitter);
+  }
+  //desc.add<edm::ParameterSetDescription>("RefitterParameters", descRefitter);
```
slava77: is this commented out code needed?

RecoMuon/GlobalTrackingTools/plugins/GlobalTrackQualityProducer.cc
```diff
@@ -311,5 +311,77 @@ GlobalTrackQualityProducer::trackProbability(Trajectory& track) const {
...
+  desc.add<double>("nSigma", 3.0);
+  desc.add<double>("MaxChi2", 100000.0);
+
+  descriptions.add("GlobalTrackQualityProducer", desc);
```
slava77: the default name should start with a lower case ```globalTrackQualityProducer```.

Perhaps more importantly: we should start using the file generated with the fillDescriptions to replace the content of RecoMuon/GlobalTrackingTools/python/GlobalTrackQuality_cfi.py

RecoMuon/GlobalTrackingTools/plugins/GlobalTrackQualityProducer.cc
```diff
@@ -311,5 +311,77 @@ GlobalTrackQualityProducer::trackProbability(Trajectory& track) const {
...
+
+    descGlbMuonRefitter.add<int>("MuonHitsOption", 1);
+    descGlbMuonRefitter.add<double>("PtCut", 1.0);
+    descGlbMuonRefitter.add<double>("Chi2ProbabilityCut", 30.0);
+    descGlbMuonRefitter.add<double>("Chi2CutCSC", 1.0);
+    descGlbMuonRefitter.add<double>("Chi2CutDT", 30.0);
+    descGlbMuonRefitter.add<double>("Chi2CutGEM", 1.0);
+    descGlbMuonRefitter.add<double>("Chi2CutME0", 1.0);
+    descGlbMuonRefitter.add<double>("Chi2CutRPC", 1.0);
+    descGlbMuonRefitter.add<int>("HitThreshold", 1);
```
slava77: these changes to reimplement RecoMuon/GlobalTrackingTools/python/GlobalMuonRefitter_cff.py here is a dead end and probably does more harm than good.
I propose to do the same as for ```GlobalMuonTrackMatcher``` on L324 ("allowAnything" so that the corresponding PSet is picked up from a central space

RecoMuon/GlobalTrackingTools/python/GlobalMuonRefitter_cff.py
```diff
@@ -26,13 +26,21 @@
...
     DYTuseAPE = cms.bool(False),
-    
+    ## Parameters for DYT threshold parametrization
+    DYTuseThrsParametrization = cms.bool(False),
+    DYTthrsParameters = cms.PSet(
+                                  eta0p8 = cms.vdouble(1, -0.919853, 0.990742),
+                                  eta1p2 = cms.vdouble(1, -0.897354, 0.987738),
+                                  eta2p0 = cms.vdouble(4, -0.986855, 0.998516),
+                                  eta2p2 = cms.vdouble(1, -0.940342, 0.992955),
+                                  eta2p4 = cms.vdouble(1, -0.947633, 0.993762),
```
slava77: instead of copy-pasting here and in GlobalTrajectoryBuilderCommon_cff.py,
better define these parameters in a separate file and load from there.

RecoMuon/GlobalTrackingTools/src/DynamicTruncation.cc
```diff
@@ -33,6 +34,14 @@ using namespace edm;
 using namespace std;
 using namespace reco;
 
+namespace dyt_utils{
+  std::map<etaRegion, std::string> etaRegionStr { {etaRegion::eta0p8, "eta0p8"},
```
slava77: add ```static const ```

RecoMuon/L3MuonProducer/src/L3MuonProducer.cc
```diff
@@ -217,6 +217,19 @@ void L3MuonProducer::fillDescriptions(edm::ConfigurationDescriptions& descriptio
...
+      psd1.add<bool>("DYTuseAPE", false );
+      psd1.add<bool>("DYTuseThrsParametrization", false);
+      {
+        edm::ParameterSetDescription psd2;
+        psd2.add<std::vector<double>>("eta0p8", {1,-0.919853, 0.990742});
+        psd2.add<std::vector<double>>("eta1p2", {1,-0.897354, 0.987738});
+        psd2.add<std::vector<double>>("eta2p0", {1,-0.986855, 0.998516});
+        psd2.add<std::vector<double>>("eta2p2", {1,-0.940342, 0.992955});
+        psd2.add<std::vector<double>>("eta2p4", {1,-0.947633, 0.993762});
+        psd1.add<edm::ParameterSetDescription>("DYTthrsParameters", psd2);
```
slava77: perhaps instead define a static method DynamicTruncation::fillDefaultThrsParameters, which will create this PSet; and call that method here.
This can then be reused in other fillDescriptions if needed.

slava77 (commented):
```static const``` in DynamicTruncation.cc should be added in this PR before merging
because it's adding a new static analyzer issue https://cmssdt.cern.ch/SDT/jenkins-artifacts/pull-request-integration/PR-26343/34257/llvm-analysis/report-396378.html#EndPath

Most of the rest can be considered for the next update

File: RecoMuon/L3MuonProducer/src/L3MuonProducer.cc [hunk 16]
```diff
@@ -217,6 +217,19 @@ void L3MuonProducer::fillDescriptions(edm::ConfigurationDescriptions& descriptio
...
       });
+      psd1.add<int>("DYTselector",1);
+      psd1.add<bool>("DYTupdator", false);
+      psd1.add<bool>("DYTuseAPE", false );
+      psd1.add<bool>("DYTuseThrsParametrization", false);
+      {
+        edm::ParameterSetDescription psd2;
+        psd2.add<std::vector<double>>("eta0p8", {1,-0.919853, 0.990742});
+        psd2.add<std::vector<double>>("eta1p2", {1,-0.897354, 0.987738});
+        psd2.add<std::vector<double>>("eta2p0", {1,-0.986855, 0.998516});
```
slava77: the first value is 4 in the python versions. Either this or the python case is a typo.

RecoMuon/GlobalTrackingTools/plugins/GlobalTrackQualityProducer.cc
```diff
@@ -311,5 +311,77 @@ GlobalTrackQualityProducer::trackProbability(Trajectory& track) const {
...
+    descGlbMuonRefitter.add<std::vector<int>>("DYTthrs",{10, 10});
+    descGlbMuonRefitter.add<int>("DYTselector",1);
+    descGlbMuonRefitter.add<bool>("DYTupdator", false);
+    descGlbMuonRefitter.add<bool>("DYTuseAPE", false );
+    descGlbMuonRefitter.add<bool>("DYTuseThrsParametrization", false);
+    {
+      edm::ParameterSetDescription descDYTthrs;
+      descDYTthrs.add<std::vector<double>>("eta0p8", {1,-0.919853, 0.990742});
+      descDYTthrs.add<std::vector<double>>("eta1p2", {1,-0.897354, 0.987738});
+      descDYTthrs.add<std::vector<double>>("eta2p0", {1,-0.986855, 0.998516});
```
slava77: the first value is 4 in the python versions. Either this or the python case is a typo.

slava77 (commented):
two more, suggestive of a bug/typo

RecoMuon/L3MuonProducer/src/L3MuonProducer.cc
clacaputo: It's a typo. Thanks

RecoMuon/GlobalTrackingTools/plugins/GlobalTrackQualityProducer.cc
clacaputo: Typo, Thanks

fabiocos:
please test

RecoMuon/GlobalTrackingTools/interface/DynamicTruncation.h
```diff
@@ -107,6 +107,7 @@ class DynamicTruncation {
   void                 useSegment(CSCSegment const &, TrajectoryStateOnSurface const &);
   void                 sort(ConstRecHitContainer&);
   void                 setEtaRegion();
+  static edm::ParameterSetDescription fillDefaultThrsParameters();
```
slava77: it looks like some commits were missed.
I do not see an implementation of this method.
Please remove to avoid undefined symbols.

clacaputo: I forgot to remove this line

slava77:
@cmsbuild please test

fabiocos:
@Martin-Grunewald do you have any further hot comment on this PR?

Martin-Grunewald:
No...

slava77:
Looking at tests for the case with ```DYTuseThrsParametrization = True```, there are some workflows where the particleFlow (muon) momentum changes. So, this can technically affect the PF quantities like jets and MET.

Here is a plot from as of bdf5109e6b599d880fd266320a45448a24df74fb
PR-26343/34185 CMSSW_10_6_X_2019-04-12-2300 
https://github.com/cms-sw/cmssw/pull/26343#issuecomment-482828922
workflow 136.731 SinglePh2016B
all_OldVSNew_RunSinglePh2016Bwf136p731c_log10recoPFCandidates_particleFlow__reRECO_obj_pt50
Notably, the affected pt is just under 2 GeV.

slava77:
related to https://its.cern.ch/jira/browse/PDMVRELVALS-43
and the setup with ```DYTuseThrsParametrization = True``` to be made from False,
the parameter is present in the following modules PSets: ```displacedGlobalMuons``` , ```glbTrackQual```, ```globalMuons```, ```globalSETMuons``` (not running, IIRC), ```tevMuons ```. 
IIUC, the revert from True to False affected all of these modules.
Please make sure to pick the full set or later modify False to True consistently with the requests given to PDMV.

slava77:
+1

for https://github.com/cms-sw/cmssw/pull/26343	a5ba439
- code changes are somewhat in line with the PR description and more so with the follow up review
    - please update the PR description to note that the feature remains off
    - please consider still unresolved comments from https://github.com/cms-sw/cmssw/pull/26343#pullrequestreview-227996851 for the follow up update
- jenkins tests pass and comparisons with the baseline show no differences (as expected for the new DYT feature turned off).

fabiocos:
@Martin-Grunewald ok, I'll wait for your signature to merge it

Martin-Grunewald:
+1

fabiocos:
+1

PR #26349, opened 2019-04-03
Bug fix for behavior after exceptions with concurrent lumis

wddgit:
#### PR description:

This fixes a bug that would be encountered when more than
one lumi is being processed concurrently and there is an
exception. If the exception is associated with a lumi other
than the last one read, an infinite wait will be entered.
The wait is in the processLumis function of the EventProcessor.
It waits on every task in the stream SerialTaskQueues.
In the above situation the serial queue is not resumed
before the wait and there are still tasks in the queues.
There is code that would clean these up in the function
endUnfinishedLumis but that only runs after the wait
is over and processLumis has returned. Too late.

This commit also fixes the maxEvents output parameter which was
broken when more than one lumi is processed concurrently. As far
as I know nothing uses this, but it is an advertised parameter
of the Framework that is supposed to work. One note about
this. If events are running concurrently when the limit is
reached, then all the events already running will complete, so
the job could write more than the requested number of events
to output.

The modified code is not executed if there are no exceptions and
the maxEvents output parameter is not used. Under normal
circumstances this should not have any effect on output or
performance.

#### PR validation:

This adds two new unit tests. One test would fail with an infinite
wait if executed in a release before this commit. It creates the
condition described above. The other tests the maxEvent output
parameter. Plus I added an assert that checks that the reference
count is one in normalEnd when we try to delete RunResources
and get endRun to execute (something to be concerned about in
concurrent lumi mode).

### DIFF
--- a/FWCore/Framework/src/EventProcessor.cc
+++ b/FWCore/Framework/src/EventProcessor.cc
@@ -243,8 +243,7 @@ namespace edm {
     fileModeNoMerge_(false),
     exceptionMessageFiles_(),
     exceptionMessageRuns_(),
-    exceptionMessageLumis_(),
-    exceptionMessageLumisIsSet_(false),
+    exceptionMessageLumis_(false),
     forceLooperToEnd_(false),
     looperBeginJobRun_(false),
     forceESCacheClearOnNewRun_(false),
@@ -280,8 +279,7 @@ namespace edm {
     fileModeNoMerge_(false),
     exceptionMessageFiles_(),
     exceptionMessageRuns_(),
-    exceptionMessageLumis_(),
-    exceptionMessageLumisIsSet_(false),
+    exceptionMessageLumis_(false),
     forceLooperToEnd_(false),
     looperBeginJobRun_(false),
     forceESCacheClearOnNewRun_(false),
@@ -319,8 +317,7 @@ namespace edm {
     fileModeNoMerge_(false),
     exceptionMessageFiles_(),
     exceptionMessageRuns_(),
-    exceptionMessageLumis_(),
-    exceptionMessageLumisIsSet_(false),
+    exceptionMessageLumis_(false),
     forceLooperToEnd_(false),
     looperBeginJobRun_(false),
     forceESCacheClearOnNewRun_(false),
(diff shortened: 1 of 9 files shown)


wddgit:
please test

Dr15Jones:
It looks like the new test failed?

```
Failure running cmsRun /build/cmsbld/jenkins/workspace/ib-any-integration/CMSSW_10_6_X_2019-04-03-1100/src/FWCore/Framework/test/testMaxEventsOutput_cfg.py: status 65
```

wddgit:
please test

I had the unit tests working, then at the last minute I experimented with some minor test changes to check something and failed to correctly restore it to what I originally intended. Try again.

FWCore/Framework/src/EventProcessor.cc
```diff
@@ -1155,7 +1158,9 @@ namespace edm {
                                                           [this,i,h = holder](std::exception_ptr const* iPtr) mutable
                   {
                     if(iPtr) {
-                      h.doneWaiting(*iPtr);
+                      WaitingTaskHolder tmp(h);
```
Dr15Jones: For my own notes: making the copy avoids having `dineWaiting` launch the task before the call to `streamEndLumiAsync`.

FWCore/Framework/src/EventProcessor.cc
```diff
@@ -1747,7 +1788,10 @@ namespace edm {
   }
 
   void EventProcessor::setExceptionMessageLumis(std::string& message) {
-    exceptionMessageLumis_ = message;
+    bool expected = false;
+    if (exceptionMessageLumisIsSet_.compare_exchange_strong(expected,true)) {
```
Dr15Jones: This is not technically thread safe. The `exceptionMessageLumisIsSet_` is not functioning as a proper guard since it does not guarantee that after writing to `exceptionMessageLumis_` that memory is synchronized across all threads that might read it. In practice, when the `tbb::task` that called `setExceptionMessageLumis` completes, it will synchronize all memory written during the task's exectution.

The better way to do this is to mae `exceptionMessageLumis_ ` a `std::atomic<std::string*>` and handle memory cleanup at the point which looks at `exceptionMessageLumis_`.

wddgit: Agreed. Although nothing was actually reading it in the concurrent parts of the code, but I agree it is a landmine we might possibly hit later. I thought of a simpler fix. The string is always the same now.  It is simpler just to delete the string data member entirely. It is not needed. All we need is an atomic<bool> data member. Fixed in the commit I just pushed.

wddgit:
I just pushed a fix to Chris's push. And as I looked at it some more I saw some other ways to improve it. It is simplified in a few places and also will behave better when there are multiple exceptions getting raised, usually a primary one and others that are side effects of the primary exception.

wddgit:
please test

Dr15Jones:
+1

fabiocos:
+1

PR #26272, opened 2019-03-27
Run2-hcx193 Changes to DDHCalBarrelAlgo by removing explicit reference to CLHEP

bsunanda:
#### PR description:

Remove explicit reference to CLHEP from DDHCalBarrelAlgo and also making changes to debug statements

#### PR validation:

Ran the dumpGeometry code with debug switched on in the existing and the new code. The debug prints out all steps within DDHcalBarrelAlgo - all parameters, positions (translation + rotation). A simple diff between the two versions yield no difference:

diff -bw /build/sunanda/CMSSW_10_6_X_2019-03-26-2300/relval/dump1.log dump.log 

1c1
< Begin processing the 1st record. Run 1, Event 1, LumiSection 1 on stream 0 at 27-Mar-2019 18:52:01.004 CET
---
> Begin processing the 1st record. Run 1, Event 1, LumiSection 1 on stream 0 at 27-Mar-2019 18:51:29.748 CET
656c656
< <<== End of DDHCalBarrelAlgo construction ...
---
> <<== End of DDHCalBarrelAlgo construction

#### if this PR is a backport please specify the original PR:

No back porting is required.

PR #26271, opened 2019-03-27
Run2-hcx192 Changes to DDHCalAngular by removing explicit reference to CLHEP

bsunanda:
#### PR description:

Remove explicit reference to CLHEP from DDHCalAngular and also making changes to debug statements

#### PR validation:

Run the code with debug On which prints out the positioning with translation and rotation in this algorithm for all calls in the new and the older versions and make a diff which results no difference:

diff -bw /build/sunanda/CMSSW_10_6_X_2019-03-26-2300/relval/dump2.log dump.log

1c1
< Begin processing the 1st record. Run 1, Event 1, LumiSection 1 on stream 0 at 27-Mar-2019 18:59:23.481 CET
---
> Begin processing the 1st record. Run 1, Event 1, LumiSection 1 on stream 0 at 27-Mar-2019 19:01:00.661 CET

#### if this PR is a backport please specify the original PR:

Need to go to master and is the first step before translating this for dd4hep

PR #26351, opened 2019-04-04
Run2-hcx194 Changes to some of the algorithms used for TB application by removing explicit reference to CLHEP

bsunanda:
#### PR description:

Three algorithms used in describing the geometry of ECAL+HCAL Test Beam setups are modified to avoid explicit calls to CLHEP. The scripts for TB simulation are updated

#### PR validation:

Some of the TB applications are run and tested

This is not for back porting

### DIFF
--- a/Geometry/HcalAlgo/plugins/DDHCalTBZposAlgo.cc
+++ b/Geometry/HcalAlgo/plugins/DDHCalTBZposAlgo.cc
@@ -6,17 +6,20 @@
 #include <cmath>
 #include <algorithm>
 
-namespace std{} using namespace std;
-
 #include "FWCore/MessageLogger/interface/MessageLogger.h"
+#include "DataFormats/Math/interface/GeantUnits.h"
 #include "DetectorDescription/Core/interface/DDLogicalPart.h"
 #include "DetectorDescription/Core/interface/DDCurrentNamespace.h"
 #include "DetectorDescription/Core/interface/DDSplit.h"
 #include "Geometry/HcalAlgo/plugins/DDHCalTBZposAlgo.h"
-#include "CLHEP/Units/GlobalSystemOfUnits.h"
+
+//#define EDM_ML_DEBUG
+using namespace geant_units::operators;
 
 DDHCalTBZposAlgo::DDHCalTBZposAlgo() {
-  LogDebug("HCalGeom") << "DDHCalTBZposAlgo test: Creating an instance";
+#ifdef EDM_ML_DEBUG
+  edm::LogVerbatim("HCalGeom") << "DDHCalTBZposAlgo: Creating an instance";
+#endif
 }
 
 DDHCalTBZposAlgo::~DDHCalTBZposAlgo() {}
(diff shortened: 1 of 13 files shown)


bsunanda:
@cmsbuild Please test

Geometry/HcalAlgo/plugins/DDHCalTBZposAlgo.cc
```diff
@@ -35,49 +38,57 @@ void DDHCalTBZposAlgo::initialize(const DDNumericArguments & nArgs,
...
  
-  double thetax = 90.*CLHEP::deg - theta;
+  double thetax = 90._deg - theta;
   double z      = zoffset + dist*tan(thetax);
   double x      = shiftX - shiftY*sin(tilt);
   double y      = shiftY*cos(tilt);
   DDTranslation tran(x,y,z);
   DDRotation rot;
-  double tiltdeg = tilt/CLHEP::deg;
+  double tiltdeg = convertRadToDeg(tilt);
```
cvuosalo: As I mentioned for #26271 and #26272, we want to eliminate conversion of angles to degrees, except in the debug code.

bsunanda: But we had the convention of reusing rotation matrices and the common ones are rotation around z-axis by an angle phi - the convention is Phi where phi is in degrees.

bsunanda:
@cmsbuild Please test

bsunanda:
@cmsbuild Please test

Geometry/HcalAlgo/plugins/DDHCalTBZposAlgo.cc
cvuosalo: There is a utility to create these rotation strings: [formatAsDegrees](https://github.com/cms-sw/cmssw/blob/6383740652b50371bd88dcd809fe8ab28e9c3510/DetectorDescription/Core/interface/DDTypes.h#L36)

If there is anything wrong with this utility or if it is declared in an unsuitable header file, then we should fix it so it can be widely used.

bsunanda:
@cmsbuild Please test

Geometry/HcalAlgo/plugins/DDHCalXtalAlgo.cc
```diff
@@ -33,60 +38,72 @@ void DDHCalXtalAlgo::initialize(const DDNumericArguments & nArgs,
...
     DDTranslation tran(pos[0], pos[1], pos[2]);
     DDName parentName = parent().name(); 
 
-    if (abs(angle) > 0.01*CLHEP::deg) {
-      LogDebug("HCalGeom") << "DDHCalXtalAlgo test: Creating a new rotation " 
-			   << rotstr << "\t" << theta[0]/CLHEP::deg << "," 
-			   << phi[0]/CLHEP::deg << "," << theta[1]/CLHEP::deg 
-			   << "," << phi[1]/CLHEP::deg << "," 
-			   << theta[2]/CLHEP::deg << ","  << phi[2]/CLHEP::deg;
+    static const double tol = 0.000174523292; // 0.01 degree
```
cvuosalo: No need to change this line now, but wouldn't using `_deg` make this line look nicer?

`static constexpr double tol = 0.01_deg;`

cvuosalo:
+1

Geometry/HcalAlgo/plugins/DDHCalXtalAlgo.cc
bsunanda: I shall remember and do that later

bsunanda:
@civanch Could you please approve this PR?

civanch:
+1

fabiocos:
+1

PR #26355, opened 2019-04-04
Updates on DQMOffline ZCounting for ultra legacy reco

davidwalter2:
#### PR description:

There are updates concerning the different configurations for the different years of the UL, these is implemented using era modifier. There are also cosmetic updates and functional updates on the module itself.

#### PR validation:

It was tested locally with the test file in /DQMOffline/Lumi/test/ZCounting_cfg.py for different years on thesingle muon primary dataset (where it is intended to run only).

### DIFF
--- a/Configuration/StandardSequences/python/Eras.py
+++ b/Configuration/StandardSequences/python/Eras.py
@@ -61,6 +61,7 @@ class Eras (object):
                            'run2_nanoAOD_94XMiniAODv1', 'run2_nanoAOD_94XMiniAODv2', 'run2_nanoAOD_94X2016',
                            'run2_miniAOD_devel', 'run2_nanoAOD_102Xv1',
                            'hcalHardcodeConditions', 'hcalSkipPacker',
+                           'run2_zcount_2016','run2_zcount_2017','run2_zcount_2018',
                            'bParking']
         internalUseModChains = ['run2_2017_noTrackingModifier']
 
--- a/Configuration/StandardSequences/python/Eras.py
+++ b/Configuration/StandardSequences/python/Eras.py
@@ -61,7 +61,7 @@ class Eras (object):
                            'run2_nanoAOD_94XMiniAODv1', 'run2_nanoAOD_94XMiniAODv2', 'run2_nanoAOD_94X2016',
                            'run2_miniAOD_devel', 'run2_nanoAOD_102Xv1',
                            'hcalHardcodeConditions', 'hcalSkipPacker',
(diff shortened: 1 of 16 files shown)


andrius-k:
please test

Configuration/StandardSequences/python/Eras.py
```diff
@@ -61,6 +61,7 @@ def __init__(self):
                            'run2_nanoAOD_94XMiniAODv1', 'run2_nanoAOD_94XMiniAODv2', 'run2_nanoAOD_94X2016',
                            'run2_miniAOD_devel', 'run2_nanoAOD_102Xv1',
                            'hcalHardcodeConditions', 'hcalSkipPacker',
+                           'run2_zcount_2016','run2_zcount_2017','run2_zcount_2018',
```
jfernan2: This kind of additions are going to be needed as a general modification to treat Trigger paths in HLT for the Run2 years. Perhaps a more general name like "run2_HLTconditions_201X"  could work for everybody otherwise we could end having every PAG/POG adding its own Modifier for the same use case: changes in HLT paths.
Just my two pence on this issue

andrius-k:
please test

davidwalter2:
I don't get it, can an expert please take a look?

fabiocos:
please test

andrius-k:
+1

DQMOffline/Lumi/src/TTrigger.cc
```diff
@@ -22,10 +22,12 @@ using namespace ZCountingTrigger;
 // the trigger can evolve and obtain a different trigger object name, but we still want this to 
 // be associated with the same leg (e.g. the trailing electron in a dielectron trigger)
 //
-TTrigger::TTrigger() { 
+TTrigger::TTrigger(std::vector<std::string> muonTriggerNames, std::vector<std::string> muonTriggerObjectNames) {
```
fabiocos: @davidwalter2 @andrius-k do we really want to pass these two vectors of strings by value? Why not by reference? 
May be as const references if they need to stay unchanged (as I imagine)?

fabiocos:
+operations

the update of StandardSequence and Era is coherent with the purpose of the PR. I agree with the name update suggested by @jfernan2

DQMOffline/Lumi/src/TTrigger.cc
davidwalter2: Ok thanks, I did another commit

fabiocos:
please test

fioriNTU:
+1

fabiocos:
+1

PR #26363, opened 2019-04-04
Hcal depth energy fraction packed candidates 10 6

hatakeyamak:
#### PR description:

Includes HCAL depth information for packed candidates in MINIAOD for high-level object use (e.g. PUPPI recomputation at the post ultra-legacy MINIAOD stage). The promising results are obtained [1,2], but needs more time to finalize it. It means it is important to have this information in packedCandidates, so that the PUPPI weights can be recomputed on-the-fly.

[1] https://indico.cern.ch/event/799365/contributions/3333038/attachments/1805670/2946660/pre.pdf
[2] https://indico.cern.ch/event/799365/contributions/3339621/attachments/1805694/2946705/JK_JetMET_03_19_2.pdf

JME leaders, please feel free to add more supporting items!

@peruzzim @fgolf  @zdemirag @ahinzmann @bendavid @mariadalfonso 

#### PR validation:

Made sure that the this information gets added to packedCandidates as intended. The size increase in TTbar sample with 2018 scenario is 0.7% with this PR.

This is NOT a backport PR.

### DIFF
--- a/DataFormats/PatCandidates/interface/PackedCandidate.h
+++ b/DataFormats/PatCandidates/interface/PackedCandidate.h
@@ -44,7 +44,7 @@ namespace pat {
       packedPuppiweightNoLepDiff_(0),
       rawCaloFraction_(0),
       hcalFraction_(0),
-      //hcalDepthEnergyFractions_({0,0,0,0,0,0,0}),
+      hcalDepthEnergyFractions_({{0,0,0,0,0,0,0}}),
       packedTime_(0),
       packedTimeError_(0),
       isIsolatedChargedHadron_(false),
@@ -57,7 +57,7 @@ namespace pat {
                               const reco::VertexRefProd &pvRefProd,
                               reco::VertexRef::key_type pvRefKey) :
       packedPuppiweight_(0), packedPuppiweightNoLepDiff_(0), rawCaloFraction_(0), hcalFraction_(0),
-      //hcalDepthEnergyFractions_({0,0,0,0,0,0,0}),
+      hcalDepthEnergyFractions_({{0,0,0,0,0,0,0}}),
       packedTime_(0), packedTimeError_(0), isIsolatedChargedHadron_(false),
       p4_( new PolarLorentzVector(c.pt(), c.eta(), c.phi(), c.mass())), 
       p4c_( new LorentzVector(*p4_)), vertex_( new Point(c.vertex())), 
@@ -73,7 +73,7 @@ namespace pat {
                               const reco::VertexRefProd &pvRefProd,
                               reco::VertexRef::key_type pvRefKey) :
       packedPuppiweight_(0), packedPuppiweightNoLepDiff_(0), rawCaloFraction_(0), hcalFraction_(0),
-      //hcalDepthEnergyFractions_({0,0,0,0,0,0,0}),
+      hcalDepthEnergyFractions_({{0,0,0,0,0,0,0}}),
       packedTime_(0), packedTimeError_(0), isIsolatedChargedHadron_(false),
       p4_( new PolarLorentzVector(p4) ), p4c_( new LorentzVector(*p4_)), 
       vertex_( new Point(vtx) ), 
@@ -91,7 +91,7 @@ namespace pat {
                               const reco::VertexRefProd &pvRefProd,
                               reco::VertexRef::key_type pvRefKey) :
       packedPuppiweight_(0), packedPuppiweightNoLepDiff_(0), rawCaloFraction_(0), hcalFraction_(0),
-      //hcalDepthEnergyFractions_({0,0,0,0,0,0,0}),
+      hcalDepthEnergyFractions_({{0,0,0,0,0,0,0}}),
       packedTime_(0), packedTimeError_(0), isIsolatedChargedHadron_(false),
       p4_(new PolarLorentzVector(p4.Pt(), p4.Eta(), p4.Phi(), p4.M())), 
       p4c_( new LorentzVector(p4)), vertex_( new Point(vtx) ) ,
@@ -652,8 +652,8 @@ namespace pat {
     float rawCaloFraction() const { return (rawCaloFraction_/100.); }    /// Raw ECAL+HCAL energy over candidate energy for isolated charged hadrons
     void setHcalFraction(float p);                      /// Set the fraction of Hcal needed for HF and neutral hadrons and isolated charged hadrons
     float hcalFraction() const { return (hcalFraction_/100.); }    /// Fraction of Ecal and Hcal for HF and neutral hadrons and isolated charged hadrons
-    void setHcalDepthEnergyFractions(std::array<float,7> p);                      /// set fraction of Hcal energies in each depth
-    float hcalDepthEnergyFraction(int iDepth) const { return (hcalDepthEnergyFractions_[iDepth]/100.); }    /// Fraction of Hcal energies in each depth
+    void setHcalDepthEnergyFractions(const std::array<float,7> & fracs);                      /// set fraction of Hcal energies in each depth
+    float hcalDepthEnergyFraction(unsigned int depth) const { return (hcalDepthEnergyFractions_[depth-1]/100.); }    /// Fraction of Hcal energies in each depth
 
      // isolated charged hadrons
     void setIsIsolatedChargedHadron(bool p);                      /// Set isolation (as in particle flow, i.e. at calorimeter surface rather than at PV) flat for charged hadrons
--- a/DataFormats/PatCandidates/interface/PackedCandidate.h
+++ b/DataFormats/PatCandidates/interface/PackedCandidate.h
@@ -653,7 +653,7 @@ namespace pat {
     void setHcalFraction(float p);                      /// Set the fraction of Hcal needed for HF and neutral hadrons and isolated charged hadrons
     float hcalFraction() const { return (hcalFraction_/100.); }    /// Fraction of Ecal and Hcal for HF and neutral hadrons and isolated charged hadrons
     void setHcalDepthEnergyFractions(const std::array<float,7> & fracs);                      /// set fraction of Hcal energies in each depth
-    float hcalDepthEnergyFraction(unsigned int depth) const { return (hcalDepthEnergyFractions_[depth-1]/100.); }    /// Fraction of Hcal energies in each depth
+    float hcalDepthEnergyFraction(unsigned int depth) const { return (hcalDepthEnergyFractions_[depth-1]/200.); }    /// Fraction of Hcal energies in each depth
 
      // isolated charged hadrons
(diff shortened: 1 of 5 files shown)


perrotta:
please test

hatakeyamak:
I think this commit above will fix the matrix test failures for 2017 and earlier scenarios.

perrotta:
please test

zdemirag:
Adding some slides from JME side for motivation purposes: 
[he_depth_reco.pdf](https://github.com/cms-sw/cmssw/files/3056535/he_depth_reco.pdf)

perrotta:
This PR adds some 2 kB to packedPFCandidates in the miniAOD output of the 11024 TTbar PU workflow (measured with 100 events, perhaps packing not optimal with such a small sample):
```
Compare packed values
    785.5 ->       794.1          9      1.1   0.01     patPackedCandidates_lostTracks__RECO.
  29804.1 ->     31890.7       2087      6.8   2.54     patPackedCandidates_packedPFCandidates__RECO.
    82060 ->       84154       2095             2.6     ALL BRANCHES
```

This represents the 2.6% of the total miniAOD size, and it is probably an acceptable load for granting the possibility of recomputing high leve objects starting from the (UL) miniAOD.

hatakeyamak:
Thank you @perrotta . I have to think why this increase estimate is larger than I estimated, but I think the expected benefit to allow recomputing high level objects starting from the (UL) miniAOD warrants this addition.

peruzzim:
@perrotta I am not sure I agree with the above.
@hatakeyamak can you please clarify for which candidates you are adding the fractions?
It was discussed and agreed to do it only for endcaps for Run2, with optimized precision - is this still the case?

In my opinion, the much larger-than-expected increase in size deserves further investigation before merging.
All this, without mentioning the fact that adding inputs that cost in terms of disk size, to be used to re-tune an algorithm that was not finalized in time, should really not be our operating mode. It would really not scale to all object developments in CMS ("if everybody adds 2%...").
This item has been discussed months, we agreed on a pragmatic solution based on different figures. This PR comes very late and changes the picture, to my eyes.

perrotta:
Thank you @peruzzim , this was the kind of input that I needed to help me concluding the review.

Up to now, this is not yet restricted to endcaps for the whole Run2, see
https://github.com/cms-sw/cmssw/pull/26363/files#diff-169aa967705aa8a91dbf632d86d37282R24
but only for Run2_2018
https://github.com/cms-sw/cmssw/pull/26363/files#diff-169aa967705aa8a91dbf632d86d37282R40

On the other hand, for all eras but Run_2018 the list of particle ID to be looked for is empty
https://github.com/cms-sw/cmssw/pull/26363/files#diff-169aa967705aa8a91dbf632d86d37282R23
and therefore I'd expect that layer fractions will not be saved in that case:
https://github.com/cms-sw/cmssw/pull/26363/files#diff-59c122468981a3091ca37a6f82fe5ae0R329

Please @hatakeyamak check and confirm what above, and also report about how did you optimize the precision.

peruzzim:
Also, I would suggest repeating the size increase test with a larger number of events.

perrotta:
> Also, I would suggest repeating the size increase test with a larger number of events.

@hatakeyamak will you take care of it?
Please use wf 11024 as benchmark (TTbar PU for 2018)

hatakeyamak:
@peruzzim @perrotta thank you very much for your comments.

] It was discussed and agreed to do it only for endcaps for Run2, with optimized precision - is this still the case?
There is an era-dependent setting, and currently adding this information only for endcap for 2018 (i.e. after HE phase1 upgrade) with limited precision of uint8. Beyond Run 2, the addition is opened up for barrel as well right now, but if JME agrees, this setting for Run 3 and beyond can be adjusted. For 2017 and earlier, the type of particles for which the hcal depth info is stored is empty, so this info won't be stored.

@perrotta how do you get the above values you quote? If you can share your recipe, I can try it  on my ttbar 2018 samples. Previously I think I used ttbar 2018 without PU. I will try PU samples next time. Previously I just generated two miniaod with and without this PR and check the output MINIAOD file size. Is it good enough?

perrotta:
assign xpog

perrotta:
> @perrotta how do you get the above values you quote? If you can share your recipe, I can try it on my ttbar 2018 samples. Previously I think I used ttbar 2018 without PU. I will try PU samples next time. Previously I just generated two miniaod with and without this PR and check the output MINIAOD file size. Is it good enough?

@hatakeyamak 
For the evaluation of the timing, I use the recipe in https://twiki.cern.ch/twiki/bin/view/CMS/RecoIntegration#Compare_timing
As a benchmark, you should better use 2018 TTbar with PU, instead of TTbar without PU. You can speed up things if you start from some AOD sample of that kind and just re-miniAOD on it: then you con compare the sizes of the miniAOD outputs produced in the baseline and with this PR.

hatakeyamak:
@perrotta @peruzzim Okay. Thanks. I will be working on it today.

hatakeyamak:
2000 events MINIAOD size (in bytes) from relval ttbar 2018 PU samples - the hcal depth info is added for packed candidates in the endcap only:
PR: 141783228
Ref: 138057801
+2.7%

2000 events MINIAOD size (in bytes) from relval ttbar 2017 PU samples - the hcal depth info is not added for any cand type:
PR: 107497290
Ref: 107232246
+0.25%

So, for the 2018 case, this observation seems consistent with @perrotta. For the 2017 case. the hcal depth info is not added for any cand type, i.e. always zero, and compression seems to work ok.
Now, with the last commit, for Run 3 and beyond, hcal depth info will not be added for any cand type for Run3 and beyond, so this addition of hcal depth info is only for 2018 and endcap.

slava77:
> Now, with the last commit, for Run 3 and beyond, hcal depth info will not be added for any cand type, so this addition of hcal depth info is only for 2018 and endcap.

This and the previous description that this is used for debugging or that the final choice is quite unclear.
This to me is a good case for placing this new data into a value map, instead of modifying the data format of the PackedCandidate.

hatakeyamak:
> This and the previous description that this is used for debugging or that the final choice is quite unclear.
As is, the addition of hcal depth info to packedCandidates will be done only for 2018 scenario and endcap.

> This to me is a good case for placing this new data into a value map, instead of modifying the data format of the PackedCandidate.
If there is an example of how to do this, I can try this weekend, but right now I don't know how to do this.

peruzzim:
In this perspective, I would support Slava's suggestion about using a value map, without changing the data format.
A ~2.5% increase in MiniAOD size seems acceptable then, being confined to one year and campaign (please evaluate whether to use a modifier to activate this only for the UL, without carrying it over to future releases) and allowing to commission updated PUPPI with 2018 datasets - in preparation of Run3.

@hatakeyamk for the value map you can do something like e.g. https://github.com/cms-sw/cmssw/blob/de2816de91a7e625b5da6bd775cd1aebb6cbebe0/PhysicsTools/NanoAOD/plugins/IsoValueMapProducer.cc#L270

hatakeyamak:
Looking into it.
Thanks for the example This  example is for nanoAOD, so it's not completely trivial to me how to do this for packedCandidates. so I am still thinking.

peruzzim:
@hatakeyamak the value map does not "belong" to a specific object, it's a map between each element of a collection (e.g. electron in the linked example, packed candidate in your case) and a value (isolation in the linked example, fractions in your case).
The example runs for NanoAOD production, but has nothing to do with NanoAOD as a specific output format.
You should create the map in a producer, and then add it to the event and output event content, with a keep statement.

hatakeyamak:
I see. Thank you. Can I use an array (of 7 for each depth) with valueMap?

slava77:
On 4/12/19 1:15 PM, Kenichi Hatakeyama wrote:
> I see. Thank you. Can I use an array (of 7 for each depth) with valueMap?
> 

if an array worked inside a PackedCandidate,
it should also work in a ValueMap<array
However, you may want to define a small struct to handle
packing-unpacking from uint8.

hatakeyamak:
Just an update. I am trying this ValueMap method, but right now I am stuck with:

Exception Message:

  edm::ValueMap<std::array<unsigned char,7> >
  edm::Wrapper<edm::ValueMap<std::array<unsigned char,7> > >

when trying to use produces< edm::ValueMap<std::array<uint8_t,7>> > ("hcalDepthEnergyFractions");

Continuing the investigation, but advice will be appreciated.

slava77:
> when trying to use produces< edm::ValueMap<std::array<uint8_t,7>> > ("hcalDepthEnergyFractions");

you will need an entry in classes_def.xml
For a basic type as you describe it should probably be in https://github.com/cms-sw/cmssw/blob/master/DataFormats/Common/src/classes_def.xml#L148
If it's some simple struct/class that encapsulates the array and allows to convert from float to the uint8 and back, then this can stay in the package where you define it.
https://twiki.cern.ch/twiki/bin/view/CMSPublic/SWGuideCreatingNewProducts#Create_a_classes_def_xml_in_the

hatakeyamak:
Thank you, @slava77 .

Unfortunately, I haven't had much luck though. In order not to mix with this branch/PR, the new ValueMap-based one is developing under:
https://github.com/hatakeyamak/cmssw/tree/HcalDepthEnergyFraction_PackedCandidates_ValueMap
It's not quite at the stage to make a new PR.

(a) it currently adds HcalDepthEnergyFractions for seven depths as seven separate floats.
(b) i still need to check mapping between these fractions and packedCandidates is correct. When packedCandidates are made from PF candidates in PhysicsTools/PatAlgos/plugins/PATPackedCandidateProducer.cc, there is some reordering, and mapping might be affected.

I am still struggling to get array of uint8_t work. Though I added some classes in https://github.com/cms-sw/cmssw/blob/master/DataFormats/Common/src/classes_def.xml#L148, I still get the "No data dictionary found" error, and I am still not sure where this possible struct/class may go, as PhysicsTools/PatAlgos/plugins/PATPackedCandidateProducer.cc is in plugins, while I understand that the rule is classes_def.xml needs to be under src.

hatakeyamak:
I think I implemented most of what @slava77 suggested under
https://github.com/hatakeyamak/cmssw/tree/HcalDepthEnergyFraction_PackedCandidates_ValueMap
The remaining part is to define and implement simple struct/class that encapsulates the array and allows to convert from float to the uint8 and back. Will try it today. If it works, I will open a new PR.

hatakeyamak:
I think I followed what @slava77 and @peruzzim suggested i.e. using adding hcal depth info via ValueMap instead of adjusting data format. The change is in:
[0] https://github.com/cms-sw/cmssw/pull/26465
If this new one [0] looks better, we can probably close this PR.

slava77:
-1

superseded by #26465

hatakeyamak:
This is now superseded by https://github.com/cms-sw/cmssw/pull/26465. Closing.

PR #26364, opened 2019-04-04
use python3 compatible syntax in DQMOffline packages

davidlange6:
#### PR description:

adapt to 
- lack of xrange function in python 3
- keys() is now an iterator
- integer division is //

### DIFF
--- a/DQMOffline/L1Trigger/python/L1TEGammaOffline_cfi.py
+++ b/DQMOffline/L1Trigger/python/L1TEGammaOffline_cfi.py
@@ -1,14 +1,15 @@
+from builtins import range
 import FWCore.ParameterSet.Config as cms
 from DQMOffline.L1Trigger.L1THistDefinitions_cff import histDefinitions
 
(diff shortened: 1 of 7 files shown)


davidlange6:
please test

andrius-k:
+1

DQMOffline/L1Trigger/python/L1TEGammaOffline_cfi.py
```diff
@@ -1,14 +1,15 @@
+from builtins import range
```
kreczko: BTW, just out of curiosity: Why this import instead of using `six.moves.range` (as you've used `six` before)?

davidlange6: hi- I guess it was the first solution suggested by google in this case. Both solutions provide the eventual python3 behavior.

fabiocos:
@rekovic please check, it looks ok to me, I will move forward with the integration for next IB

fabiocos:
+1

fabiocos:
merge

PR #26369, opened 2019-04-05
Adding digi-based shower tagging to reco::Muon

battibass:
#### PR description:

This PR adds to the muon object a variable that counts the number of digis per station that are collected in the vicinity of an extrapolated track. Based on this variable, member functions are added to `reco::Muon` to tag showers based on digi information similarly to what is done , for example, [in this study](https://indico.cern.ch/event/804955/#2-follow-up-shower-tagging-met). 

Besides the addition of one more variable to `MuonChamberMatch` (which propagates up to miniAOD), no change to `reco::Muon` objects expected.
 
#### PR validation:

The consistency of the `reco::Muon::nDigisInStation` method with the digi counting made a posteriori in ntuples (used for studies like the ones linked above) has been tested on 1K events of a flat pT muon gun and is available [here](https://battilan.web.cern.ch/battilan/HighPt/PullRequest/hit_count/), in such sample. No discrepancy between the two sources of information was found. 

Event dumps were used to test that `reco::Muon::hasShowerInStation` and `reco::Muon::numberOfShowers` behave as expected.

The increase in event size, for slimmedMuons in miniAOD and for the workflow 136.891, was checked to be less than 1%.

### DIFF
--- a/DataFormats/MuonReco/interface/Muon.h
+++ b/DataFormats/MuonReco/interface/Muon.h
@@ -186,9 +186,6 @@ namespace reco {
     enum ArbitrationType { NoArbitration, SegmentArbitration, SegmentAndTrackArbitration, SegmentAndTrackArbitrationCleaned,
 			   RPCHitAndTrackArbitration, GEMSegmentAndTrackArbitration, ME0SegmentAndTrackArbitration };
 
-    /// ranges to be used for digi-based shower counting
-    enum DigiRange { FullChamber, InRange };
-    
     ///
     /// ====================== STANDARD SELECTORS ===========================
     ///
@@ -264,11 +261,13 @@ namespace reco {
     /// same as above for given number of sigmas
     unsigned int stationGapMaskPull( float sigmaCut = 3. ) const;
     /// # of digis in a given station layer
-    /// index 0-1-2-3 = DT stations 1-2-3-4                                                             /// index 4-5-6-7 = CSC stations 1-2-3-4
-    int nDigisInStation( int index, DigiRange range = InRange ) const;
-    /// # of digis in a given station layer
-    /// index 0-1-2-3 = DT stations 1-2-3-4                                                             /// index 4-5-6-7 = CSC stations 1-2-3-4
-    bool hasShowerInStation( int index, DigiRange range = InRange, int nDtDigisCut = 20, int nCscDigisCut = 36 ) const;
+    /// index 0-1-2-3 = DT stations 1-2-3-4                                                             
+    /// index 4-5-6-7 = CSC stations 1-2-3-4
+    int nDigisInStation( int index) const;
+    /// tag a shower in a given station layer
+    /// index 0-1-2-3 = DT stations 1-2-3-4
+    /// index 4-5-6-7 = CSC stations 1-2-3-4
+    bool hasShowerInStation( int index, int nDtDigisCut = 20, int nCscDigisCut = 36 ) const;
 
     /// muon type - type of the algorithm that reconstructed this muon
     /// multiple algorithms can reconstruct the same muon
--- a/DataFormats/MuonReco/interface/Muon.h
+++ b/DataFormats/MuonReco/interface/Muon.h
@@ -268,6 +268,8 @@ namespace reco {
     /// index 0-1-2-3 = DT stations 1-2-3-4
     /// index 4-5-6-7 = CSC stations 1-2-3-4
     bool hasShowerInStation( int index, int nDtDigisCut = 20, int nCscDigisCut = 36 ) const;
+    /// count the number of showers along a muon track
+    int numberOfShowers( int nDtDigisCut = 20, int nCscDigisCut = 36 ) const;
 
     /// muon type - type of the algorithm that reconstructed this muon
     /// multiple algorithms can reconstruct the same muon
--- a/DataFormats/MuonReco/interface/Muon.h
+++ b/DataFormats/MuonReco/interface/Muon.h
@@ -185,6 +185,9 @@ namespace reco {
 
     enum ArbitrationType { NoArbitration, SegmentArbitration, SegmentAndTrackArbitration, SegmentAndTrackArbitrationCleaned,
 			   RPCHitAndTrackArbitration, GEMSegmentAndTrackArbitration, ME0SegmentAndTrackArbitration };
+
+    /// ranges to be used for digi-based shower counting
+    enum DigiRange { FullChamber, InRange };
     
     ///
     /// ====================== STANDARD SELECTORS ===========================
@@ -260,7 +263,13 @@ namespace reco {
     unsigned int stationGapMaskDistance( float distanceCut = 10. ) const;
     /// same as above for given number of sigmas
     unsigned int stationGapMaskPull( float sigmaCut = 3. ) const;
-     
+    /// # of digis in a given station layer
+    /// index 0-1-2-3 = DT stations 1-2-3-4                                                             /// index 4-5-6-7 = CSC stations 1-2-3-4
+    int nDigisInStation( int index, DigiRange range = InRange ) const;
+    /// # of digis in a given station layer
+    /// index 0-1-2-3 = DT stations 1-2-3-4                                                             /// index 4-5-6-7 = CSC stations 1-2-3-4
+    bool hasShowerInStation( int index, DigiRange range = InRange, int nDtDigisCut = 20, int nCscDigisCut = 36 ) const;
+
     /// muon type - type of the algorithm that reconstructed this muon
     /// multiple algorithms can reconstruct the same muon
     static const unsigned int GlobalMuon     =  1<<1;
--- a/DataFormats/MuonReco/interface/Muon.h
+++ b/DataFormats/MuonReco/interface/Muon.h
@@ -261,13 +261,9 @@ namespace reco {
     /// same as above for given number of sigmas
     unsigned int stationGapMaskPull( float sigmaCut = 3. ) const;
     /// # of digis in a given station layer
-    /// index 0-1-2-3 = DT stations 1-2-3-4                                                             
-    /// index 4-5-6-7 = CSC stations 1-2-3-4
-    int nDigisInStation( int index) const;
+    int nDigisInStation( int station, int muonSubdetId) const;
     /// tag a shower in a given station layer
-    /// index 0-1-2-3 = DT stations 1-2-3-4
-    /// index 4-5-6-7 = CSC stations 1-2-3-4
-    bool hasShowerInStation( int index, int nDtDigisCut = 20, int nCscDigisCut = 36 ) const;
+    bool hasShowerInStation( int station, int muonSubdetId, int nDtDigisCut = 20, int nCscDigisCut = 36 ) const;
     /// count the number of showers along a muon track
     int numberOfShowers( int nDtDigisCut = 20, int nCscDigisCut = 36 ) const;
 
--- a/DataFormats/MuonReco/src/Muon.cc
+++ b/DataFormats/MuonReco/src/Muon.cc
@@ -385,21 +385,20 @@ int Muon::nDigisInStation( int station, int muonSubdetId ) const
 
 bool Muon::hasShowerInStation( int station, int muonSubdetId, int nDtDigisCut, int nCscDigisCut ) const
 {
-  bool hasShower = muonSubdetId == MuonSubdetId::DT ? 
-    nDigisInStation(station,muonSubdetId) >= nDtDigisCut :
-    nDigisInStation(station,muonSubdetId) >= nCscDigisCut;
+  if (muonSubdetId != MuonSubdetId::DT || muonSubdetId != MuonSubdetId::CSC) return false;
+  auto nDigisCut = muonSubdetId == MuonSubdetId::DT ? nDtDigisCut : nCscDigisCut;
 
-  return hasShower;   
+  return nDigisInStation(station,muonSubdetId) >= nDigisCut ;   
 }
 
 int Muon::numberOfShowers( int nDtDigisCut, int nCscDigisCut ) const
 {
   int nShowers = 0;
-  for ( int iCh = 1; iCh < 5; ++iCh )
+  for ( int station = 1; station < 5; ++station )
     {
-      if ( hasShowerInStation(iCh,MuonSubdetId::DT,nDtDigisCut,nCscDigisCut) )
+      if ( hasShowerInStation(station,MuonSubdetId::DT,nDtDigisCut,nCscDigisCut) )
 	nShowers++;
-      if ( hasShowerInStation(iCh,MuonSubdetId::CSC,nDtDigisCut,nCscDigisCut) )
+      if ( hasShowerInStation(station,MuonSubdetId::CSC,nDtDigisCut,nCscDigisCut) )
 	nShowers++;
     }
 
--- a/DataFormats/MuonReco/src/Muon.cc
+++ b/DataFormats/MuonReco/src/Muon.cc
@@ -334,7 +334,7 @@ unsigned int Muon::stationGapMaskPull( float sigmaCut ) const
    return totMask;
 }
 
-int Muon::nDigisInStation( int index, DigiRange range ) const
+int Muon::nDigisInStation( int index ) const
 {
   int nDigis(0);
   std::map<int, int> me11DigisPerCh;
@@ -345,7 +345,7 @@ int Muon::nDigisInStation( int index, DigiRange range ) const
 	   match.detector() != MuonSubdetId::DT  )
 	continue;
 	  
-      int nDigisInCh = range == InRange ? match.nDigisInRange : match.nDigisInChamb;
+      int nDigisInCh = match.nDigisInRange;
       int iStation = match.detector() == MuonSubdetId::CSC ? index - 3 : index + 1;
 
       if( match.detector() == MuonSubdetId::CSC && iStation == 1)
@@ -381,11 +381,11 @@ int Muon::nDigisInStation( int index, DigiRange range ) const
   return nDigis;
 }
 
-bool Muon::hasShowerInStation( int index, DigiRange range, int nDtDigisCut, int nCscDigisCut ) const
+bool Muon::hasShowerInStation( int index, int nDtDigisCut, int nCscDigisCut ) const
 {
   bool hasShower = index < 4 ? 
-			   nDigisInStation(index, range) >= nDtDigisCut :
-                           nDigisInStation(index, range) >= nCscDigisCut;
+			   nDigisInStation(index) >= nDtDigisCut :
+                           nDigisInStation(index) >= nCscDigisCut;
   return hasShower;   
 }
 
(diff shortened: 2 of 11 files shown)


perrotta:
please test

fabiocos:
please test

smuzaffar:
please test
cms-bot PR https://github.com/cms-sw/cms-bot/pull/1115 should fix this

fabiocos:
@smuzaffar I think that the latest failure is unrelated to the issue with the bot

slava77:
the error in the unit test is :

```
   [5] Calling method for module MuonIdProducer/'muonsFromCosmics'
Exception Message:
Looking for type: MuonDigiCollection<DTLayerId,DTDigi>
Looking for module label: muonDTDigis

```

this looks related to this PR.

battibass:
Hi, the problem with `testTauEmbeddingProducers` should be fixed now.
As in `TauAnalysis/MCEmbeddingTools/python/customisers.py` the filling of other shower-related information in muon objects was already disabled (`process.muons.FillShoweringInfo = cms.bool(False)`), I simply switched off the DIGI-based filling of shower information as well.
Please let me know if this is OK.

civanch:
please test

civanch:
+1

DataFormats/MuonReco/interface/Muon.h
```diff
@@ -260,7 +260,17 @@ namespace reco {
     unsigned int stationGapMaskDistance( float distanceCut = 10. ) const;
     /// same as above for given number of sigmas
     unsigned int stationGapMaskPull( float sigmaCut = 3. ) const;
-     
+    /// # of digis in a given station layer
+    /// index 0-1-2-3 = DT stations 1-2-3-4                                                             
+    /// index 4-5-6-7 = CSC stations 1-2-3-4
+    int nDigisInStation( int index) const;
```
slava77: is there a plan to add RPC or GEMs?
perhaps a function with arguments ```MuonSubdetId, int station``` is more practical.

RecoMuon/MuonIdentification/plugins/MuonIdProducer.cc
```diff
@@ -876,6 +888,14 @@ void MuonIdProducer::fillMuonId(edm::Event& iEvent, const edm::EventSetup& iSetu
      matchedChamber.edgeY = chamber.localDistanceY;
 
      matchedChamber.id = chamber.id;
+
+     if (fillShowerDigis_) {
+       theShowerDigiFiller_->fill(matchedChamber);
+     }
+     else {
+       matchedChamber.nDigisInRange = 0;
```
slava77: The pattern above looks like you want to encapsulate setting of potentially more than one parameter; if that's the case,
perhaps ```theShowerDigiFiller_->fillDefault``` is more appropriate.
If there is no plan to add more variables, and only nDigisInRange is going to be computed, then instead of a ```fill(matchedChamber)`` there better be just nDigisInRange computation

RecoMuon/MuonIdentification/src/MuonShowerDigiFiller.cc
```diff
@@ -0,0 +1,153 @@
...
+//
+// constructors and destructor
+//
+
+MuonShowerDigiFiller::MuonShowerDigiFiller(const edm::ParameterSet& iConfig, edm::ConsumesCollector&& iC) : 
+  m_digiMaxDistanceX(iConfig.getParameter<double>("digiMaxDistanceX"))
+
+{
+
+  edm::InputTag label = iConfig.getParameter<edm::InputTag>("dtDigiCollectionLabel");
```
slava77: ```Label``` could be omitted and this line can also be joined with the next without a named temporary ```label`` variable.

RecoMuon/MuonIdentification/src/MuonShowerDigiFiller.cc
```diff
@@ -0,0 +1,153 @@
...
+  label = iConfig.getParameter<edm::InputTag>("cscDigiCollectionLabel");
+  m_cscDigisToken=iC.consumes<CSCStripDigiCollection>(label);
+   
+}
+
+
+MuonShowerDigiFiller::~MuonShowerDigiFiller()
+{
+
+}
```
slava77: there is no need to define an empty destructor

RecoMuon/MuonIdentification/src/MuonShowerDigiFiller.cc
```diff
@@ -0,0 +1,153 @@
...
+	}
+    }
+
+  else if(muChMatch.detector() == MuonSubdetId::CSC) 
+    {
+
+      double xTrack = muChMatch.x;
+      double yTrack = muChMatch.y;
+
+      for (int iLayer = 1; iLayer < 7; ++iLayer)
```
slava77: ```<= CSCDetId::MAX_LAYER``` instead of the magic numbers

RecoMuon/MuonIdentification/src/MuonShowerDigiFiller.cc
```diff
@@ -0,0 +1,153 @@
...
+  int nDigisInRange = 0;
+
+  // DT chamber
+  if( muChMatch.detector() == MuonSubdetId::DT ) 
+    {
+      double xTrack = muChMatch.x;
+      
+      for (int sl = 1; sl < 4; sl += 2) 
+	{
+	  for (int layer = 1; layer < 5; ++layer)
```
slava77: please use ```DTChamberId::maxLayerId ``` and ```maxSuperLayerId``` instead of the magic numbers

RecoMuon/MuonIdentification/src/MuonShowerDigiFiller.cc
```diff
@@ -0,0 +1,153 @@
...
+	  
+	  auto range =  m_cscDigis->get(layerId);
+	  
+	  for (auto digiIt = range.first ;digiIt!=range.second; ++digiIt) 
+	    {
+	    
+	      std::vector<int> adcVals = digiIt->getADCCounts();
+	      bool hasFired = false;
+	      float pedestal = 0.5*(float)(adcVals[0]+adcVals[1]);
+	      float threshold = 13.3 ;
```
slava77: should this be configurable as other selections?
If 13.3 has an origin in some calibrations, better use from the source

DataFormats/MuonReco/interface/Muon.h
battibass: I would personally not add RPC or GEM. One can disentangle the signature a shower w.r.t. some channel noise much better in a detector with many layers. My only doubt would be ME0 let me briefly discuss with MUO about this.
Anyhow OK, I will switch to a version getting `MuonSubdetId` as argument in any case.

RecoMuon/MuonIdentification/plugins/MuonIdProducer.cc
battibass: In the offline studies linked as slides at the beginning of the PR I had 4 ranges. I initially implemented this code thinking about 2 ranges, then I switched to one for simplicity. Let me briefly discuss with MUO and, depending on the outcome of the discussion I will change things accordingly.

RecoMuon/MuonIdentification/src/MuonShowerDigiFiller.cc
battibass: OK

RecoMuon/MuonIdentification/src/MuonShowerDigiFiller.cc
battibass: OK

RecoMuon/MuonIdentification/src/MuonShowerDigiFiller.cc
battibass: OK

RecoMuon/MuonIdentification/src/MuonShowerDigiFiller.cc
battibass: OK

RecoMuon/MuonIdentification/src/MuonShowerDigiFiller.cc
battibass: I got this from CSC DPG, so I have no answer, let me check with them.

ptcox: Yes, this is historical and arbitrary. It's the signal-above-pedestal threshold considered to be an actual hit rather than a pedestal fluctuation.  It's hard-coded as 13.3 in several places in CSCValidation and CSCOfflineMonitor (part of CSC DQM). This is adc counts, and a very small number relative to the dynamic range of a CSC adc channel (maximum - i.e. saturation - at 4095 counts.)
Should really be configurable because seeing it makes me realize this is something we might like to adjust if pedestal noise increases at higher luminosities (but that's irrelevant for Carlo's purposes here.)

slava77:
@battibass 
please clarify on the status of updates in this PR following the code review.
Thank you.

battibass:
@slava77 I will push an update by the end of the day (running one last test now).

RecoMuon/MuonIdentification/plugins/MuonIdProducer.cc
battibass: I discussed with @Fedespring and we agreed that, depending on the development of the studies being performed for MUO-17-001, the plan is very likely to extend this shower filler (potentially also replacing `MuonShowerInformationFiller`), hence I adopted the solution of adding a `theShowerDigiFiller_->fillDefault()` method.

slava77:
@cmsbuild please test

civanch:
+1

DataFormats/MuonReco/src/Muon.cc
```diff
@@ -334,6 +334,78 @@ unsigned int Muon::stationGapMaskPull( float sigmaCut ) const
...
+	nDigis = nMe11DigisInCh;
+    }
+  
+  return nDigis;
+}
+
+bool Muon::hasShowerInStation( int station, int muonSubdetId, int nDtDigisCut, int nCscDigisCut ) const
+{
+  bool hasShower = muonSubdetId == MuonSubdetId::DT ? 
+    nDigisInStation(station,muonSubdetId) >= nDtDigisCut :
```
slava77:
```c++
if (muonSubdetId != MuonSubdetId::DT || muonSubdetId != MuonSubdetId::CSC) return false;
auto cut = muonSubdetId == MuonSubdetId::DT ? nDtDigisCut : nCscDigisCut;
return nDigisInStation(station,muonSubdetId) >= cut;
```
is perhaps shorter and more explicit about DT vs CSC.

DataFormats/MuonReco/src/Muon.cc
```diff
@@ -334,6 +334,78 @@ unsigned int Muon::stationGapMaskPull( float sigmaCut ) const
...
+    nDigisInStation(station,muonSubdetId) >= nCscDigisCut;
+
+  return hasShower;   
+}
+
+int Muon::numberOfShowers( int nDtDigisCut, int nCscDigisCut ) const
+{
+  int nShowers = 0;
+  for ( int iCh = 1; iCh < 5; ++iCh )
```
slava77: to reduce possible confusion when reading this, perhaps ```iSt``` instead of ```iCh``` because it's supposed to be a shorthand for a station

battibass: Changed in `station` (more consistent with the member functions above).

DataFormats/MuonReco/src/Muon.cc
battibass: Right.

battibass: But the OR should be an AND.

slava77:
@cmsbuild please test

slava77:
@cmsbuild please abort test

slava77:
@cmsbuild please test

slava77:
@smuzaffar @mrodozov 

I had a request to abort  in https://github.com/cms-sw/cmssw/pull/26369#issuecomment-483736071 ,  which apparently was ignored.
Is this a normal behavior?

slava77:
+1

for https://github.com/cms-sw/cmssw/pull/26369	d719e62
    - I could not find a problem in the unitTest log file and assume that the error reported in https://github.com/cms-sw/cmssw/pull/26369#issuecomment-483887426 is unrelated; the earlier version of this PR (very similar in content) was tested without failures in https://github.com/cms-sw/cmssw/pull/26369#issuecomment-483473744
    - I added a basic plot for ```muMatches_.nDigisInRange```, which should show up in the next update of the reco comparisons for the bot soon.

fabiocos:
@mrodozov it looks like the bot is stuck in showing a unit test failure that is no more there, ideas?

davidlange6:
its not a problem of the bot being stuck - it thinks there is a unit test failure. Perhaps one should just restart the tests?

fabiocos:
@civanch @santocch could you please have a look?

fabiocos:
please test

fabiocos:
@davidlange6 the point is that I do not see any failure in the report, let's try it again...

mrodozov:
@slava77 now I see your question, only "abort" or "please abort" should do. from time to time we have this github webhook failures and that might be it, but I'll assume the bot got confused from your command

mrodozov:
@fabiocos I'm searching in the logs and it says 
```
00:11:09 ++ grep -i 'had errors\|recipe for target' /build/cmsbld/jenkins/workspace/ib-any-integration/unitTests.log
00:11:09 ++ sed 's|'\''||g;s|.*recipe for target *||;s|.*unittests_|---> test |;s| failed$| timeout|'
00:11:09 + TEST_ERRORS=
00:11:09 ++ grep -i 'had errors' /build/cmsbld/jenkins/workspace/ib-any-integration/unitTests.log
00:11:09 + TEST_ERRORS=
00:11:09 + true
00:11:09 ++ grep ALL_OK /build/cmsbld/jenkins/workspace/ib-any-integration/unitTests.log
00:11:09 + GENERAL_ERRORS=
00:11:09 + true
00:11:09 Errors in the unit tests
00:11:09 + '[' X '!=' X -o X = X ']'
00:11:09 + echo 'Errors in the unit tests'
00:11:09 + echo 'UNIT_TEST_RESULTS;ERROR'
00:11:09 + ALL_OK=false
00:11:09 + UNIT_TESTS_OK=false
```
which in fact is not true, there are no test failing. 
this part :
```
TEST_ERRORS=
00:11:09 ++ grep -i 'had errors' /build/cmsbld/jenkins/workspace/ib-any-integration/unitTests.log
00:11:09 + TEST_ERRORS=
00:11:09 + true
```
that sets the TEST_ERRORS to true doesn't make any sense, greps for any error, there is none, yet still returns true. lets see if it persists.

fabiocos:
@civanch @santocch the SIM and ANA part of this PR is restricted to the Tau embedding configuration where a new flag introduced is set to False by default...

fabiocos:
@civanch given the latest changes I will take your previous signature as valid, anyway please check and sign it again for future reference, or comment in case

fabiocos:
+1

fabiocos:
merge

RecoMuon/MuonIdentification/python/MuonShowerDigiFiller_cfi.py
```diff
@@ -0,0 +1,11 @@
+import FWCore.ParameterSet.Config as cms
+
+MuonShowerDigiFillerBlock = cms.PSet(
+  ShowerDigiFillerParameters = cms.PSet(
+      digiMaxDistanceX = cms.double(25.0),
+      dtDigiCollectionLabel  = cms.InputTag("muonDTDigis"),
+      cscDigiCollectionLabel = cms.InputTag("muonCSCDigis","MuonCSCStripDigi")
```
fabiocos: @battibass @slava77 it looks like this line is creating issues in wf 1102.0 step4 RECOFROMRECO:

```
   [7] Prefetching for module TrackProducer/'mergedDuplicateTracks'
   [8] Prefetching for module DuplicateTrackMerger/'duplicateTrackCandidates'
   [9] Prefetching for module TrackCollectionMerger/'preDuplicateMergingGeneralTracks'
   [10] Prefetching for module TrackProducer/'muonSeededTracksInOut'
   [11] Prefetching for module CkfTrackCandidateMaker/'muonSeededTrackCandidatesInOut'
   [12] Prefetching for module MuonReSeeder/'muonSeededSeedsInOut'
   [13] Calling method for module MuonIdProducer/'earlyMuons'
Exception Message:
Looking for type: MuonDigiCollection&lt;CSCDetId,CSCStripDigi&gt;
Looking for module label: muonCSCDigis
Looking for productInstanceName: MuonCSCStripDigi

   Additional Info:

```

This clearly needs to be fixed, my question is: if this is not affecting any other standard workflow as it seems, and it is not preventing the basic validation of pre4, can we consider to postpone its fix after pre4 is built? Or do you see further possible issues?

Switching this module off in the configuration allows also the step4 of 1102 to run without problems.

battibass: @fabiocos , @slava77 can surely reply more accurately than I can. 
In principle this PR implies that digis are available (hence implies that unpacking was done starting from RAW). For any workflow for which this is the case there should be no problem.
The question is, are there RelVal workflows where this is not the case?
I'm happy to work on a patch asap, but I would need inputs to understand better how I should be testing it, as my check with `runTheMatrix.py -l limited -i all` clearly failed to spot this issue, as well as the one with `TauAnalysis/MCEmbeddingTools/` that was instead spot during integration ...

slava77:
On 4/19/19 4:18 AM, Carlo Battilana wrote:
> I'm happy to work on a patch asap, but I would need inputs to understand
> better how I should be testing it, as my check with |runTheMatrix.py -l
> limited -i all| clearly failed to spot this issue, as well as the one
> with |TauAnalysis/MCEmbeddingTools/| that was instead spot during
> integration ...

just rewording the earlier message from @fabiocos

runTheMatrix.py -l 1102.0

and then look for the problem in step4


The error does imply that this PR violated assumptions of RECO sequences
design, which apparently implies that digis are not used directly and
local reco outputs are enough to make everything downstream.

Please remind me the logic why it's the digis and not the calibrated
data (which supposedly can also remove the noisy channels)
dt1DRecHits and csc2DRecHits that are used?

battibass:
@slava77 I understand well that I can run `runTheMatrix.py -l 1102.0` my question was more general: is there a wider set of tests that I should perform? If so which one?

In principle, one could use dt1DRecHits, for the DTs. They come with left/right ambiguities (there are actually 2 recHits per digi), but it can be handled. For CSC I think the situation is more complex. From a discussion I had with @ptcox when starting all this, the interpretation of csc2DRecHits is not trivial as both clustering and combination of wire and strip digi information happens at RecHit building, and it is not straightforward to understand what the counting of csc2DRecHits actually means for cases beyond the ones where you have clear muon track segments. The use of CSC strip digis was in fact chosen as it should be a better proxy for "counting" the activity in a chamber in case a shower occur.

slava77:
as discussed in the RECO meeting, we can have a quick solution to simply add the CSC and DT digis to the RECO event content.
If there are other reasons to wait with the build of pre4, perhaps this solution can be provided today.

A better choice perhaps is to allow modifications to module PSets of the RECOfromRECO workflows (which is not possible now). See #26500.
We better complete this on the time scale of pre5.

RecoMuon/MuonIdentification/python/MuonShowerDigiFiller_cfi.py
fabiocos: @battibass indeed this is a bit special case, and only wf 1102.0 implementing it shows this issue. I see that @slava77 has already mentioned it in https://github.com/cms-sw/cmssw/issues/26500 .

santocch:
+1

PR #26298, opened 2019-03-29
Update DeepAK8 to 94X training

hqucms:
#### PR description:

This PR updates the DeepAK8 tagger to a new training performed with 2017 (94X) samples. Both the full (`pfDeepBoostedJetTags`) and the mass-decorrelated (`pfMassDecorrelatedDeepBoostedJetTags`) taggers are updated. 

Needs https://github.com/cms-data/RecoBTag-Combined/pull/21.

#### PR validation:

The update has been presented at the [BTV meeting](https://indico.cern.ch/event/808258/contributions/3366869/attachments/1817445/2971091/DeepAK8_BTV_Presentation_March25_PK.pdf). The new training shows improved performance on Phase1 detector compared to the 2016 training. The new training is also evaluated on 2016 samples and no loss of performance is found compared to the 2016 training. Therefore the updated model is intended for use in UL processing for all three years.

Small improvements have been made to the training afterwards to further reduce the mass sculpting. The performance of the latest version (in this PR) can be found [here](https://www.dropbox.com/s/lhhmjjnhayp29hm/deepAK8_PR_comp_20190329.pdf?dl=0). 

Outputs from CMSSW with this PR have been verified with the training framework.

@gouskos @pkontaxa

PR #26370, opened 2019-04-05
[102X] Update DeepAK8 to 94X training

hqucms:
#### PR description:

This PR updates the DeepAK8 tagger to a new training performed with 2017 (94X) samples. Both the full (pfDeepBoostedJetTags) and the mass-decorrelated (pfMassDecorrelatedDeepBoostedJetTags) taggers are updated.

Backport of https://github.com/cms-sw/cmssw/pull/26298 for analysis.
Needs https://github.com/cms-sw/cmsdist/pull/4844


perrotta:
backport of #26298

perrotta:
please test

hqucms:
@perrotta I think this needs to be tested with cms-sw/cmsdist#4844.

perrotta:
please abort test

perrotta:
@cmsbuild please test with cms-sw/cmsdist#4844.

perrotta:
@cmsbuild please test with cms-sw/cmsdist#4844.

hqucms:
@gouskos @pkontaxa FYI

mrodozov:
please test with cms-sw/cmsdist#4844

slava77:
assign xpog

this affects outputs of the standard nanoAOD production setup.
This is nominally not allowed and some configuration switch is needed to preserve the baseline behavior.
It looks like the code changes are implemented in a way that both the old and new training setups can not be supported with just the configuration changes.

Unless the XPOG can justify an exception to the policy in this case, this PR can not be accepted as is.
To be accepted according to the policy, there should be a way to control the behavior with the configuration; in that case the new training can be delivered with ```run2_miniAOD_devel``` modifier (this exists already) or something more nanoAOD specific.

peruzzim:
I share the worries of @slava77.
Can you please clarify whether it's really impossible to recover the old training when running from 102X? If so, it looks like a showstopper to me, as we are currently rerunning the tagger on top of MINIAOD, and both for NANOAOD production and for analysis one needs to be able to recover the previous behaviour. If parameters need updating, that should go along with the tagger version as well.

peruzzim:
Actually, similar arguments also apply to #26298. I think we should really try to avoid making quantities used for physics analysis in a certain release impossible to rerun from a later release, whenever possible (e.g. changes to versioned ID like this one).

slava77:
-1

it looks like there will be no exception to the no-change policy and a different solution should be provided if this is still needed for 10_2_X.

hqucms:
@peruzzim @slava77 
https://github.com/cms-sw/cmssw/pull/26370/commits/423c6328fe7120579b0e7386bd9a35c0e15c27c4 restores the previous training and implements the new training via the `run2_miniAOD_devel` modifier.

peruzzim:
please test with cms-sw/cmsdist#4844

RecoBTag/MXNet/plugins/DeepBoostedJetTagsProducer.cc
```diff
@@ -31,8 +31,8 @@ struct MXBlockCache {
 struct PreprocessParams {
   struct VarInfo {
     VarInfo() {}
-    VarInfo(float median, float upper) :
-      center(median), norm_factor(upper==median ? 1 : 1./(upper-median)) {}
```
peruzzim: I guess you changed the configuration parameters to obtain the same result as before, loading V01.
Numerically, though, there might be small differences because now you pass the (x,1/(x-y)) instead of (x,y).
[btw naively this looks strange... if upper is very close to median then 1/(upper-median) is >> 1 so it looks kind of discontinuous...]

Have you checked how much the output score is sensitive to this change?
Binned histogram tests with low statistics might not catch something like this...

slava77:
@cmsbuild please test with cms-sw/cmsdist#4844

RecoBTag/MXNet/plugins/DeepBoostedJetTagsProducer.cc
hqucms: @peruzzim Yes, the conversion in https://github.com/cms-sw/cmssw/pull/26370/commits/5eb90b9806dd5fde155449eb8045d55b35ebfa09 actually introduced some very small differences in the outputs in rare cases, as I did the conversion in python where floating point numbers are assumed to be double precision while in the previous C++ code single precision was used. https://github.com/cms-sw/cmssw/pull/26370/commits/860b74c6514f87686adffbf9247e744c50a169c9 re-did the conversion in single precision, and now the outputs match the ones using previous version of code exactly (based on 1000 events).

peruzzim:
@hqucms please port the same modifications, allowing to run V01, to master. Then we will come back to this backport. (just discussed at the ORP meeting)

hqucms:
@peruzzim Please  see https://github.com/cms-sw/cmssw/pull/26469.

slava77:
+1

for https://github.com/cms-sw/cmssw/pull/26370 	860b74c

- code changes are in line with the PR description and the follow up review. 
    - This is now a combination backport of  #26298 and  #26469 (both merged in the master). Additional changes were made to preserve the default behavior. The new features are available for tests using ```run2_miniAOD_devel``` modifier.

peruzzim:
+xpog

fabiocos:
+1

PR #26381, opened 2019-04-08
Avoid null reference in getTrackFromChargedHadron

Dr15Jones:
#### PR description:

Calling operator* on a edm:Ref can cause it to return a reference to a nullptr. This is not allowed in C++ which allows the compiler to remove checks. This lead to crashes when clang optimized the
code. Getting a pointer instead avoids the problem.

#### PR validation:

Ran workflow 4.23 which was crashing under CMSSW_10_6_CLANG_X_2019-04-07-0000 but now
runs fine with this change.


perrotta:
please test

slava77:
+1

for https://github.com/cms-sw/cmssw/pull/26381	ac6051b

fabiocos:
+1

PR #26382, opened 2019-04-08
New slopes for S9S1filter

toropin:
#### PR description:
This PR is devoted to the definition of new slopes for the limit lines in the planes 'S9S1 vs log(ELong)' for HF S9S1 filter. The slopes are different for the different HF Ieta values.
All parameters of S9S1(S8S1) filter were defined in ~2010 with lower energy of colliding beams and lower luminosity. It is the isolation filter and with much higher luminosity we have now the isolation
parameters could be different. See report in ref. 1 from slide 29.
On slide 30 presented distributions S9S1 vs log(ELong) for different Ieta. There are 2 inclined lines in each plot for Ieta 30 - 34. The lower lines correspond to the current slope parameters, the upper
lines correspond to the proposed new parameters.
The goal of this work is to reduce populations of hits above the old slope lines. These hits could be responsible for fake jets originated in HF.
On slides 34 and 35 presented the examples of HF hits which could be marked by S9S1 filter if slopes would be increased. In the lower lines in the captions to the pictures you can see S9S1 as the value calculated by filter, S9S1cut is the limit defined with old slopes, S9S1max is the new limit defined with changed slopes.   

References:
1. https://indico.cern.ch/event/807827/contributions/3362430/attachments/1816769/2969579/HFStripFilter_22Mar2019.pdf

#### PR validation:
In this PR only python file HFPhase1Reconstructor_cfi.py was modified. S9S1 slopes were multiplied by vector of constants to increase their values. These modifications are for Run 2 data. This file is for Run 2 reconstruction and this PR cannot affect data of Run 1.

The performance of reconstruction with the new slopes was tested. On the Screenshot of EventDisplay (FireWorks) below presented HF hit (signal in Long fiber) which was not marked by S9S1 filter. For these pictures threshold = 40 GeV was applied.

image

On the picture below presented the same hit marked by S9S1 filter after slopes were increased. In the column 'flags' uppermost hit has flag value 1048577. It means it has bit0 = 1 which corresponds to the S9S1 filter decision. 

image

### DIFF
--- a/RecoLocalCalo/HcalRecProducers/python/HFPhase1Reconstructor_cfi.py
+++ b/RecoLocalCalo/HcalRecProducers/python/HFPhase1Reconstructor_cfi.py
@@ -1,5 +1,15 @@
 import FWCore.ParameterSet.Config as cms
 
+# Slopes for the S9S1 filter
+_slopes_S9S1_run1 = [-99999,0.0164905,0.0238698,0.0321383,
+                     0.041296,0.0513428,0.0622789,0.0741041,
+                     0.0868186,0.100422,0.135313,0.136289,0.0589927]
+
+_coeffs = [1.0, 2.5, 2.2, 2.0, 1.8, 1.6, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0]
+
+_slopes_S9S1_run2 = [s*c for s, c in zip(_slopes_S9S1_run1, _coeffs)]
+
+
 hfreco = cms.EDProducer("HFPhase1Reconstructor",
     # Label for the input HFPreRecHitCollection
     inputLabel = cms.InputTag("hfprereco"),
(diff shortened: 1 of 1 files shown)


perrotta:
please test

perrotta:
@toropin : jenkins tests show no differences at all in Run2 outputs.
Is this due to the low statistics of those tests?
Do you have any higher stat validation of the update proposed, and can you put a link for it?

toropin:
Yes, we are looking for a level of ~10-4.
For the last tests I used events of MET trigger with special selection for S9S1 filter. To select these events the list of events with possible fake jets presented by JetMET group were used.
The picked events of MET trigger from runs 320757 - 321006 are in file
/afs/cern.ch/work/t/toropin/public/testS9S1/pickevents_S9S1.root (59 events).
In the same directory file myfile-S9S1-old.root contains 20 reconstructed events with the "old" slopes for S9S1 filter. In file myfile-S9S1-modified.root the same 20 events were reconstructed with "new" slopes for S9S1 filter.
Actually I compared these two reconstructions with EventDisplay.
For another triggers it will be quite difficult to indicate any difference. For SingleMuon trigger it is badly possible to find any difference at all.

perrotta:
+1
- New slopes for the filter defined according to the PR description and the linked presentations
- Physics validation was provided by the authors
- Jenkins tests pass and show no differences: differences may be expected for Run2 data, but at a much smaller rate than what can be addressed by the tests statistics

fabiocos:
+1

PR #26383, opened 2019-04-08
DD4hep: DD World Built from a DD4hep-based Geometry

ianna:
#### PR description:

* DDWorld to access G4 geometry
* Converter to G4 geometry test

#### PR validation:

@mrodozov - this PR will not compile until the dd4hep-geant4 tool is setup correctly. When it does, it can be merged.

no back port is needed

### DIFF
--- a/SimG4Core/DD4hepGeometry/interface/DD4hep_DDDWorld.h
+++ b/SimG4Core/DD4hepGeometry/interface/DD4hep_DDDWorld.h
@@ -1,0 +1,30 @@
+#ifndef SIMG4CORE_DD4HEP_DDDWORLD_H
+#define SIMG4CORE_DD4HEP_DDDWORLD_H
+
+#include "G4VPhysicalVolume.hh"
+
+namespace cms {
+
+  class DDDetector;
+  
+  class DDDWorld {
+
+  public:
+    DDDWorld(const cms::DDDetector*);
+    ~DDDWorld();
+    static void SetAsWorld(G4VPhysicalVolume* pv);
+    static void WorkerSetAsWorld(G4VPhysicalVolume* pv);
+    const G4VPhysicalVolume* GetWorldVolume() const { return m_world; }
+    
+    // In order to share the world volume with the worker threads, we
+    // need a non-const pointer. Thread-safety is handled inside Geant4
+    // with TLS. Should we consider a friend declaration here in order
+    // to avoid misuse?
+    G4VPhysicalVolume* GetWorldVolumeForWorker() const { return m_world; }
+
+  private:
+    G4VPhysicalVolume* m_world;
+  };
+}
+
+#endif
--- a/SimG4Core/DD4hepGeometry/plugins/DD4hep_DDTestG4Geometry.cc
+++ b/SimG4Core/DD4hepGeometry/plugins/DD4hep_DDTestG4Geometry.cc
@@ -1,0 +1,65 @@
+#include "FWCore/Framework/interface/one/EDAnalyzer.h"
+#include "FWCore/Framework/interface/MakerMacros.h"
+#include "FWCore/Framework/interface/ESTransientHandle.h"
+#include "FWCore/Framework/interface/EventSetup.h"
+#include "FWCore/MessageLogger/interface/MessageLogger.h"
+#include "DetectorDescription/DDCMS/interface/DetectorDescriptionRcd.h"
+#include "DetectorDescription/DDCMS/interface/DDDetector.h"
+#include "DetectorDescription/DDCMS/interface/DDVectorRegistryRcd.h"
+#include "DetectorDescription/DDCMS/interface/DDVectorRegistry.h"
+#include "DDG4/Geant4Converter.h"
+#include "DD4hep/Detector.h"
+
+#include <iostream>
+#include <string>
+
+using namespace std;
+using namespace cms;
+using namespace edm;
+using namespace dd4hep;
+
+class DDTestG4Geometry : public one::EDAnalyzer<> {
+public:
+  explicit DDTestG4Geometry(const ParameterSet&);
+
+  void beginJob() override {}
+  void analyze(Event const& iEvent, EventSetup const&) override;
+  void endJob() override {}
+
+private:  
+  const ESInputTag m_tag;
+  const string m_detElementPath;
+  const string m_placedVolPath;
+
+};
+
+DDTestG4Geometry::DDTestG4Geometry(const ParameterSet& iConfig)
+  : m_tag(iConfig.getParameter<ESInputTag>("DDDetector"))
+{}
+
+void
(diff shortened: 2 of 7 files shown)


mrodozov:
@ianna I fixed the include but now the code complains about "DDG4/Detector.h" not found in the DDG4 includes: 
https://github.com/cms-sw/cmssw/pull/26383/files#diff-19bce7dd0d1ec0083d7be74adfb4e755R6
is it DD4hep/Detector.h ?

ianna:
Thanks @mrodozov ! Please, let me know when can I test it.

mrodozov:
please test

mrodozov:
I've setup latest stuff from dd4hep and now it doesn't break as before but it's missing a file:
DetectorDescription/DDCMS/interface/DetectorDescriptionRcd.h No such file or directory
To have in mind. 

DD4hep update PR is here: https://github.com/cms-sw/cmsdist/pull/4859

ianna:
@mrodozov - yes, the PR needs a re-base

ianna:
please test

ianna:
assign simulation
assign geometry

ianna:
@fabiocos - this is a new package in Simulation with dependency on a dd4hep DDG4 static library.

civanch:
@ianna , I am a bit confused. If it is a test, why these are producers in plugin directory instead of test directory? What is the purpose of addition of watchers? What they intend to watch?

SimG4Core/DD4hepGeometry/interface/DD4hep_DDDWorld.h
```diff
@@ -0,0 +1,30 @@
...
+
+  class DDDetector;
+  
+  class DDDWorld {
+
+  public:
+    DDDWorld(const cms::DDDetector*);
+    ~DDDWorld();
+    static void SetAsWorld(G4VPhysicalVolume* pv);
+    static void WorkerSetAsWorld(G4VPhysicalVolume* pv);
```
civanch: @ianna , it is the same world, why making different methods WorkerSetAsWorld(G4VPhysicalVolume* pv) and GetWorldVolumeForWorker() ? I would remove both.

SimG4Core/DD4hepGeometry/plugins/DD4hep_DDTestG4Geometry.cc
```diff
@@ -0,0 +1,65 @@
...
+  g4Geo.debugElements = true;
+  g4Geo.debugShapes = true;
+  g4Geo.debugVolumes = true;
+  g4Geo.debugPlacements = true;
+  g4Geo.create(detector.world());
+  
+  LogVerbatim("Geometry") << "Done.";
+}
+
+DEFINE_FWK_MODULE(DDTestG4Geometry);
```
civanch: @ianna , I would move this to test from plugin.

SimG4Core/DD4hepGeometry/plugins/DD4hep_GeometryESProducer.cc
```diff
@@ -0,0 +1,221 @@
...
+class DDDWorld;
+class G4RunManagerKernel;
+class SimTrackManager;
+
+using namespace edm;
+using namespace std;
+using namespace cms;
+
+static
+void createWatchers(const ParameterSet& iP, SimActivityRegistry& iReg,
```
civanch: @ianna , why we need this?

SimG4Core/DD4hepGeometry/plugins/DD4hep_GeometryESProducer.cc
```diff
@@ -0,0 +1,221 @@
...
+      m_sensCaloDets.swap(sensDets.second);
+    }
+
+    LogInfo("G4GeometryESProducer") << " Sensitive Detector building finished; found " 
+				<< m_sensTkDets.size()
+				<< " Tk type Producers, and " << m_sensCaloDets.size() 
+				<< " Calo type producers ";
+  }
+
+  for(Producers::iterator itProd = m_producers.begin();itProd != m_producers.end();
```
civanch: @ianna , what is expected from this loop?

civanch:
@ianna , if this package depends on geant4core, than geant4core cannot depend on DD4Hep. We have circular dependency.  In the current situation, SimG4Core depends on DetectorDescription/Core. I would expect we will have the same thing. For today we have SimG4Core/Geometry/test which is intended for testing of geometry only. Why not reuse the same sub-package or creating a new one?

ianna:
This package is intended to test sim dependency on dd4hep. The DDDWorld is a copy of our DDDWorld with an exception that it is in a cms namespace. So is the other class. I was wondering myself about implementations and took liberty to drop some assignments which looked useless to me. 

If there’s a circular dependency, we have to address it.

My question is if the package affects the sim jobs in any way.

SimG4Core/DD4hepGeometry/interface/DD4hep_DDDWorld.h
ianna: Good question, I’d gladly remove the code

SimG4Core/DD4hepGeometry/plugins/DD4hep_DDTestG4Geometry.cc
ianna: Ok

SimG4Core/DD4hepGeometry/plugins/DD4hep_GeometryESProducer.cc
ianna: Good question

SimG4Core/DD4hepGeometry/plugins/DD4hep_GeometryESProducer.cc
ianna: Yet another good question:-)

civanch:
@ianna , current OscarMTProducer (which do real simulation) access would volume via old DD scheme of CMSSW. Your new code seems to provide an access to the world volume. For the test it is fine. It will be good to remove things, which are not connected with geometry in this new producer, this will ensure that we have no circular dependencies (after second look, I am not sure if we really have).

SimG4Core/DD4hepGeometry/interface/DD4hep_DDDWorld.h
ianna: @civanch - all plugins ought to be in a plugin directory unrelated of their purpose. The Sim code is been copied as is to make sure that comparison is fair. I'll do a thorough cleanup at the later stage.

ianna:
@civanch - there is no circular dependency. This package is needed to move further, e.g. to make a new OscarMTProducerFromDD4hep. However, it would help if it's integrated in an IB.

fabiocos:
@ianna @civanch this PR looks now entirely embedded into #26740, which is fully approved, so I would say it may be closed

ianna:
@fabiocos - yes, it's in #26740.

PR #26350, opened 2019-04-03
Add small changes in CondCore for PPS proton reco, and serialization …

wpcarvalho:
…test code for LHCOpticalFunctions

#### PR description:

This PR includes: 

- Minor changes needed for reading(writing) optical functions data from(into) POOL databases, consumed by the proton reconstruction introduced by the recently merged PR #25495. 

- A serialization test code for the condition formats `LHCOpticalFunctionsSet` and `LHCOpticalFunctionsCollection` introduced in #25495.

#### PR validation:

Tests run: just followed https://cms-sw.github.io/PRWorkflow.html.

PR #26394, opened 2019-04-09
PPS: proton reco finalisation before UL re-reco

jan-kaspar:
#### PR description:

This PR contains several finalisation changes for PPS proton reconstruction before the UL re-reco:
  * c5c4e47: removes the hardcoded IOV as promised in #25495
  * ff4097a: switches to CondDB as the default source of LHCInfo and optical functions.
  * 9d56bb7: cherry-pick of @tocheng 's commit 3aececc to include optical functions in auto GT.
  * bfe7a0a: PPS proton-reconstruction module does nothing if input empty. This prevents loading non-existing conditions in workflows without PPS proton data.

### DIFF
--- a/CalibPPS/ESProducers/plugins/CTPPSLHCInfoESSource.cc
+++ b/CalibPPS/ESProducers/plugins/CTPPSLHCInfoESSource.cc
@@ -53,14 +53,6 @@ CTPPSLHCInfoESSource::CTPPSLHCInfoESSource(const edm::ParameterSet& conf) :
 void CTPPSLHCInfoESSource::setIntervalFor(const edm::eventsetup::EventSetupRecordKey &key,
   const edm::IOVSyncValue& iosv, edm::ValidityInterval& oValidity)
 {
-  // TODO: the IOV specified in config is temporarily replaced with a hardcoded one
-  edm::IOVSyncValue beg(edm::EventID(270293, 0, 0), edm::Timestamp(1461016800ULL << 32));
-  edm::IOVSyncValue end(edm::EventID(290872, 0, 0), edm::Timestamp(1483138800ULL << 32));
-  oValidity = edm::ValidityInterval(beg, end);
-
-  m_insideValidityRange = (beg < iosv && iosv < end);
-
-  /*
   if (edm::contains(m_validityRange, iosv.eventID()))
   {
     m_insideValidityRange = true;
@@ -83,7 +75,6 @@ void CTPPSLHCInfoESSource::setIntervalFor(const edm::eventsetup::EventSetupRecor
       oValidity = edm::ValidityInterval(edm::IOVSyncValue(beginEvent), edm::IOVSyncValue::endOfTime());
     }
   }
-  */
 }
 
 //----------------------------------------------------------------------------------------------------
--- a/CalibPPS/ESProducers/python/ctppsLHCInfo_cff.py
+++ b/CalibPPS/ESProducers/python/ctppsLHCInfo_cff.py
@@ -1,11 +1,13 @@
 import FWCore.ParameterSet.Config as cms
 
-ctppsLHCInfoLabel = cms.string("ctpps_minimal")
+# by default, LHCInfo is now loaded from CondDB using a GT
+ctppsLHCInfoLabel = cms.string("")
 
-# minimal LHCInfo for 2016 data
-ctppsLHCInfoESSource_2016 = cms.ESSource("CTPPSLHCInfoESSource",
-  label = ctppsLHCInfoLabel,
-  validityRange = cms.EventRange("270293:min - 290872:max"),
-  beamEnergy = cms.double(6500),  # GeV
-  xangle = cms.double(185)  # murad
-)
+## minimal LHCInfo for 2016 data
+#ctppsLHCInfoLabel = cms.string("ctpps_minimal")
+#ctppsLHCInfoESSource_2016 = cms.ESSource("CTPPSLHCInfoESSource",
+#  label = ctppsLHCInfoLabel,
+#  validityRange = cms.EventRange("270293:min - 290872:max"),
+#  beamEnergy = cms.double(6500),  # GeV
+#  xangle = cms.double(185)  # murad
+#)
--- a/CalibPPS/ESProducers/python/ctppsOpticalFunctions_cff.py
+++ b/CalibPPS/ESProducers/python/ctppsOpticalFunctions_cff.py
@@ -2,26 +2,28 @@ import FWCore.ParameterSet.Config as cms
 
 from CalibPPS.ESProducers.ctppsLHCInfo_cff import *
 
-from CalibPPS.ESProducers.ctppsOpticalFunctionsESSource_cfi import *
+# by default, (raw) optical functions are now loaded from CondDB using a GT
 
-# add 2016 pre-TS2 configuration
-config_2016_preTS2 = cms.PSet(
-  validityRange = cms.EventRange("273725:min - 280385:max"),
-
-  opticalFunctions = cms.VPSet(
-      cms.PSet( xangle = cms.double(185), fileName = cms.FileInPath("CalibPPS/ESProducers/data/optical_functions_2016.root") )
-  ),
-
-  scoringPlanes = cms.VPSet(
-      # z in cm
-      cms.PSet( rpId = cms.uint32(0x76100000), dirName = cms.string("XRPH_C6L5_B2"), z = cms.double(-20382.6) ),  # RP 002, strip
-      cms.PSet( rpId = cms.uint32(0x76180000), dirName = cms.string("XRPH_D6L5_B2"), z = cms.double(-21255.1) ),  # RP 003, strip
-      cms.PSet( rpId = cms.uint32(0x77100000), dirName = cms.string("XRPH_C6R5_B1"), z = cms.double(+20382.6) ),  # RP 102, strip
-      cms.PSet( rpId = cms.uint32(0x77180000), dirName = cms.string("XRPH_D6R5_B1"), z = cms.double(+21255.1) ),  # RP 103, strip
-  )
-)
-
-ctppsOpticalFunctionsESSource.configuration.append(config_2016_preTS2)
+#from CalibPPS.ESProducers.ctppsOpticalFunctionsESSource_cfi import *
+#
+## add 2016 pre-TS2 configuration
+#config_2016_preTS2 = cms.PSet(
+#  validityRange = cms.EventRange("273725:min - 280385:max"),
+#
+#  opticalFunctions = cms.VPSet(
+#      cms.PSet( xangle = cms.double(185), fileName = cms.FileInPath("CalibPPS/ESProducers/data/optical_functions_2016.root") )
+#  ),
+#
+#  scoringPlanes = cms.VPSet(
+#      # z in cm
+#      cms.PSet( rpId = cms.uint32(0x76100000), dirName = cms.string("XRPH_C6L5_B2"), z = cms.double(-20382.6) ),  # RP 002, strip
+#      cms.PSet( rpId = cms.uint32(0x76180000), dirName = cms.string("XRPH_D6L5_B2"), z = cms.double(-21255.1) ),  # RP 003, strip
+#      cms.PSet( rpId = cms.uint32(0x77100000), dirName = cms.string("XRPH_C6R5_B1"), z = cms.double(+20382.6) ),  # RP 102, strip
+#      cms.PSet( rpId = cms.uint32(0x77180000), dirName = cms.string("XRPH_D6R5_B1"), z = cms.double(+21255.1) ),  # RP 103, strip
+#  )
+#)
+#
+#ctppsOpticalFunctionsESSource.configuration.append(config_2016_preTS2)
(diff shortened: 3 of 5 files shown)


jan-kaspar:
FYI @fabferro @forthommel

slava77:
@cmsbuild please test

jan-kaspar:
I have quickly checked some of the failures - the reason is missing ESSource for CTPPSOpticsRcd. This is currently expected and will be solved by #26350 (and the following DB upload).

tocheng:
@cmsbuild please test with #26415

tocheng:
@jan-kaspar could you please take.a.look at the error?
Looks like now the reason is not due to missing record when testing with PR #26415 
For e.g.,
https://cmssdt.cern.ch/SDT/jenkins-artifacts/pull-request-integration/PR-26394/34180/runTheMatrix-results/4.53_RunPhoton2012B+RunPhoton2012B+HLTD+RECODR1reHLT+HARVESTDR1reHLT/step3_RunPhoton2012B+RunPhoton2012B+HLTD+RECODR1reHLT+HARVESTDR1reHLT.log
https://cmssdt.cern.ch/SDT/jenkins-artifacts/pull-request-integration/PR-26394/34180/runTheMatrix-results/136.8311_RunJetHT2017F_reminiaod+RunJetHT2017F_reminiaod+REMINIAOD_data2017+HARVEST2017_REMINIAOD_data2017/step2_RunJetHT2017F_reminiaod+RunJetHT2017F_reminiaod+REMINIAOD_data2017+HARVEST2017_REMINIAOD_data2017.log

wpcarvalho:
> @jan-kaspar could you please take a look at the error?
> Looks like now the reason is not due to missing record when testing with PR #26415
> For e.g.,
> https://cmssdt.cern.ch/SDT/jenkins-artifacts/pull-request-integration/PR-26394/34180/runTheMatrix-results/4.53_RunPhoton2012B+RunPhoton2012B+HLTD+RECODR1reHLT+HARVESTDR1reHLT/step3_RunPhoton2012B+RunPhoton2012B+HLTD+RECODR1reHLT+HARVESTDR1reHLT.log
> https://cmssdt.cern.ch/SDT/jenkins-artifacts/pull-request-integration/PR-26394/34180/runTheMatrix-results/136.8311_RunJetHT2017F_reminiaod+RunJetHT2017F_reminiaod+REMINIAOD_data2017+HARVEST2017_REMINIAOD_data2017/step2_RunJetHT2017F_reminiaod+RunJetHT2017F_reminiaod+REMINIAOD_data2017+HARVEST2017_REMINIAOD_data2017.log

I had a look at the log file and the error is related to the "null" payloads for those time periods for which no optical functions are defined (run<273725, for instance). In such cases,  [CTPPSInterpolatedOpticalFunctionsESSource.cc](https://github.com/cms-sw/cmssw/blob/master/CalibPPS/ESProducers/plugins/CTPPSInterpolatedOpticalFunctionsESSource.cc) expects an empty (size=0) payload but mistakenly I created payloads with size=1, having only a dummy xAngle with value 999. I have already made a fix and can provide a new set of payloads, for exactly the same IOVs but correct size=0 payloads for IOVs with no optical functions defined. How should we proceed to implement this correction?

tocheng:
@wpcarvalho @jan-kaspar  Many thanks for the check. Please upload the tag to the DB and I will make new GTs for you. Btw, current the PR has conflict with the master branch. Please fix this issue if it matters for the test. Thanks.

wpcarvalho:
Many thanks @tocheng. What is the best approach, upload the fixed payloads with a new tag or with the previous one?

tocheng:
> Many thanks @tocheng. What is the best approach, upload the fixed payloads with a new tag or with the previous one?
Please upload a new tag.

tocheng:
> Many thanks @tocheng. What is the best approach, upload the fixed payloads with a new tag or with the previous one?

@wpcarvalho Please upload a new tag.

wpcarvalho:
@tocheng New tag **PPSOpticalFunctions_v2** uploaded.

tocheng:
@cmsbuild please test with #26415

jan-kaspar:
I am going to make the rebase and tests.

jan-kaspar:
Rebase done, I now proceed with tests and notify you once successfully passed.

tocheng:
> Rebase done, I now proceed with tests and notify you once successfully passed.

Hello @jan-kaspar , could you please test with PR #26415 ?
Thanks.

jan-kaspar:
> Hello @jan-kaspar , could you please test with PR #26415 ?

Yes, I've fetched your `autoCond.py` and do my tests together with your changes. Thanks for your reminder!

fabiocos:
please test with cms-sw/cmssw#26415

jan-kaspar:
> Yes, I've fetched your `autoCond.py` and do my tests together with your changes. Thanks for your reminder!

I've run PPS tests and the output corresponds to what is currently in the DB (non final payloads).

I've run also AddOn tests, yielding `42 tests passed, 5 failed`. The failing tests use `auto:run2_mc_*` and the error message reads
```
No "LHCInfoRcd" record found in the EventSetup.
```
The LHCInfo is consumed by the proton reconstruction module `CTPPSProtonProducer`. If confirmed, we either need to supply LHCInfo to MC workflows or disable the proton reconstruction in these workflows. What would you suggest?

tocheng:
> > Yes, I've fetched your `autoCond.py` and do my tests together with your changes. Thanks for your reminder!
> 
> I've run PPS tests and the output corresponds to what is currently in the DB (non final payloads).
> 
> I've run also AddOn tests, yielding `42 tests passed, 5 failed`. The failing tests use `auto:run2_mc_*` and the error message reads
> 
> ```
> No "LHCInfoRcd" record found in the EventSetup.
> ```
> 
> The LHCInfo is consumed by the proton reconstruction module `CTPPSProtonProducer`. If confirmed, we either need to supply LHCInfo to MC workflows or disable the proton reconstruction in these workflows. What would you suggest?

@jan-kaspar I think the question is whether you want the CTPPSProtonProducer to be used for ultra legacy MC production. If you do, then you need prepare all conditions needed for the producer. If not, we should turn off proton reconstruction for MC.

jan-kaspar:
> > The LHCInfo is consumed by the proton reconstruction module `CTPPSProtonProducer`. If confirmed, we either need to supply LHCInfo to MC workflows or disable the proton reconstruction in these workflows. What would you suggest?
> 
> @jan-kaspar I think the question is whether you want the CTPPSProtonProducer to be used for ultra legacy MC production. If you do, then you need prepare all conditions needed for the producer. If not, we should turn off proton reconstruction for MC.

AFAIK, there is no official PPS simulation yet - @fabferro please confirm. This would suggest to follow the latter option - turn off proton reconstruction for MC. Could someone please give me pointers how to achieve it? Thanks in advance!

jan-kaspar:
In 9d56bb7 I've cherry-picked @tocheng 's commit 3aececc288009ab2e18437cf43de9e4b68f619ed in order to decouple this PR and #26415 .

christopheralanwest:
please test

jan-kaspar:
I have checked (some of) the matrix-test failures and they all seem to be of same type as commented previously in https://github.com/cms-sw/cmssw/pull/26394#issuecomment-483300078. I suggest to remove the proton reconstruction from MC workflows. Could someone help and indicate how to do this, please? Thanks in advance!

tocheng:
@jan-kaspar  Can you ask help from RECO people?
I don't think announcing help in the github help. Only AlCa and Orp are included in the PR review.

jan-kaspar:
@slava77 @perrotta @fabiocos I'd like to ask you for suggestion: I'd like to disable PPS proton reconstruction in MC workflows, but I don't know what the best way is. With the current configuration, PPS reconstruction leads to failures in MC workflows since the needed conditions data are not included the relative GTs, see https://github.com/cms-sw/cmssw/pull/26394#issuecomment-483300078. Furthermore, at the moment PPS has no (official) simulation, thus the proton reconstruction would not produce anything anyway.

slava77:
I recall that the removal was mentioned a few times in the past; the conclusion was to keep the pps sequences in both data/MC configurations and find either a runtime or a GT solution to make it run.

jan-kaspar:
> I recall that the removal was mentioned a few times in the past; the conclusion was to keep the pps sequences in both data/MC configurations and find either a runtime or a GT solution to make it run.

This could well be - sorry for my short memory. My suggestion was motivated by the "shortest" path to the UL re-reco, but I have no problem with other solutions (allowing us to participated in the re-reco). What could be a "runtime solution"? I am not sure whether a GT solution could work - we are now uploading conditions with real-run IOVs. Looking in the failure logs I see "fake" run numbers, e.g. 1. So we would need to have these MC-like run numbers covered with (null) conditions - this would do the job.

slava77:
On 4/17/19 6:26 AM, jan-kaspar wrote:
> So we would need to have these MC-like run numbers covered with (null)
> conditions - this would do the job.


if a fake value is the only thing available, I guess that it can be
uploaded.
The fullsim sim side of the software is supposed to appear soon,
apparently, it will eventually need to have some non-null payloads in
the GT.

Where are the failures? Are they in the run/lumi begin/end or in the
::produce?
For the latter, there is always a safe way to check for empty inputs
first and not load the conditions.
A more nuclear option is to check for the run number or the
Event::isRealData flag,
but I prefer to avoid it in order to not open a branch in logic of
developments.

jan-kaspar:
>Where are the failures? Are they in the run/lumi begin/end or in the ::produce? For the latter, there is always a safe way to check for empty inputs first and not load the conditions.

Yes, it is in produce. I will try your suggestion - thanks!

jan-kaspar:
With bfe7a0a77fe, the proton reconstruction module first checks whether the input is empty. If it is, nothing is done - in particular no attempt to load conditions is made.

I've rerun some of the previously failing matrix tests (1306.0,158.0,25202.0,10224.0), yielding `4 4 4 4 1 tests passed, 0 0 0 0 0 failed`.

I've rerun the AddOn tests, yielding `47 tests passed, 0 failed`.

Could someone trigger tests please?

christopheralanwest:
please test

slava77:
+1

for 
- code changes are somewhat in line with the PR description and then align with the follow up review and discussion
    - please update the PR description
- jenkins tests pass and comparisons with the baseline show differences in somewhat expected places
    - there are quite a few changes in JET-related quantities related to the GT changes https://cms-conddb-prod.cern.ch/cmsDbBrowser/diff/Prod/gts/106X_dataRun2_v4/106X_dataRun2_v8
    - 2017F workflow 136.8311 shows that ctppsProtons reconstruction is now running (GT change includes an extension to the optical function range)
    - 2018A wf 136.85 shows some changes in the ctppsDiamondRecHits reco and CTPPS/TimingDiamond DQM plots; apparently in line with the GT changes as well
    - the change that actually triggered a reco signature is only in RecoCTPPS/ProtonReconstruction/plugins/CTPPSProtonProducer.cc : this change is technical, to skip the reco algorithm calls for empty inputs.

tocheng:
+1

fabiocos:
+1

jan-kaspar:
> * code changes are somewhat in line with the PR description and then align with the follow up review and discussion   
>   * please update the PR description

Description updated.

fabiocos:
@jan-kaspar @tocheng @slava77 unfortunately this PR is causing an issue in the IB:

```
----- Begin Fatal Exception 18-Apr-2019 11:23:17 CEST-----------------------
An exception of category 'ProtonReconstructionAlgorithm' occurred while
   [0] Processing  Event run: 314276 lumi: 161 event: 120565217 stream: 3
   [1] Running path 'Flag_trkPOG_logErrorTooManyClusters'
   [2] Prefetching for module LogErrorEventFilter/'logErrorTooManyClusters'
   [3] Prefetching for module LogErrorHarvester/'logErrorHarvester'
   [4] Calling method for module CTPPSProtonProducer/'ctppsProtons'
Exception Message:
Optics data not available for RP 2006974464, i.e. subDet=3 arm=1 station=2 rp=4.
```

in wf 136.8562 , the test for the high beta star runs. From the message it looks like the optics info for this run is not properly loaded and triggers the exception in

https://cmssdt.cern.ch/lxr/source/RecoCTPPS/ProtonReconstruction/src/ProtonReconstructionAlgorithm.cc#0151

Is this problem expected from the provided setup?

jan-kaspar:
@fabiocos Yes, this is expected. The optics currently uploaded to DB covers the full 2018 without excluding the special runs (high beta). These runs use different RPs than the low beta runs, that's why the message. We are aware of this problem and the new optics payloads, currently under test, exclude the two special-run periods:
  https://github.com/CTPPS/cmssw/blob/proton_reco_step2/CalibPPS/ESProducers/python/ctppsOpticalFunctions_cff.py#L85
For the special-run periods, we agreed with @wpcarvalho to upload null payloads, which will inhibit the proton reconstruction module, thus avoid the error. We plan to upload the new optics payloads ~mid next week.

fabiocos:
@jan-kaspar ok, so given the limited scope of these tests, I will not consider it a show-stopper for pre4, under the assumption it will be fixed for pre5

jan-kaspar:
OK

PR #26409, opened 2019-04-09
swapping order of cuts to avoid expensive dxy/dz calc

Sam-Harper:
#### PR description:

The PhotonIDValueMap producer is a known slow module. Part of this is that it does expensive dxy/dz corrections (calling the slow ones too in miniAOD but thats another story).

This is a simple fix which speeds it up by doing a dR match first rather than a slow dxy./dz cut. Further optimisations are more than possible, however this is already significant and is obvious that it is safe.

I'm going to request this back to 102X because it'll speed up analysis user jobs significantly

#### PR validation:

A simple test job on running on (I think) EGamma Run2018 AOD and MINIAOD. This just running VID & scale and smearing in a job and writing out the result. So a typical analysis job.
AOD:
http://sharper.web.cern.ch/sharper/cgi-bin/igprof-navigator/forPRs/aodVIDRef
http://sharper.web.cern.ch/sharper/cgi-bin/igprof-navigator/forPRs/aodVIDSpeedup
time counts PhotonIDValueMapProducer::produce 186.44 -> 81.66
MINIAOD:
http://sharper.web.cern.ch/sharper/cgi-bin/igprof-navigator/forPRs/worstIsolRef
http://sharper.web.cern.ch/sharper/cgi-bin/igprof-navigator/forPRs/worstIsolSpeedup
time counts of PhotonIDValueMapProducer::produce 3,379.67 -> 433.96

### DIFF
--- a/RecoEgamma/PhotonIdentification/plugins/PhotonIDValueMapProducer.cc
+++ b/RecoEgamma/PhotonIdentification/plugins/PhotonIDValueMapProducer.cc
@@ -397,11 +397,20 @@ float PhotonIDValueMapProducer::computeWorstPFChargedIsolation(const reco::Photo
         math::XYZVector phoWrtVtx(photon.superCluster()->x() - vtx->x(),
             photon.superCluster()->y() - vtx->y(), photon.superCluster()->z() - vtx->z());
 
+	const float phoWrtVtxPhi = phoWrtVtx.phi();
+	const float phoWrtVtxEta = phoWrtVtx.eta();
+
         float sum = 0;
         // Loop over the PFCandidates
         for (auto const& aCCand : chargedCands) {
 
-            float dxy = -999;
+            auto iCand = aCCand.candidate;
+            float dR2 = deltaR2(phoWrtVtxEta, phoWrtVtxPhi, iCand->eta(), iCand->phi());
+            if (dR2 > coneSizeDR * coneSizeDR ||
+                    (options & DR_VETO && dR2 < dRveto * dRveto))
+                continue;
+
+	    float dxy = -999;
             float dz = -999;
             if (options & PV_CONSTRAINT)
                 getImpactParameters(aCCand, pv, dxy, dz);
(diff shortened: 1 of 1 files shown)


perrotta:
please test

RecoEgamma/PhotonIdentification/plugins/PhotonIDValueMapProducer.cc
```diff
@@ -397,11 +397,20 @@ float PhotonIDValueMapProducer::computeWorstPFChargedIsolation(const reco::Photo
...
+	const float phoWrtVtxEta = phoWrtVtx.eta();
+
         float sum = 0;
         // Loop over the PFCandidates
         for (auto const& aCCand : chargedCands) {
 
-            float dxy = -999;
+            auto iCand = aCCand.candidate;
+            float dR2 = deltaR2(phoWrtVtxEta, phoWrtVtxPhi, iCand->eta(), iCand->phi());
+            if (dR2 > coneSizeDR * coneSizeDR ||
```
perrotta: Some additional little gain can be achieved if you compute "coneSizeDR * coneSizeDR" and "dRveto * dRveto" only once when they are defined (as only their squares are used in this method)

Sam-Harper: if we go down that route, we'll redo the entire class, its text book on how to write slow code.  Yes there is much much more that could be done.  I figure lets get the win, bake it in and then really fix this class later.

perrotta: I am not asking to rewrite the entire class, and of course I'm happy that you plan doing it eventually.
But if the purpose of this PR is to improve timing in the less invasive way, I think that don't having to make those multiplications multiple times per event can save a few cycles in an absolutely strightforward way and without any possible counterindication: I don't see why not to apply it also.

Sam-Harper: Well two reasons 1) everytime I touch something there is a potential for error and 2) I'm concerned that because this class is such a hot mess that this will be the start of many such changes. 

Now I see from your intention that its not to go down that route, so I make those changes and call it a day?

perrotta:
please test

perrotta:
+1
- By changing the order of the cuts the final result doesn't change (jenkins tests show no differences, as expected), and the timing improves significantly (as reported in the PR description)

fabiocos:
+1

PR #26420, opened 2019-04-10
adding fixedGridRhoFastjetAllTmp to MiniAOD event content

Sam-Harper:
#### PR description:
The E/gamma energy regressions use fixedGridRhoFastJetAllTmp as an input. While this is very similar to fixedGridRhoFastJetAll (its done on PF candidates before e/gamma feeds back its energy corrected electrons) it would be nice to have it in the miniAOD as we can use it to exactly reproduce the original application of the regression. Its already in AOD, this corrects the oversight that it is not in MINIAOD

Its a single double added to the event and there are already several fixedGridRhoFastjet flavours so it shouldnt have any impact,.

### DIFF
--- a/PhysicsTools/PatAlgos/python/slimming/MicroEventContent_cff.py
+++ b/PhysicsTools/PatAlgos/python/slimming/MicroEventContent_cff.py
@@ -47,6 +47,7 @@ MicroEventContent = cms.PSet(
 
         'keep double_fixedGridRhoAll__*',
         'keep double_fixedGridRhoFastjetAll__*',
+        'keep double_fixedGridRhoFastjetAllTmp__*',
         'keep double_fixedGridRhoFastjetAllCalo__*',
         'keep double_fixedGridRhoFastjetCentral_*_*',
         'keep double_fixedGridRhoFastjetCentralCalo__*',


slava77:
@cmsbuild please test

slava77:
@Sam-Harper 
please elaborate why fixedGridRhoFastJetAll can not be used for your purposes.
I'm trying to understand if we are mitigating a logic error or if the Tmp has to be used from basic principles.

Sam-Harper:
Because fixedGridRhoFastJetAll  uses the electron/photon regression as input. Therefore using it as an input to the electron/photon regression creates a circular dependence. 

Now we can keep the status quo and during reco just fixedGridRhoFastJetAllTmp for the initial regression application and then fixedGridRhoFastJetAll for any recalculation of the regression. Admittedly the difference is very small. But I figured for the sake of one double in the event, it would be nice to be in the miniAOD and then we can exactly reproduce the regression used in reco. 

Its not a super strong request, it just would make my life a bit easier when debugging and the like.

slava77:
assign xpog

to confirm the output content changes in miniAOD

peruzzim:
+xpog

it would be better to use a more representative name than "Tmp", but I guess it's something cumbersome to change elsewhere (also maintaining compatibility for other releases)

Sam-Harper:
> it would be better to use a more representative name than "Tmp", but I guess it's something cumbersome to change elsewhere (also maintaining compatibility for other releases)

Agreed but that is the name of the quantity set in 2014.  It makes some sense if you know the provenance.  Because there is a bunch of circular dependences in particle flow, you end up having gedPhotonsTmp, gedGsfElectronsTmp, particleFlowTmp so from that convention, it makes sense. It is already available in the AOD. 

Btw adding this variable will not overlly confuse the rho situation, there are already many flavours. 
double                                "fixedGridRhoAll"           ""                "RECO"    
double                                "fixedGridRhoFastjetAll"    ""                "RECO"    
double                                "fixedGridRhoFastjetAllCalo"   ""                "RECO"    
double                                "fixedGridRhoFastjetCentral"   ""                "RECO"    
double                                "fixedGridRhoFastjetCentralCalo"   ""                "RECO"    
double                                "fixedGridRhoFastjetCentralChargedPileUp"   ""                "RECO"    
double                                "fixedGridRhoFastjetCentralNeutral"   ""                "RECO"

peruzzim:
ok, thanks for the explanation

slava77:
+1

for https://github.com/cms-sw/cmssw/pull/26420	7f41cc1
- jenkins tests pass and comparisons with the baseline show differences only in the event content of the mininAOD outputs, as expected.
The use case seems somewhat reasonable.

fabiocos:
+1

fabiocos:
merge

santocch:
+1

PR #26423, opened 2019-04-10
DD4hep: Define Vacuum as thin Air

ianna:
#### PR description:

When DD4hep geometry to G4 geometry converter is used the Vacuum as currently defined:
```
<ElementaryMaterial name="Vacuum" density="1e-13*mg/cm3" symbol=" " atomicWeight="1*g/mole" atomicNumber="1"/>
```
causes G4 to throw:
```
-------- EEEE ------- G4Exception-START -------- EEEE -------

*** ExceptionHandler is not defined ***
*** G4Exception : mat012
      issued by : G4Element::G4Element()
Fail to create G4Element CMS element with Z= 2  N= 1   N < Z is not allowed

*** Fatal Exception ***
-------- EEEE -------- G4Exception-END --------- EEEE -------
```
Instead, the following Vacuum definition as a thin Air works fine:
```
<CompositeMaterial name="Vacuum" density="1e-10*mg/cm3" symbol=" " method="mixture by weight">
   <MaterialFraction fraction="0.7494">
    <rMaterial name="materials:Nitrogen"/>
   </MaterialFraction>
   <MaterialFraction fraction="0.2369">
    <rMaterial name="materials:Oxygen"/>
   </MaterialFraction>
   <MaterialFraction fraction="0.0129">
    <rMaterial name="materials:Argon"/>
   </MaterialFraction>
   <MaterialFraction fraction="0.0008">
    <rMaterial name="materials:Hydrogen"/>
   </MaterialFraction>
  </CompositeMaterial>
```
which creates:
```
Geant4Converter        ++ Created G4  Material: materials:Vacuum    density:  0.000 mg/cm3  RadL: 5864694.266 km   Nucl.Int.Length: 13694660.844 km 
                       Imean:  85.538 eV   temperature: 273.15 K  pressure:   1.00 atm
```
#### PR validation:

when https://github.com/cms-sw/cmssw/pull/26383 is merged:
```
cmsRun SimG4Core/DD4hepGeometry/test/python/testG4Geometry.py
```

no back port is needed

### DIFF
--- a/DetectorDescription/DDCMS/data/cms-2015-muon-geometry.xml
+++ b/DetectorDescription/DDCMS/data/cms-2015-muon-geometry.xml
@@ -30,7 +30,7 @@
   </ConstantsSection>
 
   <IncludeSection>
-    <Include ref='Geometry/CMSCommonData/data/materials.xml'/>
+    <Include ref='Geometry/CMSCommonData/data/materials/2019/v1/materials.xml'/>
     <Include ref='Geometry/CMSCommonData/data/rotations.xml'/>
     <Include ref='Geometry/CMSCommonData/data/extend/cmsextent.xml'/>
     <Include ref='Geometry/CMSCommonData/data/cms.xml'/>
--- a/Geometry/CMSCommonData/data/materials/2019/v1/materials.xml
+++ b/Geometry/CMSCommonData/data/materials/2019/v1/materials.xml
@@ -1,0 +1,4709 @@
+<?xml version="1.0"?>
+<DDDefinition xmlns="http://www.cern.ch/cms/DDL" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.cern.ch/cms/DDL ../../../DetectorDescription/Schema/DDLSchema.xsd">
+ <MaterialSection label="materials.xml">
+  <ElementaryMaterial name="Aluminium" density="2.7*g/cm3" symbol=" " atomicWeight="26.98*g/mole" atomicNumber="13"/>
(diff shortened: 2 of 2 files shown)


ianna:
assign simulation

ianna:
please test

civanch:
@ianna , vacuum in a vacuum chamber is not equal to air. You need a consultation from machine experts, they know this. Moreover, you may ask if the vacuum is the same in different parts of CMS.

ianna:
@civanch which vacuum is used in CMS simulation?

civanch:
@ianna , it is:

<material name="Vacuum0x7f2d9f51e000" state="gas"><T unit="K" value="293.15"/><MEE unit="eV" value="19.2"/><D unit="g/cm3" value="9.99999473841014e-17"/><fraction n="1" ref="H0x7f2d9f6ada40"/>

where H is hydrogen - physically it is more adequate than air. Concerning density - better to ask, if the value is correct. Machine peoples know current situation better than we have in this old definition.

civanch:
@ianna , I suspect DD4hep bug, the line you show potentially correct.

ianna:
Thanks I’ll follow up

civanch:
@ianna, I do not see MEE parameter in your line, I suspect it is DD4Hep defect. They must implement material properly.

fabiocos:
@ianna @civanch any news on this subject? I fully agree with Vladimir that the standard air is not suitable as vacuum proxy, the density is incompatible with we may expect from the beam pipe ambient

ianna:
@fabiocos - yes, it's fixed it in the #26740

PR #26196, opened 2019-03-15
one::OutputModule now supports transition caches

Dr15Jones:
#### PR description:

Can now use LuminosityBlockCache and RunCache with one::OutputModule.
When used, the module does not cause a synchronization barrier at those transitions.

#### PR validation:

Framework code compiles. All framework related unit tests pass.

PR #25951, opened 2019-02-15
DAQ limited output module (10_5_X)

smorovic:
This is DAQ (HLT/Filter Farm) output module implementation based on limited::OutputModule. This implementation removes limitation of handling only one LS at the time present in current one DAQ module.

- New module also rely on changes introduced here: https://github.com/cms-sw/cmssw/pull/21830

- StreamerOutputModuleBase (inheriting one::OutputModule) base is restructured and serializer code (shared with the new, limited implementation) extracted to a common class which can be used by both one and limited module. Currently, "base" is used by: DQM test module, Streamer writer module and previous DAQ output module which we'll still keep.

- new module is called EvFOutputModule while previous DAQ module remains as ShmStreamConsumer (this name is currently still used in production configurations).

- Limited OutputModule transitions used to write event and end global lumisection are const functions. There are a few problems associated with this, for which a workaround is to apply const_cast:

1. This is too strict for streamer serializer, which is therefore included in a helper class stored in RunCache. Serializer keeps internally a buffer which is reallocated only if too small to accommodate serialized event. To allow access, const_cast is applied in event write callback.

2. File writer in LuminosityBlockCache object also needs const_cast in event write and global EoL ( where file is written to / closed). Also it needs to update the object to do per-LS event count bookkeeping.

3. JSON writer, which is initialized once with JSON definitions, is also stored in RunCache, and needs to be allowed modification at the time of global EoL.

- Those const_casts should be thread-safe since this limited output module strictly requires concurrencyLimit = 1 at runtime, and will throw exception if another configuration value is set.

PR #26425, opened 2019-04-10
DAQ multi-LS output module (10_6_X)

smorovic:
#### PR description:

This is DAQ (HLT/Filter Farm) output module implementation based on one::OutputModule watching LuminosityBlockCache. In that case limitation of handling only one LS at the time present in current DAQ module is removed. Based on #25951 which was an attempt to do this using limited output module class, and possible thanks to improvements made in #26196.

StreamerOutputModuleBase (inheriting one::OutputModule) base is restructured and serializer code (shared with the new EvFOutputModule) extracted to a common class which can be used by both module types (which inherit on different templated class of one::OutputModule). Currently, "base" is used by: DQM test module, Streamer writer module and previous DAQ output module which will still be kept for some time in the release.

New module is called EvFOutputModule while previous DAQ module remains registered ShmStreamConsumer (this name is currently still used in production configurations).

#### PR validation:

Output module was added and tested with configuration EventFilter/Utilities/test/startFU.py and also new configuration configured to work with 2 concurrent LS at the time: start_multiLS_FU.py.

It has also been verified to work in test setup emulating the Filter Farm: using both old and new version of the output module (former being tested due to changes made in the base class). Both output modules produced files or comparable size, identical header. Also, dat files from the new module could be normally read by the NewEventStreamFileReader source.


Dr15Jones:
please test

EventFilter/Utilities/interface/EvFOutputModule.h
```diff
@@ -0,0 +1,118 @@
...
+
+    std::string const& getFilePath() const {return filePath_;}
+
+    unsigned long getAccepted() const {return accepted_;}
+    void incAccepted() {accepted_++;}
+
+  private:
+    std::string filePath_;
+    unsigned long accepted_;
+    edm::propagate_const<std::shared_ptr<StreamerOutputFile>> stream_writer_events_;
```
Dr15Jones: Wouldn't std::unique_ptr be more appropriate as only this class is allowed to own the object?

EventFilter/Utilities/interface/EvFOutputModule.h
```diff
@@ -0,0 +1,118 @@
...
+
+    Trig getTriggerResults(edm::EDGetTokenT<edm::TriggerResults> const& token, edm::EventForOutput const& e) const;
+
+    edm::ParameterSet const& ps_;
+    std::string streamLabel_;
+    edm::EDGetTokenT<edm::TriggerResults> trToken_;
+
+    evf::FastMonitoringService *fms_;
+
+    std::shared_ptr<evf::EvFOutputJSONWriter> jsonWriter_;
```
Dr15Jones: `std::unique_ptr` would seem to be more appropriate.

EventFilter/Utilities/interface/EvFOutputModule.h
smorovic: I can agree it would be more appropriate.
By looking more closely at this, there is another usage of shared_ptr, "stream_writer_preamble" in beginRun. It looks like here just a local variable can be used, so I will push that change as well.

EventFilter/Utilities/interface/EvFOutputModule.h
smorovic: Agreed also on this one.

slava77:
@cmsbuild please test

slava77:
+1

for https://github.com/cms-sw/cmssw/pull/26425	4bc1f86
- reco files in the EventFilter/Utilities package shared with DAQ are not affected
- jenkins tests pass

emeschi:
+1

Dr15Jones:
+1

fabiocos:
+1

PR #26427, opened 2019-04-10
add bool method to cms.Bool

davidlange6:
#### PR description:

needed by Python3 so that if cms.Bool(False): evaluates to False

### DIFF
--- a/FWCore/ParameterSet/python/Types.py
+++ b/FWCore/ParameterSet/python/Types.py
@@ -151,7 +151,8 @@ class bool(_SimpleParameterTypeBase):
         parameterSet.addBool(self.isTracked(), myname, self.value())
     def __nonzero__(self):
         return self.value()
-
+    def __bool__(self):
+        return self.__nonzero__()
 
 
 class string(_SimpleParameterTypeBase):


davidlange6:
please test

Dr15Jones:
Could you add a unit test the show that it works?

davidlange6:
Unit tests for comparison operations in Types.py need to be strengthened in general I'm getting to it slowly.

Dr15Jones:
> Unit tests for comparison operations in Types.py need to be strengthened in general I'm getting to it slowly.
I'm fine with just adding this one test :).

davidlange6:
So the test is actually there.. but run explicitly with python2..  I can add a parallel python3 check (but after this PR goes in so that it passes...)

Dr15Jones:
+1

Dr15Jones:
So the reason this didn't get run is you are not running the unit tests using python2 and python3? Can we come up with a way to do so? The tests are just executed now from FWCore/ParameterSet/test/runPythonTests.sh. That could either be extended to run python2 and python3 or a similar versions of that script to be move to the package that gives python3 support.

davidlange6:
the unit test is a shell script that calls 'python' explicitly. Its easy to modify to run both (but now your correct question is how to find the N other cases like this...)

fabiocos:
+1

PR #26429, opened 2019-04-10
Fix clang warnings in FastSimulation/SimplifiedGeometryPropagator

fabiocos:
#### PR description:

Fix clang warnings in FastSimulation/SimplifiedGeometryPropagator: use CMS_THREAD_GUARD, reuse std::once_flag in call_once to initialize static objects in NuclearInteraction, and remove unused [this] from lambda expression in plugin NuclearInteractionFTF.

#### PR validation:

Code compiles, workflow 2018.8 runs.

### DIFF
--- a/FastSimulation/SimplifiedGeometryPropagator/plugins/NuclearInteraction.cc
+++ b/FastSimulation/SimplifiedGeometryPropagator/plugins/NuclearInteraction.cc
@@ -21,6 +21,8 @@
 #include "FWCore/Utilities/interface/Exception.h"
 #include "FWCore/ParameterSet/interface/FileInPath.h"
 
+#include "FWCore/Utilities/interface/thread_safety_macros.h"
+
 // Fast Sim headers
 #include "FastSimulation/SimplifiedGeometryPropagator/interface/Particle.h"
 #include "FastSimulation/SimplifiedGeometryPropagator/interface/SimplifiedGeometry.h"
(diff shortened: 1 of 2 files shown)


fabiocos:
please test

FastSimulation/SimplifiedGeometryPropagator/plugins/NuclearInteraction.cc
```diff
@@ -200,8 +202,8 @@ namespace fastsim
     // TODO: Is this correct?
     // Thread safety
     static std::once_flag initializeOnce;
-    [[cms::thread_guard("initializeOnce")]] std::vector< std::vector<double> > NuclearInteraction::theRatiosMap;
-    [[cms::thread_guard("initializeOnce")]] std::map<int, int> NuclearInteraction::theIDMap;
+    CMS_THREAD_GUARD(initializeOnce) std::vector< std::vector<double> > NuclearInteraction::theRatiosMap;
```
fabiocos: @Dr15Jones do you see what is different here wrt to the case of NuclearInteractionFTF ? I might simply restore the previous syntax removing the once_flag declaration, but I do not understand what is wrong here...

fabiocos: indeed the other possible solution works in case

Dr15Jones: ` CMS_THREAD_GUARD(initializeOnce)` does not actually use the variable `initializeOnce`, it just creates the string `"initializeOnce"`. As far as I can tell, the variable is not actually used.

Dr15Jones: The varaibles `theRatiosMap` and `theIDMap` are not handled in a thread-safe way.

fabiocos: @Dr15Jones ok, in the case of NuclearInteractionFTF initializeOnce is then used by std::call_once, that is the difference

fabiocos:
please test

fabiocos:
@ssekmen this is a technical fix, could you please check it?

ssekmen:
+1

fabiocos:
+1

PR #26446, opened 2019-04-12
[muon] a workaround for enum to int conversion problem in pyROOT

drkovalskyi:
There is a subtle problem with python type conversion from signed enums to basic types. In some cases it converts negative enums to an unsigned int. For example instead of -10 you get 4294967286. This problem was report to ROOT team, but it was never fixed. 
It's very easy to miss this problem and unfortunately one cannot rely on enum checks to avoid it, since for example ROOT.reco.GhostMuonFromGaugeOrHiggsBoson evaluates to -10 as it should, but simExtType() returns 4294967286.
The current workaround is documented on Muon POG twiki, but it's not a reliable solution. People (including me keep rediscovering this problem).
https://twiki.cern.ch/twiki/bin/view/CMS/SWGuideMuonIdRun2#SimHit_Muon_matching_since_CMSSW

The pull request provides a trivial fix - return int instead of the enum. I know it's ugly, but getting wrong physics results is worse.

### DIFF
--- a/DataFormats/PatCandidates/interface/Muon.h
+++ b/DataFormats/PatCandidates/interface/Muon.h
@@ -281,8 +281,10 @@ namespace pat {
       void  setSoftMvaValue(float softmva){ softMvaValue_ = softmva; }
 
       /// MC matching information
-      reco::MuonSimType simType() const { return simType_; }
-      reco::ExtendedMuonSimType simExtType() const { return simExtType_; }
+      /// returns reco::MuonSimType - we cast it to int to avoid issues in pyROOT
+      int simType() const { return simType_; }
+      /// reco::ExtendedMuonSimType - we cast it to int to avoid issues in pyROOT
+      int simExtType() const { return simExtType_; }
       //  FLAVOUR:
       //  - for non-muons: 0
       //  - for primary muons: 13


slava77:
@cmsbuild please test

drkovalskyi:
FYI @bmahakud, @folguera

slava77:
@folguera 
is this the fix in sim matching that you mentioned yesterday or is this something else?

drkovalskyi:
No, the sim-hit matching problem is still not understood. It's a different bug/problem/feature.

DataFormats/PatCandidates/interface/Muon.h
```diff
@@ -281,8 +281,10 @@ namespace pat {
       void  setSoftMvaValue(float softmva){ softMvaValue_ = softmva; }
 
       /// MC matching information
-      reco::MuonSimType simType() const { return simType_; }
-      reco::ExtendedMuonSimType simExtType() const { return simExtType_; }
+      /// returns reco::MuonSimType - we cast it to int to avoid issues in pyROOT
+      int simType() const { return simType_; }
+      /// reco::ExtendedMuonSimType - we cast it to int to avoid issues in pyROOT
+      int simExtType() const { return simExtType_; }
```
slava77: This solution degrades type safety to serve a broken external use case.

I propose to introduce/add ```simTypeInt``` and ```simExtTypeInt``` with appropriate comments that it's to be used in FWLite environment where the scoped enums are not properly recognized.

drkovalskyi: I don't think your solution will work well because the user needs to know that he has to use simTypeInt to be safe. I have personally rediscovered this problem 3 times, i.e. one tends to forget about it and use it as is. All these accessors are primarily used by endusers, i.e. there won't be many use cases where someone would really benefit from type safety. In other words the tradeoff is between a benefit that won't be used vs a real problem that affects a number of users.

slava77:
> This problem was report to ROOT team, but it was never fixed.

please share the ROOT JIRA issue number for this.

drkovalskyi:
https://sft.its.cern.ch/jira/browse/ROOT-8935

slava77:
In https://sft.its.cern.ch/jira/browse/ROOT-8935 I read that the part where
the default type is uint is apparently pointing to https://sft.its.cern.ch/jira/browse/ROOT-6988.
there is a message from Dec 17 stating that this issue is resolved (although the message is a bit too brief).

@pcanal @Dr15Jones is the enum support in pyroot reliable?

It doesn't seem appropriate to me to degrade the type correctness at the C++ level just because a fraction of the user code can not parse it correctly.

davidlange6:
> On Apr 15, 2019, at 4:02 PM, Slava Krutelyov <notifications@github.com> wrote:
> 
> In https://sft.its.cern.ch/jira/browse/ROOT-8935 I read that the part where
> the default type is uint is apparently pointing to https://sft.its.cern.ch/jira/browse/ROOT-6988.
> there is a message from Dec 17 stating that this issue is resolved (although the message is a bit too brief).
> 

resolved from inactivity it seems


> @pcanal @Dr15Jones is the enum support in pyroot reliable?

Is there something actually pyroot specific? 

root seems treats enums being uints rather than ints

root [0] enum foo { aa, bb, cc }
root [1] aa
(foo) (foo::aa) : (unsigned int) 0

which doesn't appear to follow the C++ standard (as I read about it, though the lira ticket above implies uint was expected...)

you can just take advantage of C++11 and do

root [2] enum foo2 : signed int { aa2, bb2, cc2 }
root [3] aa2
(foo2) (foo2::aa2) : (int) 0

to get the behavior you want - better than returning an int explicitly in any case.

davidlange6:
> On Apr 15, 2019, at 4:27 PM, David Lange <notifications@github.com> wrote:
> 
> 
> 
> > On Apr 15, 2019, at 4:02 PM, Slava Krutelyov <notifications@github.com> wrote: 
> > 
> > In https://sft.its.cern.ch/jira/browse/ROOT-8935 I read that the part where 
> > the default type is uint is apparently pointing to https://sft.its.cern.ch/jira/browse/ROOT-6988. 
> > there is a message from Dec 17 stating that this issue is resolved (although the message is a bit too brief). 
> > 
> 
> resolved from inactivity it seems 
> 
> 
> > @pcanal @Dr15Jones is the enum support in pyroot reliable? 
> 
> Is there something actually pyroot specific? 
> 
> root seems treats enums being uints rather than ints 
> 
> root [0] enum foo { aa, bb, cc } 
> root [1] aa 
> (foo) (foo::aa) : (unsigned int) 0 
> 
> which doesn't appear to follow the C++ standard (as I read about it, though the lira ticket above implies uint was expected...) 
> 

ah - but at the root prompt, the right thing happens given a enum with negative value

root [0] enum foo2  { aa2=-1, bb2, cc2 }
root [1] aa2
(foo2) (foo2::aa2) : (int) -1

so thats how its pyroot specific

slava77:
based on https://sft.its.cern.ch/jira/browse/ROOT-8935
it sounds like we can get the right behavior by specifying the underlying type of the enums (```enum MuonSimType : int {```) in this case.

drkovalskyi:
I’ll test the idea

drkovalskyi:
It doesn't work.

slava77:
> It doesn't work.

what exactly did not work?
- reading newly created muons
- reading old muons

Please clarify.

drkovalskyi:
Reading doesn't work. I don't see why I writing would change anything since the checksum is not changing, i.e. representation on disk is the same.

slava77:
The ticket in (py)ROOT is now reopened and perhaps will be resolved soon.
So, the resolution proposed here is likely not needed in the log run.

I think that from the code safety and usability perspective ```simTypeInt``` and ```simExtTypeInt``` can still be added (perhaos even with ```/// Safe for pyROOT``` comment).
I understand the arguments in https://github.com/cms-sw/cmssw/pull/26446#discussion_r275114050 that the change needs to be announced; however, even as proposed there should be an announcement that declares the old named methods safe to use (without the workaround in the twiki).

drkovalskyi:
simTypeInt is not a solution, since for anyone who is aware of the problem a workaround is not hard and we already provide instructions on our twiki. So I see no point to introduce parallel methods and advertise them - it won't work (I keep forgetting about this "feature" myself).
While the pyROOT thread indicates some promise it's not clear when exactly we will have this fix available to users if ever. To me the proposed solution is better because it fixes the problem and prevents mistakes on the analysis level now and it fixes it for old data as well. It also works perfectly fine in the case when people want to use enums instead of ints preserving some type safety.
Therefore I propose we integrate the change and when we do have a proper fix we revert the change. The change doesn't affect the data representation on disk, so as soon as pyROOT in CMSSW release can handle enums properly we can go back to returning enums.

slava77:
Is there some example code that I could try in pyROOT to see the problem?

drkovalskyi:
Here is a simple example: /afs/cern.ch/user/d/dmytro/public/forSlava/test_pyroot.py 
When it works right you should get simType to be -10, when it's not  you get 4294967294.
It reads data available at CERN, so it's better to run from CERN.

slava77:
It looks like there was some progress in resolving the issue with the enums
https://sft.its.cern.ch/jira/browse/ROOT-8935

From the last comment there https://sft.its.cern.ch/jira/browse/ROOT-8935#comment-104678
I'm somewhat lost in the packaging of (Py)ROOT and/or cppyy

@fabiocos @smuzaffar  please check to see if we can have the fix in our release stack.
Thank you.

slava77:
a short status update: 
since almost a week there is a PR for the ROOT master branch
https://github.com/root-project/root/pull/4121
Backports to older versions can be considered after it successfully gets merged .

mrodozov:
please test with https://github.com/cms-sw/root/pull/130
ill pick the changes in the morning

mrodozov:
now when I think about it it will need this root pr commits to be backported to 6-14 
this will probably fail

mrodozov:
please test with https://github.com/cms-sw/cmsdist/pull/5130
lets use it once and then I close it (5130)

slava77:
> Here is a simple example: /afs/cern.ch/user/d/dmytro/public/forSlava/test_pyroot.py
> When it works right you should get simType to be -10, when it's not you get 4294967294.
> It reads data available at CERN, so it's better to run from CERN.

I tested a slightly modified version in CMSSW_11_0_ROOT6_X_2019-09-03-0600
```python
#!/bin/env python
import os, re, sys, time, ROOT, subprocess
from DataFormats.FWLite import Events, Handle

muonHandle, muonLabel = Handle("std::vector<pat::Muon>"),"slimmedMuons"

ROOT.gROOT.SetBatch(True)
events = Events("root://cms-xrd-global.cern.ch//store/group/phys_muon/dmytro/tmp/store+mc+RunIIAutumn18MiniAOD+DYJetsToLL_M-50_TuneCP5_13TeV-madgraphMLM-pythia8+MINIAODSIM+102X_upgrade2018_realistic_v15-v1+80000+695EE995-7CA9-5A46-81EF-C6777A74C791.root")
nevents = 0

print "ROOT.reco.MatchedMuonFromGaugeOrHiggsBoson  = ", ROOT.reco.MatchedMuonFromGaugeOrHiggsBoson
print "ROOT.reco.GhostMuonFromGaugeOrHiggsBoson  = ", ROOT.reco.GhostMuonFromGaugeOrHiggsBoson
print "ROOT.reco.ExtendedMuonSimType.GhostMuonFromGaugeOrHiggsBoson  = ", ROOT.reco.ExtendedMuonSimType.GhostMuonFromGaugeOrHiggsBoson

for event in events:
    if nevents>=100: break
    nevents+=1
    event.getByLabel(muonLabel, muonHandle)
    muons = muonHandle.product()
    for muon in muons:
        if muon.simExtType()==ROOT.reco.GhostMuonFromGaugeOrHiggsBoson or muon.simExtType()>1e4 or muon.simExtType()<0:
            print "simExtType()==ROOT.reco.GhostMuonFromGaugeOrHiggsBoson or pos/neg: pt: %0.1f \tsimType: %d \textSimType: %d" % (muon.pt(),muon.simType(),muon.simExtType())
        if muon.simExtType()==ROOT.reco.GhostMuonFromGaugeOrHiggsBoson:
            print "simExtType()==ROOT.reco.GhostMuonFromGaugeOrHiggsBoson: pt: %0.1f \tsimType: %d \textSimType: %d" % (muon.pt(),muon.simType(),muon.simExtType())
        if muon.simExtType()==ROOT.reco.ExtendedMuonSimType.GhostMuonFromGaugeOrHiggsBoson:
            print "simExtType()==ROOT.reco.ExtendedMuonSimType.GhostMuonFromGaugeOrHiggsBoson: pt: %0.1f \tsimType: %d \textSimType: %d" % (muon.pt(),muon.simType(),muon.simExtType())
```

The output now seems to make sense:
```
ROOT.reco.MatchedMuonFromGaugeOrHiggsBoson  =  10
ROOT.reco.GhostMuonFromGaugeOrHiggsBoson  =  -10
ROOT.reco.ExtendedMuonSimType.GhostMuonFromGaugeOrHiggsBoson  =  -10
simExtType()==ROOT.reco.GhostMuonFromGaugeOrHiggsBoson or pos/neg: pt: 1.1      simType: -4     extSimType: -10
simExtType()==ROOT.reco.GhostMuonFromGaugeOrHiggsBoson: pt: 1.1         simType: -4     extSimType: -10
simExtType()==ROOT.reco.ExtendedMuonSimType.GhostMuonFromGaugeOrHiggsBoson: pt: 1.1     simType: -4     extSimType: -10
simExtType()==ROOT.reco.GhostMuonFromGaugeOrHiggsBoson or pos/neg: pt: 1.0      simType: -2     extSimType: -4
```

this also apparently works in root 6.18 branch e.g. in CMSSW_11_0_ROOT618_X_2019-09-03-0600

@drkovalskyi 
please double-check that the issue is solved (either in CMSSW_11_0_ROOT618_X_2019-09-03-0600 or in CMSSW_11_0_ROOT6_X_2019-09-03-0600).
Thank you.

slava77:
https://github.com/cms-sw/cmssw/pull/26446#issuecomment-527567587

> @drkovalskyi
> please double-check that the issue is solved (either in CMSSW_11_0_ROOT618_X_2019-09-03-0600 or in CMSSW_11_0_ROOT6_X_2019-09-03-0600).
> Thank you.

@drkovalskyi 
please let me know if you had a chance to check.
Thank you.

drkovalskyi:
It all looks good in my tests.

slava77:
-1

this is no longer necessary after the root update available starting from 
CMSSW_11_0_X_2019-10-31-2300

see notes in https://github.com/cms-sw/cmsdist/pull/5321#issuecomment-548793216
I also requested a backport of the root update for 10_6_X so that we have it in the long term support release

@drkovalskyi 
please close this PR if you are fine with the resolution.

fabiocos:
@drkovalskyi I would say that this PR can be closed at this point, please comment in case, otherwise I'll clean this from the list.

fabiocos:
hold

fabiocos:
@drkovalskyi thanks

PR #26459, opened 2019-04-15
[muon] updated multiIso working points

drkovalskyi:
An update for supported multiIso working points based on input from SUSY group (@gzevi ). Dropped unsupported working points.

### DIFF
--- a/PhysicsTools/PatAlgos/plugins/PATMuonProducer.cc
+++ b/PhysicsTools/PatAlgos/plugins/PATMuonProducer.cc
@@ -695,10 +695,7 @@ void PATMuonProducer::produce(edm::Event & iEvent, const edm::EventSetup & iSetu
 
       // multi-isolation
       if (computeMiniIso_){
-
-
-	muon.setSelector(reco::Muon::MultiIsoLoose,  miniIsoValue<0.40 && (muon.jetPtRatio() > 0.80 || muon.jetPtRel() > 7.2) );
-	muon.setSelector(reco::Muon::MultiIsoMedium, miniIsoValue<0.16 && (muon.jetPtRatio() > 0.76 || muon.jetPtRel() > 7.2) );
+	muon.setSelector(reco::Muon::MultiIsoMedium, miniIsoValue<0.11 && (muon.jetPtRatio() > 0.74 || muon.jetPtRel() > 6.8) );
       }
 
       // MVA working points


slava77:
@cmsbuild please test

slava77:
@drkovalskyi please add a link to the supporting documentation of the new working point in the PR description.
Thank you.

PhysicsTools/PatAlgos/plugins/PATMuonProducer.cc
```diff
@@ -695,10 +695,7 @@ void PATMuonProducer::produce(edm::Event & iEvent, const edm::EventSetup & iSetu
 
       // multi-isolation
       if (computeMiniIso_){
-
-
-	muon.setSelector(reco::Muon::MultiIsoLoose,  miniIsoValue<0.40 && (muon.jetPtRatio() > 0.80 || muon.jetPtRel() > 7.2) );
-	muon.setSelector(reco::Muon::MultiIsoMedium, miniIsoValue<0.16 && (muon.jetPtRatio() > 0.76 || muon.jetPtRel() > 7.2) );
+	muon.setSelector(reco::Muon::MultiIsoMedium, miniIsoValue<0.11 && (muon.jetPtRatio() > 0.74 || muon.jetPtRel() > 6.8) );
```
slava77: I found these cuts in https://twiki.cern.ch/twiki/bin/view/CMS/SUSLeptonSF#ID_IP_ISO
these are quoted as ```Medium2017data_V32JEC```.
This looks like WPs change depending on the version of miniAOD processing.
Is this setting approved/OK for (re)miniAOD of 2016 80X RECO/AOD or for 2018 102X RECO/AOD?
What about the 10_6_X itself (legacy rereco)?

@gzevi please advise or tag someone who can clarify.

drkovalskyi: We won't track JEC changes, i.e. whatever JET pt is used during processing it will be the Pt used in the selector. Muon POG will only provide scale factors for pre-computed selectors. It was discussed with @gzevi and it was concluded that it's still useful for SUSY group. If not, we will drop both working points.

slava77:
+1

for https://github.com/cms-sw/cmssw/pull/26459	e2640f9
    - the working point update is in agreement with the latest https://twiki.cern.ch/twiki/bin/view/CMS/SUSLeptonSF#ID_IP_ISO
- jenkins tests pass and comparisons with the baseline show differences only in the miniAOD muons selector flags as expected.

fabiocos:
+1

@santocch please have a look, this code is relevant for miniAOD processing but for analysis use in general

fabiocos:
merge

santocch:
+1

PR #26461, opened 2019-04-15
Add trigger bit to emulated GlobalExtBlk to indicate event was not selected …

apana:
…due to prefiring of the trigger

### DIFF
--- a/L1Trigger/L1TGlobal/plugins/L1TExtCondProducer.cc
+++ b/L1Trigger/L1TGlobal/plugins/L1TExtCondProducer.cc
@@ -36,6 +36,8 @@
 
 #include "DataFormats/L1TGlobal/interface/GlobalExtBlk.h"
 
+#include "DataFormats/TCDS/interface/TCDSRecord.h"
+
 using namespace std;
 using namespace edm;
 using namespace l1t;
@@ -71,6 +73,12 @@ using namespace l1t;
 
     unsigned long long m_l1GtMenuCacheID;
     std::map<std::string, unsigned int> m_extBitMap;
+
+    unsigned int m_triggerRulePrefireVetoBit;
+
+    bool makeTriggerRulePrefireVetoBit_;
+    edm::EDGetTokenT<TCDSRecord> tcdsRecordToken_;
+    edm::InputTag tcdsInputTag_;
   };
 
   //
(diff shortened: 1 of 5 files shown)


rekovic:
please test

rekovic:
+1

L1Trigger/L1TGlobal/plugins/L1TExtCondProducer.cc
```diff
@@ -82,8 +90,26 @@ using namespace l1t;
...
     setBptxPlus_ (iConfig.getParameter<bool>("setBptxPlus")),
     setBptxMinus_ (iConfig.getParameter<bool>("setBptxMinus")),
-    setBptxOR_ (iConfig.getParameter<bool>("setBptxOR"))
+    setBptxOR_ (iConfig.getParameter<bool>("setBptxOR")),
+    tcdsInputTag_(iConfig.getParameter<edm::InputTag>("tcdsRecordLabel"))
   {
+
+    makeTriggerRulePrefireVetoBit_=false;
+
+    m_triggerRulePrefireVetoBit=255;
```
fabiocos: @apana @rekovic this magic value looks related to https://github.com/cms-sw/cmssw/blob/master/DataFormats/L1TGlobal/interface/GlobalExtBlk.h#L47 and now everything looks consistent. But why having the two as independent, and using a LogWarning that risks to go unnoticed? If this has to be the last bit of the array just use that position no? Or I would escalate the severity to LogError at least (although still it does not seem the best solution to me...)

peruzzim:
Looking at the GlobalExtBlk constructor here:
https://github.com/cms-sw/cmssw/blob/e5aead7a3343499ddf14e980eceb36fe6a767ffe/DataFormats/L1TGlobal/src/GlobalExtBlk.cc#L37
one would need to have at least a method to get the size of the `m_extDecision` vector:
https://github.com/cms-sw/cmssw/blob/e5aead7a3343499ddf14e980eceb36fe6a767ffe/DataFormats/L1TGlobal/interface/GlobalExtBlk.h#L65
otherwise it's impossible to reverse-engineer the logic you use to set the 255-or-max-minus-1 bit of that vector.

fabiocos:
+operations

the previous circular dependency of tcdsDigis and rawDataCollector looks to have been solved,, and the TauEmbedding previously failing unit test now works smoothly

fabiocos:
+1

rekovic:
@peruzzim 
These comments are valid: avoiding setting the size by hand here and having a  public accessor which is transparent to the position of this bit.  

@apana 
Not to introduce delay in integration of the PR now, after this PR is merged (soon), we need to follow up with another PR which provide the above functionality.

apana:
Dear @peruzzim and @rekovic,

I agree this could have been coded better. I will try to put in the follow up PR asap.

PR #26465, opened 2019-04-16
Hcal depth energy fraction packed candidates value map

hatakeyamak:
#### PR description:

Includes HCAL depth information for packed candidates in MINIAOD for high-level object use (e.g. PUPPI recomputation at the post ultra-legacy MINIAOD stage). The promising results are obtained [1,2], but needs more time to finalize it. It means it is important to have this information in packedCandidates, so that the PUPPI weights can be recomputed on-the-fly from MINIAOD.

[0] https://github.com/cms-sw/cmssw/files/3056535/he_depth_reco.pdf
[1] https://indico.cern.ch/event/799365/contributions/3333038/attachments/1805670/2946660/pre.pdf
[2] https://indico.cern.ch/event/799365/contributions/3339621/attachments/1805694/2946705/JK_JetMET_03_19_2.pdf

If this ValueMap-based implementation as suggested by @slava77 and @peruzzim looks good, this may supersede:
[3] https://github.com/cms-sw/cmssw/pull/26363
Now. I also use vectors instead of arrays to store various depth information.

@peruzzim @fgolf @zdemirag @ahinzmann @bendavid @mariadalfonso

#### PR validation:

Made sure that the this information gets added to MINIAOD as intended. The size increase in TTbar sample with PU (tested in 2000 events) in 2017 & 2018 scenarios is
+ 0.0% as expected for 2017
+ 2.9% for 2018
a little larger increase than what [3] gives for 2018 (2.7%) but very similar.

This is not a backport.


slava77:
@cmsbuild please test

hatakeyamak:
FYI: I used this very simple python file to check this addition.
https://baylor.box.com/s/b3jpnwbi7xvyjs8fof05z7ivuxrbk5d3

slava77:
sorry for incomplete comments in  #26363.
All "production" persisted data has to be added in the \*DataFormats subsystem.

Given the pat:: namespace, it's more appropriate in ```DataFormats/PatCandidates```

PhysicsTools/PatAlgos/interface/PATAuxiliaryProd.h
```diff
@@ -0,0 +1,66 @@
...
+      hcalDepthEnergyFraction_.clear();
+      hcalDepthEnergyFractionI_.clear();
+      for (auto it = begin (v); it != end (v); ++it){ 
+	hcalDepthEnergyFractionI_.push_back((uint8_t)((*it)*200.));
+	hcalDepthEnergyFraction_.push_back(*it);
+      }
+    }
+
+    // provide a full vector for each depth
+    std::vector<float> hcalDepthEnergyFraction() {
```
slava77: May I suggest to be less repetitive in "HcalDepthEnergyFraction". This member method can be just ```fractions()```.

PhysicsTools/PatAlgos/interface/PATAuxiliaryProd.h
```diff
@@ -0,0 +1,66 @@
...
+      hcalDepthEnergyFraction_.clear();
+      for (auto it = begin (hcalDepthEnergyFractionI_); it != end (hcalDepthEnergyFractionI_); ++it)
+	hcalDepthEnergyFraction_.push_back(float(*it)/200.);
+      return hcalDepthEnergyFraction_;
+    }
+
+    // provide the info for each depth
+    float hcalDepthEnergyFraction(unsigned int i) {
```
slava77: this can be simply ```fraction(unsigned int i) const ```

PhysicsTools/PatAlgos/src/classes.h
```diff
@@ -0,0 +1,13 @@
+#include "DataFormats/Common/interface/Wrapper.h"
+#include "DataFormats/Common/interface/ValueMap.h"
+#include "PhysicsTools/PatAlgos/interface/PATAuxiliaryProd.h"
+
+namespace PhysicsTools_PatAlgos {
```
slava77: this (namespace ... )  is not needed anymore. Just an include is enough.

PhysicsTools/PatAlgos/src/classes_def.xml
```diff
@@ -0,0 +1,16 @@
...
+
+ <class name="pat::HcalDepthEnergyFractionProd">
+   <field name="hcalDepthEnergyFraction_" transient="true"/>
+ </class>
+ <class name="std::vector<pat::HcalDepthEnergyFractionProd>"/>
+ <class name="edm::Wrapper<std::vector<pat::HcalDepthEnergyFractionProd> >"/>
+ <class name="edm::ValueMap<pat::HcalDepthEnergyFractionProd>" />
+ <class name="edm::Wrapper<edm::ValueMap<pat::HcalDepthEnergyFractionProd> >"/>
+
+ <ioread sourceClass = "HcalDepthEnergyFractionProd" version="[1-]" targetClass="HcalDepthEnergyFractionProd" source="std::vector<uint8_t> hcalDepthEnergyFractionI_" target="hcalDepthEnergyFraction_">
```
slava77: please add newlines for the fields 
```xml
<ioread sourceClass = "HcalDepthEnergyFractionProd" version="[1-]"
              targetClass="HcalDepthEnergyFractionProd"
              source="std::vector<uint8_t> hcalDepthEnergyFractionI_"
...
```

PhysicsTools/PatAlgos/src/classes_def.xml
```diff
@@ -0,0 +1,16 @@
...
+ <![CDATA[ newObj->initUint8();
+ ]]>
```
slava77: this looks backwards: the goal should be to initialize fraction float vector from the int, not the other way around.

PhysicsTools/PatAlgos/src/classes_def.xml
```diff
@@ -0,0 +1,16 @@
+<lcgdict>
+
+ <class name="pat::HcalDepthEnergyFractionProd">
+   <field name="hcalDepthEnergyFraction_" transient="true"/>
```
slava77: a class version is needed, please start from 3

PhysicsTools/PatAlgos/interface/PATAuxiliaryProd.h
```diff
@@ -0,0 +1,66 @@
...
+#include "DataFormats/RecoCandidate/interface/RecoCandidate.h"
+#include "PhysicsTools/PatUtils/interface/StringParserTools.h"
+#include "PhysicsTools/PatUtils/interface/PATDiObjectProxy.h"
+
+namespace pat{
+
+  //
+  // Hcal depth energy fracrtion struct
+  //
+  struct HcalDepthEnergyFractionProd
```
slava77: perhaps better be a ```class```; sorry for calling this a struct in the earlier request

PhysicsTools/PatAlgos/interface/PATAuxiliaryProd.h
```diff
@@ -0,0 +1,66 @@
...
+  {
+  private:
+    //do not store
+    std::vector<float> hcalDepthEnergyFraction_;
```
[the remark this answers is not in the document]
slava77: May I suggest to be less repetitive in use of "HcalDepthEnergyFraction". This member data can be just ```fractions_;```

PhysicsTools/PatAlgos/plugins/PATPackedCandidateProducer.cc
```diff
@@ -135,6 +143,9 @@ pat::PATPackedCandidateProducer::PATPackedCandidateProducer(const edm::Parameter
   produces< std::vector<pat::PackedCandidate> > ();
   produces< edm::Association<pat::PackedCandidateCollection> > ();
   produces< edm::Association<reco::PFCandidateCollection> > ();
+
+  produces< edm::ValueMap<pat::HcalDepthEnergyFractionProd> > ("hcalDepthEnergyFractions");
```
slava77: perhaps add ```if (not pfCandidateTypesForHcalDepth_.empty() ) produces< edm::ValueMap ... ```

PhysicsTools/PatAlgos/plugins/PATPackedCandidateProducer.cc
```diff
@@ -320,7 +333,26 @@ void pat::PATPackedCandidateProducer::produce(edm::StreamID, edm::Event& iEvent,
...
+	     != pfCandidateTypesForHcalDepth_.end() ){
+	  if (!storeHcalDepthEndcapOnly_ || fabs(outPtrP->back().eta())>1.3 ){  // storeHcalDepthEndcapOnly_==false -> store all eta of selected 
+	                                                                // PF types, if true, only |eta|>1.3 of selected PF types will be stored  
+	    std::vector<float> hcalDepthEnergyFractionTmp {
+	        cand.hcalDepthEnergyFraction(1), cand.hcalDepthEnergyFraction(2),
+	        cand.hcalDepthEnergyFraction(3), cand.hcalDepthEnergyFraction(4),
+	    	cand.hcalDepthEnergyFraction(5), cand.hcalDepthEnergyFraction(6),
+	    	cand.hcalDepthEnergyFraction(7)
+		  };
+	    hcalDepthEFrac.reset(hcalDepthEnergyFractionTmp);
```
slava77: if the destination was an array, as already used in the PFCandidate, this could simply use ```cand.hcalDepthEnergyFractions()```  instead of spelling out all values.

PhysicsTools/PatAlgos/plugins/PATPackedCandidateProducer.cc
```diff
@@ -404,6 +437,14 @@ void pat::PATPackedCandidateProducer::produce(edm::StreamID, edm::Event& iEvent,
     iEvent.put(std::move(pf2pc));
     iEvent.put(std::move(pc2pf));
 
+    // HCAL depth energy fraction additions using ValueMap
+    std::unique_ptr<edm::ValueMap<HcalDepthEnergyFractionProd>> hcalDepthEnergyFractionsV(new edm::ValueMap<HcalDepthEnergyFractionProd>());
```
slava77: ```auto hcalDepthEnergyFractionsV = std::make_unique<edm::ValueMap<HcalDepthEnergyFractionProd> >();```
is a bit shorter

PhysicsTools/PatAlgos/plugins/PATPackedCandidateProducer.cc
```diff
@@ -404,6 +437,14 @@ void pat::PATPackedCandidateProducer::produce(edm::StreamID, edm::Event& iEvent,
...
+    edm::ValueMap<HcalDepthEnergyFractionProd>::Filler fillerHcalDepthEnergyFractions(*hcalDepthEnergyFractionsV);
+    fillerHcalDepthEnergyFractions.insert(cands,hcalDepthEnergyFractions_Ordered.begin(),hcalDepthEnergyFractions_Ordered.end());
+    fillerHcalDepthEnergyFractions.fill();
+    
+    iEvent.put(std::move(hcalDepthEnergyFractionsV),"hcalDepthEnergyFractions");
```
slava77: perhaps with an ```if (not pfCandidateTypesForHcalDepth_.empty() ) ```

PhysicsTools/PatAlgos/python/slimming/packedPFCandidates_cfi.py
```diff
@@ -29,6 +31,19 @@
 from Configuration.Eras.Modifier_run2_miniAOD_80XLegacy_cff import run2_miniAOD_80XLegacy
 run2_miniAOD_80XLegacy.toModify(packedPFCandidates, chargedHadronIsolation = "" )
 
+from Configuration.Eras.Modifier_run2_HCAL_2018_cff import run2_HCAL_2018
+run2_HCAL_2018.toModify(packedPFCandidates,
+    pfCandidateTypesForHcalDepth = cms.vint32(130,11,22,211,13),  # PF cand types for adding Hcal depth energy frac information
+                        # (130: neutral h, 11: ele, 22: photon, 211: charged h, 13: mu) # excluding e.g. 1:h_HF, 2:egamma_HF
+    storeHcalDepthEndcapOnly = cms.bool(True)
```
slava77: please drop type specifications for all parameters which already exist.

```python
pfCandidateTypesForHcalDepth = [130,11,22,211,13], 
storeHcalDepthEndcapOnly = True
```

PhysicsTools/PatAlgos/interface/PATAuxiliaryProd.h
```diff
@@ -0,0 +1,66 @@
...
+      for (auto it = begin (v); it != end (v); ++it){ 
+	hcalDepthEnergyFractionI_.push_back((uint8_t)((*it)*200.));
+	hcalDepthEnergyFraction_.push_back(*it);
+      }
+    }
+
+    // provide a full vector for each depth
+    std::vector<float> hcalDepthEnergyFraction() {
+      hcalDepthEnergyFraction_.clear();
+      for (auto it = begin (hcalDepthEnergyFractionI_); it != end (hcalDepthEnergyFractionI_); ++it)
```
slava77: given that the iorule and the constructor and the reset methods set the float fraction already, it seems like we do not need to reset/recompute the float vector and should simply return its value and also make the method const.

hatakeyamak:
Thanks @slava77. I will try to follow your suggestions.
One quick question: I actually thought using vector instead of array is a good idea, because when we store the depth info into packedcandiddates only in endcap, I prefer giving empty vectors to candidates in barrel and forward, rather than always-7-element arrays with all elements being zero (or some default values). But, if there is a good reason that we had better use arrays, I will be happy to try to switch.

slava77:
On 4/16/19 1:23 PM, Kenichi Hatakeyama wrote:
> One quick question: I actually thought using vector instead of array is
> a good idea, because when we store the depth info into packedcandiddates
> only in endcap, I prefer giving empty vectors to candidates in barrel
> and forward, rather than always-7-element arrays with all elements being
> zero (or some default values). But, if there is a good reason that we
> had better use arrays, I will be happy to try to switch.

OK to vectors, if the size is meant to vary.

PhysicsTools/PatAlgos/plugins/PATPackedCandidateProducer.cc
```diff
@@ -320,7 +333,26 @@ void pat::PATPackedCandidateProducer::produce(edm::StreamID, edm::Event& iEvent,
...
+	pat::HcalDepthEnergyFractionProd hcalDepthEFrac(dummyVector);
 	
+	// storing HcalDepthEnergyFraction information
+	if ( std::find(pfCandidateTypesForHcalDepth_.begin(), pfCandidateTypesForHcalDepth_.end(), abs(cand.pdgId())) 
+	     != pfCandidateTypesForHcalDepth_.end() ){
+	  if (!storeHcalDepthEndcapOnly_ || fabs(outPtrP->back().eta())>1.3 ){  // storeHcalDepthEndcapOnly_==false -> store all eta of selected 
+	                                                                // PF types, if true, only |eta|>1.3 of selected PF types will be stored  
+	    std::vector<float> hcalDepthEnergyFractionTmp {
+	        cand.hcalDepthEnergyFraction(1), cand.hcalDepthEnergyFraction(2),
+	        cand.hcalDepthEnergyFraction(3), cand.hcalDepthEnergyFraction(4),
```
slava77: this can also be ```std::vector<float> hcalDepthEnergyFractionTmp (cand.hcalDepthEnergyFractions().begin(), cand.hcalDepthEnergyFractions().end());```

hatakeyamak:
> On 4/16/19 1:23 PM, Kenichi Hatakeyama wrote: One quick question: I actually thought using vector instead of array is a good idea, because when we store the depth info into packedcandiddates only in endcap, I prefer giving empty vectors to candidates in barrel and forward, rather than always-7-element arrays with all elements being zero (or some default values). But, if there is a good reason that we had better use arrays, I will be happy to try to switch.
> OK to vectors, if the size is meant to vary.

ok!

hatakeyamak:
Thank you. I think I made all the suggested changes.

slava77:
@cmsbuild please test

PhysicsTools/PatAlgos/interface/PATAuxiliaryProd.h
hatakeyamak: For this one, I think we have to recompute when we are reading this information from MINIAOD ValueMap. At least, if I remove this recomputation part, I get an empty vector.

PhysicsTools/PatAlgos/src/classes_def.xml
hatakeyamak: > the goal should be to initialize fraction float vector from the int,

I agree and I think thta's what I am doing, but let me know if this still needs to be changed.

PhysicsTools/PatAlgos/interface/PATAuxiliaryProd.h
hatakeyamak: Now I see what you mean. After fixing the iorule, it looks like we don't have to do this recomputation, and I change it to const.

PhysicsTools/PatAlgos/src/classes_def.xml
hatakeyamak: Never mind. I understand what you mean now and I now use a new initFloatVector() here that converts the stored uint8 to float.

hatakeyamak:
I made a few additional changes after re-reading suggestions from @slava77 . We may have to run another round of jenkins test. Sorry.

fabiocos:
please test

DataFormats/PatCandidates/interface/HcalDepthEnergyFractions.h
```diff
@@ -0,0 +1,66 @@
...
+
+    // reset vectors
+    void reset(std::vector<float> v) {
+      fractions_ = v;
+      initUint8Vector();
+    }
+
+    // provide a full vector for each depth
+    std::vector<float> fractions() const {
+      return fractions_;
```
slava77: ```const std::vector<float>&``` to avoid copy

DataFormats/PatCandidates/interface/HcalDepthEnergyFractions.h
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    // provide info for individual depth
+    float fraction(unsigned int i) const {
+      if (i<fractions_.size()) return fractions_[i];
+      else return -1.;
+    }
+    
+    // provide a full vector (in uint8_t) for each depth
+    std::vector<uint8_t> fractionsI() const {
```
slava77: ```const std::vector<uint8_t>&``` to avoid copy

DataFormats/PatCandidates/interface/HcalDepthEnergyFractions.h
```diff
@@ -0,0 +1,66 @@
...
+  //
+  class HcalDepthEnergyFractions {
+
+  private:
+    //do not store
+    std::vector<float> fractions_;
+    //store
+    std::vector<uint8_t> fractionsI_;
+  public:
+    explicit HcalDepthEnergyFractions(std::vector<float> v):
```
slava77: ```const std::vector<float>&``` to avoid copy

DataFormats/PatCandidates/interface/HcalDepthEnergyFractions.h
```diff
@@ -0,0 +1,66 @@
...
+    }
+    
+    // produce vector of float from vector of uint8_t
+    void initFloatVector() {
+      fractions_.clear();
+      for (auto fracI : fractionsI_) fractions_.push_back(float(fracI)/200.);
+    }
+
+    // reset vectors
+    void reset(std::vector<float> v) {
```
slava77: ```const std::vector<float>&``` to avoid copy

PhysicsTools/PatAlgos/plugins/PATPackedCandidateProducer.cc
```diff
@@ -135,6 +143,9 @@ pat::PATPackedCandidateProducer::PATPackedCandidateProducer(const edm::Parameter
   produces< std::vector<pat::PackedCandidate> > ();
   produces< edm::Association<pat::PackedCandidateCollection> > ();
   produces< edm::Association<reco::PFCandidateCollection> > ();
+
+  produces< edm::ValueMap<pat::HcalDepthEnergyFractions> > ("hcalDepthEnergyFractions");
```
slava77: ```if (not pfCandidateTypesForHcalDepth_.empty() )```
to match L441.
Otherwise we are in a somewhat bad state of declaring that a branch is produced; but then no data is added to the event.

slava77 (commented):
we are almost there now

slava77:
@cmsbuild please test

PhysicsTools/PatAlgos/python/slimming/MicroEventContent_cff.py
```diff
@@ -146,6 +146,13 @@
 cms.untracked.PSet(branch = cms.untracked.string("EcalRecHitsSorted_reducedEgamma_reducedESRecHits_*"),splitLevel=cms.untracked.int32(99)),
 ])
 
+_run2_HCAL_2018_extraCommands = ["keep *_packedPFCandidates_hcalDepthEnergyFractions_*"]
+from Configuration.Eras.Modifier_run2_HCAL_2018_cff import run2_HCAL_2018
+run2_HCAL_2018.toModify(MicroEventContentMC, outputCommands = MicroEventContentMC.outputCommands + _run2_HCAL_2018_extraCommands)
```
slava77: is this really needed only for MC?
For data please modify ```MicroEventContent``` after L84.
I didn't realize this until I tested on data.

hatakeyamak: Oops. Thanks. Needs to be fixed. Hold on.

hatakeyamak:
I think I addressed the issue pointed out by @slava77

slava77:
@cmsbuild please test

slava77:
Here is a plot from JetHT2018C wf 136.876

all_orig-patVSsign1073-pat_RunJetHT2018Cwf136p876c_patHcalDepthEnergyFractionsedmValueMap_packedPFCandidates_hcalDepthEnergyFractions_PAT_obj_values__fraction0

It looks like the setup works OK.

hatakeyamak:
Good. I also checked locally for one data workflow before I made the last commit. So, I guess it looks good to sign off?

DataFormats/PatCandidates/src/classes_def_other.xml
```diff
@@ -71,6 +71,23 @@
...
+  <class name="edm::Wrapper<std::vector<pat::HcalDepthEnergyFractions> >"/>
+  <class name="edm::ValueMap<pat::HcalDepthEnergyFractions>" />
+  <class name="edm::Wrapper<edm::ValueMap<pat::HcalDepthEnergyFractions> >"/>
+
+  <ioread sourceClass = "pat::HcalDepthEnergyFractions" version="[1-]"
+                targetClass="pat::HcalDepthEnergyFractions"
+		source="std::vector<uint8_t> fractionsI_"
+		target="fractions_">
+  <![CDATA[ newObj->initFloatVector();
+  ]]>
```
slava77: Static analyzer picked up on this line
https://cmssdt.cern.ch/SDT/jenkins-artifacts/pull-request-integration/PR-26465/34296/llvm-analysis/report-9a8c17.html#EndPath

According to it, the code for this rule looks like the following

```c++
 pat::HcalDepthEnergyFractions* newObj = (pat::HcalDepthEnergyFractions*)target;

 if (oldObj) {}

 if (newObj) {}

  newObj->initFloatVector();

```

@pcanal @Dr15Jones 
By some magic (?) I would expect that the body of ```CDATA[``` should end up inside ```if (newObj) {```.
Are we using the syntax for ioread incorrectly or should we simply add ```if (newObj) {``` ?

We have the same pattern in a few other ioread rules in CMSSW. So, a more general solution is perhaps outside the scope of this PR.

slava77:
+1

for https://github.com/cms-sw/cmssw/pull/26465	e574dd3 
- jenkins tests pass and comparisons with the baseline show differences only in the output products. 
    - A simple monitoring of fractions was added to the reco comparisons and should appear in the next update of the cms-bot reco comparison scripts soon.

DataFormats/PatCandidates/src/classes_def_other.xml
pcanal: That code is there solely to prevent unused variable warning.  It needs to be updated to be ```(void)newObj```, etc.

fabiocos:
+1

fabiocos:
merge

santocch:
+1

PR #26485, opened 2019-04-17
New E/gamma Data members for UL

Sam-Harper:
#### PR description:
This is the E/gamma data format changes of https://github.com/cms-sw/cmssw/pull/26389 with absolutely necessary bug fixed. The VID / nano down stream changes will come in a seperate PR. 

It overhauls the photon pf isolation to be values actually used in IDs and MVAs and at the same times adds smajor,minor, alpha 2nd momement showershape variables which are useful for OOT photons.  In the electrons, it adds the HEEP flavour of track isolation and the conversion vtx probability and some helper functions for kf nhits and chi2. 

The only changes expected (other than the population of new data members) are that the selection on the miniAOD for photons is now using a different definition of charged isolation so there will be some content changes in the reducedEgamma. This is a desired feature. There will also be rare changes the reco::Photon pf photon and neutral isolation values. 

I'm running the matrix tests now.


slava77:
@cmsbuild please test

Sam-Harper:
so I couldnt see what unit test failed, looking through the log I couldnt find anything. 

Changes are as expected, slimmedPhoton changes and chargedHadIso. Also the MET uncertainties change but that is due to their improper calculation and therefore is their problem (and they have a PR in progress to fix it)

slava77:
@cmsbuild please test

with a hope to get the unitTests complete

Sam-Harper:
Locally testing, I got all the unit tests passed except for tau embedding which can not pass.

slava77:
I'm looking at the cms-bot outputs in comparisons:

DQM plots:
- chHadIso <= OK, expected
- METUnc_PhotonEnDown and METUnc_UnclusteredEnDown <= known/somewhat expected
- gedPhotonAnalyzer : here some photons appear to have small  differences in many variables; e.g in wf 4.53 (Photon2012)
wf4 53_allPho20_sIetaIeta

- somewhat of a curiosity : a change in the HCAL global hit validation
wf25 0_HBRes_globalV

I suspect that this feeds through the reduced hits
https://github.com/cms-sw/cmssw/blob/31025a354c2259e2f09358ce9741cf7be0a13323/Validation/GlobalRecHits/src/GlobalRecHitsAnalyzer.cc#L657-L660

@Sam-Harper 
Please clarify if changes in photons are expected (it wasn't obviously clear in the PR description)

> There will also be rare changes the reco::Photon pf photon and neutral isolation values.

didn't seem obviously related to the observations in the gedPhotonAnalyzer

DataFormats/EgammaCandidates/interface/GsfElectron.h
```diff
@@ -183,11 +183,18 @@ class GsfElectron : public RecoCandidate
     // forward core methods
     SuperClusterRef superCluster() const override { return core()->superCluster() ; }
     GsfTrackRef gsfTrack() const override { return core()->gsfTrack() ; }
-    virtual TrackRef closestTrack() const { return core()->ctfTrack() ; }
+    //so it turns out closestCtfTrackRef is overriden by pat::Electron but not closestTrack()
+    //this results in surprising behaviour, ie two functions which looked identical 
+    //would give different results so now closestTrack just calls closestCtfTrackRef
+    virtual TrackRef closestTrack() const { return closestCtfTrackRef() ;}
```
slava77: [responding/continuing from https://github.com/cms-sw/cmssw/pull/26389#pullrequestreview-228411055]
It's curious that the closestTrack method was added for purposes of PAT embedding https://github.com/cms-sw/cmssw/commit/ba98a5f008648802611e5b2630137ccf2d9f52a6
However, just a few days later, the PAT side decided to use the other method
https://github.com/cms-sw/cmssw/commit/fa396c23edbf071da765c823dccb8d5990102333

git grep in the baseline shows only one place where this method is used
Calibration/EcalAlCaRecoProducers/plugins/AlCaElectronTracksReducer.cc
So, apparently there was a known secret not to use closestTrack  until you stepped into it in https://github.com/cms-sw/cmssw/commit/13fb21b22c6c5130721ecd37df495771cad1dcbb

Overall, this looks to me like a good case to remove this method.

Sam-Harper: yes, this is a mess. However I prefer not to remove methods and make code not compile in 10_6 that did in 10_2, given that we want to properly fix this in 11_X

slava77:
On 4/19/19 12:21 AM, Sam Harper wrote:
> yes, this is a mess. However I prefer not to remove methods and make
> code not compile in 10_6 that did in 10_2, given that we want to
> properly fix this in 11_X
> 

Please clarify which code will not compile.

Sam-Harper:
Some hypothetical users code. Not all code is part of the release.

Sam-Harper:
I mean whats the problem here? We have this function. It probably shouldnt exist, that I agree on. But it doesnt cost anything to keep it. And it may have a downside in the fact that somebody downstream may have to adapt their code because they were using it for whatever reason. 

So why not keep it for the clean up which is happening in 11_X so users only have to adapt their code once? I think my problem is that I fail to see why this must be done now rather than later.

slava77:
On 4/19/19 4:55 AM, Sam Harper wrote:
> Some hypothetical users code. Not all code is part of the release.
> 
> —

Do you know of any recommendations from EGM that use this method?

IIUC, the concern is to keep maintain forward compatibility for a
hypothetical software use case.
Given an obvious replacement is available, I'm not sure what the problem is.

slava77:
On 4/19/19 5:10 AM, Sam Harper wrote:
> But it doesnt cost anything to keep it.

It apparently did cost something to keep it, adding from your recent
stepping into the bug and the reason that this method needs an update.

Sam-Harper:
but the bug is neutralized now. So its harmless.

Sam-Harper:
on the actual problem, no I wasnt expecting a change in photon sigmaIEtaIEta and right now I'm baffled why this would be the case. I'm investigating it (I've reproduced it locally)

slava77:
> On 4/19/19 4:55 AM, Sam Harper wrote: Some hypothetical users code. Not all code is part of the release. —
> Do you know of any recommendations from EGM that use this method? IIUC, the concern is to keep maintain forward compatibility for a hypothetical software use case. Given an obvious replacement is available, I'm not sure what the problem is.

After a bit more thought, I think the presence of that hypothetical code is a very good example to remove this method which is known to be buggy in the past.
This is a good way to communicate that there was a problem, instead of silently fixing it.

slava77:
Just to confirm on the unitTest. I see only ```testTauEmbeddingProducers had ERRORS```, attempting to read a deleted file in /store/user/swayand, which is expected to fail and is not related to this PR.

Sam-Harper:
Okay back on to the serious issue, the photon sigmaIEtaIEta histogram change. So this worried me as it made no logical sense.  I have done a raw comparison of the miniAOD files and no difference in sigmaIEtaIEta is found (this is a photon by photon comparison, not a histogram)

I have then dug into PhotonAnalyzer (which is another bit of code I greatly look forward to removing, its been added to The List). 

So it uses charged hadron isolation in its selection which will explain the change. As a test, I set it to use the old version and it fixed the issue, now perfect agreement
https://github.com/cms-sw/cmssw/blob/master/DQMOffline/EGamma/plugins/PhotonAnalyzer.cc#L1208

Anyways this confused me because this is "All Photons", as in it should have no selection added to it. A puzzle indeed. 

So what was going on. Well here is how it fills its histograms for a given variable 
https://github.com/cms-sw/cmssw/blob/master/DQMOffline/EGamma/plugins/PhotonAnalyzer.cc#L1177
as an aside truely awful name btw, 3DHistoVector means  a 3D dimensional vector containing histograms not a vector containing 3D histos. 

Anyways I then noticed this curiosity. Note how it fills the type = 0 histograms always, regardless of what type actually is. So in the case type = 0, it fills it twice.  (type 0 is the "All Photons" hist).   So "All Photons" is " All Photons with photons failing selection added in twice). You can see it clearly from that histogram, notice above >0.014 (which means it fails the selection), all entries are multiples of 2.  Type is set here btw : https://github.com/cms-sw/cmssw/blob/master/DQMOffline/EGamma/plugins/PhotonAnalyzer.cc#L770

So that explains why the selection is influencing the histogram which I had naively assumed was selection free.  

So this is an artifact of the selection used by the dqm module + a bug in how it does this.  Therefore not a problem. 

In general changing the chargedHadronIso will causes many minor things like this to happen. However this to me is a feature not a bug. Before all these selections where using a different charged hadron isolation to what the e/gamma IDs were using and now they are unified.

Sam-Harper:
I have removed the closestTrack method. Note, you did not convince me this was the correct action, simply that we would not agree on this area. In the interest of not wasting both our time on this further, I made the change.

slava77:
@cmsbuild please test

DataFormats/EgammaCandidates/interface/GsfElectron.h
```diff
@@ -183,11 +183,14 @@ class GsfElectron : public RecoCandidate
...
     GsfTrackRef gsfTrack() const override { return core()->gsfTrack() ; }
-    virtual TrackRef closestTrack() const { return core()->ctfTrack() ; }
     float ctfGsfOverlap() const { return core()->ctfGsfOverlap() ; }
     bool ecalDrivenSeed() const { return core()->ecalDrivenSeed() ; }
     bool trackerDrivenSeed() const { return core()->trackerDrivenSeed() ; }
     virtual SuperClusterRef parentSuperCluster() const { return core()->parentSuperCluster() ; }
+    bool closestTrackValid() const { return closestCtfTrackRef().isAvailable() && closestCtfTrackRef().isNonnull() ; }
+    //methods used for MVA variables
+    float closestTrackChi2() const; 
+    int closestTrackNHits() const;
```
slava77: I noticed now that the implementation is ```hitPattern().trackerLayersWithMeasurement()```.
Shouldn't the method be called ```closestTrackNLayers``` ? Otherwise this seems potentially quite confusing.

Sam-Harper: that was what it was named for the MVA variables.  Unfortunately this name is grandfathered and  I dont really want to change it.

slava77: TrackChi2 is also not a full chi2, but a normalized chi2. Slightly less confusing than the NHits which are actually layers.

Instead of introducing confusingly named methods in general use classes, it seems more practical then to let the code that fills the MVA variables call them to its liking.

slava77: Please clarify which MVA code had this naming.
Thank you.

I see one case:
https://github.com/cms-sw/cmssw/blob/570d18fd35f56f0e8865674205dacc20b1bd9328/EgammaAnalysis/ElectronTools/src/EGammaMvaEleEstimator.cc#L1155
an internal variable in the code refers to hits, but then when it's passed to the MVA for evaluation, it appropriately refers to layers here
https://github.com/cms-sw/cmssw/blob/570d18fd35f56f0e8865674205dacc20b1bd9328/EgammaAnalysis/ElectronTools/src/EGammaMvaEleEstimator.cc#L256
and here
https://github.com/cms-sw/cmssw/blob/570d18fd35f56f0e8865674205dacc20b1bd9328/EgammaAnalysis/ElectronTools/src/EGammaMvaEleEstimator.cc#L316

slava77: perhaps adding a ```mva_``` prefix and a clear note that the returned value does not match the name could be acceptable.
Also, if the implementation can be moved inline it will also be more clear.

Sam-Harper: Oh I made my peace with this decided to change it as suggested and do as much work as it takes to get this in. Means I wont do other things of course because my time is finite, but fine. 

In fact, I now named them closestCtfTrack as closestTrack is gone. 

The MVA uses it here: 
https://github.com/cms-sw/cmssw/blob/master/RecoEgamma/ElectronIdentification/interface/ElectronMVAEstimatorRun2.h#L54-L55

Sam-Harper: well that was fun! So turns out including reco::Track header file (necessary to inline the methods as requested) caused AnalysisDataFormats/Egamma not to compile. This is because GsfElectronFwd.h improperly includes GsfElectron.h  (so not really a forward declaration as much as an actual declaration, sigh).  The dictionary of AnalysisDataFormats/EGamma therefore needed the CLHEP dependence in the build file.

And this is a classic example of what looks like small simple change ending up taking significant time.

slava77:
On 4/23/19 8:57 AM, Sam Harper wrote:
> And this is a classic example of what looks like small simple change
> ending up taking significant time.
> 

sorry about that.
I guess it's done/resolved now.


I really wanted the inlines only with the bad names of the methods to
mitigate/minimize confusion.
Thank you for updating the names to be now matching the return values
meaning.

slava77:
@cmsbuild please test

slava77:
+1

for https://github.com/cms-sw/cmssw/pull/26485	ace1272
- jenkins tests as of 99f73a1 https://github.com/cms-sw/cmssw/pull/26485#issuecomment-485487174  pass and comparisons with the baseline show differences in expected places
    - reco comparisons show changes in variables with the following pattern "chargedHadronIso.png\|slimmedPhotons\|reducedEgamma\|_slimmedMETs[^$]*uncertainties\|_slimmedMETs[^$]*significance"
    - DQM comparisons have differences in plots matching the following pattern "chHadIso\|METUnc_PhotonEnDown\|METUnc_UnclusteredEnDown\|gedPhotonAnalyzer" . 
    - for both cases the changes are expected/explained earlier in the thread https://github.com/cms-sw/cmssw/pull/26485#issuecomment-485258812 https://github.com/cms-sw/cmssw/pull/26485#issuecomment-484690123
- tests made as of 99f73a1 https://github.com/cms-sw/cmssw/pull/26485#issuecomment-485487174 should cover all relevant comparisons; the latest updates require only the code compilation/building step to complete (and we are apparently facing some failures in cmsbuild eos reads, unrelated to this PR).
- local tests of nanoAOD workflows (136.7722,136.7952,136.8521,1325.6,1325.7,1325.8,1325.9,1329.1) reading from old data show no differences, as expected for this PR.

slava77:
@cmsbuild please test

Sam-Harper:
> Thank you for updating the names to be now matching the return values meaning.

No problem. In fact I should really apologise for being a bit short on this PR. It was filtered through deadline pressure but you were right about almost everything and it does significantly improve things in the long run. I do feel a large part of CMS's competitive advantage is our software and these code reviews and yourself make a huge difference to it. In fact having them is vital to CMS and I think we would be a lot worse of without them and yourself to do them. The price to pay is that we take longer to do things which can be frustrating but better than having to  redo all the ultra legacy! Thank you for this, its very much appreciated.

slava77:
@tocheng @pohsun @tlampen @christopheralanwest
alca signature is needed for this PR to confirm changes in Calibration/EcalAlCaRecoProducers/plugins/AlCaElectronTracksReducer.cc 
Please check and comment or sign.
Thank you.

slava77:
@kpedro88 @fabiocos 
there is a follow up PR #26486 waiting for this one to be integrated.
Please check and comment or merge.
Thank you.

Calibration/EcalAlCaRecoProducers/plugins/AlCaElectronTracksReducer.cc
```diff
@@ -51,11 +51,7 @@ void AlCaElectronTracksReducer::produce (edm::Event& iEvent,
 
   for (eleIt=electronCollection->begin(); eleIt!=electronCollection->end(); eleIt++) {
     // barrel
-#ifndef CMSSW42X
-    TrackRef track = (eleIt-> closestTrack() ) ;
-#else
     TrackRef track = (eleIt-> closestCtfTrackRef());
```
christopheralanwest: Is the change in the logic here intentional? L54 had an `#ifndef` statement, not an `#ifdef`.

Sam-Harper: there is no logic change. 

closestTrack() no longer exists and was removed as it is broken for pat::Electrons. For reco::GsfElectrons, its the same as closestCtfTrackRef().

slava77: The purpose of the change here is due to removal of ```closestTrack``` method, which was always equivalent for reco::GsfElectrons but was also always broken for pat::Electrons.

```closestTrack``` was introduced for PAT purposes in 5_2_0 but was apparently never made useful for PAT (and was even buggy for the pat::Electrons).
```closestCtrTrackRef``` existed since CMSSW_3_1_0.

kpedro88:
please test
(check if addon test failure was transient)

fabiocos:
@christopheralanwest any further doubt/question?

christopheralanwest:
+1

santocch:
+1

fabiocos:
+1

PR #19964, opened 2017-07-28
Adding reduced GsfTracks to miniAOD

andreypz:
Hi,

This commit is to add reduced gsf tracks collection to miniAOD. The tracks added are:
 * The main track of GsfElectronCore
 * All ambiguous tacks of GsfElectron (if there are any)

Change in the file size was estimated with *test/patMiniAOD_standard_cfg.py* script on two RelVal samples: **TTbar** and **ZtoEE**, based on 500 events:
 * In TTbar sample the increase is from 32470 Kbytes to 32524 Kbytes (0.17%). Increase per event: 108 bytes.
 * In ZEE sample the increase from 25427 Kbytes to 25460 Kbytes (0.13%). Increase per event: 65 bytes.

This addition was presented in [X-POG meeting on 18-Oct-2016](https://indico.cern.ch/event/575788/), in [this slides](https://indico.cern.ch/event/575788/contributions/2343082/attachments/1356862/2051472/Zenz-DalitzMiniAODNeeds-XPOG-v1.pdf).  
This was also privately discussed with @arizzi @gpetruc @sethzenz @bendavid.

The intent for this code to be used for miniAOD of 2017 data/MC and legacy re-miniAOD of 2016 data.

This supersedes pull #19597

PR #26486, opened 2019-04-17
Adapting VID & Nano to the new E/gamma UL format + adding HEEP V7.1

Sam-Harper:
#### PR description:
This is the adaption PR for https://github.com/cms-sw/cmssw/pull/26485 which was originally https://github.com/cms-sw/cmssw/pull/26389 (this PR is currently exactly the same as that PR just ordered). It contains all of https://github.com/cms-sw/cmssw/pull/26485, I was not sure how to proceed here as it needs those changes.  This PR starts from commit https://github.com/cms-sw/cmssw/commit/2a8a8bf5b61176328a121b3fa29f862dc7bff7eb. 

Testing on going once the first PR is tested. 

Relative to https://github.com/cms-sw/cmssw/pull/2648, there should be no change in output with the exception of the HEEP IDs which will suffer from rounding effects leading to a difference to trk isolation when now calculated using tracks rather than packed candidates.


Sam-Harper:
fixes the issue with the summer 16 HLT preselection. Also at the same time I released I broke the HEEP V5.0 however rather than fix it, I removed it from the release as it is old. Including the no-longer run tests for it.

peruzzim:
please test

I test it already to see if it runs and proceed with more detailed checks on nanoAOD content

slava77:
On 4/23/19 5:27 PM, cmsbuild wrote:
> I found errors in the following unit tests:
> 
> ---> test testRecoMETMETProducers had ERRORS
> ---> test runtestRecoEgammaPhotonIdentification had ERRORS
> 

both are due to issues in the input file reading; apparently not related
to this PR.

slava77:
@Sam-Harper 

part 1 #26485 is now merged.
Please rebase this PR to hide diffs from already merged commits as can be seen in the web interface.

perrotta:
@cmsbuild please test

RecoEgamma/ElectronIdentification/plugins/cuts/GsfEleMinEcalEtCut.cc
```diff
@@ -0,0 +1,46 @@
...
+  minEt_(c.getParameter<double>("minEt"))
+{
+
+}
+
+CutApplicatorBase::result_type 
+GsfEleMinEcalEtCut::
+operator()(const reco::GsfElectronPtr& cand) const{  
+  const reco::GsfElectron& ele = *cand;
+  const float sinTheta = ele.energy()!=0 ? ele.et()/ele.energy() : 0.;
```
slava77: pt/p ?

RecoEgamma/ElectronIdentification/python/Identification/heepElectronID_HEEPV71_cff.py
```diff
@@ -0,0 +1,95 @@
...
+from PhysicsTools.SelectorUtils.centralIDRegistry import central_id_registry
+
+# Common functions and classes for ID definition are imported here:
+from RecoEgamma.ElectronIdentification.Identification.heepElectronID_tools import HEEP_WorkingPoint_V1,configureHEEPElectronID_V71
+
+#the same HEEP V70 but now has calo et cut of >5 GeV and isolation cut is relaxed vs calo et
+
+# The cut values for the  Barrel and Endcap
+idName = "heepElectronID-HEEPV71"
+WP_HEEP71_EB = HEEP_WorkingPoint_V1(
```
slava77: why is this not just ```WP_HEEP71_EB = WP_HEEP70_EB.clone(idName)``` ?
copy-paste is not the best development strategy.

slava77:
@cmsbuild please test workflow 136.7722,136.7952,136.8521,1325.6,1325.7,1325.8,1325.9,1329.1

to check integrity of  nano wfs available in the matrix

slava77:
the errors are  apparently related to this PR changes.

both workflows are using ```--era Run2_2016,run2_miniAOD_80XLegacy ```

```
Unrunnable schedule
Module run order problem found: 
slimmedElectronsUpdated after slimmedElectronsTo106X [path nanoAOD_step], slimmedElectronsTo106X after heepIDVarValueMaps [path nanoAOD_step], heepIDVarValueMaps consumes slimmedElectronsUpdated
 Running in the threaded framework would lead to indeterminate results.
 Please change order of modules in mentioned Path(s) to avoid inconsistent module ordering.

```

slava77:
There are some changes in nanoAOD outputs compared to the baseline (CMSSW_10_6_X_2019-04-30-2300) in local tests

One variable in wf 10824 (using gensim from relval, not resim as in jenkins tests; otherwise a similar 10 event setup)
all_sign1071bVSorig_TTbar13TeV2018nanoEDMwf10824p0c_nanoaodFlatTable_photonTable__NANO_obj_floats__pfRelIso03_all_251

Quite a few  diffs in wf 136.8521 (nano on 102X 2018 data from JetHT2018A).
Note that there are no differences in the 2017 nano workflow (136.7952 checked); no statement on the 2016 case which is broken.

dr03TkSumPtHEEP is apparently not filled anymore (red histo is with this PR)
all_sign1071b-nanoVSorig-nano_RunJetHT2018AnanoEDMwf136p8521c_nanoaodFlatTable_electronTable__DQM_obj_floats__dr03TkSumPtHEEP_13


The electron mvaFall17V1Iso (and noIso and also mvaFall17V2\*) is significantly different
all_sign1071b-nanoVSorig-nano_RunJetHT2018AnanoEDMwf136p8521c_nanoaodFlatTable_electronTable__DQM_obj_floats__mvaFall17V1Iso_30

For photons:
mvaIDV is quite different  (mvaIDv2 is more similar but differs as well)
all_sign1071b-nanoVSorig-nano_RunJetHT2018AnanoEDMwf136p8521c_nanoaodFlatTable_photonTable__DQM_obj_floats__mvaID_191

Considering no changes in the 2017 nano setup, it looks like the 2018 case is missing some modifier updates as well.

arizzi:
can this be tested in cms-nanoAOD repository so that nanoaod CI is triggered?

peruzzim:
yes, I'm on it, we had agreed with @Sam-Harper that it would be tested once it would at least run and pass tests here

Sam-Harper:
Hi Slava, thank you for highlighting the issues, made it quick for me to fix. So I've fixed the unrunable tests, just a silly thing, just doing some more testing first. 

Secondly, the 2018 is an issue and will continue to be an issue because it has no modifier. Simply put it needs a modifier to be able to read pre 106X data in 106X.  What is the solution here?  I suppose define a modifier I guess?

Or we could revert nano to always running the existing code however this is problematic. Or I could look into some on demand modifier-less updater for nano.

Sam-Harper:
so just another question, is there away in CMSSW to get the class version of the object being read. Then I could use that to determine whether the updater needs to update anything.

Sam-Harper:
I suppose I could consult the provanance of the collection and do a if(CMSSW_VERSION<X)

arizzi:
@Sam-Harper can't you use IOREAD rules instead to trigger some function only on some class versions?

arizzi:
ah sorry you mean config updater? ... forget my comment then

Sam-Harper:
@arizzi, yep thats the issue. 

So I'm going to work to the following solution as an experiment. I'm always going to run the updator but the updator will now be aware of the CMSSW version of the orginal object and do nothing if its produced with the latest version.  It will be configurable (always update, update only if last version is an old CMSSW version, update only if any parent is an old CMSSW version). 

@slava77 , let me know if you think this is a terrible idea and I should just use a modifier or its worth seeing the prototype before we make decision. 

Best,
Sam

arizzi:
mhh hold on, I'm confused to write about provenance of the collection... so it is about updating a collection not a configuration? I mean the configuration is made before you read any collection... can you clarify a bit?
If you have to decide what to run or not when reading back an object from the root file based on the class version then IOREAD is the way to go. Otherwise, can you please clarify the use case?

slava77:
> Simply put it needs a modifier to be able to read pre 106X data in 106X.

should the ```run2_nanoAOD_102Xv1``` do the job?
Are we simply missing it in the wf 136.8521 definition (this wf is reading 2018A miniAODv1 and has only  ``` --era Run2_2018 ```) ?
@peruzzim

slava77:
On 5/3/19 5:57 AM, Sam Harper wrote:
> so just another question, is there away in CMSSW to get the class
> version of the object being read. Then I could use that to determine
> whether the updater needs to update anything.
> 
> —

one practical way is to add a data member to the class with a flag (or a
bitmask)
to highlight relevant integrity of some data.
This will have almost zero cost in storage due to compression.

slava77:
> mhh hold on, I'm confused to write about provenance of the collection... so it is about updating a collection not a configuration? I mean the configuration is made before you read any collection... can you clarify a bit?

changes in the configuration based on expected inputs are supported with process modifiers (era or procmodifier). We should assume support for just a set of production dataset variants and not attempt to support all possible evolutions of specific classes in intermediate releases.

Sam-Harper:
hi @slava77 , sure if run2_nanoAOD_102Xv1 is the modifier we want to read 102Xv1 data.  @peruzzim ? I'm happy to use this and the problem is solved.

@arizzi.  IOREAD is impossible unless I can somehow access event products in it which I dont see how to. 

I'm suggesting to always run the updator and nano to use the updated collection. However the updator will only actually modify things if it needs to. There are several ways to achieve this, ie the bit mask @slava77 just suggested.  Or by inspecting provenance to see what release they were produced in.

So the reason I dont want the updator to update objects which are already updated is that due to AOD/miniAOD precision changes, variables computed from miniAOD diff slightly and for isolations this can be large (rare) changes (ie if it just passes/fails an eta/phi cut). So at that point VID would differ between nano and miniAODs which is bad.

peruzzim:
> hi @slava77 , sure if run2_nanoAOD_102Xv1 is the modifier we want to read 102Xv1 data. @peruzzim ? I'm happy to use this and the problem is solved.

yes, this is what that modifier is for

slava77:
> > hi @slava77 , sure if run2_nanoAOD_102Xv1 is the modifier we want to read 102Xv1 data. @peruzzim ? I'm happy to use this and the problem is solved.
> 
> yes, this is what that modifier is for

It appears that we need a separate PR which would add run2_nanoAOD_102Xv1 to 136.8521  and similar. Then after that is merged, we can cleanly test this PR.

slava77:
> It appears that we need a separate PR which would add run2_nanoAOD_102Xv1 to 136.8521 and similar. Then after that is merged, we can cleanly test this PR.

ah, NM. These are not a part of the standard jenkins tests.
So, for private testing I can just make a baseline with the era already added.
It may be more expedient and just as safe to modify these workflows just in this PR.

Sam-Harper:
cool, this should all just work, I'll upload the fix for the 2016 in about an hour, I made another fix at the same time and wanted to be sure it didnt break to many other things.

Sam-Harper:
okay this should now work in 2016 and also for with the run2_nanoAOD_102Xv1 modifier (there was a small bug)

Sam-Harper:
oh you are kidding me, why did we put the white space formatting rules before we got 10_6_X out the way

slava77:
code-checks

there was a notice very recently that this should go away for now

Sam-Harper:
thanks. Having now seen the output, I'll be giving some feedback on that as well. While I agree that nobody will like everything about the style as its all personal taste, the 80 character limit seems to limit readability for me. Its not the 80s, monitors have more horizontal not vertical these days.

slava77:
>  the 80 character limit seems to limit readability for me.

Isn't it set to 120 
https://github.com/cms-sw/cmssw/blob/master/.clang-format#L4
?

Anything larger than the width of the github web interface display is inconvenient at least for the code review and browsing.

slava77:
@cmsbuild please test

arizzi:
well if you need access to the event you can use the IOREAD to set a "toBeUpdated" flag to different values for different class versions

Sam-Harper:
> well if you need access to the event you can use the IOREAD to set a "toBeUpdated" flag to different values for different class versions

I think this was exactly what Slava already suggested. Or could be achieved (less reliably) by looking at provenance information. Anyways it turns out it there is already a nice modifier (ie we do it at the config level) for this so this feature is not strictly necessary right now so wont be done.

slava77:
@cmsbuild please test workflow 136.7722,136.7952,136.8521,1325.6,1325.7,1325.8,1325.9,1329.1

PhysicsTools/NanoAOD/python/electrons_cff.py
```diff
@@ -438,8 +469,11 @@ def _get_bitmapVIDForEle_docstring(modules,WorkingPoints):
 electronSequence = cms.Sequence(bitmapVIDForEle + isoForEle + ptRatioRelForEle + slimmedElectronsWithUserData + finalElectrons)
 electronTables = cms.Sequence (electronMVATTH + electronTable)
 electronMC = cms.Sequence(electronsMCMatchForTable + electronMCTable)
+from RecoEgamma.ElectronIdentification.heepIdVarValueMapProducer_cfi import heepIDVarValueMaps
+_updateTo106X_sequence =cms.Sequence(heepIDVarValueMaps + slimmedElectronsTo106X)
+heepIDVarValueMaps.dataFormat = 2
```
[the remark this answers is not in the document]
slava77: it looks like comments from the review a week ago do not have any follow up.
@Sam-Harper please check.
Thank you.

RecoEgamma/ElectronIdentification/plugins/cuts/GsfEleMinEcalEtCut.cc
slava77: and here

RecoEgamma/ElectronIdentification/python/Identification/heepElectronID_HEEPV71_cff.py
slava77: and here

slava77:
I was expecting that this PR will add ```run2_nanoAOD_102Xv1``` to the 2018 nano workflows.

slava77:
After the latest updates there are fewer changes in nanoAOD outputs

- 2018 re-nano based on wf 136.8521 with ```run2_nanoAOD_102Xv1``` added to the baseline and this PR tests shows no differences. 
- 2017 re-nano (136.7952 and 1325.6) show no differences
- 2016 re-nano (136.7722 for data and  1329.1 for MC) show small differences in the MVA plots. Are these expected?

e.g. 
all_sign1071b-pass-fda2a6d-nanoVSorig-nano_RunJetHT2016HnanoEDMwf136p7722c_nanoaodFlatTable_electronTable__DQM_obj_floats__mvaFall17V1noIso_38
all_sign1071b-pass-fda2a6d-nanoVSorig-nano_ZEE13nanoEDMin2016wf1329p1c_nanoaodFlatTable_electronTable__DQM_obj_floats__mvaFall17V1noIso_42
all_sign1071b-pass-fda2a6d-nanoVSorig-nano_ZEE13nanoEDMin2016wf1329p1c_nanoaodFlatTable_photonTable__DQM_obj_floats__mvaID_237

slava77:
> I was expecting that this PR will add `run2_nanoAOD_102Xv1` to the 2018 nano workflows.

I created #26672

PhysicsTools/NanoAOD/python/electrons_cff.py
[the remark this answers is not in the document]
Sam-Harper: yes sorry I missed them as I was sort of away last week. 

The implication that nano will never run on AOD inputs, miniAOD has to be present, even if its being produced on the fly in the job.  Therefore we force it to always use the miniAOD inputs, even if AOD is present, it is more efficient this way as AOD it has to approximate the packed candidates in a special sequence to avoid rounding errors going between AOD and miniAOD 

So =2 is the correct behaviour here as we always want to run on miniAOD

RecoEgamma/ElectronIdentification/plugins/cuts/GsfEleMinEcalEtCut.cc
Sam-Harper: done

RecoEgamma/ElectronIdentification/python/Identification/heepElectronID_HEEPV71_cff.py
Sam-Harper: agreed, done

PhysicsTools/NanoAOD/python/electrons_cff.py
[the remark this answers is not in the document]
slava77: the point here is that it will not work with miniAOD on the fly because  heepIDVarValueMaps is apparently used to create some IDs in the miniAOD itself and you are modifying it here in a nano configuration, which is a wrong thing to do in the context of possibly running together with miniAOD.
So, if a miniAOD part can possibly be considered to be made on the fly together with NANO modules, this change should be made in a different way.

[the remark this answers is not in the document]
Sam-Harper: Yes, I see your point. In fact I now realise as a result nano can not run at the same time as miniAOD due to how it uses VID.  So this doesnt really do anything it always needs miniAOD and AOD is not an option.  Can I just leave it?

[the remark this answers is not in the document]
peruzzim: I take note of this, we can address it when preparing a merged MINI+NANO workflow that we don't need at the moment, though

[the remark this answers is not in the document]
slava77: OK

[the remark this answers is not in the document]
Sam-Harper: thank you. Also @peruzzim , when you are doing that let me know, we might have to revisit a few things (its not hard to do it inefficiently, just by having two seperate VID modules but perhaps not optimal)

slava77:
It looks like what remains is a response to https://github.com/cms-sw/cmssw/pull/26486#issuecomment-489786491

> 2016 re-nano (136.7722 for data and 1329.1 for MC) show small differences in the MVA plots. Are these expected?

Sam-Harper:
No I'm surprised by this, electrons should be a null op so I need to check it. I do know at one point in the miniAOD it was identical (its how I found so many fascinating bugs) so I need to figure out what happened.

slava77:
> No I'm surprised by this, electrons should be a null op so I need to check it. I do know at one point in the miniAOD it was identical (its how I found so many fascinating bugs) so I need to figure out what happened.

@Sam-Harper 
please let me know if there was some progress in understanding this remaining issue.
Thank you.

PhysicsTools/NanoAOD/python/electrons_cff.py
```diff
@@ -4,13 +4,30 @@
...
+    phoChargedHadWorstVtxIso = "",
+    phoChargedHadWorstVtxConeVetoIso = ""
+)
+#we have dataformat changes to 106X so to read older releases we use egamma updators
+slimmedElectronsTo106X = cms.EDProducer("ModifiedElectronProducer",
+    src = cms.InputTag("slimmedElectrons"),
+    modifierConfig = cms.PSet( modifications = cms.VPSet(ele9X105XUpdateModifier) )
+)
+#might as well fix 80X while we're at it although the differences are not so relavent for nano
+run2_miniAOD_80XLegacy.toModify( slimmedElectronsTo106X.modifierConfig.modifications, prependEgamma8XObjectUpdateModifier )
```
slava77: I suspect that the differences in 2016 are coming from here.
Downstream in nano ```process.slimmedElectronsUpdated``` is using ```slimmedElectons``` out of the box in the baseline. With this PR it's using slimmedElectronsTo106X instead and these now come with ```EG8XObjectUpdateModifier``` mods.

In the tested nano workflows the inputs are the miniAODs made in 80X. 
Do I understand correctly that the slimmedElectrons are not good for downstream out of the box?

It looks like this is a bugfix.
@peruzzim @Sam-Harper 
please check and confirm

peruzzim: @slava77 I understand that here Sam aimed at replacing `slimmedElectrons` with `slimmedElectronsTo106X` in all releases < 106X, including 80X.
In 80X only, `slimmedElectronsUpdated` is now built starting from `slimmedElectronsTo106X`, and then it seems to me it is used as src in all places where this is explicitly used downstream.

So what downstream should be using is:
- `slimmedElectronsUpdated` in 80X
- `slimmedElectronsTo106X` in [94X - 102X]
- `slimmedElectrons` in 106X

slava77: Answer to self: this is not the origin of the differences.
After dropping the 80X mods with ```--customise_command="process.slimmedElectronsTo106X.modifierConfig.modifications.pop(0)\n"``` I see the same output as without this customise (the same differences wrt the baseline).

slava77: > So what downstream should be using is:
> 
>     * `slimmedElectronsUpdated` in 80X
> 
>     * `slimmedElectronsTo106X` in [94X - 102X]
> 
>     * `slimmedElectrons` in 106X

I'm partly assuming that this is indeed the case because 
- 102X and 94X nano outputs show no differences
- in 80X I actually checked in the config explicitly.

Still, apparently the differences in outputs for the nano using 2016/80XLegacy are unexpected.

peruzzim: question for @Sam-Harper: are `ModifiedElectronProducer` - with list of modifications tuned for 80X - and `PATElectronUpdater` supposed to work well together? Or could there be some duplication happening there?

slava77:
I enabled debug for ElectronMVAEstimatorRun2 Fall17IsoV1
and I see that the occasional discrepancies are coming from ```ele_conversionVertexFitProbability```.
This should help with further debugging.

Sam-Harper:
Okay finally had time to have a look. Fixed it. Problem is not the updater or anything like this. Its due to the complete mess that the conversion fit probability variable. Bugs which are present in the 94X, 102X, 106X versions are not present in the 80X version, thus emulating them fails when running on 80X. 

Expect an update shortly. 

Btw on one hand, I'm regretting ever deciding to do this but on balance exposing this complete mess was probably worth it. Maybe.

Sam-Harper:
Okay here is the source of the problem:
https://github.com/cms-sw/cmssw/pull/19964

This PR added GsfTrack relinking to ReducedEGProducer, which means the Electrons GsfTrack is now a different collection to the conversion and therefore can no longer match via  the GsfTrack. 

So not present in 80X which is why 80X allows GsfTrack based matching but present for 94X+.  Any mini-AOD produced in 94X+ will have this bug and any produced before that will not. 

Fix is a modifier which turns the bugged function on / off.  Technically this is not necessary as using the same function which attempts to do a GsfTrack match with all three years is fine because it'll properly fail in 2017/2018, however that would be highly confusing so I'll make it explicit.

slava77:
@cmsbuild please test

slava77:
@cmsbuild please test workflow 136.7722,136.7952,136.8521,1325.6,1325.7,1325.8,1325.9,1329.1

Sam-Harper:
So I have patched it to work for 80X now.  There was an oddity which I've commented on, its a c++ question which I think Slava will know.

Anyway there still remains a potential problem for the photons for Spring16. For this MVA, it remakes the pf isolations from miniAOD, something which causes differences between this and AOD calculated values due to rounding issues.

While this something I have changed for new samples (ie now when a sample is produced, it uses the AOD to do this as it saves the values at AOD time sidestepping this issue for the future) and I'm okay with that, I'm somewhat reluctant to have this change simply when reading data from an old release in an analysis. 

However given its so small and really is just a rounding issue, I think I'm fine with it. Fixing it would be could complicated though.

RecoEgamma/EgammaTools/plugins/EG9X105XObjectUpdateModifier.cc
```diff
@@ -122,8 +131,12 @@ void EG9X105XObjectUpdateModifier::modifyObject(reco::GsfElectron& ele)const
     throw cms::Exception("LogicError") <<" in EG9X105ObjectUpdateModifier, line "<<__LINE__<<" electron "<<ele.et()<<" "<<ele.eta()<<" "<<ele.superCluster()->seed()->seed().rawId()<<" failed to match to the electrons the key map was keyed to, check the map collection is correct";
   }
   reco::GsfElectron::ConversionRejection convRejVars = ele.conversionRejectionVariables();
-  //its rather important to use the core function here to get the org trk ref
-  convRejVars.vtxFitProb = ConversionTools::getVtxFitProb(ConversionTools::matchedConversion(ele.core()->ctfTrack(),*conversions_.handle(),beamspot_.handle()->position(),2.0,1e-6,0));
+  if(allowGsfTrkMatchForConvs_){
+    convRejVars.vtxFitProb = ConversionTools::getVtxFitProb(ConversionTools::matchedConversion(ele.core(),*conversions_.handle(),beamspot_.handle()->position()));
```
Sam-Harper: So @slava77 , very rarely I get different results for 
`ConversionTools::matchedConversion(ele.core(),*conversions_.handle(),beamspot_.handle()->position())` and 
`ConversionTools::matchedConversion(ele.core(),*conversions_.handle(),beamspot_.handle()->position(), 2.0,1e-6,0)); ` where 2.0,1e-6,0 are the default parameters of that function. I tried passing them in as 2.0f, 1e-6f, 0 (incase the 1e-6 had some silly floating point rounding) but it didnt change anything.

Just curious why?

slava77:
Related to the "bug" in 94X+: is this really a bug or is it a feature?
In a sense that should the fix be provided in the origin or is it the downstream code that should adapt.

RecoEgamma/EgammaTools/plugins/EG9X105XObjectUpdateModifier.cc
slava77: The method is 
```c++
    static reco::Conversion const*         matchedConversion(const reco::GsfElectronCore &eleCore,
                                                  const reco::ConversionCollection &convCol, const math::XYZPoint &beamspot, bool allowCkfMatch=true, float lxyMin=2.0, float probMin=1e-6, unsigned int nHitsBeforeVtxMax=0);
```
Do you really mean passing ```2.0``` for the ```allowCkfMatch```?

Sam-Harper: urgh. thats what  I was missing. Fooled by the very similar function below which doesnt have that bool (and has different default parameters so I have to set them to be the same)

Thanks

Sam-Harper:
> Related to the "bug" in 94X+: is this really a bug or is it a feature?
> In a sense that should the fix be provided in the origin or is it the downstream code that should adapt.

Hmm "unintended consquence" then. Still something has to be done about it and not entirely sure what.  Probably some relinking  inside ReducedEGProducer.  To be honest, ConversionTools was doing this in a fragile way. 

However in 11X, this will be made significantly easier with the adoption of the unique IDs which can solve this once and for all.

Sam-Harper:
I changed my mind about the photons, sorry. I decided that for something which is only used for nano, its better that IDs in nano for samples done 80X /94X/102X in 106X gives the same as 102X. Now they are identical.

slava77:
there are merge conflicts here now, after a recent merge of #26703

Sam-Harper:
rebased now , should all work, does locally at least

slava77:
@cmsbuild please test workflow 136.7722,136.7952,136.8521,1325.6,1325.7,1325.8,1325.9,1329.1

slava77:
code-checks

trying manually.
The last attempt by the bot already took more than an hour.

slava77:
> I changed my mind about the photons, sorry. I decided that for something which is only used for nano, its better that IDs in nano for samples done 80X /94X/102X in 106X gives the same as 102X. Now they are identical.

I'm not sure what are the implications of this statement.

I see that there is a remaining discrepancy in the 2016 setup in mvaID values for photons. The origin of the difference is in ```CITK_isoPhoCorrMax2p5``` and ```CITK_isoPhotons```. Both of these are coming from the same input in RecoEgamma/PhotonIdentification/data/PhotonMVAEstimatorRun2VariablesSpring16.txt 
```diff
- CITK_isoPhotons           egmPhotonIsolation:gamma-DR030-                                     None None
+ CITK_isoPhotons           photonIso                                                           None None
- CITK_isoPhoCorrMax2p5     egmPhotonIsolation:gamma-DR030-                                     None None
+ CITK_isoPhoCorrMax2p5     photonIso                                                           None None
```
Here the ```photonIso``` is computed by ```photonIDValueMapProducer = cms.EDProducer("PhotonIDValueMapProducer"``` (via ```EG9X105XObjectUpdateModifier```), while the past value came from ```egmPhotonIsolation = cms.EDProducer( "CITKPFIsolationSumProducer"```.
Notably, both producers are running in the same job, which apparently means that different parts of the EGM ID are fed by only almost identical but different code to compute the photon iso (in wf 1329.1 , ZEE, there are 5 differences in a 4K events test; in all cases the isolation went up).

@Sam-Harper 
Please clarify if your comment (quoted at the top of the message) is related to this change in isolation for photons as well.
Commits since 502a2b53630ff1a11ae3a4f4f12005a907b4f61b (post-rebase: https://github.com/cms-sw/cmssw/compare/a271f5be8de066c5e58c2543913962a91affff2e...cffbf72ca1f286bfd167d21d01a50a1fe15388d3) did not clearly touch the photonIso.

Sam-Harper:
Ah yes the photon isolation. That I cant do anything about in a nice way. CITK  and non-CITK photon isolation rarely (about the rate you mention) can differ. Why this is the case is bit of a mystery to me, they shouldnt differ.  There may again also be packed candidate vs pf candidate rounding errors but this should be less of an issue. 

This is all part of the rotten mess that is photon isolation used in photon IDs, there are many inconsistencies. Often the variable is between do methods is close enough that you wouldnt notice any difference in efficiencies but different events will move in and out of the selection if you do change them. 

However I do not know of a way to fix this cleanly in this case other than defining a new name for Spring16 Mvas to indicate that it will change.  

It was fine for the charged hadron isolation, Spring16 MVAs only used that variable so I could feed it back into the photon for the purposes of nano. However photon isolation is used by the Fall17 MVAs and thus I can cant fix both of them at the same time. Unless I set the Spring16 MVA back to reading in the variables from ID value map producer which is exactly what I was trying to avoid. 

Thinking about it more, I could add another file for the Spring16 MVA for the purposes of nano where the MVA producer would be set to read that in nano jobs. Hmm, I like that idea. 

Okay thats what I'll  do. I had though to just ignore the issue as its small but now I'll define a new variable defination file for Spring16_Pre106X which means they will stay exactly the same. 

In 106X produced samples, the defination will be different but the effect will be so small that its washed out and not noticable.

Sam-Harper:
nope, what I wanted is infact extremely difficult.  The variables need to be in RecoEgamma/EgammaTools/src/MVAVariableHelper.cc   and are hard coded. I would have to design that class to take a config file as it accessing the photon isolation needs to be optional. 

I'm going to spend 10mins doing a userFloat solution. If that fails, well then we live with this.

Sam-Harper:
Meh, okay I could do that but rabbit hole as I need to adjust the parsers to deal with pat::Photons and have fun with dynamic casting. It could be done but it'll be nasty.  I dont think we care enough to do that?

slava77:
> Meh, okay I could do that but rabbit hole as I need to adjust the parsers to deal with pat::Photons and have fun with dynamic casting. It could be done but it'll be nasty. I dont think we care enough to do that?

OK, I'm going to leave this to @peruzzim and @fgolf to decide.

I'm fine from the reco side because it looks like the issue is understood and is not affecting what was done in the reco/miniAOD implementation.

peruzzim:
I would like to discuss with Sam in person tomorrow, I think it will be more effective - let's put this discussion on hold until then, please

slava77:
> I would like to discuss with Sam in person tomorrow, I think it will be more effective - let's put this discussion on hold until then, please

sure, your signature will be needed for this PR anyways.

slava77:
+1

for https://github.com/cms-sw/cmssw/pull/26486	cffbf72
- jenkins tests pass and comparisons with the baseline show differences in expected places:
    - new IDs are added in miniAOD photons and electrons
- tests with more events (40K) on nanoAOD workflows running on existing miniAODs from 80X/94X/102X show differences only in rare cases in 80X related to discrepancies in the way how photonIso is filled in the mvaID . This feature is understood and perhaps needs a fix in nanoAOD, but is perhaps small enough (O(1e-3)) that it's OK.

peruzzim:
+xpog

I tested this PR in the nanoAOD integration area, detailed outcome here https://github.com/cms-nanoAOD/cmssw/pull/363#issuecomment-492674482 confirming what expected

slava77:
in the meantime, this now became a 11_0_X PR.
We need a backport to 10_6_X.
I'm guessing that the same topic branch can be used and the new PR can be made almost trivially.

fabiocos:
@Sam-Harper @slava77 indeed this is one of the PRs that we need to port back to 10_6_X

fabiocos:
+1

fabiocos:
merge

santocch:
+1

Issue #26406, opened 2019-04-09
Save ECAL/HCAL energy fractions in MiniAOD for all PF charged hadrons


steggema:
@peruzzim @rmanzoni @mbluj @slava77 @fgolf 

Following a discussion in today's XPOG meeting, and based on the new possibility to rerun tau reconstruction on MiniAOD input, we propose to save ECAL and HCAL energy fractions for all PF charged hadrons in MiniAOD, which will enable us to have improved e->tauh and muon->tauh rejection for taus reconstructed from MiniAOD. This would be great to have for the MiniAODs for the upcoming ultra-legacy productions already.

Code-wise, it will be straightforward to implement by simply allowing the following condition to pass for all PF charged hadrons (pdgId ±211)

https://github.com/cms-sw/cmssw/blob/e50ff5c684d63017842fc9685e97aeb2e059cbe6/PhysicsTools/PatAlgos/plugins/PATPackedCandidateProducer.cc#L317

How can we check the impact on the MiniAOD event size? Is there a standard procedure?

related to #22594

peruzzim:
> How can we check the impact on the MiniAOD event size? Is there a standard procedure?

I would suggest you first try to run a standard MINIAOD production (take e.g. as a reference one of the ongoing campaigns) on 10000 TTBar events (running on too few events makes the estimate not very reliable, also because of compression).

Then, the standard testing workflows will also test the size automatically.

steggema:
After additional input from @mbluj and discussion with @rmanzoni, we realised that the implementation is not as straightforward as originally thought since we need to save "calibrated" energy fractions and not raw energy fractions (the latter are saved for JetMET's charged hadron calibration workflow), i.e. we need to add data members to make sure we save the fractions since the fractions JetMET saves for isolated charged hadrons are incompatible with what we need, but we need this information for all charged PF candidates including those charged hadrons.

I have prepared a branch that has the needed changes

https://github.com/cms-sw/cmssw/compare/master...steggema:CMSSW_10_6_X_PackedCandidateMoreCaloFractions?expand=1

On an Autumn 18 TTV file with 10000 events, this increases by some margin the size of MiniAOD:

Original:
patPackedCandidates_packedPFCandidates__PAT. 141015 20723.3

New:
patPackedCandidates_packedPFCandidates__PAT. 145227 21915.3

which translates to a 6% increase of the compressed size of this branch and a < 2% increase of the size of the MiniAOD output file.

[...]

PR #26506, opened 2019-04-23
Add more calorimeter energy fractions to PackedCandidates

steggema:
#### PR description:

This PR adds two new data members to PackedCandidates to be able to save two calorimeter energy fractions (one related to ECAL over HCAL energy, the other the ratio of calorimetric energy to overall candidate energy so that absolute HCAL energy fractions as well as track-ECAL energy ratios can be calculated) to charged PF candidates in MiniAOD. The additional information is the main missing information to be able to run algorithms to reject electrons and/or muons misreconstructed as taus on taus reconstructed from MiniAOD information only (see #22594).

The PR is expected to increase the size of the MiniAOD output by 1.1%, as measured on a 10000-event TTW file from the Autumn 18 production. The size of the pat::PackedCandidates branch is expected to increase from

patPackedCandidates_packedPFCandidates__PAT. 141015 20723.3

to

patPackedCandidates_packedPFCandidates__PAT. 145227 21500.8

The proposed implementation adds two new data members so that we end up with two data members that hold "raw" information and two that hold regular calibrated information. The reason is that JetMET currently also uses the regular energy fraction data member to save raw energy fractions for isolated charged hadrons so that they can run the charged hadron calibration on MiniAOD events. I found no traces of usage in CMSSW (except for b-tagging, who use the energy fractions saved for neutrals as was the original purpose of these members) and hence propose to move the JetMET content to the raw data mebers so we can save the new information in a logically consistent way in the existing data members (@ahinzmann @zdemirag this may affect you).

@mverzett @kskovpen As far as I understand the BTV code, the DNN-based b-taggers (who seem to be the only customers in CMSSW) only use the information saved for neutrals, which are unchanged.

This is to address #26406 (see also the discussion therein). @peruzzim @rmanzoni @mbluj @slava77 @fgolf @swozniewski 

#### PR validation:

Basic PR tests have been performed.

N/A


slava77:
@cmsbuild please test

ahinzmann:
Just to make sure I read the description and code correctly. This will keep saving for all isolated charged hadrons, the raw energy fractions, just under a different (+better) name, right?

slava77:
assign xpog
to sign off on the size increase and perhaps double-check the code changes as well

steggema:
> Just to make sure I read the description and code correctly. This will keep saving for all isolated charged hadrons, the raw energy fractions, just under a different (+better) name, right?

Yes, this is the intention.

slava77:
@cmsbuild please test

the last set of tests failed in root io, perhaps the errors were transient.

slava77:
```
Unable to open file  /eos/cms/store/user/cmsbuild/store/data/... .root; Network is unreachable
```

looks like there are some problems with the infrastructure.
I'd expect that most files for the jenkins tests are readable at /eos/cms/store/user/cmsbuild/store/data/

peruzzim:
> The proposed implementation adds two new data members so that we end up with two data members that hold "raw" information and two that hold regular calibrated information. The reason is that JetMET currently also uses the regular energy fraction data member to save raw energy fractions for isolated charged hadrons so that they can run the charged hadron calibration on MiniAOD events. I found no traces of usage in CMSSW (except for b-tagging, who use the energy fractions saved for neutrals as was the original purpose of these members) and hence propose to move the JetMET content to the raw data mebers so we can save the new information in a logically consistent way in the existing data members (@ahinzmann @zdemirag this may affect you).

@steggema I am not fully convinced by this naming change.
This kind of low level information (but still in MINIAOD) can be used in quite hidden places in legacy calibration, analysis codes etc. that not necessarily are in the upstream CMSSW repository, nor very actively maintained, and for sure not necessarily under JME control.
I agree that it would be better to fix the naming, but do we really want to risk breaking code for this reason?

steggema:
> I agree that it would be better to fix the naming, but do we really want to risk breaking code for this reason?

@peruzzim I'm glad we agree that fixing the naming makes sense. As soon as we add the information for all charged particles (with pT > 0.5 GeV), I'm sure people developing taggers will add that information also for charged particles, and I would regard it as highly inconsistent if that information meant something different for charged and neutral. With that said, there may indeed be a small risk that this change affects hypothetical code that is not maintained by JetMET and relies on that information. The only solution I have to that is to rename the data members so that all code using these variables breaks. Let me know if this is your preferred solution (I find it a bit cumbersome for an, in my opinion, fairly theoretical risk), or if you have any other idea.

peruzzim:
> @peruzzim I'm glad we agree that fixing the naming makes sense. As soon as we add the information for all charged particles (with pT > 0.5 GeV), I'm sure people developing taggers will add that information also for charged particles, and I would regard it as highly inconsistent if that information meant something different for charged and neutral. With that said, there may indeed be a small risk that this change affects hypothetical code that is not maintained by JetMET and relies on that information. The only solution I have to that is to rename the data members so that all code using these variables breaks. Let me know if this is your preferred solution (I find it a bit cumbersome for an, in my opinion, fairly theoretical risk), or if you have any other idea.

Well, yes, at this point I would prefer that solution: rename variables in a way that forces people to adapt to the new scheme.
@slava77 do you have suggestions about keeping the old accessor functions, and/or using IO rules?

steggema:
> do you have suggestions about keeping the old accessor functions, and/or using IO rules?

@peruzzim @slava77 

I'm not sure I understand the comment so I try to clarify my understanding. I see two solutions that address Marco's concern:
- Rename both member variables and accessors/setters (requires writing I/O read rules and changing code downstream, e.g. DeepJet)
- Rename only accessors and possibly setters (simpler to implement but may break a CMSSW rule that states that accessor/setter names must correspond to the variable names; still requires changing code downstream)

ahinzmann:
Just my opinion from the JME point of view. We added these variables to be able to do PF hadron calibration. In practice the last round of calibrations was still done with RECO, thus we'll not be affected by the renaming and can adapt during the transitioning MiniAOD for UL.
Let's hear from @gouskos @mverzett directly if the renaming would affect the DeepJet/DeepBoosted taggers?

DataFormats/PatCandidates/src/classes_def_objects.xml
```diff
@@ -334,7 +334,8 @@
   <class name="pat::PackedCandidate::PackedCovariance" ClassVersion="3">
    <version ClassVersion="3" checksum="3320406063"/>
   </class>
-  <class name="pat::PackedCandidate" ClassVersion="33">
+  <class name="pat::PackedCandidate" ClassVersion="34">
+   <version ClassVersion="34" checksum="1793021019"/>
```
slava77: Based on the discussion in this PR and past available implementations of the PATPackedCandidateProducer,
please consider to add an ioread rule ``` if(rawCaloFraction_ != 0) 	rawHcalFraction_ = hcalFraction_ ;```.
This way rawHcalFraction() will return the same value for old and new data.
This is more of a magic for support of different ways how values were filled rather than something specific to the data format change.

DataFormats/PatCandidates/interface/PackedCandidate.h
```diff
@@ -641,8 +654,13 @@ namespace pat {
     // for the neutral fractions
     void setRawCaloFraction(float p);                      /// Set the raw ECAL+HCAL energy over candidate energy for isolated charged hadrons
     float rawCaloFraction() const { return (rawCaloFraction_/100.); }    /// Raw ECAL+HCAL energy over candidate energy for isolated charged hadrons
-    void setHcalFraction(float p);                      /// Set the fraction of Hcal needed for HF and neutral hadrons and isolated charged hadrons
-    float hcalFraction() const { return (hcalFraction_/100.); }    /// Fraction of Ecal and Hcal for HF and neutral hadrons and isolated charged hadrons
+    void setRawHcalFraction(float p);                      /// Set the fraction of Hcal needed isolated charged hadrons
+    float rawHcalFraction() const { return (rawHcalFraction_/100.); }    /// Fraction of Hcal for isolated charged hadrons
+    void setCaloFraction(float p);                      /// Set the fraction of ECAL+HCAL energy over candidate energy, for charged particles
+    float caloFraction() const { return (caloFraction_/100.); }    /// Fraction of ECAL+HCAL energy over candidate energy, for charged particles
```
slava77: I would drop ```, for charged particles``` or replace with "when filled".

DataFormats/PatCandidates/src/PackedCandidate.cc
```diff
@@ -369,6 +369,14 @@ void pat::PackedCandidate::setRawCaloFraction(float p) {
     rawCaloFraction_ = 100*p;
 }
 
+void pat::PackedCandidate::setRawHcalFraction(float p) {
+  rawHcalFraction_ = 100*p;
+}
+
+void pat::PackedCandidate::setCaloFraction(float p) {
+  caloFraction_ = 100*p;
+}
```
slava77: I was looking at some events in 136.876 and see for PFCandidates_particleFlow some cases with  "(ecalEnergy+hcalEnergy)/energy" above 1.28. This will be truncated. 
Do you want to handle this separately to mark an overflow?
Note that there are negative values as well, but they apparently do not go too low.

PhysicsTools/PatAlgos/plugins/PATPackedCandidateProducer.cc
```diff
@@ -314,13 +314,23 @@ void pat::PATPackedCandidateProducer::produce(edm::StreamID, edm::Event& iEvent,
 
 	if(abs(cand.pdgId()) == 1 || abs(cand.pdgId()) == 130) {
 	  outPtrP->back().setHcalFraction(cand.hcalEnergy()/(cand.ecalEnergy()+cand.hcalEnergy()));
-        } else if(isIsolatedChargedHadron) {
-          outPtrP->back().setRawCaloFraction((cand.rawEcalEnergy()+cand.rawHcalEnergy())/cand.energy());
-          outPtrP->back().setHcalFraction(cand.rawHcalEnergy()/(cand.rawEcalEnergy()+cand.rawHcalEnergy()));
-	} else {
+        } else if (cand.charge() && cand.pt() > 0.5) {
+          outPtrP->back().setHcalFraction(cand.hcalEnergy()/(cand.ecalEnergy()+cand.hcalEnergy()));
+          outPtrP->back().setCaloFraction((cand.hcalEnergy() + cand.ecalEnergy())/cand.energy());
```
slava77: Based on 200 events in wf 136.876 made in 1060pre3
ecalEnergy can apparently be negative and the fraction may be ill defined due to truncation.
This seems to be a rather common property of charged hadrons (not sure about other charged).
The fractions will be truncated at +/-1.28.
Please check/consider the cases where it may matter.

slava77 (commented):
I think that the discussion about the "backward compatibility" is not strictly relevant to the PackedCandidate class by itself; it is rather related to the choices of  the producer to write one value or another into the available data members.
Knowledge of the past version producer behavior can be used as a magic info to fill the new rawHcalFraction_ member when reading the old class version. This should apparently resolve the conflict between the meaning of content saved in hcalFraction_ for the needs of the isolated CHs.

DataFormats/PatCandidates/src/PackedCandidate.cc
steggema: Looking at the distribution over recent simulated files, and discussing with @rmanzoni, I moved to unsigned int, just like for the raw calo fraction. This better covers the range of values encountered. The fraction of events with negative values is small, and they are also not interesting from the point of view of rejecting electrons or muons misidentified as taus, nor for other purposes we could imagine (like potential inclusion in other future jet taggers).

slava77:
@cmsbuild please test

DataFormats/PatCandidates/src/classes_def_objects.xml
steggema: Implemented as follows:
@peruzzim suggested to fill the raw HCAL fraction only for charged particles (since this is the subset of candidates for which the raw HCAL Fraction has been saved into the hcalFraction_ member variable), and he also suggested to set the HCAL fraction to 0 for charged particles such that the new interface is fully consistent

PhysicsTools/PatAlgos/plugins/PATPackedCandidateProducer.cc
steggema: See the other comment.

DataFormats/PatCandidates/interface/PackedCandidate.h
steggema: Done.

slava77:
@steggema 
there is a merge conflict here with  PhysicsTools/PatAlgos/plugins/PATPackedCandidateProducer.cc 
Please check and update this PR.
Thank you.

slava77:
@cmsbuild please test

slava77:
> @mverzett @kskovpen As far as I understand the BTV code, the DNN-based b-taggers (who seem to be the only customers in CMSSW) only use the information saved for neutrals, which are unchanged.

This doesn't appear to be correct:
I do not see any neutral ID requirement in https://github.com/cms-sw/cmssw/blob/e6ee4a8d6a569ff325d6cc0e15b83e95a562e167/RecoBTag/FeatureTools/plugins/DeepBoostedJetTagInfoProducer.cc#L308-L314

The nanoAOD (and miniAOD) comparisons are showing changes in the deepAK8 jet tags

Here is a plot from the ttbar MC nanoAOD workflow 1325.7
wf1325 7_deepTag_WvsQCD
It's not obvious if the code here was relying on the non-zero value of hcalFraction mostly for isolated CH or if the discrimination was driven by NH and we are "safe" to keep the changes as is (and absorb that this PR will return hcalFraction 0 for CH)
@gouskos @hqucms 

OTOH, there is also a change in deepFlavour
wf1325 7_btagDeepFlavB
From at least a quick glance at the code, this should not happen because the hcalFraction was supposedly used only for the neutrals, which should be unchanged.

slava77:
IIUC, there is a way to preserve the old code behavior for what's available in CMSSW when reading the hcalFraction.
This better be done in this PR.

My initial proposal did not include a reset of the old version value of the ```hcalFraction_```. This effectively kept the algorithmic choices made in the PackedCandidateProducer separated from the data format itself.

If there is a strong preference to reset the ```hcalFraction_``` in the old value, then the data format is treated more coupled to the way how it can be filled. In this case it seems to me that it's better to cleanly break the interface, introduce a new naming scheme e.g. ```hcalCaloFraction``` and ```hcalRawCaloFraction``` to be also clear to what it refers to. The current logic of the proposed ioread rule can then stay because it keeps track of the decisions made in the producer used for the previous version.

Either way, the old behavior of the code reading the old data better be preserved.

hqucms:
Just a clarification regarding DeepAK8: the `hcalFraction` variable is used for both charged and neutrals, as in the code pointed by Slava. I think it would be good to preserve the old behavior of the `hcalFraction()` accessor somehow so that we don't have unexpected behavior in the deep taggers.

slava77:
On 5/1/19 5:08 AM, Huilin Qu wrote:
> Just a clarification regarding DeepAK8: the |hcalFraction| variable is
> used for both charged and neutrals, as in the code pointed by Slava. I
> think it would be good to preserve the old behavior of the
> |hcalFraction()| accessor somehow so that we don't have unexpected
> behavior in the deep taggers.
> 
> —

Considering the changes in the proposed version of the packed cand producer,
some changes in the code using hcalFraction in the baseline to preserve
that baseline behavior will be needed. I'm thinking of a mix of checks
for NH type or isolated CH (there is a flag for it in the packedCandidate)

steggema:
> Just a clarification regarding DeepAK8: the hcalFraction variable is used for both charged and neutrals, as in the code pointed by Slava. I think it would be good to preserve the old behavior of the hcalFraction() accessor somehow so that we don't have unexpected behavior in the deep taggers.

Out of curiosity: For RECO, you get the hcalFraction as follows:
https://github.com/cms-sw/cmssw/blob/e6ee4a8d6a569ff325d6cc0e15b83e95a562e167/RecoBTag/FeatureTools/plugins/DeepBoostedJetTagInfoProducer.cc#L343
I assumed that this implied that you would not practically use it for charged particles since the variable is only saved for a fraction of charged particles, and hence
https://github.com/cms-sw/cmssw/blob/e6ee4a8d6a569ff325d6cc0e15b83e95a562e167/RecoBTag/FeatureTools/plugins/DeepBoostedJetTagInfoProducer.cc#L313
would give you a different answer for most charged particles. Maybe my assumption was wrong that you expect the same output for RECO and MINIAOD? I think it would be wise to try to use only calibrated energy fractions for any DeepX training in the future but of course I don't disagree with keeping current behaviour when reading current samples.

steggema:
> Considering the changes in the proposed version of the packed cand producer,
> some changes in the code using hcalFraction in the baseline to preserve
> that baseline behavior will be needed. I'm thinking of a mix of checks
> for NH type or isolated CH (there is a flag for it in the packedCandidate)

Just to make sure I understand properly: You would want me to change the BTV code to use the uncalibrated HCAL energy fraction explicitly for isolated charged hadrons, and then it would be up to the maintainers of that code to change it in the future if they follow the other arguments made in this thread? And would this supersede your previous comments:

> My initial proposal did not include a reset of the old version value of the hcalFraction_. This effectively kept the algorithmic choices made in the PackedCandidateProducer separated from the data format itself.

I made this change on explicit request of XPOG @peruzzim so I'll let him comment. Maybe this is obsolete if it's good enough to change the BTV code.

> If there is a strong preference to reset the hcalFraction_ in the old value, then the data format is treated more coupled to the way how it can be filled. In this case it seems to me that it's better to cleanly break the interface, introduce a new naming scheme e.g. hcalCaloFraction and hcalRawCaloFraction to be also clear to what it refers to. The current logic of the proposed ioread rule can then stay because it keeps track of the decisions made in the producer used for the previous version.

slava77:
On 5/2/19 1:23 AM, Jan Steggemann wrote:
> Just to make sure I understand properly: You would want me to change the
> BTV code to use the uncalibrated HCAL energy fraction explicitly for
> isolated charged hadrons, and then it would be up to the maintainers of
> that code to change it in the future if they follow the other arguments
> made in this thread? And would this supersede your previous comments:
> 

Yes.

slava77:
@cmsbuild please test

slava77:
> DQMHistoTests: Total failures: 9



this looks familiar from the last iteration, with a change in the nanoAOD workflow 1325.7
Screen Shot 2019-05-02 at 12 07 39 PM
https://cmssdt.cern.ch/SDT/jenkins-artifacts/baseLineComparisons/CMSSW_10_6_X_2019-05-01-2300+26506/31440/1325.7_TTbar_13_94Xv2NanoAODINPUT+TTbar_13_94Xv2NanoAODINPUT+NANOEDMMC2017_94XMiniAODv2+HARVESTNANOAODMC2017_94XMiniAODv2/Physics__NanoAODDQM_Jet_Prompt_btagDeepFlavB.png

Wasn't this difference supposed to go away?

steggema:
I seem to have found the issue: The ioread rule 
https://github.com/cms-sw/cmssw/pull/26506/files#diff-dc6b5dffe7f3674ba5ac7aeaa2f6d9d6R423
is apparently not executed when I try to read the pdgId_ field and try to write to hcalFraction_ (it's OK if I either don't read from pdgId_ or only write to rawHcalFraction_). It seems to me that one can't read a subsequent field when restoring a previous field without also writing that field, so the solution seems to be to also add pdgId_ as a target.

As an additional remark, it may even be more consistent to use the isIsolatedChargedHadron_ flag in the ioread rule, which is affected by the same issue. I'll test this version and push if it works as expected in my local test.

slava77:
@cmsbuild please test

slava77:
> -code-checks
> 
> Logs: https://cmssdt.cern.ch/SDT/code-checks/cms-sw-PR-26506/9548
> 
>     * This PR adds an extra 68KB to repository
> 
> 
> 
>     * **code-format**:
>       https://cmssdt.cern.ch/SDT/code-checks/cms-sw-PR-26506/9548/code-format.patch
>       e.g. `curl https://cmssdt.cern.ch/SDT/code-checks/cms-sw-PR-26506/9548/code-format.patch | patch -p1`
>       You can also run `scram build code-format` to apply code format directly

didn't all of it already get realigned?
@smuzaffar @Dr15Jones 
do we now have daily changes with evolving formatting rules?

smuzaffar:
@slava77 , formatting rules are fixed but looks like clang-format itself it not happy with its suggestions. One needs to have multiple passes for clang-format to get a frozen format :-(

Please apply the latest patch https://cmssdt.cern.ch/SDT/code-checks/cms-sw-PR-26506/9548/code-format.patch which is based on new build rules where scram does a multipass of clang-format.

perrotta:
@cmsbuild please test

RecoBTag/FeatureTools/plugins/DeepBoostedJetTagInfoProducer.cc
```diff
@@ -245,122 +217,159 @@ void DeepBoostedJetTagInfoProducer::produce(edm::Event& iEvent, const edm::Event
...
-    if (packed_cand){
+    if (packed_cand) {
       if (!has_puppi_weighted_daughters_) {
         puppiP4 *= puppi_wgt_cache.at(cand.key());
       }
 
-      fts.fill("pfcand_hcalFrac", packed_cand->hcalFraction());
+      fts.fill("pfcand_hcalFrac", (packed_cand->isIsolatedChargedHadron())
+                                      ? packed_cand->rawHcalFraction()
+                                      : packed_cand->hcalFraction());
```
slava77: if the goal is to preserve the old behavior which somewhat benefited from the "ID" provided by the isIsolatedChargedHadron to have hcalFraction not zero, then the change above should pass hcalFraction only for HF and neutral hadrons.

slava77:
@cmsbuild please test

DataFormats/PatCandidates/interface/PackedCandidate.h
```diff
@@ -644,92 +892,129 @@ namespace pat {
...
-    
-
-     // isolated charged hadrons
-    void setIsIsolatedChargedHadron(bool p);                      /// Set isolation (as in particle flow, i.e. at calorimeter surface rather than at PV) flat for charged hadrons
-    bool isIsolatedChargedHadron() const { return isIsolatedChargedHadron_; }    /// Flag isolation (as in particle flow, i.e. at calorimeter surface rather than at PV) flag for charged hadrons
+    void setRawCaloFraction(
+        float p);  /// Set the raw ECAL+HCAL energy over candidate energy for isolated charged hadrons
+    float rawCaloFraction() const {
+      return (rawCaloFraction_ / 100.);
+    }                                  /// Raw ECAL+HCAL energy over candidate energy for isolated charged hadrons
```
slava77: @fabiocos @davidlange6 
this is a case where the post-clang-format result looks worse than the original.
The comment should be closer or on the same line as the method definition, not at the end of the scope

DataFormats/PatCandidates/src/PackedCandidate.cc
```diff
@@ -11,442 +11,452 @@ CovarianceParameterization pat::PackedCandidate::covarianceParameterization_;
...
-      delete vertex_.exchange(nullptr);
-      unpackVtx();
-    }
+  reco::VertexRef pvRef = vertexRef();
+  Point pv = pvRef.isNonnull() ? pvRef->position() : Point();
+  float dxPV = vertex_.load()->X() - pv.X(), dyPV = vertex_.load()->Y() - pv.Y();  //, rPV = std::hypot(dxPV, dyPV);
+  float s = std::sin(float(p4_.load()->Phi()) + dphi_),
+        c = std::cos(float(
+            p4_.load()->Phi() +
+            dphi_));  // not the fastest option, but we're in reduced precision already, so let's avoid more roundoffs
```
slava77: @davidlange6 @fabiocos 
are there some options in clang-format that could make this less broken up?

DataFormats/PatCandidates/src/PackedCandidate.cc
```diff
@@ -11,442 +11,452 @@ CovarianceParameterization pat::PackedCandidate::covarianceParameterization_;
...
+  dxy_ = MiniFloatConverter::float16to32(packedDxy_) / 100.;
+  dz_ = pvRef.isNonnull() ? MiniFloatConverter::float16to32(packedDz_) / 100.
+                          : int16_t(packedDz_) * 40.f / std::numeric_limits<int16_t>::max();
+  Point pv = pvRef.isNonnull() ? pvRef->position() : Point();
+  float phi = p4_.load()->Phi() + dphi_, s = std::sin(phi), c = std::cos(phi);
+  auto vertex = std::make_unique<Point>(
+      pv.X() - dxy_ * s,
+      pv.Y() + dxy_ * c,
+      pv.Z() +
+          dz_);  // for our choice of using the PCA to the PV, by definition the remaining term -(dx*cos(phi) + dy*sin(phi))*(pz/pt) is zero
```
slava77: @davidlange6 @fabiocos 
isn't there a way in clang-format to separate the comment to be on a separate line (before or after the single-line)?
The current result looks pretty bad, breaking just a few symbols to a separate line

PhysicsTools/PatAlgos/plugins/PATPackedCandidateProducer.cc
```diff
@@ -29,431 +29,451 @@
...
-          isIsolatedChargedHadron=((cand.pt()>minPtForChargedHadronProperties_)&&(chargedHadronIsolation[reco::PFCandidateRef(cands,ic)]));
-          outPtrP->back().setIsIsolatedChargedHadron(isIsolatedChargedHadron);
+        //outPtrP->back().setTrackProperties(*ctrack,tsos.curvilinearError());
+      } else {
+        if (outPtrP->back().pt() > 0.5) {
+          if (ctrack->hitPattern().numberOfValidPixelHits() > 0)
+            outPtrP->back().setTrackProperties(
+                *ctrack, covariancePackingSchemas_[2], covarianceVersion_);  //low quality, with pixels
+          else
+            outPtrP->back().setTrackProperties(
```
slava77: @davidlange6 @fabiocos 
is there a way to add braces in clang-format for the cases of multi-line break-up of an if/else?
The result looks like a bit more dangerous pattern where and else can get misplaced

PhysicsTools/PatAlgos/plugins/PATPackedCandidateProducer.cc
```diff
@@ -1,150 +1,186 @@
...
-#include "DataFormats/RecoCandidate/interface/RecoChargedCandidate.h"
 
 /*#include "TrackingTools/TrajectoryState/interface/TrajectoryStateTransform.h"
-#include "TrackingTools/GeomPropagators/interface/AnalyticalImpactPointExtrapolator.h"
 #include "MagneticField/Engine/interface/MagneticField.h"
 #include "MagneticField/Records/interface/IdealMagneticFieldRecord.h"
-#include "TrackingTools/TrajectoryState/interface/TrajectoryStateOnSurface.h"
 #include "RecoVertex/VertexPrimitives/interface/ConvertToFromReco.h"
+#include
+"TrackingTools/GeomPropagators/interface/AnalyticalImpactPointExtrapolator.h"
```
slava77: @davidlange6 @fabiocos 
this is apparently breaking the commented out code

PhysicsTools/PatAlgos/plugins/PATPackedCandidateProducer.cc
```diff
@@ -1,150 +1,186 @@
...
+};
+} // namespace pat
+
+pat::PATPackedCandidateProducer::PATPackedCandidateProducer(
+    const edm::ParameterSet &iConfig)
+    : usePuppi_(
+          !iConfig.getParameter<edm::InputTag>("PuppiSrc").encode().empty() ||
+          !iConfig.getParameter<edm::InputTag>("PuppiNoLepSrc")
+               .encode()
+               .empty()),
```
slava77: @davidlange6 @fabiocos 
for this and similar cases is there a way to make clang-format not break up similarly looking lines in a different way?
it seems also unnecessary to have ```.encode().empty() ``` broken up to two pretty short lines. Was it intended in the configuration

slava77:
+1

for https://github.com/cms-sw/cmssw/pull/26506	3120396
- code changes are in line with the PR description and the follow up review. Some effort was made to preserve the old treatment of ```hcalFraction``` in the deep jet/b-tagging code.
- jenkins tests pass and comparisons with the baseline show differences only in the ```hcalFraction``` values, which are now also filled for most of the charged hadrons.
    - deep jet/b-tagging behavior is unchanged

peruzzim:
+xpog

PhysicsTools/PatAlgos/plugins/PATPackedCandidateProducer.cc
davidlange6: there are not an infinite number of options, no.  This presumably happens because of the greater number of characters in the second line. 

One can try by hand just to have a line break before .encode() on both lines (but that also may get changed by clang-format)

PhysicsTools/PatAlgos/plugins/PATPackedCandidateProducer.cc
davidlange6: no - clang-format changes whitespace only. There is a clang-tidy option to add braces for ifs/fors/whiles  that has been tried but so far doesn't work fully in cmssw (afaik).

DataFormats/PatCandidates/src/PackedCandidate.cc
davidlange6: i'm not aware of options for such fine grain control. Obviously this can be fixed by hand.

perrotta:
@davidlange6 here and in other recent PRs (see #26606 for another example) line breaking as implemented by clang mostly results in less easily readable code; moreover, it complicates the reviews quite a lot. Given my recent experiences I would simply avoid forcing all line breaks in the code checks resolution, unless some more clever setting can be found for them.

davidlange6:
I guess the review/discussions over the last year disagrees with that conclusion. But you are welcome to make a specific proposal to the core group.

fabiocos:
+1

fabiocos:
merge

santocch:
+1

PR #26265, opened 2019-03-26
RecoParticleFlow/PFProducer - merging PFBlockProducer and PFBlockAlgo

guitargeek:
#### PR description:

Part of the occasional PFProducer PRs to make the Particle Flow code easier to grasp.

* The PFBlockProducer was a very minimal wrapper around the PFBlockAlgo class. In auch a case it does not make sense to separate them, so the PFBlockAlgo was merged with the producer
* Moving already a lot of the code around, the whole new PFBlockProducer was formatted with Davids clang-format file shown in the last reco meeting (would be done anyway at some point)
* Remove preprocessor directives
* some spurious modernization of the code
* put the QuickUnion class into its own files, as I think it's general enough to deserve that (also clang-formatted)

Observations:
* after making the links between elements a first time, the links are stored in a hash table before going after the elements a second time and updating the links. That is not really efficient. There is definitely room for speedup and improvement in this producer! Does it take a significant portion of the reconstruction time?

#### PR validation:

CMSSW compiles and matrix tests pass.

PR #26519, opened 2019-04-23
RecoParticleFlow/PFProducer - some cleanup and modernization

guitargeek:
#### PR description:

After my [last slightly unpopular PR](https://github.com/cms-sw/cmssw/pull/26265) with the PFBlockAlgo I'm back with some other quick ideas on simplifying and modernizing the PF code :)

* __KDTreeLinkerTools__: structs with only public members don't need constructors and setters/getters
* __PFAlgo__: `unique_ptr` to sub-algos instead of raw pointers
* __PFCandConnector__: change some members to local variables which allows for `const` member functions and avoids smart pointers
* __PFProducer__: have the PFAlgo on the stack and use the PutToken at one place to accommodate for not using `unique_ptr` in the PFCandConnector
* __PFBlockLink__: remove that class. The information it stores became redundant after the Run2 rewrite of the BlockProducer, only the `dist` member is actually used => the class can just be replaced with a `double`
* __PFBlockAlgo__: adapt to the removal of the PFBlockLink class

Furthermore, the includes in the PFBlockAlgo/PFBlockProducer were cleaned up. For this, I used the necessary includes found by the include-what-you-use (iwyu) tool [in my old PR](https://github.com/guitargeek/cmssw/blob/c5fe14b2e2453dbafd312dca1c11534ae86d21fc/RecoParticleFlow/PFProducer/plugins/PFBlockProducer.cc) and spread them across the 4 files as appropriate.

#### PR validation:

Code compiles and matrix tests complete.


guitargeek:
Hi @slava77, I just saw you commented in my last PR which was closed (so I didn't get an email notification). Right, the includes! I should quickly clean them up. Moving the algo in the plugins directory would be a bit too much I think, because that would apply to all the algos actually. And moving only one would be a bit inconsistent...

slava77:
@cmsbuild please test

slava77:
> Reco comparison results: 4238 differences found in the comparisons

it looks like changes are everywhere.
e.g. in run 1 wf 1000.0  there are 4 more charged hadrons
https://cmssdt.cern.ch/SDT/jenkins-artifacts/baseLineComparisons/CMSSW_10_6_X_2019-04-23-1100+26519/31316/validateJR/all_OldVSNew_RunMinBias2011Awf1000p0/all_OldVSNew_RunMinBias2011Awf1000p0c_log10recoPFCandidates_particleFlow__RECO_obj_pt36.png

@guitargeek 
please investigate the origins and hopefully restore the old behavior or perhaps point to a bugfix leading to the difference.
Thank you.

RecoParticleFlow/PFProducer/interface/PFCandConnector.h
```diff
@@ -61,21 +61,20 @@ class PFCandConnector {
...
-       std::unique_ptr<reco::PFCandidateCollection> connect(std::unique_ptr<reco::PFCandidateCollection>& pfCand);
+       reco::PFCandidateCollection connect(reco::PFCandidateCollection& pfCand) const;
 
        
  
     private :
 
        /// Analyse nuclear interactions where a primary or merged track is present
-       void analyseNuclearWPrim(std::unique_ptr<reco::PFCandidateCollection>&, unsigned int);
+       void analyseNuclearWPrim(reco::PFCandidateCollection&, std::vector<bool> &, unsigned int) const;
```
guitargeek: I updated with force-push because my mistake that caused the differences was not interesting enough for it's own commit: I forgot to pass the mask vector by reference (also in the other function)...

slava77:
@cmsbuild please test

slava77:
+1

for https://github.com/cms-sw/cmssw/pull/26519	23dd640

kpedro88:
+1

PR #26526, opened 2019-04-24
Overlap Validation for Tracker Alignment

feingoldj:
@hroskes @jfan20 @VictordeHavenon @adewit @connorpa @mmusich @gritsan

#### PR description:

We have developed an Overlap Validation that is adept at detecting the Bowing, Radial, and Z-expansion Weak Modes. It uses the difference between residuals on overlapping modules to characterize the misalignment. 

Details can be found here: 
[Overlap Validation Presentation](https://indico.cern.ch/event/787556/contributions/3282549/attachments/1779819/2895001/Weak_Modes_-_Radial_Validation_.pdf)

#### PR validation:

The following is the Overlap Validation plot for a radial misalignment with various epsilon values:
image

We also implemented DMRs separated by track charge that highlights biases that cause Z-expansion in TEC. More information can be found here:
 [Z-Expansion Update](https://indico.cern.ch/event/787662/contributions/3362985/attachments/1815177/2966437/z_expansion.pdf)

PR #26533, opened 2019-04-25
Save Z->mumu plots as .root

hroskes:
Split from #26526 for better organization.  It's orthogonal to everything else there.

The use case for the root files is to have easy access to the numerical data in the plots.  This way we can extract the size of features in the plot to estimate the size of weak modes in the alignment


santocch:
+1 

- add simply the root file to the list of saved files (png and pdf files)

fabiocos:
+1

PR #26541, opened 2019-04-25
switch to tau id MVA2017v2 in AOD and MiniAOD

swozniewski:
#### PR description:

Switch from MVA2017v1 to MVA2017v2 tau ID in AOD and MiniAOD following the recommendation in [https://indico.cern.ch/event/810741/contributions/3384093/attachments/1827349/2991114/TauID_CMSweek_10042019.pdf#page=3](https://indico.cern.ch/event/810741/contributions/3384093/attachments/1827349/2991114/TauID_CMSweek_10042019.pdf#page=3)

#### PR validation:

Matrix tests passed.
Output of raw value has been compared to corresponding NanoAOD sample where MVA2017v2 is already available.

According differences in raw and WP values of IsolationMVArun2v1DBoldDMwLT are expected.

### DIFF
--- a/RecoTauTag/Configuration/python/HPSPFTaus_cff.py
+++ b/RecoTauTag/Configuration/python/HPSPFTaus_cff.py
@@ -526,7 +526,7 @@ hpsPFTauDiscriminationByIsolationMVArun2v1DBoldDMwLTraw = discriminationByIsolat
     PFTauProducer = cms.InputTag('hpsPFTauProducer'),
     Prediscriminants = requireDecayMode.clone(),
     loadMVAfromDB = cms.bool(True),
-    mvaName = cms.string("RecoTauTag_tauIdMVAIsoDBoldDMwLT2017v1"),
+    mvaName = cms.string("RecoTauTag_tauIdMVAIsoDBoldDMwLT2017v2"),
     mvaOpt = cms.string("DBoldDMwLTwGJ"),
     srcTauTransverseImpactParameters = cms.InputTag('hpsPFTauTransverseImpactParameters'),
     srcChargedIsoPtSum = cms.InputTag('hpsPFTauChargedIsoPtSum'),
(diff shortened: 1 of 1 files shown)


slava77:
@cmsbuild please test

slava77:
@swozniewski 
please add a link to some slides shown in a TAU POG (sub)group meeting to the PR description to document the changes.

steggema:
> @swozniewski
> please add a link to some slides shown in a TAU POG (sub)group meeting to the PR description to document the changes.

@slava77 
We have recommended the 2017v2 discriminator for a while now. The most recent set of slides on the recommendations is here: https://indico.cern.ch/event/810741/contributions/3384093/attachments/1827349/2991114/TauID_CMSweek_10042019.pdf#page=3

The main purpose of this PR is to have this discriminator also available by default in all MiniAOD to have the most useful baseline for all ultra-legacy analyses and possible new trainings that will be performed with the ultra-legacy samples. @rmanzoni @roger-wolf

swozniewski:
thank you @steggema ! I've added the link to the PR description.

RecoTauTag/Configuration/python/HPSPFTaus_cff.py
```diff
@@ -526,7 +526,7 @@
     PFTauProducer = cms.InputTag('hpsPFTauProducer'),
     Prediscriminants = requireDecayMode.clone(),
     loadMVAfromDB = cms.bool(True),
-    mvaName = cms.string("RecoTauTag_tauIdMVAIsoDBoldDMwLT2017v1"),
+    mvaName = cms.string("RecoTauTag_tauIdMVAIsoDBoldDMwLT2017v2"),
```
slava77: what is the rationale to keep the same name of the producer with "v1"?
Should it be changed as well to v2?

swozniewski: I had the impression that this version number is not really used because the year does not even show up, so v1 would not stand for 2017v1. But the naming history goes beyond my scope. @steggema @rmanzoni @mbluj could you comment?

steggema: @mbluj may know better, but we have two things to consider,

a) the list of input variables, which I think can be considered to be represented by the tag "MVArun2v1DBoldDMwLT" and then resolved to "mvaOpt = cms.string("DBoldDMwLTwGJ")"
b) the version of the training, which is 2017v2

So I think it may be fine to keep the version as Sebastian has it. OTOH, maybe a case can be made for the training tag (2017v2) representing everything, but that may need to lead to even larger overall changes...

mbluj: I confirm what @steggema wrote. Name of module in the python configuration represents version of MVA, i.e. list of inputs, while name of payloads stands for a given training of the MVA. 
From practical perspective, it is also useful to keep names of modules and change only names of payloads as it reduces number of modifications (in terms of no. of modified lines and files).

mbluj:
@swozniewski I has been not able to check the PR earlier, but it looks that __DBnewDMwLT__ and __DBdR03oldDMwLT__ are not migrated to the __2017v2__ training (whose are not present for __2017v1__). Could you confirm that this migration is considered, please? And sorry for the late hint.

swozniewski:
@mbluj I was not aware that there again is a training in 2017v2 in contrast to 2017v1 for these two IDs. I will add this now and create a PR to cms-tau-pog:CMSSW_10_6_X_tau-pog_MVA2017v2, where you can check it. I assume that the naming scheme is the same.

RecoTauTag/Configuration/python/HPSPFTaus_cff.py
slava77: Looking back to CMSSW_7_6_2, when this naming pattern was introduced, the previous payloads had the version of the MVA in the name (or at least it seemed like it).
Please share a pointer to a past example of a different case of ID where the payload does not contain the "vX" of the MVA in its name, just to be clear of a somewhat established practice.

slava77: > In this PR #21022 the 2017v1

yet, it was still a "v1".
I just wanted to understand if this "vX" is becoming more formalized just now.
Perhaps for consistency of future changes you can make a note in some tau POG area to have that as a reference.

slava77:
@cmsbuild please test

swozniewski:
The second commit addresses Michals comment that 2017v2 is also available for DBnewDMwLT, DBdR03oldDMwLT and updates them. The raw output has been compared to the existing values in NanoAOD on some test events.

RecoTauTag/Configuration/python/HPSPFTaus_cff.py
mbluj: I see what you mean I agree that some version bookkeeping will be useful for final users who usually know name of a producer or/and access name within PAT (and a release version), but not a name of a related payload. It is up to conveners to decide if it makes sense to change name of producer or it will be enough to provide a table in Tau POG recommendation TWiki.

slava77: OK, I can wait until Thursday to signoff on this PR (considering the holiday tomorrow).
I hope that by then the decision to keep the current naming or to change it can be made.

rmanzoni: @steggema and I agreed to keep the current naming.

The reason is not to force analysts to adjust their code now while we expect to move to recommend a newer DNN-based Tau ID sometime soon, which will demand code changes. 

With that, we’ll have a chance to start right away with a more coherent version naming.

slava77: OK, I will sign off, but some better formalized naming would be quite helpful.
Without being aware of this discussion and digging into the git history, it would appear more and more that the configuration may be buggy.

slava77:
+1

for https://github.com/cms-sw/cmssw/pull/26541	b11ce72 
- jenkins tests pass and comparisons with the baseline show changes only in tau MVA and ID variables
    - changes start from reco workflows in hpsPFTauDiscriminationByIsolationMVArun2v1DBoldDMwLTraw, hpsPFTauDiscriminationByIsolationMVArun2v1DBdR03oldDMwLTraw, and hpsPFTauDiscriminationByIsolationMVArun2v1DBnewDMwLTraw raw discriminant values
    - the differences then propagate to the tau IDs, including the miniAOD and nanoAOD outputs in cases where the corresponding tau "reco" MVA tag modules are running in the tested workflow.

The performance is changing somewhat moderately, e.g. in 250202.181 (ttbar with PU 2018 pmx setup)
wf250202 181_tau_summary

fabiocos:
@peruzzim @fgolf FYI

fabiocos:
+1

PR #26543, opened 2019-04-26
Remove stray includes from PFBlockAlgo.h and minor cleanups

guitargeek:
#### PR description:

Sorry, quick follow up on https://github.com/cms-sw/cmssw/pull/26519: there were still include files hiding at the  bottom of `PFBlockAlgo.h` which I forgot to remove. Also some minor cleaning of PFBlockAlgo and PFEgammaAlgo.

#### PR validation:

### DIFF
--- a/RecoParticleFlow/PFProducer/interface/PFBlockAlgo.h
+++ b/RecoParticleFlow/PFProducer/interface/PFBlockAlgo.h
@@ -38,13 +38,6 @@ class PFBlockAlgo {
  public:
   // the element list should **always** be a list of (smart) pointers
   typedef std::vector<std::unique_ptr<reco::PFBlockElement> > ElementList;
-  typedef std::unique_ptr<BlockElementImporterBase> ImporterPtr;
-  typedef std::unique_ptr<BlockElementLinkerBase> LinkTestPtr;  
-  typedef std::unique_ptr<KDTreeLinkerBase> KDTreePtr;
-  /// define these in *Fwd files in DataFormats/ParticleFlowReco?
-  typedef ElementList::iterator IE;
-  typedef ElementList::const_iterator IEC;  
-  typedef reco::PFBlockCollection::const_iterator IBC;
   //for skipping ranges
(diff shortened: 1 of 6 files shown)


slava77:
@cmsbuild please test

RecoParticleFlow/PFProducer/interface/PFBlockAlgo.h
```diff
@@ -76,10 +69,6 @@ class PFBlockAlgo {
   void packLinks(reco::PFBlock& block, 
 		 const std::unordered_map<std::pair<unsigned int,unsigned int>,double>& links) const; 
   
-  /// Avoid to check links when not useful
-  inline bool linkPrefilter(const reco::PFBlockElement* last,
```
perrotta: This method is removed because it is never used. In fact, it is the ```BlockElementLinkerBase::linkPrefilter(...)``` that is called instead.

However, if I am not missing anything in the logic and the 
```C++
std::vector<std::unique_ptr<BlockElementLinkerBase>> linkTests_;
````
does not specialize into any of the clasess derived from BlockElementLinkerBase, that ```BlockElementLinkerBase::linkPrefilter(...)``` always identically returns ```true```, see https://cmssdt.cern.ch/lxr/source/RecoParticleFlow/PFProducer/interface/BlockElementLinkerBase.h#0018

Could you please verify? If so, some further cleaning is possible (or some fix, if the original intention was different).

@bendavid @hatakeyamak

guitargeek: Hi Andrea, it's not so simple I think. Some of the derived linkers implement this method in a non-trivial way, and thanks to the virtual table these are actually used.

perrotta: > Hi Andrea, it's not so simple I think on first glance. Some of the derived linkers implement this method in a non-trivial way, and thanks to the virtual table these are actually used.

Thank you @guitargeek 
Indeed I was expecting something similar, but it was not straightforward to me to find how those derived linkers enter in the game

perrotta:
+1
- Cleanup implemented as planned

fabiocos:
+1

PR #26558, opened 2019-04-27
Add missing initialization to some local array in PPS ProtonReconstructionAlgorithm

perrotta:
#### PR description:

Catched by the static analyzer.

The missing initialization was implicitly avoided because there are always at least two tracks, and therefore those arrays will always get filled. By explicitely initializing them one can silence the static analyzer, though, and make the code more robust in case of future different usages.

#### PR validation:

It builds.
No changes are expected

@fabferro @jan-kaspar @vavati

### DIFF
--- a/RecoCTPPS/ProtonReconstruction/src/ProtonReconstructionAlgorithm.cc
+++ b/RecoCTPPS/ProtonReconstruction/src/ProtonReconstructionAlgorithm.cc
@@ -190,11 +190,10 @@ reco::ForwardProton ProtonReconstructionAlgorithm::reconstructFromMultiRP(const 
     th_x_init = 0.;
 
   // initial estimate of th_y and vtx_y
-  double y[2], v_y[2], L_y[2];
+  double y[2]={0}, v_y[2]={0}, L_y[2]={0};
   unsigned int y_idx = 0;
   for (const auto &track : tracks) {
-    if (y_idx >= 2)
-      continue;
+    if (y_idx >= 2) break;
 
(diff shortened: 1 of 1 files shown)


perrotta:
please test

VinInn:
don not understand.
the standard mandates default initialization. For scalar it is zero.
so the static analyzer is wrong.
https://en.cppreference.com/w/cpp/language/zero_initialization

slava77:
> don not understand.
> the standard mandates default initialization. For scalar it is zero.
> so the static analyzer is wrong.
> https://en.cppreference.com/w/cpp/language/zero_initialization
apparently the warning was about 
```c++
   double y[2], v_y[2], L_y[2];
```
these are not static and not brace initialized. So, there is no zero initialization.
Did I miss something?

forthommel:
@VinInn @perrotta @slava77 what would be the benefits of using a good old `std::array<double,2>` instead?

perrotta:
@forthommel @VinInn : my only intention was to silence out the static analyzer and protect in case of different use cases than what currently in the release: and this PR does it.
If you think there is anything wrong I will modify it; but as such it seems rather innocuous to me...

slava77:
On 4/27/19 6:28 AM, Laurent Forthomme wrote:
> what would be the benefits of using a good old |std::array<double,2>|
> instead?

I think that the array initialization is the same.
So, this part shouldn't change.

There are benefits from knowing the size at compile time.

perrotta:
+1
- Trivial code adjustment

fabiocos:
+1

PR #26559, opened 2019-04-27
PFEgammaAlgo - New FlaggedPtr class

guitargeek:
#### PR description:

One of the confusing aspects about the particle flow code are these intricate nested std:: types, which causes things in the code like `->first.second->` and you have no idea what you are dealing with anymore. It is not easy to improve this in one PR, therefore I propose just a little change to slightly improve the situation in PFEgammaAlgo.

The proposal is to replace the `std::pair<const T*,bool>` types with a new class `FlaggedPtr<T>`, which behaves just like a (smart) pointer wrapping the  `const T*` plus the possibility to set the flag. In particular, the `->` operator of that new class simplifies the `PFEgammaAlgo` code quite a bit.

I think I should not do more in one PR, but I hope this is the right way to the improve the PFEgamma code with new types. Maybe this could be even extended in the future to a `PfPtr` with an "index" (of the element in the block), a "type", and the flag. The other advantage is that if the flagging is not needed anymore, the `FlaggedPtr` can be replaced easily with a raw or smart pointer.

#### PR validation:

### DIFF
--- a/RecoParticleFlow/PFProducer/interface/CommutativePairs.h
+++ b/RecoParticleFlow/PFProducer/interface/CommutativePairs.h
@@ -1,0 +1,48 @@
+#ifndef RecoParticleFlow_PFProducer_CommutativePairs_h
+#define RecoParticleFlow_PFProducer_CommutativePairs_h
+
+#include <utility>
+#include <vector>
+
+/**
+ * Wrapper around std::vector<std::pair<T, T>> when the order of the pair elements is not relevant.
+ *
+ * @tparam T the type of data stored in the pairs
+ */
+
+template <class T>
+class CommutativePairs {
+public:
+  // Insert a new pair
+  void insert(T const& a, T const& b) { pairs_.emplace_back(a, b); }
+
+  // Check if this contains (a,b) or (b,a)
+  bool contains(T const& a, T const& b) const {
+    for (auto const& p : pairs_) {
+      if ((a == p.first && b == p.second) || (b == p.first && a == p.second)) {
+        return true;
+      }
+    }
+    return false;
+  }
+
+  // Check if this contains (a,b) or (b,a), where b is arbitrary
+  bool contains(T const& a) const {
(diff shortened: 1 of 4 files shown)


slava77:
@cmsbuild please test

guitargeek:
I had to update, as I forgot the header guards. In this commit, I also added a `CommutativePairs<T>` class, which replaces `std::vector<std::pair<T,T>>` when the ordering of the pair does not matter. This way, the `localMap` only needs to be half as large in memory.

Code checks seem to have changed, because things that have passed before don't pass anymore. Hence the last commit.

perrotta:
@cmsbuild please test

perrotta:
@hatakeyamak @bendavid @Sam-Harper

fabiocos:
please test

Sam-Harper:
just for the record, I fully support this. Jonas this also has driven me nuts in the past. Anything we can do to reduce the complexity is much appreciated here! Thank you.

slava77:
in wf 5.1
```
   [2] Calling method for module PFEGammaProducer/'particleFlowEGamma'
Exception Message:
The number of found ecal elements (2) in block is not the same as the number of ecal PF clusters reported by the PFSuperCluster itself (1)! This should not happen!
```
my guess is that the same is happening in the other wfs.

guitargeek:
Hello there! I'm back. Ok, that means I did a mistake when I casually tried to reorganize these `NotCloserToOther` functors.

I will just revert parts of the last commit and instead wrap up the story with the wrapped pointer: I realized the flagging is actually not used in 80 % of the `FlaggedPtr` instances, so I'll just replace them with raw pointers and update this PR.

slava77:
@cmsbuild please test

slava77:
+1

for https://github.com/cms-sw/cmssw/pull/26559	24ad716
- technical: refactoring of the code in PFEGammaAlgo

fabiocos:
+1

Issue #26562, opened 2019-04-29
Failure of wf 140.55 after merging PR 26485


fabiocos:
After the integration of #26485 a failure is observed in the HI test workflow 140.55 step4:

```
----- Begin Fatal Exception 29-Apr-2019 13:33:41 CEST-----------------------
An exception of category 'ProductNotFound' occurred while
   [0] Processing  Event run: 263400 lumi: 65 event: 1020254 stream: 0
   [1] Prefetching for module HTMonitor/'AK8PFHT750_TrimMass50_Softdropmonitoring'
   [2] Prefetching for module GEDGsfElectronFinalizer/'gedGsfElectrons'
   [3] Prefetching for module PFProducer/'particleFlowTmp'
   [4] Prefetching for module ElectronIdMVABased/'mvaElectrons'
   [5] Prefetching for module GsfElectronSelector/'electronsWithPresel'
   [6] Calling method for module GsfElectronEcalDrivenProducer/'ecalDrivenGsfElectrons'
Exception Message:
Principal::getByToken: Found zero products matching all criteria
Looking for type: std::vector<reco::Conversion>
Looking for module label: allConversions
Looking for productInstanceName: 

   Additional Info:
      [a] If you wish to continue processing events after a ProductNotFound exception,
add "SkipEvent = cms.untracked.vstring('ProductNotFound')" to the "options" PSet in the configuration.

----- End Fatal Exception -------------------------------------------------
```

fabiocos:
assign reconstruction

fabiocos:
@slava77 @perrotta @Sam-Harper apparently the update of the Egamma code breaks the old HI reconstruction sequence (140.55), as the required process.allConversions module is not run, why the 2018 HI sequence, that is basically the default pp one, is ok (140.56).

Sam-Harper:
looking at it

Sam-Harper:
okay the solution will be simply to have a flag whether this variable is to be filled or not and set it false for HION. Not elegant but no other better solution. Code is written, fixes workflow, testing other workflows.

perrotta:
+1
- Fixed with #26567,  workflow 140.55 now passes

PR #26567, opened 2019-04-29
Making filling conversion vtx fit prob optional

Sam-Harper:
#### PR description:
Fixes issue raised here https://github.com/cms-sw/cmssw/issues/26562

This makes filling the conversion vertex fit probability optional for GsfElectrons. HION doesnt have conversions and thus has issues here. So added a simple flag to disable it for cases like this rather than having to run the conversion workflow in HION.


slava77:
type bugfix

slava77:
@cmsbuild please test

slava77:
+1

for https://github.com/cms-sw/cmssw/pull/26567	739068e
- jenkins tests pass, including the workflow 140.55 with a problem in the baseline

fabiocos:
+1

@Sam-Harper thanks for the prompt reaction

PR #26569, opened 2019-04-29
add missing includes for HcalDepthEnergyFractions.h

slava77:
Recent header consistency checks showed that there is a problem with the DataFormats/PatCandidates/interface/HcalDepthEnergyFractions.h file

This PR should fix that.

### DIFF
--- a/DataFormats/PatCandidates/interface/HcalDepthEnergyFractions.h
+++ b/DataFormats/PatCandidates/interface/HcalDepthEnergyFractions.h
@@ -1,6 +1,9 @@
 #ifndef DataFormats_PatCandidates_HcalDepthEnergyFractions_h
 #define DataFormats_PatCandidates_HcalDepthEnergyFractions_h
 
+#include <vector>
+#include <cstdint>
+
 namespace pat{
 
   //


slava77:
@cmsbuild please test

hatakeyamak:
Hmm. Thank you. @slava77 Why didn't I catch it when I created this file? (so that I can avoid the same issue in future.)

slava77:
it looks like there was a (mis)configuration issue in the header checking setup, which skipped new files. So, this issue did not show up in the tests of #26465

slava77:
+1

for https://github.com/cms-sw/cmssw/pull/26569	a3f62c4
- technical
- jenkins tests pass
    - the header checks for HcalDepthEnergyFractions.h are now clean

hatakeyamak:
> it looks like there was a (mis)configuration issue in the header checking setup, which skipped new files. So, this issue did not show up in the tests of #26465

ok. i see. thank you for fixing this.

fabiocos:
+1

fabiocos:
merge

santocch:
+1

PR #26574, opened 2019-04-30
Remove dead assignments inside CSCRawToDigi

perrotta:
#### PR description:

A few dead assignments are reported by the static analyzer in the CSCRawToDigi package.
They are removed here.

However, they are sometimes so trivial, that I wonder whether they were left as such on purpose, for a reason that escapes my comprehension. That's why I would like to ask for the endorsement of the CSC reco contact, @ptcox , before signing this PR for reco.

#### PR validation:

It builds

### DIFF
--- a/EventFilter/CSCRawToDigi/src/CSCDCCExaminer.cc
+++ b/EventFilter/CSCRawToDigi/src/CSCDCCExaminer.cc
@@ -273,17 +273,13 @@ int32_t CSCDCCExaminer::check(const uint16_t* &buffer, int32_t length)
 
 
   /// Check for presence of data blocks inside TMB data
-  bool fTMB_Scope_Start = false;
   bool fTMB_MiniScope_Start = false;
   bool fTMB_RPC_Start = false;
   bool fTMB_BlockedCFEBs_Start = false;
 
-  bool fTMB_Scope = false;
   bool fTMB_MiniScope = false;
   bool fTMB_RPC = false;
   bool fTMB_BlockedCFEBs = false;
-
-  fTMB_Scope = fTMB_Scope && true; // WARNING in 5_0_X
 
   while( length>0 )
     {
@@ -633,12 +629,10 @@ int32_t CSCDCCExaminer::check(const uint16_t* &buffer, int32_t length)
           uniqueALCT   = true;
           uniqueTMB    = true;
 
-          fTMB_Scope_Start = false;
           fTMB_MiniScope_Start = false;
           fTMB_RPC_Start = false;
           fTMB_BlockedCFEBs_Start = false;
 
-          fTMB_Scope = false;
           fTMB_MiniScope = false;
           fTMB_RPC = false;
           fTMB_BlockedCFEBs = false;
@@ -1031,12 +1025,6 @@ int32_t CSCDCCExaminer::check(const uint16_t* &buffer, int32_t length)
           fTMB_RPC_Start = true;
         }
 
-      // Check for Scope data
-      if ( fTMB_Header && (scanbuf(buf0,4, 0x6B05)>=0) )
-        {
-          fTMB_Scope_Start = true;
-        }
-
       // Check for Mini-Scope data
       if ( fTMB_Header && (scanbuf(buf0,4, 0x6B07)>=0) )
         {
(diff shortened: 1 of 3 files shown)


EventFilter/CSCRawToDigi/src/CSCDCCExaminer.cc
```diff
@@ -273,18 +273,14 @@ int32_t CSCDCCExaminer::check(const uint16_t* &buffer, int32_t length)
...
   bool fTMB_MiniScope_Start = false;
   bool fTMB_RPC_Start = false;
   bool fTMB_BlockedCFEBs_Start = false;
 
-  bool fTMB_Scope = false;
   bool fTMB_MiniScope = false;
   bool fTMB_RPC = false;
   bool fTMB_BlockedCFEBs = false;
 
-  fTMB_Scope = fTMB_Scope && true; // WARNING in 5_0_X
```
perrotta: What was the original intention of this line?

EventFilter/CSCRawToDigi/src/CSCTMBData.cc
```diff
@@ -166,26 +166,21 @@ int CSCTMBData::UnpackTMB(const uint16_t *buf) {
...
   unsigned short int firmwareVersion=0;
   int Ntbins = 0 ;
-  int NHeaderFrames = 0; //WARNING in 5_0_X
   int NRPCtbins = 0; // =VB= number of RPC tbins  
   
   int b0cLine=0;///assumes that buf starts at the tmb data
                 ///this is not true if something is wrong in the data 
                 ///before TMB - then we skip the whole event
 
-  NHeaderFrames++; NHeaderFrames--;
```
perrotta: What was the original intention of this line?

EventFilter/CSCRawToDigi/src/CSCDDUEventData.cc
```diff
@@ -277,7 +271,6 @@ void CSCDDUEventData::unpack_data(const uint16_t *buf, CSCDCCExaminer* examiner)
 
     }
     theSizeInWords = dduBufSize+12;
-    buf=inputBuf+dduBufSize;
```
perrotta: If the intention was to modify the pointer passed to this function, then it should have been passed by reference. 
Otherwise, I agree with the static analyzer that this is a dead assignment as such, and it can be removed

perrotta:
please test

ptcox:
Hi,

Victor Barashko is the CSC Unpacker expert. I've already emailed him to 
ask his opinion on all these 'cleanups'. I am fairly sure that in the 
past he maintained a 'master' version outside CMSSW because it is also 
required for online and CSC stand-alone running at Point 5, as well as 
various commissioning level tasks. It could indeed be that many of the 
features noted here are associated with operation external to CMSSW. So 
let's wait for Victor's comments.

Regards, Tim

perrotta wrote on 4/30/19 12:46:

barvic:
@perrotta It looks like most of those old code lines were used as a way to prevent compiler warnings about currently unused variables, instead of commenting it out. Most likely those variables are for not fully implemented features, which were in plans initially, but were later considered as not really needed or important for actual system operation.

perrotta:
> @perrotta It looks like most of those old code lines were used as a way to prevent compiler warnings about currently unused variables, instead of commenting it out. Most likely those variables are for not fully implemented features, which were in plans initially, but were later considered as not really needed or important for actual system operation.

Thank you @barvic 
Do I understand correctly that you agree with cleaning the code as proposed in this PR, and silence out with it the static analyzer?

barvic:
> 
> > @perrotta It looks like most of those old code lines were used as a way to prevent compiler warnings about currently unused variables, instead of commenting it out. Most likely those variables are for not fully implemented features, which were in plans initially, but were later considered as not really needed or important for actual system operation.
> 
> Thank you @barvic
> Do I understand correctly that you agree with cleaning the code as proposed in this PR, and silence out with it the static analyzer?

@perotta As I see that comparison tests are passed so far, then it is OK with me.

perrotta:
+1
- Technical: just some dead wood cut away
- Jenkins tests pass

fabiocos:
+1

PR #26578, opened 2019-04-30
Decrease memory used by OscarMTProducer

Dr15Jones:
#### PR description:

Decrease the memory held by OscarMTProducer during event processing. This is accomplished by
- clearing temporary memory used for processing an event at the end of that event instead of at the beginning of the next event
- clear the G4 allocator used to store all CaloG4Hits each event rather than letting it grow for the entire job (this was found by IgProf)
- cleanup memory at the end of the job thereby making it easier to use tools to find per event memory leaks.

#### PR validation:

This was triggered from the production problem here: https://hypernews.cern.ch/HyperNews/CMS/get/simDevelopment/1891.html

I used the configuration from the hypernews and ran IgProfService to isolate memory leaks and per-event memory hoarding. Once all the changes were done, I ran it under CMSSW_10_6_ASAN_X_2019-04-26-2300 to double check there were no memory errors (it was clean).

Using the original code running with 8 threads on only 20 events, the memory kept growing and reached RSS=3.6GB After the code change, running on 8 threads over 135 events the job was still around RSS=2GB after 20 and stayed near 2.2GB until the very end of the job  when it grew temporarily to RSS=3.4.

### DIFF
--- a/SimG4CMS/Calo/interface/CaloSD.h
+++ b/SimG4CMS/Calo/interface/CaloSD.h
@@ -27,6 +27,7 @@
 
 #include <vector>
 #include <map>
+#include <memory>
 
 class G4Step;
 class G4HCofThisEvent;
@@ -163,6 +164,7 @@ private:
 
   std::map<CaloHitID, CaloG4Hit*> hitMap;
   std::map<int, TrackWithHistory*> tkMap;
+  std::vector<std::unique_ptr<CaloG4Hit>> reusehit;
 
   std::vector<CaloG4Hit*> reusehit;
   std::vector<CaloG4Hit*> hitvec;
--- a/SimG4CMS/Calo/interface/CaloSD.h
+++ b/SimG4CMS/Calo/interface/CaloSD.h
@@ -65,6 +65,7 @@ public:
 
   void clearHits() override;
   void fillHits(edm::PCaloHitContainer&, const std::string&) override;
+  void reset() override;
 
 protected:
   virtual double getEnergyDeposit(const G4Step* step);
--- a/SimG4CMS/Calo/interface/CaloSD.h
+++ b/SimG4CMS/Calo/interface/CaloSD.h
@@ -166,10 +166,6 @@ private:
   std::map<CaloHitID, CaloG4Hit*> hitMap;
   std::map<int, TrackWithHistory*> tkMap;
   std::vector<std::unique_ptr<CaloG4Hit>> reusehit;
-
-  std::vector<CaloG4Hit*> reusehit;
-  std::vector<CaloG4Hit*> hitvec;
-  std::vector<unsigned int> selIndex;
 };
 
 #endif  // SimG4CMS_CaloSD_h
--- a/SimG4CMS/Calo/src/CaloSD.cc
+++ b/SimG4CMS/Calo/src/CaloSD.cc
@@ -559,6 +559,12 @@ void CaloSD::clearHits() {
   slave.get()->Initialize();
 }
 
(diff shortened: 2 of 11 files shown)


civanch:
please test

civanch:
@Dr15Jones , great job! 
The fact, that there is no difference in regression means, that this PR is correct. Let me some time to understand, where we lost proper clean-up of intermediate calo hits - the problem appears sometime after 10_2. Concerning object deletion end of run: since migration to MT we have a G4 problem, that not all objects are properly deleted end of run. Geant4 was improving in this respect for each new release but this G4 10.4 is still not fully cleaned.  Moreover, OscarProducer should delete some G4 objects and I think what you propose in this PR is necessary and is more correct than before.

Dr15Jones:
For completion, I ran 135 events using the original code (all the numbers I've used have been from CMSSW_10_5_0 as that is where the production jobs saw the problem). In that case the job ran around RSS > 3GB for most and then at the end went up to 4GB.

civanch:
@Dr15Jones , I run ttbar 600 events for several versions of CMSSW and for this fix. In these events there is no memory grows on such statistics. What I see RSS=1.76 GB before the fix and 1.66 GB after the fix. This may be well explained by cleanup end of event. VSIZE is changed from 3.67 GB to 3.57 GB. What I do not understand - in CMSSW_10_1 for the same events but 2017 geometry we had RSS1.45 GB and VSIZE 2.34. If I rerun with this fix and 2017 global tag and era I still have RSS=1.65 GB and  VSIZE=3.57 GB. 

In summary, this PR really fix memory of G4Allocator and improve deletion at the end of run. However, I do not understand the increase RSS and VSIZE at some moment durining 10_2 developments. This may be happens not due to SIM itself but in other part of CMSSW, at least, cannot say this for sure.

Concerning G4Allocator: all objects following G4Allocator must be deleted just after usage, in this case, allocated memory will not grow and will be used efficiently.

civanch:
+1
This PR make an important fix of memory, @fabiocos , please, have a look and merge it.
Some questions still remains unclear, at least, for me, but it is a subject of further investigation.

Dr15Jones:
@civanch what I found helpful in finding why 'live' memory had increased was to add 
```
process.add_(cms.Service("IgProfService", 
                         reportEventInterval = cms.untracked.int32(10),
                         reportToFileAtPostEvent = cms.untracked.string("igtest_service_3.%E.out")))
```
to the configuration and then run
```
igprof -d -mp -z -o igprof_<something unique>.mp.gz -t cmsRunGlibC cmsRunGlibC <test_cfg.py>
```

and then compare two different outputs from the same job
```
igprof-analyse -d -v -g -r MEM_LIVE --baseline igtest_service_3.1.out --diff-mode igtest_service_3.11.out
```

You could do something similar between two different jobs. Use the IgProfService to 'snapshot' the live memory of both jobs at some point in time (say after 10 events) and then use `igprof-analyse` with `--diff-mode` to see what memory is different between the two jobs.

fabiocos:
@civanch did you perhaps compare with the same 600 events without this? Here we see identity, but usually just a  few tens of events are run (although if something really wrong happens it should likely appear)

fabiocos:
I've run 100 TTbar events in the latest IB with and without this PR, finding a set of differences that are anyway minor, what surprises me is that most of the G4 hits level plots are ok, but I see a couple of differences at GEN level (TTbar analysis, one particle of difference), and they are likely the cause of all the subsequent observed differences. Therefore I conclude that those differences are not due to this PR.

fabiocos:
+1

PR #26586, opened 2019-05-01
atomize counters in BaseIsolator.h

slava77:
some mutables are used as counters and are showing up as static analyzer warnings. In the case of  BaseIsolator, the simplest is to atomize them, as done in this PR.

The counters here are used for printing of statistics of the ```test ``` method calls. I do not see a strong need for a correctness of this in a multithreaded environment.
The ```atomic``` is effectively to just silence the warning.

### DIFF
--- a/PhysicsTools/PatAlgos/interface/BaseIsolator.h
+++ b/PhysicsTools/PatAlgos/interface/BaseIsolator.h
@@ -6,6 +6,8 @@
 #include "FWCore/Framework/interface/Event.h"
 #include "FWCore/Framework/interface/ConsumesCollector.h"
 
+#include <atomic>
+
 namespace pat { namespace helper {
 class BaseIsolator {
     public:
@@ -34,7 +36,7 @@ class BaseIsolator {
         edm::InputTag input_;
         edm::EDGetTokenT<Isolation> inputToken_;
         float cut_;
-        mutable uint64_t try_, fail_;
+        mutable std::atomic<uint64_t> try_, fail_;
 }; // class BaseIsolator
 } } // namespaces
 


slava77:
@cmsbuild please test

slava77:
@cmsbuild please test

slava77:
+1

for https://github.com/cms-sw/cmssw/pull/26586 	fb090c5
- technical; clears static analyzer warnings related to the files/classes mentioned (especially visible for the BaseIsolator which is included in many places)
- jenkins tests pass and comparisons with the baseline show no differences; the static analyzer for this PR if not showing the issues anymore with the updated variables

slava77:
@cmsbuild please test

after taking out the updates in EG\*Modifier\*

slava77:
+1

for https://github.com/cms-sw/cmssw/pull/26586	777a272
- technical
- jenkins tests pass

fabiocos:
@slava77 I am not sure I fully get the meaning of your last sentence in the PR description. Defining the counters as atomic ensures that there is no conflicting updates by multiple threads, but the global counter should still finally be correct. Or am I missing something? The problem might happen if a thread is reading the counters while another is writing them

slava77:
Ideally,
- both counters should be updated in sync.
- the print statement, which is now the only place to use the counters knows nothing about updates in other threads and one may hope for consistent results in the same thread between the call to a ```test``` method and the ```print``` method.

I do not consider having the above done appropriately as a requirement for this code full thread consistency/safety. Hence, my last message in the PR description.

fabiocos:
+1

fabiocos:
merge

santocch:
+1

PR #26589, opened 2019-05-01
Do not merge, Implement core support for concurrent IOVs

wddgit:
#### PR description:

Please do not merge this PR. This is not intended for 10_6_X. It is intended for 11_0_X. In addition, I intend to rebase this on top of the PR adds the scram code-format changes to the Framework before it gets merged into the master branch.

I am submitting this PR now to allow for review and comments so I can start work on any requested changes or questions. I'm hoping to get this merged as soon as possible after 11_0_X is open.

This implements core support for concurrent IOVs. More work needs to be done before CMSSW is actually running concurrent IOVs. Outside of core tests, there are not any ESSources or ESProducers that have been modified to actually support this kind of concurrency. Further work will be needed to convert ESSources and ESProducers outside the core. The core will continue to support and properly run existing ESSources and ESProducers, but the IOVs will not be processed concurrently.

The only changes outside the core are to CondDBESSource, FWLiteESSource, and FWLiteESRecordWriterAnalyzer. These changes are the minimal changes necessary to work with the revised core code. They do not migrate these modules to support concurrent IOVs.

There are two independent ways to control the number of concurrent IOVs:

First one can set the number of IOVs allowed to be processed at the same time for each EventSetupRecord as follows. In the configuration:

```
    numberOfConcurrentIOVs = cms.untracked.uint32(4),
    forceNumberOfConcurrentIOVs = cms.untracked.PSet(
        ESTestRecordA = cms.untracked.uint32(4)
    )
)
```

In the C++ definition of an EventSetupRecord type:

```
  static constexpr bool allowConcurrentIOVs_ = false;
```

numberOfConcurrentIOVs sets the number of IOVs for all types of EventSetupRecord (defaults to 1). The boolean defined in the C++ class overrides this for a specific EventSetupRecord type (by default this is defined true in the base class). The value in forceNumberOfConcurrentIOVs overrides both numberOfConcurrentIOVs and the value in the C++ class for specified EventSetupRecord types (by default this parameter set is empty).

The second way of controlling this relates to ESSources. Existing ESSources often rely on the time ordering of calls to setIntervalFor and the associated functions that produce data for that interval. In the past the produce function always referred to the immediately preceding setIntervalFor function call. But with concurrent IOVs, this can no longer be true. Existing ESSources will be treated as legacy ESSources. As ESSources are modified to support concurrent IOVs, the following function should be added to ESSource C++ class definition:

```
    bool isLegacyESSource() const override { return false; }
```

For legacy ESSources and when an IOV for that ESSource changes, then the Framework will wait until that IOV completes before moving forward to the next one. It would be nice someday to convert all ESSources and remove this function and this synchronization point entirely someday.

In many but not all cases, existing ESProducers will work in concurrent IOV mode without modification. Most ESSources are going to require work to migrate.

Note that the two mechanisms described above are independent of each other.

If numberOfConcurrentIOVs is set to greater than one and nothing else is done, things might fail horribly.  It will take a record by record review of the associated/dependent ESSource and ESProducer code to determine what will and will not function properly with more than one concurrent IOV.

The new EventSetupRecordIOVQueue class is at the heart of all this. It is probably the most important part of the new code in this PR.

#### PR validation:

This should not effect the results of existing tests. Without modifying things as described above, IOVs will not be running concurrently and everything should be the same. Even when IOVs are running concurrently, this should not modify behavior or results other than performance improvements caused by increasing concurrency at IOV boundaries.

### DIFF
--- a/CondCore/ESSources/plugins/CondDBESSource.cc
+++ b/CondCore/ESSources/plugins/CondDBESSource.cc
@@ -516,10 +516,13 @@ CondDBESSource::setIntervalFor( const edm::eventsetup::EventSetupRecordKey& iKey
     edm::IOVSyncValue start = cond::time::toIOVSyncValue(recordValidity.first, timetype, true);
     edm::IOVSyncValue stop = doRefresh  ? cond::time::limitedIOVSyncValue (iTime, timetype)
       : cond::time::toIOVSyncValue(recordValidity.second, timetype, false);
-       
+
+    if (start == edm::IOVSyncValue::invalidIOVSyncValue() && stop != edm::IOVSyncValue::invalidIOVSyncValue()) {
+      start = edm::IOVSyncValue::beginOfTime();
+    }
     oInterval = edm::ValidityInterval( start, stop );
-   }
-  
+  }
+
   edm::LogInfo( "CondDBESSource" ) << "Setting validity for record \"" << recordname 
 				   << "\" and corresponding label(s): starting at " << oInterval.first().eventID() << ", timestamp: " << oInterval.first().time().value()
 				   << ", ending at "<< oInterval.last().eventID() << ", timestamp: " << oInterval.last().time().value()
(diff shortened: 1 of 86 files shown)


wddgit:
please test

wddgit:
Format changes were merged faster than I expected (thanks). I am rebasing this PR now and will resubmit as a new PR soon.

Issue #26496, opened 2019-04-18
Wrong getNbins function in Utilities/RelMon


peruzzim:
The getNbins function in Utilities/RelMon/python/utils.py seems wrongly defined or used:

https://github.com/cms-sw/cmssw/blob/8c0b3deb7bf6fc4d0a30a6696fb3b0944316d0c4/Utilities/RelMon/python/utils.py#L92-L97

https://github.com/cms-sw/cmssw/blob/8c0b3deb7bf6fc4d0a30a6696fb3b0944316d0c4/Utilities/RelMon/python/utils.py#L330-L334

In this case, for instance, a 1-D histogram defined with N bins (excluding under/overflow) would give getNbins = N+1, and then range(0,N+3) which results in calling GetBinContent(N+2).

This breaks statistical tests in the same package (e.g. a file does not compare ok to itself using the BinToBin test). If you agree I can submit a PR to fix it.

Dr15Jones:
assign reconstruction

slava77:
reco is not really using relmon.
If someone will provide a fix, we'll review and sign.

fabiocos:
@peruzzim you seem to refer to the printout (nok vs ntot), not the final summary right? For what I can see the problem is in ```nbin+2```, that should be just nbin...

fabiocos:
@peruzzim the issue looks to me more complex, looking inside the code it seems to me that there is an inconsistent treatment of ranges. The second value in a range() function is the upper limit of the range *not included in the sequence* . So if we want to use 0 for the underflow and GetNbinsX()+1 for the overflow, the count of ok and total bins as it is now seems to me wrong. I am having a look at what to fix, but it seems to me there are several lines that should be modified...

davidlange6:
maybe @dpiparo still remembers this code?

fabiocos:
@davidlange6 looking at the function I think it is a matter to go systematically through the places where number of bins are used in loops and have all of them treated coherently. I am preparing a PR for this

perrotta:
+1
- Fixed with #26593

dpiparo:
@davidlange6 , saw this too late. Thanks for pinging me.

fabiocos:
@dpiparo in case have a look, I think that now the textual report is more consistent, while the histogram comparison should not be affected.

dpiparo:
@fabiocos , for me the changes are good to go (not following closely this repo yet, again, apologies for the delay in answering)

[...]

PR #26593, opened 2019-05-02
RelMon histogram comparison: adapt ranges in a compatible way with the definition of getNbins

fabiocos:
#### PR description:

The RelMon DQM histogram comparison is complaining about differences between histograms when comparing a file with itself, which is clearly incorrect, see the discussion in https://github.com/cms-sw/cmssw/issues/26496 . This issue looks linked to the inconsistent use in the utility package https://cmssdt.cern.ch/lxr/source/Utilities/RelMon/python/utils.py of number of bins and ranges. In python the range function has as a second argument not the last value of the list but the next to last. This looks likely the motivation for the function getNbins adding 1 unit to the number of bins as returned by the ROOT function getNbinsX(). There seems to be also some inconsistency between the starting point of the different range loops, in several cases bin 0 is used (underflow), sometime bin 1 (the first histogram bin). 

This PR tries to solve the issue, having all the range based loops running from the underflow (bin=0) to the overflow (bin=GetNbinsX()+1), which means setting in the range loops as last item GetNbinsX()+2. 

#### PR validation:

Running the test BinToBin to compare a file with itself does not show any more the problem. Using the DQM output of the generator workflow 503 with a syntax similar to cms-bot:
```
13:51 cmsdev25 4149> compare_using_files.py -B DQM/TimerService@3 -o out -C -R -p --no_successes -s b2b 503.0_MinBias_13TeV_pythia8+MinBias_13TeV_pythia8+HARVESTGEN/DQM_V0001_R000000001__Global__CMSSW_X_Y_Z__RECO.root 503.0_MinBias_13TeV_pythia8+MinBias_13TeV_pythia8+HARVESTGEN/DQM_V0001_R000000001__Global__CMSSW_X_Y_Z__RECO.root
Analysing Histograms located in directory  at: 
 o 503.0_MinBias_13TeV_pythia8+MinBias_13TeV_pythia8+HARVESTGEN/DQM_V0001_R000000001__Global__CMSSW_X_Y_Z__RECO.root
 o 503.0_MinBias_13TeV_pythia8+MinBias_13TeV_pythia8+HARVESTGEN/DQM_V0001_R000000001__Global__CMSSW_X_Y_Z__RECO.root
We have a Blacklist:
 o Directory TimerService at level 3
++++++++++++++++++++++++++++++
Output Directory will be  out
Studying directory Info, 1/2
Studying directory Generator, 2/2
 ->Appending Generator... Appended.
Finished

 - summary of 840 tests:
 o Failiures: 0.00% (0/840)
 o Nulls: 0.00% (0/840) 
 o Successes: 100.00% (840/840) 
 o Skipped: 0.00% (0/840) 
 o Missing objects: 0
Pickleing the directory as out.pkl in dir /build/fabiocos/106X/relmon/CMSSW_10_6_X_2019-04-29-2300/work/out
Reading directory from out.pkl
Calculating stats for the directory...
Producing html...
```

### DIFF
--- a/Utilities/RelMon/python/utils.py
+++ b/Utilities/RelMon/python/utils.py
@@ -90,11 +90,20 @@ def literal2root (literal,rootType):
 #-------------------------------------------------------------------------------
 
 def getNbins(h):
+  """
+  To be used in loops on bin number with range()
+  In case of 1-d histograms the end-of-range value should be GetNbinsX()+2 
+  In case of 2-d or 3-d histograms the product of GetNbins.()+2 gives the total number of bins, and 1 needs to be added
+  Managed using an offset variable 
+  """
+  offset=0
   biny=h.GetNbinsY()
-  if biny>1:biny+=1
+  if biny>1:
+    offset=1
+    biny+=2
   binz=h.GetNbinsZ()
-  if binz>1:binz+=1
-  return (h.GetNbinsX()+1)*(biny)*(binz)
+  if binz>1:binz+=2
+  return (h.GetNbinsX()+2)*(biny)*(binz)+offset
 
 #-------------------------------------------------------------------------------
 
@@ -168,7 +177,7 @@ class StatisticalTest(object):
 #-------------------------------------------------------------------------------
 
 def is_empty(h):
-  for i in range(0,getNbins(h)+1):
+  for i in range(0,getNbins(h)):
     if h.GetBinContent(i)!=0: return False
   return True
   #return h.GetSumOfWeights()==0
@@ -253,7 +262,7 @@ class Chi2(StatisticalTest):
   def absval(self):
     nbins=getNbins(self.h1)
     binc=0
-    for i in range(0,nbins+1):
+    for i in range(0,nbins):
       for h in self.h1,self.h2:
         binc=h.GetBinContent(i)
         if binc<0:
@@ -329,7 +338,7 @@ class BinToBin(StatisticalTest):
     equal = 1
     nbins = getNbins(self.h1)
     n_ok_bins=0.0
-    for ibin in range(0, nbins+1):
+    for ibin in range(0, nbins):
       h1bin=self.h1.GetBinContent(ibin)
       h2bin=self.h2.GetBinContent(ibin)
       bindiff=h1bin-h2bin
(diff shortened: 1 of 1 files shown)


fabiocos:
@peruzzim FYI

fabiocos:
please test

perrotta:
type bugfix

fabiocos:
please test

Utilities/RelMon/python/utils.py
```diff
@@ -90,11 +90,20 @@ def literal2root (literal,rootType):
...
+  To be used in loops on bin number with range()
+  In case of 1-d histograms the end-of-range value should be GetNbinsX()+2 
+  In case of 2-d or 3-d histograms the product of GetNbins.()+2 gives the total number of bins, and 1 needs to be added
+  Managed using an offset variable 
+  """
+  offset=0
   biny=h.GetNbinsY()
-  if biny>1:biny+=1
+  if biny>1:
+    offset=1
```
perrotta: PLease verify whether this offset is really needed.
In a 2D histogram there are (NBinsX+2)*(NBinsY+2) bins, including underflow and overflow. That is, they numbers from "0" to "(NBinsX+2)*(NBinsY+2)-1", and to loop through all them the second argument of the (python) range function must be "(NBinsX+2)*(NBinsY+2)", without such an offset.

fabiocos: @perrotta right, the point is that the range loops using this output run from 0 up to the output of this function now, which must be (NBinsX+2)(NBinsY+2)+1 to account for the fact that the last element of the range() must be the last item of the list +1. Do you agree?

perrotta: If I am not wrong,(see above)  the last item in the list is "(NBinsX+2)(NBinsY+2)-1" and the last item+1 is "(NBinsX+2)(NBinsY+2)"

fabiocos: @perrotta no, I think you are correct, range() starts from 0 but the output of this functions counts from 1, so ok

fabiocos:
please test

perrotta:
+1
- Fixed as planned in the description
- Jenkins tests pass

fabiocos:
+1

PR #26601, opened 2019-05-02
Changes from scram build format-all on FWCore

wddgit:
PR description:

The first commit includes only changes from scram build code-format-all and nothing else. This PR includes all the packages in the FWCore subsystem not previously formatted.

There were a few addition changes that showed up in code-checks in the second commit.

PR validation:

The only test I performed was checking that the revised packages compile and build.


FWCore/Skeletons/scripts/mkTemplates/EDProducer/EDProducer.cc
```diff
@@ -30,44 +29,45 @@
...
-@example_myparticle #include "DataFormats/MuonReco/interface/Muon.h"
-@example_myparticle #include "DataFormats/EgammaCandidates/interface/PixelMatchGsfElectron.h"
-@example_myparticle #include "DataFormats/Candidate/interface/Particle.h"
-@example_myparticle #include "FWCore/MessageLogger/interface/MessageLogger.h"
-@example_myparticle #include "FWCore/Utilities/interface/InputTag.h"
+@example_myparticle #include "DataFormats/MuonReco/interface/Muon.h" @example_myparticle #include
+    "DataFormats/EgammaCandidates/interface/PixelMatchGsfElectron.h" @example_myparticle #include
+    "DataFormats/Candidate/interface/Particle.h" @example_myparticle #include
+    "FWCore/MessageLogger/interface/MessageLogger.h" @example_myparticle #include
+    "FWCore/Utilities/interface/InputTag.h"
```
wddgit: This breaks the Skeletons package I think. I will take a closer look at this.

wddgit:
please test

Looks like the Skeletons package is broken by this. I'm taking a closer look at that. I'll run the tests anyway and see what happens.

wddgit:
@smuzaffar Is there a way to turn off scram code-format for certain files or packages? It seems like this can never work properly for the files that are not really code but are actually templates for code generation with special embedded text sequences that are not C++ and have a special meaning. So I would recommend turning this off for the Skeletons package. I will probably close this PR and resubmit tomorrow with the Skeletons package left out. Any other ideas?

Independent of that we need to do some updating work in these Skeletons template files, but that should be a separate PR.

davidlange6:
clang-format can let you exclude code from its checks - but perhaps non-code should be named something other than .cc? Is that technically needed?

wddgit:
It will take me a little while to understand the Python scripts in that package, but I would imagine it is possible to give the files different suffixes. That might be the best idea. I will discuss this with Chris tomorrow.

smuzaffar:
@wddgit , starting from  CMSSW_10_6_X_2019-05-02-2300 , code-format-all now only runs on
- All files compiled for release + any includes
- files changed w.r.t release
So basically existing files (even with .cc or .h extension) will not be formatted if they were not compiled. Yes if you add a new code skeleton .cc, .h file then scram will try to format it. In that case we need to exclude those from automatic checks. 
@davidlange6 , @fabiocos , should we add src/.clang-format-ignore to keep a list of files to be excluded from automated checks. Note the there are some fortran .h files which clang-format failed to format and we need to exclude such files.

Dr15Jones:
> should we add src/.clang-format-ignore to keep a list of files to be excluded from automated checks.

I'm against the idea of one central such file and that means constant editing. If we were to have such a file per directory that only references files in that directory, I'd be OK with that.

davidlange6:
I'm for using the existing clang-tidy/clang-format infrastructure instead of inventing our own

wddgit:
Close. I will resubmit a new PR. It will be the same except with the Skeletons package excluded. We can handle that separately.

PR #26501, opened 2019-04-19
Add DT digis and CSC strip digis to RECO event content

battibass:
#### PR description:

This PR is intended as a temporary fix for the problem with RECOfromRECO workflow arose in [PR 26369](https://github.com/cms-sw/cmssw/pull/26369#discussion_r276960586). It was implemented as a follow-up action from today's RECO meeting.

#### PR validation:

The increase in event size was tested in the following workflows:

**136.879** (RunSingleMu2018C)
[...]
muonCSCDigis:MuonCSCStripDigi (CSCDetIdCSCStripDigiMuonDigiCollection) 183824 18999.5
muonDTDigis: (DTLayerIdDTDigiMuonDigiCollection) 14911.6 2932.03
[...]
for a total of ~ 2.8 M event for the RECO dataset from step3

**1102.0_RR** (TTbar 8 TeV simulation with RECOfromRECO step)
[...]
uonCSCDigis:MuonCSCStripDigi (CSCDetIdCSCStripDigiMuonDigiCollection) 22364.8 2211.07
muonDTDigis: (DTLayerIdDTDigiMuonDigiCollection) 639.1 131.29
[...]
for a total of ~ 0.64 M event for the RECO dataset from step3

@Fedespring @slava77 @fabiocos

Issue #26500, opened 2019-04-19
allow modifiers for reconstruction_fromRECO


slava77:
The goal is to allow modifications to module PSets of the RECOfromRECO workflows (which is not possible now).

A process modifier is perhaps the best solution, similar to e.g. Configuration/ProcessModifiers/python/trackdnn_cff.py : 
- make a recoFromReco modifier in Configuration/ProcessModifiers
- add a relevant ```recoFromReco.toModify( `` in  
- add ```recoFromReco``` to the relevant workflows ... 

It looks like currently the "right" location for the last step is the ConfigBuilder in connection with the decision to use the fromReco sequence made in https://github.com/cms-sw/cmssw/blob/1433577299868110ae65fed1f32ac1edaf0de6f2/Configuration/Applications/python/ConfigBuilder.py#L975-L978 to have recoFromReco proc modifier to be inserted 

The need for this will also resolve the problem that was reported in https://github.com/cms-sw/cmssw/pull/26369#discussion_r276960586

@fabiocos @battibass

slava77:
assign reconstruction

preferred timescale: 10_6_0 release

slava77:
+1

(essentially) resolved by #26615

PR #26602, opened 2019-05-02
BugFix for the reconstruction_fromReco sequences: https://github.com/cms-sw/cmssw/issues/26500

Fedespring:
#### PR description:
The PR fixes the reconstruction_fromReco sequences. The fix is needed in pre4, because the #26369 takes in input collections of digis which are missing running the reRECO process. The fix consists in switching off the process which uses the digi info. The temporary fix done in PR #26501 is not needed anymore. And the #26501 can be reverted.

#### PR validation:
Validation has been successfully done on the process  runTheMatrix.py -l 1102.0.

### DIFF
--- a/RecoMuon/MuonIdentification/python/muons1stStep_cfi.py
+++ b/RecoMuon/MuonIdentification/python/muons1stStep_cfi.py
@@ -100,3 +100,11 @@ muonEcalDetIds = cms.EDProducer("InterestingEcalDetIdProducer",
 
 from Configuration.Eras.Modifier_pp_on_AA_2018_cff import pp_on_AA_2018
 pp_on_AA_2018.toModify(muons1stStep, minPt = 0.8)
+
+from Configuration.ProcessModifiers.recoFromReco_cff import recoFromReco
+recoFromReco.toModify(muons1stStep,fillShowerDigis = cms.bool(False))
+
+
+
+
+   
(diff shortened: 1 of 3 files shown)


kpedro88:
please test workflow 136.879

RecoMuon/MuonIdentification/python/muons1stStep_cfi.py
```diff
@@ -100,3 +100,11 @@
 
 from Configuration.Eras.Modifier_pp_on_AA_2018_cff import pp_on_AA_2018
 pp_on_AA_2018.toModify(muons1stStep, minPt = 0.8)
+
+from Configuration.ProcessModifiers.recoFromReco_cff import recoFromReco
+recoFromReco.toModify(muons1stStep,fillShowerDigis = cms.bool(False))
```
slava77: please drop type specifications for all parameters which already exist.

Specifically, ```fillShowerDigis = False``` in this case

slava77:
> And the #26501 can be reverted.

Please do it in this PR so that we can fully test the effect of this PR

perrotta:
type bugfix

Fedespring:
A new PR 26615 has been done, starting to the latest configuration build released today. In this way it has been possible to revert the #26501 together with the proposed bug-fix

PR #26606, opened 2019-05-03
Restore FED error treatment for phase0 (required for UL Rereco of 2016)

veszpv:
#### PR description:

The FED raw data format for phase 0 and phase 1 detectors are essentially the same with small differences for the bit-packing of the pixel hit information in layer 1. There is a difference also in the error assignment by the front-end (*). The former case has been treated in the PixelDataFormatter by setting the 'phase1' variable through python configuration files. The error decoding class, called ErrorChecker, however was not made aware of the difference, instead, was just simply updated to phase 1. 
This PR fixes that: a base class is defined for the ErrorChecker, the present default code is kept with the same name deriving from this base, and a new derived class ErrorCheckerPhase0 is introduced to restore the phase 0 treatment. The proper error checker is instantiated by the PixelDataFormatter based on the value of 'phase1', so no change needs to be made in the python configurations.

FED error 25: in phase 1, FED error type 25 has been reinterpreted and used by the FED firmware to signal "dead" links. It is used to flag stuck TBMs. This error has a totally different meaning in phase 0. In this PR, the special treatment of this error type by raw-to-digi and digi-to-raw is cancelled for phase 0. 

(*) Few error codes have different interpretation; and in phase 0 FPix, the error association with detId is ambiguous. 

#### PR validation:

During the reconstruction of phase 0 data, when a rare type 25 error was encountered, the tracking code crashed. This PR fixes that. The FED25 error packing in simulation, and unpacking in simulation and data for phase 1 should be unchanged, are being validated by @tsusa. 

@tsusa @mmusich @dkotlins

### DIFF
--- a/EventFilter/SiPixelRawToDigi/plugins/SiPixelRawToDigi.cc
+++ b/EventFilter/SiPixelRawToDigi/plugins/SiPixelRawToDigi.cc
@@ -42,284 +42,289 @@
 using namespace std;
 
 // -----------------------------------------------------------------------------
-SiPixelRawToDigi::SiPixelRawToDigi( const edm::ParameterSet& conf ) 
-  : config_(conf), 
-    badPixelInfo_(nullptr),
-    regions_(nullptr),
-    hCPU(nullptr), hDigi(nullptr)
-{
-  
+SiPixelRawToDigi::SiPixelRawToDigi(const edm::ParameterSet& conf)
+    : config_(conf), badPixelInfo_(nullptr), regions_(nullptr), hCPU(nullptr), hDigi(nullptr) {
   includeErrors = config_.getParameter<bool>("IncludeErrors");
   useQuality = config_.getParameter<bool>("UseQualityInfo");
   if (config_.exists("ErrorList")) {
-    tkerrorlist = config_.getParameter<std::vector<int> > ("ErrorList");
+    tkerrorlist = config_.getParameter<std::vector<int>>("ErrorList");
   }
   if (config_.exists("UserErrorList")) {
-    usererrorlist = config_.getParameter<std::vector<int> > ("UserErrorList");
+    usererrorlist = config_.getParameter<std::vector<int>>("UserErrorList");
   }
-  tFEDRawDataCollection = consumes <FEDRawDataCollection> (config_.getParameter<edm::InputTag>("InputLabel"));
-  
+  tFEDRawDataCollection = consumes<FEDRawDataCollection>(config_.getParameter<edm::InputTag>("InputLabel"));
+
   //start counters
   ndigis = 0;
   nwords = 0;
 
   // Products
-  produces< edm::DetSetVector<PixelDigi> >();
-  if(includeErrors){
-    produces< edm::DetSetVector<SiPixelRawDataError> >();
+  produces<edm::DetSetVector<PixelDigi>>();
+  if (includeErrors) {
+    produces<edm::DetSetVector<SiPixelRawDataError>>();
     produces<DetIdCollection>();
     produces<DetIdCollection>("UserErrorModules");
-    produces<edmNew::DetSetVector<PixelFEDChannel> >();
+    produces<edmNew::DetSetVector<PixelFEDChannel>>();
   }
-  
+
   // regions
   if (config_.exists("Regions")) {
-    if(!config_.getParameter<edm::ParameterSet>("Regions").getParameterNames().empty())
-    {
+    if (!config_.getParameter<edm::ParameterSet>("Regions").getParameterNames().empty()) {
       regions_ = new PixelUnpackingRegions(config_, consumesCollector());
     }
   }
 
   // Timing
-  bool timing = config_.getUntrackedParameter<bool>("Timing",false);
+  bool timing = config_.getUntrackedParameter<bool>("Timing", false);
   if (timing) {
-    theTimer.reset( new edm::CPUTimer );
-    hCPU = new TH1D ("hCPU","hCPU",100,0.,0.050);
-    hDigi = new TH1D("hDigi","hDigi",50,0.,15000.);
+    theTimer.reset(new edm::CPUTimer);
+    hCPU = new TH1D("hCPU", "hCPU", 100, 0., 0.050);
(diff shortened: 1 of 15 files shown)


EventFilter/SiPixelRawToDigi/plugins/SiPixelRawToDigi.cc
```diff
@@ -252,7 +252,7 @@ void SiPixelRawToDigi::produce( edm::Event& ev,
 	  for (auto const& aPixelError : errorDetSet) {
 	    // For the time being, we extend the error handling functionality with ErrorType 25
 	    // In the future, we should sort out how the usage of tkerrorlist can be generalized
-	    if (aPixelError.getType()==25) {
+	    if (usePhase1==true && aPixelError.getType()==25) {
```
slava77: is this even needed on the raw2digi side?
Do we have cases for valid pixel error =25 in phase 0?

veszpv: In phase 0, error type 25 is exotic but can happen when AFAIK the analog levels in the FED input are not interpreted correctly and the communication protocol is compromised. Such an event was actually caught by reconstruction. So this is indeed needed.

mmusich: Actually this fix is mandatory for full reprocessing of 2016. Without it you will have a small but finite chunk of unprocessable data

perrotta:
@veszpv  please apply code checks prescription (https://github.com/cms-sw/cmssw/pull/26606#issuecomment-489116317) so that tests can ge started

veszpv:
Note, it seems that "scram build code-format" gives result different from applying the downloaded patch.

The actual changes are quite minimal. 1) Two lines of "if" statements disable the reading/filling of the bad FED channel collection by the (un)packer code, 2) the current ErrorCheck class is made to inherit from a pure virtual base class and the old ErrorCheck from the last phase 0 release (CMSSW_9_3_14_patch1) is copied back as another derived class. Exact same code as used in the last phase 0 reco, only changed to compile in the present release.

perrotta:
please test

EventFilter/SiPixelRawToDigi/src/PixelDataFormatter.cc
```diff
@@ -280,228 +317,242 @@ void PixelDataFormatter::interpretRawData(bool& errorsInEvent, int fedId, const
...
-int PixelDataFormatter::digi2wordPhase1Layer1( cms_uint32_t detId, const PixelDigi& digi, 
-    std::map<int, vector<Word32> > & words) const
-{
+int PixelDataFormatter::digi2wordPhase1Layer1(
+    cms_uint32_t detId, const PixelDigi &digi,
+    std::map<int, vector<Word32>> &words) const {
   LogDebug("PixelDataFormatter")
-// <<" detId: " << detId 
-  <<print(digi);
+      // <<" detId: " << detId
```
perrotta: remove

EventFilter/SiPixelRawToDigi/src/PixelDataFormatter.cc
```diff
@@ -280,228 +317,242 @@ void PixelDataFormatter::interpretRawData(bool& errorsInEvent, int fedId, const
...
+    return fedId;
 
-  int col = ((cabling.dcol)*2)  + ((cabling.pxid)%2);
-  int row = LocalPixel::numRowsInRoc - ((cabling.pxid)/2);
+  int col = ((cabling.dcol) * 2) + ((cabling.pxid) % 2);
+  int row = LocalPixel::numRowsInRoc - ((cabling.pxid) / 2);
 
-  //if(DANEK) cout<<" layer 1: digi2raw "<<detId<<" "<<digi.column()<<" "<<digi.row()<<" "<<digi.adc()<<" "
-  //  <<cabling.link<<" "<<cabling.roc<<" "<<cabling.dcol<<" "<<cabling.pxid<<" "
+  // if(DANEK) cout<<" layer 1: digi2raw "<<detId<<" "<<digi.column()<<"
```
perrotta: Please remove these commented out lines if not needed any more

EventFilter/SiPixelRawToDigi/src/PixelUnpackingRegions.cc
```diff
@@ -58,18 +61,16 @@ void PixelUnpackingRegions::run(const edm::Event& e, const edm::EventSetup& es)
   edm::Handle<reco::BeamSpot> beamSpot;
   e.getByToken(tBeamSpot, beamSpot);
   beamSpot_ = beamSpot->position();
-  //beamSpot_ = math::XYZPoint(0.,0.,0.);
+  // beamSpot_ = math::XYZPoint(0.,0.,0.);
```
perrotta: Remove

perrotta (commented):
The most awful thing in this PR is the line splitting as it was automatically implemented by the code checks: it makes the code much less easily readable in several parts, besides complicating quite a lot the review by adding tons of unneeded (at best) or worsening changes. In the following review I tried to avoid commenting about those splittings. but it would be nice if they could be reverted here, and their implementation not enforced any more in the clang checks. or at least somehow tuned.

veszpv:
I am not sure how to undo the changes by code check. If you want, I can try to run 
git push origin +HEAD
after I removed the commits from my local repo. Would that help? Or I can submit another PR. You are commenting on code that has nothing to do with this PR. Also, I do not think it is feasible to review the entire SiPixelRawToDigi package now.

perrotta:
> I am not sure how to undo the changes by code check. If you want, I can try to run
> git push origin +HEAD
> after I removed the commits from my local repo. Would that help? Or I can submit another PR. You are commenting on code that has nothing to do with this PR. Also, I do not think it is feasible to review the entire SiPixelRawToDigi package now.

Please don't mind if I don't like the clang prescriptions of the code checks: you can just concentrate on my comments for now.

Otherwise, you could submit another PR which only contains your original commits, profiting of the fact that code checks have been now momentarily disabled in order to speed up integrations for pre5.

veszpv:
Resubmitted this PR as [PR26679](https://github.com/cms-sw/cmssw/pull/26679) without the format checks for easier review.

PR #26285, opened 2019-03-28
TrackerCommonData: Categorization script update

vargasa:
#### PR description:

`TrackerCommonData/data/Materials` package provides a script for building the categorization maps required for running `Validation/Geometry` module. This script was *a little bit* outdated (FORTRAN77) and here it is being rewritten to python3 with the same functionality (except the creation of `pdf` files, information which is now being printed directly on the console).

#### PR validation:

This update was proposed in a [Tracker DPG Meeting](https://indico.cern.ch/event/787654/contributions/3298998/attachments/1786295/2908498/Material_Budget_Categorization.pdf) and results in [the Material budget](https://indico.cern.ch/event/787656/contributions/3331625/attachments/1801554/2938673/FortranToPython_MaterialBudget.pdf) meeting. Validation was performed by comparing the output of pdf files and the new code. We'll keep the FORTRAN file until a full update of the documentation has been done.

PR #26609, opened 2019-05-03
Geometry/TrackerCommonData: Move standalone utilities to bin folder

vargasa:
#### PR description:

Move standalone utilities to bin folder as discussed [here](https://github.com/cms-sw/cmssw/pull/26285#issuecomment-481576737)

#### PR validation:

Hardcoded relative paths have been changed as well

No backport required


cvuosalo:
@cmsbuild please test

Geometry/TrackerCommonData/bin/mixture.py
```diff
@@ -185,8 +185,12 @@ def main(argv):
     radLenFile = open(fname + ".x0","w");
     intLenFile = open(fname + ".l0","w");
 
-    listOfMixedMaterials = loadMaterialsFile(inputFile="mixed_materials.input")
-    listOfPureMaterials = loadMaterialsFile(inputFile="pure_materials.input")
+    listOfMixedMaterials = loadMaterialsFile(
+        inputFile="../data/mixed_materials.input"
```
cvuosalo: It would be better to use a full path, so there is no limitation on which directory the script is run from. I think this should work in Python:
`inputFile = os.path.expandvars("$CMSSW_BASE/Geometry/TrackerCommonData/data/pure_materials.input"`

I don't know how to do this in Fortran.

vargasa: You're certainly right. I don't know how to do it in Fortran either, but I'll check

vargasa: For the record: `g77` is not available on SLC7

davidlange6:
well, no, its not general enough. you need to use the fileinpath mechanism.

vargasa:
Do you think so? This is _not_ a configuration file

andrius-k:
+1
A minimal change for DQM

davidlange6:
I think that bin would be an appropriate directory if a) these tools had a non-generic name, and b) these tools worked without the package being checked out into a developer area. As neither is the case, test is probably a better spot.. (indeed, the code doesn't belong in data/)

fabiocos:
@davidlange6 @vargasa I suggested to move the scripts away from data into a bin directory, I understand that you would consider more appropriate a script directory? As none of them exist and now the code is path independent, I think that one or the other is just a matter of choice

cvuosalo:
I think David is suggesting that the scripts be moved to `test`.

fabiocos:
fine for me the test area, strictly speaking this is not a test but I would say that we have several other examples of utilities store there. In any case for sure more appropriate than data.

cvuosalo:
please test

jfernan2:
+1

cvuosalo:
+1

fabiocos:
+1

PR #26616, opened 2019-05-03
Fix dressed lepton constituents, add tau ghosts

mseidel42:
#### PR description:

* Rivet 2.7.0 requires changes to access dressed lepton constituents (= bare lepton + photons). This did not impact any calculations done with the dressed lepton as a whole.
* Added ghost tau ancestors to identify electrons/muons from tau decays.

#### PR validation:

Tested on ttbar sample, number of dressed lepton constituents has entries >1 as expected again. Tau ghosts increase number of constituents and show up in the `tags` collection.

### DIFF
--- a/GeneratorInterface/RivetInterface/interface/RivetAnalysis.h
+++ b/GeneratorInterface/RivetInterface/interface/RivetAnalysis.h
@@ -23,7 +23,7 @@ namespace Rivet {
   class RivetAnalysis : public Analysis {
 
     public:
-      std::vector<DressedLepton> leptons() const {return _leptons;}
+      ParticleVector leptons() const {return _leptons;}
       ParticleVector photons() const {return _photons;}
       ParticleVector neutrinos() const {return _neutrinos;}
       Jets jets() const {return _jets;}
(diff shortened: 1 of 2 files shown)


qliphy:
please test

qliphy:
+1

fabiocos:
+1

peruzzim:
@intrepid42: discussing with @emanca, I understand this feature is meant to support analysis. Should it be backported to 10_2_X? Is it going to affect the products in ongoing production campaigns?

intrepid42:
@peruzzim Yes, it would be great to have this in centrally produced nanoAod. It adds one tag particle for the dressed lepton GenJets. Please let us know when it would be fine to insert this change. The code needed for 10_2_X will be a bit different because of the older Rivet version there.

peruzzim:
The question is, whether merging this code (or an equivalent version of it, because of different Rivet version) will change the output of MINIAOD production in 10_2_X. If so, it would be a problem because there are ongoing production campaigns in 10_2_X. I think we have nothing calling this code in MINIAOD production, but could you please verify this point?

intrepid42:
This is not run in **MINI**AOD produciton. Only at analysis level and **nano**Aod.

peruzzim:
good, then I would say you can backport it immediately - thanks

PR #26633, opened 2019-05-04
Use explicit type for lambda in CaloSD

Dr15Jones:
#### PR description:

This works around a compiler crash in gcc 9.

#### PR validation:

This compiles fine using gcc 9.


Dr15Jones:
please test

civanch:
+1

fabiocos:
+1

fabiocos:
@Dr15Jones for my education, is there a specific reason why the "auto" statement should not work? Obviously this seems related to a modified behaviour of the compiler, I was just wondering whether there is anything strictly wrong...

Dr15Jones:
My best guess is it is a compiler bug.

fabiocos:
@Dr15Jones ok, thanks

PR #23467, opened 2018-06-04
DeepFlavour Negative Tagger + DeepFlavour 10X training

emilbols:
This PR is concerned with the implementation of the DeepFlavour negative tagger, which is used for estimating the rate of light jets being tagged as b jets. The goal of the negative tagger is to have the same light jet distribution as the full tagger, but filter out as many b-jets as possible at high discriminator values. This is done by inverting the sign of all distributions that are symmetric under sign inversion for light jets, but not for b-jets.

A description of the negative taggers in general can be found in these presentation at the BTV meeting [1] and at the RECO meeting [2].

The DeepFlavour negative tagger distribution is shown in [3]. The negative tagger discriminator values are shown from -1 to 0, and the full tagger is shown from 0 to 1. Red is b-jets, blue and light blue are light jets, and green is c-jets.

Originally this PR was a part of PR https://github.com/cms-sw/cmssw/pull/23206 , but it was filtered out due to merge conflicts. The review comments by @slava77 left in that PR has been dealt with here.

PR https://github.com/cms-sw/cmssw/pull/23206 also introduced the 94X training which was made a while back. While that PR was ongoing we produced a 10X training, which performs better, so in this PR we also add a new DeepFlavour model trained on 10X MC samples. It is therefore required that the PR  https://github.com/cms-data/RecoBTag-Combined/pull/14  is merged first. The ROC curves can be seen in [4]. Here the dashed line is c vs b and the full line is l vs b.

[1] https://indico.cern.ch/event/713923/contributions/2938152/attachments/1620844/2578707/NegativeTagger.pdf

[2] https://indico.cern.ch/event/732060/contributions/3020340/attachments/1660464/2659922/NegativeTaggerPR.pdf

[3]
deepflavournegativetagger

[4]
rocs

PR #26634, opened 2019-05-04
drop provenance checks when not needed in DeepFlavourTagInfoProducer

slava77:
do provenance check only once per collection retrieval and only when the negative tagging is running.

in a recent performance inspection the provenance check was found to dominate the CPU use in the ```DeepFlavourTagInfoProducer::produce``` calls
https://slava77sk.web.cern.ch/slava77sk/reco/cgi-bin/igprof-navigator/CMSSW_10_6_0_pre2-sign1069.136.8311.step2.1000.pp/487

This check is used only in the analysis mode with pfNegativeDeepFlavourTagInfos. 
This PR moves the check out of the loop to be done just once per event and only with the ```flip``` flag set for running negative tagging.

### DIFF
--- a/RecoBTag/FeatureTools/plugins/DeepFlavourTagInfoProducer.cc
+++ b/RecoBTag/FeatureTools/plugins/DeepFlavourTagInfoProducer.cc
@@ -152,6 +152,15 @@ void DeepFlavourTagInfoProducer::produce(edm::Event& iEvent, const edm::EventSet
 
   edm::Handle<ShallowTagInfoCollection> shallow_tag_infos;
   iEvent.getByToken(shallow_tag_info_token_, shallow_tag_infos);
+  double negative_cut = 0;//used only with flip_
+  if (flip_){//FIXME: Check if can do even less often than once per event
+    const edm::Provenance *prov = shallow_tag_infos.provenance();
+    const edm::ParameterSet& psetFromProvenance = edm::parameterSet(*prov);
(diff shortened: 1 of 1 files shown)


slava77:
@emilbols @ferencek 
this was added in #23467.

slava77:
@cmsbuild please test

RecoBTag/FeatureTools/plugins/DeepFlavourTagInfoProducer.cc
```diff
@@ -152,6 +152,15 @@ void DeepFlavourTagInfoProducer::produce(edm::Event& iEvent, const edm::EventSet
 
   edm::Handle<ShallowTagInfoCollection> shallow_tag_infos;
   iEvent.getByToken(shallow_tag_info_token_, shallow_tag_infos);
+  double negative_cut = 0;//used only with flip_
+  if (flip_){//FIXME: Check if can do even less often than once per event
```
Dr15Jones: The framework guarantees that provenance is stable over an entire Run

slava77: nice to know.
My guess was bordering on some IOV possible boundaries.
It sounds like the beginRun can be the more safe and complete version of this code.
Thanks.

slava77:
+1

for https://github.com/cms-sw/cmssw/pull/26634	3d71f1c

fabiocos:
+1

PR #26652, opened 2019-05-06
[HeaderCheck][GenIfc] GeneratorInterface/EvtGenInterface

mrodozov:
- Add dep on evtgen external
- Change header path

### DIFF
--- a/GeneratorInterface/EvtGenInterface/BuildFile.xml
+++ b/GeneratorInterface/EvtGenInterface/BuildFile.xml
@@ -4,6 +4,7 @@
 <use name="hepmc"/>
 <use name="clhep"/>
 <use name="heppdt"/>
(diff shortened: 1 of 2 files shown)


mrodozov:
please test

qliphy:
+1

fabiocos:
+1

smuzaffar:
@qliphy, @perrozzi, @efeyazgan  looks like the header file (  GeneratorInterface/EvtGenInterface/interface/EvtGenLHCInterface.h) is not used any where in cmssw. I would suggest that instead of fixing and maintaining unused code we should delete it.

alberto-sanchez:
Indeed it is a left over of the previous implementation of EvtGen interface.
I can work on this later today.

PR #26655, opened 2019-05-06
JETMET LumiFlag

iatakisi:
#### PR description:
we set lumiflag for only one plot which is MetSig

#### PR validation:
it shouldn't affect any normal DQM

### DIFF
--- a/DQMOffline/JetMET/src/METAnalyzer.cc
+++ b/DQMOffline/JetMET/src/METAnalyzer.cc
@@ -1772,7 +1772,7 @@ void METAnalyzer::fillMonitorElement(const edm::Event& iEvent, std::string DirNa
     myLuminosityBlock = iEvent.luminosityBlock();
     //
     
-    if (subFolderName!=""){
+    if (!subFolderName.empty()){
       DirName = DirName +"/"+subFolderName;
     }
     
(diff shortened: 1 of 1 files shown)


andrius-k:
please test

schneiml:
@smuzaffar Question regarding the bot output for this PR ( https://cmssdt.cern.ch/SDT/jenkins-artifacts/baseLineComparisons/CMSSW_10_6_X_2019-05-06-1100+26655/31517/ ):

It seems that for some workflows, the workflow ran (there are log files) but there is no output directory. In this case I was looking for 136.8311 (JetHT data), but it also affects a few others. Is that expected behaviour?

schneiml:
@fabiocos We'd like to have this small change in 10_6 to get per-lumisection saved plots for JETMET. They requested lumisection information for some quantities, and if we can't get it for all of DQM, we should have at least these plots.

andrius-k:
+1

fabiocos:
@schneiml I would say that the output directory should be there, @smuzaffar could confirm.
The PR in itself looks ok to be merged, I will just launch another test to see whether the comparison in the bot is triggered

fabiocos:
please test

fabiocos:
+1

PR #26656, opened 2019-05-06
[HeaderCheck][RECO1] Cleanup headers to avoid header inconsistency errors

smuzaffar:
- removed obsolete headers
- all of these are already marked to be delete
  -  RecoTracker/Record/interface/Records.h is not used by any thing


smuzaffar:
please test

slava77:
it would be better to delete obviously obsolete files separately from those that contain real code which may be used in some analysis and could require a revert/recovery

perrotta:
+1
- Technical: 
  - all removed headers are obsolete
  - RecoTracker/Record/interface/Records.h is never used, and it is also wrong as it refers to a non-existing "RecoTracker/Records" (with final "s") package

fabiocos:
+1

PR #26184, opened 2019-03-14
Removed unused RecoParticleFlow/PFClusterTools code

guitargeek:
#### PR description:

I have not idea how controversial this PR will be. In my opinion, the first step to bring the ParticleFlow code under our control would be a serious refactor, which I already contributed to a little bit last year.

Sadly, working on the PF code is not fun at all so I didn't do that anymore, because if you try to improve something you usually break some old code. You'd also have to change this old code that is not even used anymore, which means you just waste your time. Not to mention the cognitive overhead that this old code induces if you want to understand the PF algorithms. This situation can be improved by sweeping through RecoParticleFlow and removing this unused code.

So that's what I try to do here: I removed as many files form RecoParticleFlow/PFClusterTools as possible if you still want CMSSW to compile and the matrix tests to pass.

This code included no plugins which were not used in it's own tests. Please, if you know what this code was used for or if it's still necessary, I am happy to learn about it :)

#### PR validation:

CMSSW still compiles and local matrix tests pass.

PR #26670, opened 2019-05-06
 DataFormats/ParticleFlowReco - remove unused data formats

guitargeek:
#### PR description:

Follow up on https://github.com/cms-sw/cmssw/pull/26184, where a lot of the `PFClusterTools` code was removed. Actually, the removed code relied on several DataFormats with are now obsolete and should be removed as well.

By the way, are there any CMSSW policies on how we treat the DataFormats? Is their removal allowed, or should they stick in the release forever to ensure backwards compatibility of ROOT files?

#### PR validation:

Code compiles and local matrix tests pass.


slava77:
@cmsbuild please test

slava77:
> By the way, are there any CMSSW policies on how we treat the DataFormats? Is their removal allowed, or should they stick in the release forever to ensure backwards compatibility of ROOT files?

We can remove DF classes including dictionaries for transient DFs (something that was never saved in production jobs).
It's worth to go back in history and check if the data was produced in the past.
@guitargeek do you happen to know which ones were used in the standard PF sequences?

For the  DFs that were saved in the past, but are no longer expected to be consumed, we can make a dummy dictionary to allow reading the old files. For this:  add a ```struct RemovedXYZClassName {} ;``` to classes.h.

Only if the old data is still needed for analysis it is worth to keep the DF implementation.

slava77:
@cmsbuild please test

guitargeek:
Hi @slava77, sorry I don't happen to know anything about how the classes were used, as they are from the "Snapshot of CMSSW_6_2_0_pre8" commit so it's hard to find some information in an associated PR or something. Maybe @hatakeyamak, @jpata or @bendavid are familiar with it?

The dummy dictionary sounds like a good option to me anyhow, as it probably does not add any compile time whatsoever, no? Then I would not see any drawback.

slava77:
On 5/16/19 3:17 PM, Jonas Rembser wrote:
> The dummy dictionary sounds like a good option to me anyhow, as it
> probably does not add any compile time whatsoever, no? Then I would not
> see any drawback.
> 

I think that it will be OK to add them only after we see problems in I/O.
IIUC, none of the regular production samples have these classes persisted.

slava77:
+1

for https://github.com/cms-sw/cmssw/pull/26670	e140507
- this is essentially a technical update to remove unused classes
- jenkins tests pass

fabiocos:
+1

PR #26679, opened 2019-05-07
Restore fed error treatment for phase0 pixel detector

veszpv:
Resubmitting original [PR26606](https://github.com/cms-sw/cmssw/pull/26606/) without the format checks.

PR description:

The FED raw data format for phase 0 and phase 1 detectors are essentially the same with small differences for the bit-packing of the pixel hit information in layer 1. There is a difference also in the error assignment by the front-end (*). The former case has been treated in the PixelDataFormatter by setting the 'phase1' variable through python configuration files. The error decoding class, called ErrorChecker, however was not made aware of the difference, instead, was just simply updated to phase 1.
This PR fixes that: a base class is defined for the ErrorChecker, the present default code is kept with the same name deriving from this base, and a new derived class ErrorCheckerPhase0 is introduced to restore the phase 0 treatment. The proper error checker is instantiated by the PixelDataFormatter based on the value of 'phase1', so no change needs to be made in the python configurations.

FED error 25: in phase 1, FED error type 25 has been reinterpreted and used by the FED firmware to signal "dead" links. It is used to flag stuck TBMs. This error has a totally different meaning in phase 0. In this PR, the special treatment of this error type by raw-to-digi and digi-to-raw is cancelled for phase 0.

(*) Few error codes have different interpretation; and in phase 0 FPix, the error association with detId is ambiguous.
PR validation:

During the reconstruction of phase 0 data, when a rare type 25 error was encountered, the tracking code crashed. This PR fixes that. The FED25 error packing in simulation, and unpacking in simulation and data for phase 1 should be unchanged, are being validated by @tsusa.

@tsusa @mmusich @dkotlins

### DIFF
--- a/EventFilter/SiPixelRawToDigi/plugins/SiPixelDigiToRaw.cc
+++ b/EventFilter/SiPixelRawToDigi/plugins/SiPixelDigiToRaw.cc
@@ -136,7 +136,7 @@ void SiPixelDigiToRaw::produce( edm::StreamID, edm::Event& ev,
 
   PixelDataFormatter::BadChannels badChannels;
   edm::Handle<PixelFEDChannelCollection> pixelFEDChannelCollectionHandle;
-  if (ev.getByToken(theBadPixelFEDChannelsToken, pixelFEDChannelCollectionHandle)){
+  if (usePhase1==true && ev.getByToken(theBadPixelFEDChannelsToken, pixelFEDChannelCollectionHandle)){
     for (auto const& fedChannels: *pixelFEDChannelCollectionHandle) {
       PixelDataFormatter::DetBadChannels detBadChannels;
       for(const auto& fedChannel: fedChannels) {
--- a/EventFilter/SiPixelRawToDigi/plugins/SiPixelDigiToRaw.cc
+++ b/EventFilter/SiPixelRawToDigi/plugins/SiPixelDigiToRaw.cc
@@ -136,7 +136,7 @@ void SiPixelDigiToRaw::produce( edm::StreamID, edm::Event& ev,
 
   PixelDataFormatter::BadChannels badChannels;
   edm::Handle<PixelFEDChannelCollection> pixelFEDChannelCollectionHandle;
-  if (usePhase1==true && ev.getByToken(theBadPixelFEDChannelsToken, pixelFEDChannelCollectionHandle)){
+  if (usePhase1 && ev.getByToken(theBadPixelFEDChannelsToken, pixelFEDChannelCollectionHandle)){
     for (auto const& fedChannels: *pixelFEDChannelCollectionHandle) {
       PixelDataFormatter::DetBadChannels detBadChannels;
       for(const auto& fedChannel: fedChannels) {
--- a/EventFilter/SiPixelRawToDigi/src/ErrorCheckerPhase0.cc
+++ b/EventFilter/SiPixelRawToDigi/src/ErrorCheckerPhase0.cc
@@ -1,0 +1,298 @@
+#include "EventFilter/SiPixelRawToDigi/interface/ErrorCheckerPhase0.h"
+
+#include "CondFormats/SiPixelObjects/interface/SiPixelFrameConverter.h"
+
+#include "DataFormats/DetId/interface/DetId.h"
+#include "DataFormats/SiPixelDetId/interface/PixelSubdetector.h"
+#include "DataFormats/FEDRawData/interface/FEDHeader.h"
+#include "DataFormats/FEDRawData/interface/FEDTrailer.h"
+
+#include "FWCore/MessageLogger/interface/MessageLogger.h"
+
+#include <bitset>
+#include <sstream>
+#include <iostream>
+
+using namespace std;
+using namespace edm;
+using namespace sipixelobjects;
+
+namespace {
+  constexpr int CRC_bits = 1;
+  constexpr int LINK_bits = 6;
+  constexpr int ROC_bits  = 5;
+  constexpr int DCOL_bits = 5;
+  constexpr int PXID_bits = 8;
+  constexpr int ADC_bits  = 8;
+  constexpr int OMIT_ERR_bits = 1;
+  
+  constexpr int CRC_shift = 2;
+  constexpr int ADC_shift  = 0;
+  constexpr int PXID_shift = ADC_shift + ADC_bits;
+  constexpr int DCOL_shift = PXID_shift + PXID_bits;
+  constexpr int ROC_shift  = DCOL_shift + DCOL_bits;
+  constexpr int LINK_shift = ROC_shift + ROC_bits;
+  constexpr int OMIT_ERR_shift = 20;
+  
+  constexpr cms_uint32_t dummyDetId = 0xffffffff;
+  
+  constexpr ErrorCheckerPhase0::Word64 CRC_mask = ~(~ErrorCheckerPhase0::Word64(0) << CRC_bits);
+  constexpr ErrorCheckerPhase0::Word32 ERROR_mask = ~(~ErrorCheckerPhase0::Word32(0) << ROC_bits);
+  constexpr ErrorCheckerPhase0::Word32 LINK_mask = ~(~ErrorCheckerPhase0::Word32(0) << LINK_bits);
+  constexpr ErrorCheckerPhase0::Word32 ROC_mask  = ~(~ErrorCheckerPhase0::Word32(0) << ROC_bits);
+  constexpr ErrorCheckerPhase0::Word32 OMIT_ERR_mask = ~(~ErrorCheckerPhase0::Word32(0) << OMIT_ERR_bits);
+}  
+
+ErrorCheckerPhase0::ErrorCheckerPhase0() {
+
+  includeErrors = false;
+}
+
+void ErrorCheckerPhase0::setErrorStatus(bool ErrorStatus)
+{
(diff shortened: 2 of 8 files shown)


perrotta:
please test

EventFilter/SiPixelRawToDigi/src/PixelDataFormatter.cc
```diff
@@ -100,12 +100,17 @@ PixelDataFormatter::PixelDataFormatter( const SiPixelFedCabling* map, bool phase
   PXID_mask = ~(~PixelDataFormatter::Word32(0) << PXID_bits);
   ADC_mask  = ~(~PixelDataFormatter::Word32(0) << ADC_bits);
 
+  if (phase1==true) {
```
perrotta:
```C++
  if (phase1) {
```

EventFilter/SiPixelRawToDigi/src/ErrorCheckerPhase0.cc
```diff
@@ -0,0 +1,298 @@
...
+  if LIKELY(errorType<25) return true;
+
+ switch (errorType) {
+    case(25) : {
+     LogDebug("")<<"  invalid ROC=25 found (errorType=25)";
+     errorsInEvent = true;
+     break;
+   }
+   case(26) : {
+     //LogDebug("")<<"  gap word found (errorType=26)";
```
perrotta: If these LogDebug are not neded, you can just remove them

EventFilter/SiPixelRawToDigi/src/ErrorCheckerPhase0.cc
```diff
@@ -0,0 +1,298 @@
...
+     return false;
+   }
+   case(27) : {
+     //LogDebug("")<<"  dummy word found (errorType=27)";
```
perrotta: same here

perrotta (commented):
I repeat here a few comments I already made in the former version of this PR
None of them is really fundamental, but worth being taken into account anyhow before merging this PR

perrotta:
In order not to interfere with this PR, I opened PR #26681 for the removal of the useless ```config_.exists()''' checks in SiPixelRawToDigi.cc.
Please @veszpv fill free to comment in that PR if you have any complain with it.

perrotta:
I have one last comment/request for this PR.

As far as I can see, the two classes ErrorCheckerPhase0 and ErrorChecker only differ in two methods:
- ```checkROC```
- ```errorDetId```

You can avoid a lot of (error prone during maintenance) copy/paste if you move all methods in the base class, and only specialize in the derived ones the two different ones.

Moreover, also ```errorDetId``` is basically identical in the two derived classes: it could also be moved into the base class if you specialize a function which returns ```(DetId(detIdx.rawId).subdetId() == static_cast<int>(PixelSubdetector::PixelBarrel))``` as a bool in Phase0, and just ```true``` in Phase1

veszpv:
@perrotta I would argue exactly the opposite; this is the less error prone approach. Since all the phase 0 data taking has finished (detector disassembled), the phase 0 code may be considered frozen (I doubt if we will ever use it beyond the upcoming UL, actually). At the same time, if we ever need to change anything for phase 1 in the future, we do not need to worry about being constrained by the phase 0 behavior (FED errors are very difficult to validate)... Following the same idea: we do not need to validate this particular PR for all FED errors (phase 0 and 1) either, because in both cases the encoded methods are unchanged.

perrotta:
> @perrotta I would argue exactly the opposite; this is the less error prone approach. Since all the phase 0 data taking has finished (detector disassembled), the phase 0 code may be considered frozen (I doubt if we will ever use it beyond the upcoming UL, actually). At the same time, if we ever need to change anything for phase 1 in the future, we do not need to worry about being constrained by the phase 0 behavior (FED errors are very difficult to validate)... Following the same idea: we do not need to validate this particular PR for all FED errors (phase 0 and 1) either, because in both cases the encoded methods are unchanged.

@veszpv I don't fully agree
Updates to this code can be due to the adaptation to some modified software underneath, and this would definitely benefit of having a single piece of code to touch.
Other updates specific for Phase1/Phase2 or whatsoever, if any, could be specialized even in the future by deriving the specific methods in the derived classes: and this should be done just if needed

perrotta:
@veszpv please let us know whether and when do you intend to follow up with the review comment https://github.com/cms-sw/cmssw/pull/26679#issuecomment-490084185

boudoul:
Hi @veszpv , could you please edit the title to make clear that this is for the Pixel subdetector ? thanks

mmusich:
> @veszpv please let us know whether and when do you intend to follow up with the review comment [#26679 (comment)](https://github.com/cms-sw/cmssw/pull/26679#issuecomment-490084185)

Hi Andrea @perrotta, the TkDPG endorses the current implementation as it minimizes the need for validation (the phase-0 version of the error checking is copied straight off the code that run when the detector was operating 3 years ago). 
We don't see reasons to touch the code in the foreseeable future, but we can promise to commit to systematize it, if there will ever be necessity to act on it.
For the time being, please let's try to proceed with this implementation as it is a bug fix compulsory to be able to process phase-0 data.
Thanks,

Marco & Tanja

veszpv:
Sorry for the slow response, was offline a large part. @boudoul I changed the name.
@perrotta your suggestion makes sense, of course, I would still prefer to keep the code this way in the hope that phase 0 will never need to be updated. But as Marco wrote, we will do the proper restructuring if phase 0 needs an update.

perrotta:
@veszpv @mmusich @tsusa 

I discussed the issue also with @slava77 , and we both agree that we can avoid complicate future maintenance if there are not such large chunks of copy-pasted code. Therefore, we strongly support having the common methods implemented in the base class, and only the specializations included in the derived ones.

On the other hand, if you think that it may take long time for you to implement it and you don't want to do it now, we can even stick with this implementation for the UL release, but a github issue will be opened to have it fixed for the next 11_X cycles (i.e. kind of next week, once 10_6_0 will be cut and the queue for 11_0_X will be opened).

Therefore, please let us know whether you plan to update it since now, or if you prefer to defer it to the next cycle: I will open a github issue for it in that second  case, and restart the tests here.

veszpv:
@perrotta Sorry, I will not have time to work on this. (Note that by far the most time-consuming part of modifying the error decoding is the validation, not the code-change.)

perrotta:
please test

perrotta:
Issue #26731 have been opened to keep track of the fixes for the removal of the duplicated code that will be implemented in the next cycle after the UL release

EventFilter/SiPixelRawToDigi/plugins/SiPixelDigiToRaw.cc
```diff
@@ -136,7 +136,7 @@ void SiPixelDigiToRaw::produce( edm::StreamID, edm::Event& ev,
 
   PixelDataFormatter::BadChannels badChannels;
   edm::Handle<PixelFEDChannelCollection> pixelFEDChannelCollectionHandle;
-  if (ev.getByToken(theBadPixelFEDChannelsToken, pixelFEDChannelCollectionHandle)){
+  if (usePhase1 && ev.getByToken(theBadPixelFEDChannelsToken, pixelFEDChannelCollectionHandle)){
```
perrotta: Ok, bad channels are there only for Phase1

perrotta:
@veszpv @tsusa : in the PR description you anticipate that there is an ongoing validation for this PR. Please post here a link to the results, or just a short summary of them, so that we can conclude the review and finally sign this off for reco.

tsusa:
@perrotta  @slava The validation is still ongoing. I will post the plots as soon as they are done [ETA tomorrow]

davidlange6:
if you want to reach Slava, the correct tag is @slava77....

tsusa:
Thanks @davidlange6!

fabiocos:
@tsusa any news about the validation?

tsusa:
The validation plots are uploaded. For Phase 1 detector shown are the occupancy maps w/ and w/o changes implemented in the PR and with BadFED simulation switched on and off. For Phase0 shown are the plots with BadFED simulation switched off as FED25 error did not exist in Phase 0 detector.

[pr26679_validation.pdf](https://github.com/cms-sw/cmssw/files/3170115/pr26679_validation.pdf)

mmusich:
@perrotta @fabiocos what's the plan for inclusion of this PR? 
I see that 10.6.0 has been already built: as a reminder this is a mandatory fix to process phase-0 data (also included in https://github.com/cms-sw/cmssw/issues/26721).
Does this imply that a backport will be needed?

perrotta:
+1
- ErrorChecker specialized for Phase0 and Phase1 separately, so that the original behaviour for Phase0 can be recovered by simply relying on the  "(use)phase1" switches already present in the code
- Jenkins tests show no difference: validation has been provided for both Phase0 and Phase1 configurations in https://github.com/cms-sw/cmssw/pull/26679#issuecomment-491596233

fabiocos:
+1

fabiocos:
@mmusich only the externals of 10_6_0 have been pre-built. In any case at this point there is pressure to start the GEN-SIM part of production, where this PR is irrelevant

fabiocos:
anyway, as this is now merged, unless problems appear in next IB it will be part of 10_6_0

mmusich:
> anyway, as this is now merged, unless problems appear in next IB it will be part of 10_6_0

@fabiocos thanks!

PR #26680, opened 2019-05-07
print statement fixes and preserve integer division with python3

davidlange6:
Fixes failing unit test in DEVEL IBs - the Validation/Geometry unit tests will still fail as our root build does not yet have pyroot working with python3.

### DIFF
--- a/Validation/Geometry/test/runP_Tracker_cfg.py
+++ b/Validation/Geometry/test/runP_Tracker_cfg.py
@@ -2,7 +2,7 @@
 #
 # for t in {'BeamPipe','Tracker','PixBar','PixFwdMinus','PixFwdPlus','TIB','TOB','TIDB','TIDF','TEC','TkStrct','InnerServices'}; do cmsRun runP_Tracker_cfg.py geom=XYZ label=$t >& /dev/null &; done
 
-
+from __future__ import print_function
 import FWCore.ParameterSet.Config as cms
 from FWCore.ParameterSet.VarParsing import VarParsing
 import sys, re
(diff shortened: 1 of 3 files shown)


davidlange6:
please test

andrius-k:
+1

Validation/Geometry/test/runP_Tracker_cfg.py
```diff
@@ -62,9 +62,9 @@
     )
 
 if options.label not in _ALLOWED_LABELS:
-    print "\n*** Error, '%s' not registered as a valid components to monitor." % options.label
-    print "Allowed components:", _ALLOWED_LABELS
-    print
+    print("\n*** Error, '%s' not registered as a valid components to monitor." % options.label)
+    print("Allowed components:", _ALLOWED_LABELS)
+    print()
```
vargasa: This `print` is not needed

cvuosalo:
+1

Validation/Geometry/test/runP_Tracker_cfg.py
fabiocos: I would say it can be fixed in another round

fabiocos:
+1

Validation/Geometry/test/runP_Tracker_cfg.py
davidlange6: this PR doesn't add the printout - it preserves current behavior.

fabiocos: @davidlange6 yes, I agree, if the maintainers want to modify it it can be done in another round, no point in stopping the PR already approved for this

PR #26684, opened 2019-05-07
Updated runPromptReco.py

andresfelquintero:

### DIFF
--- a/Configuration/DataProcessing/test/RunPromptReco.py
+++ b/Configuration/DataProcessing/test/RunPromptReco.py
@@ -1,10 +1,8 @@
 #!/usr/bin/env python
 """
 _RunPromptReco_
-
 Test wrapper to generate a reco config and actually push it into cmsRun for
 testing with a few input files etc from the command line
-
 """
 from __future__ import print_function
 
@@ -14,7 +12,6 @@ import traceback
 import pickle
 
 from Configuration.DataProcessing.GetScenario import getScenario
-
 
 
 class RunPromptReco:
@@ -34,6 +31,7 @@ class RunPromptReco:
         self.dqmSeq = None
         self.setRepacked = False
         self.isRepacked = False
+        self.numCores = None
 
     def __call__(self):
         if self.scenario == None:
@@ -45,6 +43,8 @@ class RunPromptReco:
         if self.inputLFN == None:
             msg = "No --lfn specified"
             raise RuntimeError(msg)
+        if self.numCores == None:
+            msg = "No --numCores specified"
 
         try:
             scenario = getScenario(self.scenario)
@@ -92,6 +92,7 @@ class RunPromptReco:
 
                 if self.alcaRecos:
                     kwds['skims'] = self.alcaRecos
+
                 if self.PhysicsSkims:
                     kwds['PhysicsSkims'] = self.PhysicsSkims
 
@@ -100,6 +101,9 @@ class RunPromptReco:
 
                 if self.setRepacked:
                     kwds['repacked'] = self.isRepacked
+                
+                if self.numCores:
+                    kwds['nThreads'] = self.numCores
 
             process = scenario.promptReco(self.globalTag, **kwds)
 
@@ -141,12 +145,11 @@ class RunPromptReco:
 
 
 if __name__ == '__main__':
-    valid = ["scenario=", "reco", "aod", "miniaod","dqm", "dqmio", "no-output",
+    valid = ["scenario=", "reco", "aod", "miniaod","dqm", "dqmio", "no-output", "numCores=", 
              "global-tag=", "lfn=", "alcarecos=", "PhysicsSkims=", "dqmSeq=", "isRepacked", "isNotRepacked" ]
     usage = \
 """
 RunPromptReco.py <options>
-
 Where options are:
  --scenario=ScenarioName
  --reco (to enable RECO output)
@@ -161,15 +164,11 @@ Where options are:
  --alcarecos=alcareco_plus_seprated_list
  --PhysicsSkims=skim_plus_seprated_list
  --dqmSeq=dqmSeq_plus_separated_list
-
+ --numCores=Number_of_cores_used
 Example:
-
 python RunPromptReco.py --scenario=cosmics --reco --aod --dqmio --global-tag GLOBALTAG --lfn=/store/whatever --alcarecos=TkAlCosmics0T+MuAlGlobalCosmics
-
 python RunPromptReco.py --scenario=pp --reco --aod --dqmio --global-tag GLOBALTAG --lfn=/store/whatever --alcarecos=TkAlMinBias+SiStripCalMinBias
-
 python RunPromptReco.py --scenario=ppEra_Run2_2016 --reco --aod --dqmio --global-tag GLOBALTAG --lfn=/store/whatever --alcarecos=TkAlMinBias+SiStripCalMinBias --PhysicsSkims=@SingleMuon
-
 """
     try:
         opts, args = getopt.getopt(sys.argv[1:], "", valid)
@@ -196,6 +195,8 @@ python RunPromptReco.py --scenario=ppEra_Run2_2016 --reco --aod --dqmio --global
             recoinator.writeDQMIO = True
         if opt == "--no-output":
             recoinator.noOutput = True
+        if opt == "--numCores":
+            recoinator.numCores = arg
         if opt == "--global-tag":
             recoinator.globalTag = arg
         if opt == "--lfn" :
@@ -212,6 +213,5 @@ python RunPromptReco.py --scenario=ppEra_Run2_2016 --reco --aod --dqmio --global
         if opt == "--isNotRepacked":
             recoinator.setRepacked = True
             recoinator.isRepacked = False
-
 
     recoinator()
--- a/Configuration/DataProcessing/test/RunPromptReco.py
+++ b/Configuration/DataProcessing/test/RunPromptReco.py
@@ -43,9 +43,6 @@ class RunPromptReco:
         if self.inputLFN == None:
             msg = "No --lfn specified"
             raise RuntimeError(msg)
-        if self.nThreads == None:
-            msg = "No --nThreads specified"
-            raise RuntimeError(msg)
 
         try:
             scenario = getScenario(self.scenario)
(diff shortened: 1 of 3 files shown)


slava77:
@cmsbuild please test

slava77:
@andresfelquintero 
please add a link to the T0 JIRA ticket in the PR description for better self-documentation

andresfelquintero:
> @andresfelquintero
> please add a link to the T0 JIRA ticket in the PR description for better self-documentation

It was added to the description as suggested.

davidlange6:
Two comments from my side
- nThreads would be a more consistent name for this option
- a default of 1 is likely better than requiring this option to be set (that way, the examples in the help will not be broken by this PR)

fabiocos:
@smuzaffar interesting situation, the bot tells us that the PR was updated but I see just one commit and no message about force updates

fabiocos:
please test

fabiocos:
please test

smuzaffar:
@fabiocos , looks like github timestamps are messed up. As you can see comments from @davidlange6 and @slava77 are also in future. 
The message from @cmsbuild is also in future but if I see the jenkins log the message was added at

```
The following comment will be made:

Latest commit sha:  9756b20ebe88c9a8c4a6ffb1ec01282bbcace35f
PR update time 2019-05-08 11:48:23
Time UTC: 2019-05-08 11:48:49.274511
```
which is correct timestamp (and matches the jenkins job's time).

andresfelquintero:
> Two comments from my side
> 
> * nThreads would be a more consistent name for this option
> * a default of 1 is likely better than requiring this option to be set (that way, the examples in the help will not be broken by this PR)

So @slava77 , should I change the name of it? For us it is better to use the name numCores because it is the same name of the variable that we obtain from the API, but if this is more useful for you then I will change the name as requested by @davidlange6 .

smuzaffar:
abort

smuzaffar:
please test

smuzaffar:
@slava77 and @davidlange6 can you please add any comment here? looks like cmsbot timestamp is correct now. I would like to confirm that your messages are also in present or not?

fabiocos:
Should we retrigger this test? Now let's see whether I can read this comment

fabiocos:
@smuzaffar the problem seems solved in the sense that I see the message at a reasonable time, but the messages that were moved in the future are still seen in the future, this probably will stay right?

davidlange6:
here is a comment

slava77:
> So @slava77 , should I change the name of it? For us it is better to use the name numCores because it is the same name of the variable that we obtain from the API, but if this is more useful for you then I will change the name as requested by @davidlange6 .

For my reference, please provide a pointer to the T0/WMAgent code where this is used as "numCores".
Thank you.

slava77:
On 5/8/19 9:17 AM, Fabio Cossutti wrote:
> the problem seems solved in the sense that I see the message at a
> reasonable time, but the messages that were moved in the future are
> still seen in the future, this probably will stay right?
> 

my messages are however sometimes moved into the past (at least in my
rendering/display).
The most recent ones are earlier than what I've sent properly earlier.
I'm trying to rely on what I get in my inbox instead of the web right
now to read the messages.

slava77:
this is what's showing up in my browser (taking the last few)

Screen Shot 2019-05-08 at 10 04 24 AM

andresfelquintero:
> > So @slava77 , should I change the name of it? For us it is better to use the name numCores because it is the same name of the variable that we obtain from the API, but if this is more useful for you then I will change the name as requested by @davidlange6 .
> 
> For my reference, please provide a pointer to the T0/WMAgent code where this is used as "numCores".
> Thank you.

Hello. You were right, there was no code set for that name, I just set that name for my wrapper based on the outcome of this service. https://cmsweb.cern.ch/t0wmadatasvc/prod/reco_config?run=326607 So I will change the name as @davidlange6 suggested.

andresfelquintero:
Is it ok for this pull request has two commits? or should I comprise them into a single one? I don't know how you handle them.

slava77:
> > > So @slava77 , should I change the name of it? For us it is better to use the name numCores because it is the same name of the variable that we obtain from the API, but if this is more useful for you then I will change the name as requested by @davidlange6 .
> > 
> > 
> > For my reference, please provide a pointer to the T0/WMAgent code where this is used as "numCores".
> > Thank you.
> 
> Hello. You were right, there was no code set for that name, I just set that name for my wrapper based on the outcome of this service. https://cmsweb.cern.ch/t0wmadatasvc/prod/reco_config?run=326607 So I will change the name as @davidlange6 suggested.

but even here the parameter is apparently called "multicore", not "numCores".



> Is it ok for this pull request has two commits? or should I comprise them into a single one? I don't know how you handle them.

either way is OK.

slava77:
@cmsbuild please test

andresfelquintero:
So the checks have passed, do you have any other comments? @slava77 @davidlange6

Configuration/DataProcessing/test/RunPromptReco.py
```diff
@@ -45,6 +43,8 @@ def __call__(self):
         if self.inputLFN == None:
             msg = "No --lfn specified"
             raise RuntimeError(msg)
+        if self.nThreads == None:
+            msg = "No --nThreads specified"
```
fabiocos: @andresfelquintero I do not see the purpose of this statement, if no action or printout is added, the "msg" string is not used as it is now

Configuration/DataProcessing/test/RunPromptReco.py
```diff
@@ -196,6 +195,8 @@ def __call__(self):
             recoinator.writeDQMIO = True
         if opt == "--no-output":
             recoinator.noOutput = True
+        if opt == "--nThreads":
+            recoinator.nThreads = arg
```
fabiocos: @andresfelquintero passing the option is not enough, it must be used within recoinator (which is not the case now)

fabiocos:
@andresfelquintero @slava77 @davidlange6 as far as I can see this PR is so far dummy, as in https://cmssdt.cern.ch/lxr/source/Configuration/DataProcessing/python/Reco.py#0045 the passed argument is used nowhere now. Adding a statement like:

```
        if 'nThreads' in args:
            options.nThreads=args['nThreads']
```

passes the argument to ConfigBuilder. Still when I have tried this I am puzzled by the fact that the option seems to disappear, even if in https://cmssdt.cern.ch/lxr/source/Configuration/Applications/python/ConfigBuilder.py#2224 the lines are properly added (verified by printing the content of self.pythonCfgCode), and this is still true at the end of the ConfigBuilder prepare function

fabiocos:
ok, in cmsDriver the pythonCfgCode is dumped into a file, if I do the same within Reco.py I get the proper configuration dumped. But if I print process.options the numberOfThreads option is just not there

slava77:
> @andresfelquintero @slava77 @davidlange6 as far as I can see this PR is so far dummy, as in https://cmssdt.cern.ch/lxr/source/Configuration/DataProcessing/python/Reco.py#0045 the passed argument is used nowhere now. Adding a statement like:

you are right.
I confused myself with some other options "magically" appearing in the ConfigBuilder args list.
I think that the Reco.py should be updated.

andresfelquintero:
I just edited Reco.py as suggested and also created the error message on RunPromptReco.py. Thanks for the feedback.

Configuration/DataProcessing/test/RunPromptReco.py
fabiocos: @andresfelquintero wait, in this case the comment of @davidlange6 is valid, you cannot have ```None``` as default value, but it should be 1, as in https://cmssdt.cern.ch/lxr/source/Configuration/Applications/python/ConfigBuilder.py#0087
otherwise you always force the user to specify explicitly the option

slava77:
@cmsbuild please test

fabiocos:
@andresfelquintero general comment: the express case should be treated in the same way as prompt reco, in general we are having even more DQM troubles in express than in prompt...

Configuration/DataProcessing/test/RunPromptReco.py
andresfelquintero: so @slava77 do you think it is better to have a default value? or shall we let that input as a completely optional one?

andresfelquintero:
> @andresfelquintero general comment: the express case should be treated in the same way as prompt reco, in general we are having even more DQM troubles in express than in prompt...

That is correct, shall I change Express as well on this pull request or should I open another one? I was just changing this file because my wrapper is working for PromptReco only, but I will create another one for Express as soon as we confirm that everything works perfectly with the first wrapper.

Configuration/DataProcessing/test/RunPromptReco.py
slava77: in configurations built with cmsDriver the parameter is optional

andresfelquintero: If this is going to be an optional input for this code, then there is no point in having the error message or the if statement on line 46. Shall I delete it then?

andresfelquintero:
Are there any other tests required to merge this into the master branch?

andresfelquintero:
Hello, any updates from this pull request? Are the commits appropriate?

fabiocos:
@andresfelquintero your PR in itself looks ok to me (at least for the prompt part, express to be added), but it is ineffective because of an apparent bug in ConfigBuilder.py . After
https://cmssdt.cern.ch/lxr/source/Configuration/Applications/python/ConfigBuilder.py#2229
you need to add:
```
            self.process.options.numberOfThreads=cms.untracked.uint32(int(self._options.nThreads))
            self.process.options.numberOfStreams=cms.untracked.uint32(int(self._options.nStreams))            
 self.process.options.numberOfConcurrentLuminosityBlocks=cms.untracked.uint32(int(self._options.nConcurrentLumis))
```
otherwise the option that you pass will never be propagated to the final configuration. I have tested that this works, without changing the behaviour of cmsDriver that does not seem to rely on process as built in prepare() but on the pythonCfgCode content (correctly set).

slava77:
@cmsbuild please test

fabiocos:
@andresfelquintero as far as I am concerned this looks ok, unless @slava77 and @davidlange6 have further comment it could move forward. I think Express should be instrumented in the same way

fabiocos:
+operations

the code provides the desired functionality, fixing a bug in ConfigBuilder when used through DataProcessing (apparently ineffective so far)

fabiocos:
+1

PR #26709, opened 2019-05-09
CondFormats/(L1T,RPC)Objects check in generated Serialization.cc

gartung:
Consider checking in generated Serialization.cc from L1TObjects and RPCObjects. This removes the need to run the serialization script under macOS which requires an incompatible version of clang under Conda builds.


gartung:
@smuzaffar is there a way to disable the autogeneration of Serialization.cc

smuzaffar:
hold
@gartung , no currently there is no way to disable it but it can be implemented.
You current approach is going to break as we will end up duplicate  definitions (one from this new .cc file and one from generated one).

PR #26725, opened 2019-05-09
Fix: dump HGCal rechit center

Mourtz:
The center was not correct for Scintillator rechits due to geometry changes.
Now the center is being dumped with the rest of the geometry.

@rovere @deguio


alja:
+1

fabiocos:
+1

PR #26726, opened 2019-05-09
DQM: Assert non-zero-width range for histograms

schneiml:
#### PR description:

This PR checks for the condition that triggered #26722 at booking time. It seems to be more common than I expected.

This indicates that the problem was not caused (or exposed) by HCAL/Phase2 DQM, but rather that something in ROOT or the framework made it show up.

#### PR validation:

This PR is mainly for validation of the issue discovered in #26722 in the PR tests. It should be save to merge into an IB, though for now, the tests will fail until all subsystem code is fixed.

### DIFF
--- a/DQMServices/Core/src/DQMStore.cc
+++ b/DQMServices/Core/src/DQMStore.cc
@@ -1098,7 +1098,7 @@ MonitorElement*
 DQMStore::book1D(char_string const& name, char_string const& title,
                  int const nchX, double const lowX, double const highX)
 {
-  assert((highX - lowX) > 0);
+  assert((highX - lowX) != 0);
   return book1D_(pwd_, name, new TH1F(name, title, nchX, lowX, highX));
 }
 
@@ -1107,7 +1107,7 @@ MonitorElement*
 DQMStore::book1S(char_string const& name, char_string const& title,
                  int const nchX, double const lowX, double const highX)
 {
-  assert((highX - lowX) > 0);
+  assert((highX - lowX) != 0);
   return book1S_(pwd_, name, new TH1S(name, title, nchX, lowX, highX));
 }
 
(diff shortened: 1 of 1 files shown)


schneiml:
please test

fabiocos:
please test

schneiml:
There are plenty of existing histograms that violate this assertion, not just in Phase2. We should remind people to fix/remove them, but this PR is unlikely to get merged any time soon...

fabiocos:
@schneiml I think that until there is no overall cleaning, you could easily turn this into a test tool, i.e. to have the assert statements conditional to some flag that is off by default, and have a unit test where it is enabled and that can be used for such a purpose

davidlange6:
Surprising since this seems to crash the merge step. Do you have a list from some workflow?




On 10 May 2019, at 11:33, Marcel Schneider <notifications@github.com<mailto:notifications@github.com>> wrote:


There are plenty of existing histograms that violate this assertion, not just in Phase2. We should remind people to fix/remove them, but this PR is unlikely to get merged any time soon...

schneiml:
@davidlange6 see the comments in #26722: The zero-width range makes TH1:: Add dispatch to TH1::Merge (if that is a bug or not is a matter of opinion I guess), and the real problem is that TH1::Merge fails on some histograms. I don't see a clear pattern there; it fails in cases where Add succeeded, and it succeeds in zero-width cases sometimes. The crash is since we hit the Merge path, *and* Merge fails on the specific histogram (`N_HE`).

srimanob:
@davidlange6 

Here is an example workflow,
https://dmytro.web.cern.ch/dmytro/cmsprodmon/workflows.php?prep_id=CMSSW_10_6_0_pre4__UPSG_Std_2023D41noPU-1556059831-ElectronGunPt2To100

PR #26729, opened 2019-05-09
Changes from scram build format-all on IOPool

wddgit:
#### PR description:

The first commit includes only changes from scram build code-format-all and nothing else. This PR includes all the packages in the IOPool subsystem.

There were a few addition changes that showed up in code-checks in the second commit.

#### PR validation:

The only test I performed was checking that the revised packages compile and build.

### DIFF
--- a/IOPool/Common/bin/CollUtil.cc
+++ b/IOPool/Common/bin/CollUtil.cc
@@ -21,7 +21,7 @@
 namespace edm {
 
   // Get a file handler
-  TFile* openFileHdl(std::string const& fname) {
+  TFile *openFileHdl(std::string const &fname) {
     TFile *hdl = TFile::Open(fname.c_str(), "read");
 
     if (nullptr == hdl) {
@@ -37,7 +37,7 @@ namespace edm {
     TList *keylist = hdl->GetListOfKeys();
     TIterator *iter = keylist->MakeIterator();
     TKey *key = nullptr;
-    while ((key = (TKey*)iter->Next())) {
+    while ((key = (TKey *)iter->Next())) {
       TObject *obj = hdl->Get(key->GetName());
       if (obj->IsA() == TTree::Class()) {
         obj->Print();
@@ -47,20 +47,19 @@ namespace edm {
   }
 
   // number of entries in a tree
-  Long64_t numEntries(TFile *hdl, std::string const& trname) {
-    TTree *tree = (TTree*)hdl->Get(trname.c_str());
+  Long64_t numEntries(TFile *hdl, std::string const &trname) {
+    TTree *tree = (TTree *)hdl->Get(trname.c_str());
     if (tree) {
       return tree->GetEntries();
     } else {
-      std::cout << "ERR cannot find a TTree named \"" << trname << "\""
-                << std::endl;
+      std::cout << "ERR cannot find a TTree named \"" << trname << "\"" << std::endl;
       return -1;
     }
   }
 
   namespace {
-    void addBranchSizes(TBranch *branch, Long64_t& size) {
-      size += branch->GetTotalSize(); // Includes size of branch metadata
+    void addBranchSizes(TBranch *branch, Long64_t &size) {
+      size += branch->GetTotalSize();  // Includes size of branch metadata
       // Now recurse through any subbranches.
       Long64_t nB = branch->GetListOfBranches()->GetEntries();
       for (Long64_t i = 0; i < nB; ++i) {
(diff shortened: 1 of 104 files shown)


IOPool/Common/bin/CollUtil.cc
```diff
@@ -112,7 +112,9 @@ namespace edm {
     std::cout << "UUID: " << fid.fid() << std::endl;
   }
 
-  static void preIndexIntoFilePrintEventLists(TFile*, FileFormatVersion const& fileFormatVersion, TTree *metaDataTree) {
+  static void preIndexIntoFilePrintEventLists(TFile *,
+                                              FileFormatVersion const &fileFormatVersion,
+                                              TTree *metaDataTree) {
```
wddgit: In this PR, I tried clang format again on IOPool. The namespace indentation looks OK now, but something still looks different from the prior times I ran this on Framework. For example, why is the symbol & next to fileFormatVersion above instead of adjacent to const?

IOPool/Input/src/ProvenanceAdaptor.cc
```diff
@@ -19,11 +19,9 @@ ProvenanceAdaptor.cc
 #include "DataFormats/Provenance/interface/ProductRegistry.h"
 #include "FWCore/Utilities/interface/Algorithms.h"
 namespace edm {
-  
-  void
-  ProvenanceAdaptor::fixProcessHistory(ProcessHistoryMap& pHistMap,
-				       ProcessHistoryVector& pHistVector) {
-    assert (pHistMap.empty() != pHistVector.empty());
+
+  void ProvenanceAdaptor::fixProcessHistory(ProcessHistoryMap& pHistMap, ProcessHistoryVector& pHistVector) {
```
wddgit: Some places it is aligning the pointer and reference symbols left. Maybe I just do not understand the algorithm it is using to decide ...

wddgit:
please test

wddgit:
clang tidy is adding override keywords where appropriate to the class IOPool/TFileAdaptor/interface/TStorageFactoryFile.h. But the class also uses the ROOT ClassDef macro that is apparently adding virtual functions that override without the override key word. This is leading to a clang compilation failure because apparently using the override keyword is required for all functions or none by the clang compiler. Is there a way to deal with this situation? The idea that occurred to me is to disable clang tidy for this one file, but I don't know how to do that.

Dr15Jones:
There is a version of the macro which also adds the override

wddgit:
Answering my own question about pointer and reference symbol alignment. We use option DerivePointerAlignment set to true from the google style guide. And here is what it means:

If true, analyze the formatted file for the most common alignment of & and *. Pointer and reference alignment styles are going to be updated according to the preferences found in the file. PointerAlignment is then used only as fallback.

In the files I noticed with right alignment, there was mixed alignment before format was run. I didn't count, but probably more right alignment than left. So I guess this means it is all working correctly and as expected with regards to the alignment of the symbols.

@Dr15Jones I will for the other version of the macro and use that. Thanks.

wddgit:
please test

Dr15Jones:
+1

IOPool/Common/bin/CollUtil.cc
smuzaffar: @wddgit , this is because of [a]. 
@fabiocos ,  @davidlange6 , @Dr15Jones should we make `DerivePointerAlignment: false` enforce a unique Alignment e.g. Left?
[a]
```
DerivePointerAlignment: true
DerivePointerAlignment (bool)
If true, analyze the formatted file for the most common alignment of & and *. 
Pointer and reference alignment styles are going to be updated according to
the preferences found in the file. PointerAlignment is then used only as fallback.
```

fabiocos:
@smuzaffar from the conversation above I understand that @wddgit is fine with the present formatting, so I move forward with the integration of this PR

fabiocos:
+1

smuzaffar:
@fabiocos , please see https://github.com/cms-sw/cmssw/pull/26729#pullrequestreview-235667473
So Alignement of `&` and `*` depends on most common alignment of the file. So in one file it could be `Right` and another one it could be `left`
My question is are we fine with this or should be enforce one

fabiocos:
As discussed, i think this is a kind of learning process. So far for this part we are sticking to the Google standard, we may discuss in future Core Software meetings possible further updates as more and more people gets experience with this new formatting

wddgit:
Now that I understand it, I am happy with the current configuration we are using for clang format for the alignment of * and &. If a developer does not like what it is doing for a particular file, then there exists the freedom to go in and manually change it. As long as more than half of them are changed, clang format will respect that.  And as long as one is consistent within a file, clang format will not change anything. I like that.

Issue #26727, opened 2019-05-09
Error in DQM step of Phase 2 relval production


fabiocos:
This issue is meant to track information in a public place about the problem reported today by PdmV experts of a massive failure of DQMIO merge step in phase2 CMSSW_10_6_0_pre4 RelVal samples:

_We observed that most of the Phase 2 relval sample workflows (for e.g. [1, 2]) encounter Error 8022:
Fatal Root Error: @SUB=TH1Merger::CopyMerge
at DQM step, which affected DQMIO output statistics.

Can you please have a look at the issue?

Many thanks,
Siddhesh Sawant, for PdmV+UPSG group_

[1] https://dmytro.web.cern.ch/dmytro/cmsprodmon/workflows.php?prep_id=CMSSW_10_6_0_pre4__UPSG_Std_2023D41PU200_highPU-1556096145-TTbar_14TeV <https://dmytro.web.cern.ch/dmytro/cmsprodmon/workflows.php?prep_id=CMSSW_10_6_0_pre4__UPSG_Std_2023D41PU200_highPU-1556096145-TTbar_14TeV>

[2] https://dmytro.web.cern.ch/dmytro/cmsprodmon/workflows.php?prep_id=CMSSW_10_6_0_pre4__UPSG_Std_2023D41noPU-1556059831-ElectronGunPt2To100 <https://dmytro.web.cern.ch/dmytro/cmsprodmon/workflows.php?prep_id=CMSSW_10_6_0_pre4__UPSG_Std_2023D41noPU-1556059831-ElectronGunPt2To100>

[3] Error 8022
https://cms-unified.web.cern.ch/cms-unified/report/ssawant_RVCMSSW_10_6_0_pre4ElectronGunPt2To100__2023D41noPU_190424_005155_567#ERROR <https://cms-unified.web.cern.ch/cms-unified/report/ssawant_RVCMSSW_10_6_0_pre4ElectronGunPt2To100__2023D41noPU_190424_005155_567#ERROR>

Full Job log
[...]

PR #26738, opened 2019-05-10
Import the precise checks done by ROOT to Add histograms.

schneiml:
This is more restrictive than what we used before, but should prevent ROOT from attempting to call Merge() (which mysteriously fails in some cases, causing a FatalRootError on CMSSW side).
The histograms that are now no longer merged seem to be meaningless either ways (0-0 X range).

This PR replaces #26722, which on closer investigation seems too dangerous. It implements the frist alternative listed there, which means we'll need to keep the checks in sync whenever ROOT changes.

It is still not clear why we can't use `TH1::Merge`, but my tests showed that it fails in some cases, even some where `Add` worked. Since semantically we don't really want a `Merge` (the histograms should really have the same "shape" when we merge them), I think it is a good idea to restrict everything to the cases handled by `TH1::Add`.

#### PR validation:

Works on the sample from issue #26727 . No other validation so far.

### DIFF
--- a/DQMServices/FwkIO/plugins/DQMRootSource.cc
+++ b/DQMServices/FwkIO/plugins/DQMRootSource.cc
@@ -71,7 +71,117 @@ namespace {
     //std::cout <<"create: hist size "<<iName <<" "<<iHist->GetEffectiveEntries()<<std::endl;
     return iStore.book1D(iName, iHist);
   }
-  //NOTE: the merge logic comes from DataFormats/Histograms/interface/MEtoEDMFormat.h
+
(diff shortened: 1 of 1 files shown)


andrius-k:
please test

fabiocos:
please test

fabiocos:
@schneiml so what is ultimately the plan here? As we see ROOT 6.14 behaves differently in this respect

schneiml:
@fabiocos I'll close it for now, since it is not really needed. I plan to integrate these changes in the bigger DQMStore restructuring coming (hopefully) soon.

If we decide to not go with the big restructuring, we can come back to this one.

PR #26740, opened 2019-05-10
DD4hep: Material Conversion

ianna:
#### PR description:

* Use material properties from an XML file definition if they are different then a default ROOT material properties

#### PR validation:
```
cmsRun SimG4Core/DD4hepGeometry/test/python/testG4Geometry.py
```

no back port is needed

### DIFF
--- a/DetectorDescription/DDCMS/plugins/DDDefinitions2Objects.cc
+++ b/DetectorDescription/DDCMS/plugins/DDDefinitions2Objects.cc
@@ -445,30 +445,41 @@ template <> void Converter<DDLElementaryMaterial>::operator()(xml_h element) con
   if(  nullptr == mat ) {
     const char* matname = nam.c_str();
     double density      = xmat.density();
-    double atomicNumber = xmat.attr<double>(DD_CMU(atomicNumber));
+    int atomicNumber = xmat.attr<int>(DD_CMU(atomicNumber));
+    double atomicWeight = xmat.attr<double>(DD_CMU(atomicWeight));
     TGeoElementTable* tab = mgr.GetElementTable();
     TGeoMixture*      mix = new TGeoMixture(nam.c_str(), 1, density);
     TGeoElement*      elt = tab->FindElement(xmat.nameStr().c_str());
 
     printout(ns.context()->debug_materials ? ALWAYS : DEBUG, "DD4CMS",
-             "+++ Converting material %-48s  Density: %.3f.",
-             ('"'+nam+'"').c_str(), density);
+             "+++ Converting material %-48s  Atomic weight %.3f, Atomic number %u, Density: %.3f.",
+             ('"'+nam+'"').c_str(), atomicWeight, atomicNumber, density);
 
-    if( !elt )  {
-      printout(WARNING,"DD4CMS",
-               "+++ Converter<ElementaryMaterial> No element present with name:%s  [FAKE IT]",
-               matname);
-      int n = int(atomicNumber/2e0);
-      if( n < 2 ) n = 2;
-      elt = new TGeoElement(xmat.nameStr().c_str(),"CMS element",n,atomicNumber);
+    bool newMatDef = false;
+    
+    if(elt) {
+      // A is Mass of a mole in Geant4 units for atoms with atomic shell
+      printout(ns.context()->debug_materials ? ALWAYS : DEBUG, "DD4CMS",
+	       "    ROOT definition of %-50s Atomic weight %.3f, Atomic number %u, Number of nucleons %u",
+	       elt->GetName(), (elt->A())*g/mole, elt->Z(), elt->N() );
+      if(atomicNumber != elt->Z() || atomicWeight != (elt->A())*g/mole)
+	newMatDef = true;
     }
-    if( elt->Z() == 0 )   {
-      int n = int(atomicNumber/2e0);
-      if( n < 2 ) n = 2;
-      elt = new TGeoElement((xmat.nameStr()+"-CMS").c_str(),"CMS element",n,atomicNumber);
+    
+    if( !elt || newMatDef)  {
+      if( newMatDef )
+	printout(WARNING,"DD4CMS",
+		 "+++ Converter<ElementaryMaterial> Different definition of a default element with name:%s [CREATE NEW MATERIAL]",
+		 matname);
+      else
+	printout(WARNING,"DD4CMS",
+		 "+++ Converter<ElementaryMaterial> No default element present with name:%s  [CREATE NEW MATERIAL]",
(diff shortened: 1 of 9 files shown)


ianna:
please test

ianna:
+1

civanch:
+1

DetectorDescription/DDCMS/plugins/DDDefinitions2Objects.cc
```diff
@@ -445,30 +445,41 @@ template <> void Converter<DDLElementaryMaterial>::operator()(xml_h element) con
...
-      if( n < 2 ) n = 2;
-      elt = new TGeoElement(xmat.nameStr().c_str(),"CMS element",n,atomicNumber);
+    bool newMatDef = false;
+    
+    if(elt) {
+      // A is Mass of a mole in Geant4 units for atoms with atomic shell
+      printout(ns.context()->debug_materials ? ALWAYS : DEBUG, "DD4CMS",
+	       "    ROOT definition of %-50s Atomic weight %.3f, Atomic number %u, Number of nucleons %u",
+	       elt->GetName(), (elt->A())*g/mole, elt->Z(), elt->N() );
+      if(atomicNumber != elt->Z() || atomicWeight != (elt->A())*g/mole)
```
cvuosalo: It looks like there is more than one unit convention in this file. I'm not sure which convention `g` and `mole` follow. Is the ROOT convention the same as the CMS convention?
The `using namespace dd4hep;` and `using namespace geant_units::operators;` statements should probably be deleted or, at least, greatly limited in scope in order to require fully qualified names so it is more clear what convention each unit constant follows. `g` is particularly difficult -- how can I search the code to find where "g" is defined?
More comments about the different unit conventions in use would be helpful.

fabiocos: @cvuosalo as far as I can see this should follow the DD4hep definition which looks a clone of CLHEP, @ianna may confirm. As this is CMS code in principle the code brought in DataFormats/Math could be used as well

fabiocos:
@ianna @civanch running the geometry test I see many differences in material definition for elementary materials:

```
DD4CMS                 +++ Converting material "materials:Aluminium"                             Atomic weight 16839.593, Atomic number 13, Density: 1685.208.
DD4CMS                     ROOT definition of AL                                                 Atomic weight 16840.553, Atomic number 13, Number of nucleons 27
DD4CMS           WARN  +++ Converter<ElementaryMaterial> Different definition of a default element with name:materials:Aluminium [CREATE NEW MATERIAL]
DD4CMS                 +++ Converting material "materials:TD_Aluminium"                          Atomic weight 16839.593, Atomic number 13, Density: 5055.623.
DD4CMS           WARN  +++ Converter<ElementaryMaterial> No default element present with name:materials:TD_Aluminium  [CREATE NEW MATERIAL]
DD4CMS                 +++ Converting material "materials:Antimony"                              Atomic weight 75990.380, Atomic number 51, Density: 4168.704.
DD4CMS                     ROOT definition of SB                                                 Atomic weight 75996.621, Atomic number 51, Number of nucleons 122
DD4CMS           WARN  +++ Converter<ElementaryMaterial> Different definition of a default element with name:materials:Antimony [CREATE NEW MATERIAL]
DD4CMS                 +++ Converting material "materials:Argon"                                 Atomic weight 24933.583, Atomic number 18, Density: 1.023.
DD4CMS                     ROOT definition of AR                                                 Atomic weight 24933.583, Atomic number 18, Number of nucleons 40
DD4CMS                 +++ Converting material "materials:Arsenic"                               Atomic weight 46762.639, Atomic number 33, Density: 3570.144.
DD4CMS                     ROOT definition of AS                                                 Atomic weight 46762.389, Atomic number 33, Number of nucleons 75
DD4CMS           WARN  +++ Converter<ElementaryMaterial> Different definition of a default element with name:materials:Arsenic [CREATE NEW MATERIAL]
DD4CMS                 +++ Converting material "materials:Barium"                                Atomic weight 85714.652, Atomic number 56, Density: 2184.528.
DD4CMS                     ROOT definition of BA                                                 Atomic weight 85712.780, Atomic number 56, Number of nucleons 137
DD4CMS           WARN  +++ Converter<ElementaryMaterial> Different definition of a default element with name:materials:Barium [CREATE NEW MATERIAL]
DD4CMS                 +++ Converting material "materials:Beryllium"                             Atomic weight 5624.973, Atomic number 4, Density: 1153.431.
DD4CMS                     ROOT definition of BE                                                 Atomic weight 5624.961, Atomic number 4, Number of nucleons 9
```
etc...
These are essentially differences in the atomic weight....

ianna:
@fabiocos - yes, that is why I create a new material rather then picking up the ROOT definition.

civanch:
@ianna , @fabiocos , if it is possible bypass ROOT element/material object creation and make G4Element/G4Material directly from xml this would be the best.

ianna:
@civanch - not in a short term - we need a Mediator (ROOT in this case).

DetectorDescription/DDCMS/plugins/DDDefinitions2Objects.cc
ianna: yes, I'll stick with the DD4hep one here. There were multiple discussions among DD4hep developers. I hope, we'll converge on something common eventually.

cvuosalo: @fabiocos: I was not suggesting changing which unit conventions are used in this file. I was just requesting they be clearly commented. 
This file contains these two lines:
```
using namespace dd4hep;
using namespace geant_units::operators;
```
From what I found in `DD4hep/Evaluator/DD4hepUnits.h`, the DD4hep units convention does not match the Geant units convention. Because of this incompatibility, it is risky to use both namespaces at once because future developers and reviewers may get confused as to which units convention a variable or constant follows.
I would delete those two `using namespace` statements and then write the conversion as:
```
double dd4hepAtomMass = elt->A() * dd4hep::g / dd4hep:mole;
// Converting elt->A() in grams/mole in the Geant units convention to the
// DD4hep units connvention.
```
I would also add a comment at the top of the file saying:
```
/**** CAUTION: This file uses both the CMS and Geant units conventions.   ****/
```
If these changes can't be made now, I will make them myself in a later PR.

ianna:
@fabiocos - I'd like this PR to be integrated before I submit the next one which depends on it.

fabiocos:
@ianna @cvuosalo @civanch thanks for all the comments. I would say that we move forward by merging this PR, and a comment is then added on top

fabiocos:
+1

PR #26763, opened 2019-05-13
updating SC reg corrector for UL regression

Sam-Harper:
#### PR description:

This PR updates the mustache supercluster energy regression for incoming UL regression. 

The 2017 regression will be coming soon to a global tag and then 2016/2018 will follow in a few weeks. 

This PR is backwards compatible with the existing supercluster regression, as in it will work just fine with what is in the global tag right now. There are two changes

1) seed eta is added to the endcap as an input variable to help the regression be more stable with high eta. This is the last variable the regression and the current 74X regressions simply ignore it. 

2) a bug was fixed in the calculation of sigmaIEtaIPhi, it was always not set as sigmaIPhiIPhi was always set to std::numeric_limits<float>::max() when it was being constructed.   The option to reapply that bug is available. 

By default, this bug will be fixed which means the 74X regressions will get this fix and therefore the correction value for a given supercluster will change slightly after this change. This I do not consider an issue as

1. it only effects mustache superclusters whose exact energy is not terribly important except for the ECAL DPG who are the main users of this (most people use refined superclusters). It does enter E/gamma seeding though so can have some minor  effect there
1. the existing regressions are regardless broken right now when being run on 106X samples as they are not adapted to the new thresholds
1. the change is on average small.
https://sharper.web.cern.ch/sharper/cms/egamma/2019/May13th/cruijffSigmaBugComp.png
https://sharper.web.cern.ch/sharper/cms/egamma/2019/May13th/cruijffMeanBugComp.png
(with the differences being just bad fits)

#### PR validation:

The result of the new regression (which is not yet active, needs GT updates)
https://indico.cern.ch/event/817696/contributions/3423912/attachments/1840813/3018110/superClusterRegressionUL_ecalFollowup.pdf
https://indico.cern.ch/event/817696/contributions/3423665/attachments/1840744/3017891/UL_10X_regV2.pdf

Validating running this regression in CMSSW vs the ntuple training, saw 57 out 19,408,000 electrons have a difference between the two of more than 1 MeV. All were <0.5 GeV except one which had a delta of 1 GeV. Not entirely sure why, but very rare process and is good enough :)

### DIFF
--- a/RecoEcal/EgammaClusterProducers/python/particleFlowSuperClusterECAL_cfi.py
+++ b/RecoEcal/EgammaClusterProducers/python/particleFlowSuperClusterECAL_cfi.py
@@ -40,7 +40,8 @@ particleFlowSuperClusterECALBox = cms.EDProducer(
        uncertaintyKeyEE = cms.string('pfscecal_EEUncertainty_offline_v2'),
        vertexCollection = cms.InputTag("offlinePrimaryVertices"),
        ecalRecHitsEB = cms.InputTag('ecalRecHit','EcalRecHitsEB'),
-       ecalRecHitsEE = cms.InputTag('ecalRecHit','EcalRecHitsEE')
+       ecalRecHitsEE = cms.InputTag('ecalRecHit','EcalRecHitsEE'),
+       applySigmaIetaIphiBug = cms.bool(False)
        ),
 
     #threshold for final SuperCluster Et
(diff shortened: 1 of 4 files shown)


perrotta:
please test

perrotta:
+1
- This PR fixes the calculation of sigmaIEtaIPhi, which was broken before
  - There is effect on seeds, and therefore on other EGamma observables, as it is visible in the results of the jenkins tests
  - Verified that when setting ```applySigmaIetaIphiBug = true``` there are no differences in output with the previous (bugged) version
- Seed eta is added to the endcap as an input variable, and it will be needed by the new regression
  - Real changes due to the new regression will only be visible as soon as the new regression will be loaded to the DB: there are no changes in output due to this addition by now

slava77:
is there a backport to 10_6_X already?

Sam-Harper:
done

fabiocos:
+1

PR #26770, opened 2019-05-13
move header in private directory

smuzaffar:
#### PR description:

Although these are public headers but are not suppose to be included directly e.g
- `*TkRotation.h`: These are included via `ThRotation.h`
- `*Basic*DVector`: These headers are exposed via `Basic*DVector.h`
- `AVXVec.h`: This is included via `SSEVec.h` 

 This PRs move these headers in interface/private directory. This will help building CXXModules for all headers of package/interface directory. 

#### PR validation:

cms-checkdeps and local build works


smuzaffar:
please test

smuzaffar:
please test

VinInn:
Is this a new practice to eventually become pervasive?
Can it be added to the CMS software rules?

smuzaffar:
@fabiocos  @VinInn , I would recommend this practice. If no objections then yes we should update CMS software packaging rules.

perrotta:
+1
- technical 
- jenkins tests pass

civanch:
+1

fabiocos:
@smuzaffar @VinInn I understand the purpose in this specific case. Which are the other possible uses cases you have in mind?

VinInn:
> On 15 May, 2019, at 2:45 PM, Fabio Cossutti <notifications@github.com> wrote:
> 
> @smuzaffar @VinInn I understand the purpose in this specific case. Which are the other possible uses cases you have in mind?
> 
> 
There are many cases (in particular with templates) where the user interface is made up of many other "implementation" classes that shall not be used directly.
Where inlining is not required typically the implementation classes reside in an anonimous namespace in the "src" .cc file itself, but in case of inlining the implementation (even if in a different namespace)
remains exposed to users.

v.

fabiocos:
+1

PR #26777, opened 2019-05-14
[HGCal] Use tiling for speeding up CLUE algo nearest neighbor search

felicepantaleo:
#### PR description:
HGCal Layer Clustering `HGCalLayerClusterProducer` is in the top 3 main offenders in computing resources in Phase 2 reconstruction.
image

Today, Layer Clustering is performed using CLUE algorithm, which is using a KD-Tree for speeding up nearest neighbor search.
However, the KD-Tree is :
- hard to build on a GPU
- expensive if the search-box is small with respect to the point cloud volume
- not optimal for a small number of dimensions and we are just using it for x and y position.

This PR replaces the KD-Tree with a Tiling data structure.
This allows to achieve a speed-up of **1 order of magnitude** in PU200 conditions, making the Layer Clustering negligible with respect to the rest of the Phase-2 reconstruction.

#### PR validation:
The validation was executed with 10_6_0 on ttbar PU200:
https://fpantale.web.cern.ch/fpantale/PRs/26777

No differences are found for Silicon-only layers, improvements in the Scintillator are found due to the change in the logic of the association.

Memory footprint per stream is **down by more than 200MB**.
image

The overall phase 2 reconstruction time (including some overhead) is also **~4% faster**:
baseline: event loop CPU/event = 224.874532 s
PR: event loop CPU/event = 217.093278 s

This is due to the **speedup by an order of magnitude** of the hgcalLayerClusters module:
baseline (s): 6.160537
PR (s): 0.797660

@rovere @ZihengChen @tonydp03 @malgeri @cseez

### DIFF
--- a/RecoLocalCalo/HGCalRecProducers/interface/HGCalCLUEAlgo.h
+++ b/RecoLocalCalo/HGCalRecProducers/interface/HGCalCLUEAlgo.h
@@ -18,6 +18,9 @@
 #include "DataFormats/EgammaReco/interface/BasicCluster.h"
 #include "DataFormats/Math/interface/Point3D.h"
 
+#include "RecoLocalCalo/HGCalRecProducers/interface/HGCalLayerTiles.h"
+
+
 #include "RecoLocalCalo/HGCalRecAlgos/interface/RecHitTools.h"
 #include "RecoLocalCalo/HGCalRecAlgos/interface/KDTreeLinkerAlgoT.h"
 
@@ -36,7 +39,6 @@ class HGCalCLUEAlgo : public HGCalClusteringAlgoBase {
       (HGCalClusteringAlgoBase::VerbosityLevel)ps.getUntrackedParameter<unsigned int>("verbosity",3),
       reco::CaloCluster::undefined),
      thresholdW0_(ps.getParameter<std::vector<double> >("thresholdW0")),
-     positionDeltaRho_c_(ps.getParameter<std::vector<double> >("positionDeltaRho_c")),
      vecDeltas_(ps.getParameter<std::vector<double> >("deltac")),
      kappa_(ps.getParameter<double>("kappa")),
      ecut_(ps.getParameter<double>("ecut")),
@@ -48,13 +50,15 @@ class HGCalCLUEAlgo : public HGCalClusteringAlgoBase {
      nonAgedNoises_(ps.getParameter<edm::ParameterSet>("noises").getParameter<std::vector<double> >("values")),
      noiseMip_(ps.getParameter<edm::ParameterSet>("noiseMip").getParameter<double>("value")),
      initialized_(false),
-     points_(2*(maxlayer+1)),
-     minpos_(2*(maxlayer+1),{ {0.0f,0.0f} }),
-     maxpos_(2*(maxlayer+1),{ {0.0f,0.0f} }) {}
+     layerTiles_(2*(maxlayer+1)),
+     cells_(2*(maxlayer+1)),
+     numberOfClustersPerLayer_(2*(maxlayer+1),0)
+     {}
 
   ~HGCalCLUEAlgo() override {}
 
   void populate(const HGCRecHitCollection &hits) override;
+
   // this is the method that will start the clusterisation (it is possible to invoke this method
   // more than once - but make sure it is with different hit collections (or else use reset)
 
@@ -63,21 +67,19 @@ class HGCalCLUEAlgo : public HGCalClusteringAlgoBase {
   // this is the method to get the cluster collection out
   std::vector<reco::BasicCluster> getClusters(bool) override;
 
-  // use this if you want to reuse the same cluster object but don't want to accumulate clusters
-  // (hardly useful?)
   void reset() override {
     clusters_v_.clear();
-    layerClustersPerLayer_.clear();
-    for (auto &it : points_) {
-      it.clear();
-      std::vector<KDNode>().swap(it);
+    for(auto& cl: numberOfClustersPerLayer_)
+    {
+      cl = 0;
     }
-    for (unsigned int i = 0; i < minpos_.size(); i++) {
-      minpos_[i][0] = 0.;
-      minpos_[i][1] = 0.;
-      maxpos_[i][0] = 0.;
-      maxpos_[i][1] = 0.;
-    }
+
+    for(auto& cells : cells_)
+      cells.clear();
+    
+    for(auto& tiles: layerTiles_)
+      tiles.clear();
+
   }
 
   Density getDensity() override;
@@ -89,11 +91,6 @@ class HGCalCLUEAlgo : public HGCalClusteringAlgoBase {
         2.9,
         2.9,
         2.9
-        });
-    iDesc.add<std::vector<double>>("positionDeltaRho_c", {
-        1.3,
-        1.3,
-        1.3
         });
     iDesc.add<std::vector<double>>("deltac", {
         1.3,
@@ -122,7 +119,6 @@ class HGCalCLUEAlgo : public HGCalClusteringAlgoBase {
  private:
   // To compute the cluster position
   std::vector<double> thresholdW0_;
-  std::vector<double> positionDeltaRho_c_;
 
   // The two parameters used to identify clusters
   std::vector<double> vecDeltas_;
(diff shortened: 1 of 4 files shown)


slava77:
@cmsbuild please test

RecoLocalCalo/HGCalRecProducers/interface/HGCalLayerTiles.h
```diff
@@ -0,0 +1,68 @@
...
+      tiles_[getGlobalBin(x[i], y[i])].push_back(i);
+    }
+  }
+
+  int getXBin(float x) const {
+    constexpr float xRange = hgcalTilesConstants::maxX - hgcalTilesConstants::minX;
+    static_assert(xRange >= 0.);
+    constexpr float r = hgcalTilesConstants::nColumns / xRange;
+    int xBin = (x - hgcalTilesConstants::minX) * r;
+    xBin = std::min(xBin, hgcalTilesConstants::nColumns);
```
kpedro88: `xBin = std::clamp(xBin, 0, hgcalTilesConstants::nColumns);` (and similarly below)

RecoLocalCalo/HGCalRecProducers/interface/HGCalTilesConstants.h
```diff
@@ -0,0 +1,29 @@
...
+#include <array>
+
+namespace hgcalTilesConstants {
+
+  constexpr int32_t ceil(float num) {
+    return (static_cast<float>(static_cast<int32_t>(num)) == num) ? static_cast<int32_t>(num)
+                                                                  : static_cast<int32_t>(num) + ((num > 0) ? 1 : 0);
+  }
+
+  // first is for CE-E, second for CE-H in cm
```
kpedro88: is this comment correct?

RecoLocalCalo/HGCalRecProducers/plugins/HGCalCLUEAlgo.cc
```diff
@@ -22,346 +25,351 @@ void HGCalCLUEAlgo::populate(const HGCRecHitCollection &hits) {
...
-    if (ci != -1) {
-      clustersOnLayer[ci].push_back(nd[i]);
-      if (verbosity_ < pINFO) {
-        LogDebug("HGCalCLUEAlgo")
-          << "Pushing hit " << i << " into cluster with index " << ci << std::endl;
-      }
+  // need to pass clusterIndex to their followers
+  while (!localStack.empty()) {
+    int endStack = localStack.back();
+    // RecHitGPU thisHit = hits[frontOfBuffer];
```
kpedro88: delete commented-out code (or add a comment about why it is kept)

kpedro88:
@felicepantaleo thanks for this impressive speedup in the layer clustering algorithm. The "Tiling" structure is basically acting as a fixed-size geometric hash, which is more efficient for the 2D layer clustering case. A few questions, mostly for my curiosity:
* do you know the average amount of hits per bucket in 200PU?
* is there any impact on the memory usage? A back of the envelope calculation suggests it should not be very noticeable, but it would be good to confirm.

RecoLocalCalo/HGCalRecProducers/interface/HGCalLayerTiles.h
felicepantaleo: thanks, done

RecoLocalCalo/HGCalRecProducers/interface/HGCalTilesConstants.h
felicepantaleo: removed

RecoLocalCalo/HGCalRecProducers/plugins/HGCalCLUEAlgo.cc
felicepantaleo: removed

felicepantaleo:
ciao @kpedro88 ,
in a PU200 there are typically 300k hits and the total number of tiles is 1M. A lot of them are always empty as they lie outside the detector (especially in EE). So I think we should not really care about the average, since in a cluster there could be ~10 cells in a bin. 
The memory usage is now completely negligible with my commit :
https://github.com/cms-sw/cmssw/pull/26777/commits/cd4ad3fdde9cbfb70b7646a9aa8fe4e8441011c6

perrotta:
please test

VinInn:
The problem with a ```v<v<i>>``` is that on average a quarter of the allocated space is always waisted. Clearing w/o never deleting exacerbates the issue as the individual capacity will end to be the max for each local bin (of course aggressive delete will just maximize memory churn and increase timing).
An optimal solution of course does exist, still the proposed implementation may well be good enough for the current purposes.

kpedro88:
If we see evidence of memory hoarding in larger jobs, we can try to cure it with `swap()` or `shrink_to_fit()`, but let's wait to see if it actually becomes necessary.

VinInn:
> On 15 May, 2019, at 3:56 PM, Kevin Pedro <notifications@github.com> wrote:
> 
> If we see evidence of memory hoarding in larger jobs, we can try to cure it with swap() or shrink_to_fit(), but let's wait to see if it actually becomes necessary.
> 
> 
Extremely expensive solutions. The only valid alternative is NOT to use a ```v<v<i>>```, just a simpler linear ds.

v.

RecoLocalCalo/HGCalRecProducers/interface/HGCalTilesConstants.h
```diff
@@ -0,0 +1,28 @@
...
+// Date: 03/2019
+// Copyright CERN
+
+#ifndef RecoLocalCalo_HGCalRecAlgos_interface_HGCalTilesConstants_h
+#define RecoLocalCalo_HGCalRecAlgos_interface_HGCalTilesConstants_h
+
+#include <cstdint>
+#include <array>
+
+namespace hgcalTilesConstants {
```
slava77: namespaces should be all lower case

felicepantaleo: thanks Slava

perrotta:
please test

perrotta:
@felicepantaleo : just for having some number recorded in this thread, do you have, or can you get, a profile of the memory usage, compared to the one without this PR, for a PU200 benchmark wf?

felicepantaleo:
The entire phase-2 reconstruction workflow is about 4% faster at PU200.
There is a non negligible impact on memory ~200MB stream.
PR_memory

I can implement Vincenzo's suggestion in another PR in about one month.

perrotta:
> The entire phase-2 reconstruction workflow is about 4% faster at PU200.
> There is a non negligible impact on memory ~200MB stream.
> I can implement Vincenzo's suggestion in another PR in about one month.

Yes, the impact on memory is not negligible, but I think it can be afforded for now. And I take note that you are committing yourself to address it within about one month.

perrotta:
please test

slava77:
On 5/21/19 2:09 AM, perrotta wrote:
> Yes, the impact on memory is not negligible, but I think it can be
> afforded for now. And I take note that you are committing yourself to
> address it within about one month.
> 

what is the current production configuration for phase-2 jobs?
We have to be perhaps 0.5GB /thread away from the mem limit for this to
be really affordable.

felicepantaleo:
The main difference in memory between this PR and the baseline is that in the baseline the kdtree for each layer is not a data member and is created in the `makeClusters` for-loop: https://github.com/cms-sw/cmssw/blob/982c595daa9fdde39b7a5925f706b457b178607d/RecoLocalCalo/HGCalRecProducers/plugins/HGCalCLUEAlgo.cc#L88

instead in this PR, tiles per layer are stored as a data member.
If you insist, I could try to construct and filling the tiles in the for-loop so that they are destroyed at the end of each iteration (increasing memory churn)

felicepantaleo:
OK, I had few minutes to do that. Memory is now 150MB smaller than the baseline and timing is negligibly affected.
PR_memory_local_tiles

kpedro88:
@felicepantaleo excellent, thanks

kpedro88:
please test

perrotta:
Great @felicepantaleo , thank you!

Could you please now update the computing and physics performance links in the PR description, so that the updated ones will remain for reference, instead of the previous ones?

kpedro88:
looks like some kind of memory corruption, but no stack trace came back in the logs:
```
A fatal system signal has occurred: segmentation violation
The following is the call stack containing the origin of the signal.

Tue May 21 21:47:57 CEST 2019

Current Modules:

Module: HGCalLayerClusterProducer:hgcalLayerClusters (crashed)

A fatal system signal has occurred: segmentation violation
```

@felicepantaleo you may need to run it in the debugger to get more info.

RecoLocalCalo/HGCalRecProducers/interface/HGCalCLUEAlgo.h
```diff
@@ -48,13 +49,15 @@ class HGCalCLUEAlgo : public HGCalClusteringAlgoBase {
      nonAgedNoises_(ps.getParameter<edm::ParameterSet>("noises").getParameter<std::vector<double> >("values")),
      noiseMip_(ps.getParameter<edm::ParameterSet>("noiseMip").getParameter<double>("value")),
      initialized_(false),
-     points_(2*(maxlayer+1)),
-     minpos_(2*(maxlayer+1),{ {0.0f,0.0f} }),
-     maxpos_(2*(maxlayer+1),{ {0.0f,0.0f} }) {}
+    //  layerTiles_(2*(maxlayer+1)),
```
perrotta: When you will integrate the fix for the segmentation violation, please also manage to remove this now unneeded commented out line

felicepantaleo:
Now I'll run the benchmark on a bigger skylake node and get some measurements on ttbar + 200PU

perrotta:
> Now I'll run the benchmark on a bigger skylake node and get some measurements on ttbar + 200PU

Thank you!

perrotta:
please test

felicepantaleo:
@perrotta could you stop the tests? I've found another bug that I've fixed and I'm testing it

perrotta:
please abort test

slava77:
@cmsbuild please test

felicepantaleo:
Memory footprint per stream is down by 5%.
image

The overall phase 2 reconstruction time (including some overhead) is also ~4% faster:
baseline: event loop CPU/event = 224.874532 s
PR: event loop CPU/event = 217.093278 s

This is due to the speedup of the hgcalLayerClusters module:
baseline (s): 6.160537
PR (s): 0.797660

perrotta:
Good to remove that assert.
For the other assert's in the same class, however, the following header is still missing from HGCalLayerTiles.h
```C++
#include <cassert>
```

perrotta:
please test

perrotta:
@felicepantaleo : are the validation plots linked in the PR description up to date?

felicepantaleo:
@perrotta , I'm rerunning the validation now with the latest greatest and I will then update the initial description of the PR

felicepantaleo:
The new validation is here with negligible differences (small improvements in the mixed Silicon/Scintillator)
https://fpantale.web.cern.ch/fpantale/PRs/26777

felicepantaleo:
(blue original, red PR)

perrotta:
Thank you Felice.

While most of the plots show something which looks like being just the effect of some fluctuations, there are a few layers which show a significant increase in efficiency, for example
image
image

Same for fakerates, e.g.
image

Those layers have differences also in "cell association", e.g.:
image

This looks like a bug fix entered somewhere. Or maybe some different geometry is used in the two cases? Do you have an explanation? (Overall, the new plots represent an improvement)

felicepantaleo:
I have removed the copyright notice, as CERN is already the Legal Entity for CMSSW and that would have been redundant.

felicepantaleo:
@perrotta yes, this is understood as I have changed the logic of association to a hit to a cluster from being "clusterId is followed by these cells" to "these cells are followers of this cluster Id". 
This allows to have slightly different clusters topology and this is important especially for small clusters.
As the clustering in the scintillator is still being tuned, these small clusters are making most of the clusters in that part. Hence, reconstructing more of these increases efficiency.

TL:DR: I'm happy that the tiles based clustering is improving physics performance in the back of HGCal, but I think the tuning of the clustering of the scintillator would have achieved better results. So this is a first step in the correct direction, but not the final step.

perrotta:
@cmsbuild please test

felicepantaleo:
initial description is now up-to-date.
Thanks for reviewing it.

perrotta:
+1
- Improvements in timing and memory footprints are impressive
- Changes in Phase2 HGCal outputs are either due to fluctuations due to the modified algo, or improvements consequential of it

kpedro88:
+upgrade

fabiocos:
+1

PR #26788, opened 2019-05-15
Run3-hcx202 Neutal density filter kept for HB in Run3

bsunanda:
#### PR description:

In presence of neutral density filter in HB, the run3 scenario should use the older layer 0 weight

#### PR validation:

Tested with runTheMatrix -l limited

No backporting is necessary


bsunanda:
@cmsbuild Please test

cvuosalo:
+1

fabiocos:
@bsunanda for my understanding, no effect is seen in the limited PR test. Was this verified on some run3 workflow?

bsunanda:
@fabiocos The changes tested with the Run3 geometry available at that time in private testing (before 19 was changed to 21 because 21 geometry was not working because of a mistake at that time). This will be used to make a new run3 geometry for the DB. Unfortunately we could not take away the neutral density filters from HB which was in the original plan and most likely that will happen now during LS3.

fabiocos:
+1

PR #26797, opened 2019-05-16
updating SC reg corrector for UL regression: 106X

Sam-Harper:
####PR description:
This PR updates the mustache supercluster energy regression for incoming UL regression.

The 2017 regression will be coming soon to a global tag and then 2016/2018 will follow in a few weeks.

This PR is backwards compatible with the existing supercluster regression, as in it will work just fine with what is in the global tag right now. There are two changes

seed eta is added to the endcap as an input variable to help the regression be more stable with high eta. This is the last variable the regression and the current 74X regressions simply ignore it.

a bug was fixed in the calculation of sigmaIEtaIPhi, it was always not set as sigmaIPhiIPhi was always set to std::numeric_limits::max() when it was being constructed. The option to reapply that bug is available.

backport  of : https://github.com/cms-sw/cmssw/pull/26763

### DIFF
--- a/RecoEcal/EgammaClusterProducers/python/particleFlowSuperClusterECAL_cfi.py
+++ b/RecoEcal/EgammaClusterProducers/python/particleFlowSuperClusterECAL_cfi.py
@@ -40,7 +40,8 @@ particleFlowSuperClusterECALBox = cms.EDProducer(
        uncertaintyKeyEE = cms.string('pfscecal_EEUncertainty_offline_v2'),
        vertexCollection = cms.InputTag("offlinePrimaryVertices"),
        ecalRecHitsEB = cms.InputTag('ecalRecHit','EcalRecHitsEB'),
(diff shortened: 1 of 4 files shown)


perrotta:
backport of #26763

perrotta:
please test

perrotta:
AddOn test errors seem to be done to some missing input file, unrelated from this PR

perrotta:
please test
(retry)

perrotta:
+1
Identical backport of #26763, copying below my notes from that PR
- This PR fixes the calculation of sigmaIEtaIPhi, which was broken before
   - There is effect on seeds, and therefore on other EGamma observables, as it is visible in the results of the jenkins tests
   - Verified that when setting applySigmaIetaIphiBug = true there are no differences in output with the previous (bugged) version
- Seed eta is added to the endcap as an input variable, and it will be needed by the new regression
   - Real changes due to the new regression will only be visible as soon as the new regression will be loaded to the DB: there are no changes in output due to this addition by now

fabiocos:
+1

PR #26802, opened 2019-05-16
Tunesetforpythia8vgt235

efeyazgan:
Updating tunes to account for changes in pythia8 version >= 2.3.5

Please see slide 7 of https://indico.cern.ch/event/810088/sessions/307594/attachments/1824497/2990874/pcgt_GEN_Apr_10_2019.pdf

### DIFF
--- a/Configuration/Generator/python/MCTunes2017/PythiaCP1Settings_cfi.py
+++ b/Configuration/Generator/python/MCTunes2017/PythiaCP1Settings_cfi.py
@@ -4,7 +4,6 @@ pythia8CP1SettingsBlock = cms.PSet(
     pythia8CP1Settings = cms.vstring(
         'Tune:pp 14',
 	'Tune:ee 7',
-	'PDF:pSet=17',
 	'MultipartonInteractions:bProfile=2',
 	'MultipartonInteractions:ecmPow=0.1543',
 	'MultipartonInteractions:pT0Ref=2.40',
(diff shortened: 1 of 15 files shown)


efeyazgan:
please test

smuzaffar:
please test

efeyazgan:
+1

fabiocos:
@efeyazgan is this update desired for the UL production? If yes this would impact the production of the GEN-SIM PU library, am I correct?

fabiocos:
@efeyazgan as far as I can see, for the pdf this is just a renaming but the actual pdf does not change, am I correct?

efeyazgan:
@fabiocos yes this is for UL production. We had already produced the minbias using these settings, So, we should be fine. Now with this, we are generalizing it to all samples. 

keeping @pgunnell in the loop.

fabiocos:
@efeyazgan @pgunnell ok, but this means that 10_6_0 cannot be used to produce the library, it needs this update at minimum

efeyazgan:
> @efeyazgan as far as I can see, for the pdf this is just a renaming but the actual pdf does not change, am I correct?

@fabiocos the access method changed from pythia8 to lhapdf method (pythia8 authors decided to get rid of our main NNLO pdf from pythia8 internal sets). But this one we put as default is the positive-definite version of our pdf set which allows us to use this pdf up to arbitrarily high scales w/o the problem of negative-weights. In the standard model phase space this pdf set and our previous default are equivalent.

efeyazgan:
> @efeyazgan @pgunnell ok, but this means that 10_6_0 cannot be used to produce the library, it needs this update at minimum

But we added these settings buy hand in the minbias production with 10_6.

fabiocos:
@efeyazgan @pgunnell so this means that the settings are passed manually so far, and we will get this in 10_6_1? Please provide a backport for this

fabiocos:
+1

Issue #26776, opened 2019-05-14
Cyclic dependency between DataFormats/TrackerCommon and DataFormats/SiStripDetId


smuzaffar:
While building CXXMODULES IB, we noticed a cyclic dependency between DataFormats/TrackerCommon and DataFormats/SiStripDetId [a]. This can be avoided if we move TrackerTopology.* in to DataFormats/SiStripDetId itself or some other common package. If move to DataFormats/SiStripDetId sound reasonable then I can provide a PR.

[a]
```
>grep -r DataFormats/TrackerCommon DataFormats/SiStripDetId
DataFormats/SiStripDetId/src/SiStripSubStructure.cc:#include "DataFormats/TrackerCommon/interface/TrackerTopology.h"
DataFormats/SiStripDetId/BuildFile.xml:<use   name="DataFormats/TrackerCommon"/>
```
```
>grep -r DataFormats/SiStripDetId DataFormats/TrackerCommon
DataFormats/TrackerCommon/interface/ClusterSummary.h:#include "DataFormats/SiStripDetId/interface/StripSubdetector.h"
DataFormats/TrackerCommon/interface/TrackerTopology.h:#include "DataFormats/SiStripDetId/interface/SiStripDetId.h"
DataFormats/TrackerCommon/interface/TrackerTopology.h:#include "DataFormats/SiStripDetId/interface/StripSubdetector.h"
DataFormats/TrackerCommon/src/TrackerTopology.cc:#include "DataFormats/SiStripDetId/interface/StripSubdetector.h"
```

smuzaffar:
assign reconstruction

smuzaffar:
assign simulation

slava77:
It looks like SiStripSubStructure should be moved somewhere else.
Moving TrackerTopology to SiStrip* does not work from "first principles", because it has to serve all tracker variants, not just SiStrip.

davidlange6:
I agree - likely to DataFormats/TrackerCommon?

smuzaffar:
moving SiStripSubStructure in to DataFormats/TrackerCommon is even easier and involves less changes.,@perrotta  @slava77 @mdhildreth @civanch any objections?

pieterdavid:
Another option would be to define the two enums that are needed from the headers in DataFormats/SiStripDetId to somewhere in DataFormats/TrackerCommon; out of
```
>grep -r DataFormats/SiStripDetId DataFormats/TrackerCommon
[...]

PR #26804, opened 2019-05-16
Move strip tracker partition and ModuleGeometry enums to TrackerTopology.h

pieterdavid:
#### PR description:

This implements the solution proposed in https://github.com/cms-sw/cmssw/issues/26776#issuecomment-492536913 to solve the cyclic dependency between TrackerTopology.h and SiStripDetId.h/StripSubstructure.h. I changed to the TrackerTopology version in the uses that picked up SiStripDetId.h or StripSubdetector.h through TrackerTopology.h (since the latter is preferred), but added a SiStripDetId include for the few nontrivial ones (except one where the code could easily be simplified, https://github.com/cms-sw/cmssw/compare/master...pieterdavid:removecyclictrackercommonsistripdetid?expand=1#diff-efb246c04e3568bf8622531fd4ea48d9).
The other uses of these enums should also be changed (and use of SiStripDetId reduced in favour of TrackerTopology, where possible). I could add that here, but maybe it is better to leave that for a separate PR, after all code formatting changes have landed?

#### PR validation:

Compiled on top of CMSSW_10_6_0 before rebasing to master (to solve a conflict with formatting changes in DQM/SiStripMonitorTrack/src/SiStripMonitorTrack.cc).

### DIFF
--- a/Alignment/CommonAlignmentMonitor/plugins/AlignmentStats.cc
+++ b/Alignment/CommonAlignmentMonitor/plugins/AlignmentStats.cc
@@ -6,6 +6,7 @@
 #include "Alignment/TrackerAlignment/interface/AlignableTracker.h"
 #include "Alignment/CommonAlignment/interface/Alignable.h"
 #include "Alignment/CommonAlignment/interface/Utilities.h"
+#include "DataFormats/SiStripDetId/interface/SiStripDetId.h"
 
 #include "DataFormats/TrackerRecHit2D/interface/SiStripRecHit1D.h"
 #include "DataFormats/TrackerRecHit2D/interface/SiStripRecHit2D.h"
--- a/Alignment/SurveyAnalysis/src/SurveyDataReader.cc
+++ b/Alignment/SurveyAnalysis/src/SurveyDataReader.cc
@@ -6,7 +6,7 @@
 
 #include "DataFormats/TrackerCommon/interface/TrackerTopology.h"
 #include "Geometry/Records/interface/IdealGeometryRecord.h"
-// #include "DataFormats/SiStripDetId/interface/StripSubdetector.h"
+#include "DataFormats/SiStripDetId/interface/StripSubdetector.h"
 
 #include "Alignment/SurveyAnalysis/interface/SurveyDataReader.h"
 
--- a/Alignment/TrackerAlignment/plugins/MCMisalignmentScaler.cc
+++ b/Alignment/TrackerAlignment/plugins/MCMisalignmentScaler.cc
@@ -46,6 +46,7 @@
 #include "CondFormats/SiPixelObjects/interface/SiPixelQuality.h"
 
 #include "DataFormats/TrackerCommon/interface/TrackerTopology.h"
+#include "DataFormats/SiStripDetId/interface/StripSubdetector.h"
 
 #include "Geometry/TrackerGeometryBuilder/interface/TrackerGeometry.h"
 #include "Geometry/TrackerGeometryBuilder/interface/TrackerGeomBuilderFromGeometricDet.h"
--- a/Alignment/TrackerAlignment/plugins/TkAlCaOverlapTagger.cc
+++ b/Alignment/TrackerAlignment/plugins/TkAlCaOverlapTagger.cc
@@ -11,6 +11,7 @@
 #include "DataFormats/Common/interface/View.h"
 #include "DataFormats/TrackerCommon/interface/TrackerTopology.h"
 #include "Geometry/Records/interface/TrackerTopologyRcd.h"
+#include "DataFormats/SiStripDetId/interface/SiStripDetId.h"
 
 #include "TrackingTools/TrajectoryState/interface/TrajectoryStateOnSurface.h"
 #include "DataFormats/TrackReco/interface/Track.h"
--- a/Alignment/TrackerAlignment/src/AlignableTrackerBuilder.cc
+++ b/Alignment/TrackerAlignment/src/AlignableTrackerBuilder.cc
@@ -14,6 +14,7 @@
 #include "Alignment/TrackerAlignment/interface/AlignableSiStripDet.h"
 #include "Alignment/TrackerAlignment/interface/TrackerAlignableIndexer.h"
 
+#include "DataFormats/SiStripDetId/interface/SiStripDetId.h"
 
 
 //=============================================================================
--- a/Alignment/TrackerAlignment/src/TrackerAlignmentLevelBuilder.cc
+++ b/Alignment/TrackerAlignment/src/TrackerAlignmentLevelBuilder.cc
@@ -7,6 +7,7 @@
 // topology and geometry
 #include "DataFormats/TrackerCommon/interface/TrackerTopology.h"
 #include "Geometry/TrackerGeometryBuilder/interface/TrackerGeometry.h"
+#include "DataFormats/SiStripDetId/interface/StripSubdetector.h"
 
 
 
--- a/Alignment/TrackerAlignment/test/TrackerGeometryAnalyzer.cc
+++ b/Alignment/TrackerAlignment/test/TrackerGeometryAnalyzer.cc
@@ -13,6 +13,8 @@
 // for creation of TrackerTopology
 #include "Geometry/Records/interface/TrackerTopologyRcd.h"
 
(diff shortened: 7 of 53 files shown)


slava77:
@cmsbuild please test

Alignment/CommonAlignmentMonitor/plugins/AlignmentStats.cc
```diff
@@ -163,7 +163,7 @@ void AlignmentStats::analyze(const edm::Event &iEvent, const edm::EventSetup &iS
       AlignmentClusterFlag inval;
 
       bool hitInPixel= (subDet==PixelSubdetector::PixelBarrel)||(subDet==PixelSubdetector::PixelEndcap);
-      bool hitInStrip=(subDet==SiStripDetId::TIB) || (subDet==SiStripDetId::TID) ||(subDet==SiStripDetId::TOB) ||(subDet==SiStripDetId::TEC);
+      bool hitInStrip=(subDet==TrackerTopology::TIB) || (subDet==TrackerTopology::TID) ||(subDet==TrackerTopology::TOB) ||(subDet==TrackerTopology::TEC);
```
mmusich: matter of personal preference, but now that I see it done,  I find it odd to have:
`TrackerTopology::TIB  but `PixelSubdetector::PixelBarrel` in the same file

L1Trigger/TrackTrigger/plugins/TTStubBuilder.cc
```diff
@@ -41,7 +41,7 @@ void TTStubBuilder< Ref_Phase2TrackerDigi_ >::produce( edm::Event& iEvent, const
   for (const auto& gd : theTrackerGeom->dets())
   {
     DetId detid = (*gd).geographicalId();
-    if(detid.subdetId()!=StripSubdetector::TOB && detid.subdetId()!=StripSubdetector::TID ) continue; // only run on OT
+    if(detid.subdetId()!=TrackerTopology::TOB && detid.subdetId()!=TrackerTopology::TID ) continue; // only run on OT
     if(!tTopo->isLower(detid) ) continue; // loop on the stacks: choose the lower arbitrarily
```
mmusich: not related to the purpose of this PR, but why do we even have TOB and TIB in phase-2 code?

venturia:
@mmusich all this business of the enumerators to identify the Tracker subdetector is pretty complicated. When I did the work to merge present and phase-2 code several years ago I tried to clean (but not drastically) and to describe (maybe not thoroughly) the situation in the README files of the geometry packages (for example here https://github.com/venturia/cmssw/tree/CMSSW_7_4_X/Geometry/TrackerGeometryBuilder )
The issue can be summarized in this way: in some parts of the code there is the need to have something which counts from 1 to 6, where 6 is the largest number of subdetectors that the Tracker can have for a given geometry (there are several vectors and arrays that have 6 as hardcoded size and in some places the code relies on that). In other parts of the code, instead, having one enumerator for each possible Tracker subdetectors, combinging together phase-0, phase-1, phase-2, other exotic geomtries,... is needed. This leads, in some cases to the ambiguity of using TIB and TOB also for phase-2 in some cases. I am available to explain what I did on my side to anyone interested to continue the clean up

Alignment/CommonAlignmentMonitor/plugins/AlignmentStats.cc
pieterdavid: I also defined `TrackerTopology::PXB` and `TrackerTopology::PXF`, with the same values as in `PixelSubdetector` in case (these are 1 and 2, while the strip partitions are 3, 4, 5, and 6), so it would be trivial to change them, but I tried to keep this PR minimal (other opinions on naming etc. are welcome - I don't know how things are called for phase2, while TrackerTopology should work for all detectors).

L1Trigger/TrackTrigger/plugins/TTStubBuilder.cc
pieterdavid: I do not know if this is correct (especially since the TrackerGeometryBuilder documentation that @venturia pointed to says that Phase2 only has 4 and 5), but https://github.com/cms-sw/cmssw/blob/master/Geometry/TrackerCommonData/data/PhaseII/trackerParameters.xml defines a scheme for all subdetectors 3, 4, 5 and 6, so at least for the purpose of DetIds (as interpreted by the TrackerTopology) they are defined

venturia: Whoever modified the code since when I wrote the documentation should have updated the README file consistently. Otherwise we oscillate between no documentation and out-of-date documentation

slava77:
@pieterdavid 
 note that this PR does not merge anymore.
Please rebase/update.
Thank you.

slava77:
If I'm not mistaken a more minimal change would be to 
- introduce DataFormats/TrackerCommon/interface/SiStripEnums.h with only header content (actually, with just the header dependency, it can even stay in DataFormats/SiStripDetId)
    - ```SiStripSubDetector { UNKNOWN = 0, TIB = 3, TID = 4, TOB = 5, TEC = 6 }```
    - ```SiStripModuleGeometry { UNKNOWNGEOMETRY, IB1, IB2, OB1, OB2, W1A, W2A, W3A, W1B, W2B, W3B, W4, W5, W6, W7 }```
- use this instead of ```SiStripDetId::SubDetector``` or ```StripSubdetector::SubDetector``` and ```SiStripDetId::ModuleGeometry``` in the DataFormats/TrackerCommon/interface/TrackerTopology.h 
- perhaps also go into DataFormats/SiStripDetId/interface/SiStripDetId.h and DataFormats/SiStripDetId/interface/StripSubdetector.h and replace the values defined there with those from DataFormats/TrackerCommon/interface/SiStripEnums.h
    - e.g. in StripSubdetector ```using SubDetector = SiStripSubDetector;```

This would avoid a need to parse the full TrackerTopology.h for all use cases where just the enums are needed.

pieterdavid:
Sorry, I rebased on the wrong branch (DEVEL IB instead of regular IB), I'm fixing it...

pieterdavid:
This version implements the suggestion by @slava77 in https://github.com/cms-sw/cmssw/pull/26804#issuecomment-494175044 .
For the files that relied on one header including the other I added the include for now; it is possible to replace SiStripDetId::SubDetector and StripSubdetector::SubDetector by SiStripSubdetector::Subdetector (that's a search-and-replace exercise involving quite some files) - let me know if you prefer that I add it here instead of in a followup PR.

slava77:
> let me know if you prefer that I add it here instead of in a followup PR.

The full update will probably touch many files.
I think that such a change better be done after the clang-format campaign is over.
For now let's focus on the minimal part to resolve  #26776

slava77:
@cmsbuild please test

slava77:
+1

for https://github.com/cms-sw/cmssw/pull/26804	76a87c5
    - some of the changes are perhaps superfluous, but I think that they are fine
- jenkins tests pass and comparisons with the baseline show

kpedro88:
@pieterdavid there are merge conflicts, can you rebase?

jfernan2:
please test

slava77:
@cmsbuild please test

slava77:
+1

for https://github.com/cms-sw/cmssw/pull/26804	5a8f1f8
- changes since the last reco signoff (https://github.com/cms-sw/cmssw/pull/26804	76a87c5) are related to the rebase 

fabiocos:
@civanch @ggovi @ssekmen @cvuosalo @kpedro88 @andrius-k @rekovic could you please have a look?

jfernan2:
+1

ssekmen:
+1

cvuosalo:
+1

slava77:
alca, DB, upgrade, and L1 signatures are still missing.

@fabiocos  perhaps you can check this at some point soon to possibly override the missing signatures.
Thank you.

christopheralanwest:
+1

kpedro88:
+upgrade

fabiocos:
@rekovic @ggovi @civanch the changes in your areas look simple to me, please check, I would like to move forward with this PR by today

civanch:
+1

fabiocos:
+1

fabiocos:
merge

PR #26810, opened 2019-05-16
[Headers][RECO] Remove unused header

smuzaffar:
Looks like this header RecoVertex/LinearizationPointFinders/interface/GenericLinearizationPointFinder.h is not used anywhere. So I propose to remove it.


slava77:
@cmsbuild please test

slava77:
+1

for https://github.com/cms-sw/cmssw/pull/26810	89cc668 
- this is a cleanup of the code -- based on the git grep -- that was perhaps meant to be used in 2007 or so, but was never really added/enabled ==> the removal is OK
- jenkins tests pass

fabiocos:
+1

PR #26823, opened 2019-05-17
DD4hep: Sim Application Based on DD4hep

ianna:
#### PR description:

* **Interim** Sim Application implementation and its full configuration
* Retrieving G4 production cuts, but not yet setting them
* No sensitive volume assignment yet

Note: the code contains commented out fragments from original classes and it will be cleaned up when an equivalent functionality is provided. I'm submitting the code in order to migrate the developers area from 10_6 to 11.

#### PR validation:

The altered GEN_SIM step1 of a 10042.0 workflow runs fine(*) as follows. The events are empty due to an absence of sensitive volumes.

```
cmsRun SimG4Core/DD4hepGeometry/test/python/testZMM_13TeV_TuneCUETP8M1_cfi_GEN_SIM.py
```
*) if **biglibs** are disabled.

no back port is needed


ianna:
please test

ianna:
please test

civanch:
@ianna , let us discuss a possibility do not clone G4 RunManagers in new directories. I believe it is possible to configure an alternative geometry in existing classes. The reason : Geant4 needs from geometry only world volume, world volume itself should have all the rest, including sensitive detectors, production cuts, in parallel field map. I do not see any reason of duplication of Run Managers.

civanch:
@ianna ,

I mean following. There is few lines of code in the current RunManagerMT:
  G4LogicalVolumeToDDLogicalPartMap map_lv;
  m_world.reset(new DDDWorld(pDD, map_lv, m_catalog, false));
  m_registry.dddWorldSignal_(m_world.get());

Ideally nothing should be changed here, when you switch to other DD. I would assume that in DD code you chose between old and new variant. In your code  you put a flag, which configured from python for the time being, which will select old or DD4Hep method of building geometry.  I would expect, that there is no problem, but if this is too difficult let discuss an alternative, which lines you need here in RunManagerMT and let us put an extra logical flag to OscarMTProducer.

ianna:
@civanch - this is a prototype that is meant to be completely separate from the standard Sim application. IMHO, we need it to run a fair comparison. I agree, there should be only one application in the long term. As we have switched the releases, I'd like this PR to be integrated to be able to move forward. 
Besides, we need something which runs, even if it's not perfect, so that other people could start looking into sensitive detectors definition.

kpedro88:
@ianna I share @civanch's discomfort with wholesale copying of large chunks of core sim code.

For GeantV I have been maintaining a separate test package outside of the CMSSW repo. This could be worth considering rather than inducing a lot of churn here.

ianna:
@kpedro88 - we have discussed a DD4hep integration plan during the Sim meetings. I'll recap. Wherever possible the classes are kept as is. Wherever it's not possible they are copied and named DD4hep_*. The plan is to minimise the impact on current production software.

I'm not the only one who is working on the migration, so maintaining a separate test package is not an option. Maintaining a special IB is an option where we'd drop the DD entirely and go for DD4hep. We have passed the test phase and other people are about to contribute in the area of sensitive detector definition. These are plugins and it does not matter if they are invoked from an interim solution package.

Unfortunately, core sim code heavily depends on DD. The classes such as G4LogicalVolumeToDDLogicalPartMap, DDCompactView, etc. are used extensively. The API of equivalent DD4hep classes is much simpler. Hence a lot of code will have to be adapted. IMHO, going for a merged solution (as long as it's not "if then") would require both re-writing this code and re-writing our current DD. BTW, we are desperately short of person-power and on a very tight schedule to compete the migration. I'd appreciate if we converge on something soonish.

davidlange6:
hi @ianna - have you evaluated what would be required to support both within one codebase? (eg, with minimal duplication but more ifs/plugins/parallel constructors, etc)

ianna:
@davidlange6 - yes, some restructuring and cleanup of the Sim core packages would be needed.

davidlange6:
Perhaps that approach should get rediscussed within the sim group? It seems the consequences of your previous discussions/choices were not fully understood.. (i'm not paying attention so can't comment - however indeed extensive code duplication is in general hard to maintain)

civanch:
I would like to try to study how to switch between two DDs within RunManagerMT. The rest of the SIM code should not be affected by this transition, only one class RunManagerMT should directly contact DD. 

Note, that now we are in the beginning of 11_X developments, this release is for LS2 developments and not for production. We may do experimental things there (which do not destroy mainstream).

ianna:
> I would like to try to study how to switch between two DDs within RunManagerMT. The rest of the SIM code should not be affected by this transition, only one class RunManagerMT should directly contact DD.
> 
> Note, that now we are in the beginning of 11_X developments, this release is for LS2 developments and not for production. We may do experimental things there (which do not destroy mainstream).

@civanch - I do not see how a separate temporary package "destroys" mainstream. Could your study go in parallel with other development to which this PR is paramount?

ianna:
please test

kpedro88:
> I'm not the only one who is working on the migration, so maintaining a separate test package is not an option.

@ianna the second statement does not follow from the first statement.

ianna:
> > I'm not the only one who is working on the migration, so maintaining a separate test package is not an option.
> 
> @ianna the second statement does not follow from the first statement.

yes, it does. We need to make it easier for developers to contribute and not make it an exercise in DevOps where they would need to checkout and recompile the whole CMSSW on a regular basis:
```
[yana@cmsdev20 src]$ ls
Alignment            CondFormats          EgammaAnalysis       HLTriggerOffline     L1Trigger         RecoEcal          RecoPixelVertexing  SimMuon
AnalysisDataFormats  CondTools            ElectroWeakAnalysis  HeavyFlavorAnalysis  L1TriggerConfig   RecoEgamma        RecoTauTag          SimPPS
CalibCalorimetry     Configuration        EventFilter          HeterogeneousCore    L1TriggerOffline  RecoHI            RecoTracker         SimTracker
CalibMuon            DPGAnalysis          FWCore               HiggsAnalysis        MagneticField     RecoJets          RecoVertex          SimTransport
CalibPPS             DQM                  FastSimulation       IOMC                 MuonAnalysis      RecoLocalCalo     SUSYBSMAnalysis     TauAnalysis
CalibTracker         DQMOffline           Fireworks            IOPool               PerfTools         RecoMET           SimDataFormats      TopQuarkAnalysis
Calibration          DQMServices          GeneratorInterface   IORawData            PhysicsTools      RecoMTD           SimG4CMS            TrackingTools
CommonTools          DataFormats          Geometry             IgTools              QCDAnalysis       RecoMuon          SimG4Core           Utilities
CondCore             DetectorDescription  HLTrigger            JetMETCorrections    RecoBTag          RecoParticleFlow  SimGeneral          Validation

```

davidlange6:
I agree - migrations should be supported in cmssw
(but this does imply some prior agreement on the design, end-state and timescale...)

kpedro88:
Here's the checkdeps from the latest test of this PR:
```
09:51:40 + git cms-checkdeps -A -a
09:51:44 >> Checking DetectorDescription/DDCMS CMSSW_11_0_X_2019-05-21-1100
09:51:44    x DetectorDescription/DDCMS/interface/DDSpecParRegistry.h
09:51:44    x DetectorDescription/DDCMS/plugins/DDTestSpecParsFilter.cc
09:51:44    x DetectorDescription/DDCMS/src/DDSpecparRegistry.cc
09:51:44 >> Checking SimG4Core/DD4hepGeometry CMSSW_11_0_X_2019-05-21-1100
09:51:44    - SimG4Core/DD4hepGeometry/plugins/DD4hep_GeometryESProducer.cc
09:51:44    x SimG4Core/DD4hepGeometry/BuildFile.xml
09:51:44    x SimG4Core/DD4hepGeometry/interface/DD4hep_DDDWorld.h
09:51:44    x SimG4Core/DD4hepGeometry/plugins/BuildFile.xml
09:51:44    x SimG4Core/DD4hepGeometry/src/DD4hep_DDDWorld.cc
09:51:44 >> Creating dummy files under /build/cmsbld/jenkins/workspace/ib-run-pr-tests/CMSSW_11_0_X_2019-05-21-1100/poison directory.
09:51:44    SimG4Core/DD4hepGeometry/plugins/DD4hep_GeometryESProducer.cc
09:51:44 Checking out these packages: 2
09:51:44 Geometry/DTGeometryBuilder (header)
09:51:44 Geometry/MuonNumbering (header)
```

The small changes in other packages from this PR could be included in CMSSW without a problem. My concern is the several thousand lines of duplicated sim code. There's no reason for this to entire the official repository.

ianna:
@kpedro88 and @smuzaffar - I'd suggest to branch a separate IB for DD4hep migration.

kpedro88:
@ianna I remain unconvinced of the necessity of anything beyond a separate test repo (or equivalently a private CMSSW branch with the duplicated code in the `SimG4Core/DD4hepGeometry` package). Nothing depends on this new package, so this has virtually zero burden on developers. Your working area shows a bunch of unnecessary dependencies probably because you updated your branch by merging instead of rebasing.

civanch:
@ianna, I would suggest schedule discussion at the next SIM meeting or offline. The ideal solution would be not a separate branch - for me no difference if it is a separate branch or a separate sub-package. 

Why two different DDs may be called from RunManager? because it is only one connection point. Geometry should be defined during initialisation, in run time, SIM does not make any call to DD. So, the task is only to initialise things properly for one or another DD.

ianna:
@kpedro88 - we have to duplicate the code to decouple current production version from a dev version to run fair comparison. 
The current code is tightly coupled to the DD types and DD API: the records the plugins observe, the volume map, the logical volumes, etc. I also try to avoid circular dependencies - unavoidable if some of the classes reside where they are. In any case this package is a **temporary** solution.

ianna:
> @ianna, I would suggest schedule discussion at the next SIM meeting or offline. The ideal solution would be not a separate branch - for me no difference if it is a separate branch or a separate sub-package.
> 
> Why two different DDs may be called from RunManager? because it is only one connection point. Geometry should be defined during initialisation, in run time, SIM does not make any call to DD. So, the task is only to initialise things properly for one or another DD.

If it were the ideal world it would be the case. Reality is different, for example: the DDG4ProductionCuts from SimG4Core/​Physics package is tightly coupled to DD and is used by SimG4Core/Application. Current Sim Application run manages observe an IdealGeometryRecord, but it's now a GeometryFileRcd. G4LogicalVolumeToDDLogicalPartMap class is highly specific to DD, very elaborate and inefficient and is replaced by a simple std::map. The list goes on.

There are two ways to approach the Sim app migration to DD4hep. One is to provide an **interim** solution (aka this PR) to allow other developers to work in parallel on unrelated to the Sim app APIs plugins, then modernise the Sim app to be scalable.

Another option is to make the Sim app scalable, then involve other developers.

On my side I can only guarantee to oversee the first.

ianna:
Screen Shot 2019-05-21 at 19 29 52

kpedro88:
@ianna I understand that duplicating the code helps you iterate faster in the testing and migration. I just don't see a reason that the duplicated code has to be included in the official CMSSW (until the time when it is ready to enter production and the changes are moved to the original code).

cvuosalo:
A separate IB for DD4hep migration would appear to be a good compromise.

ianna:
> @ianna I understand that duplicating the code helps you iterate faster in the testing and migration. I just don't see a reason that the duplicated code has to be included in the official CMSSW (until the time when it is ready to enter production and the changes are moved to the original code).

What you suggest is similar to integration of other packages/versions such as ROOT, G4, etc. in a separate IB thread. It does make sense. It has never been agreed to support both DD and DD4hep. The plan has always been to migrate to DD4hep, validate it and drop DD.

davidlange6:
This makes the eventual validation and switchover quite complicated. Better to be able to use both systems together.



On 21 May 2019, at 22:23, Ianna Osborne <notifications@github.com<mailto:notifications@github.com>> wrote:


@ianna<https://github.com/ianna> I understand that duplicating the code helps you iterate faster in the testing and migration. I just don't see a reason that the duplicated code has to be included in the official CMSSW (until the time when it is ready to enter production and the changes are moved to the original code).

What you suggest is similar to integration of other packages/versions such as ROOT, G4, etc. in a separate IB thread. It does make sense. It has never been agreed to support both DD and DD4hep. The plan has always been to migrate to DD4hep, validate it and drop DD.

kpedro88:
The difference is that 100s of CMSSW packages depend on ROOT, while 0 packages depend on SimG4Core/DD4hepGeometry.

I am not just making a philosophical point here. I have done the entire GeantV integration and testing using a separate test repo https://github.com/kpedro88/SimGVCore. This affords the flexibility to try things and fix bugs without being tied to the CMSSW development cycle and maintenance requirements. Eventually, the final product can be brought over in a minimally intrusive way.

ianna:
> The difference is that 100s of CMSSW packages depend on ROOT, while 0 packages depend on SimG4Core/DD4hepGeometry.
> 

Nothing will depend on it because it is a driver which is needed to run sensitive detector plugins (yet to be ported).

> I am not just making a philosophical point here. I have done the entire GeantV integration and testing using a separate test repo https://github.com/kpedro88/SimGVCore. This affords the flexibility to try things and fix bugs without being tied to the CMSSW development cycle and maintenance requirements. Eventually, the final product can be brought over in a minimally intrusive way.

Is it an official request to add a GitHub repository support and maintenance on my todo list? Could someone from cat A be assigned to do it?

kpedro88:
It is a recommendation for what will be, in the end, the least burdensome way to proceed with testing migrated interfaces. I have no power to assign a cat A to do anything, nor do I think it would be necessary in this case. (What maintenance does a test repo require?)

If there is some reason that the approach I suggest will not work or will cause undesirable difficulty, it has not been communicated in this conversation in a way that I can understand.

ianna:
+1

civanch:
@ianna , SIM is using GEANT4 special branch for rolling tests of new Geant4 versions. It forbidden to make commits and PRs to this branch of CMSSW. This is a useful constrain to avoid diversity in software development.

Concerning DD: Geant4 geometry is static, in sense that it is created once at initialisation and changed in run time of SIM step. There are geometry description itself, materials, regions, production cuts, and sensitive detectors. The geometry is complex. Other SIM data are trivial, moreover, these additional properties should be added after the geometry is built, except materials, which likely are filled already with geometry. We have ~300 materials, ~25 regions and production cuts, ~10 sensitive detectors. These data are 1-dimensional data structures. For example, production cut requires from 1 to 4 double for each  object. Total memory for these simple objects is negligible. The requirement to the parser is to create this small data structures (shared between threads). The code to assign logical volumes to these data exist and is correct, no sense rewrite it, you need only take part from CompactView and add to the new equivalent class. Using template it is easy to switch between CompactView and new equivalent class. Only this place should be changed in SIM packages  - it is basically few lines of code.

ianna:
@civanch - I'd like to remove ```SimG4Core/Physics``` package direct dependency on ```DetectorDescription/Core``` by moving ```DDG4ProductionCuts``` class to ```SimG4Core/Geometry```. Please, let me know if you are ok with that. Thanks.

civanch:
@ianna , please do this, it is more correct to move this class to SimG4Core/Geometry, do not forget to change SimG4Core/Application/RunManager* lenks to headers.

ianna:
@civanch - ok, done here: https://github.com/cms-sw/cmssw/pull/26901

ianna:
Most of the functionality of this PR except for the sim application itself has been integrated in the IB. The latter is under discussion and will be submitted as a separate PR.

PR #26830, opened 2019-05-19
MTD geometry: material budget validation

fabiocos:
#### PR description:

This PR updates the Validation/Geometry package by adding a class for the evaluation of the MTD material budget. A macro providing plots is also added. The clang-format cleaning has been applied to the whole package.

#### PR validation:

Results of a 10k neutrinos run on scenario D38:

[Mtd_Materials_l_vs_eta.pdf](https://github.com/cms-sw/cmssw/files/3195412/Mtd_Materials_l_vs_eta.pdf)
[Mtd_Materials_x_vs_eta.pdf](https://github.com/cms-sw/cmssw/files/3195413/Mtd_Materials_x_vs_eta.pdf)

### DIFF
--- a/Validation/Geometry/interface/MaterialBudget.h
+++ b/Validation/Geometry/interface/MaterialBudget.h
@@ -28,11 +28,12 @@ class MaterialBudget : public SimWatcher,
                        public Observer<const EndOfTrack*> {
 public:
   MaterialBudget(const edm::ParameterSet&);
+  MaterialBudget(const MaterialBudget&) = delete;                   // stop default
   ~MaterialBudget() override;
 
-private:
-  MaterialBudget(const MaterialBudget&) = delete;                   // stop default
   const MaterialBudget& operator=(const MaterialBudget&) = delete;  // ...
+
+private:
 
   void update(const BeginOfRun*) override;
   void update(const BeginOfTrack*) override;
@@ -48,7 +49,7 @@ private:
   std::vector<G4LogicalVolume*> logVolumes;
   std::vector<TProfile*> me100, me200, me300, me400, me500, me600;
   std::vector<double> stepLen, radLen, intLen;
-  double eta, phi, stepT;
+  double eta_, phi_, stepT;
 };
 
 #endif
--- a/Validation/Geometry/interface/MaterialBudget.h
+++ b/Validation/Geometry/interface/MaterialBudget.h
@@ -21,22 +21,19 @@ class EndOfTrack;
 #include <string>
 #include <vector>
 
-class MaterialBudget : public SimWatcher, 
+class MaterialBudget : public SimWatcher,
                        public Observer<const BeginOfRun*>,
                        public Observer<const BeginOfTrack*>,
-		       public Observer<const G4Step*>,
+                       public Observer<const G4Step*>,
                        public Observer<const EndOfTrack*> {
-
 public:
-
   MaterialBudget(const edm::ParameterSet&);
   ~MaterialBudget() override;
-  
-private:
 
-  MaterialBudget(const MaterialBudget&) = delete;          // stop default
-  const MaterialBudget& operator=(const MaterialBudget&) = delete; // ...
-  
+private:
+  MaterialBudget(const MaterialBudget&) = delete;                   // stop default
+  const MaterialBudget& operator=(const MaterialBudget&) = delete;  // ...
+
   void update(const BeginOfRun*) override;
   void update(const BeginOfTrack*) override;
   void update(const G4Step*) override;
@@ -44,14 +41,14 @@ private:
 
   void book(const edm::ParameterSet&);
   bool stopAfter(const G4Step*);
-  
-  std::vector<std::string>      detTypes, detNames;
-  std::vector<int>              constituents, detLevels,regionTypes,stackOrder;
-  std::vector<double>           etaRegions, boundaries;
+
+  std::vector<std::string> detTypes, detNames;
+  std::vector<int> constituents, detLevels, regionTypes, stackOrder;
+  std::vector<double> etaRegions, boundaries;
   std::vector<G4LogicalVolume*> logVolumes;
-  std::vector<TProfile*>        me100, me200, me300, me400, me500, me600;
-  std::vector<double>           stepLen, radLen, intLen;
-  double                        eta, phi, stepT;
+  std::vector<TProfile*> me100, me200, me300, me400, me500, me600;
+  std::vector<double> stepLen, radLen, intLen;
+  double eta, phi, stepT;
 };
 
 #endif
--- a/Validation/Geometry/interface/MaterialBudgetForward.h
+++ b/Validation/Geometry/interface/MaterialBudgetForward.h
@@ -28,11 +28,12 @@ class MaterialBudgetForward : public SimWatcher,
                               public Observer<const EndOfTrack *> {
 public:
   MaterialBudgetForward(const edm::ParameterSet &);
+  MaterialBudgetForward(const MaterialBudgetForward &) = delete;                   // stop default
   ~MaterialBudgetForward() override;
 
-private:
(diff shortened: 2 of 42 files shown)


fabiocos:
please test

testing really nothing here...

jfernan2:
+1

fabiocos:
please test

fabiocos:
@pmeridian @abenagli please let me know whether the format looks ok, or you would prefer to change something. I've run it for D35, D38 and D41 with 1M input neutrinos.

Validation/Geometry/src/MaterialBudgetMtdHistos.cc
```diff
@@ -0,0 +1,296 @@
...
+  hmgr->addHistoProf1(new TProfile("1610", "MB prof Eta [Other];#eta;#lambda/#lambda_{0}", 250, -5.0, 5.0));
+  hmgr->addHisto1(new TH1F("1611", "Eta [Other]", 501, -5., 5.));
+  hmgr->addHistoProf1(
+      new TProfile("1620", "MB prof Phi [Other];#varphi [rad];#lambda/#lambda_{0}", 180, -3.1416, 3.1416));
+  hmgr->addHisto1(new TH1F("1621", "Phi [Other]", 180, -3.1416, 3.1416));
+  hmgr->addHistoProf2(new TProfile2D(
+      "1630", "MB prof Eta  Phi [Other];#eta;#varphi;#lambda/#lambda_{0}", 250, -5., 5., 180, -3.1416, 3.1416));
+  hmgr->addHisto2(new TH2F("1631", "Eta vs Phi [Other]", 501, -5., 5., 180, -3.1416, 3.1416));
+  hmgr->addHistoProf1(new TProfile("1640", "MB prof R [Other];R [mm];#lambda/#lambda_{0}", 200, 0., 2000.));
+
```
cvuosalo: Our coding guidelines recommend using constants for all these magic numbers: 5, 180, 250, 501, etc.

cvuosalo:
+1

jfernan2:
+1

jfernan2:
please test

Validation/Geometry/src/MaterialBudgetMtdHistos.cc
fabiocos: @cvuosalo you are correct, this was a quick copy of the tracker class. But as it is new it makes sense to update and get it clean asap, it makes also my work easier. Please see the new version

fabiocos:
@cvuosalo @jfernan2 I do not plan further updates for this PR, I will have one more in a forthcoming PR for a small update of the MTD geometry scenario. Here are the plots coming out from this setup:

[Mtd_Materials_x_vs_eta.pdf](https://github.com/cms-sw/cmssw/files/3220144/Mtd_Materials_x_vs_eta.pdf)
[Mtd_Materials_l_vs_eta.pdf](https://github.com/cms-sw/cmssw/files/3220145/Mtd_Materials_l_vs_eta.pdf)

As discussed with @pmeridian we have now the CMS publication suggested presentation (with the CMS_lumi and tdrstyle macros slightly adjusted compared to the version available in https://twiki.cern.ch/twiki/bin/view/CMS/Internal/FigGuidelines ). And I have addressed the comment of @cvuosalo about the definition of constants (aka "magic numbers"), which is indeed correct (although their meaning in this case was quite trivial).

fabiocos:
please test

jfernan2:
+1

Validation/Geometry/interface/MaterialBudget.h
```diff
@@ -21,37 +21,34 @@ class EndOfTrack;
...
                        public Observer<const EndOfTrack*> {
-
 public:
-
   MaterialBudget(const edm::ParameterSet&);
   ~MaterialBudget() override;
-  
+
 private:
+  MaterialBudget(const MaterialBudget&) = delete;                   // stop default
```
ianna: @fabiocos - always make deleted methods public because it is an announcement to the public users that they cannot be used.

Validation/Geometry/interface/MaterialBudget.h
```diff
@@ -21,37 +21,34 @@ class EndOfTrack;
...
   void update(const EndOfTrack*) override;
 
   void book(const edm::ParameterSet&);
   bool stopAfter(const G4Step*);
-  
-  std::vector<std::string>      detTypes, detNames;
-  std::vector<int>              constituents, detLevels,regionTypes,stackOrder;
-  std::vector<double>           etaRegions, boundaries;
+
+  std::vector<std::string> detTypes, detNames;
```
ianna: @fabiocos - naming rule #2.9: private members should be named as ```member_ ```

Validation/Geometry/interface/MaterialBudgetForward.h
```diff
@@ -21,41 +21,38 @@ class EndOfTrack;
...
-  std::vector<G4LogicalVolume*> logVolumes;
-  static const int              maxSet = 25;
-  TH1F                          *me400[maxSet];
-  TH2F                          *me800[maxSet];
-  TProfile                      *me100[maxSet], *me200[maxSet], *me300[maxSet];
-  TProfile2D                    *me500[maxSet], *me600[maxSet], *me700[maxSet];
-  std::vector<double>           stepLen, radLen, intLen;
-  double                        eta, phi, stepT;
+private:
+  MaterialBudgetForward(const MaterialBudgetForward &) = delete;                   // stop default
```
ianna: @fabiocos - public, please

Validation/Geometry/interface/MaterialBudgetHcal.h
```diff
@@ -15,32 +15,29 @@ class BeginOfTrack;
...
                            public Observer<const EndOfTrack*> {
-
 public:
-
   MaterialBudgetHcal(const edm::ParameterSet&);
   ~MaterialBudgetHcal() override;
-  
+
 private:
+  MaterialBudgetHcal(const MaterialBudgetHcal&) = delete;                   // stop default
```
ianna: @fabiocos - public, please

Validation/Geometry/interface/MaterialBudgetHcal.h
```diff
@@ -15,32 +15,29 @@ class BeginOfTrack;
...
-  
   void update(const BeginOfJob*) override;
   void update(const BeginOfTrack*) override;
   void update(const G4Step*) override;
   void update(const EndOfTrack*) override;
 
   bool stopAfter(const G4Step*);
-  
-  MaterialBudgetHcalHistos*   theHistoHcal;
+
```
ianna: @fabiocos - rule No2.9

ianna (changes requested):
Please, make sure that the rule No 2.9 is implemented in naming private members.

ianna:
-1

fabiocos:
@ianna your comments are technically correct, but please notice that they have nothing to do with the real content of this PR, as they affect classes that enters only through the clang format cleaning in the first commit that I did to help the core migration and avoid interferences. I think that we may profit of the fact that we touch the whole CMSSW code base to catch old issues, and I have fixed some here and there, but we should keep their fix separate from the cleaning in order not to slow down the latter, unless it is the cleaning itself to create issues (which is not the case here). Otherwise we should likely rewrite most of out code base. Anyway your suggestions are relatively small, I will add them.

fabiocos:
@ianna although not related with PR but for the clang-format cleaning, I have addressed a couple of issues you have raised by

- promoting to public "delete" methods;

- applying rule 2.9 were needed, as it says _A trailing “_” is the preferred method to distinguish a data member from the getter method_ . I have addressed the cases where there are variables and methods with the same name (not those pointed by you).

-

fabiocos:
please test

fabiocos:
@ianna @cvuosalo any further comment? or can we move forward?

cvuosalo:
+1

jfernan2:
+1

fabiocos:
+1

PR #26883, opened 2019-05-22
[RECO] Fixes needed for new root cxxmodules

smuzaffar:
- Added missing dependency 
```
>grep -R DataFormats/SiStripDigi DataFormats/SiStripCluster
DataFormats/SiStripCluster/interface/SiStripCluster.h:#include "DataFormats/SiStripDigi/interface/SiStripDigi.h"
```
- Removed usage of pattern `ROOT::Math::*` from https://github.com/cms-sw/cmssw/blob/c52e2819d7a2bca52273e9c36fc99622313eff71/DataFormats/Math/src/classes_def.xml#L7

### DIFF
--- a/DataFormats/Math/src/classes_def.xml
+++ b/DataFormats/Math/src/classes_def.xml
@@ -4,7 +4,10 @@
   <class pattern="ROOT::Math::MatRepStd<*>" />
   <class pattern="ROOT::Math::RowOffsets<*>" />
   <class pattern="ROOT::Math::SVector<*>" />
-  <class pattern="std::vector<ROOT::Math::*>" />
+  <class pattern="std::vector<ROOT::Math::SMatrix<*>>" />
+  <class pattern="std::vector<ROOT::Math::MatRepStd<*>>" />
+  <class pattern="std::vector<ROOT::Math::RowOffsets<*>>" />
+  <class pattern="std::vector<ROOT::Math::SVector<*>>" />
   <class pattern="edm::Wrapper<*>" />
   <class pattern="edm::RefVector<*>" />
   <class pattern="edm::ValueMap<*>" />
--- a/DataFormats/SiStripCluster/BuildFile.xml
+++ b/DataFormats/SiStripCluster/BuildFile.xml
@@ -1,5 +1,6 @@
 <use   name="DataFormats/Common"/>
 <use   name="DataFormats/TrajectoryState"/>
+<use   name="DataFormats/SiStripDigi"/>
 <use   name="boost"/>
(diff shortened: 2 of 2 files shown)


smuzaffar:
please test

DataFormats/Math/src/classes.h
```diff
@@ -3,6 +3,8 @@
 #define SMATRIX_USE_CONSTEXPR
 #include <Math/SVector.h>
 #include <Math/SMatrix.h>
+#include <Math/MinimTransformVariable.h>
+#include <TGraphDelaunay2D.h>
```
smuzaffar: @slava77 , these can be avoid if we replace the `ROOT::Math::*` pattern.

slava77: do you mean the patterns used in ```DataFormats/Math/src/classes_def.xml``` ?
There was some cleanup of dictionaries by the core team a while back to drop the wildcards.
Can that be revived/completed?

smuzaffar: yes the pattern in `DataFormats/Math/src/classes_def.xml` . I am testing the following right now.
```
-  <class pattern="std::vector<ROOT::Math::*>" />
+  <class pattern="std::vector<ROOT::Math::SMatrix<*>>" />
+  <class pattern="std::vector<ROOT::Math::MatRepStd<*>>" />
+  <class pattern="std::vector<ROOT::Math::RowOffsets<*>>" />
+  <class pattern="std::vector<ROOT::Math::SVector<*>>" />
```

smuzaffar:
@slava77 , @perrotta I updated the PR and replace the usage of  pattern `ROOT::Math::*` in classes_def.xml

slava77:
@cmsbuild please test

slava77:
+1

for https://github.com/cms-sw/cmssw/pull/26883	c9711ac
- technical
- jenkins tests pass

fabiocos:
+1

slava77:
ah, so , in https://github.com/cms-sw/cmssw/pull/26883#issuecomment-494969047

> Additional comment: 'Compilation

was actually meant to sound a significant alarm. I did not notice it.
I'll try to pay more attention to the summaries.
Although perhaps a bit more verbose message would be better.

smuzaffar:
It was meant to say "Compilation Warnings: Yes" which should have also added the "warning" label but due to bug in new pr build script it did not post the full message. I will fix this soon.

fabiocos:
@smuzaffar @slava77 unfortunately I missed it as well...

PR #26892, opened 2019-05-22
Adapting EGRegression Modifiers to 106X and adding V3 modifier

Sam-Harper:
#### PR description:

This is a speculative PR which may or may not be final.  Its more alerting people to what is coming as it'll look a lot like this with perhaps some variables changed. 

First there are bug fixes to the V1 and V2 regression modifiers. These modifiers cache the event setup object.  When using ModifiedElectronProducer/  ModifiedPhotonProducer this updates every lumisection. However at somepoint after 102X, the event setup object is deleted and should not be cached. Therefore it was modified to properly cache just the calo geometry. When this code was used in reco, it just happened that setEventContent must have been called every event so we didnt see this bug.  This was a fun debug btw. 

Then it adds the V3 regression modifier. Currently its fully backwards compatible with the existing regressions but adds features necessary for the new regression. The new regression may need more changes now based on studies ongoing. 

So this might be all the code we need but then again might have some more changes coming.

Right now even though its using the V3 regression modifer for the old regressions, no changes are expected.


slava77:
@cmsbuild please test

slava77:
@Sam-Harper 
did this work in the local setup? (using the ```runTheMatrix.py -l limited``` ) ?

the errors are like
```
----- Begin Fatal Exception 22-May-2019 17:40:50 CEST-----------------------
An exception of category 'ScheduleExecutionFailure' occurred while
   [0] Calling beginJob
Exception Message:
Unrunnable schedule
Module run order problem found: 
particleFlow consumes gedGsfElectrons, gedGsfElectrons consumes fixedGridRhoFastjetAll, fixedGridRhoFastjetAll consumes particleFlow
 Running in the threaded framework would lead to indeterminate results.
 Please change order of modules in mentioned Path(s) to avoid inconsistent module ordering.
```

Sam-Harper:
ah stupid thing on my part, mixing my local testing. Fixed

slava77:
@cmsbuild please test

@smuzaffar is there a way to make the bot to notice force-pushed updates?

slava77:
> @smuzaffar is there a way to make the bot to notice force-pushed updates?

worse yet, it's not reacting to my please test either

slava77:
ah, it looks like the problem is on github side
https://www.githubstatus.com/
Screen Shot 2019-05-22 at 11 31 59 AM

slava77:
@Sam-Harper in the PR description I read
> Right now even though its using the V3 regression modifer for the old regressions, no changes are expected.

this does not match https://github.com/cms-sw/cmssw/pull/26892#issuecomment-495010234
> Reco comparison results: 3976 differences found in the comparisons
> DQMHistoTests: Total failures: 8961

it looks like the differences start from gedPhotons and/or the pf candidate photons; there are fairly significant changes in et.
e.g. in wf 1001.0 (Run2011A minbias) it looks like about a half of photons are now with et below 5 GeV
all_OldVSNew_RunMinBias2011Awf1001p0c_recoPhotons_gedPhotons__RECO_obj_et

Sam-Harper:
Opps, I had throughly tested the electrons but must have missed something in the photons! Thanks. Should be easy to fix. Will try to do so today.

RecoEgamma/EgammaTools/plugins/EGRegressionModifierV3.cc
```diff
@@ -0,0 +1,364 @@
...
+  //return the object unmodified if so
+  if(!eleRegs_) return;
+
+  const reco::SuperClusterRef& superClus = ele.superCluster();
+  
+  // skip HGCAL for now
+  if( superClus->seed()->seed().det() == DetId::Forward ) return;
+
+  const int numberOfClusters =  superClus->clusters().size();
+  const bool missingClusters = !superClus->clusters()[numberOfClusters-1].isAvailable();
```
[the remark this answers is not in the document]
Sam-Harper: so size 0 should never happen but I think its possible to happen if somebody screws up creating a supercluster in some future bit of code. So i'll add a check.  

If one cluster is available, then they all should be available, hence we just need to check one. No good reason for checking the back rather than front, I choose back incase somebody did something cute with the seed clusters (aka the front). Will make the change as suggested.

slava77:
> Opps, I had throughly tested the electrons but must have missed something in the photons! Thanks. Should be easy to fix. Will try to do so today.

@Sam-Harper 
Please clarify on the status of updates for this PR.
Thank you.

Sam-Harper:
They are going to come soon (tm). However the update that does come will be untestable till we get a new global tag.

Sam-Harper:
also I havent submitted as I only want to take out my debugging info once so I need the whole thing final first. Its epsilon close

RecoEgamma/EgammaTools/plugins/EGRegressionModifierV3.cc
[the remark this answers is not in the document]
Sam-Harper: so I found out why I did it the rather odd way I did it :)

Turns out I had a reason. Its a PtrVector and has no method "back".  Still did the size 0 check just in case.  So left as is but with a check for  size=0.

RecoEgamma/EgammaTools/plugins/EGRegressionModifierV3.cc
```diff
@@ -0,0 +1,364 @@
...
+  //return the object unmodified if so
+  if(!phoRegs_) return;
+  
+  const reco::SuperClusterRef& superClus = pho.superCluster();
+
+  // skip HGCAL for now
+  if(superClus->seed()->seed().det() == DetId::Forward ) return;
+
+  const int numberOfClusters =  superClus->clusters().size();
+  const bool missingClusters = !superClus->clusters()[numberOfClusters-1].isAvailable();
```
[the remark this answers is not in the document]
Sam-Harper: likewise "back" method does not exist for this class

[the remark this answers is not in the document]
slava77: this can also be a ```superClus->clusters().isAvailable()```
https://github.com/cms-sw/cmssw/blob/82e0ab0d145855d81d4da1251198d37dec744407/DataFormats/Common/src/PtrVectorBase.cc#L83-L99

Sam-Harper:
quickly before you test (I lost internet when I submitted). This PR will not work until we get the conditions in the Gt

Sam-Harper:
okay so longer status report:

this PR should be final, its in my testing right now but it should work. It is however awaiting the conditions entering the GT which are already being tested in the dev database and will make their way to production today assuming tests complete. 

It will not work until the GT for the workflow gets the new conditions.

slava77:
@christopheralanwest @tocheng 
please share a PR number or even propose a commit to be added to this PR 
when a new GT with the new egamma regressions is available.
Thank you.

slava77:
> @christopheralanwest @tocheng
> please share a PR number or even propose a commit to be added to this PR
> when a new GT with the new egamma regressions is available.
> Thank you.

Do we have a GT that can be used?
Please clarify.
Thank you.

christopheralanwest:
please test workflow 7.4

The tests are expected to fail but I will run them anyway for documentation of which workflows require the new conditions. I am working on generating all of the new GTs for this PR and will post a separate PR when I am finished.

Sam-Harper:
first thank you Chris! Also this was useful, although some tests failed due to an oversight in the configs in the calibrated electrons/photons re-miniAOD workflow which is enabled for some years. Fixing that now.

slava77:
@cmsbuild please test with #27089

RecoEgamma/EgammaTools/plugins/EGRegressionModifierV3.cc
```diff
@@ -63,8 +63,15 @@ class EGRegressionModifierV3 : public ModifyObjectValueBase {
...
   edm::ESHandle<CaloGeometry> caloGeomHandle_;
 
+  //because we're using a direct comp to this value to detect if its a default value
+  //we need to ensure its using the exact same type as its declared as
+  using TrkFBremType = decltype(reco::GsfElectron::ClassificationVariables::trackFbrem);
+  static const TrkFBremType kDefaultTrackFbrem_;
 };
 
+const EGRegressionModifierV3::TrkFBremType EGRegressionModifierV3::
+kDefaultTrackFbrem_ = reco::GsfElectron::ClassificationVariables().trackFbrem;
```
slava77: this better be replaced by a direct (static const or constexpr) constant defined in ```GsfElectron::ClassificationVariables```.
I'm not sure why there should be an extra layer of creating a default object here.

RecoEgamma/EgammaTools/plugins/EGRegressionModifierV3.cc
```diff
@@ -231,8 +236,10 @@ std::array<float,32> EGRegressionModifierV3::getRegData(const reco::GsfElectron&
     data[26] = iEta;
     data[27] = iPhi;
     data[28] = (iEta-signIEta)%5;
-    data[29] = (iPhi-1)%2;
-    data[30] = (std::abs(iEta)<=25)*((iEta-signIEta)) + (std::abs(iEta)>25)*((iEta-26*signIEta)%20);  
+    data[29] = (iPhi-1)%2; 
+    const int iEtaCorr = iEta - (iEta > 0 ? +1 : -1);
+    const int iEtaCorr26 = iEta - (iEta > 0 ? +26 : -26);
```
slava77: just to refresh my memory: is it true that iEta can not be equal to 0?

slava77:
IIUC, the number of used payloads about doubled.
What is the memory increase from this?
@Sam-Harper  please provide some estimate of the change in the memory used for this (a memory profile would be the best; although some general estimate based on the size of the payload could be good enough).

Sam-Harper:
> IIUC, the number of used payloads about doubled.
> What is the memory increase from this?
> @Sam-Harper please provide some estimate of the change in the memory used for this (a memory profile would be the best; although some general estimate based on the size of the payload could be good enough).

@slava77 , where do you see the number of payloads doubling? In practice we should have reduced the number (we didnt add any and in fact we reduced the ecal-trk to share the same one for low and high pt) so I'm confused and a little worried if somethings reporting such an increase. 

Best,
Sam

slava77:
> @slava77 , where do you see the number of payloads doubling? In practice we should have reduced the number (we didnt add any and in fact we reduced the ecal-trk to share the same one for low and high pt) so I'm confused and a little worried if somethings reporting such an increase.
e.g.
https://cms-conddb.cern.ch/cmsDbBrowser/diff/Prod/gts/106X_upgrade2018_realistic_forPR26892_v4/106X_upgrade2018_realistic_v4
There are 24 new or changed payloads of type GBRDWrapperRcd in the new GT, of which 12 were not present in the baseline.

Sam-Harper:
> There are 24 new or changed payloads of type GBRDWrapperRcd in the new GT, of which 12 were not present in the baseline.

Ah. So the 12 new ones replace existing payloads which there are some mild value for having around the olds ones. 

Each payload will have a barrel and endcap flavour and a mean & sigma. So it will be in multiples of 4

- electron ecal   non-saturated (aka pt <300)
- electron ecal-trk  pt 1-300
- photon ecal  non-saturated (aka pt < 300)

The ecal ones replace the existing ones. The ecal-trk  replaces 1-50 , 50-200 trainings so has 4 rather than 8 payloads. 

However the ecal-trk 1-50, 50-200 is still needed for running the scales and smearings on existing MC samples as it has to reapply the old regression. Finally I didnt bother yet requesting the removal of the electron photon ecal only ones as there is some value in being able to run the old regression for now. However when the dust settles, yes we should do a clear out.  I honestly dont know the policies, I assume we have a different GT for re-miniAODing <=102X data and UL data? Right now its easiest to have things co-exist.

slava77:
> Ah. So the 12 new ones replace existing payloads which there are some mild value for having around the olds ones.

Just to make sure, the old ones are not used in the same jobs as the old ones.
Is this correct?
Please confirm.

Sam-Harper:
> > Ah. So the 12 new ones replace existing payloads which there are some mild value for having around the olds ones.
> 
> Just to make sure, the old ones are not used in the same jobs as the old ones.
> Is this correct?
> Please confirm.

Yes. Unless somebody did an invalid combination of jobs, ie running RECO with the modifiers meant for re-miniAODing. Then in that case they would also consume the old ecal-trk regressions. But that would be an error on their part.

RecoEgamma/EgammaTools/plugins/EGRegressionModifierV3.cc
Sam-Harper: yes, it can not be true for a valid EB DetId.  It goes directly from -1 to 1

RecoEgamma/EgammaTools/plugins/EGRegressionModifierV3.cc
Sam-Harper: Done.

Sam-Harper:
okay addressed that small issue in the code review where we now define a default value in reco::GsfElectron::Classification to check against.

perrotta:
@cmsbuild please test

Sam-Harper:
@cmsbuild please test with #27089

Sam-Harper:
(needs that PR)

slava77:
@cmsbuild please test with #27089

slava77:
@Sam-Harper 
looking at the GT, e.g. https://cms-conddb-prod.cern.ch/cmsDbBrowser/diff/Prod/gts/106X_upgrade2018_realistic_v4/106X_upgrade2018_realistic_v5

I wanted to check why 8 old payloads ```{electron,photon}_e{b,e}_ECALonly{,_var}``` are still used. Is there some  coincidence that the 80X training still works OK in 10_6_X for higher ET?

Sam-Harper:
@slava77 

This is indeed intentional. And part of the agreed plan. But thank you for checking, vigalent as always. 

The high pt regressions are these days only used for e/gamma objects which are saturated which has been the case shortly after they were introduced. The reason for this is that they have the interesting property of identifying fakes as saturated e/gammas which made jetMET unhappy.  

As time is tight, we decided not to retrain these regressions but keep as is. We are working under the assumption that changing in the PF rec-hit thresholds (the major change as far as the regression is concerned) should not impact saturated e/gammas. This assumption has not been tested yet, however if it is wrong, the consquences are minimal, we are mostly worried here about getting something reasonable out of the box and if saturated electrons/photons are slightly off, then so be it, we'll fix it later. 

 This is part of the reason the regression is so similar to before, we didnt have much freedom to change things because of this approach. We did change the seed crystal defination but we think that shouldnt have a major effect here.

slava77:
>  @Sam-Harper please provide some estimate of the change in the memory used for this (a memory profile would be the best; although some general estimate based on the size of the payload could be good enough).

So, I looked at the first few payloads (electron_eb_ecalOnly_1To300_0p2To2_mean vs electron_eb_ECALonly_lowpt and  ee and their sigmas; and also the eb photon low pt).
It looks like the payload sizes are generally down by a factor of about 3 based on XML dumps.
E.g. 'electron_eb_ECALonly_lowpt',    GEDelectron_lowpt_EBCorrection_80X_EGM_v4 was about 18M compared to electron_eb_ecalOnly_1To300_0p2To2_mean electron_eb_ecalOnly_1To300_0p2To2_mean_2017ULV2 with about 5.5MB. The sigmas for this case go down from 6.1M to 2.1M.

This is good.

Sam-Harper:
@slava77 , thanks! I had the thought last night that the payloads were likely no longer the same size and I should check asap! thanks for this.

slava77:
Here are some somewhat unexpected plots/comparisons

ele 1TeV gun wf 10803
all_sign1080VSorig_SingleElectronPt1000in2018wf10803p0c_recoGsfElectrons_gedGsfElectrons__RECO_obj_correctedEcalEnergyError
ecalEnergyError() has the same change
essentially the same is showing up in gedPhotons.energyCorrections().regression[12]EnergyError.

The ele pt 35 gun wf 10802 looks \~reasonable 
all_sign1080VSorig_SingleElectronPt35in2018wf10802p0c_recoGsfElectrons_gedGsfElectrons__RECO_obj_ecalEnergyError

Some less significant but still a bit off changes are in the response
of low-R9 photons with a somewhat significant overshoot
e.g. wf10805 35 GeV gun
wf10805ph35_pho_EOgen_r9lo_EB

@Sam-Harper 
please check and comment if these are expected.
Thank you.

Sam-Harper:
> Here are some somewhat unexpected plots/comparisons
> 
> ele 1TeV gun wf 10803
> all_sign1080VSorig_SingleElectronPt1000in2018wf10803p0c_recoGsfElectrons_gedGsfElectrons__RECO_obj_correctedEcalEnergyError
> ecalEnergyError() has the same change
> essentially the same is showing up in gedPhotons.energyCorrections().regression[12]EnergyError.
> 
> The ele pt 35 gun wf 10802 looks ~reasonable
> all_sign1080VSorig_SingleElectronPt35in2018wf10802p0c_recoGsfElectrons_gedGsfElectrons__RECO_obj_ecalEnergyError
> 
> Some less significant but still a bit off changes are in the response
> of low-R9 photons with a somewhat significant overshoot
> e.g. wf10805 35 GeV gun
> wf10805ph35_pho_EOgen_r9lo_EB
> 
> @Sam-Harper
> please check and comment if these are expected.
> Thank you.

So the high pt one was not overly expected. And it seems for whatever reason, the resolution estimate is not doing great in the endcap at high pt, its over estimating it by a factor 2. The corrected energy (as you know as you would have pointed it out if it wasnt) is fine.  Its unclear how much of a problem this really is. And could be extremely difficult to fix (as in would require a rethink of the strategy, we would have to add a third regression bin). 

On a practical side of things, past 200 GeV, this value is not used anywhere in the reconstruction and can be retrained on the miniAOD if needed. 

As for the photons, will have a look. I think one thing to keep in mind here is that E/gamma can always reapply the ele/photon regressions at the miniAOD level. So it doesnt have to be perfect, just good enough not to mess up jetMET. I think having E/gamma perfect out of the box is a dream which is incompatible with the current schedule

Sam-Harper:
Yeah ,so this is why I dislike blind machine learning

image

This is for the endcap. 

The lower population numbers are fine. However the regression somehow has decided that there is an energy dependent variable which causes it to switch to random value of 4%. 

Lovely. The actual performance of the energy here is fine though, that is what is important.

 We shall see what we can do here but I think we should continue to move forward here, while this is not good, I strongly suspect not a single analysis in CMS has ever used this number TeV+ electrons and if somebody does want it, we can fix it after the fact.

Sam-Harper:
So I think I know what has gone wrong here with the training. 

The only source of high energy (ie 1.5 TeV+) electrons is in the high eta endcap (just from kinematics as the pt is trained up to 300 GeV).  Its then looked at all the 1.5 TeV electrons in energy and gone, these have bad resolution (as they are in the high endcap). Made much worse by the fact the resolution is now so poor in the high endcap. Will figure out a solution

Sam-Harper:
image

Here is R9 for photons with gen pt 30 to 35 GeV in the barrel. I think your plot was just stats.

slava77:
On 6/6/19 2:04 AM, Sam Harper wrote:
> Here is R9 for photons with gen pt 30 to 35 GeV in the barrel. I think
> your plot was just stats.
> 

well, I have essentially the same plots with an overshoot in the barrel
for the pho 35 sample in
- wf 10805 for 2018
- wf 1319 for 2016 setup
- wf 19 for Run1 setup
It's hard to blame stats for all of them

Is the plot you posted for the barrel?
(the endcap looks OK).

The tests were done in CMSSW_11_0_X_2019-06-04-1100

slava77:
2016 pho 35:
wf1319ph35_pho_EOgen_reg1_r9lo_EB

run1 pho 35
wf19ph35_pho_EOgen_reg1_r9lo_EB

Sam-Harper:
2016 photon is not expected to give any sensible results. Nor really is Run1

Sam-Harper:
To be clear. Every year needs its proper regression. At this point in time, only 2017 exists.  The others will be fixed in time.

Sam-Harper:
So on the outstanding issues:

1) I have been convinced that perhaps its not for best to change RunI / Hion. I dont think they are particularly optimal / make sense currently but fine. As such I will set their modifiers for those workflows to use the old regression.  I could supply a new retrained one appropriate to their conditions but this is seen as the more labour intensive option. Unless you dont mind me leaving them as is.  2016/2018 will be left broken intentionally as they are merely awaiting their conditions.

2) I am testing a fix for the energy error issue. I might be able to get away with capping the raw energy inputted to the BDT at 1.2 TeV, tests are running. Fixing it properly is extremely difficult. To really fix it would require a rethink on how we do regressions. We could also fix it by training with a higher pt sample but that merely pospones the problem to about 1.5 TeV in Et and might introduce new and exciting problems.

slava77:
On 6/6/19 8:14 AM, Sam Harper wrote:
> 2016/2018 will be left broken intentionally as they are merely awaiting
> their conditions.


Ah, so I was too naive to assume that an update to GTs for everything
other than 2017
was intentional and was supposed to make things good.
Now it looks like it was only made to make things run and the results
are otherwise broken and should not be really looked at.
This is a fairly strange approach.
Can we get the 2016/2018 and Run-1 back to the setup that is less broken
(assuming that before this update it was indeed less broken)?

My plots/tests did not include 2017 so far.
I've started a batch for 2017 now and will report some time later today.

Sam-Harper:
> Can we get the 2016/2018  back
Why? The release is fine for them and they are just waiting their conditions in the GT. This will come. Otherwise we'll have to undo and then redo it which is more odd no?

Sam-Harper:
Okay this is very difficult to find the information and you likely know it @slava77 .  What are the modifers for Run1? They also dont seem to be used in the matrix tests. 

Also for the record, I find it extremely odd that everything is using 2018 pf rec-hit thresholds, eg Run Hion does this as far as I can tell.

christopheralanwest:
I am working on preparing the GTs and will update PR #27089 later today.

christopheralanwest:
hold

This PR should not be merged before updating the GTs.

slava77:
> What are the modifers for Run1?

there are none. Run-1 is the default.



> Why? The release is fine for them

IMHO, "fine" is limited to consistently not knowingly worse performance.
I do not think that the release is fine because these GTs/egamma payloads are apparently broken.

christopheralanwest:
To summarize a discussion I just had with Sam, I will

- Revert the changes to the tags beginning with "pfscecal" and "ecalPFClusterCor2017V2" in all queues except those for 2017 MC (106X_mc2017_design and 106X_mc2017_realistic) and 2017 data (106X_dataRun2_2017_2018, 106X_dataRun2_relval, 106X_dataRun2_PromptLike_HEfail, 106X_dataRun2_PromptLike and 106X_dataRun2_PromptLike_HI)
- Keep 106X_mc2017_design_IdealBS and 106X_mc2017_realistic exactly as is
- For the new tags with names beginning with "electron" and "photon" in the 106X data queues, use the same payloads (there will be a technical change to allow for different synchronization requirements for the data and MC queues)

Sam-Harper:
> IMHO, "fine" is limited to consistently not knowingly worse performance.
> I do not think that the release is fine because these GTs/egamma payloads are apparently broken.

They have been broken for sometime.  I'm honestly surprised you see such a bad effect as I would have naively this regression while not perfect would be better than the existing regression for 2016/2018. But fine, something to understand. 

Anyways, I have added the modifiers for egamma now and only 2017 regression will get it at your instruction.  Before I submit, I want to run the matrix locally (its been compiling for a while now) just to check it doesnt immediately crash (very small local tests seemed to give the correct behaviour). Expect a push in a few hours. 

Then later I will finish my study on whether we can fix the ecalEnergyErr issue (again thank you for the spot) with a simple truncation of the inputs.

RecoEgamma/EgammaTools/python/regressionModifier_cfi.py
```diff
@@ -55,3 +142,11 @@
                                           uncertaintyKey_50ns = cms.vstring('gedphoton_EBUncertainty_50ns', 'gedphoton_EEUncertainty_50ns'),
                                           )
               )
+
+#by default we use the regression inappropriate to the main purpose of this release
+#life is simplier that way
+regressionModifier = regressionModifier94X.clone()
```
slava77: I thought that the EGRegressionModifierV3 can be used by all, that it was just a matter of having the right payload.
An earlier iteration of this PR was using the old payloads with EGRegressionModifierV3 and that gave the same electrons (and with the same intent but no fix, photons ).

slava77:
> This branch has conflicts that must be resolved

I suppose more commits are coming

Sam-Harper:
yes they are. happiness and joy.

slava77:
@cmsbuild please test with #27089

RecoEgamma/EgammaTools/python/regressionModifier_cfi.py
Sam-Harper: originally yes but I ended up deciding against it. It went through a few iterations, one where it lost the functionality but I think its back now. Anyways I dropped the idea as its not needed.

Sam-Harper:
so I have now pushed the changes to deal with the ecal energy error issue. It has been discussed in e/gamma today. 

https://indico.cern.ch/event/787355/contributions/3453254/attachments/1858756/3054001/egamma_2017ULRegressions_PPD.pdf (towards the end)

It is the setting I think are correct, not the settings which I've tested. Those tests are running and will complete tomorrow but I'm pretty sure they will be successful. If they are not, I'll change the python config numbers to the right value.

slava77:
@cmsbuild please test with #27089

slava77:
@cmsbuild please test with #27089

it looks like the last set of tests was aborted

Sam-Harper:
okay, the 1.2 TeV value had better performance for the endcap. Oh well, gamble failed. Will update it again shortly

Sam-Harper:
okay here is today's dose of joy and happiness in this never ending saga. 

When I set the old regression to the default for this release, my test jobs started picking up the old regression. So the less good performance I was getting was simply because I had swapped back to the old regression, which I discovered in some more validation of results. 

So now I have to re-run with what I meant to run. For added joy & happiness, crab is unusable at the moment due to the scheduler being down. 

As a  result, I've decided to push the 1.2 TeV setting as at least its a known acceptable setting and there is absolutely no eta on the validation for the 1.3 TeV setting being done.  If before this gets merged, I get the 1.3 TeV results and they are better, I'll update them, otherwise this is fine as is.  This has been agreed in the e/gamma-ecal meeting on friday.

fabiocos:
please test with #27089

Configuration/Eras/python/Era_Run2_2017_cff.py
```diff
@@ -14,8 +14,10 @@
 from Configuration.Eras.Modifier_run2_HLTconditions_2016_cff import run2_HLTconditions_2016
 from Configuration.Eras.Modifier_run2_muon_2017_cff import run2_muon_2017
 from Configuration.Eras.Modifier_run2_muon_2016_cff import run2_muon_2016
+from Configuration.Eras.Modifier_run2_egamma_2017_cff import run2_egamma_2017
+from Configuration.Eras.Modifier_run2_egamma_2016_cff import run2_egamma_2016
 
-Run2_2017 = cms.ModifierChain(Run2_2016.copyAndExclude([run2_muon_2016, run2_HLTconditions_2016]), 
+Run2_2017 = cms.ModifierChain(Run2_2016.copyAndExclude([run2_muon_2016, run2_HLTconditions_2016,run2_egamma_2017]),
```
slava77: this doesn't look right: it should be ```run2_egamma_2016``` here (of the things to be excluded from ```Run2_2016``` when making ```Run2_2017```

Configuration/Eras/python/Era_Run2_2018_cff.py
```diff
@@ -12,8 +12,11 @@
 from Configuration.Eras.Modifier_run2_HLTconditions_2017_cff import run2_HLTconditions_2017
 from Configuration.Eras.Modifier_run2_muon_2018_cff import run2_muon_2018
 from Configuration.Eras.Modifier_run2_muon_2017_cff import run2_muon_2017
+from Configuration.Eras.Modifier_run2_egamma_2018_cff import run2_egamma_2018
+from Configuration.Eras.Modifier_run2_egamma_2017_cff import run2_egamma_2017
 from Configuration.Eras.Modifier_run2_L1prefiring_cff import run2_L1prefiring
 
-Run2_2018 = cms.ModifierChain(Run2_2017.copyAndExclude([run2_HEPlan1_2017, run2_muon_2017, run2_L1prefiring, run2_HLTconditions_2017]),
+
+Run2_2018 = cms.ModifierChain(Run2_2017.copyAndExclude([run2_HEPlan1_2017, run2_muon_2017, run2_L1prefiring, run2_HLTconditions_2017, run2_egamma_2017]),
```
slava77: please drop the extra empty line

Configuration/Eras/python/Era_Run2_2017_cff.py
Sam-Harper: done, thanks for the catch

Configuration/Eras/python/Era_Run2_2018_cff.py
Sam-Harper: done

slava77:
please test with #27089

christopheralanwest:
unhold

I placed the hold because the GTs were incorrect but that has now been fixed.

slava77:
here are some plots based on 2017 MC in 10_1_0_pre1

In post cases I pick somewhat abnormal distributions.

In dijet flat-pt QCD (wf 10071)
the cluster uncertainty is blowing up
all_sign1080VSorig_QCD13TeVFlatPt15s3000in2017wf10071p0c_log10max1e-5,recoSuperClusters_particleFlowSuperClusterECAL_particleFlowSuperClusterECALBarrel_RECO_obj_correctedEnergyUncertainty
For about 180 SCs with E>200 the baseline has 0 with the uncertainty above 100, compared to 66 (1/3) with this PR.
I seem to recall the uncertainty mentioned as "can be zeroed out" in the PPD presentation last Thursday. Is this related, or do we have a similar problem as in the egamma regression?

The endcap looks less bad
all_sign1080VSorig_QCD13TeVFlatPt15s3000in2017wf10071p0c_log10max1e-5,recoSuperClusters_particleFlowSuperClusterECAL_particleFlowSuperClusterECALEndcapWithPreshower_RECO_obj_correctedEnergyUncertainty

For 1 TeV electrons the SC E uncertainty looks ~ OK
all_sign1080VSorig_SingleElectronPt1000in2017wf10003p0c_log10max1e-5,recoSuperClusters_particleFlowSuperClusterECAL_particleFlowSuperClusterECALBarrel_RECO_obj_correctedEnergyUncertainty

Compared to the earlier version  (which wasn't 2017 MC though) the electron energy error looks somewhat healthy in the 1 TeV ele gun sample
all_sign1080VSorig_SingleElectronPt1000in2017wf10003p0c_recoGsfElectrons_gedGsfElectrons__RECO_obj_correctedEcalEnergyError

In the same 1 TeV,
while the electron response apparently improves in bulk
wf10003el1Tev_pOgen
pfMET gets some extra small smearing (perhaps brem or secondary conversions are not that good) 
wf10003el1Tev_pfmetDiff
... this is small and is perhaps OK.

comparisons in photon pt 35 gun apparently show some overcorrection
for conversions
wf10005pho35_conv_eOtrue_eb
and/or for low R9
wf10005pho35_eOgen_reg1_lor9_eb
this looks pretty similar to what was posted last week in https://github.com/cms-sw/cmssw/pull/26892#issuecomment-499179510

Overall, it seems like only the SC energy uncertainty in jet data could be problematic,
while the rest is perhaps OK.

Sam-Harper:
Hi Slava,

Thank you for your checks. I'm still mystified about the photon one, my checks simply dont see this scale bias. 

The supercluster energy error in jets, it would be hard for me to care less about it :D Fakes look like badly measured electrons, this is one of the problems with the current and to a lesser extent this regression. This manifests in a large ecal uncertainty. If its not an electron, well the results will be off. Its not like this is used anywhere in CMSSW (I checked and the function is never referenced).  The 0ing was a different conversation because there we are putting in a bad value.  This is a good value which is improperly applied to fakes. 

As for the MET, I had rather hoped this time around this would improve things. Oh well.

slava77:
On 6/11/19 2:39 PM, Sam Harper wrote:
> The supercluster energy error in jets, it would be hard for me to care
> less about it :D Fakes look like badly measured electrons, this is one
> of the problems with the current and to a lesser extent this regression.
> This manifests in a large ecal uncertainty. If its not an electron, well
> the results will be off. Its not like this is used anywhere in CMSSW (I
> checked and the function is never referenced). The 0ing was a different
> conversation because there we are putting in a bad value. This is a good
> value which is improperly applied to fakes.
> 

I confirm that there are no use cases in CMSSW of the value saved in the
CaloCluster::correctedEnergyUncertainty_.
Is it correct that this uncertainty is not used in any other way by
directly calling the error function, which value is saved in the
supercluster e uncertainty?

slava77:
> Thank you for your checks. I'm still mystified about the photon one, my checks simply dont see this scale bias.

please clarify how your samples/setup were made.
Is this internal/private setup or were you using 106X_mc2017_realistic_v5 with full reco?

Sam-Harper:
20 million events of /DoublePhoton_FlatPt-5To300/RunIIWinter19PFCalibDR-2017ConditionsFlatPU0to70_105X_mc2017_realistic_v5-v1/AODSIM

I'm going to dig a bit deeper. Because this should be the final ECAL settings (except for intercalibration constants) and therefore should not change between 106X and 105X.

Sam-Harper:
Okay mystery solved. 

The conversions are filled here:
https://github.com/cms-sw/cmssw/blob/master/Validation/RecoEgamma/plugins/PhotonValidator.cc#L4450

This is the refined supercluster energy. In the future there will also be a refined energy correction, right now it gets nothing and relies on the PF cluster corrections (which is what I refer to as the one place where they matter for e/gamma).  But that is outside the scope of the UL and is what it is. Fixing it would result in a 2-3 week delay.  





> On 6/11/19 2:39 PM, Sam Harper wrote: The supercluster energy error in jets, it would be hard for me to care less about it :D Fakes look like badly measured electrons, this is one of the problems with the current and to a lesser extent this regression. This manifests in a large ecal uncertainty. If its not an electron, well the results will be off. Its not like this is used anywhere in CMSSW (I checked and the function is never referenced). The 0ing was a different conversation because there we are putting in a bad value. This is a good value which is improperly applied to fakes.
> I confirm that there are no use cases in CMSSW of the value saved in the CaloCluster::correctedEnergyUncertainty_. Is it correct that this uncertainty is not used in any other way by directly calling the error function, which value is saved in the supercluster e uncertainty?

Well the variable itself is never referenced except in the CaloCluster class (although its protected, but a quick search through supercluster and PFCluster shows no references to it ). I also looked on lxr as a check. 
https://cmssdt.cern.ch/lxr/ident?_i=correctedEnergyUncertainty_

The PF cluster one really makes no sense, as the  PF clusters are not really isolated prompt photons most of the time.  The SC makes some sense in some hypothetical use cases and it comes for free.

slava77:
> Likewise, the plot you are showing for the photon energy is the regression1 not regression2 energy. I have no idea what regression1 actually is though.

regression1 and regression2 plots are identical

It's not obvious to me how it gets set.
There is a way to overwrite 2 from 1
https://github.com/cms-sw/cmssw/blob/908b4dbce52f9cf0b5a9834d6a7fbd17990334dd/DataFormats/EgammaCandidates/src/Photon.cc#L108-L114

It is apparently happening in 
https://github.com/cms-sw/cmssw/blob/c1e01f06008132c15da8b9532dd5b921e5cb8b65/RecoEgamma/EgammaPhotonAlgos/src/PhotonEnergyCorrector.cc#L157-L164
this call happens before the final photon reg2 values are set. So, the net effect is apparently that reg1==reg2.

RecoEgamma/EgammaTools/plugins/EGRegressionModifierV3.cc
```diff
@@ -0,0 +1,398 @@
...
+  if(!pho.isEB() && maxRawEnergyForLowPtEESigma_>=0 && 
+     phoRegs_->ecalOnlySigma.useLowEtBin(rawEt,isSaturated)){
+    regData[0] = std::min(regData[0],maxRawEnergyForLowPtEESigma_);
+  }
+  const float ecalSigma = phoRegs_->ecalOnlySigma(rawEt,pho.isEB(),isSaturated,regData.data());
+    
+  const double corrEnergy = (rawEnergy + rawESEnergy)*ecalMeanCorr;
+  const double corrEnergyErr = corrEnergy*ecalSigma;
+
+  pho.setCorrectedEnergy(reco::Photon::P4type::regression2, corrEnergy, corrEnergyErr, true);
```
Sam-Harper: @slava77 , it gets set here.

slava77: but isn't it followed by https://github.com/cms-sw/cmssw/blob/6e96783e498e1d3ffad261b0bd8d67af2cab1483/RecoEgamma/EgammaPhotonProducers/src/GEDPhotonProducer.cc#L761-L773
?

slava77:
A bit more info.
In wf 10005 I select/scan
```
en->Scan("recoPhotons_gedPhotons__RECO.obj.et():recoPhotons_gedPhotons__RECO.obj[0].getCorrectedEnergy(reco::Photon::P4type::regression2):recoPhotons_gedPhotons__RECO.obj.eta():recoPhotons_gedPhotons__RECO.obj.phi()", "recoPhotons_gedPhotons__RECO.obj.r9()<0.94&&recoPhotons_gedPhotons__RECO.obj.isEB()&&recoPhotons_gedPhotons__RECO.obj.et()>30", "",1)
```
The result is 
``` 
* 35.598270 * 34.901893 * 0.0700971 * 2.7251212 *
```

Compared to the scan from the same event of the miniAOD output
```en->Scan("patPhotons_slimmedPhotons__RECO.obj.et().patPhotons_slimmedPhotons__RECO.obj.getCorrectedEnergy(reco::Photon::P4type::regression2):patPhotons_slimmedPhotons__RECO.obj.eta():patPhotons_slimmedPhotons__RECO.obj.phi()", "", "",1)```
I get
``` 
* 35.598270 * 35.685764 * 0.0700971 * 2.7251212 *
```

Is the difference in the reg2 corrected energy here expected?

slava77:
sorry, my bad, the last post had a typo in the root command line, which lead to a wrong number; I removed the post.

RecoEgamma/EgammaTools/plugins/EGRegressionModifierV3.cc
Sam-Harper: no. That is when you make the gedPhotonTmp collection. The gedPhotons get https://github.com/cms-sw/cmssw/blob/6e96783e498e1d3ffad261b0bd8d67af2cab1483/RecoEgamma/EgammaPhotonProducers/src/GEDPhotonProducer.cc#L918-L921 

This is an absolute mess this code, its almost impossible to figure out what on earth is going on.

Sam-Harper: I mean, this code here is absolutely text book example why I think your code reviews (while sometimes frustrating) are an absolute necessity and huge positive force for good in CMS and probably one of the most important things we can be doing to ensure the long term operation of our experiment

slava77:
Looking at the plot for low r9 by itself with a fit from 0.98 to 1.05
sign1080_10005_pho_eOtrue_lor9_eb
The mean of the histogram is below 1, the fit is showing that MPV is above 1, but the distribution is asymmetric.
Perhaps it's a matter of perception and the result is OK?
@Sam-Harper 
what do you think?

Sam-Harper:
Hmm it is a bit on  the edge. I need to understand why my tests dont match up. This does look funny.

Sam-Harper:
So on this, all my tests were of the regression application on 105X AOD which show things are perfect. However there could be a bug in RECO or something or something else in 106X reco.

Sadly I cant reco my 105X RAW samples in 106X due to GEM issues. I'm reproducing your test to see what happens but it'll take time.

However I think at this point, we should just go as is while trying to find the reason in parallel. This regression can always be reapplied in the re-miniAOD worst case. The important part is that jetMET get something sensible which comes at the reco step. So if we're 1% too high in low R9 photons, so be it I think and we fix it later.

slava77:
> 20 million events of /DoublePhoton_FlatPt-5To300/RunIIWinter19PFCalibDR-2017ConditionsFlatPU0to70_105X_mc2017_realistic_v5-v1/AODSIM
> 
> I'm going to dig a bit deeper. Because this should be the final ECAL settings (except for intercalibration constants) and therefore should not change between 106X and 105X.

Since this starts from AODSIM, just to be sure, is your workflow applying both the PFSC and the egamma regressions? (the matrix workflows that I use do both, the baseline has both the PF SC and egamma old regressions).

Sam-Harper:
the PF SC and PF cluster corrections should be irrelevant to this. 

The PF SC corrections arent even being applied to the Photons supercluster (which is the refined, not mustache).  I cant see where the PF cluster corrections would enter either. CaloCluster::energy() I believe is uncorrected.

Sam-Harper:
so I had a look, it enters in exactly one location, H/E where the E here is sensitive to the PF cluster corrections (basically the only bit of e/gamma that is).  Would find this surprising that this causes such an effect though.

Sam-Harper:
Oh and rho obviously as well. Hmm. Will be doing some studies.

slava77:
> so I had a look, it enters in exactly one location, H/E where the E here is sensitive to the PF cluster corrections (basically the only bit of e/gamma that is). Would find this surprising that this causes such an effect though.

my tests are without PU.
Would that matter?

Sam-Harper:
Hmm potentially yes. In theory we should be PU independent though.

slava77:
Just to discard the low-stats hypothesis, here are plots on 9K events:
sign1080_hs_10005_pho_eOtrue_lor9_eb
sign1080_hs_10005_pho_et_lor9_eb

The first one is from the photon validator.
The second is a simple plot of the photon et fitted around the 35 GeV range.
The MPV is about 0.7% above the expected, but OTOH, the distribution is a bit asymmetric and the simple mean falls below 1.0

Sam-Harper:
image

So this is #true interactions (more the inst lumi x minbias cross sec) < 0.5 for 30 to 40 GeV barrel photons with R9<0.94 reapplying to 1050 AOD samples. Now I have the low stats but I would say its fine.

slava77:
@Sam-Harper 
what's the functional form formula that I can use to fit quickly, (not in roofit)?

Sam-Harper:
Ah thats a double DCB implimented in roofitt. However  a single crystal ball is availble in normal root 
 
 `class CBFitFunc {
public:
  double operator()(double* x,double *p){
    //p0 = norm
    //p1 = mean 
    //p2 = sigma
    //p3 = alpha
    //p4 = n
    return p[0]*ROOT::Math::crystalball_function(*x,p[3],p[4],p[2],p[1]);
  }
};`

 TF1* fitFunc = new TF1("fitFunc",new CBFitFunc,0,3,5,"CBFitFunc");
  fitFunc->SetParameters(hist->Integral(),1,0.02,1,1);
  fitFunc->SetParLimits(1,0.9,1.1);
  fitFunc->SetParLimits(2,0.,0.2);
  fitFunc->SetParLimits(3,0.,100);
  fitFunc->SetParLimits(4,0.,100);
`

Sam-Harper:
the fit using that

image

slava77:
sign1080_hs_10005_pho_eOtrue_lor9_eb_CB
the CB fits well, but returns roughly the same 1.01.

I'm taking https://github.com/cms-sw/cmssw/pull/26892#issuecomment-501319309 
as an OK to move on with this PR.

slava77:
there is one more oddity:

In re-miniAOD of 2016 wf 136.7611 I see  some changes in OOT variables
all_OldVSNew_RunJetHT2016EreMINIAODwf136p7611c_patPhotons_slimmedOOTPhotons__PAT_obj_energyCorrections_scEcalEnergy
I do not see EGRegressionModifier\* in the config of this workflow.
I guess the diff is from PF SC corrections entering somewhere in the OOT reco chain.
So, this perhaps can be blamed on the choice of the GT for this workflow (auto:run2_data_relval is used , while a more appropriate should be something with payloads more consistent with what was used in the legacy 80X production).
@Sam-Harper @christopheralanwest what do you think?

Sam-Harper:
Thanks Slava, I agree we go as is. It does look like there is some interplay with PF calibrations from the OOT photons and I'll do some more tests tomorrow on this. But I think we can understand this in parallel and to be honest some relvals to debug would be useful as its something that cant be seen in the AOD applications.

slava77:
+1

for https://github.com/cms-sw/cmssw/pull/26892	02920af (in a combination with https://github.com/cms-sw/cmssw/pull/27089	a74a716)
- jenkins tests pass and comparisons with the baseline show acceptable differences
    - all data workflows are affected, even though normally only 2017 should have been affected. This is a feature of the GT PR affecting at this point only the PF SC corrections. This PR correctly applies the 2017 egamma corrections just to the 2017 parts.
        - For Run-1 case it is treated as an acceptable "damage" (or, rather, somewhat random change for a setup that has been inconsistent since a long time). 
        - For Run-2 2016 and 2018 the appropriately consistent calibrations are to be introduced later in the UL preparation. Otherwise, this data period is in a state somewhat similar to the Run-1 case.
    - supercluster corrected energy error for jet data (esp in the barrel) looks bad, but it's apparently not used
    - the OOT photon changes in the 80X reminiAOD workflow is likely a feature of the workflow using (since a while) an GT inconsistent with the 80X legacy rereco
    - the slight overshoot in response for photons with low r9 in the barrel is apparently acceptable (and in part could be a feature of tests done without PU, while the reference sample tests [not perfectly matching the UL setup] show that there is no problem)

fabiocos:
+operations

changes to Eras are coherent with the purpose of the PR

franzoni:
thanks a lot @slava77 @Sam-Harper @christopheralanwest
for the clarity of documentation along this PR.

For completeness and future reference, 
can we add equally clear statements about the impact (or better: lack thereof) on other scenarios such as run3, phase2, PbPb?
Tkx.

fabiocos:
+1

Sam-Harper:
so I hadnt noticed the non-2017 changes. There is nothing in this PR than can effect anything other than 2017! And there is nothing in the Alca changes that should effect anything but 2017. Looking into this now!

So Giovanni, the statement is there should be zero effect on run3, phase2, pbpb (unless they use a 2017 modifier)

Sam-Harper:
Ah okay hadnt had my coffee.  So its just data that is effected on other years as they share the same global tag which is just the 2017 corrections. Will see what we can do about that with some iovs to fix that.

fabiocos:
@Sam-Harper @christopheralanwest in order to finalize the 10_6_1 preparation, we will need a backport of this #26892 and of #27089 to the 10_6_X queue. In the meanwhile this code will run in the IBs

Sam-Harper:
okay expect it early in the afternoon

Sam-Harper:
so just to follow up on the scale shift at in the photons at low r9. 

I did the same pho 35 no pileup test. I confirm Slavas results (unsurprisingly). I also rule out an issue with the workflow in which I apply the regression on top of AOD (gave identical results as it should). 

I also did it with and without the PFCluster/SuperCluster corrections (and confirmed that things where different each time so I'm fully confident that the old corrections where being applied when I thought they were). 

In both cases the photon energy was almost identical (1000 dipho events) as there is a small weak dependence on the PF cluster corrections. Still negligible. There is room for error in the test but I'm confident at 2 sigma that its correct. 

So my conclusion: the scale shift we see in the tests is real. However its not present in 1050. It is not due to the new PF cluster or supercluster corrections in 1061

There are two possibilities, neither change the action which is : proceed as is.
1) its an artifact due to rho being zero causing the regression to give a scale shift (plausable) and will be fine in the pileup samples
2) something has fundamentally changed in the release causing this. In this case, we would need to retrain on 1061 samples. The timescale for this is too long and we'll just reapply at analysis/ reminiAOD level

slava77:
On 6/14/19 3:29 AM, Sam Harper wrote:
> There are two possibilities, neither change the action which is :
> proceed as is.
> 
>  1. its an artifact due to rho being zero causing the regression to give
>     a scale shift (plausable) and will be fine in the pileup samples
>  2. something has fundamentally changed in the release causing this. In
>     this case, we would need to retrain on 1061 samples. The timescale
>     for this is too long and we'll just reapply at analysis/ reminiAOD level
> 

It sounds like it would be good to rerun this with pileup.

PR #26825, opened 2019-05-17
refactor PFAlgo::processblock usePFPhotons part

jpata:
#### PR description:

Proposal to start refactoring the `PFAlgo::processBlock`, which is a single function of several thousand lines, into smaller independent subfunctions, based on physics functionality. As the original code is laid out quite clearly, simply in a long stream of consciousness, it is easy to factorize for example the part processing photons into a separate function. In a future iteration, we can factorize also the rest of the `processBlock` function. 
This can be followed by sanitizing the inputs and outputs of the functions and making sure they rely less on a global state.

#### PR validation:

We do not change the functionality of the PFAlgo code, we have validated by re-running the RECO sequence on 10k events and observed exactly the same PF reconstruction output. The full test matrix is running, I will report back with the output, I expect no new failures.

#### if this PR is a backport please specify the original PR:

This is a partial port of the original proposal PR in https://github.com/jpata/cmssw/pull/44.

PR #26894, opened 2019-05-22
refactored first part of PFAlgo::processBlock

jpata:
#### PR description:

Proposal to start refactoring the PFAlgo::processBlock, which is a single function of several thousand lines, into smaller independent subfunctions based on physics functionality. As the original code is laid out quite clearly, simply in a long stream of consciousness, it is easy to factorize.

At this stage, we tackle a part of the `PFAlgo::processBlock`, introducing the following factorizations:

-  `egammaFilters`: code in `PFAlgo::processblock { if(useEGammaFilters_) { ... } }`
-  `conversionAlgo`: code in `PFAlgo::processblock { if(usePFConversions_) { ... } }`
-  `elementLoop`: big loop in PF code ([line in code](https://github.com/cms-sw/cmssw/blob/master/RecoParticleFlow/PFProducer/src/PFAlgo.cc#L493)) that does multiple things:
  o Deals with tracks that are not associated to electrons or HCAL clusters
  o Sorts the ECAL and HCAL elements in separate vectors
  o For tracks which are connected to more than one HCAL cluster, cuts the links between the track and the cluster for all clusters but the closest one

This is meant as an inital PR to start clearing up the PFAlgo code, for which I imagine the path could be something like the following:

- [x] Factorize `PFAlgo::processBlock` to smaller subfunctions (this PR, partly)
- [ ] Sanitize the inputs and outputs of the functions and making sure they rely less on the PFAlgo state
- [ ] Given the cleaned-up implementation from above, revisit the logic/physics of the algo to rewrite it from with Run3/HL-LHC in mind
- [ ] Compare to an eventual ML PF algo that arrives in CMS

#### PR validation:

We do not change the functionality of the PFAlgo, hence the output should not change at all and timing performance should change only in as much as the C++ compiler is able to optimize the code better. We have checked this in a previous version of this PR, see [here](https://github.com/jpata/cmssw/pull/44). Nevertheless, I'm running the following comparisons:
- `runTheMatrix -l 38.0` between 43fd1cfad0b8e9e9a6abd0d1c049a1e860993887 and 3b1ae0778119c710e7bd3428e68a33b8d1fce84b: exactly the same reconstruction output on 10 events, output files have the same number of bytes.

- Complete implementation in 10_5_X: https://github.com/jpata/cmssw/pull/44.
- First cms-sw PR for the now obsolete `usePFPhotons_`: https://github.com/cms-sw/cmssw/pull/26825.

cc PF group: @hatakeyamak @bendavid

### DIFF
--- a/RecoParticleFlow/PFProducer/src/PFAlgo.cc
+++ b/RecoParticleFlow/PFProducer/src/PFAlgo.cc
@@ -446,7 +446,9 @@ void PFAlgo::conversionAlgo(const edm::OwnVector<reco::PFBlockElement> &elements
   }
 }
 
-void PFAlgo::elementLoop(const reco::PFBlock &block, reco::PFBlock::LinkData& linkData, const edm::OwnVector<reco::PFBlockElement> &elements, std::vector<bool>& active, const reco::PFBlockRef &blockref, ElementIndices& inds, std::vector<bool> &deadArea) {
+void PFAlgo::elementLoop(const reco::PFBlock &block, reco::PFBlock::LinkData& linkData,
+  const edm::OwnVector<reco::PFBlockElement> &elements, std::vector<bool>& active,
+  const reco::PFBlockRef &blockref, ElementIndices& inds, std::vector<bool> &deadArea) {
   for(unsigned iEle=0; iEle<elements.size(); iEle++) {
     PFBlockElement::Type type = elements[iEle].type();
 
@@ -495,7 +497,9 @@ void PFAlgo::elementLoop(const reco::PFBlock &block, reco::PFBlock::LinkData& li
     assert( !trackRef.isNull() );
 
     if (debug_ ) {
-      cout <<"PFAlgo:processBlock "<<" "<< inds.trackIs.size()<<" "<< inds.ecalIs.size()<<" "<< inds.hcalIs.size()<<" "<< inds.hoIs.size()<<endl;
+      cout <<"PFAlgo:processBlock "<<" "<< inds.trackIs.size()<<" "
+        << inds.ecalIs.size()<<" "<< inds.hcalIs.size()<<" "
+        << inds.hoIs.size()<<endl;
     }
 
     // look for associated elements of all types
@@ -1211,7 +1215,9 @@ void PFAlgo::elementLoop(const reco::PFBlock &block, reco::PFBlock::LinkData& li
   } // end of loop 1 on elements iEle of any type
 }
 
-int PFAlgo::decideType(const edm::OwnVector<reco::PFBlockElement> &elements, const reco::PFBlockElement::Type type, std::vector<bool>& active, ElementIndices& inds, std::vector<bool> &deadArea, unsigned int iEle) {
+int PFAlgo::decideType(const edm::OwnVector<reco::PFBlockElement> &elements,
+  const reco::PFBlockElement::Type type, std::vector<bool>& active,
+  ElementIndices& inds, std::vector<bool> &deadArea, unsigned int iEle) {
   switch( type ) {
   case PFBlockElement::TRACK:
     if ( active[iEle] ) {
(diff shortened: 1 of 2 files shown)


perrotta:
please test

RecoParticleFlow/PFProducer/src/PFAlgo.cc
```diff
@@ -461,122 +429,30 @@ void PFAlgo::processBlock( const reco::PFBlockRef& blockref,
...
-  vector<unsigned> ps2Is;
-
-  vector<unsigned> hfEmIs;
-  vector<unsigned> hfHadIs;
-
-  vector<bool> deadArea(elements.size(), false);
+  }
+}
 
+void PFAlgo::elementLoop(const reco::PFBlock &block, reco::PFBlock::LinkData& linkData, const edm::OwnVector<reco::PFBlockElement> &elements, std::vector<bool>& active, const reco::PFBlockRef &blockref, ElementIndices& inds, std::vector<bool> &deadArea) {
```
perrotta: Lines as such could be more better readable if splitted into some reasonable way.
This will be automatically done anyhow in the clang cleaning campaign ongoing

RecoParticleFlow/PFProducer/src/PFAlgo.cc
```diff
@@ -619,7 +495,7 @@ void PFAlgo::processBlock( const reco::PFBlockRef& blockref,
     assert( !trackRef.isNull() );
 
     if (debug_ ) {
-      cout <<"PFAlgo:processBlock "<<" "<<trackIs.size()<<" "<<ecalIs.size()<<" "<<hcalIs.size()<<" "<<hoIs.size()<<endl;
+      cout <<"PFAlgo:processBlock "<<" "<< inds.trackIs.size()<<" "<< inds.ecalIs.size()<<" "<< inds.hcalIs.size()<<" "<< inds.hoIs.size()<<endl;
```
perrotta: Also here some line splitting may help

RecoParticleFlow/PFProducer/src/PFAlgo.cc
```diff
@@ -1333,14 +1209,145 @@ void PFAlgo::processBlock( const reco::PFBlockRef& blockref,
       }
     } //loop hcal elements
   } // end of loop 1 on elements iEle of any type
+}
+
+int PFAlgo::decideType(const edm::OwnVector<reco::PFBlockElement> &elements, const reco::PFBlockElement::Type type, std::vector<bool>& active, ElementIndices& inds, std::vector<bool> &deadArea, unsigned int iEle) {
```
perrotta: Line splitting would allow a better readability also here

perrotta:
Please @jpata let us know whether you plan to apply here the style fixes pointed out in https://github.com/cms-sw/cmssw/pull/26894#pullrequestreview-242757200

jpata:
Done! I think clang reformat would be useful after this PR is done.

perrotta:
please test

hatakeyamak:
Thank you @jpata @perrotta .
Regarding elementLoop, well the other loops are also looping over elements, but this is the first loop to set things up for the subsequent loops. So, I don't feel strongly, but something like "initialElementLoop" or something similar can be slight improvement.

Otherwise this change looks good to me.

jpata:
Thanks for the suggestion @hatakeyamak. My take on this would be to spend a few iterations quickly over some days to factorize all the code functionally, also breaking apart `PFAlgo::elementLoop` as appropriate, and then figure out the durable function names, datastructures to pass around etc. It is possible that some functions may merge or disappear or change significantly, so I'm hesitant to spend a lot of time to choose the perfect names just now.

hatakeyamak:
@jpata Thankis. I understand that you plan to introduce additional functions for other loops and HF portion in the next PR. Probably you can consider my suggestions along the side of such subsequent step.

perrotta:
+1
- Refactoring implemented as planned
- No changes expected in output, jenkins tests pass and show no differences
- Further updates are planned for some forthcoming further update

fabiocos:
+1

PR #26896, opened 2019-05-23
Fix HGCal KDTreeLinkerAlgoT.h: in the non-leaf case both left and right sons must exist

perrotta:
#### PR description:

As pointed out by the static analyzer, there was the possibility (which shouldn't never happen by construction but still possible in practice) that the code tries to access a null pointer.
This can be avoided by enforcing the check that both pointers that the code tries to access are non-null.

#### PR validation:

The code compiles correctly

@felicepantaleo @rovere


perrotta:
please test

kpedro88:
+upgrade

perrotta:
+1
- Trivial fix which applies in some hypotetical case but with no effect on standard output
- Jenkins tests pass and show no differences with respect to the baseline

fabiocos:
+1

PR #26899, opened 2019-05-23
new scenario for Run 3 with flat distribution from 55 to 75

plujan:
A simple pileup scenario which just has a flat distribution for the average pileup (i.e., individual events will have a pileup number of events selected from a Poisson distribution whose mean is randomly selected from 55 to 75).


plujan:
just as a note -- since the automated checks don't actually test this new scenario, I tested it by hand using cmsDriver with --pileup Run3_Flat55To75_PoissonOOTPU to make sure it runs correctly, which it now does (after fixing the mistake above).

fabiocos:
please test

fabiocos:
@plujan as soon as you are satisfied with this PR, we need a backport in 10_6_X...

civanch:
+1

fabiocos:
+operations

fabiocos:
+1

PR #26905, opened 2019-05-23
Herwig7 fix seg fault master

Andrej-CMS:
#### PR description:
Dear all, 
this is a PR to fix the crash in official production with Herwig7 when the luminosityBlock changed.
Additionally, the PR adds parton shower weights for Herwig 7 as well as a corresponding configs that can be used in production.
Kind regards,
Andrej

#### PR validation:

To simulate the official production, the code was run on crab, which had the same issue. Just to be sure, we still need to test it in official production that the issue is indeed fully solved.
Output of an example job:
https://cmsweb.cern.ch/scheddmon/0197/cms1305/190523_121600:asaibel_crab_Generator_InitializeOnlyFirstLumi/job_out.3.0.txt

Before the changes, the jobs always failed after the change in luminosity block after 100events. You can see in the job that 350 events (two lumiBlock changes) are produced without failing.

The content of the PS weights was checked by checking the number of weights for each event in the GenEventInfoProduct

### DIFF
--- a/Configuration/Generator/python/Herwig7Settings/Herwig7LHECommonSettings_cfi.py
+++ b/Configuration/Generator/python/Herwig7Settings/Herwig7LHECommonSettings_cfi.py
@@ -12,6 +12,7 @@ herwig7LHECommonSettingsBlock = cms.PSet(
         'set LesHouchesHandler:DecayHandler /Herwig/Decays/DecayHandler',
         'set LesHouchesHandler:HadronizationHandler /Herwig/Hadronization/ClusterHadHandler',
         'set LesHouchesHandler:WeightOption VarNegWeight',
+        'set LesHouchesHandler:Weighted On',
         'set /Herwig/Generators/EventGenerator:EventHandler /Herwig/EventHandlers/LesHouchesHandler',
         'create ThePEG::Cuts /Herwig/Cuts/NoCuts',
         'create ThePEG::LHAPDF /Herwig/Partons/LHAPDF ThePEGLHAPDF.so',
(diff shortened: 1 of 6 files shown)


efeyazgan:
please test

efeyazgan:
+1

fabiocos:
@Andrej-CMS @efeyazgan was the output of the new code checked as well, making some DQM/Rivet comparison?

Andrej-CMS:
@fabiocos 
I didn't run these comparisons. Shouldn't this be part of the jenkins test routines to run work flow 511 ?

fabiocos:
@Andrej-CMS workflow 511 is not part of the short matrix, nobody triggered its run in this PR, and in any case we do not have a comparison for it even if we trigger its running... This is a basic check I would expect it is done a priori, looking at the output of the code before and after the update. I am not claiming that you need to expect problems, but given that the change is not totally minimal it would be better to check. Not crashing is good, but not enough to be correct.

Andrej-CMS:
Hello @fabiocos 
you can find the output of relval in the file below on lxplus. All tests show perfect agreement between the earlier version and this PR (except that the PR doesn't crash anymore when the lumiBlock changes).

I also pushed an updated version of workflow 535 to 10_2_X and 9_3_X. It would be nice to have this WF running when code related to the generators gets changed. WF 535 is very fast and takes less than a minute for 10 events.

/afs/cern.ch/work/a/asaibel/public/validation_10_2_14_OLD_vs_10_2_14_New/Generator.html

fabiocos:
+1

PR #26907, opened 2019-05-23
Skip tensorflow test that doesnt build on ppc64

mrodozov:
- Fixes build failures in the powerpc build
- Should solve this : https://cmssdt.cern.ch/SDT/cgi-bin/buildlogs/slc7_ppc64le_gcc700/CMSSW_11_0_X_2019-05-22-2300/PhysicsTools/TensorFlow so that we don't rebuild full IB every time
- Check if builds on intel and arm: builds
- Try the test on powerpc to check if cmssw PRs can be tested with different archs

### DIFF
--- a/PhysicsTools/TensorFlow/test/BuildFile.xml
+++ b/PhysicsTools/TensorFlow/test/BuildFile.xml
@@ -22,10 +22,10 @@
     <use name="PhysicsTools/TensorFlow" />
 </bin>
 
-
+<architecture name="!_ppc64le_.*">
 <bin file="tfadd_t.cpp">
   <flags DNN_NAME="test_graph_tfadd"/>
   <use name="tensorflow-runtime"/>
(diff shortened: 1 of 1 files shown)


mrodozov:
please test for slc7_ppc64le_gcc700

mrodozov:
please test
comparison failed (have to see why)

slava77:
+1

for https://github.com/cms-sw/cmssw/pull/26907	026f28b
- technical, affecting only ppc64le builds
- jenkins tests pass

smuzaffar:
@mrodozov , as discussed, we should not skip this test. We should try to understand why tfcompiler is not working on ppc.

smuzaffar:
hold
I would suggest to close this PR.

slava77:
-1

OK, I dropped my +1, which was provided assuming that this was reviewed already within the core group

mrodozov:
it's fine it's valid concern to fix the tests instead of skipping them.

PR #26915, opened 2019-05-23
[HGCAL] Perform HGCal Calibration only on the hard scatterer event

rovere:
#### PR description:

This PR will limit the HGCal Calibration to CaloParticles coming from the hard scatterer, excluding the PU contribution. This has the effect of making the plots understandable also using PU samples and a side effect of reducing the timing of that module from an average of **14s/event** to **0.2s/event** on a typical TTBar+200PU event.

Incidentally, it would be worth signalling that the current main offender in a typical Relval/runTheMatrix workflow with PU is the **OuterTrackerMonitorTrackingParticles** module, with its 400s/event. I do not know if this is known or, maybe, it is some weird effect that I happened to notice.

#### PR validation:

Run on TTbar and EGun RelVal samples of 10_6_0 with 200PU

### DIFF
--- a/Validation/HGCalValidation/plugins/HGCalHitCalibration.cc
+++ b/Validation/HGCalValidation/plugins/HGCalHitCalibration.cc
@@ -259,6 +259,13 @@ void HGCalHitCalibration::analyze(const edm::Event& iEvent,
   IfLogTrace(debug_ > 0, "HGCalHitCalibration")
     << "Number of caloParticles: " << caloParticles.size() << std::endl;
   for (const auto& it_caloPart : caloParticles) {
+    if (it_caloPart.g4Tracks()[0].eventId().event() != 0 or it_caloPart.g4Tracks()[0].eventId().bunchCrossing() != 0) {
+      LogDebug("HGCalHitCalibration") << "Excluding CaloParticles from event: "
+        << it_caloPart.g4Tracks()[0].eventId().event()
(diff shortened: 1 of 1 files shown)


jfernan2:
please test

rovere:
@kpedro88 maybe something you could be interested in.

kpedro88:
@rovere thanks, this looks like a useful speedup. I'll open an issue for the tracker module.

jfernan2:
+1

fabiocos:
@rovere fine for me to integrate this PR. I understand the CPU speedup argument driving it, but I understand less the argument of making the plots understandable. It depends of course about what you are interested to monitor, could you please clarify?

fabiocos:
+1

rovere:
@fabiocos 
These plots are meant to monitor the energy calibration of HGCAL and were meant to be used on particle gun samples. If you look at the plots produced using, e.g.,  TTbar+PU, you can still see the calibration peak at 1 but it sit on top of a huge background due to PU. 
The usefulness of the plots is to identify the MC truth coming from the hard-scatterer (ideally particle fun with HGCAL) and compare this against RECO, also with PU. This PR does exactly this, excluding PU from the quantities to be monitored to check for calibration. I hope this is more clear.

fabiocos:
@rovere ok I see what is your target

PR #26940, opened 2019-05-24
Turn off Cluster Repair for 0T data

OzAmram:
#### PR description:

This is a small PR to fix the issue that 0T data cannot be reco'ed currently due to the fact that 0T 2D templates do not exist.
This is currently causing an issue for Tier 0 replays: https://hypernews.cern.ch/HyperNews/CMS/get/tier0-Ops/2129/1/1/1/1.html

 Creating 2D 0T templates would take a significant amount of effort and there doesn't seem to be much reason to run the 2D reco for 0T data. So this fix has the 2D template ES producer always look for the default templates (instead of 0T templates) and modifies the Cluster Repair CPE to turn off the 2D reco if the mag field is not 3.8T.

We expect this PR to produce no changes except for 0T data.

### DIFF
--- a/CalibTracker/SiPixelESProducers/plugins/SiPixel2DTemplateDBObjectESProducer.cc
+++ b/CalibTracker/SiPixelESProducers/plugins/SiPixel2DTemplateDBObjectESProducer.cc
@@ -39,19 +39,9 @@ std::shared_ptr<const SiPixel2DTemplateDBObject> SiPixel2DTemplateDBObjectESProd
 	GlobalPoint center(0.0, 0.0, 0.0);
 	float theMagField = magfield.product()->inTesla(center).mag();
 
-    std::string label = "numerator";   // The correct default
-	//  std::string label = "denominator"; // Outdated. Used for MC in older GT's
-	//  std::string label = "";      // Outdated. Old default 
+        std::string label = "numerator";   // The correct default
+	if(theMagField>=4.1 || theMagField<-0.1) edm::LogWarning("UnexpectedMagneticFieldUsingDefaultPixel2DTemplate") << "Magnetic field is " << theMagField;
 
-	if(     theMagField>=-0.1 && theMagField<1.0 ) label = "0T";
-	else if(theMagField>=1.0  && theMagField<2.5 ) label = "2T";
-	else if(theMagField>=2.5  && theMagField<3.25) label = "3T";
-	else if(theMagField>=3.25 && theMagField<3.65) label = "35T";
-	else if(theMagField>=3.9  && theMagField<4.1 ) label = "4T";
-	else {
-		//label = "3.8T";
-		if(theMagField>=4.1 || theMagField<-0.1) edm::LogWarning("UnexpectedMagneticFieldUsingDefaultPixel2DTemplate") << "Magnetic field is " << theMagField;
-	}
 	ESHandle<SiPixel2DTemplateDBObject> dbobject;
 	iRecord.getRecord<SiPixel2DTemplateDBObjectRcd>().get(label,dbobject);
 
--- a/RecoLocalTracker/SiPixelRecHits/src/PixelCPEClusterRepair.cc
+++ b/RecoLocalTracker/SiPixelRecHits/src/PixelCPEClusterRepair.cc
@@ -81,14 +81,18 @@ PixelCPEClusterRepair::PixelCPEClusterRepair(edm::ParameterSet const & conf,
 	 << "\nERROR: Template ID " << forwardTemplateID_ << " not loaded correctly from text file. Reconstruction will fail.\n\n";
    }
    
-   templateDBobject2D_ = templateDBobject2D;
-   fill2DTemplIDs();
-   speed_ = conf.getParameter<int>( "speed");
-   LogDebug("PixelCPEClusterRepair::PixelCPEClusterRepair:") <<
-   "Template speed = " << speed_ << "\n";
-   
-   UseClusterSplitter_ = conf.getParameter<bool>("UseClusterSplitter");   
+   GlobalPoint center(0.0, 0.0, 0.0);
+   float theMagField = mag->inTesla(center).mag();
 
+   if(theMagField>=3.65 && theMagField<3.9){
+     templateDBobject2D_ = templateDBobject2D;
+     fill2DTemplIDs();
+     speed_ = conf.getParameter<int>( "speed");
+     LogDebug("PixelCPEClusterRepair::PixelCPEClusterRepair:") <<
+       "Template speed = " << speed_ << "\n";
+   }
+
+   UseClusterSplitter_ = conf.getParameter<bool>("UseClusterSplitter");   
 
 
    maxSizeMismatchInY_ = conf.getParameter<double>("MaxSizeMismatchInY");
@@ -102,6 +106,11 @@ PixelCPEClusterRepair::PixelCPEClusterRepair(edm::ParameterSet const & conf,
    recommend2D_.reserve(str_recommend2D.size());
    for (auto & str: str_recommend2D){
      recommend2D_.push_back(str);
+   }
+
+   // do not recommend 2D if theMagField!=3.8
+   if(theMagField>=3.65 && theMagField<3.9){
+     recommend2D_.clear();
    }
 
    // run CR on damaged clusters (and not only on edge hits)
--- a/RecoLocalTracker/SiPixelRecHits/src/PixelCPEClusterRepair.cc
+++ b/RecoLocalTracker/SiPixelRecHits/src/PixelCPEClusterRepair.cc
@@ -81,15 +81,16 @@ PixelCPEClusterRepair::PixelCPEClusterRepair(edm::ParameterSet const & conf,
 	 << "\nERROR: Template ID " << forwardTemplateID_ << " not loaded correctly from text file. Reconstruction will fail.\n\n";
    }
    
+   speed_ = conf.getParameter<int>( "speed");
+   LogDebug("PixelCPEClusterRepair::PixelCPEClusterRepair:") <<
+   "Template speed = " << speed_ << "\n";
+
    GlobalPoint center(0.0, 0.0, 0.0);
    float theMagField = mag->inTesla(center).mag();
 
    if(theMagField>=3.65 && theMagField<3.9){
      templateDBobject2D_ = templateDBobject2D;
      fill2DTemplIDs();
-     speed_ = conf.getParameter<int>( "speed");
-     LogDebug("PixelCPEClusterRepair::PixelCPEClusterRepair:") <<
-       "Template speed = " << speed_ << "\n";
    }
 
    UseClusterSplitter_ = conf.getParameter<bool>("UseClusterSplitter");   
@@ -109,7 +110,7 @@ PixelCPEClusterRepair::PixelCPEClusterRepair(edm::ParameterSet const & conf,
    }
 
    // do not recommend 2D if theMagField!=3.8
-   if(theMagField>=3.65 && theMagField<3.9){
+   if(theMagField<3.65 || theMagField > 3.9){
      recommend2D_.clear();
    }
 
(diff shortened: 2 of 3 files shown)


perrotta:
please test

RecoTracker/TransientTrackingRecHit/python/TTRHBuilderWithTemplate_cfi.py
```diff
@@ -11,7 +11,9 @@
 from Configuration.Eras.Modifier_trackingPhase2PU140_cff import trackingPhase2PU140
 trackingPhase2PU140.toModify(TTRHBuilderAngleAndTemplate, Phase2StripCPE = cms.string('Phase2StripCPE'))
 
-# uncomment these two lines to turn on Cluster Repair CPE
-from Configuration.Eras.Modifier_phase1Pixel_cff import phase1Pixel
-phase1Pixel.toModify(TTRHBuilderAngleAndTemplate, PixelCPE = cms.string('PixelCPEClusterRepair'))
+#Turn on cluster repair if not running at 0T
+from Configuration.StandardSequences.MagneticField_cff import ParabolicParametrizedMagneticFieldProducer
+if ParabolicParametrizedMagneticFieldProducer.valueOverride != 0:
```
makortel: Mostly out of curiosity, do I guess correctly that this trick works because
- for 0T jobs the generated configuration loads `MagneticField_0T_cff` instead of `MagneticField_cff`, and
- the generated configuration loads magnetic field cff before `Reconstruction_cff`

?

slava77: which jobs would these be?
```git grep MagneticField_0T_cff | grep -v /test``` doesn't show anything production-like.

Comment (OzAmram) [reply RecoTracker/TransientTrackingRecHit/python/TTRHBuilderWithTemplate_cfi.py]: Yes this was my idea behind the fix. 

There aren't any runTheMatrix tests with 0T (which is why this issue was not noticed before).
So I was testing with a config @mmusich provided : /afs/cern.ch/user/m/musich/public/forCristina/step3_RAW2DIGI_L1Reco_RECO_RECOSIM_EI_PAT_VALIDATION_DQM.py

I think it is standard to load the magnetic field before the reco, but I don't think its required right? So maybe there is some safer way to do this

Comment (slava77) [reply RecoTracker/TransientTrackingRecHit/python/TTRHBuilderWithTemplate_cfi.py]: I'm not sure what is the connection to the T0 replay.
Which job at T0 is this expected to represent?
Express or Prompt do not load MagneticField_0T_cff; the default is AutoFromDBCurrent

**Comment by davidlange6** [comment]
Right. Indeed the reco sequence cannot depend on having loaded a specific magnetic field cff. (and cannot override the field specified elsewhere)

On May 25, 2019 10:56 AM, Slava Krutelyov <notifications@github.com> wrote:

@slava77 commented on this pull request.

In RecoTracker/TransientTrackingRecHit/python/TTRHBuilderWithTemplate_cfi.py<https://github.com/cms-sw/cmssw/pull/26940#discussion_r287547084>:

> @@ -11,7 +11,9 @@
 from Configuration.Eras.Modifier_trackingPhase2PU140_cff import trackingPhase2PU140
 trackingPhase2PU140.toModify(TTRHBuilderAngleAndTemplate, Phase2StripCPE = cms.string('Phase2StripCPE'))

-# uncomment these two lines to turn on Cluster Repair CPE
-from Configuration.Eras.Modifier_phase1Pixel_cff import phase1Pixel
-phase1Pixel.toModify(TTRHBuilderAngleAndTemplate, PixelCPE = cms.string('PixelCPEClusterRepair'))
+#Turn on cluster repair if not running at 0T
+from Configuration.StandardSequences.MagneticField_cff import ParabolicParametrizedMagneticFieldProducer
+if ParabolicParametrizedMagneticFieldProducer.valueOverride != 0:


I'm not sure what is the connection to the T0 replay.
Which job at T0 is this expected to represent?
Express or Prompt do not load MagneticField_0T_cff; the default is AutoFromDBCurrent

File: RecoTracker/TransientTrackingRecHit/python/TTRHBuilderWithTemplate_cfi.py (hunk 1 above)
Comment (mmusich) [reply RecoTracker/TransientTrackingRecHit/python/TTRHBuilderWithTemplate_cfi.py]: @slava77,
that configuration pointed by Oz is not supposed to represent *any* job at Tier0.
It is actually a TTBar18 workflow modified to run without magnetic field. I stumbled in the same issue as the Tier0 replay when studying 0T reconstruction for totally orthogonal reasons, and notified the 2D template reco developers, before the replayed happened at all. 
Now the fix from Oz solves my problem, but unfortunately doesn't work in data as, as you point out, the magnetic field map is loaded according to the magnet current stored in the conditions.
@OzAmram, @cmantill I think the way to fix software-wise, if a calibration for 0T cannot be provided, is to modify the `ESProducer` not to look for labels which are not the default one **and** modify the C++ 2D template code not to run if the magnet current is not the nominal one.

Comment (slava77) [reply RecoTracker/TransientTrackingRecHit/python/TTRHBuilderWithTemplate_cfi.py]: I agree that the fix should be done in some other way, somewhere on the ES producer side.

Comment (OzAmram) [reply RecoTracker/TransientTrackingRecHit/python/TTRHBuilderWithTemplate_cfi.py]: Ok sorry for the confusion about the Tier 0.
Is there a test case any can provide then, what would represent what the Tier 0 would run? 

Is there any other way to check the magnetic field value in the event configuration then? 
I was hoping to avoid modifying the ESProducer and CPE to have a workaround in case of a different magnetic field, but if that is the only way it can be done.

Comment (mmusich) [reply RecoTracker/TransientTrackingRecHit/python/TTRHBuilderWithTemplate_cfi.py]: > Is there a test case any can provide then, what would represent what the Tier 0 would run?

Chris provided a nice recipe to reproduce the Tier-0 replay here:  https://hypernews.cern.ch/HyperNews/CMS/get/tier0-Ops/2129/1/1/1/1/1/1.html

Comment (cmantill) [reply RecoTracker/TransientTrackingRecHit/python/TTRHBuilderWithTemplate_cfi.py]: @mmusich ok I pushed my commits in this branch: https://github.com/cmantill/cmssw/tree/cpe-at-0T. Oz will cherry pick them up soon. I have tested this locally using both your recipe [1] and Chris's recipe (without modifying CPE to templateReco) [2].

[1] /afs/cern.ch/user/m/musich/public/forCristina/step3_RAW2DIGI_L1Reco_RECO_RECOSIM_EI_PAT_VALIDATION_DQM.py
[2] 
cmsrel CMSSW_10_6_0
cd CMSSW_10_6_0/src
cmsenv
scram b -j 4
cp /afs/cern.ch/work/v/vjankaus/public/20190523/Express-014a1f93-4f18-4b1c-9535-065483b74c94-2-logArchive.tar.gz  .
tar xf Express-014a1f93-4f18-4b1c-9535-065483b74c94-2-logArchive.tar.gz
cd job/WMTaskSpace/cmsRun1
edmConfigDump PSet.py > dump.py
sed -i 's|106X_dataRun3_Express_Candidate_2019_05_22_20_52_06|106X_dataRun3_Express_Candidate_2019_05_24_21_36_45|g' dump.py
cmsRun dump.py

**Comment by OzAmram** [comment]
Ok I took the commits from @cmantill for the alternative version of the fix. This now does not modify any configs, but rather has the 2D template ES producer only look for the standard templates and the CPE turns off CR if the field is not 3.8T.

**Comment by slava77** [comment]
@cmsbuild please test

**Comment by slava77** [comment]
> DQMHistoTests: Total failures: 21286

I was naively expecting that there will be no changes in the regular workflows.

**Comment by OzAmram** [comment]
Yes I was expecting there no changes to the regular workflows as well. I will try to find out what is causing the changes.

File: RecoLocalTracker/SiPixelRecHits/src/PixelCPEClusterRepair.cc [hunk 2]
```diff
@@ -104,6 +108,11 @@ PixelCPEClusterRepair::PixelCPEClusterRepair(edm::ParameterSet const & conf,
      recommend2D_.push_back(str);
    }
 
+   // do not recommend 2D if theMagField!=3.8
+   if(theMagField>=3.65 && theMagField<3.9){
+     recommend2D_.clear();
+   }
```
mmusich: Hi @OzAmram @cmantill, maybe I misread the code, but should not the logic of this clause be reversed?
Right now you are clearing `recommend2D_` if the magnetic field **is** the nominal one.

cmantill: @mmusich you are right, sorry for the bug. It was introduced when I was cleaning up the code after testing and did an incorrect translation of the logic.. https://github.com/cmantill/cmssw/commit/c78b14819bdfac71bdc324b062d160b13845a962#diff-34f19ce9486ae317e412a733351d2274R112. I pushed a fix and Oz will pick it up asap.

OzAmram:
Ok I just pushed a commit that fixes that bug. I will test the 0T workflows again

slava77:
@cmsbuild please test

slava77:
+1

for https://github.com/cms-sw/cmssw/pull/26940	db03cb9
- code changes are in line with the PR description and the follow up review. Changes for non-standard magnetic field values visually appear to be correct.
- jenkins tests pass and comparisons with the baseline show no differences (the default behavior for 3.8 T is unchanged)

CalibTracker/SiPixelESProducers/plugins/SiPixel2DTemplateDBObjectESProducer.cc
```diff
@@ -39,19 +39,9 @@ std::shared_ptr<const SiPixel2DTemplateDBObject> SiPixel2DTemplateDBObjectESProd
...
-	else if(theMagField>=1.0  && theMagField<2.5 ) label = "2T";
-	else if(theMagField>=2.5  && theMagField<3.25) label = "3T";
-	else if(theMagField>=3.25 && theMagField<3.65) label = "35T";
-	else if(theMagField>=3.9  && theMagField<4.1 ) label = "4T";
-	else {
-		//label = "3.8T";
-		if(theMagField>=4.1 || theMagField<-0.1) edm::LogWarning("UnexpectedMagneticFieldUsingDefaultPixel2DTemplate") << "Magnetic field is " << theMagField;
-	}
+        std::string label = "numerator";   // The correct default
+	if(theMagField>=4.1 || theMagField<-0.1) edm::LogWarning("UnexpectedMagneticFieldUsingDefaultPixel2DTemplate") << "Magnetic field is " << theMagField;
```
christopheralanwest: Should these values be changed to 3.9 and 3.65, respectively, to match the values in 
[PixelCPEClusterRepair.cc#L91](https://github.com/cms-sw/cmssw/blob/db03cb92a0da0c62ee5a88ab20a4d185c60b7a81/RecoLocalTracker/SiPixelRecHits/src/PixelCPEClusterRepair.cc#L91)?

mmusich: I think the numerical limits in the two classes are serving well distinct purposes:
- here an `edm::LogWarning` warning is issued if the magnetic field strength passed to the code is physically "impossible" (`<0` ,` >4`)
- in [PixelCPEClusterRepair.cc#L91](https://github.com/cms-sw/cmssw/blob/db03cb92a0da0c62ee5a88ab20a4d185c60b7a81/RecoLocalTracker/SiPixelRecHits/src/PixelCPEClusterRepair.cc#L91) those are used to decide when to turn off the CluterRepair CPE treatment due to the lack of an appropriate calibration for the not fully ON cases.  

Moreover if your suggestion is applied we'll have a warning from the `ESProducer` every time we collect 0T data (which is a completely normal and necessary data-taking mode).

christopheralanwest: OK, thanks for the explanation. Indeed, I was merely wondering if the two sets of numerical limits serve distinct purposes.

christopheralanwest:
+1

fabiocos:
@OzAmram @mmusich @boudoul even if MWGR2 has been canceled, I expect that this PR will be sooner or later needed in 10_6_X, so I would suggest to prepare a backport

mmusich:
> @OzAmram @mmusich @boudoul even if MWGR2 has been canceled, I expect that this PR will be sooner or later needed in 10_6_X, so I would suggest to prepare a backport

@fabiocos, I was actually going to suggest that once the PR to master is merged (and passed through an IB).

OzAmram:
To do a backport, I need to make another PR with the base as CMSSW_10_6_X right? I can do that once it  is added to the next IB

fabiocos:
+1

christopheralanwest:
@OzAmram Yes, to make a backport, you need to make another PR based on CMSSW_10_6_X.

OzAmram:
Ok I made another PR with the backport here: https://github.com/cms-sw/cmssw/pull/27044

PR #26941, opened 2019-05-24
HCALDQM: Updates for Run 3 HB

DryRun:
#### PR description:

This PR implements Run 3 HB in HCAL DQM. The old HBHEDigiCollection is removed, and the new HB digis are taken from QIE11DigiCollection. Since HB now overlaps with HO in (ieta, iphi, depth) space, HO digis are plotted in depth 7 for now. A future PR will improve on that hack, moving HO (and perhaps all subsystems) to their own ieta-iphi plots, rather than attempting to plot all subsystems together. 

#### PR validation:

Successfully ran over recent local and global runs with partial HB installation. Note that running in online DQM will require appropriate conditions, which are being developed in parallel.

### DIFF
--- a/DQM/HcalCommon/src/DetectorQuantity.cc
+++ b/DQM/HcalCommon/src/DetectorQuantity.cc
@@ -26,7 +26,8 @@ namespace hcaldqm {
 
     uint32_t getBin_ieta(HcalDetId const &did) { return (uint32_t)(getValue_ieta(did) + 1); }
 
-    uint32_t getBin_depth(HcalDetId const &did) { return (uint32_t)(did.depth()); }
+    uint32_t getBin_depth(HcalDetId const &did) { return (uint32_t)(
+      did.subdet() == HcalOuter ? 7 : did.depth()); }
 
     uint32_t getBin_Subdet(HcalDetId const &did) { return (uint32_t)(did.subdet()); }
 
--- a/DQM/HcalCommon/src/DetectorQuantity.cc
+++ b/DQM/HcalCommon/src/DetectorQuantity.cc
@@ -26,8 +26,9 @@ namespace hcaldqm {
 
     uint32_t getBin_ieta(HcalDetId const &did) { return (uint32_t)(getValue_ieta(did) + 1); }
 
-    uint32_t getBin_depth(HcalDetId const &did) { return (uint32_t)(
-      did.subdet() == HcalOuter ? 7 : did.depth()); }
+    uint32_t getBin_depth(HcalDetId const &did) { //return (uint32_t)(did.depth());}
+      return (uint32_t)(did.subdet() == HcalOuter ? 7 : did.depth());
+    }
 
     uint32_t getBin_Subdet(HcalDetId const &did) { return (uint32_t)(did.subdet()); }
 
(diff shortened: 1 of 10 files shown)


abdoulline:
@DryRun   
David, I suppose backporting it for MWGR#2 is planned?

jfernan2:
please test

jfernan2:
@DryRun it seems there are many RelVals with errors due to this PR, can you please have a look? Thanks

fabiocos:
@DryRun as the DQM is running on many different input years, is this updated code compatible?

abdoulline:
... that was my qiestion (about back-compatibility, era-wise ) to David 
in HCAL e-mail thread, didn't get an answer yet.

DryRun:
Hi @abdoulline - replying here and in the thread. I'm not sure about strict backwards compatibility -- e.g. for HEP17 and HE 2018, we just ensured that DQM would run over old data without crashing, not that it would reproduce the histograms. If that's desired for UL, I can try to incorporate that via eras (but note that we already don't run over HE digis in 2016 and 2017 data, so that would have to be restored).

abdoulline:
@DryRun    
Hi David, I'd personally think _not crashing_ is enough for DQM, which is primarily for data-taking online and Prompt. We're using  OfflineDQM/Hcal (era-customized) for UL.

DryRun:
Hi @jfernan2 , I believe I've fixed the crash in the RelVals. It was due to a filter intended to select HO channels doing the opposite. Could you relaunch the testing?

Also @fabiocos - sorry to have missed your message here. This PR should be backwards compatible, in the sense of not crashing on inputs from different years. As said above, I've deleted the code that runs on Run1 and Run2 HB/HE digis; if it's desired to preserve that functionality, I can put it back with an eras specifier, but I don't think that's needed for the online/prompt DQM modules.

jfernan2:
please test

jfernan2:
Hi @DryRun I am afraid there is some issue still:
Exception Message:
Requested conditions of type HcalQIEData for cell (0x49387427) (HF 29,39,3) got conditions for cell (0x49187827) (HF 30,39,1)

abdoulline:
@DryRun  
David, we shouldn't have HF depth3 (HF 29,39,3) in 2016 Geometry from Run2 era, right? 
There was no HF dual readout  (with 4 depths) in 2016 yet, it arrived in 2017.

DryRun:
Hi @abdoulline - indeed, in 2016 there was only a pilot dual anode channel in HF, but if I recall correctly, it was in ieta=41 (+/-?). Investigating again...

abdoulline:
Ah... indeed, there were two HF PMTs (in addition 22 others split in QIE10 and QIE8 with only QIE8 channels in emap, so no problem at the point)  with both anodes -> QIE10 
See slide #10
https://indico.cern.ch/event/505065/contributions/2166338/attachments/1334807/2013510/HCAL_Phase1_Upgrade_S.Abdullin.pdf
Now we need to figure out which channels and whether they were included in emap...

abdoulline:
@DryRun 
found, sent you e-mail

DryRun:
I've identified the problem, but unfortunately this PR is based on an IB that no longer exists, and `git fetch` doesn't seem to work. I'm not sure how to fix that, so I'll close this PR and opening a new one.

PR #26945, opened 2019-05-26
Fix the storage of the header record from the SiPixel template in case it is less than 80 chars long

perrotta:
#### PR description:

As catched by the static analyzer automatically run during the standar PR tests, there would be garbage values assigned in case the header record of the template calibration file contains a carriage return before its 80th character.

This PR contains a proposal to fix that hypothetical case, which had quite likely never happened (otherwise I'd expect we would have encountered it already in current tests and productions).

Please @pmaksim1 @OzAmram verify that this does not conflict with the original purpose of this code.

#### PR validation:

It compiles, and short matrix tests run to completion.
No differences are expected in output, although this was not actually checked explicitly.

### DIFF
--- a/CalibTracker/SiPixelESProducers/plugins/SiPixelFakeGenErrorDBObjectESSource.cc
+++ b/CalibTracker/SiPixelESProducers/plugins/SiPixelFakeGenErrorDBObjectESSource.cc
@@ -59,10 +59,9 @@ std::unique_ptr<SiPixelGenErrorDBObject> SiPixelFakeGenErrorDBObjectESSource::pr
 			  title_char[iter] ='\n';
 			} else  {
 			  for (unsigned int it=0; it!=3-(iter%4); it++) {
-			    iter++;
 			    title_char[iter] =' ';
+			    iter++;
 			  }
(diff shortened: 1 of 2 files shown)


perrotta:
please test

pmaksim1:
This looks good.  Tomorrow is a holiday in the US.  On Tuesday, we can run this by Morris, and if he has no objections, we can get this integrated.

perrotta:
please test

pmaksim1:
@perrotta 
Morris says this is OK, so please merge at your earliest convenience.  (For the sake of expediency, you can skip the comment and we can add it later.)  Thanks!

perrotta:
> @perrotta
> Morris says this is OK, so please merge at your earliest convenience. (For the sake of expediency, you can skip the comment and we can add it later.) Thanks!

Thank you @pmaksim1 . 
This has to be signed by Alca first: @christopheralanwest, @tocheng, @cmsbuild, @franzoni, @tlampen, @pohsun

pohsun:
+1

fabiocos:
+1

PR #26960, opened 2019-05-27
[SIMULATION] fixes needed to build cxxmodules

smuzaffar:
#### PR description:

- Added missing include to avoid clang error about unknown type `uint32_t`


smuzaffar:
please test

civanch:
+1

DataFormats/HcalDetId/interface/HcalTestNumbering.h
```diff
@@ -6,6 +6,7 @@
 ///////////////////////////////////////////////////////////////////////////////
 
 #include <boost/cstdint.hpp>
+#include <stdint.h>
```
kpedro88: usually we recommend using `<c*>` over `<*.h>` headers for STL.

In addition, this file already includes `<boost/cstdint.hpp>`, which seems to define `uint32_t`. I am confused why the specific added include is necessary.

smuzaffar: @kpedro88 , when CXX Modules are generated (using clang) then clang complains about unknown type `uint32_t`

fabiocos: @smuzaffar ok, but is this behaviour related to the use of <*.h> vs <c*> ? I would naively expect it is not...

civanch: @smuzaffar , we have in many other places

#include <cstdint>

likely it is fully equivalent, do we have a guideline what to use?

smuzaffar: I will try cstdint and update the PR if works

fabiocos: @civanch I do not think we have any official written guideline, I think this is more a habit, @smuzaffar @davidlange6 do you confirm?

smuzaffar:
please test

DataFormats/HcalDetId/interface/HcalTestNumbering.h
smuzaffar: fixed, it was enough to releance boost/cstdint with csdtint

civanch:
+1

fabiocos:
@smuzaffar thanks!

fabiocos:
+1

PR #26999, opened 2019-05-29
TkAl All in One tool: don't import crabWrapper

hroskes:
#### PR description:

Basically: don't mess with the environment.  This caused 10+ messages like
```
sh: module: line 1: syntax error: unexpected end of file
sh: error importing function definition for `BASH_FUNC_module'
```
every time we ran the code.

On slc7 this was a more serious issue, because these lines were added to the output of the call to [dasgoclient](https://github.com/cms-sw/cmssw/blob/5b62bf9f89f2df1e899a0d5c813efe6c1c5b35e5/Utilities/General/python/cmssw_das_client.py#L85) and causing it not to parse because it wasn't a valid json string.

#### PR validation:

Certain segments of validateAlignments won't work anymore because they're missing this import, but we don't use the configuration options that lead to those sections.  Gregor Mittag tried to implement crab a long time ago but it never worked properly.  I am not deleting crabWrapper and associated functionality from `validateAlignments.py` yet in case this conflicts with the condor migration, but if it doesn't, as far as I'm concerned someone can go ahead and delete it.

@adewit @connorpa @mmusich

### DIFF
--- a/Alignment/OfflineValidation/scripts/validateAlignments.py
+++ b/Alignment/OfflineValidation/scripts/validateAlignments.py
@@ -11,7 +11,6 @@ import fnmatch
 import six
 import Alignment.OfflineValidation.TkAlAllInOneTool.configTemplates \
     as configTemplates
-import Alignment.OfflineValidation.TkAlAllInOneTool.crabWrapper as crabWrapper
 from Alignment.OfflineValidation.TkAlAllInOneTool.TkAlExceptions \
     import AllInOneError
 from Alignment.OfflineValidation.TkAlAllInOneTool.helperFunctions \


christopheralanwest:
please test

christopheralanwest:
@adewit @connorpa @mmusich AlCa cannot review this PR in any sensible way. Whether or not it should be merged depends more on whether the subsystem experts want the change. AlCa will sign off if you confirm that the change is indeed desired.

adewit:
I'm assuming that @hroskes verified that all the parts of validateAlignments that we actually use still work without this, so fine for me from that point of view. Just to be absolutely sure I'd like @henriettepetersen and @DaveBrun94 can confirm that this import is not used in any way in the HTCondor setup they're working on (I don't think it is, but better safe than sorry ;-) )

hroskes:
Yes, the only places where this import is used are sections of the code that only happens if you put `jobmode = crab` or if add `--status` on the command line, which is also only supposed to be used to check crab jobs.  I didn't want to create conflicts, but Henriette and David can go ahead and delete those lines.

DaveBrun94:
@adewit We can confirm that this import is not used in the HTCondor setup

christopheralanwest:
+1

fabiocos:
+1

PR #27000, opened 2019-05-29
correct missing commas in dflt initializations for ghostTrack in IPProducer

slava77:
This is a follow up to https://github.com/cms-sw/cmssw/issues/26913#issuecomment-497027588

@ferencek

### DIFF
--- a/RecoBTag/ImpactParameter/plugins/IPProducer.h
+++ b/RecoBTag/ImpactParameter/plugins/IPProducer.h
@@ -391,11 +391,11 @@ IPProducer<Container,Base,Helper>::produce(edm::Event& iEvent, const edm::EventS
                                                         VertexState(p2, e2));
            trackIP.ghostTrackWeight = pos->weight();
          } else {
-           trackIP.distanceToGhostTrack = Measurement1D(-1. -1.);
+           trackIP.distanceToGhostTrack = Measurement1D(0., -1.);
            trackIP.ghostTrackWeight = 0.;
          }
        } else {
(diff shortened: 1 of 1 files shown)


slava77:
@cmsbuild please test

slava77:
as expected, in a few cases the ghostTrack tag infos moved from -2 to 0.
E.g. in Run2018A EGamma wf 136.85

all_OldVSNew_RunEGamma2018Awf136p85c_recoCandidateedmPtrsrecoJetTagInforecoIPTagInfos_pfImpactParameterTagInfos__reRECO_obj_m_data_distanceToGhostTrack_value

slava77:
+1

for https://github.com/cms-sw/cmssw/pull/27000	de52341
- differences observed in the jenkins tests are as expected: in some rare cases where the defaults are not reset the values are now at 0 instead of -2

fabiocos:
@slava77 this looks a genuine bug fix that I assume we would like to port to 10_6_X, am I correct?

fabiocos:
+1

slava77:
> @slava77 this looks a genuine bug fix that I assume we would like to port to 10_6_X, am I correct?

I do not really see a reason for a backport.
From https://github.com/cms-sw/cmssw/issues/26913#issuecomment-496877602
> The ghost track feature is not really in active use which probably explains why this went unnoticed all this time.

fabiocos:
@slava77 ok, in this case I agree, although strictly speaking it remains a bug

PR #27090, opened 2019-06-04
Modified OuterTrackerMonitorTrackingParticles to run much faster

### LINKED ISSUE #26932: Speed up OuterTrackerMonitorTrackingParticles with high PU
In #26915 it was reported that the module [OuterTrackerMonitorTrackingParticles](https://github.com/cms-sw/cmssw/blob/master/Validation/SiOuterTrackerV/plugins/OuterTrackerMonitorTrackingParticles.cc) takes ~400 s/evt in high PU.

It looks like there is a nested loop, which could be the cause. There may be a way to perform the same operations more efficiently. attn: @emacdonald16

emacdonald16:
#### PR description:
Heavily modified nLayers variable; runs much faster now (~2.5 s/evt at 200PU, compared to ~400)
resolves #26932 
#### PR validation:
Compared output file before and after changes, and they were identical. Only difference is in run time.

### DIFF
--- a/Validation/SiOuterTrackerV/interface/OuterTrackerMonitorTrackingParticles.h
+++ b/Validation/SiOuterTrackerV/interface/OuterTrackerMonitorTrackingParticles.h
@@ -29,7 +29,6 @@ public:
   ~OuterTrackerMonitorTrackingParticles() override;
   void analyze(const edm::Event &, const edm::EventSetup &) override;
   void bookHistograms(DQMStore::IBooker &, edm::Run const &, edm::EventSetup const &) override;
-  int Layer(const float R_, const float Z_) const;
 
   // Tracking particle distributions
   MonitorElement *trackParts_Eta = nullptr;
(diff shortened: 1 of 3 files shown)


jfernan2:
please test

jfernan2:
@emacdonald16 does this PR need a backport?

emacdonald16:
@jfernan2 A backport would probably be a very good idea, if it's the main offender for a typical runTheMatrix workflow with PU, at least for CMSSW_10_6_X

jfernan2:
> @jfernan2 A backport would probably be a very good idea, if it's the main offender for a typical runTheMatrix workflow with PU, at least for CMSSW_10_6_X

I would definitely vote for it, please

jfernan2:
+1

fabiocos:
@emacdonald16 @jfernan2 are the differences observed in the DQM comparisons understood and expected? The descriptions says there should be none...

jfernan2:
@fabiocos most of the DQM differences are due to fine tuning of the Bin by Bin tool (WIP to fully black list the EventInfo checks), with the exception of different number of events processed for 13 MEs in Phase2 workflows, e.g.
http://tinyurl.com/yy23luoo

@emacdonald16 can you comment on those? Thanks

emacdonald16:
@fabiocos @jfernan2 The differences are understood. I modified a few other things in the code as well, that have no effect when the L1 TTTracks are included (which isn't the case in central CMSSW, only when run privately), but allow you to now see the denominator of the efficiency plots that are made from the tracking particles if the TTTracks are not included.

fabiocos:
+1

PR #27116, opened 2019-06-05
Validation fixes

perrotta:
#### PR description:

Starting from some issue reported by the static analyzer, I think there are two real bugs in these Validation files which are fixed here:
- L732 of Validation/EventGenerator/plugins/TauValidation.cc, where with the original formulation the case with two negative and one positive pion is never hit
- The fix in  Validation/EventGenerator/src/HepMCValidationHelper.cc 

A few other minor adjustments are also included here.

Please check these fixes, and accept them if you think they are correct.
I expect some difference in the affected Validation outputs

#### PR validation:

It compiles

### DIFF
--- a/Validation/EventGenerator/plugins/TauValidation.cc
+++ b/Validation/EventGenerator/plugins/TauValidation.cc
@@ -650,19 +650,19 @@ void TauValidation::countParticles(const reco::GenParticle *p,
     allCount++;
     if (abs(pid) == 11)
       eCount++;
-    if (abs(pid) == 13)
+    else if (abs(pid) == 13)
       muCount++;
-    if (abs(pid) == 111)
+    else if (abs(pid) == 111)
       pi0Count++;
-    if (abs(pid) == 211)
+    else if (abs(pid) == 211)
       piCount++;
-    if (abs(pid) == 213)
+    else if (abs(pid) == 213)
       rhoCount++;
-    if (abs(pid) == 20213)
+    else if (abs(pid) == 20213)
       a1Count++;
-    if (abs(pid) == 321)
+    else if (abs(pid) == 321)
       KCount++;
-    if (abs(pid) == 323)
+    else if (abs(pid) == 323)
       KstarCount++;
     countParticles(dau, allCount, eCount, muCount, pi0Count, piCount, rhoCount, a1Count, KCount, KstarCount);
   }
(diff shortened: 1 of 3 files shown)


perrotta:
please test

perrotta:
type bugfix

jfernan2:
@perrotta the only changes I spot are in Generator/MBUEandQCD   e.g.

http://tinyurl.com/y34vvpay

Are they what you expect?

perrotta:
> @perrotta the only changes I spot are in Generator/MBUEandQCD e.g.
> 
> http://tinyurl.com/y34vvpay
> 
> Are they what you expect?

Maybe the tau validation touched is not run centrally...

Yes, I do expect some change in Generator DQM, because of the bug fix in https://github.com/cms-sw/cmssw/pull/27116/files#diff-4125b620f030e56d690fab0dadaab92cR732. However, I am not able to evaluate whether they are correct or not, an expert from Generator DQM should check and bless it. And an expert from Tau as well, even if that code is not run centrally.

All other changes in this PR should not affect the final results, instead

jfernan2:
+1

fabiocos:
@perrotta @jfernan2 the tau validation should be run as well

https://cmssdt.cern.ch/lxr/source/Configuration/StandardSequences/python/Validation_cff.py#0047
https://cmssdt.cern.ch/lxr/source/Validation/EventGenerator/python/BasicGenValidation_cff.py#0049

but I may imagine that on 10 events it is not obvious to have a tau decay and see a difference. This might be easily tested with more events in GEN-only workflows like 505.0 (DY to LL)

perrotta:
@fabiocos 

I run on 500 events with fw 505, and compared the DQM output with the one of the baseline.

Still I don't see differences in the tau validation plots, while some better resolution can be obtained in the generator ones, that I attach here for reference:

DJR10

DJR21

DJR32

DJR43

perrotta:
On the other hand, the plots ```TauSpinEffectsWUpsilonA1``` and  ```TauSpinEffectsHpmUpsilonA1``` are always empty also in that sample, and therefore they are trivially identical.

Maybe the fix (which looks rather trivial to me, just a fix of bugged c+= code) should be rather evaluated by a generator or tau expert...

fabiocos:
@perrotta those histograms depend on the mother of the tau, for Ws one may use 506.0 and indeed I see on a few hundred events a difference.

Anyway I agree that @efeyazgan @qliphy @alberto-sanchez GEN experts should have an independent look and sign or comment this PR

fabiocos:
@efeyazgan @qliphy @alberto-sanchez  ping

efeyazgan:
+1

fabiocos:
+1

fabiocos:
@efeyazgan @perrotta should we consider a backport to 10_6_X? I may imagine that GEN validation might be useful there for a while...

PR #27084, opened 2019-06-03
 Validation/Configuration/python/autoValidation.py Phase2 modification

archiron:
#### PR description:

add electronValidation sequence in Phase2 because RECO branches did not appear into the root files (only the miniAOD ones). One line added in Validation/Configuration/python/autoValidation.py and one one sequence (electronOnlyValidation) added in autoValidation. 

#### PR validation:

tests with :

27434.0_TTbar_14TeV+TTbar_14TeV_TuneCUETP8M1_2023D35_GenSimHLBeamSpotFull14+DigiFullTrigger_2023D35+RecoFullGlobal_2023D35+HARVESTFullGlobal_2023D35/
29434.0_TTbar_14TeV+TTbar_14TeV_TuneCUETP8M1_2023D42_GenSimHLBeamSpotFull14+DigiFullTrigger_2023D42+RecoFullGlobal_2023D42+HARVESTFullGlobal_2023D42/
20434.0_TTbar_14TeV+TTbar_14TeV_TuneCUETP8M1_2023D19_GenSimHLBeamSpotFull14+DigiFullTrigger_2023D19+RecoFullGlobal_2023D19+HARVESTFullGlobal_2023D19/
20034.0_TTbar_14TeV+TTbar_14TeV_TuneCUETP8M1_2023D17_GenSimHLBeamSpotFull14+DigiFullTrigger_2023D17+RecoFullGlobal_2023D17+HARVESTFullGlobal_2023D17/
11624.0_TTbar_13+TTbar_13TeV_TuneCUETP8M1_2021_GenSimFull+DigiFull_2021+RecoFull_2021+HARVESTFull_2021+ALCAFull_2021/
10824.0_TTbar_13+TTbar_13TeV_TuneCUETP8M1_2018_GenSimFull+DigiFull_2018+RecoFull_2018+HARVESTFull_2018+ALCAFull_2018+NanoFull_2018/

and all ran perfectly (the 2 last tests (11624 & 10824) are made to verify if the introduction did not alter the results.). 

#### if this PR is a backport please specify the original PR:

Before submitting your pull requests, make sure you followed this checklist:
- verify that the PR is really intended for the chosen branch -> OK
- verify that changes follow [CMS Naming, Coding, And Style Rules](http://cms-sw.github.io/cms_coding_rules.html) -> OK
- verify that the PR passes the basic test procedure suggested in the [CMSSW PR instructions](https://cms-sw.github.io/PRWorkflow.html) -> OK

@beaudett

PR #27120, opened 2019-06-05
Enable tau validation for phase 2

steggema:
#### PR description:

This PR enables tau validation for phase 2 and re-adds the RecoTauV directory in the DQM output. It also includes a more optimal setting of the binning for the summary plots in RecoTauV subdirectories. It required changing a few names of duplicated modules for the auto-validation to work since either the wrong one ended up being added to the process or the wrong one was changed by an era modifier.

It may interfere with https://github.com/cms-sw/cmssw/pull/27084/files but merging by hand should be trivial.

#### PR validation:

Checked that the RecoTauV folder is added with appropriate content for the relevant matrix workflows 27434.0, 29434.0, 20434.0, 20034.0

-

Adding @rmanzoni @roger-wolf @mbluj @aspiezia


Validation/RecoTau/src/TauTagValidation.cc
```diff
@@ -129,7 +129,7 @@ void TauTagValidation::bookHistograms(DQMStore::IBooker& ibooker,
   ibooker.setCurrentFolder("RecoTauV/" + TauProducer_ + extensionName_ + "_Summary");
   hinfo summaryHinfo = (histoSettings_.exists("summary"))
                            ? hinfo(histoSettings_.getParameter<edm::ParameterSet>("summary"))
-                           : hinfo(21, -0.5, 20.5);
+                           : hinfo(discriminators_.size(), -0.5, discriminators_.size() - 0.5);
```
jfernan2: @steggema is this change creating a non null number of bins histogram in all cases?

Validation/RecoTau/python/dataTypes/ValidateTausOnZTT_cff.py
```diff
@@ -5,9 +5,11 @@
 from PhysicsTools.HepMCCandAlgos.genParticles_cfi import *
 import PhysicsTools.PatAlgos.tools.helpers as helpers
 
+tauGenJetsForVal = tauGenJets.clone()
```
jfernan2: @steggema I don't understand why this clone is needed since no modifications are done afterwards

steggema: The point is that a MiniAOD era modifier changes process.tauGenJets, so it needs to have a different name and can't be the same object (i.e. if you write tauGenJetsForVal = tauGenJets it will just be a pointer). Maybe there's a better solution (... import as ... didn't work either).

Validation/RecoTau/src/TauTagValidation.cc
steggema: For all regular cases yes, but it may be a good idea to add a check here.

Validation/RecoTau/python/dataTypes/ValidateTausOnZTT_cff.py
jfernan2: I see thanks

Validation/RecoTau/src/TauTagValidation.cc
jfernan2: Please do so, we are having issues while merging such histograms: https://github.com/cms-sw/cmssw/issues/26727

steggema: This was addressed by https://github.com/cms-sw/cmssw/pull/27120/commits/af33e54928329fff9a803c9389938f148dfae4fe.
Also merged with the latest master to get in sync with the meanwhile merged E/Gamma PR mentioned in the PR description).

jfernan2:
please test

jfernan2:
@steggema Before approving, can you please do a last check that the 4069 histograms added to Phase2 wf are what you expect? E.g. in http://tinyurl.com/y4yc6ffk

steggema:
> @steggema Before approving, can you please do a last check that the 4069 histograms added to Phase2 wf are what you expect? E.g. in http://tinyurl.com/y4yc6ffk

@jfernan2 Yes, they're as expected for MC workflows.

ssekmen:
+1

jfernan2:
+1

civanch:
+1

fabiocos:
+1

the memory addition to the total is relatively small
```
< Total bytes: 675105.24 KiB
---
> Total bytes: 673661.26 KiB
```
and dispersed in a large number of small subfolders (not visible with the standard threshold of 512 KiB).

PR #27123, opened 2019-06-05
PF tracks to HF extrapolation for Phase2 (and suppress unnecessary warnings)

hatakeyamak:
#### PR description:

Suppress unnecessary warning for tracks with |eta|>3 not extrapolated to ECAL and HCAL which happens in the phase2 scenarios. Also, update PFTrackTransformer to allow extrapolation of tracks to HF.

#### PR validation:

Compiles. Matrix tests run. Also made sure no change to reco as we are not using extrapolated tracks to HF yet. This is a step 1.
http://hep.baylor.edu/hatake/PFval/val_10_6_0_FwdTrkExtrapolation/

#### if this PR is a backport please specify the original PR:

This is not backport.

@bendavid

PR #27131, opened 2019-06-06
PF tracks to HF extrapolation for Phase2 (and suppress unnecessary warnings)

hatakeyamak:
#### PR description:

Suppress unnecessary warning for tracks with |eta|>3 not extrapolated to ECAL and HCAL which happens in the phase2 scenarios. Also, update PFTrackTransformer to allow extrapolation of tracks to HF. [Superseded #27123]

#### PR validation:

Compiles. Matrix tests run. Also made sure no change to reco as we are not using extrapolated tracks to HF yet. This is a step 1.
http://hep.baylor.edu/hatake/PFval/val_10_6_0_FwdTrkExtrapolation/

This is not backport.

@bendavid


slava77:
@cmsbuild please test

RecoParticleFlow/PFTracking/src/PFTrackTransformer.cc
```diff
@@ -147,8 +150,13 @@ bool PFTrackTransformer::addPoints(reco::PFRecTrack& pftrack,
                                        meanShower,
                                        math::XYZTLorentzVector(theOutParticle.particle().momentum())));
   } else {
-    if (PT > 5. && msgwarning)
-      LogWarning("PFTrackTransformer") << "KF TRACK " << pftrack << " PROPAGATION TO THE ECAL HAS FAILED";
+    if (PT > 5. && theOutParticle.particle().cos2ThetaV() < 0.99014 && msgwarning)
```
slava77: please define a named constant to replace the magic number ```0.99014```

slava77: @hatakeyamak 
just in case this review was not noticed initially.

hatakeyamak:
Thanks. Just return from vacation. Will work on it.

RecoParticleFlow/PFTracking/interface/PFTrackTransformer.h
```diff
@@ -58,6 +58,7 @@ class PFTrackTransformer {
   ///B field
   math::XYZVector B_;
   PFGeometry pfGeometry_;
+  double cos2ThetaV_Endcap_HiEnd_;
```
perrotta: Defining here as ```static constexpr double``` could be a better choice

hatakeyamak: Tried it now.

perrotta:
@cmsbuild please test

RecoParticleFlow/PFTracking/interface/PFTrackTransformer.h
```diff
@@ -58,6 +58,8 @@ class PFTrackTransformer {
   ///B field
   math::XYZVector B_;
   PFGeometry pfGeometry_;
+  static constexpr double cos2ThetaV_Endcap_HiEnd_ = 0.99014;
```
slava77: it would be nice to eventually drop this here and define the constants in BaseParticlePropagator, which is the origin of these magic numbers.

hatakeyamak: @slava77 
Do you mean I define a constant in BaseParticlePropagator, and somehow use that in PFTrackTransformer as well?

slava77:
+1

for 
- jenkins tests pass and comparisons with the baseline show differences only in 2023 workflows related to the reduction in the LogWarning messages

RecoParticleFlow/PFTracking/interface/PFTrackTransformer.h
slava77: right. the value 0.99014 here is apparently supposed to match exactly the same value used (5 times) in BaseParticlePropagator.cc

hatakeyamak: I will do this in the next PR. OK?

slava77: OK.

fabiocos:
+1

PR #27149, opened 2019-06-07
HCALDQM: Updates for Run 3 HB (take 2)

DryRun:
This PR is a retry of https://github.com/cms-sw/cmssw/pull/26941. 

**PR description:**
This PR implements Run 3 HB in HCAL DQM. The old HBHEDigiCollection is removed, and the new HB digis are taken from QIE11DigiCollection. Since HB now overlaps with HO in (ieta, iphi, depth) space, HO digis are plotted in depth 7 for now. A future PR will improve on that hack, moving HO (and perhaps all subsystems) to their own ieta-iphi plots, rather than attempting to plot all subsystems together.

**PR validation:**
Successfully ran over recent local and global runs with partial HB installation. Note that running in online DQM will require appropriate conditions, which are being developed in parallel.
runTheMatrix was successful, particularly the one that failed in the last PR (136.731).

@abdoulline 

Backport to 10_6_X: https://github.com/cms-sw/cmssw/pull/27152

### DIFF
--- a/DQM/HcalCommon/src/DetectorQuantity.cc
+++ b/DQM/HcalCommon/src/DetectorQuantity.cc
@@ -26,7 +26,8 @@ namespace hcaldqm {
 
     uint32_t getBin_ieta(HcalDetId const &did) { return (uint32_t)(getValue_ieta(did) + 1); }
 
-    uint32_t getBin_depth(HcalDetId const &did) { return (uint32_t)(did.depth()); }
+    uint32_t getBin_depth(HcalDetId const &did) { return (uint32_t)(
+      did.subdet() == HcalOuter ? 7 : did.depth()); }
 
     uint32_t getBin_Subdet(HcalDetId const &did) { return (uint32_t)(did.subdet()); }
 
--- a/DQM/HcalCommon/src/DetectorQuantity.cc
+++ b/DQM/HcalCommon/src/DetectorQuantity.cc
@@ -26,8 +26,9 @@ namespace hcaldqm {
 
     uint32_t getBin_ieta(HcalDetId const &did) { return (uint32_t)(getValue_ieta(did) + 1); }
 
-    uint32_t getBin_depth(HcalDetId const &did) { return (uint32_t)(
-      did.subdet() == HcalOuter ? 7 : did.depth()); }
+    uint32_t getBin_depth(HcalDetId const &did) { //return (uint32_t)(did.depth());}
+      return (uint32_t)(did.subdet() == HcalOuter ? 7 : did.depth());
+    }
 
     uint32_t getBin_Subdet(HcalDetId const &did) { return (uint32_t)(did.subdet()); }
 
--- a/DQM/HcalTasks/interface/DigiTask.h
+++ b/DQM/HcalTasks/interface/DigiTask.h
@@ -39,16 +39,14 @@ protected:
   void _process(edm::Event const &, edm::EventSetup const &) override;
   void _resetMonitors(hcaldqm::UpdateFreq) override;
 
-  edm::InputTag _tagHBHE;
-  edm::InputTag _tagHE;
+  edm::InputTag _tagQIE11;
   edm::InputTag _tagHO;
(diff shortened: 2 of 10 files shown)


jfernan2:
please test

jfernan2:
@DryRun, same comment as in the backport:
The PR is apparently creating the ADC plots from QIE11DigiCollection empty (see e.g. http://tinyurl.com/y3vehmf3 which compare base QIE8 http://tinyurl.com/y2adbjqo with the base+PR http://tinyurl.com/y42aohxg and other MEs in the links). Can you confirm this is expected?

Can you also check if the output produced is OK across workflows? In particular I don't see any ME related to depth 7 as you mentioned in the PR description:
https://cmssdt.cern.ch/SDT/jenkins-artifacts/baseLineComparisons/CMSSW_10_6_X_2019-06-07-2300+6588f4/32161/dqm-histo-comparison-summary.html
Thanks!

DryRun:
Hi @jfernan2 - 

> The PR is apparently creating the ADC plots from QIE11DigiCollection empty (see e.g. http://tinyurl.com/y3vehmf3 which compare base QIE8 http://tinyurl.com/y2adbjqo with the base+PR http://tinyurl.com/y42aohxg and other MEs in the links). Can you confirm this is expected?
Yes, I think it's expected, as I removed the block that processed QIE8 digis in this PR. (Though I don't know specifically how these RelVals are configured)

> Can you also check if the output produced is OK across workflows? In particular I don't see any ME related to depth 7 as you mentioned in the PR description:
Yes, this is also OK. The depth maps are created from the emap, so for example, if an old RelVals uses an emap with 2 depths, then only 2 depth histograms are created. But again, I'm not very familiar with how these RelVals are configured...

jfernan2:
Hi @DryRun you can see the configuration of every workflow in
https://cmssdt.cern.ch/SDT/jenkins-artifacts/pull-request-integration/PR-f391d9/842/runTheMatrix-results/

jfernan2:
+1

fabiocos:
please test

fabiocos:
@DryRun @jfernan2 I understand that this development is targeting Run3, but it affects test workflows for past runs as well, as it can be seen in the DQM checks https://cmssdt.cern.ch/SDT/jenkins-artifacts/baseLineComparisons/CMSSW_11_0_X_2019-06-16-0000+f391d9/32265/validateJR.html touching both data and MC. If we would merge this code, many checks on Run1/Run2 Hcal DIGIs that are still in use would become just useless, am I correct? Is this that we really want? 
For sure I do not see how this might be seamlessly backported on the release to be used on UL

mariadalfonso:
@fabiocos 
This PR target the MWGR of july.
we need the backport #27152 of this PR in the 10_6_1

fabiocos:
@mariadalfonso you need this PR in some 10_6_X release, not necessarily 10_6_1 (whose main purpose is UL reprocessing for 2017). But my questions should be addressed before, even for master. Thanks

DryRun:
Hi @fabiocos (also @mariadalfonso , @abdoulline) -- we had some discussion of this above with @jfernan2 , the conclusion of which is that these DQM modules are only used for online DQM, while offline validation is handled by other modules (perhaps @abdoulline can comment more on that). While processing of the older Run1/Run2-style digis could be preserved with eras, we decided it's not worth the extra complexity to make plots that are not used by anyone.

jfernan2:
@DryRun but these DQM modules are also produced in Offline DQM (not only Online DQM) since the workflows are modfied with the PR. Are you sure this is not used for DC?

abdoulline:
They are used in Offline DQM (the goal of this PR is also to provide 
Offline DQM for >= MWGR#3), but not for previous eras.
In particular, for UL we've used era-customized OfflineDQM/Hcal.

fabiocos:
As agreed at the ORP meeting with HCAL DPG and PPD, it is ok to modify the behaviour on past runs of this DQM part, even if it is not simply related to online

fabiocos:
+1

PR #27128, opened 2019-06-06
Extend and fix RecHitTools

clelange:
- add bhOffset_,
- fix getLayerWithOffset for new geometry,
- add firstLayerBH(), lastLayerBH()

#### PR description:

This PR introduces some important functionality and fixes in/to the RecHitTools required to eventually remove magic numbers in other places of the HGCal reconstruction code.
Related to https://github.com/cms-sw/cmssw/issues/26225

#### PR validation:

Ran some test code with D41 geometry and got the following output:

```text
bhOffset_: 36
EE layers: 28
FH layers: 22
BH layers: 14
first EE layer: 1
first FH layer: 1
first BH layer: 9
EE layer at z = 10000: 28
FH layer at z = 10000: 24
BH layer at z = 10000: 24
lastLayerEE: 28
lastLayerFH: 50
lastLayerBH: 50
firstLayerBH: 37
```

There are some discrepancies that need to be understood, which, however, seem to be in the geometry, underlining the importance of this PR for geometry validation.
For example: D41 should have 50 HGCal layers only, 22 FH layers, 14 BH layers.
After talking to @bsunanda , the error might also be in my setup and there are no issues in D41 geoemtry. We will investigate, but that's independent of the PR.

PR #27026, opened 2019-05-31
TICL Reorganisation

rovere:
##### TICL Reorganization

Several changes entered this PR.
    
  * The data structures used by TICL have been moved into
      DataFormats/TICL, that has to be added as a CMSSW package.
      Tracksters and Tile are currently present. The encapsulation of
      std::array as a transient data member of a Tile is due to the fact
      that ROOT natively supports those and refuses to create ad-hoc
      dictionaries. Moreover, at present, an std::array cannot sit at the
      top of a ROOT branch requiring, once again, its encapsulation.
  * The "master Tile" has been added as a single producer. A unique Tile
      containing all the 2D reconstructed clusters is created once and put
      into the Event. All iterations will fetch it from the event and mask
      the 2D clusters on the fly. This required also some changes to the
      clusters filters logic. Now the filters produce also a vector<float>
      that represents the result of merging their input mask with their
      own selection. TICL iterations will make use of the filter's
      vector<float> to dynamically mask clusters from the "master Tile".
  * The fractions have been added to the Tracksters. The sharing of the
      energy of 2D clusters is flat and based on their multiplicity, i.e,
      a fraction of energy/multiplicity is assigned to all Tracksters that
      use the 2D clusters. Better splitting algorithms can be studies or
      Machine Learning techniques could be put at work here.
  * The hits and fraction of the Trackster are now correctly populated.
      This required an additional loop over the components of each 2D
      cluster. In the future the CaloCluster, PFCluster and hits and
      fractions data structures could be reviewed or, at least, TICL could
      use some other, more efficient, internal data representation and
      collapse into the legacy ones only at the very last stage.

#### PR validation:

`runTheMatrix.py -l limited -i all -j 4`

and other samples.

**Note to the reviewers**: all changes in `RecoLocalCalo/HGCalRecAlgos` are purely due to the code-format. No other changes have been committed there. I did so so that PR #26991 could be closed (if and when this PR gets merged)

PR #27160, opened 2019-06-11
TICL Reorganization, reloaded.

rovere:
##### TICL Reorganization

Several changes entered this PR.
    
  * The data structures used by TICL have been moved into
      DataFormats/TICL, that has to be added as a CMSSW package.
      Tracksters and Tile are currently present. The encapsulation of
      std::array as a transient data member of a Tile is due to the fact
      that ROOT natively supports those and refuses to create ad-hoc
      dictionaries. Moreover, at present, an std::array cannot sit at the
      top of a ROOT branch requiring, once again, its encapsulation.
  * The "master Tile" has been added as a single producer. A unique Tile
      containing all the 2D reconstructed clusters is created once and put
      into the Event. All iterations will fetch it from the event and mask
      the 2D clusters on the fly. This required also some changes to the
      clusters filters logic. Now the filters produce also a vector<float>
      that represents the result of merging their input mask with their
      own selection. TICL iterations will make use of the filter's
      vector<float> to dynamically mask clusters from the "master Tile".
  * The fractions have been added to the Tracksters. The sharing of the
      energy of 2D clusters is flat and based on their multiplicity, i.e,
      a fraction of energy/multiplicity is assigned to all Tracksters that
      use the 2D clusters. Better splitting algorithms can be studies or
      Machine Learning techniques could be put at work here.
  * The hits and fraction of the Trackster are now correctly populated.
      This required an additional loop over the components of each 2D
      cluster. In the future the CaloCluster, PFCluster and hits and
      fractions data structures could be reviewed or, at least, TICL could
      use some other, more efficient, internal data representation and
      collapse into the legacy ones only at the very last stage.

#### PR validation:

`runTheMatrix.py -l limited -i all -j 4`

and other samples.

**Note to the reviewers**: all changes in `RecoLocalCalo/HGCalRecAlgos` are purely due to the code-format. No other changes have been committed there. I did so so that PR #26991 could be closed (if and when this PR gets merged)


rovere:
@slava77 for some weird reason PR #27026 got a little confused after my last rebase. I decided to close that and open a fresh new PR, hoping that this time the review process will go smoother.
I'll read the comments you made on #27026 and implement some of the suggestions.

As for moving stuff out of DataFormats/TICL into DataFormats/HGCRecHits, I disagree. That package has a very specific name and I'd not mix stuff together. I see the value of not adding too many packages, yet the proposed solution is not optimal, from my point of view. Either we change the name of the current DataFormats/HGCRecHit into something more generic, like HGCRecoCommonn, or any other name you could suggest, or I'd prefer to keep things logically separated into different folders.

I'll go through the other comments in the coming days.

rovere:
@slava77 Also for the move of TICL algorithms into RecoLocalCalo, I believe this is misleading. TICL, or more in general HGCAL reconstruction, will have internally links and connections with other detectors, like tracker/muons. I'd prefer to have it in a separate package that explicitly declares its purpose.
RecoLocalCalo/HGCalRecAlgos, instead, contains correctly algorithms to perform real local (i.e. 2d, eventually 3D) clustering.

perrotta:
please test

slava77:
> @slava77 Also for the move of TICL algorithms into RecoLocalCalo, I believe this is misleading. TICL, or more in general HGCAL reconstruction, will have internally links and connections with other detectors, like tracker/muons. I'd prefer to have it in a separate package that explicitly declares its purpose.
> RecoLocalCalo/HGCalRecAlgos, instead, contains correctly algorithms to perform real local (i.e. 2d, eventually 3D) clustering.

I was apparently incorrect that the current status of this package represents the target as a skeleton.
From the comment above it sounds like the data formats are incomplete; that there will be objects linking HGCAL, tracker, and muon. So, it's not really a reco of HGCAL clusters.
If that's the case, RecoParticleFlow subsystem is perhaps more appropriate subsystem than RecoHGCAL.

About the DataFormats. Please remind me the motivation to move the transient-only products to the DataFormats subsystem.
It seems like even the Trackster is actually transient (even though it's the only one not marked to be as such).
https://twiki.cern.ch/twiki/bin/view/CMSPublic/SWGuideCreatingNewProducts#Restrictions_on_a_package_defini
There is much more freedom for transient products definitions and developments.

rovere:
@slava77 In the long term the plan is to have links between HGCal and the nearby detectors, following a ParticleFlow like reconstruction approach. At present and on a timescale of ~year, this will be difficult to be fully implemented and we will use information from other detectors to trigger specific iterations (defining specific seeds) inside TICL. In view of this, I'd not start adding stuff directly into ParticleFlow related packages, because this could give the wrong impression that the outcome of TICL is already a particle-flow product, which is not (yet). It's like tracking using calorimetric and muon-detector information to trigger dedicated iterations, yet it's not considered part of particle-flow. Could you clarify what exactly you are trying to achieve by scattering pieces of a novel reconstruction for a new detector into already existing packages?

*DataFormats*
Tracksters are temporarily transient and may be persisted in future, especially if their graph-based nature (once we agree on an optimal representation of that information) will be exploited by ML algorithms.
The TICL Layers will be totally transient since they are useful to navigate 2D clusters in an efficient way but will add no additional information. I believe gathering all TICL related objects that will be put into the Event into a dedicated DataFormats package is a natural choice. Could you explain what is the added value in not following this path?

slava77:
On 6/12/19 12:24 AM, Marco Rovere wrote:
> Could you clarify what exactly you are trying to achieve by scattering
> pieces of a novel reconstruction for a new detector into already
> existing packages?

Just to clarify, I want this to be moved to an existing subsystem.
I do not see a need for a new subsystem.
Either we assume that current status (At present and on a timescale of
~year), and it is RecoLocalCalo,
or look for a longer term target of linking more detectors,
and then it is RecoParticleFlow subsystem.
It certainly deserves to stay in a  separate  TICL package.

slava77:
On 6/12/19 12:24 AM, Marco Rovere wrote:
> Could you explain what is the added value in not following this path?

I tried to clarify this already:
There are less restrictions for code outside of the DataFormats subsystem.

slava77:
@rovere 
based on your inputs, here is my updated proposal:
-  RecoHGCal subsystem stays to collect not just TICL but also future high-level HGCal-specific algorithms. HGC is a complex detector and it reco does not quite fit in the RecoLocalCalo, but on the higher level synthesis, its reco algorithms are not necessarily RecoParticleFlow either
- The DataFormats/TICL contents need to be moved to DataFormats/RecoHGCal so that the attribution to HGC is clear and we are not polluting the general DataFormats package subsystem with small packages.

What do you think?

rovere:
I'll wait that also #27128 is merged before rebasing. Please put that in higher priority, thanks.

slava77:
@cmsbuild please test

slava77:
@cmsbuild please test

slava77:
@rovere 
please clarify on the status of updates in this PR to address https://github.com/cms-sw/cmssw/pull/27026#pullrequestreview-246631511.
Compared to the reviewed code ( #27026	b20310a), I see that the only change is DataFormats/TICL => DataFormats/RecoHGCal.

rovere:
@slava77 now that I see no more rebase ahead (but #27194, which should be trivial), I can go over them and either implement or comment on this PR.

rovere:
@slava77 I think I covered almost all comments. Few were left out:

* The ES validity of the rechittool is, in fact, constant for the full run (it access the geometry).
* the assert on the size of filters and layercluster is not necessary, since the former is always constructed rom the latter.

slava77:
@cmsbuild please test

slava77:
I was going to update the categories_map.py file for the bot and now I realize that  DataFormats/RecoHGCal is breaking the pattern of other package names in the DF subsystem. They all have "Reco" as a suffix not as a prefix. Please move to DataFormats/HGCalReco. Sorry for not realizing this initially when the package name was proposed.

rovere:
@slava77 no problem.

> I was going to update the categories_map.py file for the bot and now I realize that DataFormats/RecoHGCal is breaking the pattern of other package names in the DF subsystem. They all have "Reco" as a suffix not as a prefix. Please move to DataFormats/HGCalReco. Sorry for not realizing this initially when the package name was proposed.

slava77:
@cmsbuild please test

@rovere 
thank you for the quick update

DataFormats/HGCalReco/src/classes_def.xml
```diff
@@ -0,0 +1,19 @@
+
+<lcgdict>
+  <class name="ticl::Trackster" ClassVersion="5">
+   <version ClassVersion="5" checksum="3878166595"/>
```
slava77: FWIW, this is a new class and the version name should start from 3.

DataFormats/HGCalReco/src/classes_def.xml
```diff
@@ -0,0 +1,19 @@
...
+  </class>
+  <class name="std::vector<ticl::Trackster>" />
+  <class name="edm::Wrapper<ticl::Trackster>" />
+  <class name="edm::Wrapper<std::vector<ticl::Trackster> >" />
+
+  <class name="TICLLayerTile" ClassVersion="4" persistent="false">
+    <field name="tile_" transient="true"/>
+    <version ClassVersion="4" checksum="63861365"/>
```
slava77: classVersion=3, if we need to keep versioning.
However, given that this is transient, I propose to drop the version support.

DataFormats/HGCalReco/src/classes_def.xml
```diff
@@ -0,0 +1,19 @@
...
+  </class>
+  <class name="TICLLayerTiles" ClassVersion="4" persistent="false">
```
slava77: same here: either change to version 3, or, better yet, drop the version support.

slava77 (commented):
we are almost there

slava77:
@cmsbuild please test

DataFormats/HGCalReco/src/classes_def.xml
```diff
@@ -1,19 +1,16 @@
 
 <lcgdict>
-  <class name="ticl::Trackster" ClassVersion="5">
-   <version ClassVersion="5" checksum="3878166595"/>
+  <class name="ticl::Trackster">
```
slava77: Just to be sure, if the Trackster class is planned to be persisted in production, then the class version is still useful.
From earlier comments I understood that it may be used for storage mostly for tests or otherwise cases where the format evolution would not matter.
@rovere 
please clarify/confirm.
Thank you.

rovere: @slava77 that's difficult to say at the moment. The idea is to use them as intermediate products to deliver something more complex and with more links with the outside world.
I'll add it back and, in case we will end up never persisting it, we can remove it at a later time. Would that be ok?

slava77: >  I'll add it back and, in case we will end up never persisting it, we can remove it at a later time. Would that be ok?

Yes, please.

slava77:
@cmsbuild please test

slava77:
+1

for https://github.com/cms-sw/cmssw/pull/27160	fba76c7
- jenkins tests pass and comparisons with the baseline show no differences (as seen in a slightly earlier test of somewhat the same code; the TICL code is not running in production anyways).

@kpedro88 please check this PR at your earliest so that we can then proceed with #27194

RecoHGCal/TICL/plugins/PatternRecognitionbyCA.cc
```diff
@@ -6,54 +6,49 @@
...
-                                   min_cos_theta_, min_cos_pointing_, missing_layers_,
-                                   rhtools_.lastLayerFH());
-  theGraph_.findNtuplets(foundNtuplets, min_clusters_per_ntuplet_);
+  std::vector<uint8_t> layer_cluster_usage(layerClusters.size(), 0);
+  theGraph_->makeAndConnectDoublets(tiles,
+                                    ticl::constants::nEtaBins,
+                                    ticl::constants::nPhiBins,
+                                    layerClusters,
+                                    mask,
+                                    2,
```
kpedro88: could these magic numbers be named constants?

RecoHGCal/TICL/plugins/PatternRecognitionbyCA.cc
```diff
@@ -63,22 +58,32 @@ void PatternRecognitionbyCA::makeTracksters(const edm::Event &ev, const edm::Eve
...
+      layer_cluster_usage[i]++;
+      LogDebug("HGCPatterRecoByCA") << "LayerID: " << i << " count: " << (int)layer_cluster_usage[i] << std::endl;
+    }
     // Put back indices, in the form of a Trackster, into the results vector
     Trackster tmp;
     tmp.vertices.reserve(effective_cluster_idx.size());
-    std::copy(std::begin(effective_cluster_idx), std::end(effective_cluster_idx),
-              std::back_inserter(tmp.vertices));
+    tmp.vertex_multiplicity.resize(effective_cluster_idx.size(), 0);
+    std::copy(std::begin(effective_cluster_idx), std::end(effective_cluster_idx), std::back_inserter(tmp.vertices));
```
kpedro88: I think this would be simpler (and maybe better performing) as:
```
tmp.vertices.insert(tmp.vertices.end(),effective_cluster_idx.begin(),effective_cluster_idx.end());
```

RecoLocalCalo/HGCalRecAlgos/interface/ClusterTools.h
```diff
@@ -13,50 +13,53 @@
...
 
-    std::vector<size_t> sort_by_z(const reco::HGCalMultiCluster&v) const {
+  private:
+    std::vector<size_t> sort_by_z(const reco::HGCalMultiCluster &v) const {
       std::vector<size_t> idx(v.size());
-      std::iota (std::begin(idx), std::end(idx), 0);
-      sort(idx.begin(), idx.end(),
-	   [&v](size_t i1, size_t i2) {return v.clusters()[i1]->z() < v.clusters()[i2]->z();});
+      std::iota(std::begin(idx), std::end(idx), 0);
+      sort(
```
kpedro88: `std::sort`

kpedro88 (commented):
These few minor comments can be addressed in a followup PR.

kpedro88:
+upgrade

slava77:
@fabiocos 
now that this is fully signed,
perhaps you can check this some time today or soon after.
Thank you.

fabiocos:
@makortel there is an overlap with you #27194, but this is not creating conflicts

fabiocos:
@slava77 @kpedro88 I understand that you are satisfied enough with the present status of the code, to integrate it closing a long-lasting saga, and further improvements may come later. I will integrate this PR in next IB.

slava77:
On 6/26/19 4:56 PM, Fabio Cossutti wrote:
> I will integrate this PR in next IB.

Thank you.

fabiocos:
+1

PR #27174, opened 2019-06-11
Reco vertex adjustments

perrotta:
#### PR description:

Starting from the inspection of some static analyzer warning message, the following adjustments are proposed for two files in RecoVertex:

-  RecoVertex/NuclearInteractionProducer/src/NuclearLikelihood.cc
   - Do not let undefined idBest even if no secondary track is found in 

- RecoVertex/V0Producer/src/V0Fitter.cc
  - Avoid non necessary calculations of sqrt

#### PR validation:

It compiles
No changes expected in output

### DIFF
--- a/RecoVertex/NuclearInteractionProducer/src/NuclearLikelihood.cc
+++ b/RecoVertex/NuclearInteractionProducer/src/NuclearLikelihood.cc
@@ -4,15 +4,16 @@ void NuclearLikelihood::calculate(const reco::Vertex& vtx) {
   likelihood_ = 0.0;
   if (vtx.isValid()) {
     if (vtx.tracksSize() > 1) {
-      likelihood_ = 0.3;
-      int idBest;
+      int idBest = 0;
       int secMaxHits = secondaryTrackMaxHits(vtx, idBest);
-      if (secMaxHits > 3)
-        likelihood_ = 0.5;
-      if (secMaxHits > 4)
-        likelihood_ = 0.7;
       if ((*(vtx.tracks_begin() + idBest))->normalizedChi2() < 3.0)
         likelihood_ = 1.0;
+      else if (secMaxHits > 4)
+        likelihood_ = 0.7;
+      else if (secMaxHits > 3)
+        likelihood_ = 0.5;
+      else
+        likelihood_ = 0.3;
     }
   }
 }
--- a/RecoVertex/V0Producer/src/V0Fitter.cc
+++ b/RecoVertex/V0Producer/src/V0Fitter.cc
@@ -175,7 +175,7 @@ void V0Fitter::fitAll(const edm::Event& iEvent,
 
       // the POCA should at least be in the sensitive volume
       GlobalPoint cxPt = cApp.crossingPoint();
-      if ( (cxPt.x() * cxPt.x() + cxPt.y() * cxPt.y()) > 120.*120. || std::abs(cxPt.z()) > 300.)
+      if ((cxPt.x() * cxPt.x() + cxPt.y() * cxPt.y()) > 120. * 120. || std::abs(cxPt.z()) > 300.)
         continue;
 
       // the tracks should at least point in the same quadrant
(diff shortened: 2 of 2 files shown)


RecoVertex/NuclearInteractionProducer/src/NuclearLikelihood.cc
```diff
@@ -4,15 +4,16 @@ void NuclearLikelihood::calculate(const reco::Vertex& vtx) {
...
     if (vtx.tracksSize() > 1) {
-      likelihood_ = 0.3;
-      int idBest;
+      int idBest = 0;
       int secMaxHits = secondaryTrackMaxHits(vtx, idBest);
-      if (secMaxHits > 3)
-        likelihood_ = 0.5;
-      if (secMaxHits > 4)
-        likelihood_ = 0.7;
       if ((*(vtx.tracks_begin() + idBest))->normalizedChi2() < 3.0)
```
perrotta: While editing this code, I was surprised of this line (that I did not modify): if there are no secondary tracks in the vertex (i.e. secMaxHits=0) and the chi2 of the main track is below 3 then likelihood_ is equal 1.

Could anybody confirm that this the wanted behaviour?
@mtosi

perrotta:
please test

perrotta:
While editing this code, I was surprised of this line (that I did not modify): 
https://github.com/cms-sw/cmssw/pull/27174/files#diff-bf19904422a1a6c6832a80ca71eb839bR9

If there are no secondary tracks in the vertex (i.e. secMaxHits=0) and the chi2 of the main track is lower than 3, then ```likelihood_``` is put equal 1.

Could anybody confirm that this is indeed the wanted behaviour? (But I suspect it can be a bug, still to be fixed...)

@mtosi @JanFSchulte @ferencek @arizzi

ferencek:
I really have no familiarity with this code and as far as I see from https://github.com/cms-cvs-history/RecoVertex-NuclearInteractionProducer/commits/master/src/NuclearLikelihood.cc and https://github.com/cms-sw/cmssw/commits/master/RecoVertex/NuclearInteractionProducer/src/NuclearLikelihood.cc, until these recent format changes, this code hasn't been touched for 11 years. I'm not sure among the people still around who might know more about this code.

RecoVertex/NuclearInteractionProducer/src/NuclearLikelihood.cc
slava77: I tried to grep a bit, and I did not find any active use cases of this code.

perrotta: Indeed...
- "NuclearLikelihood" is used inside ```RecoVertex/NuclearInteractionProducer/plugins/NuclearInteractionEDProducer.cc```
- ... whose plugin is used in ```RecoVertex/NuclearInteractionProducer/python/NuclearInteraction_cfi.py```
- ... which is imported by ```RecoParticleFlow/PFTracking/python/nuclear_cff.py```
- ... entering in a sequence defined in ```RecoParticleFlow/​PFTracking/​python/​nuclearRemainingHits_cff.py```
- ... and that file is never accessed in CMSSW

All those modules/configs were never updated since 2013 (at least), excluding a few technical adaptations to the developing framework from time to time. I suspect that they can all be erased, in particular now that some PF code cleanup is ongoing: @hatakeyamak @bendavid

perrotta:
+1
- Just remove the origin of a few static analyzer warnings, without modifying the behaviour itself of the code
- Even if during the review I noticed something that looked like a bug in the original code of ```RecoVertex/NuclearInteractionProducer/src/NuclearLikelihood.cc```, given https://github.com/cms-sw/cmssw/pull/27174#discussion_r293587169 and https://github.com/cms-sw/cmssw/pull/27174#discussion_r293695610, and taking also into account that apparently nobody knows what this code is supposed to do by now, I'd suggest to merge this PR as it is, just to remove the static analyzer warnings, and avoid loosing time in inspecting the code further.
  - In a forthcoming iteration one could simply decide to remove all the code which is obsolete.

fabiocos:
+1

PR #27181, opened 2019-06-12
update Pixel digitizer thresholds for Run3

mmusich:
#### PR description:

Phase1 Pixel digitization thresholds for Run 3 era are modified to match the expected behavior during LHC Run3.
This PR contains:
   * updated BPix L1 thresholds: decrease from 3000 to 1300 electrons
   * updated BPix L2 thresholds: decrease from 2600 to 1600 electrons
   * updated BPix L3/L4 thresholds: decrease from 2000 to 1600 electrons
   * updated FPix thresholds: decrease from 2000 to 1600 electrons

This PR de facto undoes the worsening of the resolution introduced in PR https://github.com/cms-sw/cmssw/pull/20775 where thresholds were increased to match the prelminiary resolution measurements in 2017 data.
We do expect an improvement for the BPix L1 and L2 thresholds from the better timing alignment of the two layers readouts with the installation of the new BPix L1 over LS2.
The expected impact on physics will be a slight improvement of pixel related tracking quantities.

#### PR validation:

Code compiles and the dumping of cmsDriver output configuration confirms the desired thresholds are assigned. 
There is not yet a physics output based validation (in progress), that is why this PR is proposed as draft. We will make it regular once the comparisons are available.

#### if this PR is a backport please specify the original PR:

This is not a backport

cc: @tsusa @tvami @leaca

PR #27183, opened 2019-06-12
Activate IrradiationBiasCorrection in Pixel CPE generic for Run3

mmusich:
#### PR description:

In this PR the `IrradiationBiasCorrection` parameter for the Pixel CPE Generic algorithm is activated for the Run 3 era. 
The motivation for activating the correction is due to the expected comparatively large loss of resolution along the local y [global-z in barrel] direction for an heavily irradiated detector (as expected for the end of run3 and extended end or run3 simulation scenarios), in which carrier trapping significantly reduces Lorentz drift induced charge sharing.
As shown in [this presentation](https://indico.cern.ch/event/808415/contributions/3447906/attachments/1853234/3043117/Run3_Template_Update.pdf), without this correction most clusters are reconstructed at ±100-200μm from true y, caused by charge loss from the backplane side.

As pointed out in the review (see https://github.com/cms-sw/cmssw/pull/27183#issuecomment-502780836) the Pixel CPE conditions need to be updated to take into account the parameter is switched on. 
The `phase1_2021_realistic` and `phase1_2021_cosmics` are updated accordingly.
The diff of the two Global Tags is the following:
```bash
$ conddb diff 106X_upgrade2021_realistic_Candidate_2019_06_17_18_50_50 106X_upgrade2021_realistic_v5
[2019-06-17 21:26:38,571] INFO: Connecting to pro [frontier://PromptProd/CMS_CONDITIONS]
Record                        Label        pro::106X_upgrade2021_realistic_Candidate_2019_06_17_18_50_50 Tag  pro::106X_upgrade2021_realistic_v5 Tag                          
----------------------------  -----------  -----------------------------------------------------------------  --------------------------------------------------------------  
SiPixel2DTemplateDBObjectRcd  denominator  SiPixel2DTemplateDBObject_phase1_BoR3_HV350_Tr1300_den             SiPixel2DTemplateDBObject_phase1_38T_mc_v4_den_unirradiated     
SiPixel2DTemplateDBObjectRcd  numerator    SiPixel2DTemplateDBObject_phase1_BoR3_HV350_Tr1300_num             SiPixel2DTemplateDBObject_phase1_38T_2018_ultralegacymc_v1_num  
SiPixelGenErrorDBObjectRcd    -            SiPixelGenErrorDBObject_phase1_BoR3_HV350_Tr1300                   SiPixelGenErrorDBObject_phase1_38T_2018_ultralegacymc_v1        
SiPixelLorentzAngleRcd        -            SiPixelLorentzAngle_phase1_BoR3_HV350_Tr1300                       SiPixelLorentzAngle_phase1_2018_ultralegacymc_v1                
SiPixelLorentzAngleSimRcd     -            SiPixelLorentzAngleSim_phase1_BoR3_HV350_Tr1300                    SiPixelLorentzAngleSim_phase1_mc_v4                             
SiPixelTemplateDBObjectRcd    -            SiPixelTemplateDBObject_phase1_BoR3_HV350_Tr1300                   SiPixelTemplateDBObject_phase1_38T_2018_ultralegacymc_v1      
```
```bash
$ conddb diff 106X_upgrade2021cosmics_realistic_deco_Candidate_2019_06_17_19_08_59 106X_upgrade2021cosmics_realistic_deco_v2[2019-06-17 21:27:39,951] INFO: Connecting to pro [frontier://PromptProd/CMS_CONDITIONS]
Record                        Label        pro::106X_upgrade2021cosmics_realistic_deco_Candidate_2019_06_17_19_08_59 Tag  pro::106X_upgrade2021cosmics_realistic_deco_v2 Tag              
----------------------------  -----------  -----------------------------------------------------------------------------  --------------------------------------------------------------  
SiPixel2DTemplateDBObjectRcd  denominator  SiPixel2DTemplateDBObject_phase1_BoR3_HV350_Tr1300_den                         SiPixel2DTemplateDBObject_phase1_38T_mc_v4_den_unirradiated     
SiPixel2DTemplateDBObjectRcd  numerator    SiPixel2DTemplateDBObject_phase1_BoR3_HV350_Tr1300_num                         SiPixel2DTemplateDBObject_phase1_38T_2018_ultralegacymc_v1_num  
SiPixelGenErrorDBObjectRcd    -            SiPixelGenErrorDBObject_phase1_BoR3_HV350_Tr1300                               SiPixelGenErrorDBObject_phase1_38T_2018_ultralegacymc_v1        
SiPixelLorentzAngleRcd        -            SiPixelLorentzAngle_phase1_BoR3_HV350_Tr1300                                   SiPixelLorentzAngle_phase1_2018_ultralegacymc_v1                
SiPixelLorentzAngleSimRcd     -            SiPixelLorentzAngleSim_phase1_BoR3_HV350_Tr1300                                SiPixelLorentzAngleSim_phase1_mc_v4                             
SiPixelTemplateDBObjectRcd    -            SiPixelTemplateDBObject_phase1_BoR3_HV350_Tr1300                               SiPixelTemplateDBObject_phase1_38T_2018_ultralegacymc_v1   
```

#### PR validation:
Code compiles and the dumping of cmsDriver output configuration confirms the desired parameter value is assigned.
The activation of the parameter has been tested (together with the changes proposed in #27181) with:
```
cmsDriver.py -s GEN,SIM,DIGI,L1,DIGI2RAW,RAW2DIGI,RECO --evt_type SingleMuPt10_pythia8_cfi --fileout file:GENSIMRECO_MuPt10.root --python_filename=Run3_2023_generic_noIBC.py -n 10 --no_exec --era Run3 --beamspot Run3RoundOptics25ns13TeVHighSigmaZ --nThreads=4 —customise_commands="process.mix.digitizers.pixel.ThresholdInElectrons_BPix_L1 = cms.double(1300.0);  process.mix.digitizers.pixel.ThresholdInElectrons_BPix_L2 = cms.double(1600.0);process.mix.digitizers.pixel.ThresholdInElectrons_BPix = cms.double(1600.0); process.mix.digitizers.pixel.ThresholdInElectrons_FPix = cms.double(1600.0);process.PixelCPEGenericESProducer.IrradiationBiasCorrection = True”   --conditions 106X_mcRun3_2023_realistic_Candidate_2019_06_07_21_52_54 
```
~~There is not yet a physics output based validation (in progress), We will update the description once comparisons are available.~~
Physics validation has been shown in this presentation at the RECO / AT meeting of  14 Jun 2019 [link](https://indico.cern.ch/event/827369/contributions/3467663/attachments/1862678/3061762/tanjaReco_14062019.pdf).

This PR is not a backport

cc: @tsusa @tvami @leaca @pmaksim1 @cmantill @OzAmram

### DIFF
--- a/Configuration/AlCa/python/autoCond.py
+++ b/Configuration/AlCa/python/autoCond.py
@@ -66,9 +66,9 @@ autoCond = {
     # GlobalTag for MC production with perfectly aligned and calibrated detector for Phase1 2021
     'phase1_2021_design'       : '106X_upgrade2021_design_v3', # GT containing design conditions for Phase1 2021
     # GlobalTag for MC production with realistic conditions for Phase1 2021
-    'phase1_2021_realistic'    : '106X_upgrade2021_realistic_v7', # GT containing realistic conditions for Phase1 2021
+    'phase1_2021_realistic'    : '106X_upgrade2021_realistic_Candidate_2019_06_17_18_50_50', # GT containing realistic conditions for Phase1 2021
     # GlobalTag for MC production (cosmics) with realistic conditions for Phase1 2021,  Strip tracker in DECO mode
-    'phase1_2021_cosmics'      : '106X_upgrade2021cosmics_realistic_deco_v2',
+    'phase1_2021_cosmics'      : '106X_upgrade2021cosmics_realistic_deco_Candidate_2019_06_17_19_08_59',
     # GlobalTag for MC production with realistic conditions for Phase2 2023
     'phase2_realistic'         : '106X_upgrade2023_realistic_v5'
 }
--- a/RecoLocalTracker/SiPixelRecHits/python/PixelCPEGeneric_cfi.py
+++ b/RecoLocalTracker/SiPixelRecHits/python/PixelCPEGeneric_cfi.py
@@ -62,7 +62,7 @@ PixelCPEGenericESProducer = cms.ESProducer("PixelCPEGenericESProducer",
 # This customizes the Run3 Pixel CPE generic reconstruction in order to activate the IrradiationBiasCorrection
 # because of the expected resolution loss due to radiation damage
 from Configuration.Eras.Modifier_run3_common_cff import run3_common
-run3_common.toModify(PixelCPEGenericESProducer, IrradiationBiasCorrection = cms.bool(True))
+run3_common.toModify(PixelCPEGenericESProducer, IrradiationBiasCorrection = True)
 
 # This customization will be removed once we get the templates for phase2 pixel
 # FIXME::Is the Upgrade variable actually used?
@@ -76,5 +76,3 @@ phase2_tracker.toModify(PixelCPEGenericESProducer,
   DoCosmics = False,
   Upgrade = cms.bool(True)
 )
-
-
--- a/RecoLocalTracker/SiPixelRecHits/python/PixelCPEGeneric_cfi.py
+++ b/RecoLocalTracker/SiPixelRecHits/python/PixelCPEGeneric_cfi.py
@@ -59,6 +59,11 @@ PixelCPEGenericESProducer = cms.ESProducer("PixelCPEGenericESProducer",
     MagneticFieldRecord = cms.ESInputTag(""),
 )
 
+# This customizes the Run3 Pixel CPE generic reconstruction in order to activate the IrradiationBiasCorrection
+# because of the expected resolution loss due to radiation damage
+from Configuration.Eras.Modifier_run3_common_cff import run3_common
+run3_common.toModify(PixelCPEGenericESProducer, IrradiationBiasCorrection = cms.bool(True))
+
 # This customization will be removed once we get the templates for phase2 pixel
 # FIXME::Is the Upgrade variable actually used?
 from Configuration.Eras.Modifier_phase2_tracker_cff import phase2_tracker
@@ -71,4 +76,5 @@ phase2_tracker.toModify(PixelCPEGenericESProducer,
   DoCosmics = False,
   Upgrade = cms.bool(True)
 )
+
 


perrotta:
please test

perrotta:
Phase2 workflows are failing in the tests...

mmusich:
> Phase2 workflows are failing in the tests...

@perrotta please send the tests again...

perrotta:
please test
(I had missed that there was already an update since the previous launch of the tests)

pmaksim1:
Not to be flip, but I can't imagine us not needing this in Phase 2...

mmusich:
> Not to be flip, but I can't imagine us not needing this in Phase 2...

Hi @pmaksim1, sure (indeed that was my original intention), but this results in:

```
Exception Message:

ERROR: UseErrorsFromTemplates_ is set to False in PixelCPEGeneric_cfi.py.  In this case it does not make sense to set any of the following to True:  TruncatePixelCharge_, IrradiationBiasCorrection_, DoCosmics_, LoadTemplatesFromDB_ !!!
``` 

I suppose we first need to activate the `UseErrorsFromTemplates_` for Phase-2 workflows (which requires having the proper GerErrors in GlobalTag and is already in the plans) and then activate the IrradiationBiasCorrection for phase-2. Does it sound like a plan?

pmaksim1:
@mmusich  Yes, that sounds like a plan.  We are going to produce the Phase 2 templates; we started with that but then detoured into reorganizing and streamlining the template production (which is globally useful, since half of was still in Fortran...)

mmusich:
@perrotta we need to propagate this change also in the HLT menu, but only for Run3, i.e. I don't want to change the frozen menus run in relvals
What is the policy for that? Shall I create a customize function in
 `/HLTrigger/Configuration/python/customizeHLTforCMSSW.py`?
Thanks in advance

RecoLocalTracker/SiPixelRecHits/python/PixelCPEGeneric_cfi.py
```diff
@@ -59,6 +59,11 @@
     MagneticFieldRecord = cms.ESInputTag(""),
 )
 
+# This customizes the Run3 Pixel CPE generic reconstruction in order to activate the IrradiationBiasCorrection
+# because of the expected resolution loss due to radiation damage
+from Configuration.Eras.Modifier_run3_common_cff import run3_common
+run3_common.toModify(PixelCPEGenericESProducer, IrradiationBiasCorrection = cms.bool(True))
```
slava77: please drop type specifications for all parameters which already exist.

slava77:
> we need to propagate this change also in the HLT menu, but only for Run3, i.e. I don't want to change the frozen menus run in relvals
> What is the policy for that? Shall I create a customize function in
> `/HLTrigger/Configuration/python/customizeHLTforCMSSW.py`?

If it goes through customizeHLTforCMSSW, I guess it would have to be a method that loops over all instances of types PixelCPEGenericESProducer. It can be picked up with ```run3_common.makeProcessModifier(methodName or lambda)```

The run3 HLT in the matrix tests is currently fake. So, this thing will have no effect.

I guess a more appropriate way is to have a proper table with the right parameter set via confdb.

@Martin-Grunewald please advise.

slava77:
> The motivation for activating the correction is due to the expected comparatively large loss of resolution along the local y [global-z in barrel] direction for an heavily irradiated detector (as expected for the end of run3 and extended end or run3 simulation scenarios), in which carrier trapping significantly reduces Lorentz drift induced charge sharing.

Is the correction additionally driven by some GT/database parameters?
If not, we need something more creative than just ```run3_common``` so that the start of run3 is not too pessimistic.

Is there a significant effect for Run2? If so, perhaps this should be enabled more broadly.

mmusich:
@slava77 

> Is the correction additionally driven by some GT/database parameters?

it is not "driven" by GT/database parameter, but the GenError we plan to use for the simulation of Run3 have been derived explicitly by assuming that this parameter is switched on in the CPE configuration.

> so that the start of run3 is not too pessimistic.

in what sense too pessimistic? This is an additional correction for high irradiation, it should not make anything worse than it already is.

> Is there a significant effect for Run2? If so, perhaps this should be enabled more broadly.

I think it needs to be studied. 
By the way, to give a bit of context @tsusa is preparing a set of slides about the impact of the correction the run3 simulation aiming to show them tomorrow's RECO/AT meeting.

RecoLocalTracker/SiPixelRecHits/python/PixelCPEGeneric_cfi.py
mmusich: done

slava77:
> in what sense too pessimistic? This is an additional correction for high irradiation, it should not make anything worse than it already is.

The way I read it, is there is no dependence on other external parameters and only run3_common is needed. At the same time, the correction refers to the end of run 3.
I'm not sure how the correction for "high irradiation" be agnostic of the value of the irradiation.
Perhaps I should just wait for tomorrow to see more details .

(should I add you and @tsusa  to the agenda?)

mmusich:
> The way I read it, is there is no dependence on other external parameters and only run3_common is needed. At the same time, the correction refers to the end of run 3.

Well, we studied the effect on projected 2021 simulation and it does close to nothing, but it improves things substantially on 2023 and 2024 projections. IF we had year-dependent eras for run3 as well we could consider having this switched on only later, but this idea was met with overwhelmingly negative reactions during the next-to-last PPD coordination meeting. 

> (should I add you and @tsusa to the agenda?)

yes, please!

slava77:
@cmsbuild please test

mmusich:
> > we need to propagate this change also in the HLT menu, but only for Run3, i.e. I don't want to change the frozen menus run in relvals
> > What is the policy for that? Shall I create a customize function in
> > `/HLTrigger/Configuration/python/customizeHLTforCMSSW.py`?
> 
> If it goes through customizeHLTforCMSSW, I guess it would have to be a method that loops over all instances of types PixelCPEGenericESProducer. It can be picked up with `run3_common.makeProcessModifier(methodName or lambda)`
> 
> The run3 HLT in the matrix tests is currently fake. So, this thing will have no effect.
> 
> I guess a more appropriate way is to have a proper table with the right parameter set via confdb.
> 
> @Martin-Grunewald please advise.


@slava77 @Martin-Grunewald 
Could you please advise how to move on?
I tried this:

```diff
diff --git a/HLTrigger/Configuration/python/customizeHLTforCMSSW.py b/HLTrigger/Configuration/python/customizeHLTforCMSSW.py
index 4f429b96d56..6c731ba6717 100644
--- a/HLTrigger/Configuration/python/customizeHLTforCMSSW.py
+++ b/HLTrigger/Configuration/python/customizeHLTforCMSSW.py
@@ -47,11 +47,21 @@ def customiseFor2017DtUnpacking(process):
 
     return process
 
+def customiseFor27183(process) :
 
+    def _activateIrradiationBiasCorrection(process):
+        for producer in esproducers_by_type(process, "PixelCPEGenericESProducer"):
+            producer.IrradiationBiasCorrection = cms.bool(True)
+
+    from Configuration.Eras.Modifier_run3_common_cff import run3_common
+    run3_common.makeProcessModifier(_activateIrradiationBiasCorrection)
+ 
 # CMSSW version specific customizations
 def customizeHLTforCMSSW(process, menuType="GRun"):
 
     # add call to action function in proper order: newest last!
     # process = customiseFor12718(process)
 
+    process = customiseFor27183(process)
+
```
but it fails the addOnTests. What is the right way of using `makeProcessModifier`?

slava77:
ehm, perhaps it's easier to ```run3_common.toModify(process, _activateIrradiationBiasCorrection)``` ?

Martin-Grunewald:
Indeed we do not have a formal Run3 menu yet. We need to discuss within TSG on how to proceed in general.
However, you could simply use fillDescriptions, with code default OFF and reco-modifiers used to turn it on for Run3. Then you do not need any special HLT customisation.

slava77:
On 6/17/19 12:02 PM, Martin Grunewald wrote:
> However, you could simply use fillDescriptions, with code default OFF
> and reco-modifiers used to turn it on for Run3. Then you do not need any
> special HLT customisation.

This is not a case of a new parameter. This parameter is present since a
long time.
The question was how to change its value for Run3 HLT.
The proposal was to change it using an era.
However, if the preference is to do it based on a specific Run-3 HLT table,
then this will have to be deferred to the time when that table is built.
Perhaps some JIRA or similar ticket should be open to not forget this
change.

slava77:
I guess my last comment was from a lack of knowledge of the procedure how an HLT table is built.
If an instance of ```ESProducer("PixelCPEGenericESProducer"``` in HLT is coming from parsing RecoLocalTracker/SiPixelRecHits/python/PixelCPEGeneric_cfi.py (e.g. cloning from the ```PixelCPEGenericESProducer``` defined there, then we are already done with this PR.

Martin-Grunewald:
It comes from parsing, but the values of existing parameters of menus in ConfDB are not changed when the cfi is parsed again. Shouldn't that ensure the 'old' behaviour (old value for the parameter)?

slava77:
On 6/17/19 12:36 PM, Martin Grunewald wrote:
> It comes from parsing, but the values of existing parameters of menus in
> ConfDB are not changed when the cfi is parsed again. Shouldn't that
> ensure the 'old' behaviour (old value for the parameter)?
> 

The old behavior will indeed be preserved.
However, from physics considerations, this parameter has to be changed
to have an appropriate response.

Martin-Grunewald:
OK, so for now no change in any HLT customisation (and in this PR) until/unless we get a Run-3 menu.

mmusich:
> OK, so for now no change in any HLT customisation (and in this PR) until/unless we get a Run-3 menu.

Thanks @Martin-Grunewald. Shall we open a JIRA ticket to remind us about it in the future? If yes where should it be posted? Thanks again.

Martin-Grunewald:
Please submit here: https://its.cern.ch/jira/projects/CMSHLT

slava77:
Looking at the jenkins outputs for comparisons in the run3 ttbar workflow (11624, no PU),
it looks like the hit resolution in y is getting significantly worse in BRL1
wf11624_brl1_y

Changes in the x coordinate are less significant.

What does this mean?
Are the corrections inconsistent with biases introduced in the SIM/DIGI step?

mmusich:
> Looking at the jenkins outputs for comparisons in the run3 ttbar workflow (11624, no PU),
it looks like the hit resolution in y is getting significantly worse in BRL1

can you point to the actual jenkins comparison for this plot?

slava77:
> can you point to the actual jenkins comparison for this plot?

It can be seen in the DQM bin-to-bin 
http://tinyurl.com/y5pztemo

mmusich:
> What does this mean?
> Are the corrections inconsistent with biases introduced in the SIM/DIGI step?

@slava77 
thanks to @tsusa we have produced the SimHit-RecHit comparison for the setup tested in the PR jenkins tests and we are able to reproduce the resolution loss effect in the 2021 workflow (compared with what already shown by Tanja during the [last RECO/AT meeting](https://indico.cern.ch/event/827369/contributions/3467663/attachments/1862678/3061762/tanjaReco_14062019.pdf)):

image
 
The reason for such resolution loss is that the Pixel Generic Error payload derivation needs to take into account the fact that the `IrradiationBiasCorrection` is switched on in the reconstruction. 
When this is properly taken into account there is no resolution loss (see blue curve in the plot posted above).
We have done that explicitly when deriving the updated pixel conditions for 2021 (already publicly announced to AlCa HN [here](https://hypernews.cern.ch/HyperNews/CMS/get/calibrations/4146.html)), but these conditions haven't made yet the autoCond in release.
Now, we don't have a version of the GenError payload for the current 2021 GenError condition in GT (which is outdated anyway) and it would be very anti-economic to produce one just to make this PR pass, as in the long run we need to update them.
I can propose either one of these several options:
   1. as AlCa needs to update the 2021 GT for the `Run3Summer19DR` production (cf. [here](https://hypernews.cern.ch/HyperNews/CMS/get/prep-ops/6490/1/1/1.html)) I can propose the full updated GT in this PR as well (please notice that it will contain also other updates in addition to the CPE conditions)
   2. we can wait for the update of the conditions to percolate in release and then activate the parameter here. 
   3. I can propose a minimal GT update in this PR in which only the CPE conditions are changed.

Please advice on what would be the optimal commit flow.

slava77:
> I can propose either one of these several options:
> 
>     1. as AlCa needs to update the 2021 GT for the `Run3Summer19DR` production (cf. [here](https://hypernews.cern.ch/HyperNews/CMS/get/prep-ops/6490/1/1/1.html)) I can propose the full updated GT in this PR as well (please notice that it will contain also other updates in addition to the CPE conditions)
> 
>     2. we can wait for the update of the conditions to percolate in release and then activate the parameter here.
> 
>     3. I can propose a minimal GT update in this PR in which only the CPE conditions are changed.

IIUC, option 1. will introduce changes in some other payloads and there will be many more changes in the comparisons.

I can imagine that 3. may not be preferred by AlCa because it's an additional burn of a GT version number for something relatively small.

Option 2. will just take longer.

Since either of the 3 provides a consistent final result, all are somewhat acceptable.
I'd personally prefer the option 3.

slava77:
this is still coming out with a merge conflict

mmusich:
@slava77 
> this is still coming out with a merge conflict

sorry for the clumsy commit... now should be OK.

mmusich:
the PR description has been updated to take into account the GT update as well, I implemented choice n. 3 of https://github.com/cms-sw/cmssw/pull/27183#issuecomment-502802729

slava77:
@cmsbuild please test

slava77:
The error does not really look related to this PR

```
===== Test "test_CreateFileLists" ====
=======================================
This script is moved to its new home in
tkal_create_file_lists.py              
(in Alignment/CommonAlignment/scripts).
=======================================

  >>> Please create proxy via 'voms-proxy-init -voms cms -rfc'.

---> test test_CreateFileLists had ERRORS
```

@fabiocos @smuzaffar 
do you recognize this as something appearing in regular IBs?

mmusich:
> Comparison is ready
> https://cmssdt.cern.ch/SDT/jenkins-artifacts/pull-request-integration/PR-ef1fd2/992/summary.html
> 
> Comparison Summary:
> 
>     * No significant changes to the logs found
> 
>     * Reco comparison results: 2151 differences found in the comparisons
> 
>     * DQMHistoTests: Total files compared: 33
> 
>     * DQMHistoTests: Total histograms compared: 3254096
> 
>     * DQMHistoTests: Total failures: 4745
> 
>     * DQMHistoTests: Total nulls: 0
> 
>     * DQMHistoTests: Total successes: 3249017
> 
>     * DQMHistoTests: Total skipped: 334
> 
>     * DQMHistoTests: Total Missing objects: 0
> 
>     * DQMHistoSizes: Histogram memory added: 0.0 KiB( 32 files compared)
> 
>     * Checked 137 log files, 14 edm output root files, 33 DQM output files

I was hoping to check the resolution from the DQM bin-by-bin comparison as suggested previously, but I see nothing in the GUI: http://tinyurl.com/yy4t6l48 ... 
Is this expected?

smuzaffar:
@slava77 , looks like /etc/vomses directory on cmsbuildXX VMs is empty. that is why we are getting messages like 
```
+ voms-proxy-init -voms cms
Your identity: /DC=ch/DC=cern/OU=Organic Units/OU=Users/CN=cmsbot/CN=545661/CN=Robot: CMS Build Bot
Cannot find file or dir: /home/cmsbld/.glite/vomses
VOMS Server for cms not known!
```
in IBs. I haev updated bot to make use of /cvmfs/grid.cern.ch/etc/grid-security/vomses is local /etc/vomses is empty.

mmusich:
>  I have updated bot to make use of /cvmfs/grid.cern.ch/etc/grid-security/vomses is local /etc/vomses is empty.

so can tests be re-triggered here?

smuzaffar:
@cmsbuild please test

andrius-k:
@mmusich I ran the comparison locally for the workflow 11624.0 which contains lots of changes. The results are in the GUI and the url from the results page will now work.

fabiocos:
@slava77 the problem with the unit test has happened in the last days, I've already communicated with @smuzaffar about this

mmusich:
@slava77 
thanks to @andrius-k that has manually re-run the DQM bin-by-bin comparison one can see now the comparison with the updated PR: http://tinyurl.com/y2ar7bgw
image
It seems to me the resolution problem is solved.

jfernan2:
@mmusich apart from the fix on Pixel resolution, the changes in this PR are so deep that they are affecting almost all subsystems in the Run3 wf, even HCAL and ECAL RecHits and CaloTowers...
http://tinyurl.com/y3x626bd
Is this expected?

jfernan2:
OK, answering myself, I realize now that ALCA conditions are different too, please ignore my previos message, sorry

mmusich:
> OK, answering myself, I realize now that ALCA conditions are different too, please ignore my previous message, sorry.

@jfernan2 the changes in the ALCA conditions are affecting **only** the Pixel subsystem. 
The fact that there are changes in other subdetectors is due to the fact that since the (pixel) digitizer conditions are updated there are fluctuation due to the random number generator service usage in the digitization step.

christopheralanwest:
@mmusich Your candidate GTs do not contain two tags that the Hypernews thread [[MC][GT][Run3] Pixel conditions for Run3 studies](https://hypernews.cern.ch/HyperNews/CMS/get/calibrations/4146/2.html) indicates are needed. If you confirm that these tags should be used, I can create auto:phase1_2021_realistic (106X_upgrade2021_realistic_v8) and auto:phase1_2021_cosmics (106X_upgrade2021cosmics_realistic_deco_v3) GTs for you that includes all of the 2021 pixel updates.

mmusich:
> @mmusich Your candidate GTs do not contain two tags that the Hypernews thread [[MC][GT][Run3] Pixel conditions for Run3 studies](https://hypernews.cern.ch/HyperNews/CMS/get/calibrations/4146/2.html) indicates are needed. 

@christopheralanwest it is done on purpose, please try to read carefully https://github.com/cms-sw/cmssw/pull/27183#issuecomment-502780836 and https://github.com/cms-sw/cmssw/pull/27183#issuecomment-502802729 ... 
(only the CPE conditions are actually needed for this PR, the other two conditions mentioned in the thread would just add spurious differences)

>If you confirm that these tags should be used, I can create auto:phase1_2021_realistic (106X_upgrade2021_realistic_v8) and auto:phase1_2021_cosmics (106X_upgrade2021cosmics_realistic_deco_v3) GTs for you that includes all of the 2021 pixel updates.

as far as I can see these two GTs do not exist, but if you want to make my candidates real GTs that would actually be better. You can create another set of GTs with the full package (including the other two) at a later time and update autoCond in another PR.

christopheralanwest:
@mmusich Thanks for pointing me to the relevant section of the thread. I would rather use the candidate GTs for now in case some other development arises and needs to be merged before this PR. Once this PR has a reco signature, I will submit a PR that cherry-picks your GT update and then updates the two GTs to use the full 2021 pixel conditions. That will allow the two PRs to be tested separately or together as needed, without generating merge conflicts.

Configuration/AlCa/python/autoCond.py
```diff
@@ -66,9 +66,9 @@
     # GlobalTag for MC production with perfectly aligned and calibrated detector for Phase1 2021
     'phase1_2021_design'       : '106X_upgrade2021_design_v3', # GT containing design conditions for Phase1 2021
```
slava77: shouldn't there also be an update for phase1_2021_design or is this GT filled with some other independent/trivial conditions?

mmusich: @slava77, at the moment there is no need to update the 2021 design GT, since conditions there are "trivial" -  in the sense that we don't simulate any radiation damage. At any rate @tsusa has run the resolution analysis using Single Muon pT=10 gun, in the following setup:
```
cmsDriver.py -s GEN,SIM,DIGI,L1,DIGI2RAW,RAW2DIGI,RECO --evt_type SingleMuPt10_pythia8_cfi --fileout file:GENSIMRECO_MuPt10.root --python_filename=Run3_2021_generic_IBC.py -n 10 --no_exec --era Run3 --beamspot  GaussSigmaZ4cm --nThreads=4 —customise_commands="process.mix.digitizers.pixel.ThresholdInElectrons_BPix_L1 = cms.double(1300.0);  process.mix.digitizers.pixel.ThresholdInElectrons_BPix_L2 = cms.double(1600.0);process.mix.digitizers.pixel.ThresholdInElectrons_BPix = cms.double(1600.0); process.mix.digitizers.pixel.ThresholdInElectrons_FPix = cms.double(1600.0);process.PixelCPEGenericESProducer.IrradiationBiasCorrection = True”  --conditions auto:phase1_2021_design
```

and we don't observe significant deviations from the current performance. You can find the results of the test [here](https://cernbox.cern.ch/index.php/s/LP1Tu9Kg30xVG2l).

slava77:
Here are some highlights from ttbar no-PU workflow 11624 (1K events).

The track uncertainties are down, which is reflected in the PV position uncertainties (by about 20%, if I'm reading off the log scale right)
all_sign1081VSorig_TTbar13TeV2021wf11624p0c_log10recoVertexs_offlinePrimaryVertices__RECO_obj_xError

About 7% fewer initialStep tracks
all_sign1081VSorig_TTbar13TeV2021wf11624p0c_recoTracks_generalTracks__RECO_obj_originalAlgo
This is partly due to loss of efficiency
wf11624_initialStep_hp_eff_eta


It looks like more things than just what's immediately related to the irradiation bias correction have changed.

At the pixel level (also can be seen with less stats in  in the DQM gui posted earlier http://tinyurl.com/y5hjwlbw ):
There is more charge in the on-track clusters (most visible in BPIX1)
wf11624_BPIX1_onTrack

The x resolutions for the hits have some bias
wf11624_BPIX1_resx
The y direction in BPIX1 looks OK for the resolution
wf11624_BPIX1_rh_resy

There is some small bias in FPIX x:
wf11624_FPIX1_m_pullx
wf11624_FPIX1_p_pullx


The y pulls look better, closer to 1 (the most significant effect is in BPIX1: Gaussian width changes from 0.61 to 0.94 for a fit within -2, 2)
wf11624_BPIX1_pully
a similar improvement is visible in FPIX as well
wf11624_FPIX1_p_pully

The cluster prob changed, most visibly in BPIX1.
wf11624_BPIX1_cluProb
From the loss of efficiency I'm guessing that this change is for the worse.


Overall, the efficiency loss (driven by cluster prob?) seems a bit of a showstopper.


I ran with ```IrradiationBiasCorrection = False``` and it seems like by itself it has a fairly small effect compared to the first order large effects mentioned above.
[In the following several plots the baseline in black is ```IrradiationBiasCorrection = False``` with otherwise this PR and red is this PR as is]
x and y resolutions get a little bit worse in the inner layers
wf11624_noibc_BPIX1_resx
wf11624_noibc_BPIX1_rh_resy
wf11624_noibc_BPIX1_pully

Another significant effect is in FPIX, but here it's going in the right direction to reduce the average bias/offset from 0.
wf11624_noibc_FPIXn1_resx
wf11624_noibc_FPIXp1_resx

So, incrementally, just doing ```IrradiationBiasCorrection = True``` on top of the GT seems better now (y res in BPIX1 is only a little worse and an increased bias in x res is probably too small to worry much).
However, relative to the baseline, the GT change seems either incomplete or partly incorrect (see cluster probs and initialStep efficiency changes).

mmusich:
> However, relative to the baseline, the GT change seems either incomplete or partly incorrect (see cluster probs and initialStep efficiency changes).

@slava77 I think we are simply *correctly* simulating more radiation damage now...

slava77:
+1

for https://github.com/cms-sw/cmssw/pull/27183	9310ec1
- this includes a coherent GT update together with enabling the irradiation correction in the pixel CPE. Only Run3 is affected.
- jenkins tests pass and comparisons with the baseline show roughly expected behavior
- local tests with more events also show roughly expected behavior as described in https://github.com/cms-sw/cmssw/pull/27183#issuecomment-504928586
    - some loss in tracking efficiency is attributed to more correct modeling of expected irradiation effects https://github.com/cms-sw/cmssw/pull/27183#issuecomment-504946665

mmusich:
> It looks like more things than just what's immediately related to the irradiation bias correction have changed.

@slava77 thanks for the sign-off.
Just to give a more detailed explanation on https://github.com/cms-sw/cmssw/pull/27183#issuecomment-504946665 (maybe a bit too dry) I detailed what has changed in the GT in the PR description (please have a look at:  https://github.com/cms-sw/cmssw/pull/27183#issue-287467997). 
The radiation damage simulation which is going through the 2D templates has changed:
```
SiPixel2DTemplateDBObjectRcd  denominator  SiPixel2DTemplateDBObject_phase1_BoR3_HV350_Tr1300_den             SiPixel2DTemplateDBObject_phase1_38T_mc_v4_den_unirradiated     
SiPixel2DTemplateDBObjectRcd  numerator    SiPixel2DTemplateDBObject_phase1_BoR3_HV350_Tr1300_num             SiPixel2DTemplateDBObject_phase1_38T_2018_ultralegacymc_v1_num  
```
The whole CPE package (reconstruction **and** digitization conditions) have been updated. These come as an indivisible block as they are all derived with a target integrated dose (in this case end of 2021).   
Also, for completeness, we still expect to update the tracker alignment on top of these changes. 
By the way this is what will be used for the `Run3Summer19DR` campaign.

christopheralanwest:
+1

fabiocos:
+1

PR #27211, opened 2019-06-14
FWCore/Framework: Move function call outside of assert

gartung:
Found while building with spack. If NDEBUG is set by an external product header before assert is included then the assert macro effectively does not run whatever is inside the assert macro. In this case the function call
moduleChanger()->changeModule(m_tag.label(), newPSet)
was never compiled into the code which causes the test to fail.


gartung:
please test

Dr15Jones:
+1

fabiocos:
+1

PR #27215, opened 2019-06-14
Pf cluster validation

akanugan:
**PR description:**

Added an analyzer to get  PFCluster summed energies ratio plots and some profile histograms. 

**PR validation:**

Added an analyzer to get PFCluster summed energies plots and some profile histograms. runTheMatrix test workflows. The plots produced locally good enough. Plugin, interface and cfg files are created. Finally, modified macro and .html files to get the final set of plots.
Added plots are in the last 4 rows in the below link
http://hep.baylor.edu/cmshcal/akanugan/PFClusterValidation3/10_6_0_pre4_vs_10_6_0_pre4_SinglePi/

@mariadalfonso 
@abdoulline
 @bendavid
@arapyan

### DIFF
--- a/Validation/RecoParticleFlow/plugins/PFClusterValidation.cc
+++ b/Validation/RecoParticleFlow/plugins/PFClusterValidation.cc
@@ -86,15 +86,13 @@ void PFClusterValidation::bookHistograms(DQMStore::IBooker & ibooker, edm::Run c
 } // BOOKING HISTOS
 
 
-double   phi_MC = 9999.;
-double   eta_MC = 9999.;
-double partR  = 0.3;
-double Rmin   = 9999.;
  
 void PFClusterValidation::analyze(edm::Event const& event, edm::EventSetup const& c) {
   
   nevent++;
-  
+  //double   phi_MC = 9999.;
+  //double   eta_MC = 9999.;
+ 
    
   if (imc != 0){
     edm::Handle<edm::HepMCProduct> evtMC;
@@ -115,9 +113,14 @@ void PFClusterValidation::analyze(edm::Event const& event, edm::EventSetup const
       double phip = (*p)->momentum().phi();
       double etap = (*p)->momentum().eta();
       double pt  = (*p)->momentum().perp();
-      if(pt > maxPt) { npart++; maxPt = pt; phi_MC = phip; eta_MC = etap; }
+      if(pt > maxPt) { 
+	npart++; 
+	maxPt = pt; 
+	phi_MC = phip; 
+	eta_MC = etap; 
+      }
     }
-    //  std::cout << "*** Max pT = " << maxPt <<  std::endl;  
+    //std::cout << "eta and phi MC: " << eta_MC << phi_MC <<  std::endl;  
   }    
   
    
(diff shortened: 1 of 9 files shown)


akanugan:
@cmsbuild please test

civanch:
please test

jfernan2:
please test

akanugan:
please test

akanugan:
I have added a config file which was missing in the repo.
Now, the above matrix tests are tested and all of them passed.
I think it will pass the Jenkins test now.

jfernan2:
please test

Validation/RecoParticleFlow/plugins/PFClusterValidation.cc
```diff
@@ -0,0 +1,271 @@
...
+      consumes<reco::PFClusterCollection>(conf.getUntrackedParameter<edm::InputTag>("pflowClusterECAL"));
+  PFClusterHCALTok_ =
+      consumes<reco::PFClusterCollection>(conf.getUntrackedParameter<edm::InputTag>("pflowClusterHCAL"));
+  PFClusterHOTok_ = consumes<reco::PFClusterCollection>(conf.getUntrackedParameter<edm::InputTag>("pflowClusterHO"));
+  PFClusterHFTok_ = consumes<reco::PFClusterCollection>(
+      conf.getUntrackedParameter<edm::InputTag>("pflowClusterHF"));  // cms.InputTag("particleFlowClusterECAL");
+
+  outputFile_ = conf.getUntrackedParameter<std::string>("outputFile", "myfile.root");
+  mc_ = conf.getUntrackedParameter<std::string>("mc", "yes");
+
```
jfernan2: @akanugan why do you use here a string instead of a bool?

akanugan: @jfernan2 Hi, I can make that change but the same string was used in another module which I carried forward. If it's necessary, I can change it.

hatakeyamak: I think @akanugan meant 
https://cmssdt.cern.ch/lxr/source/Validation/CaloTowers/src/CaloTowersValidation.cc
Well, maybe we can fix this as well, i.e. use bool.
I think
@mariadalfonso
@abdoulline
@arapyan
will be happy if @akanugan does that?

abdoulline:
Yes, Kenich. This string "yes" comes from old (dark) times...

Validation/RecoParticleFlow/plugins/PFClusterValidation.cc
akanugan: The change has been made and also it passed the tests.

jfernan2:
please test

civanch:
+1

jfernan2:
+1

fabiocos:
@smuzaffar @jfernan2 in our guidelines we state "Do not split plugins into header and source files. If you do split them, keep the header files in the .../plugins directory." Here we do split the new PFClusterValidation, as the code is relatively long (not terribly but) I think this could be anyway acceptable.

fabiocos:
+1

PR #27220, opened 2019-06-17
add the ability to use a seed just as a 5DHit

VinInn:
With this PR we add the ability (controlled by a configurable in the trajectory builder) to use the seed ad a 5DHit.
It targets mainly HLT when the seeds come from an outside source.
w/r/t standard hitless seeds it allows to fully use the seed information in the trajectory fit.

while technically working it will require to be really operational:
   * seeds with a meaningful covariance matrix
   * retuning of the trajectory filter
   * retuning of the track classifier
   * and most probably additional work at level of HitPattern and of clients that rely on specific hits to be present in the track

tested offline and hit with default configuration and with 5DHit switched on

### DIFF
--- a/HLTrigger/Configuration/python/customizeHLTforCMSSW.py
+++ b/HLTrigger/Configuration/python/customizeHLTforCMSSW.py
@@ -18,7 +18,7 @@ from HLTrigger.Configuration.common import *
 #     return process
 
 # use 5DHit as seed
-def customiseFor99999(process):
+def customiseFor27220(process):
    for pset in process._Process__psets.values():
        if hasattr(pset,'ComponentType'):
              if (pset.ComponentType == 'CkfTrajectoryBuilder' or pset.ComponentType == 'GroupedCkfTrajectoryBuilder'):
@@ -60,6 +60,6 @@ def customiseFor2017DtUnpacking(process):
 def customizeHLTforCMSSW(process, menuType="GRun"):
 
     # add call to action function in proper order: newest last!
-    process = customiseFor99999(process)
+    process = customiseFor27220(process)
 
     return process
--- a/HLTrigger/Configuration/python/customizeHLTforCMSSW.py
+++ b/HLTrigger/Configuration/python/customizeHLTforCMSSW.py
@@ -21,7 +21,9 @@ from HLTrigger.Configuration.common import *
 def customiseFor27220(process):
    for pset in process._Process__psets.values():
        if hasattr(pset,'ComponentType'):
-             if (pset.ComponentType == 'CkfTrajectoryBuilder' or pset.ComponentType == 'GroupedCkfTrajectoryBuilder'):
+             if (pset.ComponentType == 'CkfTrajectoryBuilder' or 
+                 pset.ComponentType == 'GroupedCkfTrajectoryBuilder' or
+                 pset.ComponentType == 'MuonCkfTrajectoryBuilder'):
                  if not hasattr(pset,'seedAs5DHit'):
                      pset.seedAs5DHit = cms.bool(False)
    return process
--- a/HLTrigger/Configuration/python/customizeHLTforCMSSW.py
+++ b/HLTrigger/Configuration/python/customizeHLTforCMSSW.py
@@ -17,6 +17,14 @@ from HLTrigger.Configuration.common import *
 #                     pset.minGoodStripCharge = cms.PSet(refToPSet_ = cms.string('HLTSiStripClusterChargeCutNone'))
 #     return process
 
+# use 5DHit as seed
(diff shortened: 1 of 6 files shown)


VinInn:
How the is supposed to work with pixel GPU products?

1) TrajectorySeeds will be built directly from SOA transferred from GPU.
1a) Skipping TrajectorySeeds or even thinking to change its content to host directly a Hit5D is asking for infinite trouble 
1b) in the operation some additional filter can be added (pixel tracks from gpu can be of various quality: some sort of iterative tracking can be built just selecting first the best and then the one of lower quality)
1c) some sort of hit-proxy (for instance just recording the layer or even the det) can be stored in the seed to be consumed by those building quality out of this information

2) standard tracking can then start on CPU w/o any need to access pixel clusters or hits
2a) of course the whole workflow should be fully consistent in using only gpu products for what pixel is concerned: mixing 5DHits iterations with standard ones will not work (unless the cluster-remover is modified to account for the hits already used on GPU) 

questions and comments are welcome

slava77:
>     2\. standard tracking can then start on CPU w/o any need to access pixel clusters or hits
>         2a) of course the whole workflow should be fully consistent in using only gpu products for what pixel is concerned: mixing 5DHits iterations with standard ones will not work (unless the cluster-remover is modified to account for the hits already used on GPU)

I think that ability to run seeding for all (at least CA-based) pixel iterations on GPU is important.
We can not rely on quadruplets-only from hltPixelTracks.

VinInn:
Producing triplets is trivial (the current code is already able to produce inner triplets)
It is just a matter of proper configuration
I think is better to produce them all in one go, then CPU clients selects those that fit best their workflow

pixelpairs are there as "Cells" : before moving them to CPU a proper filtering is required (depending on the use case)

slava77:
> Producing triplets is trivial (the current code is already able to produce inner triplets)
> It is just a matter of proper configuration
> I think is better to produce them all in one go, then CPU clients select those that fit best their workflow

This is nice.
Will this then cover what's currently used for seeding hltIter[0,1,2] in one single step?
Is this still hypothetical or was such setup studied already?

VinInn:
> 
> Will this then cover what's currently used for seeding hltIter[0,1,2] in one single step?
That's the idea
> Is this still hypothetical or was such setup studied already?
> 
I would say still "hypothetical"
there is no obvious show stopper, essentially one needs to
1) tune cuts for the missing doublets (1,3, 2,4)
2) tune cuts for the triplets 1,2,4 and 1,3,4
3) make the whole thing configurable
4) mark "used" the cells and hits in the (good?) quadruplets
5) build triplets

after the whole saga about PixelPair in 2017&18 data I would prefer FIRST to have clear statement of their real usefulness before going in implementation

v.

>

makortel:
> I think is better to produce them all in one go, then CPU clients select those that fit best their workflow

Does this mean then that the GPU seeding would use the most relaxed tracking region of the targeted iterations? (assuming the current iteration structure is more-or-less kept, which does not necessarily have to be the case)

VinInn:
> On 17 Jun, 2019, at 3:45 PM, Matti Kortelainen <notifications@github.com> wrote:
> 
> Does this mean then that the GPU seeding would use the most relaxed tracking region of the targeted iterations? (assuming the current iteration structure is more-or-less kept, which does not necessarily have to be the case)
I was thinking to try to go global at least for quads and triplets "high-pt"
The effective tracking region can then been applied on CPU to limit the pattern-recognition in the outer-tracker
we need to understand what to do for the specialized seeding with very relaxed cuts (for the time being do them fully on CPU if they affect a minimal event fraction)
We wil never be able to port on GPU all use-cases (trigger paths) : we do not need to, enought to address the majority of the events ....


v.

slava77:
@cmsbuild please test

VinInn:
do not understand the logic of the failures: why 10824.0 step2 fails and  10024.0 does not?
will have to check one by one all the failing modules....
v.

slava77:
10024is running a fake hlt
Perhaps that's the reason

VinInn:
```
33 32 31 25 12 2 1 1 1 tests passed, 0 0 0 0 0 0 0 0 0 failed
```

VinInn:
bizzare the code check: first times passed, second time askes indentation to be fixed...

VinInn:
@cmsbuild , please test

slava77:
+1

for https://github.com/cms-sw/cmssw/pull/27220	9a15a2b
- code changes are in line with the PR description; the optional (disabled) implementation for ```seedAs5DHit``` looks visually correct

Martin-Grunewald:
+1

fabiocos:
+1

Issue #27219, opened 2019-06-17
Failure in wf 136.7801 (and others) after integration of PR #26814


fabiocos:
After the integration of #26814 in CMSSW_11_0_X_2019-06-16-2300 there are a few failures in some IB test workflows (136.7801, 136.7803, 1040.0) all with the same problem:

```
----- Begin Fatal Exception 17-Jun-2019 10:20:24 CEST-----------------------
An exception of category 'ProductNotFound' occurred while
   [0] Processing  Event run: 297227 lumi: 1 event: 48150 stream: 3
   [1] Running path 'dqmoffline_step'
   [2] Calling method for module PhotonIDValueMapProducer/'photonIDValueMapProducer'
Exception Message:
Principal::getByToken: Found zero products matching all criteria
Looking for a container with elements of type: reco::Photon
Looking for module label: slimmedPhotons
Looking for productInstanceName: 

   Additional Info:
      [a] If you wish to continue processing events after a ProductNotFound exception,
add "SkipEvent = cms.untracked.vstring('ProductNotFound')" to the "options" PSet in the configuration.

----- End Fatal Exception -------------------------------------------------
```

The problem happens in the DQM sequence when the photonIdValueMapProducer is invoked

fabiocos:
assign reconstruction,dqm

@guitargeek please have a look at this

fabiocos:
The problem is reproducible, and adding #26814 on top of CMSSW_11_0_X_2019-06-16-0000 is enough to see it

guitargeek:
Ok no big deal I hope, there was a usage of the `photonIDValueMapProducer` which was not tested in the bot tests and that requires the gedPhotons collection instead of the slimmedPhotons, which I made the new default instead of auto-detecting the right collection:

https://github.com/cms-sw/cmssw/blob/master/DQMOffline/Trigger/python/HLTEGTnPMonitor_cfi.py#L992

I will just fix it in the same way as the `egmPhotonIDs` were treated two likes below: cloning the object and then resetting the input collection. Or do I even need to clone it? I will go the safe way and just clone it for my fixup PR.

guitargeek:
Hi @fabiocos, I proposed a bugfix: https://github.com/cms-sw/cmssw/pull/27221.

fabiocos:
Problem fixed

PR #26814, opened 2019-05-16
Adapting MVAVariableHelper to new E/gamma UL format and removal of MultiToken

guitargeek:
#### PR description:

As the Egamma MVA infrastructure does not need to compute object specific variables anymore that are not stored in the objects themselves (thanks to https://github.com/cms-sw/cmssw/pull/26486), the `MVAVariableHelper` is now an overly complicated piece of code which should be adapted.

I also suggest to remove the MultiToken class I introduced last year, as it's not really used much anymore after all the changes in the last months. It was also used in the ElectronIDValueMapProducer, which can now be removed that all the variables are in the object. I also think I made a mistake introducing the MultiToken to begin with. The consumes should be better configured with Python... That makes the MVAValueMapProducer also a bit simpler to use as it's not ambiguous anymore how it can be configured.

Other changes:
* adapted Electron MVA FWLite python code and test (thanks to the appropriate relvals being available now)
* merge header and source of some plugins in Electron-PhotonIdentification
* put all the plugin template definitions in Electron-PhotonIdentification into plugins.cc file

#### PR validation:

CMSSW compiles and local matrix tests pass.

#### if this PR is a backport please specify the original PR:
 
Maybe the FWLite part should get backported to 10_6_X.

PR #27221, opened 2019-06-17
Bugfix after "Adapting MVAVariableHelper to new E/gamma UL format and removal of MultiToken"

guitargeek:
#### PR description:

PR https://github.com/cms-sw/cmssw/pull/26814 caused some workflows in the IB to crash, see issue https://github.com/cms-sw/cmssw/issues/27219.

There was a usage of the `photonIDValueMapProducer` which was not tested in the bot tests and that requires the gedPhotons collection instead of the slimmedPhotons, which I made the new default instead of auto-detecting the right collection:

https://github.com/cms-sw/cmssw/blob/master/DQMOffline/Trigger/python/HLTEGTnPMonitor_cfi.py#L992

I propose to just fix it in the same way as the `egmPhotonIDs` were treated a few lines below: cloning the object and then resetting the input collection. This  should  be the safest way, as I can't run the matrix tests due to not being able to access the necessary files right now...

__Note:__ needs to be tested with workflows 136.7801, 136.7803, 1040.0 explicitly, as these are the ones that were crashing and they are not covered by the limited matrix tests performed by the bot.

#### PR validation:

```
python  DQMOffline/Trigger/python/EgammaMonitoring_cff.py
```


jfernan2:
please test

guitargeek:
Hi @jfernan2, sorry I forgot to mention this originally, but it seems the standard bot tests don't cover the crashing workflows. I wrote more in the PR description now. So please can you test the 3 crashing workflows after the standard tests are done? Thank you!

jfernan2:
please abort

jfernan2:
please test workflow 136.7801,136.7803,1040.0

jfernan2:
@guitargeek slimmedPhotons are not available:
https://cmssdt.cern.ch/SDT/jenkins-artifacts/pull-request-integration/PR-2e8516/981/runTheMatrix-results/136.7801_RunHLTPhy2017B_AOD+RunHLTPhy2017B_AOD+DQMHLTonAOD_2017+HARVESTDQMHLTonAOD_2017/step2_RunHLTPhy2017B_AOD+RunHLTPhy2017B_AOD+DQMHLTonAOD_2017+HARVESTDQMHLTonAOD_2017.log
I suggest testing your PR off-topic, you should be able to do it yourself with this kind of commands:
https://cmssdt.cern.ch/SDT/jenkins-artifacts/pull-request-integration/PR-2e8516/981/runTheMatrix-results/136.7801_RunHLTPhy2017B_AOD+RunHLTPhy2017B_AOD+DQMHLTonAOD_2017+HARVESTDQMHLTonAOD_2017/cmdLog

guitargeek:
Yep I will do that now, I was a bit overconfident that this commit here would fix it already. I will test locally and push anew when my solution works locally.

guitargeek:
Okay with this update it works now locally. I was mislead by the line `egmPhotonIDsForDQM.physicsObjectSrc == cms.InputTag('gedPhotons')` in the existing code which I copy pasted, but it actually does not seem to work. For `egmPhotonIDsForDQM` that was not a problem because the default collection was `gedPhotons` anyway. I figured out that I need to pass the parameters I want to change directly in the `clone()` method.

jfernan2:
please test workflow 136.7801,136.7803,1040.0

fabiocos:
@jfernan2 according to the test this PR looks to solve the failures in the IB. Do you have comments? Or may we move forward with it?

jfernan2:
+1

fabiocos:
+1

PR #27222, opened 2019-06-17
DD4hep: Sim Application Based on DD4hep

ianna:
#### PR description:

* Sim Application implementation and its full configuration for discussion with @civanch 
* Retrieving G4 production cuts, but not yet setting them
* No sensitive volume assignment yet

### DIFF
--- a/SimG4Core/Geometry/src/DDDWorld.cc
+++ b/SimG4Core/Geometry/src/DDDWorld.cc
@@ -33,21 +33,18 @@ DDDWorld::DDDWorld(const DDCompactView *cpv,
   catalog = ret.sdCatalog();
 }
 
-DDDWorld::DDDWorld(const cms::DDDetector* ddd, dd4hep::sim::Geant4GeometryMaps::VolumeMap& map) {
+DDDWorld::DDDWorld(const cms::DDDetector *ddd, dd4hep::sim::Geant4GeometryMaps::VolumeMap &map) {
+  LogVerbatim("SimG4CoreApplication") << "DD4hep_DDDWorld: initialization of DDDWorld...";
 
-  LogVerbatim("SimG4CoreApplication") 
-    << "DD4hep_DDDWorld: initialization of DDDWorld...";
-  
   DetElement world = ddd->description()->world();
-  const Detector& detector = *ddd->description();
+  const Detector &detector = *ddd->description();
   Geant4Converter g4Geo(detector);
-  Geant4GeometryInfo* geometry = g4Geo.create(world).detach();
-  map = geometry->g4Volumes;   
+  Geant4GeometryInfo *geometry = g4Geo.create(world).detach();
+  map = geometry->g4Volumes;
   m_world = geometry->world();
   SetAsWorld(m_world);
 
-  LogVerbatim("SimG4CoreApplication") 
-    << "DD4hep_DDDWorld: initialization of DDDWorld done.";
+  LogVerbatim("SimG4CoreApplication") << "DD4hep_DDDWorld: initialization of DDDWorld done.";
 }
 
 DDDWorld::~DDDWorld() {}
--- a/SimG4Core/Geometry/src/DDDWorld.cc
+++ b/SimG4Core/Geometry/src/DDDWorld.cc
@@ -2,6 +2,12 @@
 #include "SimG4Core/Geometry/interface/DDG4Builder.h"
 
 #include "DetectorDescription/Core/interface/DDCompactView.h"
+#include "DetectorDescription/DDCMS/interface/DDDetector.h"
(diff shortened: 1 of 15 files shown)


ianna:
please test

ianna:
+1

Discussed with Vladimir, the PR should go in to allow further migration. Noted comments will be added later as a separate PR.

civanch:
+1
private runs shows proper results

fabiocos:
@ianna thank you for this new implementation, this indeed goes along the direction we have discussed.
I have a question also for @civanch @Dr15Jones : obviously the geometry implementation should be equivalent in the long term between DD and DD4help. But for the time being I wonder whether we aren't interested in keeping track of which of these tools has been used to produce a given result. In this case the flag
```
m_pGeoFromDD4hep(iConfig.getUntrackedParameter<bool>("g4GeometryDD4hepSource", false)),
```
should be tracked. What do you think?

fabiocos:
@kpedro88 of course the question is also for you...

Dr15Jones:
Making the parameter tracked seems fine to me.

civanch:
I think "tracked", I also sure it is not the last PR on these files (we had a long discussion this Monday with Carl and Ianna). So, I would propose to merge this and we will continue.

kpedro88:
The changes proposed here look reasonable, but I would still like to see a full roadmap for the migration. Are more details available?

ianna:
@kpedro88 - this will be discussed during the planning meeting next week. In the meantime both me and Vladimir need this PR for further development.

SimG4Core/Geometry/src/DDDWorld.cc
```diff
@@ -25,6 +33,44 @@ DDDWorld::DDDWorld(const DDCompactView *cpv,
...
+  printout(INFO,"SimDD4CMS",
+	   "+++ DDDWorld::DDDWorld start... %s",
+	   world.name());
+  const Detector& detector = *ddd->description();
+  Geant4Converter g4Geo(detector);
+  Geant4GeometryInfo* geometry = g4Geo.create(world).detach();
+  map = geometry->g4Volumes;
+  
+  auto it = geometry->g4Volumes.find(detector.worldVolume().ptr());
+  std::cout << "The world is " << it->first.name() << "\n";
```
kpedro88: MessageLogger should be used rather than cout
(you can put all the messages into a stringstream and then just pass stringstream.str() to the MessageLogger)

Dr15Jones: There is no need to use a `stringstream` as a MessageLogger message is not sent until the local logger guard is destroyed
```C++
edm::LogInfo("Geometry") << "The world is " << it->first.name();
```
is perfectly fine.

kpedro88: In this case there is a message filled (below these lines) in a for loop. (I guess this can also be done with a lambda now, but maybe a stringstream is easier.)

ianna: yes, these and also the DDDWorld API changes discussed and agreed with @civanch are due as a next PR.

Dr15Jones:
```C++
auto l = edm::LogInfo("Geometry");
for(...) {
    l << "some message";
}
```
would accumulate the message and only output after the loop.

kpedro88:
@ianna in general it's not clear why minor review comments have to wait for a subsequent PR. (I am quite reluctant to sign anything that uses cout in potential production code, i.e. anything outside of a test directory.)

ianna:
please test

SimG4Core/Geometry/src/DDDWorld.cc
civanch: To say more: this line is not needed at the end, when DD4Hep will work, so may be changed to DEBUG.

Let us clean up in the next iteration, as I said before we will work on clean up of this.

ianna:
please test

ianna:
+1

civanch:
+1

ianna:
@kpedro88 - updated as requested, please review

kpedro88:
+upgrade

ianna:
@fabiocos - who is signing for the operations? It's an orthogonal to the production workflows configuration.

fabiocos:
@ianna it's me, this is in the queue for next IB. Apparently my comment about the tracked parameter was ignored so far, I will not block the PR because of this but I think we should consider it, unless there is an explicit argument against

fabiocos:
+operations

the update of Configuration/StandardSequences looks coherent with the purpose of the PR

ianna:
@fabiocos - thanks, your request was noted and will be implemented next.

fabiocos:
+1

PR #27237, opened 2019-06-18
Revert "Add some reset calls to TrackingMonitorClient."

schneiml:
This reverts commit a0c43f3ea4e248d9b0b318bc1b566f1e8f36e5c0.

#### PR description:

@mmusich suspects that this is causing too many plots to be reset. So let's try without it.
This might cause double counting in other places, lets see.

#### PR validation:

none.

Needs forward port if tests agree.

### DIFF
--- a/DQM/TrackingMonitorClient/src/TrackingQualityChecker.cc
+++ b/DQM/TrackingMonitorClient/src/TrackingQualityChecker.cc
@@ -363,17 +363,6 @@ void TrackingQualityChecker::fillTrackingStatus(DQMStore::IBooker & ibooker, DQM
     }
   }
 
-  // After harvesting, all per-lumi MEs are reset, to make sure we only get
-  // events of the new lumisection next time.
-  for (std::map<std::string, TrackingMEs>::iterator it = TrackingMEsMap.begin();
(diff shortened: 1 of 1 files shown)


schneiml:
please test

mmusich:
thanks @schneiml  but is not only me to have this suspicion . 
We are running some other tests without the lines and see what happens.
Let me add @jandrea @hbecerri and @arossi83

mmusich:
see also ongoing thread at: https://hypernews.cern.ch/HyperNews/CMS/get/dqmDevel/2412/1/2/1.html

mmusich:
@schneiml is there a master PR as well?

schneiml:
@mmusich no master PR for now, we'll need to decide if we go with this or #27239 based on the test results. I hope the PR tests actually do something useful on these plots.

mmusich:
@schneiml I am sorry, but I don't follow how you actually validate the improvement. You won't see it in standard relval tests.

schneiml:
http://tinyurl.com/y2m5mbxb shows the original, filled state. Since #27239 does the same but more cleanly, let's go with that.

PR #27239, opened 2019-06-18
DQM/Tracking: Improve the per-lumi fix [10_6_X]

schneiml:
#### PR description:

The fix for double counting in a0c43f3ea4e248d seems to be a bit aggressive. It seems to have reset some MEs that are not per-lumi, so check for that.

#### PR validation:
Bin-to-bin comparison shows the plots in question re-populated.

#### if this PR is a backport please specify the original PR:

This is used instead of #27237, and a backport of #27246.

PR #27246, opened 2019-06-19
DQM/Tracking: Improve the per-lumi fix

schneiml:
#### PR description:

The fix for double counting in a0c43f3ea4e248d seems to be a bit aggressive. It seems to have reset some MEs that are not per-lumi, so check for that.

#### PR validation:

After the tests of #27239 and #27237, it looks like this should bring back the missing plots, while not risking to bring back bugs due to double-counting.

Old test comparison: http://tinyurl.com/yyanjesa

### DIFF
--- a/DQM/TrackingMonitorClient/src/TrackingQualityChecker.cc
+++ b/DQM/TrackingMonitorClient/src/TrackingQualityChecker.cc
@@ -441,7 +441,9 @@ void TrackingQualityChecker::fillTrackingStatus(DQMStore::IBooker& ibooker, DQMS
     std::string localMEdirpath = it->second.HistoDir;
     std::vector<MonitorElement*> tmpMEvec = igetter.getContents(ibooker.pwd() + "/" + localMEdirpath);
     for (auto ime : tmpMEvec) {
-      ime->Reset();
+      if (ime->getLumiFlag()) {
+        ime->Reset();
+      }
     }
   }
 


schneiml:
please test

fioriNTU:
please test

mmusich:
@andrius-k, @kmaeshima, @schneiml, @jfernan2, @fioriNTU 
What is preventing this from being signed?

fioriNTU:
+1

fioriNTU:
I guess we forgot about it, sorry @mmusich

fabiocos:
@schneiml @fioriNTU @mmusich @mtosi could you please clarify why in the DQM test differences are observed only for a subset of sample? Are they as expected?

schneiml:
@fabiocos most likely because the specific harvesting code only runs under special circumstances, or only when more than one lumisection is processed. I have not investigated this specific case, but i am not surprised.

fabiocos:
@schneiml well it would be good that we understand what we see before putting this into the release. I am not claiming that it is wrong, I just want to ensure that it is understood.

jfernan2:
The plots are now filled in wf 4.22 and 7.3 (Cosmics) which were not filled before the PR, those are the 103 plots being declared as modified by DQM bin by bin tool, not sure why RelMon is not declaring them as changed

mmusich:
> @schneiml well it would be good that we understand what we see before putting this into the release. I am not claiming that it is wrong, I just want to ensure that it is understood.

@fabiocos without these changes the Strip offline client is broken. Do we want to go blind for MWGR?

schneiml:
@fabiocos We also got the hint from @mmusich that this depends on the `commonSiStripZeroBias` sequence which is not used on the data relvals (though it should be, eventually). The tests where we see changes are cosmics data, where the respective functionality is enabled as well.

Also as far as I see @hbecerri has locally run additional tests to prove that the changes fix what they are supposed to (while the PR tests show that they at least don't break any of the tested wfs).

fabiocos:
@mmusich let's avoid useless panic please, MWGR is in about one month from now, I am just asking experts to confirm that what is seen in tests (whose reason is not so obvious to me, and apparently not even to @schneiml ) is what is expected before merging the code, and explaining why it changes one workflow and not another one.
I see the argument above, I understand that Tracker DPG sign off this update. The PR will enter next IB.

fabiocos:
+1

PR #27276, opened 2019-06-19
MTD geometry: post D41 technical fixes

fabiocos:
#### PR description:

This PR adds a revised version of the ```mtd.xml``` file for scenario D41 where:

- the ETL material budget is adjusted for a missing aluminum ring (~5% of X0), used for material budget plots for TDR;

- the warnings for duplicated definition of material are removed by a redefinition, no impact on SimHits is anyway observed in checks.

This is the stripped version of a complete new scenario implementation https://github.com/fabiocos/cmssw/tree/fc-MTDpostD41 based on this revised geometry. At present it does not seem really necessary, and I prefer to avoid proliferation of not strictly needed new scenarios,
but I would like to have an official reference for this revised implementation.

#### PR validation:

No overlap is present (even if plugged in tracker geometry T15), the numbering scheme is obviously untouched, if plugged in a new scenario the SingleMu workflow runs smoothly.


fabiocos:
@kpedro88 this is the updated code I was telling you, I do not think that in itself it justifies a new scenario 9although fully ready), unless there are concurrent needs

kpedro88:
@fabiocos sure, since it is here it can be incorporated later (after the big cleanup etc.)

fabiocos:
please test 

dummy formal test, the added file is used nowhere at present

kpedro88:
+upgrade

fabiocos:
@ianna @cvuosalo so far this is essentially a do-nothing documentation PR, please let me know whether you have objections or we may move forward with it

cvuosalo:
+1

fabiocos:
+1

PR #27279, opened 2019-06-19
Phase2-hgx200 Add a new Phase2 scenario D43

bsunanda:
#### PR description:

Add a new scenario with the V9 geometry of HGCal and T14 geometry of the tracker. This is similar to the D41 scenario except that HGCal resembles the TDR design. The corresponding workflow is created at 29834.0

#### PR validation:

Tested using the workflow 29838.0

To be used for testing the impact of HGCal design on muon chambers. No backport is needed


bsunanda:
@cmsbuild Please test

Configuration/Geometry/python/dict2023Geometry.py
```diff
@@ -1104,6 +1104,7 @@
     ("O2","T14","C3","M2","F2","I1") : "D40",
     ("O3","T14","C8","M3","F2","I9") : "D41",
     ("O2","T15","C3","M2","F2","I1") : "D42",
+    ("O2","T14","C4","M2","F2","I7") : "D43",
```
kpedro88: is it intentional to use older muon and MTD versions here? This makes it not an exact copy of D41.

bsunanda: The only difference was in the RPC specpars. Now I have made it M3

bsunanda:
@cmsbuild Please test

Configuration/Geometry/python/dict2023Geometry.py
kpedro88: Ah, and now I recall that I9 and I7 are the same except for the HGCal positioning.

kpedro88:
+upgrade

Configuration/Geometry/python/dict2023Geometry.py
fabiocos: @kpedro88 @bsunanda I9 and I7 are identical as far as BTL is concerned, but ETL is moved and "hammered" (i.e. less active modules) to fit in the new mother volume in D41.  So HGCal v9 cannot use MTD I9

kpedro88: @fabiocos thanks for confirming

Configuration/PyReleaseValidation/python/relval_2023.py
```diff
@@ -25,5 +25,6 @@
 numWFIB.extend([28634.0]) #2023D40
 numWFIB.extend([29034.0]) #2023D41
 numWFIB.extend([29434.0]) #2023D42
+numWFIB.extend([29834.0]) #2023D43
```
fabiocos: @bsunanda I repeat my question: do you need this scenario to be routinely run in every PR test? I understood that having only the upgrade set with ```--what upgrade``` was enough...

kpedro88: The above change will just run it in IB tests, not PR tests (not added to short matrix)

fabiocos:
+operations

the update of Configuration/Geometry is correct and coherent with the purpose of this PR

Configuration/PyReleaseValidation/python/relval_2023.py
fabiocos: you are right, this is not updating the limited selection in runTheMatrix.py https://cmssdt.cern.ch/lxr/source/Configuration/PyReleaseValidation/scripts/runTheMatrix.py#0081 , so this will just enter the standard list. One more workflow is not a big deal, I withdraw my question.

fabiocos:
@cvuosalo @prebello @zhenhu could you please check and sign or comment? After this PR @kpedro88 will move forward with a Phase2 scenario cleaning...

cvuosalo:
+1

fabiocos:
@prebello @zhenhu could you please have a look?

fabiocos:
+1

fabiocos:
please test workflow 29834.0

fabiocos:
merge

PR #27288, opened 2019-06-20
LZMA and ZSTD support for Streamer format

smorovic:
#### PR description:
PR adds new compression formats and also optimizes buffer usage in Streamer event writing.

Output buffering changes:
- Output buffering changed to reduce extra copy of the payload. Instead, small header is copied to the front of the payload after compression. Initial default of 7 MB is not needed and can be sized dynamically (usually to much less). New initial size is 50 kb to reserve for event header.

- Reduced ROOT TBuffer (TBufferFile object) size after serialization of the INI file, which, due to typically large INI payloads in HLT, results in significant reduction in memory usage of the process during event processing.

- In testing with the realistic 2018 HLT payload and data with the 4-thread HLT (pp Physics) CMSSW job, these two changes reduced total memory footprint from 2.7 GB to 2 GB (per process).

Compression changes:
- LZMA compression algorithm added. To facilitate automatic detection in reader, a
4-byte header is present at the beginning of LZMA content (this does not
conflict with zlib header which starts with a different sequence). Compression
/decompression functions are based on ROOT implementation, but we turn off
calculation of CRC32/64 checksum (adler32 is already used to checksum whole payload). Error handling is also improved and 16 MB buffer size limitation in ROOT routine is removed.

- Zstandard compression is added. Compression factor is comparable to zlib, but is performing faster at lower compression levels making it potentially interesting for HLT in situations where CPU usage is tight. Similar 4-byte header ("ZS") is added to the payload.

- "compression_algorithm" parameter was added to the output module allowing to choose between algorithms.

- Decompression for both new algorithms with format autodetection is added to the StreamerInputSource.

- several "scram b code-checks-all" changes to Streamer and Utililities modules has also been included with the PR.

#### PR validation:

Buffering and compression changes have been verified to produce correct files by producing output files  with different algorithms (DAQ and Streamer output module) and reading them with Streamer input source to verify integrity.

### DIFF
--- a/IOPool/Streamer/interface/StreamSerializer.h
+++ b/IOPool/Streamer/interface/StreamSerializer.h
@@ -18,7 +18,6 @@
 #include "DataFormats/Provenance/interface/SelectedProducts.h"
 #include "FWCore/Utilities/interface/get_underlying_safe.h"
 
-
 // Data structure to be shared by all output modules for event serialization
 struct SerializeDataBuffer {
   typedef std::vector<char> SBuffer;
--- a/IOPool/Streamer/interface/StreamSerializer.h
+++ b/IOPool/Streamer/interface/StreamSerializer.h
@@ -18,12 +18,12 @@
 #include "DataFormats/Provenance/interface/SelectedProducts.h"
 #include "FWCore/Utilities/interface/get_underlying_safe.h"
 
-const int init_size = 0;  //will be allocated on first event
-const unsigned int reserve_size = 50000;
 
 // Data structure to be shared by all output modules for event serialization
 struct SerializeDataBuffer {
   typedef std::vector<char> SBuffer;
+  static constexpr int init_size = 0;  //will be allocated on first event
+  static constexpr unsigned int reserve_size = 50000;
 
   SerializeDataBuffer()
       : comp_buf_(reserve_size + init_size),
--- a/IOPool/Streamer/interface/StreamSerializer.h
+++ b/IOPool/Streamer/interface/StreamSerializer.h
@@ -18,45 +18,52 @@
 #include "DataFormats/Provenance/interface/SelectedProducts.h"
 #include "FWCore/Utilities/interface/get_underlying_safe.h"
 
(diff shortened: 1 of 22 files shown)


slava77:
@cmsbuild please test

IOPool/Streamer/interface/StreamSerializer.h
```diff
@@ -18,77 +18,100 @@
 #include "DataFormats/Provenance/interface/SelectedProducts.h"
 #include "FWCore/Utilities/interface/get_underlying_safe.h"
 
-const int init_size = 1024 * 1024;
+const int init_size = 0;  //will be allocated on first event
```
Dr15Jones: `constexpr` would be better

IOPool/Streamer/interface/StreamSerializer.h
```diff
@@ -18,77 +18,100 @@
...
+const unsigned int reserve_size = 50000;
```
Dr15Jones: `constexpr`

Dr15Jones: Better to add both inside `struct SerializeDataBuffer` to avoid having these names be in the global namespace.

Dr15Jones (changes requested):
Would be nice to see tests running the different compression options.

IOPool/Streamer/interface/StreamSerializer.h
smorovic: Thanks for comments. I moved them inside this struct.

smorovic:
@Dr15Jones there are now unit tests for each compression algorithm option.

slava77:
@cmsbuild please test

slava77:
+1

for https://github.com/cms-sw/cmssw/pull/27288	6b04fa1
- reco code is essentially not affected; the reco signature requirement was triggered via a shared package EventFilter/Utilities. 

mommsen:
+1

Dr15Jones:
+1

fabiocos:
+1

PR #27289, opened 2019-06-20
Cleanup of RecoLuminosity/LumiDB

plujan:
#### PR description:

Most of the files in RecoLuminosity/LumiDB are obsolete -- they were scripts for working with the Run 1 lumi DB, which no longer exists and has been superseded by brilcalc. In this PR:
- the scripts in plotdata/ are removed, since these are now maintained independently by the lumi POG (<https://github.com/CMS-LUMI-POG/PublicPlots>)
- all of the unused scripts (which is nearly all of them) are removed
- split the constants used in lumiQueryAPI.py into their own file so those can be saved
- estimatePileup_makeJSON_2015.py has been extensively rewritten, since there was a lot of dead or otherwise not useful code, and renamed to makePileupJSON.py, since it's not 2015 any more!)
- pileupCalc.py has been slightly modified (mostly just giving more helpful informational and warning messages, and more sensible defaults for Run 2)
- the other few items are kept unchanged

#### PR validation:

I've run through the normal process of generating a pileup histogram (brilcalc -> makePileupJSON.py -> pileupCalc.py) and verified that the resulting histogram is exactly the same. Also checked to make sure that nothing outside RecoLuminosity/LumiDB depends on any of these modules.


slava77:
@cmsbuild please test

plujan:
sorry for the late push -- I realized there was another feature I should add, can you retest?

fabiocos:
please test

ggovi:
+1

fabiocos:
+1

Issue #26669, opened 2019-05-06
UBSAN and boost serialization are incompatible


Dr15Jones:
I looked at the UBSAN failure for CondFormats/Common. When running edmCheckClassVersion we get a segmentation fault during the shared library load. Running the code in gdb and looking a the source code

```
From
 boost::serialization::void_cast_detail::void_caster_primitive<cond::GenericSummary, cond::Summary>::void_caster_primitive (this=0x3d36f70) at /cvmfs/cms-ib.cern.ch/nweek-02574/slc7_amd64_gcc700/external/boost/1.67.0-pafccj/include/boost/serialization/void_cast.hpp:187

184	        // note:I wanted to displace from 0 here, but at least one compiler
185	        // treated 0 by not shifting it at all.
186	        reinterpret_cast<std::ptrdiff_t>(
187	            static_cast<Derived *>(
188	                reinterpret_cast<Base *>(8)
189	            )
190	        ) - 8
```

So what the code appears to be doing is seeing if the `Derived` and `Base` class parts of an object start at the same address or if they are displaced (which can happen with multiple inheritance). 

I then dumped the assembly language for the same. Here I guess that UBSAN has injected code that checks to see if the `static_cast<Derived*>` or `reinterpret_cast<Base*>` is actually pointing to the type specified. It seems to do this by taking the address of the object and then look to see if the virtual table is as expected (at least that is my guess). The dereferencing of `8` leads to the segmentation fault.

In principal, this is OK (but a bit strange) C++ without the UBSAN check. With the check it crashes.

So it looks like no code that includes `boost/serialization/void_cast.hpp` and causes the `void_caster_primitive` to be instantiated can be compiled using UBSAN. 

I don't know if this means any package directly or indirectly using boost serialization can't be compiled with UBSAN or if we can be more isolating.

Dr15Jones:
assign core

Dr15Jones:
FYI @smuzaffar

Dr15Jones:
From the debugger, the problematic stack is

```
#5  0x00007f5329e59ac2 in boost::serialization::void_cast_detail::void_caster_primitive<cond::GenericSummary, cond::Summary>::void_caster_primitive (this=0x3d36f70) at /cvmfs/cms-ib.cern.ch/nweek-02574/slc7_amd64_gcc700/external/boost/1.67.0-pafccj/include/boost/serialization/void_cast.hpp:187
[...]

PR #27312, opened 2019-06-22
Remove UBSAN flags from CondFormats/HcalObjects

Dr15Jones:
#### PR description:

CondFormats/HcalObjects directly uses boost serialization which is known to cause problems when compiled with UBSAN.

#### PR validation:

The code now compiles and links using CMSSW_11_0_UBSAN_X_2019-06-21-2300 IB. Previously we were getting a segmentation fault during the edmCheckClassVersion call.

### DIFF
--- a/CondFormats/HcalObjects/BuildFile.xml
+++ b/CondFormats/HcalObjects/BuildFile.xml
@@ -1,4 +1,7 @@
 <flags   GENREFLEX_ARGS="--"/>
+<ifrelease name="_UBSAN_">
+  <flags   REM_CXXFLAGS="-fno-omit-frame-pointer -fsanitize=undefined"/>
+</ifrelease>
 <use   name="CondFormats/Common"/>
 <use   name="DataFormats/DetId"/>
 <use   name="DataFormats/HcalDetId"/>


Dr15Jones:
This is related to #26669

Dr15Jones:
@smuzaffar I had to directly use `REM_CXXFLAGS` instead of being able to say 'exclude from UBSAN'. Is this how you ultimately want to handle this case?

Dr15Jones:
please test

Dr15Jones:
Note, this is the only package that directly uses boost serialization. Is this how the db L2's want this to be?

smuzaffar:
@Dr15Jones , you do not need to do this. SCRAM now does not pass UBSAN flags when boost serialization file (Serialization.cc) is compiled.

Dr15Jones:
@smuzaffar This case is different. For the rest of the CondFormats packages, they are only indirectly dependent on boost serialization. That is, the code in the package itself does not depend on boost serialization, it is only the autogenerated Serialization.cc file which has the dependency.

CondFormats/HcalObjects is different. The source code in the package itself directly calls boost serialization macros. Under UBSAN those macros are causing segmentation faults when running edmCheckClassVersion (which is what was also happening to the other packages because of Serialization.cc). So for this case we need to explicitly stop UBSAN from compiling the package source directly.

smuzaffar:
@Dr15Jones , in this case then please use
```
<ifrelease name="_UBSAN_">
  <flags   REM_CXXFLAGS="-fno-omit-frame-pointer -fsanitize=undefined"/>
</ifrelease>
```

Dr15Jones:
please test

fabiocos:
@ggovi @christopheralanwest @tocheng could you please check?

ggovi:
+1

pohsun:
+1

fabiocos:
+1

PR #27336, opened 2019-06-25
Override the default value of the input file for wf 250203.18

fabiocos:
#### PR description:

Override the default value of the input file for wf 250203.18 to solve the persistent DAS error in IB after #27206 .

#### PR validation:

Until the files are cached at CERN the wf will not work out of the box, but the correct query string is produced:
```
10:07 cmsdev25 661> cat cmdLog 

# in: /build/fabiocos/110X/crash/CMSSW_11_0_X_2019-06-24-1100/work going to execute cd 250203.18_H125GGgluonfusion_13UP18+H125GGgluonfusion_13UP18INPUT+DIGIPRMXUP18_PU25+RECOPRMXUP18_PU25+HARVESTUP18_PU25
 dasgoclient --limit 0 --query 'file dataset=/RelValH125GGgluonfusion_13/CMSSW_10_6_0-106X_upgrade2018_realistic_v4-v2/GEN-SIM site=T2_CH_CERN' | sort -u > step1_dasquery.log  2>&1
 
```

and

```
10:07 cmsdev25 662> dasgoclient --limit 0 --query 'file dataset=/RelValH125GGgluonfusion_13/CMSSW_10_6_0-106X_upgrade2018_realistic_v4-v2/GEN-SIM'
/store/relval/CMSSW_10_6_0/RelValH125GGgluonfusion_13/GEN-SIM/106X_upgrade2018_realistic_v4-v2/10000/B9F429B4-1EC9-A049-B70F-0F53757926FA.root
/store/relval/CMSSW_10_6_0/RelValH125GGgluonfusion_13/GEN-SIM/106X_upgrade2018_realistic_v4-v2/10000/2C8B55AC-9316-8844-970B-653596580104.root
```

PR #27341, opened 2019-06-25
Adding Run 330098 (2019 MWGR3) in data relvals

boudoul:
#### PR description:
Greetings
This PR aims to introduce a new run 3 data workflow based on fresh run3 data taken during MWGR3 .
The goal is to test the run3 reconstruction with real data and anticipate any issue without waiting for the usual Tier0 replay prior the global runs - Eventually we can update with more fresh runs once we get some.

#### PR validation:
I ran the following workflow runTheMatrix.py -l 138.1 --command "-n 100"  

It may need to be backported in 10_6 (depending on the MWGR 4 plan) but let's make a review in master first

### DIFF
--- a/Configuration/AlCa/python/autoCond.py
+++ b/Configuration/AlCa/python/autoCond.py
@@ -43,6 +43,8 @@ autoCond = {
     'run2_hlt_relval_hi'   :   '103X_dataRun2_HLT_relval_HI_v4',
     # GlobalTag for Run2 HLT for HI (not 2018 HI): it points to the online GT
     'run2_hlt_hi'       :   '101X_dataRun2_HLTHI_frozen_v9',
+    # GlobalTag for Run3 data relvals
+    'run3_data_promptlike'         :   '106X_dataRun3_Prompt_v2',
     # GlobalTag for MC production with perfectly aligned and calibrated detector for Phase1 2017 (and 0,0,~0-centred beamspot)
     'phase1_2017_design'       :  '110X_mc2017_design_v1',
     # GlobalTag for MC production with realistic conditions for Phase1 2017 detector
--- a/Configuration/PyReleaseValidation/python/relval_standard.py
+++ b/Configuration/PyReleaseValidation/python/relval_standard.py
@@ -410,6 +410,8 @@ workflows[136.896] = ['',['RunCharmonium2018D','HLTDR2_2018','RECODR2_2018reHLT_
 # multi-run harvesting
 workflows[137.8] = ['',['RunEGamma2018C','HLTDR2_2018','RECODR2_2018reHLT_skimEGamma_Offline_L1TEgDQM',
                         'RunEGamma2018D','HLTDR2_2018','RECODR2_2018reHLT_skimEGamma_Prompt_L1TEgDQM','HARVEST2018_L1TEgDQM_MULTIRUN']]
+### LS2 - MWGR ###
+workflows[138.1] = ['',['RunCosmics2019','RECOCOSDRUN3','ALCACOSDRUN3','HARVESTDCRUN3']]
 
 ### fastsim ###
 workflows[5.1] = ['TTbar', ['TTbarFS','HARVESTFS']]
--- a/Configuration/PyReleaseValidation/python/relval_steps.py
+++ b/Configuration/PyReleaseValidation/python/relval_steps.py
@@ -475,6 +475,9 @@ steps['RunSingleMu2015HLHS']={'INPUT':InputInfo(dataSet='/SingleMuon/Run2015D-v1
 steps['RunCosmics2015C']={'INPUT':InputInfo(dataSet='/Cosmics/Run2015C-v1/RAW',label='2015C',run=[256259],events=100000,location='STD')}
 steps['RunCosmics2016B']={'INPUT':InputInfo(dataSet='/Cosmics/Run2016B-v1/RAW',label='2016B',run=[272133],events=100000,location='STD')}
 
+### LS2 - MWGR ###
+steps['RunCosmics2019']={'INPUT':InputInfo(dataSet='/ExpressCosmics/Commissioning2019-Express-v1/FEVT',label='2019GR3',run=[330098],events=100000,location='STD')}
+
 def gen(fragment,howMuch):
     global step1Defaults
     return merge([{'cfg':fragment},howMuch,step1Defaults])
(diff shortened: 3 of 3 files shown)


boudoul:
I would need : 
- comments from PdmV  (@pgunnell,@prebello,@zhenhu) : if they are happy with the way I'm introducing this WF ( WF number, naming etc..) 
- comments from Alca ( @christopheralanwest,@franzoni,@tlampen,@pohsun,@tocheng )  : I  used a candidate GT kindly prepared by @christopheralanwest a couple of weeks ago, and defined it as run3_data in autocond. Would it be possible to get a 'real GT' ( so that I can update this PR with it and not leave a candidate GT in this workflow )  - Thanks.

Configuration/AlCa/python/autoCond.py
```diff
@@ -44,6 +44,8 @@
     # GlobalTag for Run2 HLT for HI (not 2018 HI): it points to the online GT
     'run2_hlt_hi'       :   '101X_dataRun2_HLTHI_frozen_v9',
     # GlobalTag for MC production with perfectly aligned and calibrated detector for Phase1 2017 (and 0,0,~0-centred beamspot)
+    'run3_data'         :   '106X_dataRun3_Express_Candidate_2019_05_24_21_36_45',
```
mmusich: @boudoul in the past 'run2_data' key was reserved for offline / re-reco Global Tags. I think to preserve the nomenclature something like `run3_data_promptlike` would be more appropriate. By the way do we have a Run3 prompt queue?

fabiocos:
The overlap with #27336 is not going to produce conflicts

Configuration/AlCa/python/autoCond.py
christopheralanwest: I agree that `run3_data` should be reserved for offline global tags. 

@boudoul Do we plan to have prompt reco for the MWGRs or only express? We do have both express and prompt queues for Run 3: "106X_dataRun3_Express_Queue" and "106X_dataRun3_Prompt_Queue". But if we are not planning to have prompt reco in the near future, we should have an express GT here: there's no point in validating a GT that we don't plan to use. I would be inclined to use the key name `run3_data_promptlike` in either case.

Configuration/AlCa/python/autoCond.py
```diff
@@ -44,6 +44,8 @@
...
+    # GlobalTag for Run3 data relvals
```
christopheralanwest: The comments in L46 and L48 should be switched so that the comment is above the corresponding key-value pair. Please try to maintain the same whitespace in the unchanged lines to avoid irrelevant merge conflicts.

boudoul: Hi @christopheralanwest , thanks for your comments (sorry for the delay, CMS week..) We should consider having prompt reconstruction in MWGR, do you have a name of a GT I should use  ?  (I will also use un3_data_promptlike as you recommended)   - And for the line switching , thanks for spotting , it was not intentionnal to change , I will fix in a next commit. Thanks again .

boudoul:
Dear  Alca ( @christopheralanwest,@franzoni,@tlampen,@pohsun,@tocheng )- Could you please let me know the name of the GT I should use, in order for me to edit this PR with this new GT and also take care of the comments at the same time. MWGR is approaching .... 
Thank you .

tlampen:
Dear Gaëlle, my understanding is that  it should be 106X_upgrade2021_realistic_v8 . I hope @christopheralanwest can confirm to be sure.

boudoul:
Hi @tlampen , thanks. This sounds more like a MC GT and not a data GT, no ? Anyway I let @christopheralanwest comment when he can in these busy days .

christopheralanwest:
We currently don't have any better data GT for MWGR#1 than the candidate `106X_dataRun3_Express_Candidate_2019_05_24_21_36_45` that you have proposed. When MWGR#1 was taken, proper Run-3 HCAL conditions for data were not yet available. The candidate GT referenced here includes HCAL test tags with IOV from [1, inf) so that a test of the new HCAL conditions could be performed. But because of synchronization rules needed to ensure reproducibility, we cannot append new IOVs to a tag with an express synchronization with an IOV later than (last run taken at P5+1).

boudoul:
ok thanks @christopheralanwest - I believe I fixed in my last comment autocond according to your previous comments

christopheralanwest:
please test

prebello:
please test workflow 138.1

christopheralanwest:
The change of `auto:run3_data` to `auto:run3_data_promptlike` needs to be propagated to the other files in the commit.

The maintenance of this new relval will present some awkward issues as the proper conditions for this run cannot be provided when following the usual rules for handling prompt queues. There are several options:

- Generate new HCAL tags that introduce the HCAL conditions for the HB upgrade, setting the last IOV to some time prior to run 328788. This violates the usual requirement that the history in prompt queues be preserved. This is only an option because prompt reco has never been run on 2019 data. If we do not do this, it will not be possible to reconstruct MWGR#2 data except with this ad hoc candidate. On the other hand, it is likely have no one will ever look at MWGR#2 data after this month.
- Change the run number in the future so that it points to some run from MWGR#3 instead.
- Do not merge this PR but only use it for offline tests and reconsider it once MWGR#3 data is available.

Which is the preferred option? I'm also adding @abdoulline to this discussion to get an opinion from HCAL.

abdoulline:
Thank you for adding me to the discussion, Chris.
To me arranging RelVal for MWGR#3 would be the preferred option, if there is no particular importance(rush) of including MWGR#1 specifically...

fabiocos:
@boudoul as we are having several concurrent updates proposed for PyReleaseValidation , what is the plan for this one? From the thread I understand that there are issues that would suggest to keep this on hold until MWGR2

boudoul:
Hi @fabiocos ,  I failed to make understand the goal of this PR to the participants, which was actually to run in an IB the only data we have so far (we cannot invent new ones before the run) in order to test in IBs the workflow *before* a replay (Tier0 is not meant to validate a workflow, it should be tested first...) . But at this point I will anyway request a replay likely today with this run  and we will see . We will decide what do do later with this PR. 

So we can put it on hold (and anyhow I will need to fix the auto:run3_data_promptlike that I forgot to propagate) - Thanks

fabiocos:
hold

christopheralanwest:
I guess this PR could now be revisited using new data from the latest MWGR and the prompt GT 106X_dataRun3_Prompt_v2?

boudoul:
Hi @christopheralanwest , definitely - I will update with a fresh run and the GT (on my list for tomorrow..)

boudoul:
Hello @christopheralanwest - After a rebase, I have updated the GT , fixed the run3_data_promptlike , use a fresh run 330098 and updated title & doc - At your convenance , could you please look at it ? Thanks

christopheralanwest:
please test

christopheralanwest:
please test workflow 138.1

christopheralanwest:
+1

boudoul:
Thanks @christopheralanwest !
@fabiocos , can we unhold now ? and if the remaining signatures (@prebello, @zhenhu, @pgunnell, @kpedro88) could also take a look , that's would be great. Thanks.

kpedro88:
+upgrade

kpedro88:
unhold

boudoul:
Hi PdmV  (@prebello, @zhenhu, @pgunnell ), we need your comments here and eventually your signature

prebello:
+1

all looks ok. 138.1 tested and approved. sorry for the delay @boudoul. This github announcement skipped my daily check.

boudoul:
Thanks @prebello !

kpedro88:
+1

PR #27373, opened 2019-06-27
HGCal CLUE Algorithm - make the ordering reproducible in case the energy density is the same

ZihengChen:
#### PR description:

in the case where two or more cells are close-by and have the same energy density, we choose their density order based on their detid.

### DIFF
--- a/RecoLocalCalo/HGCalRecProducers/plugins/HGCalCLUEAlgo.cc
+++ b/RecoLocalCalo/HGCalRecProducers/plugins/HGCalCLUEAlgo.cc
@@ -271,8 +271,7 @@ void HGCalCLUEAlgo::calculateDistanceToHigher(const HGCalLayerTiles& lt, const u
           unsigned int otherId = lt[binId][j];
 
           float dist = distance(i, otherId, layerId);
-          bool foundHigher = cellsOnLayer.rho[otherId] > cellsOnLayer.rho[i];
-
+          bool foundHigher = (cellsOnLayer.rho[otherId] > cellsOnLayer.rho[i]) || (cellsOnLayer.rho[otherId] == cellsOnLayer.rho[i] && cellsOnLayer.detid[otherId] > cellsOnLayer.detid[i]);
           // if dist == i_delta, then last comer being the nearest higher
           if (foundHigher && dist <= i_delta) {
             // update i_delta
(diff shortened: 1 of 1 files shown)


ZihengChen:
@felicepantaleo

ZihengChen:
pr

Before three close-by cells with the same density are all promoted as seeds, splitting a single cluster into three smaller clusters. After this fix, only one of these three cells are promoted as seed. Thus the cluster is not split into three.

slava77:
@ZihengChen 
please follow https://github.com/cms-sw/cmssw/pull/27373#issuecomment-506284237
this PR can not be tested until this is addressed.

kpedro88:
please test

kpedro88:
+upgrade
very minor changes in layer clusters and downstream objects

slava77:
@ZihengChen  
was this presented in a DPG meeting already? If so, please add a link to the slides in the PR description.

@rovere @felicepantaleo Is this OK for integration from the DPG perspective?

RecoLocalCalo/HGCalRecProducers/plugins/HGCalCLUEAlgo.cc
```diff
@@ -271,8 +271,9 @@ void HGCalCLUEAlgo::calculateDistanceToHigher(const HGCalLayerTiles& lt, const u
           unsigned int otherId = lt[binId][j];
 
           float dist = distance(i, otherId, layerId);
-          bool foundHigher = cellsOnLayer.rho[otherId] > cellsOnLayer.rho[i];
-
+          bool foundHigher =
+              (cellsOnLayer.rho[otherId] > cellsOnLayer.rho[i]) ||
+              (cellsOnLayer.rho[otherId] == cellsOnLayer.rho[i] && cellsOnLayer.detid[otherId] > cellsOnLayer.detid[i]);
```
slava77: Is the case with equal rho supposed to be ordered by detid or is this an attempt to introduce a predictable order and a random choice is a more appropriate choice?
If a random choice is more preferable, perhaps some better effort to randomize the choice in a predictable way should be made.

felicepantaleo:
@slava77 yes, this was discussed during the DPG meeting as it fixes weird behaviors, mostly visible in the scintillator as there was a saturation problem.
Normally, the effect should be negligible.

RecoLocalCalo/HGCalRecProducers/plugins/HGCalCLUEAlgo.cc
felicepantaleo: @slava77 the order does not matter as soon as only one seed is created instead of 2 or 3. The effect is that these clusters will be merged in a single one.

slava77:
+1

for https://github.com/cms-sw/cmssw/pull/27373	ef390a1
- jenkins tests pass and comparisons with the baseline show differences only in the phase-2 workflows; starting from HGC clusters
    - e.g. in wf 29034.0 (D41), there are about 0.2% fewer hgcalLayerClusters and about 0.4% less multiclusters. This is consistent with the expected merging of clusters.

fabiocos:
+1

PR #27377, opened 2019-06-27
Fix ADC range for HGC scintillators

deguio:
The ADC range was fixed to [0, 68.5] MIPs which isn't enough for large energy deposits. the range is now extended to 275 MIPs. the current implementation of the readout electronics for the HGC scintillators is a simplistic one so this choice is arbitrary.

this PR also includes:
- updated conversion of the kev2mip factor --> transparent for the reco as rechits are autocalibrated
- subtraction of the pedestal mean at digi level --> this is following some discussion we had recently and it is relevant for aged scenario only

the PR was validated locally and passes:
`scram build code-checks`
`scram b runtests`
`runTheMatrix.py -l limited -i all --ibeos` (still running)

@rovere @jingyucms @pfs @cseez

### DIFF
--- a/SimCalorimetry/HGCalSimProducers/python/hgcalDigitizer_cfi.py
+++ b/SimCalorimetry/HGCalSimProducers/python/hgcalDigitizer_cfi.py
@@ -188,9 +188,9 @@ hgchebackDigitizer = cms.PSet(
             # 0 only ADC, 1 ADC with pulse shape, 2 ADC+TDC with pulse shape
             fwVersion       = cms.uint32(0),
             # n bits for the ADC (same as the silicon ROC)
-            adcNbits        = cms.uint32(10),
+            adcNbits        = cms.uint32(12),
             # ADC saturation : in this case we use the same variable but fC=MIP
-            adcSaturation_fC = cms.double(68.5), #value chosen to have 1MIP at 15ADC
+            adcSaturation_fC = cms.double(275.0), #value chosen to have 1MIP at 15ADC
             # threshold for digi production : in this case we use the same variable but fC=MIP
             adcThreshold_fC = cms.double(0.5),
             thresholdFollowsMIP = cms.bool(False)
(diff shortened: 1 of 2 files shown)


kpedro88:
please test

kpedro88:
Comparison differences are isolated to 2023 workflows and appear to be statistical, likely due to changes in random number sequences in the DIGI step.

The exception is the HEB RecHits, where there appears to be a systematic shift:
log10 HEB RecHits energy

Is this expected?

civanch:
+1

deguio:
> Is this expected?

hi @kpedro88 
I'm not sure yet. The difference you've highlighted seems to be the effect of updating the `keV2MIP` param, but I was expecting to see only minor changes at rechit level as the rechit energy should be automatically re-calibrated for scintillators according to [this](https://github.com/cms-sw/cmssw/blob/master/RecoLocalCalo/HGCalRecProducers/plugins/HGCalRecHitWorkerSimple.cc#L163). @rovere is this a correct assumption?

I've already updated the same parameter in a previous PR, but the value wasn't accurate.

To cross check this PR I've generated few events using a muon particle gun and when summing the PCaloHit energy deposited in each DetID the single MIP peak sits around 675 MeV:
simhits

when plotting the DIGI distribution in ADC the peak is now around 15 ADC which is what is expected (see red histo) compared to the peak obtained with the conversion at 500 MeV (black):
digi

kpedro88:
@deguio @rovere what is the status of this PR? (is the shift in the HEB RecHit energy spectrum expected?)

deguio:
ciao @kpedro88,
I've looked into more details and I've convinced myself that the shift is indeed expected.
I was wrongly assuming that for scintillators the recHit energy is automatically re-calibrated, but that doesn't seem to be the case because in the end the parameter `kev2mip` is never re-used at (uncalib)rechit level.

For scintillators the other factors that come into play are the [weights](https://github.com/cms-sw/cmssw/blob/CMSSW_11_0_X/RecoLocalCalo/HGCalRecProducers/python/HGCalRecHit_cfi.py#L129-L179) which should depend only on the material.

So in the end updating this value seems the right thing to do to restore the proper conversion from `keV` (from G4) to `MIPs`

kpedro88:
@deguio to make sure I understand, the conclusion is that your update `keV2MIP` fixes some behavior that was wrong before? (I want to make sure we know how to keep this value in sync with other changes...)

deguio:
hi @kpedro88 
yes. I've recently (PR #26811) updated the same `keV2MIP` value from 1/616 to 1/500 but I've made a mistake and the measurement wasn't correct. The new value I'm proposing is a more accurate measurement of the conversion from simHits energy to MIPs based on geometry v10.

kpedro88:
+upgrade

fabiocos:
+1

PR #27380, opened 2019-06-27
Fix SiPixelGenError Interpolation for Irradiation Bias Correction

mmusich:
#### PR description:

This PR is intended to fix a bug in the GenError object that is responsible for the poor performance of the generic reconstruction with Irradiation Bias Correction switched on. This is very relevant for the performance of the Pixel detector in Run3 studies.
The fix has been provided by M. Swartz and JHU group (@pmaksim1 @cmantill @OzAmram )

#### PR validation:

Enclosed are the plots illustrating the issue and the performance after the fix is applied. 
Results are shown for templates and generic reconstruction with and without the Irradiation Bias Correction and with the bug [left hand side] and with the bug fixed [right hand side]. As you can see, without the  bug fix the resolution in the central region is significantly degraded,

image

This PR is not a backport, but it should be ported to 10_6_X as well, to be used for the `Run3Summer19DR` production studies. 

cc:
@tvami @tsusa @leaca

### DIFF
--- a/CondFormats/SiPixelTransient/src/SiPixelGenError.cc
+++ b/CondFormats/SiPixelTransient/src/SiPixelGenError.cc
@@ -1,5 +1,5 @@
 //
-//  SiPixelGenError.cc  Version 2.20
+//  SiPixelGenError.cc  Version 2.30
 //
 //  Object stores Lorentz widths, bias corrections, and errors for the Generic Algorithm
 //
@@ -10,6 +10,7 @@
 //  V2.10 - Update the variable size [SI_PIXEL_TEMPLATE_USE_BOOST] option so that it works with VI's enhancements
 //  V2.20 - Add directory path selection to the ascii pushfile method
 //  V2.21 - Move templateStore to the heap, fix variable name in pushfile()
+//  V2.30 - Fix interpolation of IrradiationBias corrections
 
 //#include <stdlib.h>
 //#include <stdio.h>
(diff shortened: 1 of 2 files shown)


mmusich:
type bug-fix

srimanob:
@boudoul @fabiocos 
FYI

CondFormats/SiPixelTransient/src/SiPixelGenError.cc
```diff
@@ -754,6 +755,23 @@ int SiPixelGenError::qbin(int id,
   sy1 = (1.f - yratio) * thePixelTemp_[index].enty[ilow].syone + yratio * thePixelTemp_[index].enty[ihigh].syone;
   sy2 = (1.f - yratio) * thePixelTemp_[index].enty[ilow].sytwo + yratio * thePixelTemp_[index].enty[ihigh].sytwo;
 
+  if (irradiationCorrections) {
+    auto yavggen = (1.f - yratio) * thePixelTemp_[index].enty[ilow].yavggen[binq] +
```
slava77: a comment line may help to be more obvious why this code should be here rather than 40 lines below.
Unless I missed something, the code here is the same as in the old version, but it apparently depends on what's happening with the used values between here and the old place

mmusich: @slava77 the code is indeed the same. 
The problem lies in the fact that "the parameters used in the y-interpolation were changed prior to the x-interpolation.  The minimal change was to move the y-related quantities to before this happens." (quoting Morris here)

slava77:
@cmsbuild please test

CondFormats/SiPixelTransient/src/SiPixelGenError.cc
OzAmram: Staring at the code for a bit I think what is different between here and 40 lines below are the indices "ilow" and "ihigh" which designate which two template entries are being interpolated between. These are changed on what is now line 812. 

These are different between the x and y interpolation and the previous version had the irradiation bias correction for the y-interpolation use the indices for the x-interpolation.

ggovi:
+1

slava77:
Running locally on 1K events for the 2021/Run3 workflow (11624) shows very small changes in in pixel hit residuals in y direction, as somewhat expected from the code changes. This workflow is set up for the beginning of Run-3, which means that the irradiation effects are fairly small. 

Some examples of changes:
wf11624_BPIX1_res_y
wf11624_FPIXm1_res_y


@mmusich @tsusa 
please clarify which GT should be used to test the effects of larger irradiation effects.

mmusich:
@slava77 indeed the 2021 worfklow uses conditions for a mildly irradiated pixel detector.
All plots shown [here](https://indico.cern.ch/event/830302/contributions/3483706/attachments/1871334/3079441/tanjaReco_28062019.pdf) by @tsusa and myself use the simulation for end-of-2023: the corresponding Global Tag candidate used for the test is: `106X_mcRun3_2023_realistic_Candidate_2019_06_22_09_48_28`

slava77:
@mmusich 
thanks for the pointer.
BTW, for the comparisons of the fix in I was rerunning from the same step2 (DIGI) input in the baseline.
Do the changes in SiPixelGenError affect the DIGI step in any way ? (my guess is not, but I wanted to double-check).

mmusich:
> Do the changes in SiPixelGenError affect the DIGI step in any way ? (my guess is not, but I wanted to double-check).

@slava77 no, it should affect only the CPE code which is exercised in the RECO step.

slava77:
+1

for https://github.com/cms-sw/cmssw/pull/27380	6ac62d5
- code changes are in line with the PR description and the presentation in the RECO meeting on June 28 https://indico.cern.ch/event/830302/contributions/3483706/attachments/1871334/3079441/tanjaReco_28062019.pdf
- jenkins tests pass and comparisons with the baseline show small differences only in 2021 workflow (11624).
- local test with more events (1K) in the same wf 11624 confirms that the changes remain small (see https://github.com/cms-sw/cmssw/pull/27380#issuecomment-507156203)
- local tests using 2023 conditions show expected behavior with a significant improvement in BPIX1 y direction; this also leads to an improved tracking efficiency in the initialStep

wf11624_2023_BPIX1_resy
wf11624_2023_trk_initialHP_eff_eta

fabiocos:
+1

PR #27386, opened 2019-06-27
HGCAL TICL time compatibility for doublets building

amartelli:
Add time compatibility between layer clusters when building doublets 
- only apply if both inner and outer considered layer clusters have time information
- parameter for dT cut is configurable 

@rovere @felicepantaleo @Shameena01

### DIFF
--- a/RecoHGCal/TICL/plugins/HGCGraph.cc
+++ b/RecoHGCal/TICL/plugins/HGCGraph.cc
@@ -5,18 +5,21 @@
 #include "PatternRecognitionbyCA.h"
 #include "HGCDoublet.h"
 #include "HGCGraph.h"
+#include "DataFormats/Common/interface/ValueMap.h"
 
 void HGCGraph::makeAndConnectDoublets(const TICLLayerTiles &histo,
                                       int nEtaBins,
                                       int nPhiBins,
                                       const std::vector<reco::CaloCluster> &layerClusters,
                                       const std::vector<float> &mask,
+				      const edm::ValueMap<float> &layerClustersTime,
                                       int deltaIEta,
                                       int deltaIPhi,
                                       float minCosTheta,
                                       float minCosPointing,
                                       int missing_layers,
-                                      int maxNumberOfLayers) {
+                                      int maxNumberOfLayers,
+				      float maxDeltaTime) {
   isOuterClusterOfDoublets_.clear();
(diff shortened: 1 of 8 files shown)


slava77:
@cmsbuild please test

RecoHGCal/TICL/plugins/HGCGraph.cc
```diff
@@ -90,6 +96,21 @@ void HGCGraph::makeAndConnectDoublets(const TICLLayerTiles &histo,
...
 }
 
+bool HGCGraph::areTimeCompatible(int innerIdx,
+                                 int outerIdx,
+                                 const edm::ValueMap<float> &layerClustersTime,
+                                 float maxDeltaTime) {
+  float timeIn = layerClustersTime.get(innerIdx);
+  float timeOut = layerClustersTime.get(outerIdx);
+
+  if (timeIn == -99 || timeOut == -99)
```
kpedro88: this can be simplified:
```
return (timeIn == -99 || timeOut == -99 || std::abs(timeIn - timeOut) < maxDeltaTime));
```

kpedro88:
please test

RecoHGCal/TICL/plugins/HGCGraph.cc
```diff
@@ -53,6 +56,9 @@ void HGCGraph::makeAndConnectDoublets(const TICLLayerTiles &histo,
                     if (mask[innerClusterId] == 0.)
                       continue;
                     auto doubletId = allDoublets_.size();
+                    if (maxDeltaTime != -1 &&
```
slava77: Isn't it too deep to define a magic number?
Perhaps use ```<= 0``` to be less restrictive

amartelli: -1 was meant to be "don't check the time compatibility", to leave the option to switch this on or off and instead of adding a bool I used a negative value.
I can add a comment here, or add a bool instead.

slava77:
+1

for https://github.com/cms-sw/cmssw/pull/27386	8ddad78 
- jenkins tests pass and comparisons with the baseline show no differences as expected (the TICL code is not enabled in the standard workflows; pending a merge of #27392 )

kpedro88:
+upgrade

fabiocos:
+1

PR #27387, opened 2019-06-27
Run3-sim34 Update the codes which make more meaningful comparisons between G4 and GV

bsunanda:
### PR description:

Update the codes which make more meaningful comparisons between G4 and GV

#### PR validation:

Tested in the context where Geant4 results are compared with GeantV results

No back porting is needed. The code is kept for a futuristic use if we take a GeantV like approach in future.

### DIFF
--- a/SimG4CMS/Calo/macros/MakeGVPlots.C
+++ b/SimG4CMS/Calo/macros/MakeGVPlots.C
@@ -1,0 +1,255 @@
+#include <TCanvas.h>
+#include <TChain.h>
+#include <TFile.h>
+#include <TFitResult.h>
+#include <TFitResultPtr.h>
+#include <TGraphAsymmErrors.h>
+#include <TH1D.h>
+#include <TH2D.h>
+#include <TLegend.h>
+#include <TPaveStats.h>
+#include <TPaveText.h>
+#include <TProfile.h>
+#include <TROOT.h>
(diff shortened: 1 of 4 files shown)


bsunanda:
@cmsbuild Please test

civanch:
+1

fabiocos:
@bsunanda besides the macro update, you make a number of adjustment to the use of unit of measurements in CaloSteppingAction, keeping CLHEP but changing their use (from the Birk constants to the input parameters). Although this is essentially a standalone test class, I assume you have verified the result are the same. For my understanding what is the reason for this change?

bsunanda:
@fabiocos Geant4 and GeantV come with totally different set of units. We need to isolate the use of units in the specific code where Geant4 class is used. In case of Birk's law application the basic constants supplied as parameters require step size in cm, density in g/cc -> so I changed them in appropriate place. For Light collection efficiency or other geometry parameters the internal unit is in mm and for energy deposit the unit is in MeV till it is changed at the time of storage. This way I could match the units in Geant4 and GeantV. My latest comparisons are much more positive - most likely I can show that in the next simulation meeting

fabiocos:
@bsunanda thanks for the clarification, this PR will be integrated in next IB

fabiocos:
+1

PR #27382, opened 2019-06-27
Bug fixes in TICL

rovere:
#### PR description:

During the review process, 2 bugs were introduced. This PR fixes them.

#### PR validation:

runTheMatrix.py -l limited -i all -j 4 -t 12
(still running)

PR #27392, opened 2019-06-28
Closeby Workflows with TICL Variations in runTheMatrix

rovere:
#### PR description:

This PR does the following:

   * Fix the Closeby snippet to match the name with the actual particles generated
   * Add Closeby Workflow to the upgrade session of runTheMatrix
   * Add 2 variations to this workflow that will add TICL to the full Reco (.52) or run TICL reconstruction alone (.51)

This PR, in order to be fully tested, needs #27382.

Since runTheMatrix is a little outside of my area of expertise, please provide alternative solutions if you are not satisfied with this PR.

Finally, given the deadline of the 6th Patatrack hackathon next week, it would be nice to have an IB with this and #27382 integrated.

#### PR validation:

```
runTheMatrix.py -l limited -i all --ibeos
```

and

```
runTheMatrix.py -w upgrade -l 29093.51,29093.52
```

### DIFF
--- a/Configuration/PyReleaseValidation/python/relval_upgrade.py
+++ b/Configuration/PyReleaseValidation/python/relval_upgrade.py
@@ -89,6 +89,16 @@ for year in upgradeKeys:
                 elif '2018' in key:
                     workflows[numWF+upgradeSteps['pixelTrackingOnly']['offset']] = [ upgradeDatasetFromFragment[frag], stepList['pixelTrackingOnly']]
 
+            # special workflows for HGCAL/TICL
+            if (upgradeDatasetFromFragment[frag]=="CloseByParticleGun"):
+                TICLVariations = ['TICLOnly', 'TICLFullReco']
+                # Skip Hharvesting for TICLOnly
+                for tv in TICLVariations:
+                    if 'TICLOnly' in tv:
+                        stepList[tv] = [s for s in stepList[tv] if ("HARVEST" not in s)]
+                for tv in TICLVariations:
+                    workflows[numWF+upgradeSteps[tv]['offset']] = [ upgradeDatasetFromFragment[frag], stepList[tv]]
+
             # special workflows for HE
             if upgradeDatasetFromFragment[frag]=="TTbar_13" and '2018' in key:
                 workflows[numWF+upgradeSteps['heCollapse']['offset']] = [ upgradeDatasetFromFragment[frag], stepList['heCollapse']]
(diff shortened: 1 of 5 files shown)


Configuration/PyReleaseValidation/python/relval_upgrade.py
```diff
@@ -89,6 +89,16 @@ def makeStepName(key,frag,step,suffix):
                 elif '2018' in key:
                     workflows[numWF+upgradeSteps['pixelTrackingOnly']['offset']] = [ upgradeDatasetFromFragment[frag], stepList['pixelTrackingOnly']]
 
+            # special workflows for HGCAL/TICL
+            if (upgradeDatasetFromFragment[frag]=="CloseByParticleGun"):
```
kpedro88: can you restrict this to 2023 workflows? (We don't need to run TICL for Phase 1)

kpedro88:
@rovere if you want these workflows to be tested in IBs, you should add the WF numbers to relval_2023.py

Configuration/PyReleaseValidation/python/relval_upgrade.py
rovere: Sure!

rovere:
> @rovere if you want these workflows to be tested in IBs, you should add the WF numbers to relval_2023.py

Thanks, I just add 1 TICL workflow to the D41 section

kpedro88:
please test workflow 29093.52

rovere:
> please test workflow 29093.52

Without #27382 it will fail in creating the python configuration and will also fail in the assert.
That's why I explicitly required #27382.
I tested locally with #27382 on top and it works. Not sure how to proceed in this case.

kpedro88:
please abort

kpedro88:
please test workflow 29093.52 with #27382

kpedro88:
+upgrade

fabiocos:
@prebello, @zhenhu, @efeyazgan, @pgunnell, @agrohsje, @alberto-sanchez, @qliphy could you please check this PR?

prebello:
+1
@rovere are these new workflows aimed to be tested only in IB?

qliphy:
+1

rovere:
> +1
> @rovere are these new workflows aimed to be tested only in IB?

For the time being yes. The plan is to request some specific sample in the form of RelVal at some point in the (near) future.

kpedro88:
@fabiocos can we merge this now?

fabiocos:
+1

fwyzard:
@kpedro88 @fabiocos please do not use the modifiers .51, .52, .53 .

They are being used in the Patatrack branch (after we had to move from the .7 etc.) we were using before.

rovere:
Indeed, I overlooked the numbering. Will shift it in a coming PR (unless @kpedro88 or someone else wants to take care of that)

kpedro88:
@fwyzard maybe it would be useful for you to submit a PR updating upgradeWorkflowComponents.py with your usage of these numbers, even if they aren't applied to any actual workflows, so that we can keep track of reserved values more easily.

fabiocos:
@rovere @kpedro88 so what would be your alternative proposal for this PR?

kpedro88:
We just need to change the numbering to something like .101 and .102. This is not urgent, just a point for the longer term to avoid the conflict with patatrack.

PR #27383, opened 2019-06-27
Ecal Geometry: DD Dependency Cleanup v2

ianna:
#### PR description:

 * Move all plugins **ASIS** in a plugin directory to avoid direct dependency of the package on DD
 * Run code-checks

_Note_: the code needs a thorough revision: CLHEP usage, code modernization, etc. This will be done as a next step.

Replaces https://github.com/cms-sw/cmssw/pull/27376

#### if this PR is a backport please specify the original PR:

Before submitting your pull requests, make sure you followed this checklist:
- verify that the PR is really intended for the chosen branch
- verify that changes follow [CMS Naming, Coding, And Style Rules](http://cms-sw.github.io/cms_coding_rules.html)
- verify that the PR passes the basic test procedure suggested in the [CMSSW PR instructions](https://cms-sw.github.io/PRWorkflow.html)

PR #27395, opened 2019-06-28
Ecal Geometry: Algorithms Migration from CLHEP to GeantUnits

ianna:
#### PR description:

* Use GeantUnits instead of CLHEP

_Note_: This PR is based on https://github.com/cms-sw/cmssw/pull/27383

#### PR validation:

Screen Shot 2019-06-28 at 12 16 59
Screen Shot 2019-06-28 at 12 17 14
Screen Shot 2019-06-28 at 12 17 24
Screen Shot 2019-06-28 at 12 17 53
Screen Shot 2019-06-28 at 12 18 05
Screen Shot 2019-06-28 at 12 18 17
Screen Shot 2019-06-28 at 12 18 31
Screen Shot 2019-06-28 at 12 18 41
Screen Shot 2019-06-28 at 12 18 51


ianna:
please test

ianna:
please test

slava77:
> Reco comparison results: 12007 differences found in the comparisons

apparently only 2023 workflows are affected.
Is this because these are built from xml, while the others (run1/2) are from DB and the DB part is somehow unchanged?

ianna:
@slava77 - yes, the rest of the workflows use the DB geometry. It's strange that there are changes observed. It must be a change from a double/float to a long double.

ianna:
@slava77 - please, let me know if I shall I roll back the changes. Apparently, the initialisation using long double calculations takes a second longer (~17 sec)

ianna:
+1

According to validation plots the Ecal geometry description did not change.

slava77:
assign upgrade
because it looks like only the phase-2 workflows are affected

kpedro88:
@ianna two questions:
1. why is long double needed?
2. will these changes also appear in non-2023 workflows if/when the geometry tags in the database are updated?

ianna:
> @ianna two questions:
> 
>     1. why is long double needed?
Please see https://en.cppreference.com/w/cpp/language/user_literal:
```

Only the following parameter lists are allowed on literal operators:
 
--
( const char * ) | (1) |  
 
( unsigned long long int ) | (2) |  
 
( long double ) | (3) |  
 
( char ) | (4) |  
 
( wchar_t ) | (5) |  
 
( char8_t ) | (6) | (since C++20)
 
( char16_t ) | (7) |  
 
( char32_t ) | (8) |  
 
( const char * , std::size_t ) | (9) |  
 
( const wchar_t * , std::size_t ) | (10) |  
 
( const char8_t * , std::size_t ) | (11) | (since C++20)
 
( const char16_t * , std::size_t ) | (12) |  
 
( const char32_t * , std::size_t ) | (13) |  

```
> 
>     2. will these changes also appear in non-2023 workflows if/when the geometry tags in the database are updated?

yes, I'd think so.

kpedro88:
+upgrade

DataFormats/Math/interface/GeantUnits.h
```diff
@@ -32,6 +32,8 @@ namespace geant_units {
     constexpr long double operator"" _m(long double length) { return length * 1000.; }
     constexpr long double operator"" _cm3(long double length) { return length * 1._cm * 1._cm * 1._cm; }
     constexpr long double operator"" _m3(long double length) { return length * 1._m * 1._m * 1._m; }
+    constexpr long double operator"" _mm(unsigned long long int length) { return length * 1; }
+    constexpr long double operator"" _cm(unsigned long long int length) { return length * 10; }
```
cvuosalo: Shouldn't integer arithmetic be preserved in these cases?
```
constexpr unsigned long long int operator"" _mm(unsigned long long int length) { return length * 1; }
```

ianna: This is consistent with what we already have:
https://cmssdt.cern.ch/lxr/source/DataFormats/Math/interface/deltaPhi.h?v=CMSSW_11_0_X_2019-07-01-2300#0065

ianna:
@fabiocos - I think, this is ready to go in.

ianna:
@slava77 - I'll be needing the PR for a subsequent one. If you are happy to sign this one off, please, do.

fabiocos:
@ianna in the scrutiny of #27383 I had a comment about moving the .h of DDEcalEndcapTrap into the .cc which you told would be addressed in a subsequent PR https://github.com/cms-sw/cmssw/pull/27383#issuecomment-506865226 . I do not see this done here, neither in following PRs, could you please clarify?

Furthermore, how do we check the consistency of the new geometry with the previous one? The relvals clearly show differences, they might be minor negligible fluctuations, but we should have a way to compare those differences, right?

civanch:
@ianna , @fabiocos , @kpedro88, @cvuosalo , if we see the difference, normally we should open PPD campaign, which we do, when switch between Geant4 versions or switch compilers. However, here will have a simple technical change from one representation of const values to another in only one sub-detector. How many similar modifications can be foreseen? It is why for this type of PRs I would request full identity and not accept PR if the result is not identical. If we will agree with this change without full validation we will have problem with validation of other technical PRs, moreover, if each time when migrate units our results are changed a bit we come to very grey zone where any control will be lost.

ianna:
> @ianna in the scrutiny of #27383 I had a comment about moving the .h of DDEcalEndcapTrap into the .cc which you told would be addressed in a subsequent PR [#27383 (comment)](https://github.com/cms-sw/cmssw/pull/27383#issuecomment-506865226) . I do not see this done here, neither in following PRs, could you please clarify?

This is a helper call used by the plugins hence it was moved to plugins directory. Merging the header to its cc would not work.

> 
> Furthermore, how do we check the consistency of the new geometry with the previous one? The relvals clearly show differences, they might be minor negligible fluctuations, but we should have a way to compare those differences, right?

Please, see the histograms attached to the PR. IMHO, the geometry did not change.

@civanch - I appreciate your comment. I will roll back the units change for Ecal as non-essential for the  DD4hep migration. It's more work for me at this stage, but it's ok.

slava77:
@ianna 
it looks like you now have 3 coupled PRs
and also from the https://github.com/cms-sw/cmssw/pull/27395#issuecomment-508713042 it sounds like this one is going to be updated.
Perhaps it's more productive to leave only the most advanced PR and close the other two.

fabiocos:
@ianna @civanch @cvuosalo a PPD  campaign would statistically check the final effect of the change, here I am speaking of a more basic a priori check which we should anyway do to compare the old and new geometries. 

Concerning the comment of @slava77 I understand the use of the same branch adding more and more material in subsequent PRs, but I agree that this makes confusing to understand what is new and what is already considered in a previous one...

ianna:
> @ianna @civanch @cvuosalo a PPD campaign would statistically check the final effect of the change, here I am speaking of a more basic a priori check which we should anyway do to compare the old and new geometries.
> 
@fabiocos - what is wrong with the comparison of the Ecal crystal centers and corners which I've done before submitting this PR? The plots are attached to the PR description. Let's discuss it at the Sim meeting today.

> Concerning the comment of @slava77 I understand the use of the same branch adding more and more material in subsequent PRs, but I agree that this makes confusing to understand what is new and what is already considered in a previous one...

Yes, when it takes more then a week to converge on a single PR it creates more work for everyone.

ianna:
To be reopened later when there is decision to do an appropriate validation of technical changes.

PR #27407, opened 2019-06-29
Fixed missing link in DQM/SiPixelPhase1Common

Dr15Jones:
#### PR description:

Need to link with DataFormats/TrackerRecHit2D so UBSAN works.

#### PR validation:

Compiles and links using the CMSSW_11_0_UBSAN_X_2019-06-28-2300 IB.

### DIFF
--- a/DQM/SiPixelPhase1Common/BuildFile.xml
+++ b/DQM/SiPixelPhase1Common/BuildFile.xml
@@ -2,6 +2,7 @@
 <use   name="FWCore/PluginManager"/>
 <use   name="FWCore/ParameterSet"/>
 <use   name="DataFormats/SiPixelDetId"/>
+<use   name="DataFormats/TrackerRecHit2D"/>
 <use   name="DQMServices/Core"/>
 <use   name="Geometry/TrackerGeometryBuilder"/>
 <use   name="CalibTracker/Records"/>


Dr15Jones:
please test.

Dr15Jones:
please test

fabiocos:
+1

fabiocos:
merge

@jfernan2 @andrius-k the update looks technical and straightforward

schneiml:
@fabiocos @Dr15Jones  I am mildly surprised by this, the idea was that all the tracking-related things are isolated in `DQM/SiPixelPhase1Track`. But I guess the dependency does not really cause any harm and there is not much point trying to enforce the separation.

So, fine with me, even though things where planned to not need that.

Dr15Jones:
@schneiml I didn't add the dependency, it was their before:

https://github.com/cms-sw/cmssw/blob/master/DQM/SiPixelPhase1Common/interface/SiPixelCoordinates.h#L29

All I did was tell the linker about the dependency (which was needed to allow UBSAN to work). It only happened to work before because only inlined functions were being used by the code in SiPixelPhase1Common.

schneiml:
@Dr15Jones ah, now I also see where this is used. As said, I don't think there is any problem with this dependency (and we should totally have it if it is used), I was just surprised since the code was originally split with the idea of avoiding such dependencies.

jfernan2:
+1
just for the records...

PR #27421, opened 2019-07-02
[GCC9] fix warning about initializing base class in copy constructor

smuzaffar:
#### PR description:

This should fix the following GCC 9 compilation warning
```
DataFormats/GeometrySurface/interface/Surface.h:58:3: warning: base class 'class BasicReferenceCounted' should be explicitly initialized in the copy constructor [-Wextra]
    58 |   Surface(const Surface& iSurface)
```
#### PR validation:

Local compilation in gcc 9 Ibs shows that warnings is gone

### DIFF
--- a/DataFormats/GeometrySurface/interface/Surface.h
+++ b/DataFormats/GeometrySurface/interface/Surface.h
@@ -56,7 +56,10 @@ protected:
       : Base(pos, rot), theMediumProperties(mp), theBounds(bounds) {}
 
   Surface(const Surface& iSurface)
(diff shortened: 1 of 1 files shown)


smuzaffar:
please test

smuzaffar:
please test

VinInn:
what about the move constructor?

civanch:
+1

smuzaffar:
@VinInn , compiler was not complaining that is why I did not pay attention to that but you are right. Base class should be initialized for move constructior too.

smuzaffar:
please test

civanch:
+1

fabiocos:
+1

PR #27428, opened 2019-07-02
Fix jet energy fractions after jet resolution smearing

ahinzmann:
#### PR description:

As currently implemented, jet resolution smearing breaks the jet energy fraction variables of pat::Jets.
This line of code changes the four-vector of a pat::Jet, but not the correction factors stored in the pat::Jet:
https://github.com/cms-sw/cmssw/blob/CMSSW_8_0_25/PhysicsTools/PatUtils/interface/SmearedJetProducerT.h#L267
Jet energy fraction calculation relies on the fact that one can retried the raw jet energy with the stored correction factors:
https://github.com/cms-sw/cmssw/blob/master/DataFormats/PatCandidates/interface/Jet.h#L376

This fix to the scaleEnergy method adjusts the stored correction factors, to be able to retrieve the raw jet energy again.

Jet resolution smearing is run during MiniAOD production and also in this test:
https://github.com/cms-sw/cmssw/blob/master/PhysicsTools/PatAlgos/test/corMETFromMiniAOD.py

Small changes are expected in MET corrections and uncertainties related to jet resolution smearing in MiniAOD.

#### PR validation:

It was checked that jet energy fractions are the same before and after jet energy resolution smearing with this fix.

### DIFF
--- a/DataFormats/PatCandidates/interface/Jet.h
+++ b/DataFormats/PatCandidates/interface/Jet.h
@@ -174,7 +174,7 @@ namespace pat {
       return correctedJet(level, flavor, set).p4();
     };
     /// Scale energy and correspondingly adjust raw jec factors
-    void scaleEnergy(double fScale) override { scaleEnergy(fScale, "scaled"); };
+    void scaleEnergy(double fScale) override { scaleEnergy(fScale, "Unscaled"); };
     void scaleEnergy(double fScale, const std::string& level);
 
   private:
--- a/DataFormats/PatCandidates/interface/Jet.h
+++ b/DataFormats/PatCandidates/interface/Jet.h
@@ -174,7 +174,8 @@ namespace pat {
       return correctedJet(level, flavor, set).p4();
     };
     /// Scale energy and correspondingly adjust raw jec factors
-    void scaleEnergy(double fScale) override;
+    void scaleEnergy(double fScale) override { scaleEnergy(fScale, "scaled"); };
+    void scaleEnergy(double fScale, std::string level);
 
   private:
     /// index of the set of jec factors with given label; returns -1 if no set
--- a/DataFormats/PatCandidates/interface/Jet.h
+++ b/DataFormats/PatCandidates/interface/Jet.h
@@ -173,6 +173,8 @@ namespace pat {
                                     const unsigned int& set = 0) const {
       return correctedJet(level, flavor, set).p4();
     };
+    /// Scale energy and correspondingly adjust raw jec factors
+    void scaleEnergy(double fScale) override;
 
   private:
     /// index of the set of jec factors with given label; returns -1 if no set
--- a/DataFormats/PatCandidates/interface/Jet.h
+++ b/DataFormats/PatCandidates/interface/Jet.h
@@ -175,7 +175,7 @@ namespace pat {
     };
     /// Scale energy and correspondingly adjust raw jec factors
     void scaleEnergy(double fScale) override { scaleEnergy(fScale, "scaled"); };
-    void scaleEnergy(double fScale, std::string level);
+    void scaleEnergy(double fScale, const std::string& level);
 
   private:
     /// index of the set of jec factors with given label; returns -1 if no set
--- a/DataFormats/PatCandidates/interface/Jet.h
+++ b/DataFormats/PatCandidates/interface/Jet.h
@@ -174,7 +174,7 @@ namespace pat {
       return correctedJet(level, flavor, set).p4();
     };
     /// Scale energy and correspondingly adjust raw jec factors
-    void scaleEnergy(double fScale) override { scaleEnergy(fScale, "Unscaled"); };
+    void scaleEnergy(double fScale) override { scaleEnergy(fScale, "Unscaled"); }
     void scaleEnergy(double fScale, const std::string& level);
 
   private:
--- a/DataFormats/PatCandidates/interface/JetCorrFactors.h
+++ b/DataFormats/PatCandidates/interface/JetCorrFactors.h
@@ -53,7 +53,7 @@ namespace pat {
     // constructor by value
     JetCorrFactors(const std::string& label, const std::vector<CorrectionFactor>& jec);
     // add correction factor
-    void addFactor(const CorrectionFactor& factor);
+    void insertFactor(const unsigned int& position, const CorrectionFactor& factor);
 
     // instance label of the jet energy corrections set
     std::string jecSet() const { return label_; }
--- a/DataFormats/PatCandidates/interface/JetCorrFactors.h
+++ b/DataFormats/PatCandidates/interface/JetCorrFactors.h
@@ -35,8 +35,6 @@
 namespace pat {
 
   class JetCorrFactors {
-    friend class Jet;
-
   public:
     // jet energy correction factor. For flavor independent jet energy corrections the
     // std::vector<float> holds just a single entry. From the first flavor dependent entry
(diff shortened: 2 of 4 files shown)


ahinzmann:
In addition to the original problem, this solution also stores the smearing factor for bookkeeping in pat::Jets that contain JECs.

perrotta:
please test

ahinzmann:
@rappoccio @knash @clelange FYI

ahinzmann:
sorry for the hick-up. fixed and re-ran the tests.

slava77:
@cmsbuild please test

ahinzmann:
The ExpressionEvaluatorUnitTest seems to fail also without this PR.

DataFormats/PatCandidates/interface/Jet.h
```diff
@@ -173,6 +173,9 @@ namespace pat {
                                     const unsigned int& set = 0) const {
       return correctedJet(level, flavor, set).p4();
     };
+    /// Scale energy and correspondingly adjust raw jec factors
+    void scaleEnergy(double fScale) override { scaleEnergy(fScale, "scaled"); };
+    void scaleEnergy(double fScale, std::string level);
```
slava77: ```static const& std::string```

DataFormats/PatCandidates/interface/JetCorrFactors.h
```diff
@@ -35,6 +35,8 @@
 namespace pat {
 
   class JetCorrFactors {
+    friend class Jet;
```
slava77: why is this needed?

DataFormats/PatCandidates/interface/Jet.h
ahinzmann: const std::string& level -> yes
why static? This method modifies the object.

DataFormats/PatCandidates/interface/JetCorrFactors.h
ahinzmann: otherwise JecCorrFactors don't allow addition of new levels after creation

DataFormats/PatCandidates/interface/Jet.h
slava77: sorry, no ```static``` necessary

DataFormats/PatCandidates/interface/JetCorrFactors.h
slava77: it would be still better not to expose data members outside of the class.
Is there a significant problem to add a relevant modifier method?

slava77:
@cmsbuild please test

slava77:
> No changes are expected as it only affects workflows that first apply jet energy resolution smearing and then apply a jet ID (does not happen in MiniAOD/NanoAOD production itself).

This is inconsistent with 
>Reco comparison results: 290 differences found in the comparisons

This shows changes in all MC miniAOD outputs in MET corrections (TXYForT1Smear = 6, TXYForT01Smear = 7, Smear = 8) and uncertainties (JetResUpSmear = 16, JetResDownSmear = 17).
e.g. in wf 250202.181 
all_mini_OldVSNew_TTbar13TeVPUppmx2018wf250202p181c_patMETs_slimmedMETs__RECO_obj_0__corrections__8__dsumEt43
the smear correction becomes very close to 0.
Please check and confirm that this is expected.

A related question is why are these not monitored in the regular JME DQM plots?

ahinzmann:
Ok, looks like the Jet smearing for MET Type1 corrections, assumes that the "Uncorrected" level in a PAT jet is resolution smeared:
https://github.com/cms-sw/cmssw/blob/master/JetMETCorrections/Type1MET/interface/JetCleanerForType1METT.h#L152
https://github.com/cms-sw/cmssw/blob/bad8e6d3af19b4521011bcb9d26c97e4e8a8872a/PhysicsTools/PatUtils/plugins/PATCaloJetMETcorrInputProducer.cc#L28

On the other hand the jet ID variables assume that the 0th level is really Raw without any smearing:
https://github.com/cms-sw/cmssw/blob/master/DataFormats/PatCandidates/interface/Jet.h#L376

Thus adapted the smearing to this convention:
level = "Uncorrected" = no JEC applied. for smeared jets JER is applied.
level = 0 = no JEC and no JER applied.

We will follow up to have this monitored in DQM.

perrotta:
@cmsbuild please test

DataFormats/PatCandidates/interface/Jet.h
```diff
@@ -174,7 +174,7 @@ namespace pat {
       return correctedJet(level, flavor, set).p4();
     };
     /// Scale energy and correspondingly adjust raw jec factors
-    void scaleEnergy(double fScale) override { scaleEnergy(fScale, "scaled"); };
+    void scaleEnergy(double fScale) override { scaleEnergy(fScale, "Unscaled"); };
```
slava77: the trailing semi-colon is not needed

slava77:
> Reco comparison results: 61 differences found in the comparisons

Looking at the reco comparisons
- differences in MET corrections/uncertainties are still present in 2017 and 2018 pp MC; they appear in the same set of corrections 6,7,8 (TXYForT1Smear = 6, TXYForT01Smear = 7, Smear = 8) and uncertainties 16,17	(JetResUpSmear = 16, JetResDownSmear = 17)
- there are no more differences in MET corrections/uncertainties in earlier (before 2017) and Run-3 workflows
- there is a new difference in the data re-miniAOD workflow 136.7611 in one of the AK8 b-tag discriminant (apparently one of DeepAK8, but could be smth else)

slava77:
one difference from 136.7611
all_OldVSNew_RunJetHT2016EreMINIAODwf136p7611c_min2,max-2,patJets_slimmedJetsAK8__PAT_obj___pairDiscriVector__31__second

this looks like maybe some numerical precision level change

slava77:
> one difference from 136.7611

there is a high probability that this is a random change
I started collecting inputs in #27504

ahinzmann:
So it turns out this PR is in fact a bug fix for the MET corrections/uncertainties, since energy fraction variables of jet energy resolution smeared jets are used here:
https://github.com/cms-sw/cmssw/blob/bad8e6d3af19b4521011bcb9d26c97e4e8a8872a/JetMETCorrections/Type1MET/interface/PFJetMETcorrInputProducerT.h#L174

-> The observed changes are expected
Rerunning 10024.0 locally, I confirm that this line of code is where the difference comes from.

We should backport this, however it changes MiniAOD content my a nuisance. I'm creating a backport to 10_6. What's the best strategy, to have analyzers use this bug-fix at least in analysis code in 10_2?

perrotta:
@cmsbuild please test

slava77:
> What's the best strategy, to have analyzers use this bug-fix at least in analysis code in 10_2?

The best strategy will have to adhere to the no-change policy (the code running in production should not change physics results).
I'm not sure how this can be done in the context of coding in this PR.
For a backport there better be a config parameter to switch the behavior.

slava77:
> > What's the best strategy, to have analyzers use this bug-fix at least in analysis code in 10_2?
> 

Is this particularly important for 10_2_X?

slava77:
+1

for https://github.com/cms-sw/cmssw/pull/27428	56e606d
- code changes are in line with the PR description and the follow up review. Changes are expected only in miniAOD MC workflows in smeared MET uncertainties/corrections
- jenkins tests pass and comparisons with the baseline show differences only in 2017,2018 and phase-2 MC in the smeared MET uncertainties/corrections. The differences are typically only on single events (out of 10 tested, as in most of the workflows with MC); so, the other MCs are apparently just lucky to have no diff.

fabiocos:
@santocch could you please review this PR?

@ahinzmann for backports we need to be careful in not changing the existing miniAOD default behaviour, as @slava77 says

slava77:
> @ahinzmann for backports we need to be careful in not changing the existing miniAOD default behaviour, as @slava77 says

I think that for 10_6_X we should still be OK as is. Not for 10_2_X.

fabiocos:
+1

fabiocos:
merge

santocch:
+1

PR #27439, opened 2019-07-04
HGCAL TICL  seeded iteration 

amartelli:
This PR adds a (track) seeded iteration in the TICL 

* Add a SeedingProducer
- track seeded: propagate tracks to the first HGCAL layer and save entry point and direction
- global for self-seeded: 2 regions, one for each z-side

* Doublets finding: 
- for each seeding region, loop over the 2dcl in the tiles in the proper window
- save the seed information in the Doublet and only consider pairs with same seed
- doublet pointing criteria updated to point to the track entry direction (track-seeded) or 0,0,0 (self-seeded) 

* Propagate the information of the seed to the trackster saved in the event, the information is in the form of ProductID (collection) and int (index in the collection)

* Add iteration track-seeded in the python sequence

@felicepantaleo, @rovere


slava77:
@cmsbuild please test workflow 29093.52

slava77:
@amartelli 
the TICL workflow is broken.
Please take a look and check 29093.52 before updating this PR.
Thank you.

amartelli:
tested 29093.52 and solved the bug that broke step3. Thanks

slava77:
@cmsbuild please test workflow 29093.52

DataFormats/HGCalReco/interface/TICLSeedingRegion.h
```diff
@@ -0,0 +1,29 @@
...
+#ifndef RecoHGCal_HGCalReco_TICLSeedingRegion_h
+#define RecoHGCal_HGCalReco_TICLSeedingRegion_h
+
+#include "DataFormats/HGCalReco/interface/Common.h"
+#include "DataFormats/Math/interface/normalizedPhi.h"
+#include "DataFormats/GeometryVector/interface/GlobalPoint.h"
+#include "DataFormats/GeometryVector/interface/GlobalVector.h"
+#include "DataFormats/Provenance/interface/ProductID.h"
+
+namespace ticl {
```
kpedro88: I think we have decided that if a class has TICL in the name, it does not also need to be in the ticl namespace.

RecoHGCal/TICL/interface/SeedingRegionAlgoBase.h
```diff
@@ -0,0 +1,38 @@
...
+  class SeedingRegionAlgoBase {
+  public:
+    SeedingRegionAlgoBase(const edm::ParameterSet& conf, edm::ConsumesCollector& sumes)
+        : algo_verbosity_(conf.getParameter<int>("algo_verbosity")) {}
+    virtual ~SeedingRegionAlgoBase(){};
+
+    virtual void makeRegions(const edm::Event& ev,
+                             const edm::EventSetup& es,
+                             std::vector<ticl::TICLSeedingRegion>& result) = 0;
+    enum VerbosityLevel { None = 0, Basic, Advanced, Expert, Guru };
```
kpedro88: better to use enum class

RecoHGCal/TICL/plugins/SeedingRegionByTracks.cc
```diff
@@ -0,0 +1,74 @@
+// Author: Arabella Martelli, Felice Pantaleo, Marco Rovere - arabella.martelli@cern.ch, felice.pantaleo@cern.ch, marco.rovere@cern.ch
+// Date: 06/2019
+#include <algorithm>
+#include <set>
+#include <vector>
+
+#include "FWCore/MessageLogger/interface/MessageLogger.h"
+#include "SeedingRegionByTracks.h"
```
kpedro88: plugins should not have separate headers

RecoHGCal/TICL/plugins/SeedingRegionByTracks.cc
```diff
@@ -0,0 +1,74 @@
...
+}
+
+void SeedingRegionByTracks::buildFirstLayers() {
+  float zVal = hgcons_->waferZ(1, true);
+  std::pair<double, double> rMinMax = hgcons_->rangeR(zVal, true);
+
+  for (int iSide = 0; iSide < 2; ++iSide) {
+    float zSide = (iSide == 0) ? (-1. * zVal) : zVal;
+    firstDisk_[iSide] =
+        new GeomDet(Disk::build(Disk::PositionType(0, 0, zSide),
```
kpedro88: looks like a memory leak, please use smart pointers (if pointers are actually necessary here)

RecoHGCal/TICL/plugins/SeedingRegionByTracks.cc
```diff
@@ -0,0 +1,74 @@
...
+    const reco::Track &tk = (*tracks_h)[i];
+    if (!cutTk_((tk))) {
+      continue;
+    }
+
+    FreeTrajectoryState fts = trajectoryStateTransform::outerFreeState((tk), bFieldProd);
+    int iSide = int(tk.eta() > 0);
+    TrajectoryStateOnSurface tsos = prop.propagate(fts, firstDisk_[iSide]->surface());
+    if (tsos.isValid()) {
+      result.emplace_back(ticl::TICLSeedingRegion(
```
kpedro88: This is not a good use of `emplace_back`. Provide a constructor for the `TICLSeedingRegion` struct, then you can use `emplace_back` correctly (just providing the arguments).

RecoHGCal/TICL/plugins/SeedingRegionGlobal.cc
```diff
@@ -0,0 +1,28 @@
...
+SeedingRegionGlobal::~SeedingRegionGlobal(){};
+
+void SeedingRegionGlobal::makeRegions(const edm::Event &ev,
+                                      const edm::EventSetup &es,
+                                      std::vector<ticl::TICLSeedingRegion> &result) {
+  // for unseeded iterations create 2 global seeding regions
+  // one for each endcap
+  for (int i = 0; i < 2; ++i) {
+    result.emplace_back(
+        ticl::TICLSeedingRegion({GlobalPoint(0., 0., 0.), GlobalVector(0., 0., 0.), i, -1, edm::ProductID()}));
```
kpedro88: see above comment about use of `emplace_back`

RecoHGCal/TICL/plugins/SeedingRegionGlobal.cc
```diff
@@ -0,0 +1,28 @@
+// Author: Arabella Martelli, Felice Pantaleo, Marco Rovere - arabella.martelli@cern.ch, felice.pantaleo@cern.ch, marco.rovere@cern.ch
+// Date: 06/2019
+#include <algorithm>
+#include <set>
+#include <vector>
+
+#include "FWCore/MessageLogger/interface/MessageLogger.h"
+#include "SeedingRegionGlobal.h"
```
kpedro88: plugins should not have separate headers

RecoHGCal/TICL/plugins/TICLSeedingRegionProducer.cc
```diff
@@ -0,0 +1,72 @@
...
+  int algoId_;
+  std::string seedingId_;
+};
+DEFINE_FWK_MODULE(TICLSeedingRegionProducer);
+
+TICLSeedingRegionProducer::TICLSeedingRegionProducer(const edm::ParameterSet& ps)
+    : algoId_(ps.getParameter<int>("algoId")) {
+  auto sumes = consumesCollector();
+
+  switch (algoId_) {
```
kpedro88: eventually a factory should be used to construct algorithms by name on demand (can be addressed in a subsequent PR)

RecoHGCal/TICL/plugins/HGCGraph.cc
```diff
@@ -24,17 +25,39 @@ void HGCGraph::makeAndConnectDoublets(const TICLLayerTiles &histo,
...
+      auto firstLayerOnZSide = maxNumberOfLayers * zSide;
+      const auto &firstLayerHisto = histo[firstLayerOnZSide];
+
+      int entryEtaBin = firstLayerHisto.etaBin(r.origin.eta());
+      int entryPhiBin = firstLayerHisto.phiBin(r.origin.phi());
+      startEtaBin = std::max(entryEtaBin - deltaIEta, 0);
+      endEtaBin = std::min(entryEtaBin + deltaIEta, nEtaBins);
+      startPhiBin = entryPhiBin - deltaIPhi;
+      endPhiBin = entryPhiBin + deltaIPhi;
+    }
```
rovere: I'm not sure I follow the logic here. You open a window in [eta,phi] here around the impact point. But later on, at lines 65,66 and 70, we open it up again, once for each bin in the original [eta,phi] space. This will likely both kill the performance and will add too many clusters (unless the compatibility kills them all, in which case we only waste time).
I'd suggest a more conservative approach of having a single bin here (so +1 for both ends) and open the window later on.

amartelli: Hi @rovere, 
lines 57-59 define the outer loop. Previously it was by default over the full range 0-nEtaBins, 0-nPhiBins.
This remains the same for the global seeded region, and it is restricted to a window +/- 2 (as the following loop on the inner region, but could be changed) around the entry point on the first layer.
The window is needed to keep efficiency in case the first compatible 2dcl is few layers beyond the first, while the track is propagate only to the first HGCAL layer. +/-2 might be too much, but the only bin containing the entry position can also be too tight. If I understand your comment...
Maybe we should define dedicated deltaIPhi and deltaIPhi for the outer loop and make these configurable.
Do I understand correctly your point? thanks

amartelli:
Hi @kpedro88 thanks for the comments, 
I will try to implement the changes in the next days, but I'll be to a conference and not super efficient

RecoHGCal/TICL/plugins/HGCGraph.cc
rovere: @amartelli I understand the logic, but from preliminary time-measurement I did on TTbar+200PU, the time taken by this iteration is a little too much.
I'd suggest for the moment to restrict the region to the single bin and pay, in case, some penalty on the efficiency side. Unless you want to give a try with a fixed 3x3 tower.

RecoHGCal/TICL/plugins/HGCDoublet.cc
```diff
@@ -23,9 +25,12 @@ bool HGCDoublet::checkCompatibilityAndTag(std::vector<HGCDoublet> &allDoublets,
       xi[j] = otherDoublet.innerX();
       yi[j] = otherDoublet.innerY();
       zi[j] = otherDoublet.innerZ();
+      seedi[j] = otherDoublet.seedIndex();
     }
     for (int j = 0; j < vs; ++j) {
-      ok[j] = areAligned(xi[j], yi[j], zi[j], xo, yo, zo, minCosTheta, minCosPointing, debug);
+      if (seedi[j] != seedIndex_)
+        continue;
+      ok[j] = areAligned(xi[j], yi[j], zi[j], xo, yo, zo, minCosTheta, minCosPointing, refDir, debug);
```
slava77: apparently the "continue" above is leading to the static analyzer warning at https://cmssdt.cern.ch/SDT/jenkins-artifacts/pull-request-integration/PR-d5a760/1333/llvm-analysis/report-687371.html#EndPath
This looks real

RecoHGCal/TICL/plugins/SeedingRegionByTracks.cc
```diff
@@ -0,0 +1,74 @@
+// Author: Arabella Martelli, Felice Pantaleo, Marco Rovere - arabella.martelli@cern.ch, felice.pantaleo@cern.ch, marco.rovere@cern.ch
```
slava77: a line break

RecoHGCal/TICL/plugins/SeedingRegionByTracks.h
```diff
@@ -0,0 +1,54 @@
+// Author: Felice Pantaleo,Marco Rovere - felice.pantaleo@cern.ch, marco.rovere@cern.ch
+// Date: 09/2018
+
+#ifndef __RecoHGCal_TICL_PRbyCA_H__
+#define __RecoHGCal_TICL_PRbyCA_H__
```
slava77: Please use an include guard matching the full file name subsystem_package_fileName_h

Alternatively, this file can be removed in line with https://github.com/cms-sw/cmssw/pull/27439/files#r301672902

RecoHGCal/TICL/plugins/SeedingRegionByTracks.h
```diff
@@ -0,0 +1,54 @@
...
+                     std::vector<ticl::TICLSeedingRegion>& result) override;
+
+  private:
+    void buildFirstLayers();
+
+    edm::EDGetTokenT<reco::TrackCollection> tracks_token_;
+    std::once_flag initializeGeometry_;
+    const HGCalDDDConstants* hgcons_;
+    const StringCutObjectSelector<reco::Track> cutTk_;
+    std::string detectorName_ = "HGCalEESensitive";
```
slava77: with the current use case, this can be a static const

RecoHGCal/TICL/plugins/SeedingRegionByTracks.cc
slava77: Isn't there an actual usable GeomDet for this?
Do we need to make a fake one?

RecoHGCal/TICL/plugins/SeedingRegionByTracks.cc
```diff
@@ -0,0 +1,74 @@
...
+      continue;
+    }
+
+    FreeTrajectoryState fts = trajectoryStateTransform::outerFreeState((tk), bFieldProd);
+    int iSide = int(tk.eta() > 0);
+    TrajectoryStateOnSurface tsos = prop.propagate(fts, firstDisk_[iSide]->surface());
+    if (tsos.isValid()) {
+      result.emplace_back(ticl::TICLSeedingRegion(
+          {GlobalPoint(tsos.globalPosition()),
+           GlobalVector(tsos.globalMomentum().x(), tsos.globalMomentum().y(), tsos.globalMomentum().z()),
```
slava77: why are these explicit constructors needed?
tsos globalPosition is already a GlobalPoint and a globalMomentum is of type GlobalVector

RecoHGCal/TICL/plugins/SeedingRegionByTracks.cc
```diff
@@ -0,0 +1,74 @@
...
+#include "DataFormats/TrackReco/interface/Track.h"
+#include "TrackingTools/TrajectoryState/interface/TrajectoryStateTransform.h"
+
+using namespace ticl;
+
+SeedingRegionByTracks::SeedingRegionByTracks(const edm::ParameterSet &conf, edm::ConsumesCollector &sumes)
+    : SeedingRegionAlgoBase(conf, sumes),
+      cutTk_(conf.getParameter<std::string>("cutTk")),
+      propName_(conf.getParameter<std::string>("propagator")) {
+  tracks_token_ = sumes.consumes<reco::TrackCollection>(conf.getParameter<edm::InputTag>("tracks"));
```
slava77: this can be in the initializer list

RecoHGCal/TICL/plugins/SeedingRegionByTracks.h
```diff
@@ -0,0 +1,54 @@
...
+#include "FWCore/ParameterSet/interface/ParameterSetDescription.h"
+#include "DataFormats/TrackReco/interface/TrackFwd.h"
+#include "FWCore/Framework/interface/ConsumesCollector.h"
+#include "TrackingTools/GeomPropagators/interface/Propagator.h"
+#include "MagneticField/Engine/interface/MagneticField.h"
+#include "Geometry/HGCalCommonData/interface/HGCalDDDConstants.h"
+#include "TrackingTools/Records/interface/TrackingComponentsRecord.h"
+#include "Geometry/CommonDetUnit/interface/GeomDet.h"
+#include "DataFormats/GeometrySurface/interface/BoundDisk.h"
+#include "DataFormats/GeometrySurface/interface/SimpleDiskBounds.h"
```
slava77: the two includes above are not used

RecoHGCal/TICL/plugins/SeedingRegionByTracks.cc
```diff
@@ -0,0 +1,74 @@
...
+}
+
+SeedingRegionByTracks::~SeedingRegionByTracks(){};
+
+void SeedingRegionByTracks::makeRegions(const edm::Event &ev,
+                                        const edm::EventSetup &es,
+                                        std::vector<ticl::TICLSeedingRegion> &result) {
+  std::call_once(initializeGeometry_, [&]() {
```
slava77: this doesn't look right: other instances of SeedingRegionByTracks in other threads should have bfield_ etc initialized;  IOVs can change.
Use ESWatchers instead.

RecoHGCal/TICL/plugins/SeedingRegionByTracks.cc
```diff
@@ -0,0 +1,74 @@
...
+
+  edm::Handle<reco::TrackCollection> tracks_h;
+  ev.getByToken(tracks_token_, tracks_h);
+  edm::ProductID trkId = tracks_h.id();
+  auto bFieldProd = bfield_.product();
+  const Propagator &prop = (*propagator_);
+
+  int nTracks = tracks_h->size();
+  for (int i = 0; i < nTracks; ++i) {
+    const reco::Track &tk = (*tracks_h)[i];
```
slava77: for range loop with an index may be better

RecoHGCal/TICL/plugins/SeedingRegionGlobal.cc
```diff
@@ -0,0 +1,28 @@
+// Author: Arabella Martelli, Felice Pantaleo, Marco Rovere - arabella.martelli@cern.ch, felice.pantaleo@cern.ch, marco.rovere@cern.ch
```
slava77: add a linebreak, please

RecoHGCal/TICL/plugins/SeedingRegionGlobal.h
```diff
@@ -0,0 +1,43 @@
...
+  class SeedingRegionGlobal final : public SeedingRegionAlgoBase {
+  public:
+    SeedingRegionGlobal(const edm::ParameterSet& conf, edm::ConsumesCollector& sumes);
+    ~SeedingRegionGlobal() override;
+
+    void makeRegions(const edm::Event& ev,
+                     const edm::EventSetup& es,
+                     std::vector<ticl::TICLSeedingRegion>& result) override;
+
+  private:
```
slava77: not needed

RecoHGCal/TICL/plugins/SeedingRegionGlobal.h
```diff
@@ -0,0 +1,43 @@
...
+#include "DataFormats/TrackReco/interface/TrackFwd.h"
+#include "FWCore/Framework/interface/ConsumesCollector.h"
+#include "TrackingTools/GeomPropagators/interface/Propagator.h"
+#include "MagneticField/Engine/interface/MagneticField.h"
+#include "Geometry/HGCalCommonData/interface/HGCalDDDConstants.h"
+#include "TrackingTools/Records/interface/TrackingComponentsRecord.h"
+#include "Geometry/CommonDetUnit/interface/GeomDet.h"
+#include "DataFormats/GeometrySurface/interface/BoundDisk.h"
+#include "DataFormats/GeometrySurface/interface/SimpleDiskBounds.h"
+#include "CommonTools/Utils/interface/StringCutObjectSelector.h"
```
slava77: please cleanup the includes, most are not necessary

RecoHGCal/TICL/plugins/SeedingRegionGlobal.h
```diff
@@ -0,0 +1,43 @@
...
+
+class HGCGraph;
```
slava77: this is not needed/used

RecoHGCal/TICL/plugins/TICLSeedingRegionProducer.cc
```diff
@@ -0,0 +1,72 @@
...
+      myAlgo_ = std::make_unique<SeedingRegionGlobal>(ps, sumes);
+      break;
+    default:
+      break;
+  }
+  produces<std::vector<ticl::TICLSeedingRegion>>();
+}
+
+void TICLSeedingRegionProducer::fillDescriptions(edm::ConfigurationDescriptions& descriptions) {
+  // hgcalMultiClusters
```
slava77: what does this comment refer to ?

RecoHGCal/TICL/plugins/TICLSeedingRegionProducer.cc
```diff
@@ -0,0 +1,72 @@
...
+  edm::ParameterSetDescription desc;
+  desc.add<int>("algo_verbosity", 0);
+  desc.add<edm::InputTag>("tracks", edm::InputTag("generalTracks"));
+  desc.add<std::string>("cutTk",
+                        "1.48 < abs(eta) < 3.0 && pt > 2. && p > 1 && quality(\"highPurity\") && "
+                        "hitPattern().numberOfLostHits(\"MISSING_OUTER_HITS\") < 10");
+  desc.add<std::string>("propagator", "PropagatorWithMaterial");
+  desc.add<int>("algoId", 1);
+  descriptions.add("seedingRegionProducer", desc);
```
slava77: the default name should match the class name (modulo the capitalization rules) => ```ticlSeedingRegionProducer```

RecoHGCal/TICL/python/ticl_iterations.py
```diff
@@ -22,6 +22,39 @@ def TICL_iterations_withReco(process):
 
   process.TICLLayerTileProducer = ticlLayerTileProducer.clone()
 
+  process.SeedingTrk = seedingRegionProducer.clone(
```
slava77: module instances should start with lower case (this comment applies to everything in this file)

RecoHGCal/TICL/interface/SeedingRegionAlgoBase.h
amartelli: Hi @kpedro88 
I see the same enum VerbosityLevel { None = 0, Basic, Advanced, Expert, Guru }; 
syntax in RecoHGCal/TICL/interface/PatternRecognitionAlgoBase.h
why should it be different for  SeedingRegionAlgoBase.h? Or can you clarify better what you mean? thanks

RecoHGCal/TICL/plugins/SeedingRegionByTracks.cc
amartelli: Also in this case I see the same logic applied to RecoHGCal/TICL/plugins/PatternRecognitionbyCA.cc and .h 
The plugin/producer is RecoHGCal/TICL/plugins/TICLSeedingRegionProducer.cc and for this there is no .h

and same for RecoHGCal/TICL/plugins/SeedingRegionGlobal.cc

RecoHGCal/TICL/interface/SeedingRegionAlgoBase.h
kpedro88: I should have made the same comment on `PatternRecognitionAlgoBase`, but must have missed it.

It would be even better not to duplicate this enum and just reference it from a central header (or something).

RecoHGCal/TICL/plugins/SeedingRegionByTracks.cc
slava77: you are not using a factory pattern. So, the header file is needed after all.

DataFormats/HGCalReco/interface/TICLSeedingRegion.h
amartelli: ok

RecoHGCal/TICL/plugins/SeedingRegionGlobal.cc
amartelli: ok

slava77:
@amartelli 
please clarify on the status/plan to update this PR.
Thank you.

amartelli:
@slava77 
I will followup this coming week. thanks

RecoHGCal/TICL/interface/SeedingRegionAlgoBase.h
amartelli: Hi @kpedro88 
I looked for enum class, but then the use of Verbosity level as int like in https://github.com/cms-sw/cmssw/blob/f30411b0af6599cb68961bccbc04b9b42f0f9d4c/RecoHGCal/TICL/plugins/PatternRecognitionbyCA.cc#L36
is no more allowed.
To keep this nice feature, do you have any suggestion?
also, @rovere

RecoHGCal/TICL/plugins/SeedingRegionByTracks.cc
amartelli: @kpedro88  replaced with smart pointer

@slava77 it seems there is no geomdet for the HGCAL layers. All this is built with the correct parameters from the geometry, so to get the surface() from the GeomDet that is needed to propagate the track. Also add @bsunanda if he has something more clever in mind. 

thanks

RecoHGCal/TICL/plugins/SeedingRegionByTracks.cc
amartelli: yes ok

RecoHGCal/TICL/interface/SeedingRegionAlgoBase.h
kpedro88: The type of the `algo_verbosity_` variable should be changed to the enum class type, rather than int. This is actually a benefit, because it prevents improper comparisons with other int variables whose values do not actually come from the enum.

DataFormats/HGCalReco/interface/TICLSeedingRegion.h
```diff
@@ -0,0 +1,29 @@
+// Authors: Felice Pantaleo, Marco Rovere
+// Emails: felice.pantaleo@cern.ch, marco.rovere@cern.ch
+// Date: 06/2019
+
+#ifndef RecoHGCal_HGCalReco_TICLSeedingRegion_h
+#define RecoHGCal_HGCalReco_TICLSeedingRegion_h
```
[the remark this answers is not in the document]
amartelli: yes, thanks

RecoHGCal/TICL/interface/SeedingRegionAlgoBase.h
rovere: I do not see a huge added value, especially in this debugging only case.
@amartelli you could implement the comparison operator, templated, to accommodate comparison among enums and ints. something along the lines:

```c++
template<typename T>
bool operator>(const VerbosityLevel  lhs, const T  rhs) {return static_cast<int>(lhs) > static_cast<int>(rhs);}
```

which will somehow defeat the whole purpose :) 
Use the non-templated version, i.e. using `VerbosityLevel` on both arguments explicitly to prevent compilation against int-comparison.

RecoHGCal/TICL/plugins/HGCGraph.cc
amartelli: changed to +/-1 by default for all iterations at the moment.

DataFormats/HGCalReco/src/classes_def.xml
```diff
@@ -14,4 +14,11 @@
     <field name="tiles_" transient="true"/>
   </class>
   <class name="edm::Wrapper<TICLLayerTiles> " />
+
+  <class name="ticl::TICLSeedingRegion" ClassVersion="5">
+    <version ClassVersion="5" checksum="1952737206"/>
```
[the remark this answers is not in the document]
amartelli: ok done

RecoHGCal/TICL/plugins/HGCDoublet.cc
```diff
@@ -81,12 +87,18 @@ int HGCDoublet::areAligned(double xi,
...
+  // or the direction wrt (0,0,0)
   // The compatibility is checked only for the innermost doublets: the
   // one with the outer doublets comes in by the alignment requirement of
   // the doublets themeselves
-  auto dot_pointing = dx2 * xi + dy2 * yi + dz2 * zi;
-  auto mag_pointing = std::sqrt(xi * xi + yi * yi + zi * zi);
+
+  const GlobalVector firstDoublet(dx2, dy2, dz2);
+  const GlobalVector pointingDir =
+      (seedIndex_ == -1) ? (GlobalPoint(innerX(), innerY(), innerZ()) - GlobalPoint(0, 0, 0)) : refDir;
```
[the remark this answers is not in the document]
amartelli: ok

RecoHGCal/TICL/plugins/HGCDoublet.h
```diff
@@ -56,8 +61,12 @@ class HGCDoublet {
 
   void tagAsInnerNeighbor(unsigned int otherDoublet) { innerNeighbors_.push_back(otherDoublet); }
 
-  bool checkCompatibilityAndTag(
-      std::vector<HGCDoublet> &, const std::vector<int> &, float, float minCosPointing = 1., bool debug = false);
+  bool checkCompatibilityAndTag(std::vector<HGCDoublet> &,
+                                const std::vector<int> &,
+                                const GlobalVector &refDir,
+                                float,
```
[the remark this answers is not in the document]
amartelli: ok

RecoHGCal/TICL/plugins/HGCDoublet.h
```diff
@@ -67,9 +76,10 @@ class HGCDoublet {
                  double zo,
                  float minCosTheta,
                  float minCosPointing,
+                 const GlobalVector &refDir,
                  bool debug = false) const;
 
-  void findNtuplets(std::vector<HGCDoublet> &, HGCntuplet &);
+  void findNtuplets(std::vector<HGCDoublet> &, HGCntuplet &, int);
```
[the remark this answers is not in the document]
amartelli: ok

RecoHGCal/TICL/plugins/HGCDoublet.cc
amartelli: yes thanks. bug fixed

RecoHGCal/TICL/plugins/SeedingRegionByTracks.cc
amartelli: ok

RecoHGCal/TICL/plugins/SeedingRegionByTracks.h
amartelli: ok  included guard matching the full file name

RecoHGCal/TICL/plugins/SeedingRegionByTracks.h
amartelli: ok

RecoHGCal/TICL/plugins/SeedingRegionByTracks.cc
amartelli: ok thanks

RecoHGCal/TICL/plugins/SeedingRegionByTracks.cc
amartelli: ok

RecoHGCal/TICL/plugins/SeedingRegionByTracks.h
amartelli: ok

RecoHGCal/TICL/plugins/SeedingRegionByTracks.cc
amartelli: replaced with beginRun

RecoHGCal/TICL/plugins/SeedingRegionByTracks.cc
amartelli: this I don't understand. the "i" is used afterwards so a counter is anyway needed...

RecoHGCal/TICL/plugins/SeedingRegionGlobal.cc
amartelli: ok

RecoHGCal/TICL/plugins/SeedingRegionGlobal.h
amartelli: ok

RecoHGCal/TICL/plugins/SeedingRegionGlobal.h
amartelli: ok

RecoHGCal/TICL/plugins/SeedingRegionGlobal.h
amartelli: ok

RecoHGCal/TICL/plugins/TICLSeedingRegionProducer.cc
amartelli: ok

RecoHGCal/TICL/plugins/TICLSeedingRegionProducer.cc
amartelli: ok

RecoHGCal/TICL/python/ticl_iterations.py
amartelli: ok done

amartelli:
Hi @slava77 @kpedro88 @rovere 
I implemented most of the comments (and thanks for the checks)
missing parts: 
- (enum class) https://github.com/cms-sw/cmssw/pull/27439#discussion_r301670752
- (factory instead of switch) https://github.com/cms-sw/cmssw/pull/27439#discussion_r301675994
- (struct to preserve the interface) https://github.com/cms-sw/cmssw/pull/27439#discussion_r302734128

I won't be able to focus on these soon.

RecoHGCal/TICL/plugins/SeedingRegionByTracks.cc
slava77: an index can be followed inside as well; not a strong request anyways

slava77:
@cmsbuild please test

slava77:
@cmsbuild please test workflow 29093.52

amartelli:
https://github.com/cms-sw/cmssw/pull/27439#issuecomment-514247223
it seems that these failings are not related to the PR

RecoHGCal/TICL/interface/PatternRecognitionAlgoBase.h
```diff
@@ -31,7 +32,9 @@ namespace ticl {
                                 const std::vector<float>& mask,
                                 const edm::ValueMap<float>& layerClustersTime,
                                 const TICLLayerTiles& tiles,
+                                const std::vector<ticl::TICLSeedingRegion>& regions,
```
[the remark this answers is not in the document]
amartelli: @slava77  can this be implemented for the next change in the interface?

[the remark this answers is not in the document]
slava77: Please clarify what is the limiting factor to do it now.
This looks like a matter of O(1hr) of work with a benefit of not having to touch many files the next time we update the code in the release.

kpedro88:
please test workflow 29093.52

RecoHGCal/TICL/plugins/TICLSeedingRegionProducer.cc
```diff
@@ -0,0 +1,75 @@
...
+  }
+  produces<std::vector<TICLSeedingRegion>>();
+}
+
+void TICLSeedingRegionProducer::fillDescriptions(edm::ConfigurationDescriptions& descriptions) {
+  edm::ParameterSetDescription desc;
+  desc.add<int>("algo_verbosity", 0);
+  desc.add<edm::InputTag>("tracks", edm::InputTag("generalTracks"));
+  desc.add<std::string>("cutTk",
+                        "1.48 < abs(eta) < 3.0 && pt > 2. && p > 1 && quality(\"highPurity\") && "
```
slava77: as written, ```p > 1``` does not select anything (but wastes CPU). Please remove or fix if some other value larger than pt was intended.

slava77:
> please test workflow 29093.52

this was supposed to be 20493.52

slava77:
please test workflow 20493.52

RecoHGCal/TICL/plugins/SeedingRegionByTracks.h
amartelli: ok, moved to .cc 
and keep only BoundDisk.h, needed to define Disk and it already includes SimpleDiskBounds.h

RecoHGCal/TICL/python/ticl_iterations.py
```diff
@@ -10,155 +10,155 @@
...
 
-  process.TrackstersMIP = trackstersProducer.clone(
-      filtered_mask = cms.InputTag("FilteredLayerClustersMIP", "MIP"),
-      seeding_regions = cms.InputTag("SeedingGlobal"),    
+  process.trackstersMIP = trackstersProducer.clone(
+      filtered_mask = cms.InputTag("filteredLayerClustersMIP", "MIP"),
+      seeding_regions = cms.InputTag("seedingGlobal"),
       missing_layers = 3,
       min_clusters_per_ntuplet = 15,
       min_cos_theta = 0.985 # ~10 degrees
```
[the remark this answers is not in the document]
amartelli: probably all parameters will be optimised again, but no reason to have these different by default. thanks

kpedro88:
please test workflow 20493.52

RecoHGCal/TICL/interface/PatternRecognitionAlgoBase.h
```diff
@@ -26,14 +26,28 @@ namespace ticl {
...
-                                std::vector<Trackster>& result) = 0;
+    struct Inputs {
+      const edm::Event& ev;
+      const edm::EventSetup& es;
+      const std::vector<reco::CaloCluster>& layerClusters;
+      const std::vector<float>& mask;
+      const edm::ValueMap<float>& layerClustersTime;
+      const TICLLayerTiles& tiles;
+      const std::vector<TICLSeedingRegion>& regions;
+      std::vector<Trackster>& result;
```
slava77: this is an output.

My request to aggregate was trying to address the fluid/growing part of the interface which is also apparently necessary only for a fraction of the derived algorithms.
"results" are somewhat distinct and non-const and should be used by every implementation. Better keep them separate.

BTW, is the ```result``` supposed to be empty on entry into ```makeTracksters(``` , if so, why not return by value?

RecoHGCal/TICL/interface/PatternRecognitionAlgoBase.h
```diff
@@ -26,14 +26,28 @@ namespace ticl {
...
+             const std::vector<reco::CaloCluster>& lC,
+             const std::vector<float>& mS,
+             const edm::ValueMap<float>& lT,
+             const TICLLayerTiles& tL,
+             const std::vector<TICLSeedingRegion>& rG,
+             std::vector<Trackster>& rS)
+          : ev(eV), es(eS), layerClusters(lC), mask(mS), layerClustersTime(lT), tiles(tL), regions(rG), result(rS) {}
+    };
+
+    virtual void makeTracksters(const Inputs& input) = 0;
```
slava77: (since we are touching the interface anyways)
is there a reason to keep ```makeTracksters``` non-const? If some thread-safety can be imposed by the API, it should be worthwhile.

RecoHGCal/TICL/plugins/SeedingRegionByTracks.cc
```diff
@@ -20,7 +21,7 @@ SeedingRegionByTracks::SeedingRegionByTracks(const edm::ParameterSet &conf, edm:
 
 SeedingRegionByTracks::~SeedingRegionByTracks(){};
```
slava77: the last semicolon is unnecessary

RecoHGCal/TICL/python/ticl_iterations.py
```diff
@@ -10,155 +10,155 @@
...
+      process.filteredLayerClustersMIP,
+      process.trackstersMIP,
+      process.multiClustersFromTrackstersMIP,
+      process.filteredLayerClusters,
+      process.tracksters,
+      process.multiClustersFromTracksters)
   process.schedule.associate(process.TICL_Task)
   return process
 
 def TICL_iterations(process):
```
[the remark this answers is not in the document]
slava77: > please add a comment to denote the purpose of this method.

this would help to clarify the purpose of the method.
Thanks for adding a note for the first method

slava77 (commented):
about ticl_iterations.py and module instance names there,
the naming is still rather nondescript in the context of running in the full reco processing job.
It should be relatively clear that all modules in the sequence are either more generically from HGCAL or at least more specifically from the TICL framework.

I recognize that this is a bit of a prototyping stage and some intermediate PRs may have some slack. 
Still, by the time defs from this .py file migrate to proper places that define modules/tasks for production environment, they should be all named clearly so that they are easily recognizable during execution and/or by their product module names.

RecoHGCal/TICL/python/ticl_iterations.py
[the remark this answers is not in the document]
slava77: >  It seems more practical to merge the two.

is there a good reason to keep this separated?
From https://github.com/cms-sw/cmssw/pull/27439#discussion_r310090538 I somehow (mis?)read that the rest can be the same as well

RecoHGCal/TICL/python/ticl_iterations.py
```diff
@@ -10,109 +10,153 @@
...
+  process.filteredLayerClustersTrk = filteredLayerClustersProducer.clone(
+    clusterFilter = "ClusterFilterByAlgo",
+    algo_number = 8,
+    iteration_label = "Trk"
+  )
 
-  process.TICLLayerTileProducer = ticlLayerTileProducer.clone()
+  process.trackstersTrk = trackstersProducer.clone(
+    filtered_mask = cms.InputTag("filteredLayerClustersTrk", "Trk"),
+    seeding_regions = cms.InputTag("ticlSeedingTrk"),
```
slava77: please drop type specifications for all parameters which already exist.

slava77:
@amartelli 
please note that there is a conflict now in the /ticl_iterations.py file
Please resolve this

slava77:
> @amartelli
> please note that there is a conflict now in the /ticl_iterations.py file
> Please resolve this

just a reminder

slava77:
please test workflow 20493.52

RecoHGCal/TICL/interface/PatternRecognitionAlgoBase.h
```diff
@@ -34,20 +34,18 @@ namespace ticl {
...
              const TICLLayerTiles& tL,
-             const std::vector<TICLSeedingRegion>& rG,
-             std::vector<Trackster>& rS)
-          : ev(eV), es(eS), layerClusters(lC), mask(mS), layerClustersTime(lT), tiles(tL), regions(rG), result(rS) {}
+             const std::vector<TICLSeedingRegion>& rG)
+          : ev(eV), es(eS), layerClusters(lC), mask(mS), layerClustersTime(lT), tiles(tL), regions(rG) {}
     };
 
-    virtual void makeTracksters(const Inputs& input) = 0;
+    virtual const void makeTracksters(const Inputs& input, std::vector<Trackster>& result) = 0;
```
slava77: I meant the method qualifier, not the return value qualifier.
```virtual void makeTracksters(const Inputs& input) const = 0;```

RecoHGCal/TICL/python/ticl_iterations.py
slava77: > please drop type specifications for all parameters which already exist.
> This is a safer way to protect from parameter name mistakes and will also help in possible parameter migrations.

this was not addressed yet

amartelli: Sorry for not commenting before, I don't understand what should be dropped here.
Removing "seeding_regions = cms.InputTag("ticlSeedingTrk")," produces the following exception...
Exception Message:
Looking for type: std::vector<TICLSeedingRegion>
Looking for module label: ticlSeedingRegionProducer

RecoHGCal/TICL/interface/PatternRecognitionAlgoBase.h
amartelli: with this change (const as method qualifier, not return value qualifier), it arise the compilation error 
RecoHGCal/TICL/plugins/PatternRecognitionbyCA.cc:26:34:  error: passing 'const hgcal::RecHitTools' as 'this' argument discards qualifiers [-fpermissive]
   rhtools_.getEventSetup(input.es);

Is there an easy fix, or should I remove the const?

RecoHGCal/TICL/python/ticl_iterations.py
slava77: ```cms.InputTag``` is a type specification, I was asking to drop it. => ```seeding_regions = "ticlSeedingTrk"```

RecoHGCal/TICL/interface/PatternRecognitionAlgoBase.h
slava77: I see. So, it looks like at least for some of the derived classes the methods are not reentrant
and we can not make it const (or thread safe).
It seems fine now to just revert this attempt to add a const.

RecoHGCal/TICL/python/ticl_iterations.py
amartelli: ok, thanks for the clarification

RecoHGCal/TICL/interface/PatternRecognitionAlgoBase.h
amartelli: ok!

felicepantaleo:
please test workflow 20493.52

felicepantaleo:
is there anything else to be addressed in this PR?
Once this is merged we can submit the TICL-PF interface.

PR #26399, opened 2019-04-09
Import additions for UL from CMSSW_9_4_13_UL1 (#26168) and #26211

fabiocos:
#### PR description:

This PR is meant to import into the main CMSSW_9_4_X branch the additions for UL tested separately in the test build CMSSW_9_4_13_UL1 (based on #26168) and on the further addition #26211 .

#### PR validation:

This code has been used in tests and running without issues in CMSSW_9_4_MAOD_X IBs for a while.

PR #25171, opened 2018-11-09
backported StoredMergeableRunProductMetadata

Dr15Jones:
In order to attempt to make CMSSW_10_X files forward compatible with
CMSSW_8_0 (for possible use in the ultra legacy processing) ROOT
needs the dictionary for the type StoredMergeableRunProductMetadata.

PR #27446, opened 2019-07-05
import additions for UL-HLT in 10_2_X

srimanob:
#### PR description:

Preparation for the UL-HLT in 10_2_X. The content of this PR is based on https://github.com/cms-sw/cmssw/pull/26399 (except no need of backport of CSCDigi) and https://github.com/cms-sw/cmssw/pull/25171

#### PR validation:

This PR is tested by producing GEN-SIM-RAW using 10_6, then pass to 10_2 for the HLT step. The RAW is re-written from 10_2 and read again in 10_6 for RECO and DQM.

GEN-SIM; 10_6_1_patch1
`cmsDriver.py ZEE_13TeV_TuneCUETP8M1_cfi --conditions auto:phase1_2018_realistic -n 10 --era Run2_2018 --eventcontent RAWSIM --relval 9000,50 -s GEN,SIM --datatier GEN-SIM --beamspot Realistic25ns13TeVEarly2018Collision --geometry DB:Extended --io ProdZEE_13UP18.io --python ProdZEE_13UP18.py --no_exec --fileout file:step1.root --nThreads 4`

DIGI; 10_6_1_patch1
`cmsDriver.py step2 --datamix PreMix --conditions auto:phase1_2018_realistic --pileup_input das:/RelValPREMIXUP18_PU25/CMSSW_10_6_0-PU25ns_106X_upgrade2018_realistic_v4-v1/PREMIX --era Run2_2018 --eventcontent FEVTDEBUGHLT --procModifiers premix_stage2 -s DIGI:pdigi_valid,DATAMIX,L1,DIGI2RAW --datatier GEN-SIM-DIGI-RAW-HLTDEBUG --io DIGIPRMXUP18_PU25.io --python DIGIPRMXUP18_PU25.py -n -1 --no_exec --filein file:step1.root --fileout file:step2.root --nThreads 4`

HLT; CMSSW_10_2_15_patch2 with contents of this PR
`cmsDriver.py HLT --conditions auto:phase1_2018_realistic --era Run2_2018 --eventcontent FEVTDEBUGHLT -s HLT:@relval2018 --datatier GEN-SIM-DIGI-RAW-HLTDEBUG --io step2a_HLT.io --python step2a_HLT.py -n -1 --no_exec --filein filelist:step2.root --fileout file:step2a.root --nThreads 4`

RECO; 10_6_1_patch1
`cmsDriver.py step3 --conditions auto:phase1_2018_realistic -n 10 --era Run2_2018 --eventcontent RECOSIM,MINIAODSIM,DQM --runUnscheduled --procModifiers premix_stage2 -s RAW2DIGI,L1Reco,RECO,RECOSIM,EI,PAT,VALIDATION:@standardValidation+@miniAODValidation,DQM:@standardDQM+@ExtraHLT+@miniAODDQM --datatier GEN-SIM-RECO,MINIAODSIM,DQMIO --io RECOPRMXUP18_PU25.io --python RECOPRMXUP18_PU25.py --no_exec --filein file:step2a.root --fileout file:step3.root --nThreads 4`

### DIFF
--- a/Configuration/DataProcessing/python/Merge.py
+++ b/Configuration/DataProcessing/python/Merge.py
@@ -27,6 +27,8 @@ def mergeProcess(*inputFiles, **options):
     - newDQMIO : specifies if the new DQM format should be used to merge the files
     - output_file : sets the output file name
     - output_lfn : sets the output LFN
+    - mergeNANO : to merge NanoAOD
+    - bypassVersionCheck : to bypass version check in case merging happened in lower version of CMSSW (i.e. UL HLT case). This will be TRUE by default.
 
     """
     #  //
@@ -39,6 +41,7 @@ def mergeProcess(*inputFiles, **options):
     dropDQM = options.get("drop_dqm", False)
     newDQMIO = options.get("newDQMIO", False)
     mergeNANO = options.get("mergeNANO", False)
+    bypassVersionCheck = options.get("bypassVersionCheck", True)
     #  //
     # // build process
     #//
@@ -52,6 +55,7 @@ def mergeProcess(*inputFiles, **options):
         process.add_(Service("DQMStore"))
     else:
         process.source = Source("PoolSource")
+        process.source.bypassVersionCheck = CfgTypes.untracked.bool(bypassVersionCheck)
         if dropDQM:
             process.source.inputCommands = CfgTypes.untracked.vstring('keep *','drop *_EDMtoMEConverter_*_*')
     process.source.fileNames = CfgTypes.untracked(CfgTypes.vstring())
--- a/Configuration/DataProcessing/test/RunMerge.py
+++ b/Configuration/DataProcessing/test/RunMerge.py
@@ -22,7 +22,9 @@ class RunMerge:
         self.outputLFN = None
         self.inputFiles = []
         self.newDQMIO = False
-        
+        self.mergeNANO = False
+        self.bypassVersionCheck = True
+
 
     def __call__(self):
         if self.inputFiles == []:
@@ -36,7 +38,8 @@ class RunMerge:
                 output_file = self.outputFile,
                 output_lfn = self.outputLFN,
                 newDQMIO = self.newDQMIO,
-                mergeNANO = self.mergeNANO)
+                mergeNANO = self.mergeNANO
+                bypassVersionCheck = self.bypassVersionCheck)
         except Exception as ex:
             msg = "Error creating process for Merge:\n"
             msg += str(ex)
@@ -52,7 +55,7 @@ class RunMerge:
 
 
 if __name__ == '__main__':
-    valid = ["input-files=", "output-file=", "output-lfn=", "dqmroot", "mergeNANO" ]
+    valid = ["input-files=", "output-file=", "output-lfn=", "dqmroot", "mergeNANO", "bypassVersionCheck" ]
              
     usage = """RunMerge.py <options>"""
     try:
@@ -76,7 +79,9 @@ if __name__ == '__main__':
             merger.outputLFN = arg
         if opt == "--dqmroot" :
             merger.newDQMIO = True
-        if  opt == "--mergeNANO" :
+        if opt == "--mergeNANO" :
             merger.mergeNANO = True
+        if opt == "--bypassVersionCheck" :
+            merger.bypassVersionCheck = True
 
     merger()
--- a/DataFormats/Provenance/interface/StoredMergeableRunProductMetadata.h
+++ b/DataFormats/Provenance/interface/StoredMergeableRunProductMetadata.h
@@ -1,0 +1,152 @@
+#ifndef DataFormats_Provenance_StoredMergeableRunProductMetadata_h
+#define DataFormats_Provenance_StoredMergeableRunProductMetadata_h
+
+/** \class edm::StoredMergeableRunProductMetadata
+
+This class holds information used to decide how to merge together
+run products when multiple run entries with the same run number
+and ProcessHistoryID are read from input files contiguously. This
+class is persistent and stores the information that needs to be
+remembered from one process to the next. Most of the work related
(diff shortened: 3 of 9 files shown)


Dr15Jones:
@wddgit do you foresee any problems including the meta data object in the back port without also backporting the code that uses the meta data?

srimanob:
Hi @Dr15Jones 
What am I missing, as I use https://github.com/cms-sw/cmssw/pull/25171 as the reference?
Thanks.

wddgit:
I think there would be no problems with anything that is not a mergeable run product. No crashes or obvious bad behavior.  This metadata object itself is stored in the MetaData TTree. So if you read a file created with 10_3_X or later with 10_2_X as modified by this code, then this object would simply be dropped. There would be nothing to read or write the object or the branch that contains it. Note that this implies any information it contains would be lost.

Mergeable run products would not behave properly in some cases. For example, if you had split a run in one of a series of 10_3_X processes, then processed the file with 10_2_X the mergeable run products would be merged via the old merging algorithm in that 10_2_X process or any subsequent process, whether or not the subsequent process is 10_2_X or 10_3_X. The information about how the runs were split would be lost and I am pretty sure it would just use the old algorithm when merging those run products. This is what is already done when reading file created and only processed with 10_2_X  which is later read by something from after 10_2_X. Now it is likely that no cares about this problem ... I don't know.

I'll also add in a warning that I did not test this and only thought about it for about 30 minutes so I could be missing something ...

franzoni:
please test

srimanob:
Hi @Dr15Jones @wddgit @fabiocos 
Regarding the concern, could you please explain a bit more if something we should think in 94 release also that we did the same backports, or this is a concern on 10_2 only. Do I miss anything?

Thanks very much in advance.

fabiocos:
@Dr15Jones @wddgit @srimanob the backport of the provenance code is part of both 80X and 94X UL versions, so if there is an issue it is a general one with the UL processing strategy I would say. As far as I can see in practice this would affect LHERunInfoProduct.h and GenLumiInfoProduct.h correct me in case. @srimanob @franzoni were these objects checked in the validation of the UL chain for 94X for instance?

srimanob:
@pgunnell @qliphy @efeyazgan 
In case, you can answer @fabiocos question. Thanks.

Dr15Jones:
@fabiocos you'd just get the regular 8_0_X and 9_4_X behavior for merging and not the improved algorithm used in 10_3_X. So from the framework side we do not think it will cause any (new) problems.

Configuration/DataProcessing/python/Merge.py
```diff
@@ -52,6 +55,7 @@ def mergeProcess(*inputFiles, **options):
         process.add_(Service("DQMStore"))
     else:
         process.source = Source("PoolSource")
+        process.source.bypassVersionCheck = CfgTypes.untracked.bool(bypassVersionCheck)
```
fabiocos: @srimanob this is not the final version we have in master and 10_6_X, although it is consistent with 9_4_X and 8_0_X. As far as I can see they are anyway equivalent as final functionality

fabiocos:
@efeyazgan @qliphy any comment on this PR? What about the run products handling? Otherwise are you ready to sign it?

@Dr15Jones I understand that there is no issue in your view. Do you suggest any further check? Or can we merge it as it is as far as core is concerned?

qliphy:
Adding @agrohsje

We can check more, but it may take some time. And as we agree with @Dr15Jones that  it will not cause new problems, so if it is urgent, we can sign it first.

Configuration/DataProcessing/python/Merge.py
srimanob: @fabiocos Thanks for checking. This is following the discussion we had before that we turn on the bypassversioncheck by default for the release that will be used for UL-HLT. 


If we follow the master, we need to handle at the cmsDriver level.

Dr15Jones:
> I understand that there is no issue in your view. Do you suggest any further check? Or can we merge it as it is as far as core is concerned?

It should be fine as far as core is concerned.

Configuration/DataProcessing/python/Merge.py
fabiocos: @srimanob yes, just let's keep in mind that while new 8_0_X and 9_4_X will now be used basically only for UL, 10_2_X is still being used for other purposes, and this change will affect not only UL but whatever other production updated to this version, there is no special UL branch

fabiocos:
@efeyazgan @qliphy @agrohsje the check above is not only for this PR, actually as I suggested at the PPD general meeting I would make the check for the 2017 chain using 94X first, this PR will make 10_2_X not better nor worse than that. It depends on the urgency to have the release for tests, I would say that we could move forward with the integration and have the tests in parallel

Configuration/DataProcessing/python/Merge.py
srimanob: @fabiocos 
This option can't config from McM side as McM don't touch the Merge process. We also don't have the switch in wmcore too. What we agreed on 
https://github.com/dmwm/WMCore/issues/9054
is to set the default value to true by default. The original PR was done in such a way that we can make it configurable in future.

qliphy:
@fabiocos  Thanks!  GEN will then sign this PR and in the meantime we will try the checks you suggested.

qliphy:
+1

fabiocos:
+operations

Configuration/DataProcessing/python/Merge.py
fabiocos: @srimanob right, thank you for the reminder, the important things is that we are all aware of the implication

fabiocos:
@Dr15Jones so is core ready to sign this PR? I understand PPD would like to have a new version to start testing the 2018 chain

Dr15Jones:
+1

fabiocos:
+1

PR #27448, opened 2019-07-05
Fix PixelCPEBase::fillDetParams() for Phase-2 IT, using new GeomDetEnumerators methods

mmusich:
#### PR description:

While trying to move the reading of the Lorentz Angle for the phase-2 Inner Tracker detector to DataBase, we noticed that when not providing the LA value for all the `DetId`s (even OuterTracker ones), the ` SiPixelRecHitConverter` was complaining with errors of the type:

```
%MSG-e SiPixelLorentzAngle: SiPixelRecHitConverter:siPixelRecHits 04-Jul-2019 18:56:11 CEST Run: 1 Event: 3
SiPixelLorentzAngle for DetID 419697669 is not stored
%MSG
```
while not effectively crashing or giving bogus results.
We traced back the issue to the logic implemented in the `fillDetParams` method of `PixelCPEBase` which looks for the first not-Inner Tracker `DetUnit` to determine the size of the DetUnit list to loop over.
Currently the list is stopped at the first not Strip `DetUnit` via the method `GeomDetEnumerators::isTrackerStrip`. 
This unfortunately doesn't work in the case of the Phase-2 detector as the `GeomDetEnumerators::isTrackerStrip` takes into account only the current Strip Subdetectors.
We solved this inconvenience by add new methods:
  * `isInnerTracker`
  * `isOuterTracker`

in `GeomDetEnumerators` and `GeomDetType` and use them to fix the logic to hold also for Phase-2.
These methods seem to be quite handy, therefore I am wondering if there is a specific reason why they were not implemented before?
The reason why the current setup is not producing these error messages is because the Lorentz Angle used in the reconstruction comes from a Fake conditions ESProducer: [SiPixelFakeLorentzAngleESSource](https://github.com/cms-sw/cmssw/blob/master/CalibTracker/SiPixelESProducers/plugins/SiPixelFakeLorentzAngleESSource.cc) which produces a value for all the DetId list from “Skimmed Geometry” [external file](https://github.com/cms-data/SLHCUpgradeSimulations-Geometry/tree/master/PhaseII/Tilted) (so **all** of them).
Some more information about this can be found in this presentation at the [Tracker Phase-2 simulation,modeling and performance meeting ](https://indico.cern.ch/event/809436/#2-impelentation-on-innertracke).

I profit of this PR to make two other light fixes:
   * apply the same logic fix also the `PixelCPEClusterRepair::fill2DTemplIDs()` method (72cee37)
   * purely technical change, use `MagneticField::nominalValue()` as suggested [here](https://github.com/cms-sw/cmssw/pull/27043#pullrequestreview-248117632) (59ee713)

#### PR validation:

Code proposes passes all tests of:
```
runTheMatrix.py -l limited -t 4 -j 8
```

This is not a backport.

cc:
@tsusa, @pmaksim1, @OzAmram, @cmantill, @emiglior

### DIFF
--- a/RecoLocalTracker/SiPixelRecHits/src/PixelCPEClusterRepair.cc
+++ b/RecoLocalTracker/SiPixelRecHits/src/PixelCPEClusterRepair.cc
@@ -80,10 +80,13 @@ PixelCPEClusterRepair::PixelCPEClusterRepair(edm::ParameterSet const& conf,
   speed_ = conf.getParameter<int>("speed");
   LogDebug("PixelCPEClusterRepair::PixelCPEClusterRepair:") << "Template speed = " << speed_ << "\n";
 
-  GlobalPoint center(0.0, 0.0, 0.0);
-  float theMagField = mag->inTesla(center).mag();
+  // this returns the magnetic field value in kgauss (1T = 10 kgauss)
+  int theMagField = mag->nominalValue();
 
-  if (theMagField >= 3.65 && theMagField < 3.9) {
+  if (theMagField >= 36 && theMagField < 39) {
+    LogDebug("PixelCPEClusterRepair::PixelCPEClusterRepair:")
+        << "Magnetic field value is: " << theMagField << " kgauss. Algorithm is being run \n";
+
     templateDBobject2D_ = templateDBobject2D;
     fill2DTemplIDs();
   }
@@ -103,8 +106,8 @@ PixelCPEClusterRepair::PixelCPEClusterRepair(edm::ParameterSet const& conf,
     recommend2D_.push_back(str);
   }
 
-  // do not recommend 2D if theMagField!=3.8
-  if (theMagField < 3.65 || theMagField > 3.9) {
+  // do not recommend 2D if theMagField!=3.8T
+  if (theMagField < 36 || theMagField > 39) {
     recommend2D_.clear();
   }
 
--- a/RecoLocalTracker/SiPixelRecHits/src/PixelCPEClusterRepair.cc
+++ b/RecoLocalTracker/SiPixelRecHits/src/PixelCPEClusterRepair.cc
@@ -123,10 +123,10 @@ void PixelCPEClusterRepair::fill2DTemplIDs() {
         << "Subdetector " << i << " GeomDetEnumerator " << GeomDetEnumerators::tkDetEnum[i] << " offset "
         << geom_.offsetDU(GeomDetEnumerators::tkDetEnum[i]) << " is it strip? "
         << (geom_.offsetDU(GeomDetEnumerators::tkDetEnum[i]) != dus.size()
-                ? dus[geom_.offsetDU(GeomDetEnumerators::tkDetEnum[i])]->type().isTrackerStrip()
+                ? dus[geom_.offsetDU(GeomDetEnumerators::tkDetEnum[i])]->type().isOuterTracker()
                 : false);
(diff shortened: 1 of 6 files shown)


VinInn:
>These methods seem to be quite handy, therefore I am wondering if there is a specific reason why they were not implemented before?

Laziness...

perrotta:
please test

perrotta:
@mmusich 
Test outputs show a lot of differences in the non-Phase2 workflows (and none in the Phase2 ones...).
According to the PR description this is not what we should have expected: please have a look.

mmusich:
> Test outputs show a lot of differences in the non-Phase2 workflows (and none in the Phase2 ones...).
> According to the PR description this is not what we should have expected: please have a look.

@perrotta: please read carefully again the description. No changes are expected in the Phase2 workflows because of how those workflows are configured.
As for the changes in the Run2 workflows, since there is no change as far as I can see in the Run1 and in the Phase2  I presume it's due to the changes in the `PixelCPEClusterRepair` which is Run only in run2. I'll have a look.

RecoLocalTracker/SiPixelRecHits/src/PixelCPEClusterRepair.cc
```diff
@@ -80,8 +80,7 @@ PixelCPEClusterRepair::PixelCPEClusterRepair(edm::ParameterSet const& conf,
   speed_ = conf.getParameter<int>("speed");
   LogDebug("PixelCPEClusterRepair::PixelCPEClusterRepair:") << "Template speed = " << speed_ << "\n";
 
-  GlobalPoint center(0.0, 0.0, 0.0);
-  float theMagField = mag->inTesla(center).mag();
+  float theMagField = mag->nominalValue();
```
mmusich: @perrotta, apparently this:
```
theMagField = mag->nominalValue();
```
returns the nominal magnetic field in kGauss and not in Tesla... I'll fix it in the next commit

mmusich:
https://github.com/cms-sw/cmssw/pull/27448/commits/600fdef482e2e95f1955ecfb959cf9ab198425b3 should resolve the undesired differences. I have also tested a 0T workflow ( via wf  138.1 proposed in https://github.com/cms-sw/cmssw/pull/27341/)

perrotta:
please test workflow 138.1

mmusich:
OK, now should be fixed for good. I tested one of the workflows which was giving still differences ( see [136.788](https://cmssdt.cern.ch/SDT/jenkins-artifacts/baseLineComparisons/CMSSW_11_0_X_2019-07-06-1100+cba38e/32556/136.788_RunSinglePh2017B+RunSinglePh2017B+HLTDR2_2017+RECODR2_2017reHLT_skimSinglePh_Prompt+HARVEST2017/PixelPhase1_Tracks.html) ) and now I obtain bitwise same results:

image
image

full comparison at [link](http://musich.web.cern.ch/musich/display/diff_PR27448_final.pdf)

perrotta:
please test workflow 138.1

perrotta:
+1
- Replaced the usage of the ```GeomDetEnumerators::isTrackerStrip``` method (which would give buggy results in Phase2) with newly defined ```isOuterTracker```one
- ```isInnerTracker``` implemented as well, for possible future usage
- use ```MagneticField::nominalValue()``` to check whether the B is on, instead of the more complex ```MagneticField::inTesla(...)``` method
- Jenkins tests pass and show no differences, as they should

cvuosalo:
+1

fabiocos:
+1

PR #27459, opened 2019-07-08
Run3-fwk02 Make use of name or fullname as option

bsunanda:
#### PR description:

Now the usage of name vs full name is given through an option. Default is to use full name. To use name rather than full name one need to set fullName=False in the cfg

#### PR validation:

Tested dumping some geometries

No bak porting is required


bsunanda:
@cmsbuild Please test

alja:
+1

fabiocos:
+1

PR #27465, opened 2019-07-09
Update tau anti electron discriminators in RECO and MiniAOD

swozniewski:
#### PR description:

This PR updates the version of the tau AntiElectronDiscriminator in RECO and miniAOD to the latest 2018 version. This is to have the latest tau discriminator versions ready for a coming integration of tau ID versions into global tag.
The NanoAOD config is updated such that 2018 version is taken from PAT taus and the old 2015 version is recalculated. For already existing input samples from former productions a switch is introduced to do it the other way round (as it was before this PR).

#### PR validation:
Reco and PAT step were run on a test set and output compared to corresponding NanoAOD sample that already contains both versions of AntiEleDiscriminator. NanoAOD step was also run on a test set with different config eras and output checked.


slava77:
@cmsbuild please test

RecoTauTag/Configuration/python/HPSPFTaus_cff.py
```diff
@@ -327,86 +327,86 @@
...
-hpsPFTauDiscriminationByMVA6LooseElectronRejection.mapping[1].cut = cms.string("RecoTauTag_antiElectronMVA6v1_gbr_NoEleMatch_wGwoGSF_BL_WPEff96")
-hpsPFTauDiscriminationByMVA6LooseElectronRejection.mapping[2].cut = cms.string("RecoTauTag_antiElectronMVA6v1_gbr_woGwGSF_BL_WPEff96")
-hpsPFTauDiscriminationByMVA6LooseElectronRejection.mapping[3].cut = cms.string("RecoTauTag_antiElectronMVA6v1_gbr_wGwGSF_BL_WPEff96")
-hpsPFTauDiscriminationByMVA6LooseElectronRejection.mapping[4].cut = cms.string("RecoTauTag_antiElectronMVA6v1_gbr_NoEleMatch_woGwoGSF_EC_WPEff96")
-hpsPFTauDiscriminationByMVA6LooseElectronRejection.mapping[5].cut = cms.string("RecoTauTag_antiElectronMVA6v1_gbr_NoEleMatch_wGwoGSF_EC_WPEff96")
-hpsPFTauDiscriminationByMVA6LooseElectronRejection.mapping[6].cut = cms.string("RecoTauTag_antiElectronMVA6v1_gbr_woGwGSF_EC_WPEff96")
-hpsPFTauDiscriminationByMVA6LooseElectronRejection.mapping[7].cut = cms.string("RecoTauTag_antiElectronMVA6v1_gbr_wGwGSF_EC_WPEff96")
+hpsPFTauDiscriminationByMVA6LooseElectronRejection.mapping[0].cut = cms.string("RecoTauTag_antiElectronMVA6v3_noeveto_gbr_NoEleMatch_woGwoGSF_BL_WPeff90")
```
slava77: can't all of this be rolled up in a simple loop? smth like
```python
for m in hpsPFTauDiscriminationByMVA6LooseElectronRejection.mapping : 
    m.cut = m.cut.replace("WPeff98", "WPeff90")
```

PhysicsTools/NanoAOD/python/taus_updatedMVAIds_cff.py
```diff
@@ -534,15 +538,273 @@
...
+        cms.PSet(
+            category = cms.uint32(15),
+            cut = cms.string('RecoTauTag_antiElectron'+antiElectronDiscrMVA6_version+'_gbr_wGwGSF_EC_WPEff99'),
+            variable = cms.string('pt')
+        )
+    )
+)
+# Loose
+patTauDiscriminationByLooseElectronRejectionMVA62015 = patTauDiscriminationByVLooseElectronRejectionMVA62015.clone(
+    mapping = cms.VPSet(
```
slava77: can't all of the below be rolled up in a simple loop? smth like
```python
patTauDiscriminationByLooseElectronRejectionMVA62015 = patTauDiscriminationByVLooseElectronRejectionMVA62015.clone()
for m in patTauDiscriminationByLooseElectronRejectionMVA62015.mapping : 
    m.cut = m.cut.replace("WPeff99", "WPeff96")
```

slava77 (commented):
the implementations to change WPeffXX to WPeffYY appear to be extremely verbose.
Unless I missed some irregularity in the pattern, it seems more practical to compactify

PhysicsTools/NanoAOD/python/taus_updatedMVAIds_cff.py
swozniewski: Indeed, why not. Update follows.

swozniewski: cms.string does not seem to support replace or conversion to python string. Do you have an idea, how to do this in an elegant way?

swozniewski: if it gets too ugly, we can also wait for the new tau discriminator format, where these WPs and categories will be separated in the configurations and then 'multiplied' by the module. https://github.com/swozniewski/cmssw/blob/restructure_PFTauDiscriminators_v3/RecoTauTag/Configuration/python/HPSPFTaus_cff.py#L320-L376

slava77: .pythonValue may be needed on rhs (before .replace)
https://github.com/cms-sw/cmssw/blob/master/FWCore/ParameterSet/python/Types.py#L300
@Dr15Jones @makortel

swozniewski: Thanks for the hint, this could be used! However, this function puts the string in single quotes, which I have to remove:
`m.cut = cms.string(m.cut.configValue().strip("'").replace("WPeff98", "WPeff90"))`
Does not look very nice but would work...

slava77: there is also a .configValue return.

Dr15Jones: Use `m.cut.value()` which returns the held python string.  Remember, you can use python to look at any of the classes to find what methods they have
```bash
bash>python
>>> import FWCore.ParameterSet.Config as cms
>>> help(cms.string)
```

swozniewski: Thanks! This looks good

swozniewski:
sorry, I forgot to remove the print statements

PhysicsTools/NanoAOD/python/taus_updatedMVAIds_cff.py
```diff
@@ -355,194 +359,124 @@
...
+patTauDiscriminationByVTightElectronRejectionMVA62018 = patTauDiscriminationByVLooseElectronRejectionMVA62018.clone()
+for m in patTauDiscriminationByVTightElectronRejectionMVA62018.mapping:
+    m.cut = m.cut.value().replace("WPeff98", "WPeff60")
+### Define v1 anit-e discriminants (2015)
+antiElectronDiscrMVA6_version = "MVA6v1"
+## Raw
+patTauDiscriminationByElectronRejectionMVA62015Raw = patTauDiscriminationAgainstElectronMVA6.clone(
+    Prediscriminants = noPrediscriminants, #already selected for MiniAOD
+    vetoEcalCracks = True, #don't keep tau candidates in EB-EE cracks for v1
+    mvaName_NoEleMatch_wGwoGSF_BL = 'RecoTauTag_antiElectron'+antiElectronDiscrMVA6_version+'_gbr_NoEleMatch_wGwoGSF_BL',
```
slava77: this looks almost exactly like patTauDiscriminationByElectronRejectionMVA62018Raw, except for the veto.
If that's the intent, please change to 
```
patTauDiscriminationByElectronRejectionMVA62015Raw = patTauDiscriminationByElectronRejectionMVA62018Raw.clone(vetoEcalCracks = False)
```

PhysicsTools/NanoAOD/python/taus_updatedMVAIds_cff.py
```diff
@@ -355,194 +359,124 @@
...
+    PATTauProducer = patTauDiscriminationByElectronRejectionMVA62015Raw.PATTauProducer,
+    Prediscriminants = patTauDiscriminationByElectronRejectionMVA62015Raw.Prediscriminants,
+    toMultiplex = cms.InputTag("patTauDiscriminationByElectronRejectionMVA62015Raw"),
+    key = cms.InputTag("patTauDiscriminationByElectronRejectionMVA62015Raw","category"),
     mapping = cms.VPSet(
         cms.PSet(
             category = cms.uint32(0),
-            cut = cms.string('RecoTauTag_antiElectron'+antiElectronDiscrMVA6_version+'_gbr_NoEleMatch_woGwoGSF_BL_WPeff60'),
+            cut = cms.string('RecoTauTag_antiElectron'+antiElectronDiscrMVA6_version+'_gbr_NoEleMatch_woGwoGSF_BL_WPEff99'),
             variable = cms.string('pt')
```
slava77: it looks like the mapping values are identical to what already exists in this file in patTauDiscriminationByVLooseElectronRejectionMVA62018 with 98 -> 98.
It would be nice to respect this similarity and use a similar string replace

PhysicsTools/NanoAOD/python/taus_updatedMVAIds_cff.py
swozniewski: please note that the variable antiElectronDiscrMVA6_version changes. Therefore, the mva_Name parameters also have to be set, so this would not save much space and add additional dependency which makes future changes a bit more complicated. If you agree, I'd keep it like that.

slava77: ok for me (partly also because it's not my signoff area).

PhysicsTools/NanoAOD/python/taus_updatedMVAIds_cff.py
swozniewski: Ok, I have added this in the latest commit

slava77:
@cmsbuild please test

slava77:
+1

for https://github.com/cms-sw/cmssw/pull/27465	626a98e
- jenkins tests pass and comparisons with the baseline show differences in expected places
    - only in the tau discriminants of ByMVA6\*ElectronRejection kind; the changes are fairly small on the limited set of events used in the PR tests
    - the only nanoAOD wf based on the old sample (1325.7 using run2_nanoAOD_94XMiniAODv2) in the jenkins tests shows no changes in line with the PR description and implementation to preserve the results

santocch:
+1

fabiocos:
@peruzzim @fgolf could you please scrutinize this PR? The nanoAOD part needs your review

slava77:
> @peruzzim @fgolf could you please scrutinize this PR? The nanoAOD part needs your review

ping

slava77:
> > @peruzzim @fgolf could you please scrutinize this PR? The nanoAOD part needs your review
> 
> ping

peruzzim:
1) We have to really try to limit this kind of PR, that change things "in cascade" through several data formats. It's very difficult to review. Are the changes to nanoAOD only needed for future miniAOD productions, or are you also touching the output on past productions?
2) What do you plan to do for 102X?

mbluj:
Hello Marco,
I think that there is some misunderstanding, let try to explain what was done and then answer your comments.
What this PR does is an update of anti-electron tauID to the newer training in AOD and hence in MiniAOD and NanoAOD which base on AOD. But, the anti-electron tauID with the new training has been already run on top of MiniAOD and stored to NanoAOD, Therefore, to keep content of NanoAOD unchanged when produced with new inputs it was mandatory to swap new and old anti-e tauID in NanoAOD sequences, i.e. read new from MiniAOD and produce old on top of it. For old eras noting is changed. 

>     1. We have to really try to limit this kind of PR, that change things "in cascade" through several data formats. It's very difficult to review. Are the changes to nanoAOD only needed for future miniAOD productions, or are you also touching the output on past productions?
I think it is impossible to avoid such "cascade" PRs if changes are introduced to any pre-Nano level in case they were already intdouced to NanoAOD. I think it happens because release cycle of Nano is shorther than for AOD and MiniAOD. 
Answering your question: the changes do not modify NanoAOD content regardless if it is produced on top of new MiniAOD samples with >=110X or old ones with 94X/102X as long as appropriate era modifiers are used. 
> 
>     2. What do you plan to do for 102X?
Nothing, i.e. no backport. It is as changes in AOD/MiniAOD are not foreseen/allowed for closed releases.

Finally, Tau POG plans to clear NanoAOD from outdated Run-2 tauIDs, but it is beyond scope of this PR (and will be performed via NanoAOD repo)

peruzzim:
Thanks for clarifying. I am going to test this and then sign, but I don't foresee any issues.

slava77:
> Thanks for clarifying. I am going to test this and then sign, but I don't foresee any issues.

@peruzzim 
please clarify on the status of your checks, or perhaps sign if all is OK.
Thank you.

fabiocos:
@peruzzim any news about this?

slava77:
> @peruzzim any news about this?

ping

peruzzim:
+xpog

I have not detected changes that would break the existing setup, the status of discriminants stored in nanoAOD for different eras will be rediscussed for the future when x-checking the sync between 11_0 and previous releases.

kpedro88:
+1

fabiocos:
@mbluj @swozniewski @slava77 @perrotta @peruzzim the PPD/PdmV validation report about CMSSW_11_0_0_pre7 gave positive feedback on this PR. Is its backport to 10_6_X forseen? With protections like those added in #27882?

mbluj:
> @mbluj @swozniewski @slava77 @perrotta @peruzzim the PPD/PdmV validation report about CMSSW_11_0_0_pre7 gave positive feedback on this PR. Is its backport to 10_6_X forseen? With protections like those added in #27882?

No, backport of this is not expected. There are two reasons: the main one is the fact that DeepTauID overperforms all current BDT-based TauIDs (including anti-e 2018) so it is going to be the main POG recomended set of TauID, the other reason is that anti-e MVA 2018 is already available for users via NanoAOD (or can be run on top of MiniAOD by whose are not using NanoAOD).

slava77:
> > @mbluj @swozniewski @slava77 @perrotta @peruzzim the PPD/PdmV validation report about CMSSW_11_0_0_pre7 gave positive feedback on this PR. Is its backport to 10_6_X forseen? With protections like those added in #27882?
> 
> No, backport of this is not expected. There are two reasons: the main one is the fact that DeepTauID overperforms all current BDT-based TauIDs (including anti-e 2018) so it is going to be the main POG recomended set of TauID, the other reason is that anti-e MVA 2018 is already available for users via NanoAOD (or can be run on top of MiniAOD by whose are not using NanoAOD).

I'm curious if this expectation changed since Sep 20 of last year.
While investigating differences between the variables in re-miniAOD of UL (with features collected in #27889) I noticed that againstElectron\*MVA6 discriminants differ the most

mbluj:
> 
> 
> > > @mbluj @swozniewski @slava77 @perrotta @peruzzim the PPD/PdmV validation report about CMSSW_11_0_0_pre7 gave positive feedback on this PR. Is its backport to 10_6_X forseen? With protections like those added in #27882?
> > 
> > 
> > No, backport of this is not expected. There are two reasons: the main one is the fact that DeepTauID overperforms all current BDT-based TauIDs (including anti-e 2018) so it is going to be the main POG recomended set of TauID, the other reason is that anti-e MVA 2018 is already available for users via NanoAOD (or can be run on top of MiniAOD by whose are not using NanoAOD).
> 
> I'm curious if this expectation changed since Sep 20 of last year.
> While investigating differences between the variables in re-miniAOD of UL (with features collected in #27889) I noticed that againstElectron*MVA6 discriminants differ the most

I think that the issue is related with the fact that for some time payloads of BDT-based tauIDs (including `againstElectron*MVA6` ones) in 11XY series are being mapped via a GT and the newest training is used in this mapping. In 106X series mappig is set by python config and newest training is used for `IsolationMVArun2*` discriminants, but not `againstElectron*MVA6` ones. To agree behavior of re-miniAOD of UL with 106X and 11XY one needs rerun the `againstElectron*MVA6` discriminants with the newest training (and adjust NanoAOD) with 106X - can be done if requested, but an effort needed.

slava77:
> I think that the issue is related with the fact that for some time payloads of BDT-based tauIDs (including `againstElectron*MVA6` ones) in 11XY series are being mapped via a GT and the newest training is used in this mapping. In 106X series mappig is set by python config and newest training is used for `IsolationMVArun2*` discriminants, but not `againstElectron*MVA6` ones. To agree behavior of re-miniAOD of UL with 106X and 11XY one needs rerun the `againstElectron*MVA6` discriminants with the newest training (and adjust NanoAOD) with 106X - can be done if requested, but an effort needed.

Thank you for clarifying that the main difference is in the payloads.
This means that an update for UL workflows can be rather straightforward (configuration/GT switch).
Would any analysis using UL re-miniAOD be recommended to use the `againstElectron*MVA6` ?
If so, then an update is in order. If these are instead unused tags computed just in case, then it may be fine to leave them without an update.

mbluj:
> 
> 
> > I think that the issue is related with the fact that for some time payloads of BDT-based tauIDs (including `againstElectron*MVA6` ones) in 11XY series are being mapped via a GT and the newest training is used in this mapping. In 106X series mappig is set by python config and newest training is used for `IsolationMVArun2*` discriminants, but not `againstElectron*MVA6` ones. To agree behavior of re-miniAOD of UL with 106X and 11XY one needs rerun the `againstElectron*MVA6` discriminants with the newest training (and adjust NanoAOD) with 106X - can be done if requested, but an effort needed.
> 
> Thank you for clarifying that the main difference is in the payloads.
> This means that an update for UL workflows can be rather straightforward (configuration/GT switch).
> Would any analysis using UL re-miniAOD be recommended to use the `againstElectron*MVA6` ?
> If so, then an update is in order. If these are instead unused tags computed just in case, then it may be fine to leave them without an update.

After some thinking, we find that it makes sense to update `againstElectron*MVA6` to most recent training also for UL re-miniAOD with 10_6_X (at least for coherency with 11_X_Y). A related PR will be prepared soon.

PR #27476, opened 2019-07-10
DQM: header and namespace cleanup

schneiml:
#### PR description:

This PR separates a few changes out of the big `DQMStore` restructuring project [1] that can be merged safely. No behaviour changes. These changes include things that need to be fixed/explicitly declared in subsystem code, therefore this PR touches many packages and will conflict almost certainly with others.

This is the reason to split out these changes and merge them now.

There are two big sets of changes:
- Simplify includes. The only files to be included from the `DQMServices/Core` package are `DQMStore.h` and the module base classes (`DQMEDAnalyzer.h`, `DQMEDHarvester.h`, `oneDQMEDAnalyzer.h`, `DQMGlobalEDAnalyzer.h`). All others should the replaced with `DQMStore.h`.
    - This implies that `MonitorElement.h` should never be included, the majority the majority of changes is removing these.
    - There are still a lot of redundant includes and a few non-standard ones, that I did not bother to hunt down.
    - `DQMStore` and `MonitorElement` should never be forward-declared, to prevent problems in the next steps.

- Namespace `MonitorElement` and `DQMStore` into `dqm::reco::`, `dqm::harvesting::` and `dqm::legacy::` (and `dqm::impl::`, which is the actual implementation).
    - There is (intentionally) now more `DQMStore` and `MonitorElement` in the default namespace. Most code will continue to work unchanged since `DQMEDAnalyzer` and `DQMEDHarvester` import them into class scope.
    - For now all of these are the same class, but in the future they will start to expose different APIs. 
    - All _legacy_ modules, that is those using DQM but not deriving from `DQMEDAnalyzer` or `DQMEDHarvester`, need to import the correct definition themselves. `typedef`s to to that are the second big batch of changes.
    - Many of the classes that now import `dqm::legacy::` are actually helper classes and not legacy modules by themselves. These could be changed to `dqm::reco::` or `dqm::harvesting::` based on the APIs that they use, but that will be easier once the reduced APIs are actually in place.
    - The only visible change in non-legacy code is that functions returning `MonitorElement` defined outside the class declaration now need to qualify the `MonitorElement` in their return type. This is due to how C++ scoping works and hard to avoid.

In addition,
- A lot of `DQMServices/Examples` is deleted, since it demonstrates the usage of legacy APIs, some of which are only used in the example code...
- `MonitorElement::Kind` got turned into an `enum class` to keep things a bit more clean in the future.

#### PR validation:
Code compiles.

None of the changes should affect the output, but bugs and mistakes are possible, so we need to watch the comparisons.

PR comparison shows no significant changes.

[1] https://github.com/cms-sw/cmssw/compare/master...schneiml:dqm-new-dqmstore-on-CMSSW_11_0_0_pre1-from-andrius


CalibTracker/SiPixelTools/plugins/SiPixelErrorsDigisToCalibDigis.cc
```diff
@@ -129,22 +129,21 @@ void SiPixelErrorsDigisToCalibDigis::endJob() {
...
-MonitorElement* SiPixelErrorsDigisToCalibDigis::bookDQMHistogram2D(uint32_t detid,
-                                                                   std::string name,
-                                                                   std::string title,
-                                                                   int nchX,
-                                                                   double lowX,
-                                                                   double highX,
-                                                                   int nchY,
-                                                                   double lowY,
-                                                                   double highY) {
+SiPixelErrorsDigisToCalibDigis::MonitorElement* SiPixelErrorsDigisToCalibDigis::bookDQMHistogram2D(uint32_t detid,
```
Dr15Jones: I’m fairly certain you could have used ```auto``` as the return type.

Alignment/LaserDQM/plugins/LaserDQM.h
```diff
@@ -15,12 +15,13 @@
...
 // DQM
 #include "DQMServices/Core/interface/DQMStore.h"
-#include "DQMServices/Core/interface/MonitorElement.h"
 
 #include <iostream>
 
 class LaserDQM : public edm::EDAnalyzer {
 public:
+  typedef dqm::legacy::DQMStore DQMStore;
+  typedef dqm::legacy::MonitorElement MonitorElement;
```
Dr15Jones: Something to keep in mind for the future. We highly encourage the use of ```using``` instead of ```typedef``` in all future code development.

CalibTracker/SiPixelTools/plugins/SiPixelErrorsDigisToCalibDigis.cc
schneiml: Probably yes, but that requires changes in two places (header and implementation). Also the return type might not (always) be obvious, so I went the most obvious route.

Alignment/LaserDQM/plugins/LaserDQM.h
schneiml: I used `using` in some places, but then switched to `typedef` for symmetry with those cases where the declarations are put in namespaces (fewer, but still some).

schneiml:
please test

hoping that this at least compiles...

schneiml:
I am rather surprised by these results. Seems the output turned almost (but not entirely) empty. Investigating.

schneiml:
Some changes, but still no fix for the issue. Seems the legacy saving code does not find it's MEs any more...

schneiml:
please test

Turns out there where some unrelated changes in on of the commits, which removed some functionality related to `enableMultiThread`. Should be better now.

schneiml:
+1

RecoParticleFlow/Benchmark/src/GenericBenchmark.cc
```diff
@@ -1,9 +1,8 @@
 #include "RecoParticleFlow/Benchmark/interface/GenericBenchmark.h"
 #include "RecoParticleFlow/Benchmark/interface/BenchmarkTree.h"
-#include "DQMServices/Core/interface/MonitorElement.h"
+#include "DQMServices/Core/interface/DQMStore.h"
```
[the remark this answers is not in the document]
schneiml: Note that all these changes are machine generated, I will integrate your comments but there might be more.

schneiml:
please test

... to see that things still compile with the removed includes.

perrotta:
+1
- Ok for the (quite limited and technical) reco changes now
- Perhaps, if these machine generated code changes require some manual cleaning also in other sectors I would personally fix them now that you have everything at hand, instead of waiting that they get hidden in the separate packages: but this is up to you to decide

Martin-Grunewald:
+1

kpedro88:
+upgrade
for SimTracker/SiPhase2Digitizer

santocch:
+1

fabiocos:
code-checks

Dr15Jones:
+1

For Validation/Performance

Alignment/LaserDQM/plugins/LaserDQM.h
fabiocos: @schneiml could you please clarify the rationale behind this choice? Is there a specific reason I am missing, or is it mostly a matter of taste?

fabiocos:
@schneiml is dqm ready to sign this PR?

@christopheralanwest @pohsun @ggovi @civanch @peruzzim @fgolf @ssekmen @sbein @rekovic the changes in your areas look simple, could you please check and comment or sign it? We need to integrate this PR asap in order to prevent new conflicts

FastSimulation/MaterialEffects/interface/NuclearInteractionSimulator.h
```diff
@@ -14,6 +14,7 @@
  */
 
 #include "FastSimulation/MaterialEffects/interface/MaterialEffectsSimulator.h"
+#include "DQMServices/Core/interface/DQMStore.h"
```
fabiocos: @schneiml @ssekmen @sbein as ME are commented, is this (and the previous update to BuildFile) really needed? Otherwise it would be good to clean it (possibly in a separate PR)

fabiocos:
@qliphy @efeyazgan @alberto-sanchez please check also the update in your domain, it looks straightforward

pohsun:
+1

alberto-sanchez:
+1

indeed the changes are associated to RivetInterface and are straightforward

ggovi:
+1

fabiocos:
@schneiml I assume this is ok for you, please sign it, I would like to merge it tomorrow

@civanch @sbein @ssekmen @rekovic @peruzzim the updates in your are look ok to me, please check them anyway

ssekmen:
+1

fabiocos:
@schneiml @andrius-k @jfernan2 ping

jfernan2:
+1
@fabiocos   Marcel is on holiday this week, I believe he was OK with merging it, we clearly forgot to discuss this on last Friday meeting, I cannot comment on the PR notes made though

civanch:
+1

fabiocos:
+1

fabiocos:
merge

Alignment/LaserDQM/plugins/LaserDQM.h
schneiml: Regarding typedef vs. using: `using` only works in class scopes, but some of these declarations also appear in namespaces, where afaik `using` can't be used. So I chose to use the form of declaration that works everywhere. 

Another option is to not import the names but just use full names. I also did that in a few places where the declarations caused trouble (e.g. due to multi-inheritance), but in general I preferred to have the import since it leads to less changes.

Dr15Jones: `using` can also be used outside of class scope.

PR #27370, opened 2019-06-26
RecoEgamma/ElectronIdentification cleanup

guitargeek:
#### PR description:

This is one of the occasional PRs that removes old electron IDs as the goal is still to have only the VIDs and other currently used IDs left in the release. This time, I remove the old Likelihood based IDs and some old cut based IDs. One working point is still used in PF, so I move it there.

@Sam-Harper

#### PR validation:

CMSSW compiles and local matrix tests pass.

#### if this PR is a backport please specify the original PR:

No backport intended.

PR #26967, opened 2019-05-28
Primitive RPC trigger

maseguracern:
This is the first version of the RPCTriggerPrimitives for the RPC detector. Which uses as an input the RPCDigis as an input and the RPCRecHits (New collection) as an output. The output of this module is an edm branch named RPCPrimitivesDigis, following the RPCRecHit format already committed in CMSSW_10_6_0. We apply the cluster size cut and emulate max two clusters per link board. The module can be tuned by the parameters LinkBoardCut and ClusterSizeCut for phaseII. 

People interested: @maseguracern @BrieucF @abrinke1

PR #27477, opened 2019-07-10
RecoEgamma/EgammaElectronProducers cleanup

guitargeek:
#### PR description:

In the same spirit as https://github.com/cms-sw/cmssw/pull/27370 I propose another package-centric RecoEgamma cleanup. This time I took a look at the EgammaElectronProducers package and made the following cleanups/changes:

* remove unused python configuration fragment files
* merge some configuration fragment files that only get used at one place anyway
* change the `fillDescriptions` methods of the LowPt* producers to contain what is actually used in the configs so separate `cfi` files are not needed anymore
* implement a `produce` function in the GsfElectronBaseProducer so it can be used directly in configs. This way, the GsfElectronEcalDrivenProducer and LowPtGsfElectronProducer can be removed. They were exact duplicates so one of them should not have existed anyway
* take the SiStripElectron related code out of the release as suggested by @Sam-Harper at some point if I remember correctly
* turn the LowPtGsfElectronIDProducer into a global producer so the `LowPtGsfElectronIDHeavyObjectCache` could be integrated in it. I think that's good because the concept of a HeavyObjectCache is a bit technical, and having the functionality just in a global producer is easier to understand for non-experts

#### PR validation:

No backport intended.


Sam-Harper:
thank you Jonas!

perrotta:
please test

guitargeek:
That was a bit unexpected, the RelVals failed with the following error:
```
Unable to find plugin 'RPCRecHitStandardAlgo' because the category 'RPCRecHitAlgoFactory' has no known plugins
```
`RPCRecHitStandardAlgo` and `RPCRecHitAlgoFactory` are from the muon reconstruction, I really don't understand how my changes  could affect that and after all, some workflows that failed above worked fine locally. Is everything alright with the bot tests? I will run some local tests again just to make sure.

slava77:
> That was a bit unexpected, the RelVals failed with the following error:

the IB is broken by #26967

guitargeek:
Thanks, no need to rerun local tests in that case!

perrotta:
please test
(offending PR has been reverted in the last IB CMSSW_11_0_X_2019-07-10-2300)

slava77:
@cmsbuild please test
update for reco comparisons in phase-2 wf

RecoEgamma/EgammaElectronProducers/python/ecalDrivenElectronSeeds_cfi.py
```diff
@@ -4,7 +4,94 @@
 # module to produce pixel seeds for electrons from super clusters
 # Author:  Ursula Berthon, Claude Charlot
 #
-from RecoEgamma.EgammaElectronProducers.ecalDrivenElectronSeedsParameters_cff import *
+
+from RecoTracker.IterativeTracking.ElectronSeeds_cff import newCombinedSeeds as _newCombinedSeeds
+
+ecalDrivenElectronSeedsParameters = cms.PSet(
```
slava77: this PSet was apparently defined outside in view of possible reuse elsewhere.
If there is no need to reuse, it seems more practical to simply inline this in the ecalDrivenElectronSeeds definition

RecoEgamma/EgammaElectronProducers/python/gsfElectronSequence_cff.py
```diff
@@ -19,8 +20,89 @@
...
 #gsfElectronMergingSequence = cms.Sequence(gsfElectronCores*gsfElectrons)
 
-from RecoEgamma.EgammaElectronProducers.edBasedElectronIso_cff import *
 from RecoEgamma.EgammaElectronProducers.pfBasedElectronIso_cff import *
+ 
+from CommonTools.ParticleFlow.pfParticleSelection_cff import *
+from RecoEgamma.EgammaElectronProducers.electronPFIsolationDeposits_cff import *
+
+elEDIsoDepositCharged=elPFIsoDepositCharged.clone()
+elEDIsoDepositCharged.src="ecalDrivenGsfElectrons"
```
slava77: what is the purpose of exploding this file in size and merging the content defined in  RecoEgamma/EgammaElectronProducers/python/electronEDIsolationDeposits_cff.py and  RecoEgamma/EgammaElectronProducers/python/electronEDIsolationValues_cff.py

The isolation defs are fairly self-contained and could stay in their corresponding files, exposing just the final sequence needed to build the electronIsoTask.
Overall the purpose of these isolation modules/tasks in gsfElectonSequence_cff is fairly nebulous.
I couldn't find valid usecases for either electronIsoSequence or gsfElectronMergingSequence

RecoEgamma/EgammaElectronProducers/python/lowPtGsfElectronSequence_cff.py
```diff
@@ -57,19 +57,23 @@
 # RecoParticleFlow/PFClusterProducer/python/particleFlowClusterECALUncorrected_cfi.py
 # RecoParticleFlow/PFClusterProducer/python/particleFlowClusterECAL_cff.py
 # (particleFlowClusterECAL_cfi is generated automatically)
-from RecoEgamma.EgammaElectronProducers.lowPtGsfElectronSuperClusters_cff import lowPtGsfElectronSuperClusters
+from RecoEgamma.EgammaElectronProducers.lowPtGsfElectronSuperClusters_cfi import lowPtGsfElectronSuperClusters
 
 # Low pT electron cores
-from RecoEgamma.EgammaElectronProducers.lowPtGsfElectronCores_cff import *
+from RecoEgamma.EgammaElectronProducers.lowPtGsfElectronCores_cfi import lowPtGsfElectronCores
```
slava77: I think that it's better to keep this and the toModify in a lowPtGsfElectronCores_cff so that the modifiers are the closest to their module defs.
This need for modifiers and lack of support for them in the fillDescriptions is a somewhat minor reason to keep defaultLowPtGsfElectronCores for the system-generated cfi and reduce possible confusion for proper consistent include of lowPtGsfElectronCores from a single place lowPtGsfElectronCores_cff (if it's ever needed more than just here).

slava77 (commented):
Is the removal of SiStripElectron\* done assuming the pixel detector is going to always work in the remaining Run-3 time? Or is it so hopelessly obsolete that if something happens, it will be easier to rewrite?

guitargeek:
> Is the removal of SiStripElectron* done assuming the pixel detector is going to always work in the remaining Run-3 time? Or is it so hopelessly obsolete that if something happens, it will be easier to rewrite?

Hi @slava77, that is a good point. Tracking starting from the SiStrips might still be a good strategy, in particular with higher pileup as the occupancy in the pixel gets really high. At least that's what I remember from some talk I saw once. It would be nice to see what @Sam-Harper thinks who proposed taking the SiStripElectron code out of the release.

RecoEgamma/EgammaElectronProducers/python/gsfElectronSequence_cff.py
guitargeek: Hi @slava77, thanks for the comments! You are right, here it should have stayed in a separate file because it is rather self-contained as you say. Putting the isolation definitions in that file was my way of making explicit that they are only used in gsfElectonSequence_cff, but now that the tests passed and we had a first look at it I'll revisit.

RecoEgamma/EgammaElectronProducers/plugins/GEDGsfElectronCoreProducer.cc
```diff
@@ -22,10 +22,12 @@ using namespace reco;
 
 void GEDGsfElectronCoreProducer::fillDescriptions(edm::ConfigurationDescriptions &descriptions) {
   edm::ParameterSetDescription desc;
-  GsfElectronCoreBaseProducer::fillDescription(desc);
-  desc.add<edm::InputTag>("GEDEMUnbiased", edm::InputTag("GEDPFCandidates"));
-
-  descriptions.add("produceEcalDrivenGsfElectronCores", desc);
+  desc.add<edm::InputTag>("gsfPfRecTracks", edm::InputTag("pfTrackElec"));
+  desc.add<edm::InputTag>("gsfTracks", edm::InputTag("electronGsfTracks"));
+  desc.add<edm::InputTag>("ctfTracks", edm::InputTag("generalTracks"));
```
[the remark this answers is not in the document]
guitargeek: Having `GsfElectronCoreBaseProducer::fillDescription` for just two entries is not really worth it I think. What do you think about just removing the `GsfElectronCoreBaseProducer::fillDescription` which is unused anyway and then there is no confusion and just 2 duplicate entries? I think having two repeated entries in 2 derived producers `fillDescriptions` is more explicit and easy than trying to put something together with inheritance. I'm not a bit fan of inheritance with producers anyway...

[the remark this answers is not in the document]
slava77: There are 4 parameters (by name, if not by value) in common, not 2.
There are 3 derived classes (GsfElectronCoreProducer doesn't have a fillDescriptions yet).
So, the repetition seems excessive.

Sam-Harper:
Well given we havent used it to my knowledge since RunI and we have pixel-less electrons (they are called particle flow electrons) I dont see the need to keep old code around.  I think at this point it is hopelessly obsolete. Of course if we have this potential problem and we really want to do this, we can always add the code back in?

Sam-Harper:
In general I would be very pro in aggressively decluttering things with a better to seek forgiveness than ask permission approach. Having a large code base with many obsolete files makes it difficult and intimidating for new comers.

RecoEgamma/EgammaElectronProducers/plugins/GEDGsfElectronCoreProducer.cc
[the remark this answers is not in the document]
guitargeek: I don't understand how we can have the parameters in the base `fillDescription` that will have different values in the derived plugins without re-introducing the manual cfi files to change them again. Isn't this what we want to avoid? That's why I said we can only put the 2 parameters that do not change in the base `fillDescription`. Maybe you can clarify what you mean? Or just take a look at my most recent commit and tell me if it goes in the right direction? I have now `fillDescriptions` for all the derived producers and therefore the `gsfElectronCores_cfi` configuration is not needed anymore. Everything is now configured with the C++ code and the era modifiers.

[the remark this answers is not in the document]
slava77: I meant to mod the base fillDescription to be ```fillDesccription(desc, const string& pfTk, gsfTk)``` and pass appropriate arguments. Recall that only the values of gsfPfRecTracks and gsfTracks change, the parameter names and types themselves are unchanged.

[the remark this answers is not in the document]
guitargeek: I finally see what you mean. Nice we can do it like that! I'll update the last commit.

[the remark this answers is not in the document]
guitargeek: Alright I did now what you suggested and updated the last commit so you can take a look. Now I'm doing local tests. I'll let you know if everything worked.

slava77:
@guitargeek 
thank you for the quick updates.
I will not post "please test" until you say it's OK.

guitargeek:
No problem! Local tests went fine.

slava77:
@cmsbuild please test

guitargeek:
Hi @slava77, I'm sorry I had to update again. I forgot to address your comment with the trailing underscores... Furthermore I had to do a rebase to get rid of the invalid states: I just kept the isolation config files in the first commit that I brought back in the 9th commit. The integral of commits 1-9 is the  same. I hope this does not complicate the review too much...

perrotta:
please test

RecoEgamma/EgammaElectronProducers/plugins/GsfElectronCoreBaseProducer.cc
```diff
@@ -13,11 +13,14 @@
 
 using namespace reco;
 
-void GsfElectronCoreBaseProducer::fillDescription(edm::ParameterSetDescription& desc) {
-  desc.add<edm::InputTag>("gsfPfRecTracks", edm::InputTag("pfTrackElec"));
-  desc.add<edm::InputTag>("gsfTracks", edm::InputTag("electronGsfTracks"));
+edm::ParameterSetDescription GsfElectronCoreBaseProducer::makeBaseDescription(std::string const& gsfPfRecTracks,
```
slava77: why is this copy of a ParameterSetDescription needed?
The old ```fillDescription( ``` was lighter

RecoEgamma/EgammaElectronProducers/plugins/GsfElectronCoreBaseProducer.h
```diff
@@ -27,10 +27,10 @@ namespace edm {
 
 class GsfElectronCoreBaseProducer : public edm::stream::EDProducer<> {
 public:
-  static void fillDescription(edm::ParameterSetDescription&);
+  static edm::ParameterSetDescription makeBaseDescription(std::string const& gsfPfRecTracks,
+                                                          std::string const& gsfTracks);
```
slava77: FWIW, "pfTrackElec" and "electronGsfTracks" can be default values for these arguments (so that it does not have to be repeated in 3 places)

RecoEgamma/EgammaElectronProducers/plugins/GsfElectronCoreBaseProducer.cc
guitargeek: There should be no copy there due to return value optimization. So it's just as light but the input and output is more explicit.

guitargeek: Okay! I was not aware that this is a reoccurring pattern, then it should of course stay consistent.

slava77:
@cmsbuild please test

slava77:
@cmsbuild please test

slava77:
+1

for https://github.com/cms-sw/cmssw/pull/27477	49d5663
- code changes are in line with the PR description and the follow up review; the PR is essentially technical

pohsun:
+1

Configuration/StandardSequences/python/Reconstruction_cff.py
```diff
@@ -299,4 +299,4 @@
 # define a standard candle. please note I am picking up individual
 # modules instead of sequences
 #
-reconstruction_standard_candle = cms.Sequence(localreco*globalreco*vertexreco*recoJetAssociations*btagging*electronSequence*photonSequence)
+reconstruction_standard_candle = cms.Sequence(localreco*globalreco*vertexreco*recoJetAssociations*btagging*gsfElectronSequence*photonSequence)
```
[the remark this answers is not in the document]
fabiocos: indeed this sequence appears only in here in the whole CMSSW repository, so I do not see issues with removing it

fabiocos:
+operations

the update of StandardSequences is minimal, removing an unused sequence in the RECO domain, and coherent with the purpose of the PR

fabiocos:
+1

Issue #27417, opened 2019-07-01
Test wf 20234.99 frequently failing in HGCal validation


fabiocos:
In recent days, starting with CMSSW_11_0_X_2019-06-27-1100, test wf 20234.99 (old scenario D17 with PU) has started to frequently fail with a TimeOut error:
```
%MSG-e ExcessiveTime:  HGCalValidator:hgcalValidator 30-Jun-2019 20:00:56 CEST  Run: 1 Event: 109
ExcessiveTime: Module used 2551.57 seconds of time which exceeds the error threshold configured in the Timing Service of 600 seconds.
%MSG
```
This issues appears after the integration of #27337, where the HGCal validation code was updated. Even if the module does not fail, one should compare the TimeReport before this update
```
TimeReport   3.735532     3.735532     3.735532  hgcalValidator
```
with that for instance in CMSSW_11_0_X_2019-06-30-0000:
```
TimeReport 2106.703824  2106.703824  2106.703824  hgcalValidator
```

This problem should be urgently addressed.

fabiocos:
assign dqm,upgrade

@apsallid FYI

fabiocos:
@kpedro88 we have few Phase2 tests with PU so far in the IB:
```
10:39 cmsdev25 1435> runTheMatrix.py -n | grep ^2 | grep D[0-9][0-9] | grep PU
20234.99 TTbar_14TeV_TuneCUETP8M1_2023D17PU_GenSimHLBeamSpotFull14+SingleNuE10_cf_2023D17PU_PremixHLBeamSpotFull14PU_Premix+DigiFullTriggerPUPRMXCombined_2023D17PU+RecoFullGlobalPUPRMX_2023D17PU+HARVESTFullGlobalPU_2023D17PU 
20261.97 SingleNuE10_cf_2023D17PU_PremixHLBeamSpotFullPU_Premix 
27634.21 TTbar_14TeV_TuneCUETP8M1_2023D35PU_GenSimHLBeamSpotFull14+DigiFullTriggerPU_2023D35PU+RecoFullGlobalPU_ProdLike_2023D35PU+MiniAODFullGlobalPU_ProdLike_2023D35PU 
```
and in 27634.21 (scenario D35) as far as I can see the hgcalValidation does not seem to run:
```
TimeReport   0.002477     0.002477     0.002477  hfreco
TimeReport   0.166499     0.166499     0.166499  hgcalDigis
[...]

PR #27337, opened 2019-06-25
[HGCal] Updating HGCalValidator with SimHits, Digis, RecHits and MultiClusters

apsallid:
#### PR description:

This PR updates the HGCalValidator package to be able to handle: 

1. SimHits, Digis, RecHits (@tmengke, @deguio, @skunori)

As part of the Validation of campaigns regularly announced from PdmV and UPSG, we want to monitor and validate SimHits, Digis and Rechits objects. Usually, RelMon and DQM GUI is being used for this task. However, in order to quantify the performance of certain clustering algorithms or optimizing the same algorithm or checking different calibration algorithms or inspect the plots in the DQM file easier, we include in our webpage results from the above mentioned objects.

@bsunanda: This slightly touches the [HGCalSimHitValidation.cc](https://github.com/cms-sw/cmssw/blob/master/Validation/HGCalValidation/plugins/HGCalSimHitValidation.cc), [HGCalDigiValidation.cc](https://github.com/cms-sw/cmssw/blob/master/Validation/HGCalValidation/plugins/HGCalDigiValidation.cc), [HGCalRecHitValidation.cc](https://github.com/cms-sw/cmssw/blob/master/Validation/HGCalValidation/plugins/HGCalRecHitValidation.cc) by adding a zero to the names of the DQM histograms, so the results in our web page to be ordered by layer number. 

2. Multiclusters (@rovere, @felicepantaleo, @apsallid, @amartelli)

With the inclusion of TICL multiclusters in CMSSW being imminent, we update the HGCalValidator package with the capability of monitoring as well as associating MultiClusters wiith CaloParticles. In the HGCal webpage [1] there is a "Validation" section where we document and describe all the relevant plots. This is a first work based on the standard "hgcalMultiClusters" object that if needed will be refined in the very near future with the study of the TICL multiclusters. 

#### PR validation:

We tested locally that all objects plots are populated [2-4] and will continue the validation of the code as soon as TICL multiclusters are available.

#### if this PR is a backport please specify the original PR:

Not a backport.

[1] http://hgcal.web.cern.ch/hgcal/

[2] http://tmengke.web.cern.ch/tmengke/HGCHitValidation_front/

[3] http://apsallid.web.cern.ch/apsallid/HGCValidmulticlusters_testing/

[4] http://apsallid.web.cern.ch/apsallid/HGCValidlayerclusters_testing/

PR #27478, opened 2019-07-10
[HGCal] Run HGCalValidator only on the hard scatterer event and correcting the mess from merging

apsallid:
#### PR description:

This PR addresses the issue #27417. We decided (@rovere , @felicepantaleo , @amartelli )
to follow PR #26915 and consider only CaloParticles coming from the hard scatterer, 
excluding the PU contribution. Running the Timing Service on 10 events of workflow 20234.99 
we see that the timing of the hgcalValidator module is on average 2.88 s/event. 

We should also note that the hgcalValidator module will for sure be even faster in cooperation 
with TICL multiclusters, since at the moment the number of multiclusters produced is huge 
(~10^4 multiclusters per event for wf 20234.99). 

Furthermore, we noted: 

- A duplication of a loop coming from the git rebasing/merging. 
- An incorrect initialization of the 3D score, shared energy and shared energy fraction inside the CaloParticles loop.
- An incorrect iteration over the full collection of multiclusters. At that point we should iterate only to the multiclusters that are related to CaloParticles. Iterating over the full multicluster collection contained cases for the score where the numerator and denominator had their initialized values,which is zero and lead to nan's in the histograms. 

#### PR validation:

We tested the PR with wf 20234.99 (TTbar with 200PU).

### DIFF
--- a/Validation/HGCalValidation/plugins/HGCalValidator.cc
+++ b/Validation/HGCalValidation/plugins/HGCalValidator.cc
@@ -192,6 +192,18 @@ void HGCalValidator::dqmAnalyze(const edm::Event& event,
     histoProducerAlgo_->fill_info_histos(histograms.histoProducerAlgo, totallayers_to_monitor_);
   }
 
+  //Consider CaloParticles coming from the hard scatterer, excluding the PU contribution.
+  std::vector<CaloParticle> caloParticlesFromHardScat; 
+  for ( auto& it_caloPart : caloParticles) {
+    if (it_caloPart.g4Tracks()[0].eventId().event() != 0 or it_caloPart.g4Tracks()[0].eventId().bunchCrossing() != 0) {
+      LogDebug("HGCalValidator") << "Excluding CaloParticles from event: "
+				 << it_caloPart.g4Tracks()[0].eventId().event()
+				 << " with BX: " << it_caloPart.g4Tracks()[0].eventId().bunchCrossing() << std::endl;
+      continue;
+    }
+    caloParticlesFromHardScat.push_back(it_caloPart);
+  }
+
   // ##############################################
   // fill caloparticles histograms
   // ##############################################
--- a/Validation/HGCalValidation/plugins/HGCalValidator.cc
+++ b/Validation/HGCalValidation/plugins/HGCalValidator.cc
@@ -192,17 +192,20 @@ void HGCalValidator::dqmAnalyze(const edm::Event& event,
     histoProducerAlgo_->fill_info_histos(histograms.histoProducerAlgo, totallayers_to_monitor_);
   }
 
-  //Consider CaloParticles coming from the hard scatterer, excluding the PU contribution.
-  // std::vector<CaloParticle*> caloParticlesFromHardScat;
-  std::vector<CaloParticle> caloParticlesFromHardScat;
-  for (const auto& it_caloPart : caloParticles) {
-    if (it_caloPart.g4Tracks()[0].eventId().event() != 0 or it_caloPart.g4Tracks()[0].eventId().bunchCrossing() != 0) {
+  auto nCaloParticles = caloParticles.size();
+  std::vector<size_t> cPIndices;
+  //Consider CaloParticles coming from the hard scatterer
+  //excluding the PU contribution and save the indices.
+  for (unsigned int cpId = 0; cpId < nCaloParticles; ++cpId) {
+    if (caloParticles[cpId].g4Tracks()[0].eventId().event() != 0 or
+        caloParticles[cpId].g4Tracks()[0].eventId().bunchCrossing() != 0) {
       LogDebug("HGCalValidator") << "Excluding CaloParticles from event: "
-                                 << it_caloPart.g4Tracks()[0].eventId().event()
-                                 << " with BX: " << it_caloPart.g4Tracks()[0].eventId().bunchCrossing() << std::endl;
+                                 << caloParticles[cpId].g4Tracks()[0].eventId().event()
+                                 << " with BX: " << caloParticles[cpId].g4Tracks()[0].eventId().bunchCrossing()
+                                 << std::endl;
       continue;
     }
-    caloParticlesFromHardScat.emplace_back(it_caloPart);
+    cPIndices.emplace_back(cpId);
   }
 
   // ##############################################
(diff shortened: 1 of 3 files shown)


jfernan2:
please test

Validation/HGCalValidation/plugins/HGCalValidator.cc
```diff
@@ -192,6 +192,18 @@ void HGCalValidator::dqmAnalyze(const edm::Event& event,
     histoProducerAlgo_->fill_info_histos(histograms.histoProducerAlgo, totallayers_to_monitor_);
   }
 
+  //Consider CaloParticles coming from the hard scatterer, excluding the PU contribution.
+  std::vector<CaloParticle> caloParticlesFromHardScat;
+  for (auto& it_caloPart : caloParticles) {
```
kpedro88: could be const auto?

Validation/HGCalValidation/plugins/HGCalValidator.cc
```diff
@@ -192,6 +192,18 @@ void HGCalValidator::dqmAnalyze(const edm::Event& event,
...
+    if (it_caloPart.g4Tracks()[0].eventId().event() != 0 or it_caloPart.g4Tracks()[0].eventId().bunchCrossing() != 0) {
+      LogDebug("HGCalValidator") << "Excluding CaloParticles from event: "
+                                 << it_caloPart.g4Tracks()[0].eventId().event()
+                                 << " with BX: " << it_caloPart.g4Tracks()[0].eventId().bunchCrossing() << std::endl;
+      continue;
+    }
+    caloParticlesFromHardScat.push_back(it_caloPart);
```
kpedro88: This is making a copy of each hard-scatter caloparticle. Some alternatives:
* store pointers to hard-scatter caloparticles
* store vector<size_t> with a list of hard-scatter indices, and then select only those indices from the original vector
Both of these alternatives have worse memory locality, but reduce memory usage and copying overhead. It would be interesting to see if these improve the performance any further.

jfernan2:
@apsallid from the point of view of DQM, this PR seems to modify 6 plots on folder HGCAL /HGCalValidator /hgcalMultiClusters    but they show NaN, do you understand why?

This is what I mean:   https://tinyurl.com/y54lv4m2

I also fail to find out when these plots were introduced since I don't find them in any RelVal sample, Do you have a reference to compare? Thanks

fabiocos:
please test workflow 20234.99

Validation/HGCalValidation/plugins/HGCalValidator.cc
apsallid: Hi @kpedro88 , 

Using pointers or the selected indices has far reaching consequences to the whole package since the input to the vast 
majority of the methods in Validation/HGCalValidation/src/HGVHistoProducerAlgo.cc should be changed and subsequently 
a lot of code. Could we please go on with the emplace_back?

apsallid:
Hi @jfernan2 , 

These plots were first introduced in #27337 (14 days ago) so CMSSW_11_0_X_2019-06-30-0000 after 
CMSSW_11_0_0_pre2 . i do not know if there are pre3 RelVals out there. For pre2 there are I think. 
For the plots you mention the nans i haven't seen them before. I have started a personal production 
with no PU events to investigate further. 
We are starting now to look at the PU samples and this the reason issue #27417 missed our attention. 
Many things run in parallel with TICL multiclusters and it will take some time until we fully understand 
the situation, especially with PU.

Validation/HGCalValidation/plugins/HGCalValidator.cc
kpedro88: It's fine for now, but if the memory usage becomes concerning, these alternatives should be revisited.

kpedro88:
please test workflow 20634.99

Validation/HGCalValidation/plugins/HGCalValidator.cc
rovere: @apsallid move the filter on the CaloParticles where you actually used them, i.e. one level down the stack.
In this way, you do not have to copy anything.

apsallid: @rovere i tested that but for some reason it is really slow. Maybe because we need that check in 5 different places while the code in these places will loop in any case through all CaloParticles which is of the order of 10000. By reducing the collection to ~200 CaloParticles once in the beginning we gain speed. I do not know what else to think.

rovere: @apsallid doesn't sound like a good explanation. Are you sure, by any chance, you left some part of the code w/o the selection?

rovere: @apsallid otherwise use the vector of indices down the stack, so that you don't need to loop over and over and make the very same selection.

jfernan2:
Hi @apsallid I am afraid there are no RelVals with your change made yet, so it would be great if you can investigate the NaN issue before approving this PR so that the plots be meaningful for you

apsallid:
@rovere , @kpedro88 I implemented the indices method and the timing of the hgcalValidator 
module is on average 4.31 s/event. I will update the number in the description also. 

@jfernan2 I will start a production with PU and try to understand the nan issue of these 6 histos.

jfernan2:
please test

apsallid:
@jfernan2  We understood the cause of the nan's and it should be fine now. I will update the description so that it is clear what was causing the problem as well as with the new timing average of the module, which is a little faster.

jfernan2:
please test

jfernan2:
+1
thanks @apalid you have indeed fixed the NaN
https://tinyurl.com/y3qukje9

fabiocos:
In a private test on wf 20634.99 I see in step4:
```
18:42 cmsdev25 2949> grep hgcalV step4_TTbar_14TeV+TTbar_14TeV_TuneCUETP8M1_2023D41PU_GenSimHLBeamSpotFull14INPUT+SingleNuE10_cf_2023D41PU_PremixHLBeamSpotFull14PU_Premix+DigiFullTriggerPUPRMXCombined_2023D41PU+RecoFullGlobalPUPRMX_2023D41PU+HARVESTFullGlobalPU_2023D41PU.log 
TimeModule> 106 1 hgcalValidator HGCalValidator 2.73505
TimeModule> 103 1 hgcalValidator HGCalValidator 3.22121
TimeModule> 104 1 hgcalValidator HGCalValidator 3.35191
TimeModule> 105 1 hgcalValidator HGCalValidator 3.08943
TimeModule> 107 1 hgcalValidator HGCalValidator 3.06279
TimeModule> 101 1 hgcalValidator HGCalValidator 2.84508
TimeModule> 108 1 hgcalValidator HGCalValidator 2.68395
TimeModule> 110 1 hgcalValidator HGCalValidator 3.53003
TimeModule> 102 1 hgcalValidator HGCalValidator 2.82613
```
So I would say that the original problem is fixed.

fabiocos:
+1

PR #27505, opened 2019-07-12
PFAlgo thread-safe logging

jpata:
#### PR description:

As suggested earlier in https://github.com/cms-sw/cmssw/pull/27153#issuecomment-501324699, I'm refactoring the logging in PFAlgo to be thread-safe. This is a first commit to let you know I'm working on this and to get initial feedback. I hope to revamp the whole PFAlgo.cc from the logging point of view (also removing useless comments etc) in the next few days.

#### PR validation:

WIP, please don't trigger tests yet on this PR.

### DIFF
--- a/RecoParticleFlow/PFProducer/src/PFAlgo.cc
+++ b/RecoParticleFlow/PFProducer/src/PFAlgo.cc
@@ -132,7 +132,7 @@ void PFAlgo::reconstructParticles(const reco::PFBlockHandle& blockHandle, PFEGam
   // reset output collection
   pfCandidates_->clear();
 
-  LogDebug("PFAlgo::reconstructParticles") << "start of function";
+  LogDebug("PFAlgo::reconstructParticles") << "start of function, blocks.size()=" << blocks.size();
 
   // sort elements in three lists:
   std::list<reco::PFBlockRef> hcalBlockRefs;
@@ -210,6 +210,8 @@ void PFAlgo::reconstructParticles(const reco::PFBlockHandle& blockHandle, PFEGam
   //Add Missing muons
   if (muonHandle_.isValid())
     pfmu_->addMissingMuons(muonHandle_, pfCandidates_.get());
+
+  LogDebug("PFAlgo::reconstructParticles") << "end of function, pfCandidates_->size()=" << pfCandidates_->size();
 }
 
 void PFAlgo::egammaFilters(const reco::PFBlockRef& blockref,
@@ -355,7 +357,7 @@ void PFAlgo::egammaFilters(const reco::PFBlockRef& blockref,
 }
 
 void PFAlgo::conversionAlgo(const edm::OwnVector<reco::PFBlockElement>& elements, std::vector<bool>& active) {
-  LogDebug("PFAlgo::conversionAlgo") << "start of function";
+  LogDebug("PFAlgo::conversionAlgo") << "start of function, elements.size()=" << elements.size();
   for (unsigned iEle = 0; iEle < elements.size(); iEle++) {
     PFBlockElement::Type type = elements[iEle].type();
     if (type == PFBlockElement::TRACK) {
@@ -391,8 +393,8 @@ bool PFAlgo::recoTracksNotHCAL(const reco::PFBlock& block,
                                unsigned int iTrack,
                                std::multimap<double, unsigned>& ecalElems,
                                reco::TrackRef& trackRef) {
-  LogDebug("PFAlgo::recoTracksNotHCAL") << "Now deals with tracks linked to no HCAL clusters. Was HCal active? "
-                                        << (!hasDeadHcal);
+  LogDebug("PFAlgo::recoTracksNotHCAL")
+      << "start of function, now dealing with tracks linked to no HCAL clusters. Was HCal active? " << (!hasDeadHcal);
   // vector<unsigned> elementIndices;
   // elementIndices.push_back(iTrack);
 
@@ -471,15 +473,15 @@ bool PFAlgo::recoTracksNotHCAL(const reco::PFBlock& block,
     unsigned nHits = elements[iTrack].trackRef()->hitPattern().trackerLayersWithMeasurement();
     unsigned int NLostHit = trackRef->hitPattern().trackerLayersWithoutMeasurement(HitPattern::TRACK_HITS);
 
-    if (debug_)
-      LogDebug("PFAlgo::recoTracksNotHCAL")
-          << "A track (algo = " << trackRef->algo() << ") with momentum " << trackMomentum << " / "
-          << elements[iTrack].trackRef()->pt() << " +/- " << Dpt << " / " << elements[iTrack].trackRef()->eta()
-          << " without any link to ECAL/HCAL and with " << nHits << " (" << NLostHit
-          << ") hits (lost hits) has been cleaned";
+    LogDebug("PFAlgo::recoTracksNotHCAL")
+        << "A track (algo = " << trackRef->algo() << ") with momentum " << trackMomentum << " / "
+        << elements[iTrack].trackRef()->pt() << " +/- " << Dpt << " / " << elements[iTrack].trackRef()->eta()
+        << " without any link to ECAL/HCAL and with " << nHits << " (" << NLostHit
+        << ") hits (lost hits) has been cleaned";
+
     active[iTrack] = false;
     return true;
-  }
+  }  //rejectTracks_Step45_ && ...
 
   tmpi.push_back(reconstructTrack(elements[iTrack]));
 
@@ -835,6 +837,7 @@ bool PFAlgo::recoTracksNotHCAL(const reco::PFBlock& block,
       (*pfCandidates_)[tmpi[ic]].addElementInBlock(blockref, kTrack[ic]);
     }
   }
+  LogDebug("PFAlgo::recoTracksNotHCAL") << "end of function";
   return false;
 }
 
@@ -845,7 +848,7 @@ void PFAlgo::elementLoop(const reco::PFBlock& block,
                          const reco::PFBlockRef& blockref,
                          ElementIndices& inds,
                          std::vector<bool>& deadArea) {
-  LogDebug("PFAlgo::elementLoop") << "start of function";
+  LogDebug("PFAlgo::elementLoop") << "start of function, elements.size()" << elements.size();
   for (unsigned iEle = 0; iEle < elements.size(); iEle++) {
     PFBlockElement::Type type = elements[iEle].type();
 
@@ -983,7 +986,6 @@ void PFAlgo::elementLoop(const reco::PFBlock& block,
     // ECAL cluster and an HCAL cluster, link the track to the HCAL cluster for
     // later analysis
     if (hcalElems.empty() && !ecalElems.empty() && !hasDeadHcal) {
-      // debug_ = true;
       unsigned ntt = 1;
       unsigned index = ecalElems.begin()->second;
       std::multimap<double, unsigned> sortedTracks;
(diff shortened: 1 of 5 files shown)


slava77:
@cmsbuild please test

perrotta:
@jpata , please let us know when this will not be a work in progress any more, so that we can start the tests and the review
(Maybe then remove "WIP" from the title)

jpata:
please test now :)

perrotta:
please test

RecoParticleFlow/PFProducer/src/PFAlgo.cc
```diff
@@ -1076,30 +1047,29 @@ void PFAlgo::elementLoop(const reco::PFBlock& block,
...
       unsigned index = ecal.second;
       // Sanity checks and optional printout
       PFBlockElement::Type type = elements[index].type();
-      if (debug_) {
-        double dist = ecal.first;
-        cout << "\telement " << elements[index] << " linked with distance = " << dist << endl;
-        if (!active[index])
-          cout << "This ECAL is already used - skip it" << endl;
-      }
+#ifdef EDM_ML_LOGDEBUG
```
perrotta: What is the scope of these directives? Since you are going to use them, why not to let the previous ```cout```'s, once protected by them?
Why there are both  ```#ifdef EDM_ML_LOGDEBUG``` and ```#ifdef EDM_ML_DEBUG```? Is it a typo?

perrotta (commented):
Furthermore: the ```debug_``` private variable is not needed any more, and can be removed from both the header file and from the constructor here in the implementation.

RecoParticleFlow/PFProducer/src/PFAlgo.cc
jpata: my bad, should be `EDM_ML_DEBUG`

slava77:
@cmsbuild please test

RecoParticleFlow/PFProducer/src/PFCandConnector.cc
```diff
@@ -72,74 +72,70 @@ reco::PFCandidateCollection PFCandConnector::connect(PFCandidateCollection& pfCa
...
-      cout << "====================         Cand Connector         ===============" << endl;
-      cout << "==================== ------------------------------ ===============" << endl;
-      cout << "====================   \tfor " << pfCand.size() << " Candidates\t  =============" << endl;
-      cout << "==================== primary calibrated " << bCalibPrimary_ << "     =============" << endl;
-    }
+    LogDebug("PFCandConnector::connect") << "pfCand.size()=" << pfCand.size() << "bCalibPrimary_=" << bCalibPrimary_;
 
     for (unsigned int ce1 = 0; ce1 < pfCand.size(); ++ce1) {
       if (isPrimaryNucl(pfCand.at(ce1))) {
         if (debug_)
```
perrotta: This "debug_" should also get removed.
Then, if you do so, you can further modify the interface of PFCandConnector, and remove such "debug_" from the class members, as well as from the constructor (PFCandConnector is only accessed by PFAlgo, therefore you don't have to propagate this change anywhere else)

jpata: good point, thanks. done.

jpata:
Ok, should be all now! :) Please test again.

perrotta:
please test

perrotta:
I think this PR is ready to get approved.
Please @hatakeyamak and/or @bendavid confirm that the move to LogDebug everywhere is ok with you, and then I will sign (as soon as the automatic tests will have completed)

hatakeyamak:
I think this is good which makes debug printing thread-safe.
Just for my education, is there a known downside of moving from cout to LogDebug?

perrotta:
> I think this is good which makes debug printing thread-safe. I knew @jpata is working on it.
> Just for my education, I don't know any, but is there a known downside of moving from cout to LogDebug, which we should keep in mind?

No downside. Since this is your realm, I wondered whether you wanted to propose some different LogSomething instead (e.g. LogInfo): for me LogDebug is perfectly fine.

hatakeyamak:
Okay.

@jpata @perrotta do you have example output of logdebug messages with this PR when LogDebug messages are enabled (just for these modules preferably)? It will be nice to see how debugging actually looks together with %MSG-i etc lines.

LogDebug add endl automatically afaik. If so, some endl had better be removed? For example, lines 578, 971 of PFAlgo.cc and lines 83, 117, etc of PFCandConnector.cc? Is it intentional to keep it?

jpata:
Here is a sample log output: [debugs2.log](https://github.com/cms-sw/cmssw/files/3407490/debugs2.log)

To enable debug logging, the following needs to be added to the config:
 ```python
process.MessageLogger.debugModules = cms.untracked.vstring(["particleFlowTmp"])
process.MessageLogger.debugs = cms.untracked.PSet(threshold = cms.untracked.string('DEBUG'))
```

and the code must be compiled with
```bash
scram b -j8 USER_CXXFLAGS="-DEDM_ML_DEBUG"
```

The additional `endl` are a remnant of the old code and not intentional - I can remove them.

hatakeyamak:
Thank you. Why do I see only messages from PFCandConnector and not from PFAlgo?

jpata:
Right, this was the first 10k lines, here is the full file for one event (6 MB): [debugs.log](https://github.com/cms-sw/cmssw/files/3407556/debugs.log)

hatakeyamak:
@jpata Thank you. The large number of %MSG-d lines in additions to the real message lines makes the printouts a little hard to read, but I guess we cannot do much about it.

When threshold = cms.untracked.string('DEBUG') is set, it looks like we also get many many %MSG-i messages from other modules too. Is there a way to get only %MSG-d from particleFlowTmp in a separate output than other messages from MSG-i? If so, it will be very helpful. If not, I guess we have to live with it, though I find it a little painful...

jpata:
@hatakeyamak `process.MessageLogger.debugModules = cms.untracked.vstring(["particleFlowTmp"])` should enable logging only for the particle flow module, according to the docs. It takes a bit of time to configure MessageLogger correctly but I believe it should be possible to redirect all PFAlgo LogDebug to a separate file. I also don't like the very verbose printout but this is what we have in CMSSW :)

perrotta:
+1
- Debug output of PFAlgo moved from cout to the thread safe LogDebug
- Jenkins tests pass and show no differences, as expected

hatakeyamak:
> but I believe it should be possible to redirect all PFAlgo LogDebug to a separate file. I also don't like the very verbose printout but this is what we have in CMSSW :)

If you can figure that out, it will be great to make this LogDebug outputs more useful.

RecoParticleFlow/PFProducer/src/PFAlgo.cc
```diff
@@ -171,18 +166,16 @@ void PFAlgo::reconstructParticles(const reco::PFBlockHandle& blockHandle, PFEGam
     }
   }  //loop blocks
 
-  if (debug_) {
-    cout << "# Ecal blocks: " << ecalBlockRefs.size() << ", # Hcal blocks: " << hcalBlockRefs.size()
-         << ", # HO blocks: " << hoBlockRefs.size() << ", # Other blocks: " << otherBlockRefs.size() << endl;
-  }
+  LogDebug("PFAlgo::reconstructParticles")
```
slava77: were you able to select on categories with ```::``` in the message logger?
There are some special characters that can not be a part of the category and I'm not sure if colons are among them or now.

jpata: In the end I turned PFAlgo on or off just through particleFlowTmp. I used `grep` later to filter on a specific function.

jpata:
@hatakeyamak if we want, we can change LogDebug to LogTrace, which does not attach the %MSG, but also does not attach any line numbering or such, which I think are useful. I would prefer to keep LogDebug, as the extra info can be removed with grep. What do you think?

Reading through the CMSSW MessageLogger [docs](https://twiki.cern.ch/twiki/bin/view/CMSPublic/SWGuideMessageLoggerSetupExamples#Suppressing_Messages_From_Specif), I don't find a way to enable debug-level (LogDebug or LogTrace) just for one module, but maybe I miss the magic commands. 

Let me know how you want to take this PR forward :)

hatakeyamak:
Thank you for the information.

It looks like, with your setup, LogDebug messages only come form particleFlowTmp. but I cannot find a way to suppress so many LogInfo messages from other modules, which I find rather unfortunate.

Thank you for info about LogTrace/LogVerbatim/LogPrint flavor. Now knowing more about various options, completely being honest with you, I think I prefer using this LogPrint and turn on/off these LogPrints as done before using debug flag in PFAlgo. These couts were very carefully formatted by original authors, and I think it's nice to keep it using LogTrace/LogVerbatim/LogPrint flavor (I understand they care be removed by grep), and I find on/off of LogDebug/LogInfo/LogDebug/LogVerbatim not very flexible (and we end up provoking many extra messages from other modules).

So, this is my personal preference (basically using LogPrint as thread-safe cout in a similar way as before), but I will not insist too strongly.

jpata:
Sorry, I'm confused by your last message. Are you sure you want LogPrint for debugging messages?

I quote:
`LogPrint is identical in all ways to LogWarning, except that absolutely no message formatting is performed, and likewise for LogProblem and LogImportant relative to LogError.`

Probably we don't want to print low-level PFAlgo debug info at a WARNING threshold. The options I see are:

1) keep `cout` and `debug_` printing: output always in stdout, not thread safe
2) Use LogDebug: verbose, thread-safe, output can be in a file or stdout - all modules compiled with debug flags will print. With debugging disabled, these logging messages are not even compiled.
3) Use LogTrace: like LogDebug, but no formatting.
4) LogInfo: like LogDebug, but logging code will always be compiled into cmssw, resulting in boolean switches at runtime (as with cout)
5) LogTrace: like LogInfo, but no formatting

So the real question to me is if and how MessageLogger outputs can be put into different files on a per-module basis. So far I have not found a way, but maybe I'm not clever enough to understand MessageLogger

Any suggestions? @Dr15Jones

perrotta:
I wrote my "+1" when I tought I understood that a decision was taken to stick on LogDebug. I'm removing it now, and I will resign once you have converged  on any of them

jpata:
Sure. I also thought it's all good, until I got the last message.

jpata:
How about this? Switched to LogTrace, filter according to categories.

[debugs.log](https://github.com/cms-sw/cmssw/files/3413381/debugs.log)

```python
process.MessageLogger = cms.Service("MessageLogger"
...
    categories = cms.untracked.vstring(
        'FwkJob',
        'FwkReport',
        'FwkSummary',
        'Root_NoDictionary',
        'PFAlgo',
        'PFCandConnector'
    ),
    debugModules = cms.untracked.vstring("particleFlowTmp"),
    debugs = cms.untracked.PSet(
         INFO =  cms.untracked.PSet(limit = cms.untracked.int32(0)),
         DEBUG   = cms.untracked.PSet(limit = cms.untracked.int32(0)),
         PFAlgo = cms.untracked.PSet(limit = cms.untracked.int32(-1)),
         PFCandConnector = cms.untracked.PSet(limit = cms.untracked.int32(-1)),
         threshold = cms.untracked.string('DEBUG')
    ),
...
)
```

Dr15Jones:
@jpata you appear to have answered the question. As far as I can tell from reading the MessageLogger twiki pages, what you have should do what you want.

perrotta:
please test

hatakeyamak:
Thank you @jpata. It looks like you found a nice way to control and sort out log outputs from different modules. I have small minor suggestion (will add inline), but it looks good to me overall. Thank you for looking into this.

perrotta:
> Thank you @jpata. It looks like you found a nice way to control and sort out log outputs from different modules. I have small minor suggestion (will add inline), but it looks good to me overall. Thank you for looking into this.

@hatakeyamak do I understand correctly that you are sending a few comments related to this PR, and therefore we should wait for them and  the following further updates before signing it?

hatakeyamak:
I already did send one above, but it's so minor. So, I left it to @jpata to implement it or not. I am ok even if the current PR goes as is.

perrotta:
+1
- Debug output of PFAlgo moved from cout to the thread safe LogTrace: the format of the Log was agreed after a discussion with the PF convener
- Jenkins tests pass and show no differences, as expected
- (I will re-sign in case of further aestetic updates)

fabiocos:
+1

Issue #27393, opened 2019-06-28
Developers overriding production location for retrieving conditions in configurations


davidlange6:
Breaks centrally determined location for retrieving conditions (which of course can change even if it rarely does). All of these should rely on the global tag to determine from where the conditions are to be taken (or they should not be loaded into production workflows)

https://github.com/cms-sw/cmssw/blob/02d4198c0b6615287fd88e9a8ff650aea994412e/L1Trigger/L1TMuon/python/simGmtStage2Digis_cfi.py#L23

https://github.com/cms-sw/cmssw/blob/82c4a527471bca6fc4897fd4feb3f7d7fcb73928/RecoTauTag/RecoTau/python/tools/runTauIdMVA.py#L41

https://github.com/cms-sw/cmssw/blob/8587c76d0a2ee041152b0c0ac289d851a13ef247/L1Trigger/L1TCalorimeter/python/simDigis_cff.py#L49

davidlange6:
assign l1,db,reco

davidlange6:
assign reconstruction

ggovi:
My personal opinion is that a cfg file containing a Condition Customization ( defined with one or more toGet statements ) should be NEVER invoked ( directly or included ) in a production workflow, since:
- it overrides the mapping record/tag provided by the official Global Tag, referencing some Tag that could stay outside the control ( in terms of testing and validation ) 
- it potentially breaks reproducibility
For this reason, I think that the first measure required is to find out the reason of the concerned customisation, and remove them from the production workflows.
This will potentially imply:
- to replace some tag in existing GTs (?)
- to create additional GTs (?)
I think AlCa people should be involved in this review procedure

slava77:
for cmssw/RecoTauTag/RecoTau/python/tools/runTauIdMVA.py
@mbluj @steggema @swozniewski please comment and perhaps follow up with a resolution

fabiocos:
for the L1Trigger instances: @rekovic could you please address this issue?

ggovi:
Can we add alca in this loop?

kpedro88:
assign alca

mbluj:
Hello,
comments on https://github.com/cms-sw/cmssw/blob/82c4a527471bca6fc4897fd4feb3f7d7fcb73928/RecoTauTag/RecoTau/python/tools/runTauIdMVA.py#L41
[...]

PR #27516, opened 2019-07-13
[RFC]  Phase-2 InnerTracker conditions from DB (version 1)

mmusich:
#### PR description:
Upcoming developments in the Phase-2 Inner Tracker digitization and local reconstruction require access to several conditions related to the CPE, namely per-module Pixel Lorentz Angle (both for simulation and reconstruction) and in future also Pixel 1D, 2D Templates and Generic Errors.
Currently only the generic CPE algorithm is used for the Inner Tracker local reconstruction (hence only the Lorentz Angle is accessed) and it is is provided via a fake conditions `ESSource`:  [SiPixelFakeLorentzAngleESSource](https://github.com/cms-sw/cmssw/blob/master/CalibTracker/SiPixelESProducers/plugins/SiPixelFakeLorentzAngleESSource.cc).  This `ESSource` makes use of the `DetId` list from a “Skimmed Geometry” external file: one DetId list for each geometry ([example](https://github.com/cms-sw/cmssw/blob/master/SLHCUpgradeSimulations/Geometry/python/fakeConditions_phase2TkT14_cff.py#L11)) and only one single LA (μH =0.106) for all IT modules is used ([hardcoded](https://github.com/cms-sw/cmssw/blob/master/CalibTracker/SiPixelESProducers/plugins/SiPixelFakeLorentzAngleESSource.cc#L53)).
Moreover the `“forWidth” ` label is not taken from DB  (see [PixelCPEGeneric_cfi ](https://github.com/cms-sw/cmssw/blob/master/RecoLocalTracker/SiPixelRecHits/python/PixelCPEGeneric_cfi.py#L72) configuration) hence no correction is applied, and the simulation LA value is passed as a configuration parameter in the digitizer (see [phase2TrackerDigitizer_cfi.](https://github.com/cms-sw/cmssw/blob/master/SimTracker/SiPhase2Digitizer/python/phase2TrackerDigitizer_cfi.py#L50-L51)).
Finally the Generic Errors are also not taken from DB (see [PixelCPEGeneric_cfi](https://github.com/cms-sw/cmssw/blob/master/RecoLocalTracker/SiPixelRecHits/python/PixelCPEGeneric_cfi.py#L72)) but are  hard-coded in C++ (see [PixelCPEGeneric.cc](https://github.com/cms-sw/cmssw/blob/master/RecoLocalTracker/SiPixelRecHits/src/PixelCPEGeneric.cc#L114)).
This PR is meant to change part of this situation, switching on the reading of conditions from Database where payloads are available, and provide such payloads in the production DB.
The implementation is complicated by the co-existence in release of several geometries, with different `DetId`s lists, and all need to be supported, hence the needed conditions cannot simply be added to the `phase2_realistic` autoCond Global Tag key to satisfy all use-cases. 
In this PR we propose a solution based on an additional `ESSource` configured differently for each Tracker `T*` geometry via the `fakeConditions_phase2TkT*_cff.py` files (already included in the geometry definition).  
For the D45 geometry a new `SLHCUpgradeSimulations.Geometry.fakeConditions_phase2TkT15_cff` file is created, in order to allow for a different set of LA values (μH =0.053) to be applied (T14 geometry is supposed to stay with the current value μH =0.106).
A complementary approach, based on the dynamical creation of more autoCond keys one for each of the Tracker geometry, has been implemented in PR #27517, though **we prefer this implementation as it automatically binds the conditions to the geometry**. 
More information has been collected in [this presentation](https://indico.cern.ch/event/832785/contributions/3492905/attachments/1876286/3089720/ts_mm_AlCaDBMeeting_8Jul2019.pdf) at the AlCa/DB meeting of 7th July 2019.
This PR is RFC and in draft state as it is at odds with the purpose of issue https://github.com/cms-sw/cmssw/issues/27393.

#### PR validation:
The PR was tested running 1 event of SingleMuPt10 for all the currently supported phase-2 geometry combinations: D35 (T6), D41 (T14), D43 (T14), D44 (T14), D45 (T5) via:
```
runTheMatrix.py --what upgrade -l 20007.0,20407.0,20807.0,21207.0,21607.0 --command='-n 1' -j 4
```

This PR is not a backport
cc:
@tsusa @emiglior @pwittich @christopheralanwest @tlampen @ggovi

### DIFF
--- a/Configuration/Geometry/python/GeometryExtended2023D45_cff.py
+++ b/Configuration/Geometry/python/GeometryExtended2023D45_cff.py
@@ -5,7 +5,7 @@ import FWCore.ParameterSet.Config as cms
 
 from Geometry.CMSCommonData.cmsExtendedGeometry2023D45XML_cfi import *
 from Geometry.TrackerNumberingBuilder.trackerNumberingGeometry_cfi import *
-from SLHCUpgradeSimulations.Geometry.fakeConditions_phase2TkT14_cff import *
+from SLHCUpgradeSimulations.Geometry.fakeConditions_phase2TkT15_cff import *
 from Geometry.HcalCommonData.hcalParameters_cfi      import *
 from Geometry.HcalCommonData.hcalDDDSimConstants_cfi import *
 from Geometry.HGCalCommonData.hgcalParametersInitialization_cfi import *
--- a/RecoLocalTracker/SiPixelRecHits/python/PixelCPEGeneric_cfi.py
+++ b/RecoLocalTracker/SiPixelRecHits/python/PixelCPEGeneric_cfi.py
@@ -68,7 +68,7 @@ run3_common.toModify(PixelCPEGenericESProducer, IrradiationBiasCorrection = True
 # FIXME::Is the Upgrade variable actually used?
 from Configuration.Eras.Modifier_phase2_tracker_cff import phase2_tracker
 phase2_tracker.toModify(PixelCPEGenericESProducer, 
-  useLAWidthFromDB = False,
+  #useLAWidthFromDB = False,
   UseErrorsFromTemplates = False,
   LoadTemplatesFromDB = False,
   TruncatePixelCharge = False,
--- a/RecoLocalTracker/SiPixelRecHits/python/PixelCPEGeneric_cfi.py
+++ b/RecoLocalTracker/SiPixelRecHits/python/PixelCPEGeneric_cfi.py
@@ -68,7 +68,6 @@ run3_common.toModify(PixelCPEGenericESProducer, IrradiationBiasCorrection = True
 # FIXME::Is the Upgrade variable actually used?
 from Configuration.Eras.Modifier_phase2_tracker_cff import phase2_tracker
 phase2_tracker.toModify(PixelCPEGenericESProducer, 
-  #useLAWidthFromDB = False,
   UseErrorsFromTemplates = False,
   LoadTemplatesFromDB = False,
   TruncatePixelCharge = False,
--- a/SLHCUpgradeSimulations/Configuration/python/loadInnerTrackerConditionFromDB.py
+++ b/SLHCUpgradeSimulations/Configuration/python/loadInnerTrackerConditionFromDB.py
@@ -20,7 +20,7 @@ loadPhase2InneTrackerConditions = cms.ESSource( "PoolDBESSource",
 #%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
 class DBConfiguration():
     '''Helper class to store the configuration object DB'''
-    def __init__(self,vGeometry=None,vLA=None,vLAwidth=None,vSimLA=None,vGenError=None,vTemplate1D=None,vTemplate2Dnum=None,vTemplate2Dden=None):
+    def __init__(self,vGeometry=None, vLA=None, vLAwidth=None, vSimLA=None, vGenError=None, vTemplate1D=None, vTemplate2Dnum=None, vTemplate2Dden=None):
         self._vGeometry      = vGeometry
         self._vLA            = vLA 
         self._vLAwidth       = vLAwidth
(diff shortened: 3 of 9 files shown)


Configuration/Geometry/python/GeometryExtended2023D45_cff.py
```diff
@@ -5,7 +5,7 @@
 
 from Geometry.CMSCommonData.cmsExtendedGeometry2023D45XML_cfi import *
 from Geometry.TrackerNumberingBuilder.trackerNumberingGeometry_cfi import *
-from SLHCUpgradeSimulations.Geometry.fakeConditions_phase2TkT14_cff import *
+from SLHCUpgradeSimulations.Geometry.fakeConditions_phase2TkT15_cff import *
```
kpedro88: Please note that directly changing this file is forbidden. The changes should be propagated into [dict2023Geometry.py](https://github.com/cms-sw/cmssw/blob/master/Configuration/Geometry/python/dict2023Geometry.py) and the corresponding script [generate2023Geometry.py](https://github.com/cms-sw/cmssw/blob/master/Configuration/Geometry/scripts/generate2023Geometry.py) should be run to regenerate the geometry files.

kpedro88 (commented):
As mentioned in #27517, I prefer version 2 to version 1. The setup of the ESSource and ESPrefer here seems overcomplicated and not so easy to understand/maintain.

Configuration/Geometry/python/GeometryExtended2023D45_cff.py
mmusich: Hi @kpedro88 I am not familiar with the procedure for updating these files. 
For my own education why is it forbidden to change it? 
Also it is not clear to me what does it mean to regenerate the geometry files? Do you have instructions on how to run this script? 
By the way I am not changing the geometry, just providing a different set of conditions.

kpedro88: @mmusich because you have edited this file by hand, if someone were to add a new phase 2 detector version using the T15 tracker, they would not pick up this change and would get incorrect results. In fact, there is a unit test that enforces this, so this PR would not pass the test. The file also states at the top:
https://github.com/cms-sw/cmssw/blob/4590f5d4ba0a20f4fc13295e2d376820cab49d9e/Configuration/Geometry/python/GeometryExtended2023D45_cff.py#L3-L4

The generate2023Geometry.py script generates the following files for each phase 2 detector version:
https://github.com/cms-sw/cmssw/blob/f46916dae7609d7808eb6b9d4c9a996c5dcf981b/Configuration/Geometry/scripts/generate2023Geometry.py#L39-L41

It does this using the information in dict2023Geometry.py, to maintain the consistency. (If you checkout the packages Configuration/Geometry and Geometry/CMSCommonData in advance, it will even put the files in the right places, so you just have to `git add` them and then commit.)

slava77:
@mmusich 
please clarify on the status of this PR

slava77:
I noticed that #27517 is not marked as a draft anymore.
Perhaps it's time to close this PR.
@mmusich please check/close.
Thank you.

Configuration/Geometry/python/GeometryExtended2023D45_cff.py
mmusich: @kpedro88,  so just for my understanding, the correct way to change anything in a phase-2 workflow is to create a new an entirely new `D*` version by using the scripts you point above?

mmusich:
@slava77 I am still not entirely sure that the approach proposed in  #27517 is completely OK with PPD (see pending question at https://github.com/cms-sw/cmssw/pull/27517#issuecomment-516595542), so if you don't mind I would prefer to keep this draft open until that gets resolved.

Configuration/Geometry/python/GeometryExtended2023D45_cff.py
kpedro88: If the actual detector geometry is changed, indeed a new subdetector version should be added, and then a new D* full detector version that uses the new subdetector.

In this case, you're not changing the geometry, but adding some related information. New D* versions are not necessary in that case, but updating the component files is still not allowed. Only [dict2026Geometry.py](https://github.com/cms-sw/cmssw/blob/master/Configuration/Geometry/python/dict2026Geometry.py) can be updated, and then you have to rerun [generate2026Geometry.py](https://github.com/cms-sw/cmssw/blob/master/Configuration/Geometry/scripts/generate2026Geometry.py) with the existing D* versions to update the generated files. (While this is not the most automatic procedure, it was the only agreement that could be made between all relevant focus areas to keep things consistent, and it is enforced by a unit test.)

slava77:
-1

this is apparently superseded by #27517 (which is close to be fully signed now)

fabiocos:
@mmusich given that we have converged on the solution #27517 I guess that this proposal is obsolete and can be closed

mmusich:
closing in favor of #27517 (just merged)

PR #27517, opened 2019-07-13
Phase-2 InnerTracker conditions from DB (version 2)

mmusich:
#### PR description:
Upcoming developments in the Phase-2 Inner Tracker digitization and local reconstruction require access to several conditions related to the CPE, namely per-module Pixel Lorentz Angle (both for simulation and reconstruction) and in future also Pixel 1D, 2D Templates and Generic Errors.
Currently only the generic CPE algorithm is used for the Inner Tracker local reconstruction (hence only the Lorentz Angle is accessed) and it is is provided via a fake conditions `ESSource`:  [SiPixelFakeLorentzAngleESSource](https://github.com/cms-sw/cmssw/blob/master/CalibTracker/SiPixelESProducers/plugins/SiPixelFakeLorentzAngleESSource.cc).  This `ESSource` makes use of the `DetId` list from a “Skimmed Geometry” external file: one DetId list for each geometry ([example](https://github.com/cms-sw/cmssw/blob/master/SLHCUpgradeSimulations/Geometry/python/fakeConditions_phase2TkT14_cff.py#L11)) and only one single LA (μH =0.106) for all IT modules is used ([hardcoded](https://github.com/cms-sw/cmssw/blob/master/CalibTracker/SiPixelESProducers/plugins/SiPixelFakeLorentzAngleESSource.cc#L53)).
Moreover the `“forWidth” ` label is not taken from DB  (see [PixelCPEGeneric_cfi ](https://github.com/cms-sw/cmssw/blob/master/RecoLocalTracker/SiPixelRecHits/python/PixelCPEGeneric_cfi.py#L72) configuration) hence no correction is applied, and the simulation LA value is passed as a configuration parameter in the digitizer (see [phase2TrackerDigitizer_cfi.](https://github.com/cms-sw/cmssw/blob/master/SimTracker/SiPhase2Digitizer/python/phase2TrackerDigitizer_cfi.py#L50-L51)).
Finally the Generic Errors are also not taken from DB (see [PixelCPEGeneric_cfi](https://github.com/cms-sw/cmssw/blob/master/RecoLocalTracker/SiPixelRecHits/python/PixelCPEGeneric_cfi.py#L72)) but are  hard-coded in C++ (see [PixelCPEGeneric.cc](https://github.com/cms-sw/cmssw/blob/master/RecoLocalTracker/SiPixelRecHits/src/PixelCPEGeneric.cc#L114)).
This PR is meant to change part of this situation, switching on the reading of conditions from Database where payloads are available, and provide such payloads in the production DB.
The implementation is complicated by the co-existence in release of several geometries, with different `DetId`s lists, and all need to be supported, hence the needed conditions cannot simply be added to the `phase2_realistic` autoCond Global Tag key to satisfy all use-cases. 
In this PR we propose a solution based on the dynamical creation of new autoCond Global Tag keys, one for each of the geometries to be supported, by appending the CPE-related conditions in a similar fashion to what already happens for the HLT GT keys in [autoCondHLT](https://github.com/cms-sw/cmssw/blob/master/Configuration/HLT/python/autoCondHLT.py).
A complementary approach based on an additional `ESSource` configured differently for each Tracker `T*` geometry via the `fakeConditions_phase2TkT*_cff.py` files (already included in the geometry definition) has been implemented  in PR #27516 .
More information has been collected in [this presentation](https://indico.cern.ch/event/832785/contributions/3492905/attachments/1876286/3089720/ts_mm_AlCaDBMeeting_8Jul2019.pdf) at the AlCa/DB meeting of 7th July 2019.
Please notice that a different set of LA values (μH =0.053) is be applied to the T15 geometry with respect to the current value (μH =0.106), hence changes are expected in D45-related workflows.
This PR is RFC and in draft state as it is at odds with the purpose of issue https://github.com/cms-sw/cmssw/issues/27393.

#### PR validation:
The PR was tested running 1 event of SingleMuPt10 for all the currently supported phase-2 geometry combinations: D35 (T6), D41 (T14), D43 (T14), D44 (T14), D45 (T15), D46 (T15)  via:
```
runTheMatrix.py --what upgrade -l 20007.0,20407.0,20807.0,21207.0,21607.0,22007.0 --command='-n 1' -j 4
```

This PR is not a backport
cc:
@tsusa @emiglior @pwittich @christopheralanwest @tlampen @ggovi


Configuration/AlCa/python/autoCondPhase2.py
```diff
@@ -0,0 +1,70 @@
+# default value for all customizations
+connectionString = "frontier://FrontierProd/CMS_CONDITIONS"
+
+SiPixelLARecord           =   "SiPixelLorentzAngleRcd"                  
+SiPixelSimLARecord        =   "SiPixelLorentzAngleSimRcd"   
+SiPixelGenErrorRecord     =   "SiPixelGenErrorDBObjectRcd"       
+SiPixelTemplatesRecord    =   "SiPixelTemplateDBObjectRcd"       
+SiPixel2DTemplatesRecord  =   "SiPixel2DTemplateDBObjectRcd"     
+
+LA = {
```
kpedro88: These individual dicts could be combined into a single dict of dicts, with a list of actively used keys. Then the `phase2GTs` dict could be filled automatically from the active list. (The list of active subdet names could also be provided for this purpose.)

Configuration/AlCa/python/autoCond.py
```diff
@@ -86,6 +86,9 @@
 from Configuration.HLT.autoCondHLT import autoCondHLT
 autoCond = autoCondHLT(autoCond)
 
+from Configuration.AlCa.autoCondPhase2 import autoCondPhase2
+autoCond = autoCondPhase2(autoCond)
```
kpedro88: an explanatory comment should be added here indicating that autoCond keys "phase2_realistic_T*" are provided

kpedro88 (commented):
This solution seems more parsimonious and maintainable than version 1.

Ultimately, we may want to move to labeled conditions in a single phase 2 GT, but this seems like a useful intermediate step.

It would be good to integrate this more with the phase 2 geometry/workflow scripts ([dict2023Geometry.py](https://github.com/cms-sw/cmssw/blob/master/Configuration/Geometry/python/dict2023Geometry.py), [generate2023Geometry.py](https://github.com/cms-sw/cmssw/blob/master/Configuration/Geometry/scripts/generate2023Geometry.py)) in order to automate the relationship of a GT with a geometry scenario and corresponding workflow. For now, this could be done in the same way the suggestion of the Era modifiers is done (a printed message when the geometry is generated, based on a field in the subdetector components dictionary).

Configuration/AlCa/python/autoCondPhase2.py
mmusich: For our case, the nice thing of keeping the dictionaries separate is that at the moment we are not yet using the full list of possible customizable records, but only the Lorentz Angle related ones. So one can piece by piece increase the number of "customizations" by adding more payload to each of the  `phase2GTs` keys. 
Did I miss your point? Would you care to suggest the structure of this dict of dicts?

kpedro88: This is covered by part of my suggestion: "a list of actively used keys". For example:
```python
activeKeys = ["LA","LAWidth","SimLA"]
activeDets = ["T6","T14","T15"]
phase2GTs = {}
for det in activeDets:
    appendedTags = ()
    for key in activeKeys:
       appendedTags += allTags[key][det]
    phase2GTs["phase2_realistic_"+det] = ('phase2_realistic', appendedTags)
```

slava77:
@mmusich 
please clarify on the status of this PR

christopheralanwest:
I don't know that these symbolic GTs are supported in official MC production; I've asked PdmV experts for their input. Is it intended that the samples generated using the symbolic GTs are only used in local tests?

mmusich:
> I don't know that these symbolic GTs are supported in official MC production.

These GTs can be supported in official MC production, whatever is run in IB tests can also be run in official relval productions.

> Is it intended that the samples generated using the symbolic GTs are only used in local tests?

no. they are meant for the general public.

christopheralanwest:
please test

christopheralanwest:
please test workflow 20007.0,20407.0,20807.0,21207.0,21607.0

slava77:
@mmusich 
please clarify if you are selecting this option for the IT conditions.
If so, perhaps it's time to change from the draft mode to the normal one.
Thank you.

mmusich:
@slava77 I am sorry for the late reply on this. I would still want to take Kevin's comment: https://github.com/cms-sw/cmssw/pull/27517#discussion_r304009041, but I haven't had time so far. I'll try to update the branch as soon as possible.

Configuration/AlCa/python/autoCond.py
mmusich: done

Configuration/AlCa/python/autoCondPhase2.py
mmusich: done few lines below.

christopheralanwest:
please test

Configuration/AlCa/python/autoCondPhase2.py
```diff
@@ -0,0 +1,86 @@
...
+}
+
+Template2Dden = {
+    'T6'  : ( ','.join( [ 'SiPixel2DTemplateDBObject_phase2_T6_v0_den'  ,SiPixel2DTemplatesRecord,connectionString, "denominator", "2019-07-15 12:00:00.000"] ), ),
+    'T14' : ( ','.join( [ 'SiPixel2DTemplateDBObject_phase2_T14_v0_den' ,SiPixel2DTemplatesRecord,connectionString, "denominator", "2019-07-15 12:00:00.000"] ), ),
+    'T15' : ( ','.join( [ 'SiPixel2DTemplateDBObject_phase2_T15_v0_den' ,SiPixel2DTemplatesRecord,connectionString, "denominator", "2019-07-15 12:00:00.000"] ), ),
+}
+
+#combines in a single dict of dict the tags defined above
+allTags={
```
kpedro88: would be a bit more concise to define `allTags = {}` at the top and then directly assign `allTags["LA"] = {...}` etc. (eliminating individual variables entirely)

kpedro88:
please test

christopheralanwest:
please test workflow 20007.0,20407.0,20807.0,21207.0,21607.0

christopheralanwest:
In a recent meeting, PdmV experts made a verbal statement that symbolic GTs such as these are supported by the MC submission infrastructure. Since there are no datasets that have a name that contains "auto" or any of the autoCond keys (aside from cases where the string corresponding to the key is a substring of the value), it seems that this functionality has never been used. I would like PdmV experts (attn: @prebello @zhenhu @pgunnell) to explicitly confirm here that symbolic GTs are supported prior to signing off for AlCa.

mmusich:
The last commit removes the fake conditions `ESProducers` rendered unnecessary by the availability of them in the symbolic Global Tag.

christopheralanwest:
please test

mmusich:
>     /build/cmsbld/jenkins/workspace/compare-root-files-short-matrix/results/JR-comparison/PR-e83448/22034.0_TTbar_14TeV+TTbar_14TeV_TuneCUETP8M1_2026D46_GenSimHLBeamSpotFull14+DigiFullTrigger_2026D46+RecoFullGlobal_2026D46+HARVESTFullGlobal_2026D46
> 
> Comparison Summary:
> 
>     You potentially added 114107 lines to the logs

OK, apparently now there is also a D46 which needs to be taken care of. Will fix in the next commit.

slava77:
@mmusich 
this is no longer marked as a draft PR.
Perhaps it's time to change the title to drop "[RFC]". Please edit the title, if you agree.
Thank you.

mmusich:
Rebased on top of `CMSSW_11_0_X_2019-07-31-1100`, in order to have also `D46` (introduced in https://github.com/cms-sw/cmssw/pull/27555 ) and raken care of it (contains `T15`).

kpedro88:
please test
thanks @mmusich !

mmusich:
> please test
> thanks @mmusich !

@smuzaffar is the bot ignoring commands?

smuzaffar:
bot is waiting for code-checks results which ran fine ( https://cmssdt.cern.ch/jenkins/job/run-pr-code-checks/11214/console ) but for some unknown reason the comment was not propogated to github. I have force run the check now.

mmusich:
> bot is waiting for code-checks results which ran fine ( https://cmssdt.cern.ch/jenkins/job/run-pr-code-checks/11214/console ) 

@smuzaffar I've noticed in this and other PR that completing the code-check is taking signifcantly longer recently than in the past (up to several hours).

cvuosalo:
+1

kpedro88:
+upgrade

zhenhu:
+1

christopheralanwest:
@zhenhu Could you respond to the comment at https://github.com/cms-sw/cmssw/pull/27517#issuecomment-516595542? How does the MC submission infrastructure handle the "auto:" string, particularly the colon?

siddhesh86:
> In a recent meeting, PdmV experts made a verbal statement that symbolic GTs such as these are supported by the MC submission infrastructure. Since there are no datasets that have a name that contains "auto" or any of the autoCond keys (aside from cases where the string corresponding to the key is a substring of the value), it seems that this functionality has never been used. I would like PdmV experts (attn: @prebello @zhenhu @pgunnell) to explicitly confirm here that symbolic GTs are supported prior to signing off for AlCa.

My understanding of the GT for phase2 relval is in the following. @zhenhu or @prebello can confirm it or add to it.
GT in cmsDriver is defined as auto:phase2_realistic (as for e.g. [1]), and phase2_realistic is read from autoCond.py [2]

[1] https://github.com/cms-sw/cmssw/blob/master/Configuration/PyReleaseValidation/python/upgradeWorkflowComponents.py#L332
[2] https://github.com/cms-sw/cmssw/blob/master/Configuration/AlCa/python/autoCond.py#L79

christopheralanwest:
@siddhesh86 I think that there's a misunderstanding. The point of this PR is to modify the procedure that you describe. This PR proposes to instead use keys such as https://github.com/cms-sw/cmssw/pull/27517/files#diff-0b98ab5f4c0e35c17624487441dbab08R332 that are _not_ defined by a true global tag in autoCond. Instead, they use a symbolic global tag that is based on `auto:phase2_realistic`. The value that corresponds to the autoCond key in these cases key is no longer a string but a tuple. The 0th element of this tuple is the usual global tag and the rest describes conditions from that global tag that should be overridden in the symbolic GT.

siddhesh86:
@christopheralanwest I understand your point now. But I don't know at the moment what you asked.

chayanit:
> In a recent meeting, PdmV experts made a verbal statement that symbolic GTs such as these are supported by the MC submission infrastructure. Since there are no datasets that have a name that contains "auto" or any of the autoCond keys (aside from cases where the string corresponding to the key is a substring of the value), it seems that this functionality has never been used. I would like PdmV experts (attn: @prebello @zhenhu @pgunnell) to explicitly confirm here that symbolic GTs are supported prior to signing off for AlCa.

Hi @christopheralanwest, @mmusich 
So far, the files (autoCond.py and upgradeWorkflowComponents.py) modified by @mmusich only affect to MC "relvals" production/submission and we usually use these autoCond keys in MC relval production. I think if everything can pass IB test, we are also fine with these new symbolic GTs (I'm personally curious how cmsDriver looks like with these new symbolic GTs tho)
However, for MC official production is another story. It never handle auto:GTs. As fas as I can see, these symbolic GTs will work for relval production but not in the MC official production. Is that what you worry Chris?
Cheers,
Chayanit

mmusich:
Hi @chayanit 

> I think if everything can pass IB test, we are also fine with these new symbolic GTs 

great!

> However, for MC official production is another story. It never handle auto:GTs. As fas as I can see, these symbolic GTs will work for MC relval production but not in the MC official production.

The case of relval productions is what I am targeting here, to limit the number of queues with different geometry content to be maintained by AlCa. 
I see the point with the official production, but couldn't that case (that always involves a single geometry) be handled with an ad-hoc GT, that - given the low frequency of such campaigns - AlCa can certainly afford to create?

chayanit:
Hi @mmusich,
> The case of relval productions is what I am targeting here, to limit the number of queues with different geometry content to be maintained by AlCa.
I see and I agreed, I just wanna make sure that we are all the same page with AlCa too that we are only referring to RelVal production here to make use of these symbolic GTs. I foresee It will also make our relval submission more effective. Thanks.

> I see the point with the official production, but couldn't that case (that always involves a single geometry) be handled with an ad-hoc GT, that - given the low frequency of such campaigns - AlCa can certainly afford to create?
I guess so...

christopheralanwest:
+1

My concerns were primarily with official production. If only RelVals are intended to be produced with these symbolic GTs, I think that this approach is fine.

slava77:
> Reco comparison results: 1553 differences found in the comparisons

only D46 phase-2 workflow has differences. 
Is this expected? I didn't find any clarifying comments .

mmusich:
@slava77 
quoting part of the PR description:
> Please notice that a different set of LA values (μH =0.053) is be applied to the T15 geometry with respect to the current value (μH =0.106), hence changes are expected in D45-related workflows.

This applies to `D45` as well as `D46`, and all future D's using `T15`

slava77:
+1

for https://github.com/cms-sw/cmssw/pull/27517	8931b00
- code changes are in line with the PR description. Changes in the reco code are simply a change in a config parameter in RecoLocalTracker/SiPixelRecHits/python/PixelCPEGeneric_cfi.py
- jenkins tests pass and comparisons with the baseline show differences only in phase-2 D46 (T15 tracker) workflow, which is expected. The changes are somewhat small, they start from tracking distributions.

fabiocos:
@civanch could you please have a look?

civanch:
+1

fabiocos:
I believe there is consensus that this mechanism is useful if it is a way of simplifying tests, and it stays restricted to RelVal-level tests where many different possibilities are probed. For full production I think we should rely on full GTs, at that point it is clear which is the scenario one wants to produce and a corresponding GT may be built for it. In this sense I think that this update does not worsen the situation discussed in #27393 because it is localized in a centrally maintained configuration meant to be used only for tests. 

In the long term, when we converge on a single final scenario, this mechanism will not be any more needed.

Configuration/AlCa/python/autoCondPhase2.py
```diff
@@ -0,0 +1,80 @@
+# default value for all customizations
+connectionString = "frontier://FrontierProd/CMS_CONDITIONS"
```
fabiocos: @mmusich could not we get this directly from the central GlobalTag configuration?

fabiocos (commented):
after a clarification with @davidlange6 about his comments in #27393 I understand that a basic remark is about spreading this default is several places, instead of retrieving it from a single source.
Besides a possible wider discussion with AlCa, @kpedro88 et others about possible alternative ways to manage GTs for upgrade, this seems a specific point that could be addressed

Configuration/AlCa/python/autoCondPhase2.py
mmusich: how?

mmusich: you mean from here: https://github.com/cms-sw/cmssw/blob/a6191a11f9a39a95324f883b4a35363354d92324/CondCore/ESSources/python/CondDBESSource_cfi.py#L7 ?

fabiocos: right,

```
from Configuration.StandardSequences.CondDBESSource_cff import GlobalTag as essource
connectionString = essource.connect.value()
```

should work for what I can see in a quick test of mine

mmusich: Hi @fabiocos, OK. I came to the same conclusion. 
If I change the configuration now (which will make the PR loose all the signature acquired in the last month) will it be sufficient for this to be merged?

slava77:
@cmsbuild please test

fabiocos:
@kpedro88 @slava77 @perrotta @civanch @cvuosalo I'll consider your previous signatures as valid
@christopheralanwest @pohsun I believe this is a cleaner solution, in the spirit of the discussion #27393 please comment or sign

fabiocos:
@christopheralanwest @pohsun the differences are seen only for the scenario D46 (tracker geometry T15) as in the previously approved tests

christopheralanwest:
+1

fabiocos:
+1

fabiocos:
merge

cvuosalo:
+1

slava77:
+1

for https://github.com/cms-sw/cmssw/pull/27517	9e4013e
- the last update was not affecting reco https://github.com/cms-sw/cmssw/pull/27517/files/8931b0038825156e2ee40fb7401a89ffeee9379b..9e4013e660c5cad1108e7f4ca8837911f779630e

kpedro88:
+upgrade

PR #27520, opened 2019-07-15
TRK POG Phase2 : limit highPurity tracks up to |eta| < 4.0

mtosi:
#### PR description:
as suggested/requested by PF group at the UPSG workshop last week (@hatakeyamak @bendavid @JanFSchulte @jalimena),
the highPurity definition for the tracks in Phase2 is updated in order to consider *only* tracks w/ |eta| < 4.0
currently the highPurity is handle by the MultiTrackSelector, therefore such requirement can be easily integrated

it would be particularly useful to  apply such a selection already before the pattern recognition,
in order to avoid the waste of CPU resources on tracks which then are discarded

#### PR validation:
runTheMatrix.py -w upgrade -l 21234.1 (w/ and w/o this update)
runTheMatrix.py -l limited -i all --ibeos


mtosi:
upgrade

slava77:
>  the highPurity definition for the tracks in Phase2 is updated in order to consider _only_ tracks w/ |eta| < 4.0

If I'm not mistaken, this was justified by the large increase in the fake rate in the MTV plots.
Just to make sure this is real, are we saving enough tracking particles in that region to be able to measure the efficiency?

slava77:
> be able to measure the efficiency?

I meant the fake rate as well

perrotta:
assign upgrade

mtosi:
not that many, indeed, but still enough
https://tinyurl.com/y533ygzb

eff: https://tinyurl.com/y294j44f
fake: https://tinyurl.com/y2pqhvaa

kpedro88:
please test

slava77:
> not that many, indeed, but still enough
> https://tinyurl.com/y533ygzb
> 
> eff: https://tinyurl.com/y294j44f
> fake: https://tinyurl.com/y2pqhvaa

based on these plots I'd argue that 4.1 is perhaps a bit more appropriate.

VinInn:
In my opinion kinematic cuts should not be used to define quality. 
They can always added later and independently...

slava77:
> In my opinion kinematic cuts should not be used to define quality.
> They can always added later and independently...

Isn't BDT/DNN selection using track eta? 

For these phase-2 cases we are still using cut-based selections. So, the flexibility  for eta-dependent requirements is perhaps lacking and it seems like an easy solution is to "categorize" as everything fails above some value.

hatakeyamak:
> > In my opinion kinematic cuts should not be used to define quality.
> > They can always added later and independently...
> 
> Isn't BDT/DNN selection using track eta?
> 
> For these phase-2 cases we are still using cut-based selections. So, the flexibility for eta-dependent requirements is perhaps lacking and it seems like an easy solution is to "categorize" as everything fails above some value.

My comment at the workshop was that we wanted highPurity to be better defined and reduce the fake rate at high eta so that we don't have to apply extra cleaning at the PF level.
Probably this |eta|<4.0 cut is ok for now, and tracking experts may be able to come up with a cleaner selection based on track quality in future...

kpedro88:
+upgrade

Minor changes occur in tracks and downstream quantities (PF candidates, jets) at high eta, as expected:
PFCandidate eta
jet eta

slava77:
@mtosi @JanFSchulte 
please confirm if you prefer to stay with the originally proposed choice of 4.0 or if you are considering a different selection.
Thank you.

JanFSchulte:
Well, we somewhat misunderstood @hatakeyamak 's comment and thought he had requested a hard cut. A better track selection would be something I would prefer over hard cuts, and I guess the DNN will provide that at some point. But if something has to happen now, I am fine with the cut.

slava77:
Looking iter by iter in https://tinyurl.com/y294j44f 
- only detachedQuad really needs the cut: this iter has some visible fakerate in this eta range and essentially no efficiency
fakerate_d4HP 
- the highPt triplet appears to be hurt significantly
fakerate_h3HP
effic_h3HP

The highPt triplet provides essentially all of the efficiency in the range |eta|>4, as can be seen in the HP eff  plot
https://cmsweb.cern.ch/dqm/relval/plotfairy/archive/1/RelValTTbar_14TeV/CMSSW_10_6_0_patch2-PU25ns_106X_upgrade2023_realistic_v3_2023D41PU200_v2-v1/DQMIO/Tracking/Track/cutsRecoHp_trackingParticleRecoAsssociation/effic?session=GBxnQt;v=1561070026489696129;w=814;h=525
effic_HP


Based on this, it seems to me that the cut can be applied safely for all except highPtTriplet:
- it will have an impact only in the detachedQuad case
- it looks noticeably harmful for the highPtTriplet
- it will do essentially nothing in the remaining iterations

@mtosi @JanFSchulte 
if you agree, please update the PR.
Thank you.

slava77:
just in case, "ByOriginalAlgo" lead to the same conclusion 
https://cmsweb.cern.ch/dqm/relval/plotfairy/archive/1/RelValTTbar_14TeV/CMSSW_10_6_0_patch2-PU25ns_106X_upgrade2023_realistic_v3_2023D41PU200_v2-v1/DQMIO/Tracking/Track/cutsRecoHighPtTripletStepByOriginalAlgoHp_trackingParticleRecoAsssociation/effic?session=GBxnQt;v=1561070026489696129;w=814;h=525

slava77:
> Based on this, it seems to me that the cut can be applied safely for all except highPtTriplet:
> 
>     * it will have an impact only in the detachedQuad case
> 
>     * it looks noticeably harmful for the highPtTriplet
> 
>     * it will do essentially nothing in the remaining iterations
> 
> 
> @mtosi @JanFSchulte
> if you agree, please update the PR.
> Thank you.

Please comment.

mtosi:
ciao
thanks for the support, i totally agree w/ you
and i’m going to update this PR in the upcoming hours

mtosi:
all selections have been updated, but the one for the DetachedQuadStep tracks
for the one for the HighPtTripletStep tracks it has been removed
while for the other the value has been updated to 4.1

slava77:
@cmsbuild please test

RecoTracker/IterativeTracking/python/HighPtTripletStep_cff.py
```diff
@@ -302,8 +302,8 @@
         RecoTracker.FinalTrackSelectors.multiTrackSelector_cfi.highpurityMTS.clone(
             name = 'highPtTripletStep',
             preFilterName = 'highPtTripletStepTight',
-            min_eta = -4.0,
-            max_eta = 4.0,
+#            min_eta = -4.0, # this steps the fake/eff is small for |eta|>4.0, and such a cut looks noticeably harmful
+#            max_eta = 4.0,
```
slava77: ehm, I was actually implying that here we can use 4.1 
fr https://user-images.githubusercontent.com/4676718/61416295-0d24d300-a8a8-11e9-8fd1-66d230949be9.png
and eff https://user-images.githubusercontent.com/4676718/61416296-0d24d300-a8a8-11e9-8a3d-f4616b9fc166.png
or, better, a log-scale of eff https://tinyurl.com/y56bx7j9
eff_h3HP_logy

plots suggest that the balance of FR vs eff is bad for |eta|>4.1

mtosi: opsi, right, thanks @slava77 
I got confused

I've just updated the PR w/ the following

InitialStep_cff.py:                   min_eta = -4.1,
HighPtTripletStep_cff.py:      min_eta = -4.1
LowPtQuadStep_cff.py:         min_eta = -4.0,
LowPtTripletStep_cff.py:       min_eta = -4.0,
DetachedQuadStep_cff.py:   min_eta = -4.0,
PixelPairStep_cff.py:#            min_eta = -4.0, # useless, the eta distribution does not even reach |eta| = 3 (!)

slava77:
@cmsbuild please test

kpedro88:
+upgrade
extremely minor change in PFcandidates

slava77:
+1

for https://github.com/cms-sw/cmssw/pull/27520	09f8739
- jenkins tests pass and comparisons with the baseline show differences only in phase-2 workflows in plots related to tracking and downstream in agreement with the code (python) changes.

fabiocos:
+1

PR #27522, opened 2019-07-15
New GsfEleEBEECut class to generalize some ElectronIdentification cuts

guitargeek:
#### PR description:

I was never a big fan of the army of cut plugins in the Egamma ID department, so I'd like to suggest replacing some of them with a more general cut class based on the cut parser. One could generalize further, but I didn't want to take too many steps at the time and risk causing differences.

#### PR validation:

CMSSW compiles and local matrix tests pass. There are no changes expected.

No backport intended.


slava77:
@cmsbuild please test

slava77:
> I'd like to suggest replacing some of them with a more general cut class based on the cut parser.

please provide some assessments/jutifications related to the following:
- thread safety
- memory use
- CPU use

Normally, hardcoded/precompiled implementations can be the most optimal. If the string parser is just as good, then a bulk migration to it should be OK.

guitargeek:
These "plugin not found" errors are very mysterious to me.. Probably I will come back to this later and then also comment on performance.

perrotta:
@guitargeek , do you have any plan for this PR?

guitargeek:
This `PluginLibraryLoadError` was really strange to me because I didn't even touch the BuildFile for the RecoEgammaElectronIdentificationPlugins_cuts plugin! I actually had the same problem popping up in the local matrix tests too. One time it somehow worked, so I did the PR.

I was hoping that this problem was temporary and tested these developments again later with a newer base release. I did this today, but the problem persists. Here again the relevant message:

```
----- Begin Fatal Exception 30-Jul-2019 14:48:54 CEST-----------------------
An exception of category 'PluginLibraryLoadError' occurred while
   [1] Constructing module: class=VersionedGsfElectronIdProducer label='egmGsfElectronIDs'
Exception Message:
unable to load /scratch/rembser/cmssw/dev/CMSSW_11_0_0_pre4/lib/slc7_amd64_gcc700/poisoned/plugin-poisoned-RecoEgammaElectronIdentificationPlugins_cuts.so because /scratch/rembser/cmssw/dev/CMSSW_11_0_0_pre4/lib/slc7_amd64_gcc700/poisoned/plugin-poisoned-RecoEgammaElectronIdentificationPlugins_cuts.so: cannot open shared object file: No such file or directory

```

Do you have an idea what this problem is or who might know the solution? I even tried renaming the cuts plugin library in the BuildFile and doing a `scram b distclean`, but even after that it looks for the old name ('RecoEgammaElectronIdentificationPlugins_cuts.so') and complains it does not find it in the `poisoned` directory (which does not even exist by the way).

perrotta:
> These "plugin not found" errors are very mysterious to me.. Probably I will come back to this later and then also comment on performance.

Uhm, I also made some test but I've no idea why compilation ends up with those "poisoned" libraries.
@Dr15Jones , any hint?

slava77:
perhaps there is a problem when one /plugins builds two separate plugins.
All outputs of it are poisoned.
Perhaps touching or adding a space somewhere in one of the the cuts/ files will kludge around the problem.

Dr15Jones:
The first question is, what is the failing module attempting to load? You can determine that by adding
`process.add_(cms.Service("PrintLoadingPlugins"))`

to the configuration.

Dr15Jones:
So `VersionedGsfElectronIdProducer` is calling `CutApplicatorFactory` 
https://github.com/cms-sw/cmssw/blob/162772335acb92d0a9585a6d887ecb0fb64826ed/PhysicsTools/SelectorUtils/interface/VersionedSelector.h#L216

and files in the `plugins/cuts` directory which you removed are plugins loaded by the factory. So it appears modules outside of this package are dependent upon plugins in this package.

guitargeek:
Hi Slava and Chris, thanks a lot for the comments! I can't try anything out right now, but I thought about that thing with the factory before ant therefore I made sure to checkout the SelectorUtils package. If that would be the problem, that would suffice, right?

guitargeek:
Ok don't mind my previous message, I was looking at it again now, and you were right of course. I forgot about the HEEP ID and accidentally removed some plugins that were used by it. The error message with the poisoned directory didn't guide my thoughts in the right direction, but now it's obvious. Sorry for loosing time here. I will test locally again with the relevant plugins brought back and then update the PR. If you don't mind we can let the bot test it, and if we know the reco output is the same I'll do a round of IgProf.

Dr15Jones:
_poison_ files are meant to catch the case where you remove a plugin from your work area but the same plugin is in the release area. It works the following
- when you checkout a package with a plugin, a .edmplugincache file containing the plugin seen in the release and the library is added to the _poison_ directory in your personal area.
- when you build the library, the plugins that are actually in your version of the library are added to  the .edmplugincache file in your regular lib directory.

When cmsRun attempts to load a plugin, it first checks your lib directory's  .edmplugincache file for a reference to the plugin. If it is not found, it then checks your _poison_ directory's .edmplugincache. If that fails, it checks the release area's lib directory.

The use of the _poison_ directory was the easiest way for us to stop accidently picking up a removed plugin from the release area as the _fallback-to-other-directories-when-searching-for-plugins_ was already built into the plugin system. The downside is the error message isn't very informative (but we do catch the error).

guitargeek:
Thank you @Dr15Jones for that nice explanation you provided! That is very good to know. Even though the error message is not very informative, I will know to interpret it right the next time.

guitargeek:
I brought the mistakenly deleted plugins back. Could we please have one round of tests before discussing performance implications?

Dr15Jones:
please test

perrotta:
@guitargeek : it is not really urgent, also given the vacation period, but do you plan to fix the related issue with the unit test any soon?

guitargeek:
Hi @perrotta sure that was not a big deal and I fixed it already a week ago. Sorry for not pushing it before, but originally I wanted to push it together with my profiling report. The profiling report never came to be due to my eternal struggle with igprof problems and then the holidays began, so that didn't work out :) In two weeks I'll try again!

perrotta:
please test
(Thank you @guitargeek )

perrotta:
@guitargeek (just in case you are already back from vacation), please let us know when you'll have news about the profiling report announced in https://github.com/cms-sw/cmssw/pull/27522#issuecomment-520019606
I must say that a quick comparison of CPU performances with the TTbar PU wf 11024  didn't show any appreciable difference wrt baseline

guitargeek:
Hi @perrotta, yes I'm back. I'll retry the profiling later this week. Like you, I don't expect to see significant differences. The Twiki hints there is no significant performance overhead [1], and also we didn't see significant differences when I did the same thing last year with the MVA which parses way more variables.

[1] https://twiki.cern.ch/twiki/bin/view/CMSPublic/SWGuidePhysicsCutParser#Performance_issues

perrotta:
please test
(to refresh the tests)

guitargeek:
I can confirm now that the performance is not impacted. I ran the ElectronMVANtuplizer where I changed the config a bit to use all Run 2 cut based IDs [1] on one Drell-Yan MiniAOD file (32418 electrons), should be enough stats. the igprof reports are here.

[igreport_reference.txt](https://github.com/cms-sw/cmssw/files/3562212/igreport_reference.txt)
[igreport_target.txt](https://github.com/cms-sw/cmssw/files/3562213/igreport_target.txt)

Reference is CMSSW_11_0_X_2019-08-30-1100 and target is CMSSW_11_0_X_2019-08-30-1100 plus this PR. We can observe that in the reference report, the *VersionedIdProducer* takes 44.8 % while it takes 44.5 % in the target report*.

In the config file I created, I enabled all the Run 2 cut based electron IDs, even the one disabled in the official workflows. So I took the opportunity to validate the output by comparing the root files that I got out of the igprof tests:

```Python
import uproot
import numpy

tree_reference = uproot.open("electron_ntuple_reference.root")["ntuplizer/tree"]
tree_target = uproot.open("electron_ntuple_target.root")["ntuplizer/tree"]

for branch in tree_target.keys():
    a_target = tree_target.array(branch)
    a_reference = tree_reference.array(branch)

    numpy.testing.assert_array_almost_equal(a_target, a_reference)
```

Everything seems fine from the asserts.

[1] https://gist.github.com/guitargeek/d0ebda273faf5fc84d60bf7a8d347ea9

*if you add the two entries in the report that correspond to VersionedIdProducer::produce

perrotta:
+1
- A few cut plugins in the Egamma ID area are replaced with a more general cut class based on the cut parser
- Jenkins tests show that the results of the cuts are unchanged
- Performance are not affects, as verified with a simple CPU time measurement and some dedicated igprof profiling

fabiocos:
+1

PR #27526, opened 2019-07-15
drop constness for modifiers in RecoTauQualityCuts; move auto_ptr to unique_ptr

slava77:
- the first and last commits are related to the static analyzer category of issues "ConstThreadSafety | mutable member if accessed via const pointer" for RecoTauQualityCuts:
    - setXYZ methods are no longer declared const (and the corresponding members are no longer mutable)
    - calls to RecoTauQualityCuts in the client methods qualified as const now go via pointers to the instances of RecoTauQualityCuts. Incidentally, this was already used this way in some RecoTau code

The migration from auto_ptr to unique_ptr is a somewhat aside cleanup of a deprecated C++ feature (removed in C++17).

This was tested in re-miniAOD workflows and showed no differences in tau reco outputs.


slava77:
@cmsbuild please test

slava77:
+1

for https://github.com/cms-sw/cmssw/pull/27526	a901c48
- jenkins tests passed and comparisons with the baseline show no (relevant) differences

fabiocos:
+1

Issue #27528, opened 2019-07-16
DQM merge failure in CMSSW_11_0_0_pre3


fabiocos:
The PdmV team reports a massive failure of the validation in CMSSW_11_0_0_pre3 in the DQMIO merge step, with errors like:
```
Fatal Exception (Exit code: 8022) 
An exception of category 'FatalRootError' occurred while
[0] Calling InputSource::readRun_
Additional Info:
[a] Fatal Root Error: @SUB=TProfileHelper::Merge
Cannot merge profiles 1 dim - limits are inconsistent:
first: (31, -9.000000, 15.000000), second: (31, -15.000000, 9.000000)
```
seen for instance in

https://cms-unified.web.cern.ch/cms-unified/report/muahmad_RVCMSSW_11_0_0_pre3Higgs200ChargedTaus_13_190711_065022_9820

or

https://cms-unified.web.cern.ch/cms-unified/report/chayanit_RVCMSSW_11_0_0_pre3RunCharmonium2018D__RelVal_2018D_190710_083031_7155

(but the problem is general).

In the set of IB workflows we have now test 137.8 which performs DQM harvesting on two separate files from two periods, without failing. I have also tested the DQMIO merge (missing, I would add it), but it does not fail. But if I try to split 136.85 in two pieces of 50 events each, and I try the merge, I get the failure.

This looks a show-stopper for a meaningful validation of the release, and should be solved asap.

fabiocos:
assign dqm

for the problem itself

fabiocos:
assign pdmv 

for the extension of release tests

fabiocos:
@jfernan2 this looks the continuation of #26727 . I will close that, keeping here there reference to it

fabiocos:
repeating the test with wf 136.85 in the recent IB, using the addition by @Dr15Jones #27473 👍 
```
----- Begin Fatal Exception 16-Jul-2019 13:13:29 CEST-----------------------
An exception of category 'FatalRootError' occurred while
   [0] Calling InputSource::readRun_
   [1] While reading element Tracking/TrackParameters/generalTracks/GeneralProperties/DistanceOfClosestApproachToBSVsEta_GenTk
   Additional Info:
      [a] Fatal Root Error: @SUB=TProfileHelper::Merge
Cannot merge profiles 1 dim - limits are inconsistent:
 first: (31, -15.000000, 9.000000), second: (31, -9.000000, 15.000000)

----- End Fatal Exception -------------------------------------------------
[...]

PR #27330, opened 2019-06-24
New plots for Tracking Workspace 

hbecerri:
PR description:

Add new plots in the Tracking Workspace

dz with respect to BS
dxy with respect to the BS vs eta and phi,
q over p
track quality (isLoose, isTight, isHighQuality),
"inclusively" but also vs pT and eta
number of tracks per vertex, for the hard vertex and
for the other (PU) vertices
tracking efficiency plots from the folding method, for the
"eta vs phi map"
tracking efficiency vs pT of the tracks

Validated with  runTheMatrix.py -l 10024

PR #27531, opened 2019-07-16
remove SetCanExtend for the histogram DistanceOfClosestApproachToBSVsEta

fabiocos:
#### PR description:

This PR addresses the problem #27528 by removing the option ```SetCanExtend``` for the ME ```DistanceOfClosestApproachToBSVsEta``` added in #27330. 

#### PR validation:

Splitting the step3 of test wf 136.85 into two parts, and merging the two DQM outputs, does not fail any more.

### DIFF
--- a/DQM/TrackingMonitor/src/TrackAnalyzer.cc
+++ b/DQM/TrackingMonitor/src/TrackAnalyzer.cc
@@ -863,7 +863,7 @@ void TrackAnalyzer::bookHistosForBeamSpot(DQMStore::IBooker& ibooker) {
       histname = "DistanceOfClosestApproachToBSVsEta_";
       DistanceOfClosestApproachToBSVsEta = ibooker.bookProfile(
           histname + CategoryName, histname + CategoryName, EtaBin, EtaMin, EtaMax, DxyBin, DxyMin, DxyMax, "");
-      DistanceOfClosestApproachToBSVsEta->getTH1()->SetCanExtend(TH1::kAllAxes);
+      //      DistanceOfClosestApproachToBSVsEta->getTH1()->SetCanExtend(TH1::kAllAxes);   // causing issues with DQMIO merge
       DistanceOfClosestApproachToBSVsEta->setAxisTitle("Track #eta", 1);
       DistanceOfClosestApproachToBSVsEta->setAxisTitle("Track d_{xy} wrt beam spot (cm)", 2);
     }


fabiocos:
@hbecerri this is meant to be a quick fix to circumvent a problem that is preventing the validation of CMSSW_11_0_0_pre3. If you have an alternative solution please make a suggestion.

fabiocos:
please test workflow 137.8

fabiocos:
type bugfix

mtosi:
opsi, thanks @fabiocos for this fix

how do other DQM codes handle such a problem ?
@fioriNTU @schneiml

schneiml:
My approximate recommendation would be to never ever use `SetCanExtend`.

What is the use case here? Maybe there is a better way.

fabiocos:
unfortunately browsing the DQM code this is used in several places... 
https://cmssdt.cern.ch/lxr/search?%21v=CMSSW_11_0_X_2019-07-15-2300&_filestring=&_string=SetCanExtend

schneiml:
Yes, and now it makes me wonder if ROOT always fails to merge histograms when they where extended in different ways. Actually, I would be surprised by that. Maybe that is specific to `TProfile`?

mmusich:
@fabiocos @schneiml @hbecerri  I don't understand...
3 lines below there is exactly the same code:
```
histname = "DistanceOfClosestApproachToBSVsPhi_";
    DistanceOfClosestApproachToBSVsPhi = ibooker.bookProfile(
        histname + CategoryName, histname + CategoryName, PhiBin, PhiMin, PhiMax, DxyBin, DxyMin, DxyMax, "");
    DistanceOfClosestApproachToBSVsPhi->getTH1()->SetCanExtend(TH1::kAllAxes);
    DistanceOfClosestApproachToBSVsPhi->setAxisTitle("Track #phi", 1);
    DistanceOfClosestApproachToBSVsPhi->setAxisTitle("Track d_{xy} wrt beam spot (cm)", 2);
```

in what way is this different?

fabiocos:
@mmusich empirically in the fact that they do not trigger a failure. I may just guess, to be verified, that in one case the range is extended, while in another it never happens, but this should be checked.

mmusich:
@hbecerri I actually don't understand why you need to extend the profile. It's a profile vs eta, we certainly don't expect the detector to get any bigger. Same goes true for the profile vs phi.
Can you clarify?

hbecerri:
Hello @mmusich. The fact that the range can be extended for profile vs phi (despite that the detector don't get any bigger as you said) makes me think that it was a well stablished format for this kind of plots that I should respect. I agree if you decide to not use SetCanExtend at all

mmusich:
Hi @hbecerri thanks for the explanation, so if I understand correctly it's some sort of copy-n-paste error and we unfortunately hit a case where the plot gets actually extended. 
Could you make a PR where all the occurrences of `SetCanExtend` are removed from this file?

hbecerri:
@mmusich Sure, I am on it

davidlange6:
does root actually keep track of when it has extended a histogram? I haven't found it.. if not, I guess the merging of two histograms that started as the same thing but got extended differently works or fails just as the merging of two histograms with different histories.

jfernan2:
The histogram was created within an unphysical eta range:
https://tinyurl.com/yy6g7z9x
despite being created in a proper range. As I said in:
https://github.com/cms-sw/cmssw/pull/27535#issuecomment-511890811
it would be good to know the reason. 
Anyway the extension of the bin edges is clearly not needed

jfernan2:
+1

mmusich:
This PR is superseded by https://github.com/cms-sw/cmssw/pull/27535

fabiocos:
@mmusich @hbecerri thanks, that was my first attempt, but then I saw that just this instance was added by the recent PR and triggered the crash, so I let you experts to look deeper into it. So we migth just integrate #27535 once approved

fabiocos:
hold

fabiocos:
unhold

fabiocos:
superseded by #27535

PR #27535, opened 2019-07-16
Remove all the ocurrences of SetCanExtend from TrackAnalyzer

hbecerri:
#### PR description:

This PR removes all the occurrences of SetCanExtend from TrackAnalyzer.cc. Solves the issue discussed in #27531

#### PR validation:

tested using runTheMatrix.py -l 10024

### DIFF
--- a/DQM/TrackingMonitor/src/TrackAnalyzer.cc
+++ b/DQM/TrackingMonitor/src/TrackAnalyzer.cc
@@ -863,7 +863,6 @@ void TrackAnalyzer::bookHistosForBeamSpot(DQMStore::IBooker& ibooker) {
       histname = "DistanceOfClosestApproachToBSVsEta_";
       DistanceOfClosestApproachToBSVsEta = ibooker.bookProfile(
           histname + CategoryName, histname + CategoryName, EtaBin, EtaMin, EtaMax, DxyBin, DxyMin, DxyMax, "");
-      DistanceOfClosestApproachToBSVsEta->getTH1()->SetCanExtend(TH1::kAllAxes);
       DistanceOfClosestApproachToBSVsEta->setAxisTitle("Track #eta", 1);
       DistanceOfClosestApproachToBSVsEta->setAxisTitle("Track d_{xy} wrt beam spot (cm)", 2);
     }
@@ -877,7 +876,6 @@ void TrackAnalyzer::bookHistosForBeamSpot(DQMStore::IBooker& ibooker) {
     histname = "DistanceOfClosestApproachToBSVsPhi_";
     DistanceOfClosestApproachToBSVsPhi = ibooker.bookProfile(
         histname + CategoryName, histname + CategoryName, PhiBin, PhiMin, PhiMax, DxyBin, DxyMin, DxyMax, "");
-    DistanceOfClosestApproachToBSVsPhi->getTH1()->SetCanExtend(TH1::kAllAxes);
     DistanceOfClosestApproachToBSVsPhi->setAxisTitle("Track #phi", 1);
     DistanceOfClosestApproachToBSVsPhi->setAxisTitle("Track d_{xy} wrt beam spot (cm)", 2);
 
@@ -960,7 +958,6 @@ void TrackAnalyzer::bookHistosForBeamSpot(DQMStore::IBooker& ibooker) {
     histname = "DistanceOfClosestApproachToPVVsPhi_";
     DistanceOfClosestApproachToPVVsPhi = ibooker.bookProfile(
         histname + CategoryName, histname + CategoryName, PhiBin, PhiMin, PhiMax, DxyBin, DxyMin, DxyMax, "");
-    DistanceOfClosestApproachToPVVsPhi->getTH1()->SetCanExtend(TH1::kAllAxes);
     DistanceOfClosestApproachToPVVsPhi->setAxisTitle("Track #phi", 1);
     DistanceOfClosestApproachToPVVsPhi->setAxisTitle("Track d_{xy} w.r.t. PV (cm)", 2);
 
@@ -996,7 +993,6 @@ void TrackAnalyzer::bookHistosForBeamSpot(DQMStore::IBooker& ibooker) {
       histname = "TESTDistanceOfClosestApproachToBSVsPhi_";
       TESTDistanceOfClosestApproachToBSVsPhi = ibooker.bookProfile(
           histname + CategoryName, histname + CategoryName, PhiBin, PhiMin, PhiMax, DxyBin, DxyMin, DxyMax, "");
-      TESTDistanceOfClosestApproachToBSVsPhi->getTH1()->SetCanExtend(TH1::kAllAxes);
       TESTDistanceOfClosestApproachToBSVsPhi->setAxisTitle("Track #phi", 1);
       TESTDistanceOfClosestApproachToBSVsPhi->setAxisTitle("Track d_{xy} wrt beam spot (cm)", 2);
     }
@@ -1047,7 +1043,6 @@ void TrackAnalyzer::bookHistosForBeamSpot(DQMStore::IBooker& ibooker) {
       histname = "DistanceOfClosestApproachVsPhi_";
       DistanceOfClosestApproachVsPhi = ibooker.bookProfile(
           histname + CategoryName, histname + CategoryName, PhiBin, PhiMin, PhiMax, DxyMin, DxyMax, "");
-      DistanceOfClosestApproachVsPhi->getTH1()->SetCanExtend(TH1::kAllAxes);
       DistanceOfClosestApproachVsPhi->setAxisTitle("Track #phi", 1);
       DistanceOfClosestApproachVsPhi->setAxisTitle("Track d_{xy} wrt (0,0,0) (cm)", 2);
     }


jfernan2:
@hbecerri thanks for this PR, however, despite the discussion in #27535 it would be nice to know why the wf was creating eta histograms from first: (31, -15.000000, 9.000000), second: (31, -9.000000, 15.000000) as shown by @fabiocos on https://github.com/cms-sw/cmssw/issues/27528#issuecomment-511773666 
if you are actually setting 26 bins from -2.5 to 2.5 in:
https://cmssdt.cern.ch/lxr/source/DQM/TrackingMonitor/python/TrackingMonitor_cfi.py#L258
Is this line the conflicting one and a right setting?
https://cmssdt.cern.ch/lxr/source/DQM/TrackingMonitor/python/TrackingMonitor_cfi.py#L452

mmusich:
@jfernan2
> know why the wf was creating eta histograms from first: (31, -15.000000, 9.000000), second: (31, -9.000000, 15.000000)

I think this happened exactly because of the fact that SetCanExtend was activated for this profile (while there is no reason for it to be)

> Is this line the conflicting one and a right setting?

What do you mean? It's an era modifier. The phase0 detector had smaller pseudorapidity coverage hence the |eta|<2.5 axis limits, which was extended to |eta|<3 for the new phase1 detector.

mmusich:
Can tests be started here please?

fioriNTU:
please test

jfernan2:
hold

jfernan2:
@mmusich  It is disappointing that you skim my comments
What I meant is that the histogram is created with 26 bins from -2.5 to 2.5 and then the crash reports 31 bins, exactly as the Era modifier...
We need to understand what is going on before moving onward, even if the histogram is allowed to be extended, what module is causing the extension?

mmusich:
@jfernan2 

> skim my comments

what does "skim" mean in this context?

>  created with 26 bins from -2.5 to 2.5 

for my education, how can you say it is created with 26 bins? 
With this PR it gets **correctly** booked with |eta|<3 see https://tinyurl.com/y5behksq

image


> even if the histogram is allowed to be extended

there is no point in extending this particular histogram, I don't follow your logic. 
At any rate I think that David provided a plausible explanation of what is happening in https://github.com/cms-sw/cmssw/pull/27531#issuecomment-511869414

jfernan2:
> what does "skim" mean in this context?
overlook, but never mind
> for my education, how can you say it is created with 26 bins?
https://cmssdt.cern.ch/lxr/source/DQM/TrackingMonitor/python/TrackingMonitor_cfi.py#0258
> With this PR it gets **correctly** booked with |eta|<3 see https://tinyurl.com/y5behksq
I was talking about Phase0 Pixel plots
> there is no point in extending this particular histogram, I don't follow your logic.
I don't say it is logic to extend it but that the code was allowing to extend it in first instance, and it gets actually extended if you allow it:   https://tinyurl.com/y4gaqlsg
> At any rate I think that David provided a plausible explanation of what is happening in [#27531 (comment)](https://github.com/cms-sw/cmssw/pull/27531#issuecomment-511869414)
My point is WHERE they get extended in your code, to avoid problems in the future

mmusich:
> > for my education, how can you say it is created with 26 bins?
> 
> https://cmssdt.cern.ch/lxr/source/DQM/TrackingMonitor/python/TrackingMonitor_cfi.py#0258
> 

but this is valid only for phase-0 plots and not for phase-1 plots, where the era modifier prevails.

> > With this PR it gets **correctly** booked with |eta|<3 see https://tinyurl.com/y5behksq
> 
> I was talking about Phase0 Pixel plots

right, but I am still not getting your point. Sorry I must be very dense :(

> 
> > there is no point in extending this particular histogram, I don't follow your logic.
> 
> I don't say it is logic to extend it but that the code was allowing to extend it in first instance, and it gets actually extended if you allow it: https://tinyurl.com/y4gaqlsg

but isn't it true that any plot can be extended?
In my understanding that's a ROOT TH1 feature that we pass onto CMSSW DQM MonitorElements:
https://root.cern.ch/doc/master/classTH1.html#a9f608543655d452dafd1cf6bce83c683

> 
> > At any rate I think that David provided a plausible explanation of what is happening in [#27531 > :(comment)](https://github.com/cms-sw/cmssw/pull/27531#issuecomment-511869414)
> 
> My point is WHERE they get extended in your code, to avoid problems in the future

They get extended all over the place where it makes sense, like in time or LS series: (see e.g. https://github.com/cms-sw/cmssw/blob/master/DQM/TrackingMonitor/src/TrackingMonitor.cc#L348 for an example).
In my understanding the advantage of using the `SetCanExtend` is that users do not need to code their own version of the histogram merging, relying instead on the one provided by ROOT as it has been commented yesterday by @fioriNTU here: https://github.com/cms-sw/cmssw/issues/27528#issuecomment-511814521

fabiocos:
@jfernan2 @mmusich @hbecerri this PR is indeed addressing an issue that is worthwhile to investigate. But in practice #27531 is solving a problem that has broken a pre-release making it unusable, while the other instances of SetCanExtend are not for what we can see. So I consider solving the former issue as urgent. Given that the discussion on this PR might require a bit more time, I will move forward with merging #27531, this PR can just be rebased on that, and possibly update it in case

mmusich:
> Given that the discussion on this PR might require a bit more time, I will move forward with merging #27531, this PR can just be rebased on that, and possibly update it in case.

OK, provide that @hbecerri is fine in rebasing. Frankly I don't understand why the due investigations on the issue are a reason enough to hold this PR, which is basically doing the same thing as https://github.com/cms-sw/cmssw/pull/27531 (which is already DQM signed) just more cleanly. 
It seems to me PRs are reviewed differently based on the submitter...

jfernan2:
@mmusich   the line you point only allows it CAN be extended, and it is a LS based plot which makes sense since you don't know the limits a priori, but not the ones we are talking about which have a fixed range from the beginning.
You are not saying WHERE the code was extended it in a random manner in the range: -15 < eta < 15 
Just browse in the 4 plots here and you will see what I am talking about: https://tinyurl.com/y3latly4

And not, I am not reviewing PRs based on the submitter but on the urgency.  Just to avoid disappointments to you in the future ;-)

mmusich:
>  You are not saying WHERE the code was extended it in a random manner in the range: -15 < eta < 15

Now I finally see what you mean! I think it's ROOT that eventually decides where to put the boundary, could it be that ther is some nans propagating here? 
I think it can debugged by printing the x and y axis ranges of the profile every time the plot is filled.
To reproduce the plot in https://tinyurl.com/y3latly4 which wf shall I use?

jfernan2:
This is in particular is wf 4.53 but  you can see it is happening in half of the workflows, those with a non zero number in:
https://cmssdt.cern.ch/SDT/jenkins-artifacts/baseLineComparisons/CMSSW_11_0_X_2019-07-16-2300+d27f61/32673/dqm-histo-comparison-summary.html
Left column is original code, central PR, right comparision using common bin range.
Thanks

mmusich:
Hi @jfernan2 so I re-run step3 of wf 4.53 by adding these changes:
```diff
diff --git a/DQM/TrackingMonitor/src/TrackAnalyzer.cc b/DQM/TrackingMonitor/src/TrackAnalyzer.cc
index f1abe45ff75..67e628e20fd 100644
--- a/DQM/TrackingMonitor/src/TrackAnalyzer.cc
+++ b/DQM/TrackingMonitor/src/TrackAnalyzer.cc
@@ -1287,6 +1287,9 @@ void TrackAnalyzer::analyze(const edm::Event& iEvent, const edm::EventSetup& iSe
 
     if (Folder == "Tr") {
       DistanceOfClosestApproachToBSdz->Fill(track.dz(bs.position()));
+      std::cout << "====================================================================="<< std::endl;
+      std::cout << "track.eta():" << track.eta() << "  track.dxy(bs.position())):" <<  track.dxy(bs.position()) << std::endl;
+      std::cout<< "DistanceOfClosestApproachToBSVsEta->getTH1()->GetXaxis()->GetXmax(): " << DistanceOfClosestApproachToBSVsEta->getTH1()->GetXaxis()->GetXmax () <<  "    |  DistanceOfClosestApproachToBSVsEta->getTH1()->GetXaxis()->GetXmin(): " << DistanceOfClosestApproachToBSVsEta->getTH1()->GetXaxis()->GetXmin () << std::endl;
       DistanceOfClosestApproachToBSVsEta->Fill(track.eta(), track.dxy(bs.position()));
     }
```

and I identified two transitions where the x-axis range got updated:

```bash
=====================================================================
track.eta():2.50979  track.dxy(bs.position())):0.0306698
DistanceOfClosestApproachToBSVsEta->getTH1()->GetXaxis()->GetXmax(): 2.5    |  DistanceOfClosestApproachToBSVsEta->getTH1()->GetXaxis()->GetXmin(): -2.5
=====================================================================
track.eta():2.34284  track.dxy(bs.position())):-0.0247759
DistanceOfClosestApproachToBSVsEta->getTH1()->GetXaxis()->GetXmax(): 7.5    |  DistanceOfClosestApproachToBSVsEta->getTH1()->GetXaxis()->GetXmin(): -2.5
=====================================================================
```

and:
```bash
=====================================================================
track.eta():-2.50028  track.dxy(bs.position())):0.0215483
DistanceOfClosestApproachToBSVsEta->getTH1()->GetXaxis()->GetXmax(): 7.5    |  DistanceOfClosestApproachToBSVsEta->getTH1()->GetXaxis()->GetXmin(): -2.5
=====================================================================
track.eta():-2.22177  track.dxy(bs.position())):-0.0364012
DistanceOfClosestApproachToBSVsEta->getTH1()->GetXaxis()->GetXmax(): 7.5    |  DistanceOfClosestApproachToBSVsEta->getTH1()->GetXaxis()->GetXmin(): -12.5
=====================================================================
```
so it seems that indeed ROOT automatically updated the range (by some internal heuristic I guess) when we hit a track with a pseudo-rapidity very close to the boundary eta = 2.5 (which is the correct one for a phase-0 pixel detector workflow such as 4.53 (2012B)).
I don't know exactly what else we can learn from this, a part from the fact that we should in general put an epsilon more than the physical detector boundary to be able to capture tracks whose pseudo-rapidity is enhanced probably by BeamSpot z - widths effects.
Please let me know if you need more debugging...

jfernan2:
ok @mmusich that explains why the phi plots are not affected: there are no boundaries there...
Thanks!

jfernan2:
unhold

jfernan2:
+1

fabiocos:
+1

PR #27538, opened 2019-07-16
Improve L1MuGMTMatrix

Dr15Jones:
#### PR description:

- Use just a single contiguous memory location instead of an array of arrays.
- Use std::unique_ptr<T[]> to manage the memory
- Initialize matrix elements in the constructor
- Added a unit test to prove no functional changes happened

Not initializing the matrix elements in the constructor was causing a complaint by UBSAN.
#### PR validation:

Code compiles and the new unit test passes.

This is a technical change and should have no impact on the results. It should take less memory and be faster.

### DIFF
--- a/L1Trigger/GlobalMuonTrigger/src/L1MuGMTDebugBlock.cc
+++ b/L1Trigger/GlobalMuonTrigger/src/L1MuGMTDebugBlock.cc
@@ -40,8 +40,9 @@ L1MuGMTDebugBlock::L1MuGMTDebugBlock(int minbx, int maxbx)
       _phisel(maxbx - minbx + 1, std::vector<unsigned>(32, 0)),
       _etasel(maxbx - minbx + 1, std::vector<unsigned>(32, 0)),
       _isMIPISO(maxbx - minbx + 1, std::vector<unsigned>(32, 0)),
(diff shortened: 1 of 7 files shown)


Dr15Jones:
please test

Dr15Jones:
ping  @rekovic @benkrikler

fabiocos:
@rekovic I would like to move forward with this PR. The code update looks ok to me, and the tests do not show any change. Please have a look asap, I plan to merge this PR in the coming IBs unless you have explicit objections.

fabiocos:
+1

@rekovic please review this PR for possible future comments

fabiocos:
merge

davidlange6:
hi @Dr15Jones - unfortunately L1Trigger/GlobalMuonTrigger/src defines a plugin so you can't use it in the tests  here (causing a warning in release builds now)

https://github.com/cms-sw/cmssw/pull/27538/files#diff-ad9585c8e3b32fba9922df3e129141f8R17

Dr15Jones:
It would seem to me that the correct solution is to move the plugin out of /src and into a new /plugin directory (I'm on vacation today and for the next few days).

Dr15Jones:
@davidlange6 see #27828

PR #27550, opened 2019-07-18
Updates to E(H)CalEnergy definitions for PF hadrons.

hatakeyamak:
#### PR description:

Based on recent discussion in the PF meeting, the following changes are being implemented:
+ Improve rawE(H)CalEnergy definitions for neutral hadrons.
   - Neutral hadrons (photons) from "loop hcal" are not associated with ECAL (HCAL) clusters, as they are defined based on excess energies in HCAL (ECAL) energies only. rawE(H)CalEnergy is now set to zero for neutral hadrons (photons), as done for corrected energies to avoid confusions.
+ protect against the major source of negative ecal energy for charged hadrons.
   - Occasionally ecal energy for charged hadrons went negative, when photons are also created and the ecal energy corrected by PF hadron calibration is smaller than ecal energy from egamma's ecal cluster calibration. Now this is protected against going negative.
+ also protect against negative energy scaling.
   - Very rarely, charged hadron momentum scaling went negative, after rejecting of "bad" tracks. Now we protect against it as well. 

https://indico.cern.ch/event/832017/contributions/3485635/attachments/1871517/3102604/PF_NeutralHadron_CaloFractions_v4.pdf
for further information.

#### PR validation:

It compiles. Also, some validation plots in the above pdf file.

This is not a backport.

@bendavid @zdemirag @ahinzmann

### DIFF
--- a/RecoParticleFlow/PFProducer/src/PFAlgo.cc
+++ b/RecoParticleFlow/PFProducer/src/PFAlgo.cc
@@ -2315,8 +2315,7 @@ void PFAlgo::createCandidatesHCAL(const reco::PFBlock& block,
           particleDirection.push_back(std::get<1>(pae));
           ecalEnergy.push_back(mergedPhotonEnergy * clusterEnergyCalibrated / sumEcalClusters);
           hcalEnergy.push_back(0.);
-          rawecalEnergy.push_back(mergedPhotonEnergy * clusterEnergyCalibrated /
-                                  sumEcalClusters);  // "raw" not well defined. use the corrected value.
+          rawecalEnergy.push_back(totalEcal);
           rawhcalEnergy.push_back(0.);
           pivotalClusterRef.push_back(elements[std::get<0>(pae)].clusterRef());
           iPivotal.push_back(std::get<0>(pae));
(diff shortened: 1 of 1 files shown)


perrotta:
please test

fabiocos:
@slava77 @mrodozov the recent update of workflow names 2023 -> 2026 in #27488 requires to update the map in the bot for comparisons

fabiocos:
see https://github.com/cms-sw/cms-bot/pull/1170

slava77:
@cmsbuild please test

to refresh with now appropriately configured reco comparisons in the cms-bot

RecoParticleFlow/PFProducer/src/PFAlgo.cc
```diff
@@ -2311,8 +2315,9 @@ void PFAlgo::createCandidatesHCAL(const reco::PFBlock& block,
           particleDirection.push_back(std::get<1>(pae));
           ecalEnergy.push_back(mergedPhotonEnergy * clusterEnergyCalibrated / sumEcalClusters);
           hcalEnergy.push_back(0.);
-          rawecalEnergy.push_back(totalEcal);
-          rawhcalEnergy.push_back(totalHcal);
+          rawecalEnergy.push_back(mergedPhotonEnergy * clusterEnergyCalibrated /
+                                  sumEcalClusters);  // "raw" not well defined. use the corrected value.
+          rawhcalEnergy.push_back(0.);
```
slava77: this breaks the HO energy settings .

Some protections are necessary below in 
https://github.com/cms-sw/cmssw/blob/18425f38932959335ace0f686b663eacfa129e69/RecoParticleFlow/PFProducer/src/PFAlgo.cc#L2374-L2377

hatakeyamak: Thank you @slava77 . I made the change.

perrotta:
@cmsbuild please test

slava77:
+1

for https://github.com/cms-sw/cmssw/pull/27550	0458365
- jenkins tests pass and comparisons with the baseline show differences primarily in the PF candidate energy fractions, as expected
    - in AOD/RECO: tau emFraction is now non-negative (ignoring default "-1" values), which propagates to various tau discriminants as small changes
    - in miniAOD caloFraction and hcalFraction of PF candidates change in sync and hcalFraction is now bound in [0,1] range
    - also, in some rate cases (only in wf 25202 among all the tested wfs, including  jenkins and also some limited particle guns, dijets, and ttbar with PU and 2018 data in local tests) there is a change in PF charged momentum

kpedro88:
+1

PR #27558, opened 2019-07-18
Use default compiler generated methods in TrajectorySeed

Dr15Jones:
#### PR description:

The hand written move based operator= was flagging an UBSAN error when RVO was not able to be used. The uninitialized final value was having its members being swapped to the returned object.

#### PR validation:

The code compiles.

### DIFF
--- a/DataFormats/TrajectorySeed/interface/TrajectorySeed.h
+++ b/DataFormats/TrajectorySeed/interface/TrajectorySeed.h
@@ -41,21 +41,13 @@ public:
     std::swap(dir_, rh.dir_);
   }
 
-  TrajectorySeed(TrajectorySeed const& o) : hits_(o.hits_), tsos_(o.tsos_), dir_(o.dir_) {}
+  TrajectorySeed(TrajectorySeed const& o) = default;
 
-  TrajectorySeed& operator=(TrajectorySeed const& o) {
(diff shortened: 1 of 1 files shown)


Dr15Jones:
please test

Dr15Jones:
The UBSAN report was complaining that an uninitialized value was being used in the call to `operator=` when assigning to the return value here

https://github.com/cms-sw/cmssw/blob/ae97882c71933048f2a47a9eaeac124813029f8f/RecoMuon/MuonSeedGenerator/src/MuonCSCSeedFromRecHits.cc#L220

ultimately the object assigned to `seed` came from here
https://github.com/cms-sw/cmssw/blob/ae97882c71933048f2a47a9eaeac124813029f8f/RecoMuon/MuonSeedGenerator/src/MuonCSCSeedFromRecHits.cc#L18

As the default constructor does not set any values, this lead to the UBSAN error. The compiler generated move `operator=` does not suffer from that problem since it would not swap the values and instead just overwrite the uninitialized values.

slava77:
+1

for https://github.com/cms-sw/cmssw/pull/27558	d12ba21
- technical; in line with the PR description and https://github.com/cms-sw/cmssw/pull/27558#issuecomment-512977337

fabiocos:
+1

PR #27559, opened 2019-07-18
SimPPS/PPSPixelDigiProducer package for PPS Full simulation

mundim:
#### PR description:
New package with Digi producer for the Pixel sensor of PPS, needed for the full simulation

#### PR validation:
The following tests have been run:
scram b runtests
scram b code-checks
scram b code-formats (which triggered some style changes in the files)

runTheMatrix.py --job-reports -j 14 -s -l 101.0,10224.0,1306.0,250202.181,25202.0,9.0
which produced the output
20 15 14 13 7 1 0 0 0 tests passed, 12 4 0 0 0 0 0 0 0 failed


mundim:
just to report the result of:
runTheMatrix.py --job-reports -j 14 -s -l limited -i all --ibeos

produced the result:
32 30 29 24 12 2 1 1 1 tests passed, 0 1 0 0 0 0 0 0 0 failed

civanch:
please test

SimPPS/PPSPixelDigiProducer/plugins/PPSPixelDigiAnalyzer.cc
```diff
@@ -0,0 +1,309 @@
...
+class TFile;
+class PSimHit;
+
+namespace edm {
+  class ParameterSet;
+  class Event;
+  class EventSetup;
+}  // namespace edm
+
+class PPSPixelDigiAnalyzer : public edm::EDAnalyzer {
```
civanch: @mundim , in this analizer you open root files. In that case, better to inherit of "one" type of analyzers

class PPSPixelDigiAnalyze : public edm::one::EDAnalyzer<edm::one::SharedResources> {

SimPPS/PPSPixelDigiProducer/BuildFile.xml
```diff
@@ -0,0 +1,21 @@
...
+<use   name="DataFormats/CTPPSDigi"/>
+<use   name="SimDataFormats/CrossingFrame"/>
+<use   name="SimTracker/Common"/>
+<use   name="Geometry/VeryForwardGeometry"/>
+<use   name="Geometry/VeryForwardGeometryBuilder"/>
+<use   name="clhep"/>
+<use   name="Mixing/Base"/>
+<use   name="gsl"/>
+<use   name="boost"/>
+<use   name="geant4"/>
```
civanch: @mundim , you not need dependence on Geant4 (also dependences on other unused libraries better to remove)

SimPPS/PPSPixelDigiProducer/interface/RPixChargeShare.h
```diff
@@ -0,0 +1,32 @@
...
+private:
+  uint32_t det_id_;
+  std::vector<double> signalCoupling_;
+  CTPPSPixelSimTopology theRPixDetTopology_;
+  CTPPSPixelIndices pxlInd;
+  const int pxlRowSize_ = pxlInd.getDefaultRowDetSize();
+  const int pxlColSize_ = pxlInd.getDefaultColDetSize();
+  int no_of_pixels_;
+
+  double sqrt_2;
```
civanch: @mundim , is it const? or it is a variable, if const not necessary to be class member.

SimPPS/PPSPixelDigiProducer/interface/RPixChargeShare.h
```diff
@@ -0,0 +1,32 @@
...
+  int verbosity_;
+
+  std::string ChargeMapFile2E_[4];
+  double chargeMap2E[4][60][40];
```
civanch: @mundim , in this line and in several places below local member name does not corresponds to convention. For example, "verbose_" is OK but "chargeMap2E" is not.

SimPPS/PPSPixelDigiProducer/interface/RPixEnergyDepositUnit.h
```diff
@@ -0,0 +1,32 @@
...
+  RPixEnergyDepositUnit(double energy, double x, double y, double z) : energy_(energy), position_(x, y, z){};
+
+  RPixEnergyDepositUnit(double energy, const Local3DPoint& position) : energy_(energy), position_(position){};
+
+  inline double X() const { return position_.x(); }
+  inline double Y() const { return position_.y(); }
+  inline double Z() const { return position_.z(); }
+  inline double& Energy() { return energy_; }
+  inline double Energy() const { return energy_; }
+  inline Local3DPoint Position() const { return position_; }
```
civanch: @mundim , I do not see a reason to implement to methods "Energy" and "Position". I would left one for each case. 

double Energy() const;
Local3DPoint& Position() const;

SimPPS/PPSPixelDigiProducer/interface/RPixSignalPoint.h
```diff
@@ -0,0 +1,32 @@
...
+public:
+  RPixSignalPoint() : pos_(0, 0), sigma_(0), charge_(0){};
+
+  RPixSignalPoint(double x, double y, double s, double charge) : pos_(x, y), sigma_(s), charge_(charge){};
+
+  inline LocalPoint Position() const { return pos_; }
+  inline LocalPoint &Position() { return pos_; }
+  inline double X() const { return pos_.x(); }
+  inline double Y() const { return pos_.y(); }
+  inline double Sigma() const { return sigma_; }
```
civanch: @mundim , here as well: not needed duplication of methods.

SimPPS/PPSPixelDigiProducer/plugins/PPSPixelDigiProducer.cc
```diff
@@ -0,0 +1,260 @@
...
+ Description: <one line class summary>
+
+ Implementation:
+     <Notes on implementation>
+*/
+//
+// Original Author:  F.Ferro
+//
+
+#include "boost/shared_ptr.hpp"
```
civanch: @mundim , we use std::shared_ptr everywhere, boost should not be used for that.

SimPPS/PPSPixelDigiProducer/src/RPixChargeShare.cc
```diff
@@ -0,0 +1,162 @@
...
+          double neighbour_pixel_upper_y = 0;
+          double neighbour_pixel_center_x = 0;
+          double neighbour_pixel_center_y = 0;
+          //      Check the hit approach to the neighbours
+          theRPixDetTopology_.pixelRange(
+              k, l, neighbour_pixel_lower_x, neighbour_pixel_upper_x, neighbour_pixel_lower_y, neighbour_pixel_upper_y);
+          neighbour_pixel_center_x = neighbour_pixel_lower_x + (neighbour_pixel_upper_x - neighbour_pixel_lower_x) / 2.;
+          neighbour_pixel_center_y = neighbour_pixel_lower_y + (neighbour_pixel_upper_y - neighbour_pixel_lower_y) / 2.;
+          hit2neighbour[m] =
+              1 / sqrt(pow((*i).X() - neighbour_pixel_center_x, 2.) + pow((*i).Y() - neighbour_pixel_center_y, 2.));
```
civanch: @mundim , better avoid using pow() for power 2 - in some compilers it is transformed to right thing but at the same time easy to write a code without pow in this case.

SimPPS/PPSPixelDigiProducer/src/RPixDummyROCSimulator.cc
```diff
@@ -0,0 +1,79 @@
...
+
+      // Avoid exception due to col > 103 in case of 2x2 plane. To be removed
+      if (col >= DetCalibs.getNCols())
+        continue;
+
+      if (doSingleCalibration_) {
+        adc = int(round(i->second / electron_per_adc_));
+      } else {
+        if (DetCalibs.getDetId() != 0) {
+          gain = DetCalibs.getGain(col, row) * 1800. / 260.;  // *highRangeCal/lowRangeCal
```
civanch: @mundim , 1800, 260 and 255 better to define as const value like highRangeCal, lowRangeCal

SimPPS/PPSPixelDigiProducer/src/RPixLinearChargeDivider.cc
```diff
@@ -0,0 +1,107 @@
...
+
+void RPixLinearChargeDivider::FluctuateEloss(int pid,
+                                             double particleMomentum,
+                                             double eloss,
+                                             double length,
+                                             int NumberOfSegs,
+                                             std::vector<RPixEnergyDepositUnit>& elossVector) {
+  double particleMass = 139.6;  // Mass in MeV, Assume pion
+  pid = std::abs(pid);
+  if (pid != 211) {  // Mass in MeV
```
civanch: @mundim , in PPS we handle only protons, logic of pid and particleMass should correspond for proton only.

SimPPS/PPSPixelDigiProducer/interface/RPixEnergyDepositUnit.h
mundim: It is really needed both kinds of methods, otherwise, a deeper analysis in the code will be needed. Removing one of them (either one) produces compilation errors.

SimPPS/PPSPixelDigiProducer/interface/RPixSignalPoint.h
mundim: It is really needed both kinds of methods, otherwise, a deeper analysis in the code will be needed. Removing one of them (either one) produces compilation errors.

SimPPS/PPSPixelDigiProducer/plugins/PPSPixelDigiProducer.cc
mundim: The code seems to use methods that is not present in the std::shared_ptr, basically the "find" apart from that, moving to std version produces a compilation error concerning a wrong template argument (?). So, if it is mandatory, a longer time will be needed to track down the problems.

SimPPS/PPSPixelDigiProducer/plugins/PPSPixelDigiAnalyzer.cc
mundim: done

SimPPS/PPSPixelDigiProducer/BuildFile.xml
mundim: done. In fact, many other unneeded libraries were removed...

SimPPS/PPSPixelDigiProducer/interface/RPixChargeShare.h
mundim: done.

SimPPS/PPSPixelDigiProducer/src/RPixChargeShare.cc
mundim: What about changing 1/pow(x,2) to pow(x,-2). Would that comply with the rules?

SimPPS/PPSPixelDigiProducer/src/RPixDummyROCSimulator.cc
mundim: done

SimPPS/PPSPixelDigiProducer/src/RPixLinearChargeDivider.cc
mundim: In fact, due to interaction with the material, other particles might occur. Debugging the code it was found many other kinds of particles present.

mundim:
Waiting for comments in the open issues above to commit the changes.

civanch:
@mundim , some my comments are strong (dependency, shared_ptr), another are only code style but I do not think it is difficult to implement all. Access method like "double Energy() const" are practically in all CMSSW classes, if in all classes we will duplicate all methods it will be a mess. Your compilation problem likely only because of usage of const. The method should be const if class members are not changed, if they may be modified, the method should not be declared as "const". If a method return "const&" return reference object should be "const".

SimPPS/PPSPixelDigiProducer/plugins/PPSPixelDigiProducer.cc
```diff
@@ -0,0 +1,260 @@
...
+  void produce(edm::Event&, const edm::EventSetup&) override;
+
+  // ----------member data ---------------------------
+  std::vector<std::string> RPix_hit_containers_;
+  typedef std::map<unsigned int, std::vector<PSimHit>> simhit_map;
+  typedef simhit_map::iterator simhit_map_iterator;
+
+  edm::ParameterSet conf_;
+
+  std::map<uint32_t, boost::shared_ptr<RPixDetDigitizer>> theAlgoMap;  //DetId = uint32_t
```
civanch: @mundim , I suspect that this map is not thread safe. It may look as not very efficient solution but I would propose instead of "boost::shared_ptr" to use "std::unique_ptr", the code will be nearly the same. This will mean that the map is created in each stream separately, a bit more memory but guaranteed thread safety.

SimPPS/PPSPixelDigiProducer/interface/RPixChargeShare.h
mundim: done.

SimPPS/PPSPixelDigiProducer/plugins/PPSPixelDigiProducer.cc
mundim: done

SimPPS/PPSPixelDigiProducer/plugins/PPSPixelDigiProducer.cc
mundim: done

SimPPS/PPSPixelDigiProducer/interface/RPixEnergyDepositUnit.h
mundim: Done... replaced the method with the reference by a setter method.

SimPPS/PPSPixelDigiProducer/interface/RPixSignalPoint.h
mundim: Done. Replaced the reference method by a setter method.

SimPPS/PPSPixelDigiProducer/src/RPixChargeShare.cc
mundim: this suggestion was motivated by an old technique of code optimization, where the inverse is better implemented with negative power.

fabiocos:
please test

civanch:
+1

SimPPS/PPSPixelDigiProducer/plugins/PPSPixelDigiAnalyzer.cc
```diff
@@ -0,0 +1,309 @@
...
+  CTPPSPixelSimTopology theRPixDetTopology_;
+  unsigned int found_corresponding_digi_count_;
+  unsigned int cumulative_cluster_size_[3];
+};
+
+PPSPixelDigiAnalyzer::PPSPixelDigiAnalyzer(const ParameterSet &pset) : theRPixDetTopology_() {
+  label_ = pset.getUntrackedParameter<string>("label");
+  verbosity_ = pset.getParameter<int>("Verbosity");
+#ifdef USE_MIDDLE_OF_PIXEL
+  file = new TFile("CTPPSPixelDigiPlots2.root", "RECREATE");
```
fabiocos: @mundim why not using the https://twiki.cern.ch/twiki/bin/view/CMSPublic/SWGuideTFileService provided by the framework?

fabiocos (commented):
@mundim a few questions. How has this code been really tested? Do you have a reference for results?

fabiocos:
```
10:07 cmsdev25 922> cmsRun FromGun2DigiAnal_.py 
----- Begin Fatal Exception 26-Jul-2019 10:07:23 CEST-----------------------
   [0] Processing the python configuration file named FromGun2DigiAnal_.py
Exception Message:
ImportError: No module named TotemCommon.RandomNumbers_cfi

At:
  /cvmfs/cms-ib.cern.ch/week0/slc7_amd64_gcc700/cms/cmssw/CMSSW_11_0_X_2019-07-24-2300/python/FWCore/ParameterSet/Config.py(649): load
  FromGun2DigiAnal_.py(59): <module>

```

Was this tried? In which environment is this supposed to work? 

```
process.load("Configuration.TotemCommon.g4SimHits_cfi")
```
?

SimPPS/PPSPixelDigiProducer/interface/RPixSignalPoint.h
```diff
@@ -0,0 +1,32 @@
...
+ * An elementar charge point, with position, sigma from diffusion and tof.
+ */
+class RPixSignalPoint {
+public:
+  RPixSignalPoint() : pos_(0, 0), sigma_(0), charge_(0){};
+
+  RPixSignalPoint(double x, double y, double s, double charge) : pos_(x, y), sigma_(s), charge_(charge){};
+
+  inline LocalPoint Position() const { return pos_; }
+  inline double X() const { return pos_.x(); }
```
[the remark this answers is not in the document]
mundim: This code comes from old TOTEM code. For sure I can make such a change, but it would delay the merge even further. I would kindly ask to keep this change as a change for after the merge, would it be possible, please?

SimPPS/PPSPixelDigiProducer/plugins/PPSPixelDigiAnalyzer.cc
mundim: @fabferro 
The same as above for SimPPS/PPSPixelDigiProducer/interface/RPixSignalPoint.h, which mean, adding this to the list of change for next releases, after the merge. We need urgently the PPS full simulation integrated.

mundim:
> ```
> 10:07 cmsdev25 922> cmsRun FromGun2DigiAnal_.py 
> ----- Begin Fatal Exception 26-Jul-2019 10:07:23 CEST-----------------------
> An exception of category 'ConfigFileReadError' occurred while
>    [0] Processing the python configuration file named FromGun2DigiAnal_.py
> Exception Message:
>  unknown python problem occurred.
> ImportError: No module named TotemCommon.RandomNumbers_cfi
> 
> At:
>   /cvmfs/cms-ib.cern.ch/week0/slc7_amd64_gcc700/cms/cmssw/CMSSW_11_0_X_2019-07-24-2300/python/FWCore/ParameterSet/Config.py(649): load
>   FromGun2DigiAnal_.py(59): <module>
> 
> ----- End Fatal Exception -------------------------------------------------
> ```
> 
> Was this tried? In which environment is this supposed to work?
> 
> ```
> process.load("Configuration.TotemCommon.g4SimHits_cfi")
> ```
> 
> ?

This is part of a much bigger PR that was tried in the past, but due to its size, it was requested to split it into smaller pieces. The adopted solution was to make PRs with code only, which would not break the cms chain, but the consequence is that each piece can not be tested until all pieces are integrated. The last PR would merge the configuration files, only then the full PPS simulation would be testable ( @fabferro).

There is already another PR, with the code for the Si strip sensor, then it would be submitted the SimG4CMS/PPS and finally, all the config files that will really include PPS in the CMS simulation.

fabiocos:
@fabferro @mundim sorry, I understand some urgency, but we cannot go beyond some threshold of acceptance. The addition of useless method looks quite simple to remove, probably faster than discussing, I assume that then you need to adapt some other code. Concerning the TFileService, I would say that your approach is outside the basic advices to manage ROOT IO in the FWK, @Dr15Jones please correct me if I misinterpret the instructions. 

Could you please clarify the timescale and target for your production that does not afford to make some relatively simple update?

mundim:
This code is in the process of integration since mid-2018. It was considered too big, so we started to split it in several other PRs. We already missed the deadline for the ultra legacy, mainly because basically there is only one part time person to reply the reviewers comments. 

Right now, I'm trying to implementing the change concerning the TFileService, hopefully it will not take too long.

As for the LocalPoint, the code is very similar to SimTracker/SiPixelDigitizer/plugins/SiPixelDigitizer.{h,cc}. I'll try to reduce the number of methods, using the one from LocalPoint as much as possible.
My concern is not just the change in the code, which is not so simple, I mean, it is not just a few lines, but it will require change in other part of the code, testing, and pushing to git, which will restart all the test chain again, which mean a couple of days to get it merged (if everything goes fine) and I have only a few more days of my vacations to get all the other PRs done.

But, if it is mandatory, let's to do it. Thank you anyway.

Dr15Jones:
> Concerning the TFileService, I would say that your approach is outside the basic advices to manage ROOT IO in the FWK, @Dr15Jones please correct me if I misinterpret the instructions.

@fabiocos I haven't taken a deep look at the code. Could you summarize how this code is using TFileService in a way we do not recommend? Thanks.

mundim:
> > Concerning the TFileService, I would say that your approach is outside the basic advices to manage ROOT IO in the FWK, @Dr15Jones please correct me if I misinterpret the instructions.
> 
> @fabiocos I haven't taken a deep look at the code. Could you summarize how this code is using TFileService in a way we do not recommend? Thanks.

@Dr15Jones, I guess I have put the code compliant with the recommendation, changing to TFileService. Just to mention, it involves an analyzer that is not supposed to be included in the CMS simulation chain, it is part of the validation of PPS Digis.

Dr15Jones:
@mundim thanks for the clarification.

kpedro88:
please test

civanch:
+1

kpedro88:
+1

smuzaffar:
@mundim , @fabiocos looks like this code is not used by any test. It refers to data files ( https://github.com/cms-sw/cmssw/pull/27559/files#diff-aa30946b6ed613a8fd274e39c9ee067eR137-R140 )  of https://github.com/cms-data/SimPPS-PPSPixelDigiProducer/pull/1 but that PR is not merged.

PR #24920, opened 2018-10-18
Add Workflow for multi-run harvesting

schneiml:
This PR adds a workflow that exercises multi-run HARVESTING on DQMIO files as number 137.8.

It is based on the 136 data relvals and mixes 2018C and D data; I think this is not guaranteed to work, but it saves us from adding a second run to the relval data.  Currently, I only added it for one PD and 2018C/D data, that should be enough to catch the obvious problems. We could do the full combinatoric expansion as for the 136 WFs as well.

As expected, this WF fails at the moment, as far as I see for a different issue than the ones that @cerminar observed. PRs to fix the crashes will follow. We can wait with integrating this until all plugins are fixed for multi-run harvesting.

PR #27565, opened 2019-07-19
Remove SiStripBadModuleFedErrESSource

pieterdavid:
#### PR description:

SiStripBadModuleFedErrESSource was directly accessing the DQMStore to build a SiStripQuality object (from a histogram with FED errors) - as pointed out by @makortel this should be avoided since it is not guaranteed to continue working with multi-threaded DQM and EventSetup.
Since the "merged" SiStripQuality with FED error contribution is only used in standalone cases (e.g. to make tracker maps of the number of bad components) over a single run, and the different bad component sources are never more granular than per-run (unless I missed something), the easiest solution seemed to be to derive and merge the FED error SiStripQuality only from the dqmEndJob method (caching the required event setup  products) of SiStripBadComponentInfo and SiStripQualityStatistics.

#### PR validation:

Compiles, and unit tests run. @suchandradutta could you please have a look and test some of the bad component workflows that may be affected by these changes?


jfernan2:
please test

jfernan2:
please test

jfernan2:
@pieterdavid it seems the PR is only modifying SiStrip/EventInfo /BadComponentContents histograms in SOME wf: 4.53  140.56 136.85  136.731 136.788
All are data workflows but not ALL the data workflows are modified.
Is this expected? Thanks

pieterdavid:
@jfernan2 thanks, no, this was supposed to be a purely technical change, so I must have missed something, I will check.

pieterdavid:
With the last commit the difference in the comparison should hopefully disappear (I checked for 136.85 and found a typo in the configuration, and missing calls to the methods that actually merge the SiStripQuality objects - I suppose the problem with the other ones was the same).

jfernan2:
please test

jfernan2:
@pieterdavid now the PR is not modifying any ME... I am not sure if this is expwected somehow...

pieterdavid:
@jfernan2 I'm sorry for not being clearer from the beginning... this is a purely technical change, so no modified ME are expected.
The only case where there would be changes is when SiStripBadComponentInfo or SiStripQualityStatistics sees multiple runs and uses the FED errors as an input (because this is only retrieved at endJob instead of endRun with this PR), but I didn't find any instances of that (it may be that this would not be a well-defined thing to do anyway).

jfernan2:
+1

mmusich:
> The only case where there would be changes is when SiStripBadComponentInfo or SiStripQualityStatistics sees multiple runs and uses the FED errors as an input (because this is only retrieved at endJob instead of endRun with this PR), but I didn't find any instances of that (it may be that this would not be a well-defined thing to do anyway).

@jfernan2 @pieterdavid 
what happens in the case of Multi-Run harvesting? 
Is there a workflow exercising that?

venturia:
Why can't it be harvested in endRun ? Or both endJob and endRun?

pieterdavid:
@mmusich APV are marked "bad" if they had an error in at least some fraction (80% by default) of the events in a run, so unless this is done per-run or the APV is giving errors at a constant rate over the whole run range, the result would be different. For the per-run outputs of SiStripQualityStatistics (it produces the same files with and without the run number included) the FED errors bad component contribution would not be included.
I did not find any configs that actually do this (those that loop on runs don't include the FED errors), but this is why it would be best if the DQM/bad components experts (@suchandradutta et al.) could also have a look at this, and verify it doesn't break their workflows  (if some rely on the SiStripQualityStatistics outputs with the run number to include the FED errors contribution, this change may cause problems).
@venturia DQMEDHarvester only provides dqmEndLuminosityBlock (for lumi-based histograms) and dqmEndJob (for run-based histograms), I followed [this documentation](https://twiki.cern.ch/twiki/bin/view/CMS/DQMOfflineThreadedMigration#The_DQMEDHarvester). From what I see on the [DQMOffline](https://twiki.cern.ch/twiki/bin/view/CMS/DQMOffline) twiki, it is possible to save the DQM histograms per run, but not to run code for each run separately in the harvesting step (I may have missed something, though).

pieterdavid:
@mmusich APV are marked "bad" if they had an error in at least some fraction (80% by default) of the events in a run, so unless this is done per-run or the APV is giving errors at a constant rate over the whole run range, the result would be different. For the per-run outputs of SiStripQualityStatistics (it produces the same files with and without the run number included) the FED errors bad component contribution would not be included.
I did not find any configs that actually do this (those that loop on runs don't include the FED errors), but this is why it would be best if the DQM/bad components experts (@suchandradutta et al.) could also have a look at this, and verify it doesn't break their workflows  (if some rely on the SiStripQualityStatistics outputs with the run number to include the FED errors contribution, this change may cause problems).
@venturia DQMEDHarvester only provides dqmEndLuminosityBlock (for lumi-based histograms) and dqmEndJob (for run-based histograms), I followed [this documentation](https://twiki.cern.ch/twiki/bin/view/CMS/DQMOfflineThreadedMigration#The_DQMEDHarvester). From what I see on the [DQMOffline](https://twiki.cern.ch/twiki/bin/view/CMS/DQMOffline) twiki, it is possible to save the DQM histograms per run, but not to run code for each run separately in the harvesting step (I may have missed something, though).

mmusich:
I am wondering just to make sure there are no issues with the MRH if it is possible to run wf 137.8:
```
137.8  RunEGamma2018C+HLTDR2_2018+RECODR2_2018reHLT_skimEGamma_Offline_L1TEgDQM+RunEGamma2018D+HLTDR2_2018+RECODR2_2018reHLT_skimEGamma_Prompt_L1TEgDQM+HARVEST2018_L1TEgDQM_MULTIRUN 
```
which was created explicitly for this purpose: https://github.com/cms-sw/cmssw/pull/24920

jfernan2:
please test workflow 137.8

christopheralanwest:
please test workflow 137.8

venturia:
Indeed the best thing to do is to ask @suchandradutta and whoever else used this list of bad channels routinely. The purpose is to produced the trend plots of the bad channels, including the ones with too many APV errors, per run. This is done reading a DQM plot (one per run). If what is proposed in this pull request does not break this functionality I am fine.

pieterdavid:
@mmusich APV are marked "bad" if they had an error in at least some fraction (80% by default) of the events in a run, so unless this is done per-run or the APV is giving errors at a constant rate over the whole run range, the result would be different. For the per-run outputs of SiStripQualityStatistics (it produces the same files with and without the run number included) the FED errors bad component contribution would not be included.
I did not find any configs that actually do this (those that loop on runs don't include the FED errors), but this is why it would be best if the DQM/bad components experts (@suchandradutta et al.) could also have a look at this, and verify it doesn't break their workflows  (if some rely on the SiStripQualityStatistics outputs with the run number to include the FED errors contribution, this change may cause problems).
@venturia DQMEDHarvester only provides dqmEndLuminosityBlock (for lumi-based histograms) and dqmEndJob (for run-based histograms), I followed [this documentation](https://twiki.cern.ch/twiki/bin/view/CMS/DQMOfflineThreadedMigration#The_DQMEDHarvester). From what I see on the [DQMOffline](https://twiki.cern.ch/twiki/bin/view/CMS/DQMOffline) twiki, it is possible to save the DQM histograms per run, but not to run code for each run separately in the harvesting step (I may have missed something, though).

[I tried to post this a few hours ago, but github had some issues... I guess I should have marked it 'work in progress' because some more feedback/testing is needed indeed]

pieterdavid:
I rebased to solve the conflicts after https://github.com/cms-sw/cmssw/pull/27476 was merged

jfernan2:
please test workflow 137.8

jfernan2:
+1

christopheralanwest:
please test workflow 137.8

The errors in the AddOn test look like transient file access problems so let's run them again.

schneiml:
> @venturia DQMEDHarvester only provides dqmEndLuminosityBlock (for lumi-based histograms) and dqmEndJob (for run-based histograms), I followed [this documentation](https://twiki.cern.ch/twiki/bin/view/CMS/DQMOfflineThreadedMigration#The_DQMEDHarvester). From what I see on the [DQMOffline](https://twiki.cern.ch/twiki/bin/view/CMS/DQMOffline) twiki, it is possible to save the DQM histograms per run, but not to run code for each run separately in the harvesting step (I may have missed something, though).

`dqmEndJob` will run at the end of each run in all configurations except
the ones doing multi-run harvesting. In multi-run harvesting, all things
that normally happen at end of the run are moved to the end of the job,
so that all DQM produces accumulated output. Multi-run harvesting is not
very well supported in CMSSW and DQM, and there is no (reliable) way to
do things at the end of run when multi-run harvesting.

pohsun:
+1

kpedro88:
+1

PR #27572, opened 2019-07-19
Avoid unrepresentable int value in BitSet

Dr15Jones:
#### PR description:

The bit manipulation used to iterate over the BitSet required a negative value to be negated and the number obtained could not be represented by an int when the bit in question was # 31. The new bit manipulation gives the same results (as tested in the unit test) but avoids the undefined behavior.

This was found by UBSAN.

#### PR validation:

The code compiles and the new unit test succeeds. The unit test also shows that the UBSAN error goes away.


Dr15Jones:
please test

santocch:
+1

fabiocos:
+1

PR #27577, opened 2019-07-20
Avoid dereference of nullptr in METAnalyzer

Dr15Jones:
#### PR description:

This was uncovered by UBSAN.

#### PR validation:

Ran the same workflow which was reporting the problem in the IB and no report happened with this change.


Dr15Jones:
please test

jfernan2:
+1

DQMOffline/JetMET/src/METAnalyzer.cc
```diff
@@ -1504,14 +1504,17 @@ void METAnalyzer::analyze(const edm::Event& iEvent, const edm::EventSetup& iSetu
   //met=&(tcmetcoll->front());
   //}
   if (isPFMet_) {
+    assert(!pfmetcoll->empty());
```
fabiocos: @Dr15Jones @jfernan2 are we expecting this instruction to abort the job? Is this the behaviour that we want in this DQM analyzer? Or should we set a different type of protection?

Dr15Jones: The lines following lead to undefined behavior. In the case of programming errors I use `assert`. In the case of runtime problems I'd use exceptions. At the cursory glance I gave the code, it looked like if you got here with an empty collection it was a logic problem with the algorithm. Hence why I chose `assert`.

fabiocos:
+1

PR #27582, opened 2019-07-22
Improvements to the HGC digitizer and implementation of a end-of-life scenario for the HGC silicon sector

deguio:
#### PR description:

Following PR #26811, this development allows to simulate an end-of-life scenario for Silicon sensors as well. In particular:

- the computation of charge collection efficiency and leakage current makes use of a parametrized fluence map from fluka
- an `HGCalRadiationMap` class is developed with specializations for Si and Sci sectors
- proper treatment of the noise components is added together with the implementation of the gain switch mechanism for the HGCROC
- standalone analyzers are added/updated to allow simple testing of each feature

This set of changes will allow to study the evolution of occupancy and bandwidth for an end-of-life configuration of the HGC silicon and scintillator sectors.

#### PR validation:

the changes included in this PR have been presented and discussed at the HGCal DPG [here](https://indico.cern.ch/event/827633/contributions/3463807/attachments/1861038/3058615/20190612_deguio_HGCalElectronicsSimulation.pdf) and [here](https://indico.cern.ch/event/832849/contributions/3492966/attachments/1878035/3095136/20190709_deguio_HGCalElectronicsSimulation.pdf)

the performance isn't expected to change for the standard non-aged scenario:
- [HGCDigitizer::finalizeEvent - plain 11_0_0_pre2 - non-aged](http://deguio.web.cern.ch/deguio/cgi-bin/igprof-navigator/HGCalDigi/Jul15/outputAlgo2_0_geom.11_0_0_pre2.pp/18)
- [HGCDigitizer::finalizeEvent - plain 11_0_0_pre4+thisPR - non-aged](http://deguio.web.cern.ch/deguio/cgi-bin/igprof-navigator/HGCalDigi/Jul15/outputAlgo2_0_geom.a0e819999d9/18)

while for the aged scenario most of the slow down comes from `HGCalDDDConstants::locateCell` or math operations:
- [HGCDigitizer::finalizeEvent - plain 11_0_0_pre4+thisPR - aged](http://deguio.web.cern.ch/deguio/cgi-bin/igprof-navigator/HGCalDigi/Jul15/outputAlgo2_3000_geom.a0e819999d9/18)

@pfs @rovere @cseez


kpedro88:
please test

kpedro88 (commented):
As a general comment, eventually these maps should be provided by ESProducers in case they need to be accessed in multiple places consistently.

deguio:
hey @kpedro88
I think I've implemented all your comments except the one about the hash function for which I should think more. do you expect non negligible speed-ups from it?

kpedro88:
@deguio it depends how many entries are in the map. Do you have a rough estimate?

deguio:
not many.. that's basically just the content of the file
`doseParams_3000fb_fluka-3.5.15.9.txt`

kpedro88:
> not many.. that's basically just the content of the file `doseParams_3000fb_fluka-3.5.15.9.txt`

Ah okay, if that's the case it probably doesn't matter then. Though in that case, you could just use something like pair.first*1000 + pair.second as the key and then hash that.

deguio:
I could..  ;)

kpedro88:
https://github.com/cms-sw/cmssw/pull/27582/files#r306039084 still needs to be addressed

deguio:
hi @kpedro88 
ok, I've missed it. what if I add a comment to the code to explain the factors?
would something like the following be acceptable?

```
  //formula is: A = A0 * exp( -D^0.65 / 199.6)                                                                                                                                                
  //where A0 is the response of the undamaged detector, D is the dose                                                                                                                               
  int layer = cellId.layer();
  double cellDose = getDoseValue(DetId::HGCalHSc, layer, radius);  //in kRad                                                                                                                  
  constexpr double expofactor = 1. / 199.6;
  double scaleFactor = std::exp(-std::pow(cellDose, 0.65) * expofactor);

  //formula is: N = 2.18 * sqrt(F * A / 2e13)                                                                                                                                                 
  //where F is the fluence and A is the SiPM area                                                                                                                                             
  double cellFluence = getFluenceValue(DetId::HGCalHSc, layer, radius);  //in 1-Mev-equivalent neutrons per cm2                                                                               

  constexpr double factor = 2. / (2 * 1e13);  //SiPM area = 2mm^2                                                                                                                             
  double noise = 2.18 * sqrt(cellFluence * factor);
```

kpedro88:
http://cms-sw.github.io/cms_coding_rules.html

7.6 specifies that magic numbers should be replaced with named constants.

deguio:
hi @kpedro88 
ok, but those are already in a `constexpr` and I'm not sure what meaningful name I could give to the single numbers (and I'm not sure if that's useful). I can make the `2.18` const, but that's the only one I see which is not `const` or part of a `constexpr`.
did you have something else in mind?

kpedro88:
0.65 should also be a named constant

deguio:
ok, I'll do that.

kpedro88:
please test

kpedro88:
please test

civanch:
@deguio , should  SimCalorimetry/HGCalSimProducers/data/doseParams_3000fb_fluka-3.5.15.9.txt be part of this PR or may be better to keep this file as an external data? The decision depends on the size of the file and on the estimated possible number of such files.

deguio:
Hi Vladimir,
that file shouldn't grow anymore at this point so I hope it is acceptable to have it in cmssw for now. it allows to compute dose and fluence per cell needed for the simulation of the aged detector.

kpedro88:
@deguio as we move closer to integrating this PR, can you follow the instructions from the bot for the new package? The categories should be simulation and upgrade.
> Please create a PR for https://github.com/cms-sw/cms-bot/blob/master/categories_map.py to assign category

deguio:
hi @kpedro88 
here it is: https://github.com/cms-sw/cms-bot/pull/1177

civanch:
+1

kpedro88:
@deguio somehow almost all of the silicon hits are gone. I do not think this is intended.

HGCEE RecHits
HGCHEF RecHits
HGCHEB RecHits

kpedro88:
ping @cmsbuild

deguio:
@kpedro88
I admit that I didn’t check rechits when I validated this pr, but indeed this looks strange. I’m traveling for one week with limited connectivity. I’ll look into it next week at some point. 
Thanks for reporting.

kpedro88:
hold

deguio:
hi @kpedro88 
the issue should be cured with the last commit. there was a bug in the comparison of the charge with the ZS threshold where different units (fC vs ADC) were compared. the comparison is now performed in ADC everywhere which is a more realistic behaviour. the validation against the baseline should now give good agreement for a non-aged scenario except for some rounding effect due to the above.

@pfs

fabiocos:
please test

fabiocos:
given the size of the .txt file 
```
13:50 cmsdev25 2347> du -h -a SimCalorimetry/HGCalSimProducers/data/
8.0K    SimCalorimetry/HGCalSimProducers/data/doseParams_3000fb_fluka-3.5.15.9.txt
4.0K    SimCalorimetry/HGCalSimProducers/data/sipmParams_geom-10.txt
16K     SimCalorimetry/HGCalSimProducers/data/
```
I do not see the need for an external package

civanch:
+1

kpedro88:
unhold

SimCalorimetry/HGCalSimProducers/interface/HGCFEElectronics.h
```diff
@@ -34,7 +34,7 @@ class HGCFEElectronics {
                         hgc::HGCSimHitData& chargeColl,
                         hgc::HGCSimHitData& toa,
                         CLHEP::HepRandomEngine* engine,
-                        int thrADC = -1,
+                        uint32_t thrADC = -1,
```
kpedro88: This is effectively setting the default value to the highest possible uint32_t. Is this desirable?

kpedro88:
The RecHit distributions look mostly reasonable now. There are some changes at very low energy that are probably due to the ADC rounding that was mentioned.

HGCEE RecHits
HGCHEF RecHits
HGCHEB RecHits

deguio:
thanks @kpedro88,
things look good from our side. if there are no other comments could we have this PR merged soon?

SimCalorimetry/HGCalSimProducers/interface/HGCFEElectronics.h
kpedro88: @deguio can you comment on this?

deguio: thanks @kpedro88 for catching this. this is now fixed and initialized to 0 in case a value is not explicitly passed.

deguio:
hi @kpedro88 
I've added one commit. its content is:
- unify the `thresholdFollowsMIP` parameter for Si and Sci removing the `calibDigi` one
- turn the `thresholdFollowsMIP` on for all subdetectors (this was intended since the beginning and done in our tests, but not propagated to the branch)
- fix a bug in how the ADC threshold is computed for Scintillators

I hope there won't be additional adjustments on top of this.

deguio:
@kpedro88 
yes I can squash the 25 commits into few. I'm running some tests just to make sure I'm not missing anything else. will do that later today.

deguio:
ok, it should be done now

deguio:
yes, I'm fixing it..

kpedro88:
+upgrade

kpedro88:
@civanch please sign (hopefully for the final time)

civanch:
+1

fabiocos:
+1

PR #27585, opened 2019-07-22
Phase2-hgx207 Have uniform wafer type from HGCalDDDConst and reduce usage of magic numbers

bsunanda:
#### PR description:

Have uniform wafer type definition from HGCalDDDConst and reduce usage of magic numbers

#### PR validation:

Use standard matrix for testing this

Nothing special

### DIFF
--- a/Geometry/HGCalCommonData/src/HGCalDDDConstants.cc
+++ b/Geometry/HGCalCommonData/src/HGCalDDDConstants.cc
@@ -372,8 +372,6 @@ HGCalParameters::hgtrap HGCalDDDConstants::getModule(unsigned int indx, bool hex
                                    << " ***** ERROR *****";
     unsigned int type =
         ((indx < hgpar_->waferTypeL_.size()) ? hgpar_->waferTypeL_[indx] - 1 : HGCSiliconDetId::HGCalCoarseThick);
-    //  if (type > 0)
-    //    --type;
     mytr = hgpar_->getModule(type, reco);
   } else {
     mytr = hgpar_->getModule(indx, reco);
--- a/Geometry/HGCalCommonData/src/HGCalDDDConstants.cc
+++ b/Geometry/HGCalCommonData/src/HGCalDDDConstants.cc
@@ -370,9 +370,10 @@ HGCalParameters::hgtrap HGCalDDDConstants::getModule(unsigned int indx, bool hex
       edm::LogWarning("HGCalGeom") << "Wafer no. out bound for index " << indx << ":" << (hgpar_->waferTypeL_).size()
                                    << ":" << (hgpar_->waferPosX_).size() << ":" << (hgpar_->waferPosY_).size()
                                    << " ***** ERROR *****";
-    unsigned int type = ((indx < hgpar_->waferTypeL_.size()) ? hgpar_->waferTypeL_[indx]-1 : HGCSiliconDetId::HGCalCoarseThick);
-//  if (type > 0)
-//    --type;
+    unsigned int type =
+        ((indx < hgpar_->waferTypeL_.size()) ? hgpar_->waferTypeL_[indx] - 1 : HGCSiliconDetId::HGCalCoarseThick);
+    //  if (type > 0)
+    //    --type;
     mytr = hgpar_->getModule(type, reco);
   } else {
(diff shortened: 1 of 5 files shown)


bsunanda:
@cmsbuild Please test

Geometry/HGCalCommonData/src/HGCalDDDConstants.cc
```diff
@@ -369,9 +370,10 @@ HGCalParameters::hgtrap HGCalDDDConstants::getModule(unsigned int indx, bool hex
       edm::LogWarning("HGCalGeom") << "Wafer no. out bound for index " << indx << ":" << (hgpar_->waferTypeL_).size()
                                    << ":" << (hgpar_->waferPosX_).size() << ":" << (hgpar_->waferPosY_).size()
                                    << " ***** ERROR *****";
-    unsigned int type = ((indx < hgpar_->waferTypeL_.size()) ? hgpar_->waferTypeL_[indx] : 3);
-    if (type > 0)
-      --type;
+    unsigned int type =
+        ((indx < hgpar_->waferTypeL_.size()) ? hgpar_->waferTypeL_[indx] - 1 : HGCSiliconDetId::HGCalCoarseThick);
+    //  if (type > 0)
```
kpedro88: delete commented-out code

bsunanda: Done

bsunanda:
@cmsbuild Please test

kpedro88:
+upgrade

Geometry/HGCalCommonData/src/HGCalDDDConstants.cc
```diff
@@ -1241,15 +1246,15 @@ std::pair<double, double> HGCalDDDConstants::waferPosition(int waferU, int wafer
 int HGCalDDDConstants::waferType(DetId const& id) const {
```
perrotta: The return value of this function is the previous one minus 1. This is correctly propagated to the HGCalDigitizer, the only users of this method as far as I could tell: therefore the net effect is that no changes are expected. (OK!)

Geometry/HGCalCommonData/src/HGCalDDDConstants.cc
```diff
@@ -1241,15 +1246,15 @@ std::pair<double, double> HGCalDDDConstants::waferPosition(int waferU, int wafer
...
   } else if ((mode_ == HGCalGeometryMode::Hexagon) || (mode_ == HGCalGeometryMode::HexagonFull)) {
-    type = waferTypeL(HGCalDetId(id).wafer());
+    type = waferTypeL(HGCalDetId(id).wafer()) - 1;
   }
   return type;
 }
 
 int HGCalDDDConstants::waferType(int layer, int waferU, int waferV) const {
-  int type(-1);
+  int type(HGCSiliconDetId::HGCalCoarseThick);
```
perrotta: Here your are changing the default value from "-1" to "2". Is that default value ever accessed? Was anything else modified in the rest of the code to take into account this change? (I couldn't easily find it...)

Geometry/HGCalCommonData/src/HGCalDDDConstants.cc
bsunanda: This function has not changed at all except it uses enum values instead of numbers. The o/p is indeed unchanged

Geometry/HGCalCommonData/src/HGCalDDDConstants.cc
bsunanda: -1 does not mean anything - it was used earlier so that all variables are initialized to some value. That default value is never accessed

perrotta: Ok, thank you Sunanda. I just wanted to be sure that such a modified initialization default is never accessed, which is what I expected but not obvious from the code itself.

perrotta:
+1
- Centralized enum values replace previous hardcoded numbers, with the same meaning
- Jenkins tests pass and show no difference with respect to the baseline, as it should

cvuosalo:
+1

civanch:
+1

fabiocos:
+1

PR #27592, opened 2019-07-24
Extended test on Geant4 geometry 

civanch:
#### PR description:
Extended test on Geant4 geometry is added, it is expected , that the test will help in Phase2 development and for DD4Hep migration. As a result of test few files are created with dumps of information: 
   1) elements/material/cuts
   2) Geant4 regions
   3) Geant4 mother volumes (according to request it may be a world or some logical volume)
   4) Geant4 overlap check

Some classes are moved from SimG4Core/Application to SimG4Core/Geometry in order to have a correct dependecy between sub-libraries. Some files are renamed to avoid classes started from "G4".

Mainstream simulation should not be affected. 

#### PR validation:
private


civanch:
please test

civanch:
please test

civanch:
please test

civanch:
+1

SimG4Core/Application/python/g4SimHits_cfi.py
```diff
@@ -52,12 +52,20 @@
     PhysicsTablesDirectory = cms.string('PhysicsTables'),
     StorePhysicsTables = cms.bool(False),
     RestorePhysicsTables = cms.bool(False),
-    CheckOverlap = cms.untracked.bool(False),
+    CheckGeometry = cms.bool(False),
```
fabiocos: @civanch why this and other verbosity flags are moved from untracked to tracked? What is the rationale behind it?

SimG4Core/Application/python/g4SimHits_cfi.py
```diff
@@ -81,7 +89,7 @@
     ),
     MagneticField = cms.PSet(
         UseLocalMagFieldManager = cms.bool(False),
-        Verbosity = cms.untracked.bool(False),
+        Verbosity = cms.bool(False),
```
fabiocos: @civanch similar question here for instance

SimG4Core/Application/python/g4SimHits_cfi.py
```diff
@@ -118,7 +126,7 @@
         type = cms.string('SimG4Core/Physics/FTFP_BERT_EMM'),
         DummyEMPhysics = cms.bool(False),
         CutsPerRegion = cms.bool(True),
-        CutsOnProton  = cms.untracked.bool(True),
+        CutsOnProton  = cms.bool(True),
```
fabiocos: @civanch this I understand more

fabiocos (commented):
@civanch a very useful piece of work, I have just a couple of questions about the change of nature of some parameters

fabiocos:
@civanch if I want to recursively navigate into the volume hierarchy as before which settings should I provide now?

SimG4Core/Application/python/g4SimHits_cfi.py
civanch: @fabiocos, "verbosity" parameters were added by different peoples and historically some are tracked, another untracked. The way how these parameters are used is well defined: only for debuging. They mainly are inside this base python file. Likely some may be removed, another may stay "tracked".

civanch:
@fabiocos, in this test we may call Geant4 facility to check overlaps starting from some physical volume, it will go via full depth but due to migration to VecGeom this become extremely slow. With native Geant4 geometry this check takes ~20 minutes for full CMS and > 1 day for VecGeom. The possibility to limit depth allowing fast check on volume of interest. If depth=-1 full check will be applied.

In parallel here we have a possibility to make our own printout of physical and/or logical volumes, we define how deep we want to go. I need consult with Geant4 geometry experts if Geant4 internal facility still exist but experts are at vacations.

fabiocos:
@civanch ok, I see that with Depth set to -1 I may recover for instance a full scan of the FastTimerRegion. Of course it isn't fast, but it depends on what we need to check.

fabiocos:
+1

PR #27612, opened 2019-07-25
Run3-sim37 Enable the useful analyzer to run in the same step as simulation

bsunanda:
#### PR description:

Enable the useful analyzers in SimG4CMS/Calo to run in the same step as simulation

#### PR validation:

The crashes in earlier runs are removed. It is essentially rearrangement of files. The standard workflows are unaffected.

Nothing special


bsunanda:
@cmsbuild Please test

civanch:
+1

fabiocos:
@bsunanda what is the plan here? To optionally add some of these new plugins in some test workflow? I would imagine this is not for production

bsunanda:
@fabiocos The analyzers are used in validating newer versions of Geant4. These are run in the Sim step. But with current biglib's this does not work if the analyzers stay in the test directory - that is why they are moved to plugins and now one can run the analyzers in the same step. It started happening in 10_6 (or may be even in 10_5). Only now we understand the reason of earlier crashes. So these analyzers will not be added in standard production runs but in testing newer Geant4 or even GeantV versions.

kpedro88:
+1

PR #27613, opened 2019-07-25
Run3-sim38 Enable plugins from the TB areas to work

bsunanda:
#### PR description:

Enlarge biglib of simulation to make it work for the test beam applications

#### PR validation:

Tested it with TB applications

Nothing special

### DIFF
--- a/BigProducts/Simulation/BuildFile.xml
+++ b/BigProducts/Simulation/BuildFile.xml
@@ -1,6 +1,8 @@
 <use name="SimG4CMS/Calo"/>
 <use name="SimG4CMS/FP420"/>
 <use name="SimG4CMS/Forward"/>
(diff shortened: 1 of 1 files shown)


bsunanda:
@cmsbuild Please test

Dr15Jones:
@smuzaffar could you review this one?

smuzaffar:
+core
yes, these packaegs should be part of biglib as they link against geant4. 
@bsunanda , looks like these two packages[a] are only used by their plugins. May be better to move the src and interface of these to plugins directory (may be in a different PR).

[a]
SimG4CMS/HcalTestBeam
SimG4CMS/HGCalTestBeam

kpedro88:
+1
I find it a bit odd that BigProducts/Simulation is not in the simulation category...

PR #27619, opened 2019-07-26
including ctpps in DQMOffline standard conf

fabferro:
#### PR description:

Change of DQM configuration in order to have (CT)PPS properly configured also for Offline plots. 

#### PR validation:

Tested on data with runTheMatrix on workflows 136.867 (2018) and 136.801 (2017)

@forthommel @jfernan2

PR #27620, opened 2019-07-26
[106x] including ctpps in DQMOffline standard conf

fabferro:
#### PR description:

Change of DQM configuration in order to have (CT)PPS properly configured also for Offline plots.
To be tested with #27591 

#### PR validation:

Tested on data with runTheMatrix on workflows 136.867 (2018) and 136.801 (2017)

@forthommel @jfernan2

Backport of #27619


jfernan2:
please test

fabferro:
The tests failing are Run1 MC. Crash due to missing CTPPS geometry record in the GT. No CTPPS during run 1, so that's normal. How can we prevent these "normal" crashes??

jfernan2:
@fabferro  you can drive that with Eras, like the lines you removed but in the proper place

fabferro:
@jfernan2: ok, any suggesting on where?

jfernan2:
What about in DQMOffline_cff but like Tau in lines 186-188? @fiorintu what do hoy think?

fioriNTU:
@jfernan2 @fabferro if you need some general info about the use of eras, please have a look here https://twiki.cern.ch/twiki/bin/view/CMSPublic/SWGuideCmsDriverEras

I think the best place where to set the era modifiers is in ctppsDQM_cff.py , it is more clean. I think what you have to do is to define an empty sequence for Run1 data and include a modifier, say "ctppsRun1" which switch to the empty sequence in the case run1 data are used. I need to dig a bit in the code to find some suitable example. 

You can have a look to the Pixel code, as an example, this is the sequence ran in Run1 https://github.com/cms-sw/cmssw/blob/master/DQM/SiPixelCommon/python/SiPixelOfflineDQM_source_cff.py#L121 with the old detector. Then era modifiers are defined here for the varius data taking cases https://github.com/cms-sw/cmssw/blob/master/DQM/SiPixelCommon/python/SiPixelOfflineDQM_source_cff.py#L121
At this point it is sufficient to import the modifier in the DQMOffline_cff.py (https://github.com/cms-sw/cmssw/blob/master/DQMOffline/Configuration/python/DQMOffline_cff.py#L72) and automatically the configuration switches to the Phase1 once the era is set to 2017 or later. I would define a default empty sequence for ctppsDQM, and then modify it to the actual sequence in eras > 2016 (here I am not sure when ctpps entered CMS data). I can help more on Monday

fabferro:
@fioriNTU @jfernan2: I've implemented some changes in ctppsDQM_cff that enable ctppsDQM only in Run2. The default is an empty sequence. I could not find a Run 1 era to be used to disable ctpps DQM in Run 1 only case. If we stay with this change, a new modification will have to be scheduled for Run 3.
Let me know...

fioriNTU:
@fabferro this looks ok to me, thanks.

fioriNTU:
please test

fioriNTU:
+1

fioriNTU:
@fabiocos would be great to integrate this PR as soon as possible, it could enter UL 2017 after the stop. Is there any reason to not merge it?

fabiocos:
@fioriNTU this code has never run in a RelVal campaign, as it is part of 11_0_0_pre5 just built yesterday. At this point we need to be careful about how we treat backports in 10_6_X. Anyway I have already notified PPD about possible further additions, and this is going to be discussed in a dedicated meeting today

jfernan2:
@fabiocos I am not sure if this was finally discussed in a dedicated meeting but it seems from PPD now that this PR is needed in 10_6_X for UL2018 production.

srimanob:
PPD validates this PR with relvals from 11_0_0_pre5, and it seems to be OK.

2018 data) 
Validation (private) OK for
11_0_0_pre5 + 110X_dataRun2_v3

We will ask them to fill the valdb.

fabiocos:
+1

PR #27625, opened 2019-07-29
shared_ptr has arrived to the standard. NFC

vgvassilev:
We replace the CPP11_shared_ptr macro with its actual expansion to
std::shared_ptr nowadays.

We replace the inclusion of <CPP11_shared_ptr.hh> with <memory> and order
the includes from less generic to generic ones where applicable.

Depends on: #27626

cc @davidlange6

PR #27626, opened 2019-07-29
shared_ptr has arrived to the standard. NFC

vgvassilev:
We replace the CPP11_shared_ptr macro with its actual expansion to
std::shared_ptr nowadays.


davidlange6:
please test

fabiocos:
@vgvassilev this PR is effectively embedded into #27625 so if that is merged this one becomes useless, right?

fabiocos:
effectively merged as the commit was the same as in #27625

santocch:
+1

PR #27610, opened 2019-07-25
Update RivetInterface for Rivet 3.x

mseidel42:
#### PR description:

* Updated interface for new major Rivet release 3.x
* Needs https://github.com/cms-sw/cmsdist/pull/5107
* Add multiweight handling for PS (default) and ME/PDF (optional) variations.

#### PR validation:

* Processed Powheg ttbar with PS, ME, and PDF (NNPDF31_nnlo) uncertainties:
image

* Checked that output of `ParticleLevelProducer` remains as expected.

PR #27634, opened 2019-07-29
Add energy regression and particle identification to RecoHGCal/TICL.

riga:
#### PR description

This PR adds energy regression and particle identification for trackster objects to the `TrackstersProducer` of the `RecoHGCal/TICL` package. Both the regressed energy and ID probabilities for different types of particles (e, mu, gamma, hadron) are the outputs of a single neural network, stored as a TensorFlow model. The first version of this model is proposed in

https://github.com/cms-data/RecoHGCal-TICL/pull/1.

It does not yet produce a regressed energy, however, this PR allows the `TrackstersProducer` to be configured such that it skips the regression / ID / both. Therefore it does not require any changes in the future once the regression is included in the model.

The inference itself supports multi-threading by means of the usual splitting of TensorFlow `Graph` (stored in the global cache) and `Session` objects (one session per stream).

I ran clang-format on the parts of the code I added or updated.


DataFormats/HGCalReco/interface/Trackster.h
```diff
@@ -24,6 +24,21 @@ namespace ticl {
...
+    std::array<float, 4> id_probabilities;
+
+    // convenience methods to return certain id probabilities
+    inline float photon_probability() { return id_probabilities[0]; };
+
+    inline float electron_probability() { return id_probabilities[1]; };
+
+    inline float muon_probability() { return id_probabilities[2]; };
+
+    inline float hadron_probability() { return id_probabilities[3]; };
```
slava77: is this final?
Will there be a charged_hadron vs neutral_hadron if tracks are added to the game?

RecoHGCal/TICL/plugins/PatternRecognitionbyCA.h
```diff
@@ -6,13 +6,16 @@
 #include <memory>  // unique_ptr
 #include "RecoHGCal/TICL/interface/PatternRecognitionAlgoBase.h"
 #include "RecoLocalCalo/HGCalRecAlgos/interface/RecHitTools.h"
+#include "PhysicsTools/TensorFlow/interface/TensorFlow.h"
+
+namespace tf = tensorflow;
```
slava77: this is pretty close to "using namespace" in a header file, which is not allowed.

slava77:
please test workflow 20493.52 with cms-data/RecoHGCal-TICL#1

kpedro88:
@riga the bot reports that there are still code-format checks that need to be applied. The tests cannot start until this is done.

RecoHGCal/TICL/plugins/PatternRecognitionbyCA.h
riga: Ok. I will remove it also in the .cc files for consistency.

DataFormats/HGCalReco/interface/Trackster.h
riga: Good point. I will split this into `charged_hadron_probability()` and `neutral_hadron_probability()` and adjust the current model so that it outputs 5 values with the last one being constant 0 for the moment.

riga:
@kpedro88 Sorry for the delay, I had to apply some changes to the model added in https://github.com/cms-data/RecoHGCal-TICL/pull/1.

felicepantaleo:
@ZihengChen @tonydp03 fyi

DataFormats/HGCalReco/interface/Trackster.h
slava77: to avoid remapping, to the PF particle types, perhaps it would make sense to at least magically align them to the same order (ch, el, mu, ph, nh).

BTW, are these probabilities expected to be uniquely usable only for the Trackster class? I'm thinking if there is a good reason not to wrap them in a struct.

slava77:
please test workflow 20493.52 with cms-data/RecoHGCal-TICL#1

kpedro88:
@smuzaffar @mrodozov can you check this?

smuzaffar:
@kpedro88 , we are looking in to it.

mrodozov:
please test workflow 20493.52 with https://github.com/cms-sw/cmsdist/pull/5120, cms-data/RecoHGCal-TICL#1

RecoHGCal/TICL/plugins/PatternRecognitionbyCA.cc
```diff
@@ -90,4 +102,134 @@ void PatternRecognitionbyCA::makeTracksters(const edm::Event &ev,
...
+  if (!eidOutputNameId_.empty()) {
+    outputNames.push_back(eidOutputNameId_);
+  }
+
+  // fill input tensor (4)
+  for (int i : tracksterIndices) {
+    // get features per layer and cluster and store them in a nested vector
+    // this is necessary since layer information is stored per layer cluster, not vice versa
+    // also, this allows for convenient re-ordering
+    std::vector<std::vector<std::array<float, 3> > > tracksterFeatures;
```
kpedro88: The memory locality of nested vectors is poor. Maybe unavoidable in this case, but it would be good to consider better structures.

kpedro88: Also, the number 3 is repeated a lot. Probably better to make it a named constant.

RecoHGCal/TICL/plugins/PatternRecognitionbyCA.cc
```diff
@@ -90,4 +102,134 @@ void PatternRecognitionbyCA::makeTracksters(const edm::Event &ev,
...
+
+    // store id probabilities per trackster (6)
+    if (!eidOutputNameId_.empty()) {
+      // get the pointer to the id probability tensor, dimension is batch x 5
+      // (photon, electron, muon, charged hadron, neutral hadron)
+      size_t probsIdx = eidOutputNameEnergy_.empty() ? 0 : 1;
+      float *probs = outputs[probsIdx].flat<float>().data();
+
+      for (int i : tracksterIndices) {
+        tracksters[i].id_probabilities[0] = *(probs++);
```
kpedro88: this could be done in a loop

RecoHGCal/TICL/plugins/TrackstersProducer.cc
```diff
@@ -43,8 +50,29 @@ class TrackstersProducer : public edm::stream::EDProducer<> {
 };
 DEFINE_FWK_MODULE(TrackstersProducer);
 
-TrackstersProducer::TrackstersProducer(const edm::ParameterSet& ps)
-    : myAlgo_(std::make_unique<PatternRecognitionbyCA>(ps)) {
+std::unique_ptr<TrackstersCache> TrackstersProducer::initializeGlobalCache(const edm::ParameterSet& params) {
+  // this method is supposed to create, initialize and return a TrackstersCache instance
+  TrackstersCache* cache = new TrackstersCache(params);
```
kpedro88: don't use `new`. Just define the `unique_ptr` right away:
```
auto cache = std::make_unique<TrackstersCache>(params);
```

riga: Ok, will add that in the next commit.

RecoHGCal/TICL/plugins/PatternRecognitionbyCA.cc
riga: ✔︎

RecoHGCal/TICL/plugins/PatternRecognitionbyCA.cc
riga: True. We might get rid of that nested structure in a future PR anyway. Maybe it is okay to keep it for the moment.

riga:
Sorry for the single pushes, the PR should be ready now for testing.

perrotta:
please test workflow 20493.52 with cms-sw/cmsdist#5120, cms-data/RecoHGCal-TICL#1

DataFormats/HGCalReco/src/classes_def.xml
```diff
@@ -1,7 +1,7 @@
 
 <lcgdict>
-  <class name="ticl::Trackster" ClassVersion="3">
-   <version ClassVersion="3" checksum="3878166595"/>
```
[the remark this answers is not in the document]
riga: Ok, I added it again.

mrodozov:
allow @riga test rights

riga:
please test workflow 20493.52 with cms-sw/cmsdist#5120, cms-data/RecoHGCal-TICL#1

DataFormats/HGCalReco/src/classes_def.xml
```diff
@@ -3,6 +3,9 @@
   <class name="ticl::Trackster" ClassVersion="3">
    <version ClassVersion="3" checksum="3878166595"/>
   </class>
+  <class name="ticl::Trackster" ClassVersion="4">
```
perrotta: The exact syntax should be (see https://twiki.cern.ch/twiki/bin/view/CMSPublic/SWGuideCreatingNewProducts#Class_versioning):

```
  <class name="ticl::Trackster" ClassVersion="4">
   <version ClassVersion="4" checksum="522492601"/>
   <version ClassVersion="3" checksum="3878166595"/>
  </class>
```

riga: Thanks for pointing me to the docs, I fixed it in the last commit.

DataFormats/HGCalReco/src/classes_def.xml
```diff
@@ -1,10 +1,8 @@
 
 <lcgdict>
   <class name="ticl::Trackster" ClassVersion="3">
```
perrotta: No!
The default version must remain the most recent one, i.e. ```ClassVersion="4"```

riga:
please test workflow 20493.52 with cms-sw/cmsdist#5120, cms-data/RecoHGCal-TICL#1

perrotta:
The error does not depend on this PR, see https://github.com/cms-sw/cmssw/pull/27691#issuecomment-520812883

perrotta:
please test workflow 20493.52 with cms-sw/cmsdist#5120, cms-data/RecoHGCal-TICL#1,cms-sw/cmsdist#5107

intrepid42:
Funny, this one looks like the opposite problem compared to this morning. It seems to pick up the Rivet changes from cmsdist but not from cmssw.

So you can try to test with https://github.com/cms-sw/cmssw/pull/27610

perrotta:
please test
(let retry in a fresh IB)

RecoHGCal/TICL/plugins/PatternRecognitionbyCA.h
```diff
@@ -31,6 +34,15 @@ namespace ticl {
     int missing_layers_;
     int min_clusters_per_ntuplet_;
     float max_delta_time_;
+    tensorflow::Session* eidSession_;
+    std::string eidInputName_;
+    std::string eidOutputNameEnergy_;
+    std::string eidOutputNameId_;
+    double eidMinClusterEnergy_;
```
perrotta: This ```eidMinClusterEnergy_``` is not got as parameter in ```PatternRecognitionbyCA::PatternRecognitionbyCA```, and as such it remains uninitialized: please add
By the way: why double, if it is only to be compared with float's? (see ```max_delta_time_``` above)

RecoHGCal/TICL/plugins/PatternRecognitionbyCA.cc
```diff
@@ -90,4 +102,133 @@ void PatternRecognitionbyCA::makeTracksters(const edm::Event &ev,
...
+      for (int cluster = 0; cluster < eidNClusters_; cluster++) {
+        // if there are not enough clusters, fill zeros
+        if (cluster < (int)tracksterFeatures[layer].size()) {
+          std::array<float, eidNFeatures_> &features = layerData[cluster];
+          for (int j = 0; j < eidNFeatures_; j++) {
+            *(inputData++) = features[j];
+          }
+        } else {
+          for (int j = 0; j < eidNFeatures_; j++) {
+            *(inputData++) = float(0);
```
perrotta: ```0f```

RecoHGCal/TICL/python/ticl_iterations.py
```diff
@@ -54,7 +54,14 @@ def TICL_iterations_withReco(process):
       missing_layers = 2,
       min_clusters_per_ntuplet = 15,
       min_cos_theta = 0.94, # ~20 degrees
-      min_cos_pointing = 0.7
+      min_cos_pointing = 0.7,
+      eid_graph_path = cms.string("RecoHGCal/TICL/data/tf_models/energy_id_v0.pb"),
```
perrotta: These are all the default values for the newly added parameters, that you don't need to overwrite here in these explicit configs (the same as you do for ```TrackstersMIP```)

RecoHGCal/TICL/python/ticl_iterations.py
```diff
@@ -108,7 +115,14 @@ def TICL_iterations(process):
...
```
perrotta: same here

perrotta (commented):
Sorry, I didn't notice before the issue with eidMinClusterEnergy_, which has to be fixed (together with the other minor ones I'm also listing).
I will re-launch the tests with the TICL workflow once these fixes will be implemented

RecoHGCal/TICL/plugins/PatternRecognitionbyCA.h
riga: ✔︎ I set it to float now.

RecoHGCal/TICL/plugins/PatternRecognitionbyCA.cc
riga: Changed (to `0.f`).

RecoHGCal/TICL/python/ticl_iterations.py
riga: ✔︎

RecoHGCal/TICL/python/ticl_iterations.py
riga: ✔︎

perrotta:
please test workflow 20493.52 with cms-data/RecoHGCal-TICL#1

riga:
please test workflow 20493.52 with cms-sw/cmsdist#5120, cms-data/RecoHGCal-TICL#1

DataFormats/HGCalReco/interface/Trackster.h
```diff
@@ -24,6 +24,23 @@ namespace ticl {
     // the first element is on the inner layer and
     // the outer element is on the outer layer.
     std::vector<std::array<unsigned int, 2> > edges;
+
+    // regressed energy
+    float regressed_energy;
+
+    // trackster ID probabilities
+    std::array<float, 5> id_probabilities;
```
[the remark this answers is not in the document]
hatakeyamak: Sorry for being late.

We do have enum ParticleType defined in: https://cmssdt.cern.ch/lxr/source/DataFormats/ParticleFlowCandidate/interface/PFCandidate.h
I haven't looked at this PR closely, but I believe it will help if TICL's particle ID uses the same convention when we try to integrate TICL in PF as a whole.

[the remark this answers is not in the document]
riga: Thanks for the reference.

We will probably need the two additional classes (`unknown` and `ambiguous`) internally in  TICL for further studies. However, there will be a PFTICLProducer (@steggema) that is fully compliant with the ParticleType enum.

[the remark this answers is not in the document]
hatakeyamak: OK. I will keep eyes on it when it comes out.

DataFormats/HGCalReco/interface/Trackster.h
```diff
@@ -24,6 +24,27 @@ namespace ticl {
     // the first element is on the inner layer and
     // the outer element is on the outer layer.
     std::vector<std::array<unsigned int, 2> > edges;
+
+    // regressed energy
+    float regressed_energy;
+
+    // trackster ID probabilities
+    std::array<float, 7> id_probabilities;
```
kpedro88: I would still prefer to have these probabilities use standard enums. The ParticleType from PF has an "undefined" category, though not an "ambiguous" category: https://github.com/cms-sw/cmssw/blob/master/DataFormats/ParticleFlowCandidate/interface/PFCandidate.h#L43-L52

RecoHGCal/TICL/plugins/PatternRecognitionbyCA.cc
```diff
@@ -90,4 +103,133 @@ void PatternRecognitionbyCA::makeTracksters(const edm::Event &ev,
...
+  if (!eidOutputNameId_.empty()) {
+    outputNames.push_back(eidOutputNameId_);
+  }
+
+  // fill input tensor (4)
+  for (int i : tracksterIndices) {
+    // get features per layer and cluster and store them in a nested vector
+    // this is necessary since layer information is stored per layer cluster, not vice versa
+    // also, this allows for convenient re-ordering
+    std::vector<std::vector<std::array<float, eidNFeatures_> > > tracksterFeatures;
```
kpedro88: Repeating comment: memory locality of nested vectors is poor. It would be good to consider better structures.

perrotta:
@riga please also consider the comments in https://github.com/cms-sw/cmssw/pull/27634#pullrequestreview-278536067

perrotta:
One of the outcomes of the discussion at the [reco meeting on Aug 23](https://indico.cern.ch/event/841640) was that with the fix in the number of tracksters implemented in #27819 some performance measurement at PU200 could have been attempted. Please @riga let us know if and when either you or someone else in the group plans to perform such measurement: results could be posted either her or in #27819, and cross-referenced

perrotta:
@riga please notice that there are also conflicts to be fixed now, in 
- DataFormats/HGCalReco/interface/Trackster.h
- DataFormats/HGCalReco/src/classes_def.xml
- RecoHGCal/TICL/plugins/BuildFile.xml

riga:
@perrotta @kpedro88 

Here is the plan for the next steps to finalize this PR:

- [ ] Possible conflicts: A simple merge should do the trick.

- [ ] High-PU performance: I created a test sample with 200 pileup and will repeat the simple memory / cpu checks with that sample, including the changes proposed in #27819.

- [ ] Nested vectors: The reason for using nested vectors in the first place was the need for sorting input data before feeding it to the network. But thinking about it, there is no reason why this should not be done as the first step in the computational graph itself (and probably benefit from the better / parallel implementation in TF ootb). I will update the model another time to include sorting and directly fill the input tensor instead of the intermediate vectors.

- [ ] PF enum type: @kpedro88 do you suggest to store the probabilities in a map `< PFCandidate::ParticleType, float>` or to have (e.g.) a getter such as `id_probability(PFCandidate::ParticleType type)`?

rovere:
> Regarding enums, either option is fine. My main concern is avoiding hardcoded magic numbers that could become inconsistent between different parts of the code, leading to bugs.

At this stage, I do not particularly like bounding these probabilities, that are still internal to TICL, already to Particle Flow. There'll be a module that will link TICL to Particle Flow and, at that level, we can handle all the connections in the proper way.

perrotta:
@riga thank you for presenting your plan, which I am perfectly fine with: please go ahead.
While I believe that eventually the particle types used here to define probabilities will have to get synchronized with the PF ones, I don't have any particular preference about how this is dealt with right now, and I'd rather adhere to the conclusions that you will reach together with @kpedro88

kpedro88:
@rovere in that case, a separate enum class could be defined and used.

riga:
Sorry for the commit duplication, I actually only did a merge of my own branch into the latest IB. rebase-topic lead to conflicts in parts of the code I didn't touch.

All of the review comments are addressed.

I introduced a `ParticleType` enum to the Trackster struct that can be used to query certain id probabilities. `neutral_pion` was added to the list of particle types predicted by the model (which was also updated).

Also, I refactored the code quite a bit to do the sorting of clusters per layer without using nested vectors. It turned out that some kind of sorting has to be done anyway to select the leading clusters per layer of one trackster.

perrotta:
@riga ```git cms-merge-topic``` should be avoided when updating pull requests.
This anyhow need to be rebased now: please refer to http://cms-sw.github.io/tutorial-resolve-conflicts.html

riga:
Something in the merge/rebase workflow seemed odd. The [comparison](https://github.com/cms-sw/cmssw/compare/master...riga:ticl_tracksterEnergyID_11_0_0_pre3) looked actually fine, but the view here in the PR didn't.

I closed and will reopen an other PR.

kpedro88:
> I actually only did a merge of my own branch into the latest IB.

Never do this. rebase-topic will not encounter spurious conflicts if you maintain the PR correctly from the start. Any merges with master can introduce problems. The workflow in http://cms-sw.github.io/tutorial-resolve-conflicts.html should be followed exclusively.

PR #27639, opened 2019-07-30
Fireworks: Change logs description

vargasa:
#### PR description:

Let the `std::runtime_exception` shows its own description as the actual printout could be misleading. 
Same exception with different description are thrown in `try` block. `catch` block needs to provide exception description instead


fabiocos:
please test

alja:
+1

kpedro88:
+1

PR #27649, opened 2019-07-30
Bug fix for setRunNumberForEachLumi

wddgit:
#### PR description:

The PoolSource configuration option setRunNumberForEachLumi had a bug that would cause an assert failure if there was an output module running (and possibly other problems because the wrong run number was being used at endRun). Probably no one was using this option before now because no one noticed the failure before and there is nothing in the repository referencing it other than a Framework unit test.

This bug should not affect anything if that option is not configured. The code fragment with the bug would not be executed.

#### PR validation:

I added an output module to the unit test so this failure will be detected in the IB if it recurs.


wddgit:
please test

Dr15Jones:
+1

kpedro88:
+1

PR #27666, opened 2019-08-01
HGCal - Minimum Cluster size for Pattern Recognition

felicepantaleo:
#### PR description:

This PR is introducing a new filter on the minimum layer clusters size that a hgcalLayerCluster requires in order to be passed to the next step of the Pattern Recognition in TICL. This is useful to make PR more robust against single cell inefficiency and to remove duplicate tracksters.
#### PR validation:

This PR was tested with 20490.0_SingleGammaPt25Eta1p7_2p7 checking that the number of hgcalLayerClusters fed to the pattern recognition is halved, and that the number of duplicate tracksters is decreased. 
Another PR will be made in order to make the Pattern Recognition by CA more robust as well.

@rovere @cseez @amartelli fyi

### DIFF
--- a/RecoHGCal/TICL/python/ticl_iterations.py
+++ b/RecoHGCal/TICL/python/ticl_iterations.py
@@ -43,6 +43,8 @@ def TICL_iterations_withReco(process):
   )
 
   process.FilteredLayerClusters = filteredLayerClustersProducer.clone(
+      clusterFilter = "ClusterFilterByAlgoAndSize",
+      min_cluster_size = 2,
       algo_number = 8,
       iteration_label = "algo8",
       LayerClustersInputMask = "TrackstersMIP"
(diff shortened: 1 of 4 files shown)


slava77:
please test workflow 20493.52

kpedro88:
+upgrade

RecoHGCal/TICL/python/ticl_iterations.py
```diff
@@ -43,6 +43,8 @@ def TICL_iterations_withReco(process):
   )
 
   process.FilteredLayerClusters = filteredLayerClustersProducer.clone(
+      clusterFilter = "ClusterFilterByAlgoAndSize",
```
slava77: this is already a part of filteredLayerClustersProducer PSet. Is this redundant change really necessary here?

slava77:
+1

for https://github.com/cms-sw/cmssw/pull/27666	b3cbe30
- code changes are in line with the PR description (although it could be more clear that this is replacing an old filter, not just "introducing a new filter")
- jenkins tests pass and comparisons with the baseline show no differences because the TICL workflow is not used in comparisons in the tests

https://github.com/cms-sw/cmssw/pull/27666#pullrequestreview-272133903 is minor enough and applies only to an experimental config/customization.

fabiocos:
+1

PR #27445, opened 2019-07-04
Gem digitization update

yeckang:
#### PR description:

This PR modifies GEMDigitization simulation.
* The DigiProducers of GEM and ME0 has moved to plugins.
* GEMTrivialModel has removed, because it isn't be used anymore.
* GEMSimpleModel has spliced as a signal model and noise models.
* The fillDescription has added for GEMDigiProducer.

#### PR validation:

* Code style has fixed with `scram build code-format` command.
* runTheMatrix.py is runnable. 
* The result has not changed for GEM.
* The simulation result has validated in [GEM DPG](https://indico.cern.ch/event/823672/contributions/3444716/attachments/1852072/3040845/GEMDigiUpdate.pdf).

@jshlee @watson-ij

PR #27667, opened 2019-08-01
revert SignalPropagationSpeed at GEMDigiProducer

yeckang:
#### PR description:

This PR is minor revert of the PR #27445.
For consistency with the comment line, the value of signalPropagationSpeed at GEMDigiProducer has reverted to the old version. And `cspeed` is converted to mm/ns to cm/ns for it. @kpedro88

### DIFF
--- a/SimMuon/GEMDigitizer/plugins/GEMDigiProducer.cc
+++ b/SimMuon/GEMDigitizer/plugins/GEMDigiProducer.cc
@@ -85,7 +85,7 @@ void GEMDigiProducer::fillDescriptions(edm::ConfigurationDescriptions& descripti
   desc.add<std::string>("inputCollection", "g4SimHitsMuonGEMHits");
   desc.add<std::string>("mixLabel", "mix");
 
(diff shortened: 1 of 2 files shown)


kpedro88:
Does this change also need to be propagated to https://github.com/cms-sw/cmssw/blob/caf88b0ed55bbf3910ff210aba3b4a884083c41d/SimMuon/GEMDigitizer/src/GEMSignalModel.cc#L101 ?

If so, maybe the conversion should be applied where `cspeed` is defined, so the value is always in mm/ns: https://github.com/cms-sw/cmssw/blob/caf88b0ed55bbf3910ff210aba3b4a884083c41d/SimMuon/GEMDigitizer/src/GEMSignalModel.cc#L26

yeckang:
Ah, okay. Then I will change the initialization of `cspeed` as cm/ns.

kpedro88:
please test

kpedro88:
+upgrade

civanch:
+1

kpedro88:
+1

PR #27644, opened 2019-07-30
enabling UL regression for 2018 (and likely Run3)

Sam-Harper:
#### PR description:
This PR enables the 106X UL regressions for 2018, given the imminent update of these conditions in the GT. 

The 2017UL is already in the global tags so matrix tests ran locally without errors.  Of course they are not fully correct yet as they are not the 2018UL corrections.

Also from looking around at the modifiers it appears Run3 activates the egamma 2018 modifier so Run3 also gets this. Which is good, they should be using this.

PR #27678, opened 2019-08-02
fix short matrix for 2021

kpedro88:
#### PR description:

One-line fix for #27433 (based on comment in https://github.com/cms-sw/cmssw/pull/27644#issuecomment-517689881)

### DIFF
--- a/Configuration/PyReleaseValidation/scripts/runTheMatrix.py
+++ b/Configuration/PyReleaseValidation/scripts/runTheMatrix.py
@@ -74,7 +74,7 @@ if __name__ == '__main__':
                      10024.0, #2017 ttbar
                      10224.0, #2017 ttbar PU
                      10824.0, #2018 ttbar
-                     11624.0, #2021 ttbar
+                     11634.0, #2021 ttbar
                      20034.0, #2026D35 ttbar (MTD TDR baseline)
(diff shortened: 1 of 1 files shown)


kpedro88:
please test

kpedro88:
+upgrade

zhenhu:
+1

Configuration/PyReleaseValidation/scripts/runTheMatrix.py
```diff
@@ -74,7 +74,7 @@ def runSelected(opt):
                      10024.0, #2017 ttbar
                      10224.0, #2017 ttbar PU
                      10824.0, #2018 ttbar
-                     11624.0, #2021 ttbar
+                     11634.0, #2021 ttbar
```
slava77: can we make the runner throw if the workflow is not available?

fabiocos: @slava77 this is a good suggestion, I would manage it independently on this PR

fabiocos:
+1

Configuration/PyReleaseValidation/scripts/runTheMatrix.py
slava77: given some failures in 2023,2024 in some other tests, should we add at least one to the limited set?

PR #27682, opened 2019-08-02
add MTD related hits plots

mtosi:
#### PR description:
add 3 new histograms to the MTV for having the MTD hits profile vs eta

@makortel do you think we should add a flag for the phase2 (so that we add those plots only for the relevant workflows) ?

### DIFF
--- a/Validation/RecoTrack/python/MTVHistoProducerAlgoForTrackerBlock_cfi.py
+++ b/Validation/RecoTrack/python/MTVHistoProducerAlgoForTrackerBlock_cfi.py
@@ -148,21 +148,20 @@ def _modifyForPhase1(pset):
     pset.maxEta = 3
     pset.nintEta = 60
 from Configuration.Eras.Modifier_phase1Pixel_cff import phase1Pixel
+#phase1Pixel.toModify(MTVHistoProducerAlgoForTrackerBlock, dict(minEta = -3, maxEta = 3, nintEta = 60) )
 phase1Pixel.toModify(MTVHistoProducerAlgoForTrackerBlock, _modifyForPhase1)
 
 def _modifyForPhase2(pset):
     pset.minEta = -4.5
     pset.maxEta = 4.5
     pset.nintEta = 90
-
 from Configuration.Eras.Modifier_phase2_tracker_cff import phase2_tracker
+#phase2_tracker.toModify(MTVHistoProducerAlgoForTrackerBlock, dict(minEta = -4.5, maxEta = 4.5, nintEta = 90) )
 phase2_tracker.toModify(MTVHistoProducerAlgoForTrackerBlock, _modifyForPhase2)
 
 def _modifyForPhase2wMTD(pset):
-    pset.minEta = -4.5
-    pset.maxEta = 4.5
-    pset.nintEta = 90
-    doMTDPlots = cms.untracked.bool(True)
-
+    pset.doMTDPlots = True
 from Configuration.Eras.Modifier_phase2_timing_layer_cff import phase2_timing_layer
-phase2_timing_layer.toModify(MTVHistoProducerAlgoForTrackerBlock, _modifyForPhase2wMTD )
+#phase2_timing_layer.toModify(MTVHistoProducerAlgoForTrackerBlock, dict(doMTDPlots = True) )
+phase2_timing_layer.toModify(MTVHistoProducerAlgoForTrackerBlock, _modifyForPhase2wMTD)
+
--- a/Validation/RecoTrack/python/MTVHistoProducerAlgoForTrackerBlock_cfi.py
+++ b/Validation/RecoTrack/python/MTVHistoProducerAlgoForTrackerBlock_cfi.py
@@ -139,6 +139,8 @@ MTVHistoProducerAlgoForTrackerBlock = cms.PSet(
     nintDzpvsigCumulative = cms.int32(200),
 
     seedingLayerSets = cms.vstring(),
+
+    doMTDPlots = cms.untracked.bool(False), # meant to be switch on in Phase2 workflows
 )
 
 def _modifyForPhase1(pset):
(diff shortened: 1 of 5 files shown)


mtosi:
@lgray @pmeridian

makortel:
> do you think we should add a flag for the phase2 (so that we add those plots only for the relevant workflows) ?

Hmm, I'm not sure. It's a bit of a hassle, but would save some memory for non-phase2 workflows. If more MTD plots are foreseen, the flag would start to be more useful.

mtosi:
well, in view of a more 4D tracking (and vertexing), I would expect more plots which are relevant only with the MTD detectors

but, this will not happen in the short term, I’m afraid

if you agree, I’ll add a flag
or are there more brilliant ways ?

makortel:
> if you agree, I’ll add a flag
> or are there more brilliant ways ?

Sounds good, a configuration flag customized with an era is the best way.

mtosi:
thanks for the confirmation 

  mia

fabiocos:
@mtosi what are the plans for this PR? It looks dormant...

mtosi:
@fabiocos , thanks for the message
apologies for the delay, but I'm just back from Holidays ...

mtosi:
@cmsbuild, please test

Validation/RecoTrack/python/MTVHistoProducerAlgoForTrackerBlock_cfi.py
```diff
@@ -147,5 +149,20 @@ def _modifyForPhase1(pset):
...
+
 from Configuration.Eras.Modifier_phase2_tracker_cff import phase2_tracker
-phase2_tracker.toModify(MTVHistoProducerAlgoForTrackerBlock, minEta=-4.5, maxEta=4.5, nintEta = 90)
+phase2_tracker.toModify(MTVHistoProducerAlgoForTrackerBlock, _modifyForPhase2)
+
+def _modifyForPhase2wMTD(pset):
+    pset.minEta = -4.5
+    pset.maxEta = 4.5
+    pset.nintEta = 90
+    doMTDPlots = cms.untracked.bool(True)
```
makortel: Isn't the binning of eta more a property of the tracker than the MTD? If that is the case, it is redundant to repeat the eta binning customization here (AFAIK all phase2 eras with `phase2_timing_layer` include `phase2_tracker`).

Also, it is preferred to avoid specifying the type (`cms.untracked.bool`) if not necessary (here it is not).

Validation/RecoTrack/python/MTVHistoProducerAlgoForTrackerBlock_cfi.py
```diff
@@ -147,5 +149,20 @@ def _modifyForPhase1(pset):
...
 phase1Pixel.toModify(MTVHistoProducerAlgoForTrackerBlock, _modifyForPhase1)
+
+def _modifyForPhase2(pset):
+    pset.minEta = -4.5
+    pset.maxEta = 4.5
+    pset.nintEta = 90
+
 from Configuration.Eras.Modifier_phase2_tracker_cff import phase2_tracker
-phase2_tracker.toModify(MTVHistoProducerAlgoForTrackerBlock, minEta=-4.5, maxEta=4.5, nintEta = 90)
+phase2_tracker.toModify(MTVHistoProducerAlgoForTrackerBlock, _modifyForPhase2)
```
makortel: In general dictionary syntax is preferred over functions.

Validation/RecoTrack/python/TrackValidation_cff.py
```diff
@@ -861,15 +861,20 @@ def _uniqueFirstLayers(layerList):
                               timeResosTag = cms.InputTag('trackTimeValueMapProducer:generalTracksConfigurableFlatResolutionModelResolution'), 
                               nSigmaDtVertex = cms.double(3) )
 phase2_timing_layer.toModify( trackValidatorStandalone,
-                              label_vertex = cms.untracked.InputTag('offlinePrimaryVertices4D') )
+                              label_vertex = cms.untracked.InputTag('offlinePrimaryVertices4D'),
+                              doMTDPlots = cms.untracked.bool(True) )
```
makortel: Are these now unnecessary? (if yes, without `cms.untracked.bool` the configuration system would have complained)

jfernan2:
@mtosi it seems no histograms are being added with this PR in any workflow

jfernan2:
please test

jfernan2:
@mtosi are you aware that you add 3 plots X 63 folders = 189 histograms? And that in some cases they are empty? It was not clear to me from your PR description
https://cmssdt.cern.ch/SDT/jenkins-artifacts/baseLineComparisons/CMSSW_11_0_X_2019-09-16-2300+b2133b/33509/dqm-histo-comparison-summary.html

mtosi:
ciao
thanks for the comment, indeed it is expected and meant to validate / monitor the hits from the MTD detector. they are not used in the current pattern recognition, but the MTD and TRK POG are planning to make use of them as well

jfernan2:
+1

Validation/RecoTrack/src/MTVHistoProducerAlgoForTracker.cc
```diff
@@ -2241,6 +2252,14 @@ void MTVHistoProducerAlgoForTracker::fill_simAssociated_recoTrack_histos(const H
   histograms.nSTRIPhits_vs_eta[count].fill(eta, tibHits + tidHits + tobHits + tecHits);
   histograms.nLayersWithMeas_vs_eta[count].fill(eta, track.hitPattern().trackerLayersWithMeasurement());
   histograms.nPXLlayersWithMeas_vs_eta[count].fill(eta, track.hitPattern().pixelLayersWithMeasurement());
+  if (doMTDPlots_) {
+    //  const auto mtdHits = track.hitPattern().numberOfValidTimingHits();
+    const auto btlHits = track.hitPattern().numberOfValidTimingBTLHits();
+    const auto etlHits = track.hitPattern().numberOfValidTimingETLHits();
+    histograms.nMTDhits_vs_eta[count].fill(eta, btlHits + etlHits);
```
fabiocos: @mtosi sorry, the 3 histograms are identical in definition, just filled with either btl, etl or both. At the least the last histogram is redundant (or the first two). Could you please clarify the rationale behind this choice, that ahs an impact in memory?

mtosi:
as always, we make plots for subsystem and for the whole
(as we do for the 4 strip subsystem, and and the whole strip detector)

fabiocos:
@mtosi @jfernan2 in case we start to be compressed in memory a choice to be reconsidered...

mtosi:
the only thing i can add is that
if one wants to be independent, and check the hitPattern filling,
then we should use the hitPattern directly instead of passing through the sum

point taken, thanks
  mia

fioriNTU:
@mtosi @fabiocos one possible workaround would be to fill etl and btl plots in step1 and then sum up them in Harvesting. This would save 1/3 of the memory added. However the increase in memory of about 600 KB seems completely acceptable, so we can safely integrate it for the time being.

fabiocos:
+1

PR #27695, opened 2019-08-05
throw a meaninful error if LFN isn't found in FileCatalog

dan131riley:
#### PR description:

When the StreamerFileReader tries to process an LFN that doesn't have a valid PFN translation, it currently tries to open the file "", leading to difficult to diagnose errors at the Tier0, see this thread:

https://hypernews.cern.ch/HyperNews/CMS/get/edmFramework/3876/1/1/1/2/1/1/2/1.html

This PR throws a LogicalFileNameNotFound error if the PFN is empty.  Some code changes needed to be made to the error message could include the LFN that resulted in an empty PFN.

#### PR validation:

Code was tested on the Tier0 repack job configuration from the HN thread, confirming that a working config still worked, and a config with an invalid LFN resulted in a meaningful error message.

In addition the ReadStreamerFile test was repaired (it wasn't running the tests and always returned success), and a new test case added to verify that an invalid LFN results in the correct exception.

### DIFF
--- a/IOPool/Streamer/test/BuildFile.xml
+++ b/IOPool/Streamer/test/BuildFile.xml
@@ -5,13 +5,18 @@
   </bin>
   <bin   file="ReadStreamerFile.cpp">
     <use   name="IOPool/Streamer"/>
-    <flags   TEST_RUNNER_ARGS=" bin/bash teststreamfile.dat"/>
+    <use   name="FWCore/Catalog"/>
+    <use   name="FWCore/PluginManager"/>
+    <use   name="FWCore/ServiceRegistry"/>
+    <use   name="FWCore/Services"/>
+    <use   name="FWCore/ParameterSetReader"/>
+    <flags   TEST_RUNNER_ARGS="all"/>
   </bin>
   <bin   file="WriteStreamerFile.cpp">
     <use   name="IOPool/Streamer"/>
   </bin>
   <bin   file="RunThis_t.cpp" name="NewStreamerUNCOMPRESSED">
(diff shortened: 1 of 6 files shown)


fabiocos:
please test

IOPool/Streamer/test/BuildFile.xml
```diff
@@ -5,13 +5,18 @@
   </bin>
   <bin   file="ReadStreamerFile.cpp">
     <use   name="IOPool/Streamer"/>
-    <flags   TEST_RUNNER_ARGS=" bin/bash teststreamfile.dat"/>
+    <use   name="FWCore/Catalog"/>
+    <use   name="FWCore/PluginManager"/>
+    <use   name="FWCore/ServiceRegistry"/>
+    <use   name="FWCore/Services"/>
+    <use   name="FWCore/ParameterSetReader"/>
+    <flags   TEST_RUNNER_ARGS="all"/>
```
Dr15Jones: is `all` a special value known by scram?

Dr15Jones: I see that it is an option known by ReadStreamerFile

dan131riley: Right, it's just an arg to the test program, scram doesn't know anything about the value.  Alternatively we could make all the default and get rid of the arg, which might be safer--previously the arg was set to a nonsense string and as a result the tests weren't even being run.

Dr15Jones:
+1

fabiocos:
+1

PR #27705, opened 2019-08-06
TemplateReco: Fix out of bounds memory write for very large clusters

OzAmram:
This is a fix for the issue reported a few days ago: https://github.com/cms-sw/cmssw/issues/27685.

This out of bounds memory write could happen if there was a very large cluster that essentially filled up the whole length of the cluster matrix and this clusters center was 'misaligned' with the templates expectation. The code previously checked for these oversized clusters and return an error if it found them but this was after a memory write happened. This fix, from Morris, does the check before any writing happens so that no invalid writes can happen. 

I would expect that there is no difference in output after this change. The same clusters will still get the same error and the invalid write was only to the local stack of the function, which should have been returning very soon anyway so it likely caused no real damage. 

@Dr15Jones @mmusich @tsusa @pmaksim1 @cmantill @tvami

### DIFF
--- a/RecoLocalTracker/SiPixelRecHits/src/SiPixelTemplateReco.cc
+++ b/RecoLocalTracker/SiPixelRecHits/src/SiPixelTemplateReco.cc
@@ -1,5 +1,5 @@
 //
-//  SiPixelTemplateReco.cc (Version 10.00)
+//  SiPixelTemplateReco.cc (Version 10.01)
 //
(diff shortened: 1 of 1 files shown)


Dr15Jones:
please test

perrotta:
+1
- Bug fix implemented as in the description
- As such, it should fix the issue reported in #27685

fabiocos:
+1

PR #27717, opened 2019-08-07
Added missing return statement in switch for SiStripPayloadInspectorHelper

Dr15Jones:
#### PR description:

gcc 9 gave a warning that the case statement was falling through to the next statement. Looking at the git history showed that before the routine was returning a std::pair it returned a string and for the case in question it returned the value "undefined". This change now matches the earlier behavior.

#### PR validation:

Tested compilation under a gcc 9 IB.

This change would only cause a results change if incorrect values are being passed to the routine which was modified.


Dr15Jones:
please test

fabiocos:
@ggovi this looks a trivial bug fix to me, I will merge it in next IB

fabiocos:
+1

fabiocos:
merge

PR #27719, opened 2019-08-07
Fix copy constructor for TRecHit1DMomConstraint

Dr15Jones:
#### PR description:

The gcc 9 compiler complained that the base class copy constructor was not being called. As the code was doing the same work as the default compiler written copy constructor, switched to using `= default`.

#### PR validation:

Tested compilation using a gcc 9 IB.
The change is technical and should not affect results unless the results were dependent on an incorrect copy constructor.


Dr15Jones:
please test

perrotta:
+1
- Technical
- Jenkins tests pass

fabiocos:
+1

PR #27721, opened 2019-08-07
Explicitly use default c'tor in muon DetId copy c'tors

Dr15Jones:
#### PR description:

The gcc 9 compiler was warning that inheriting class copy constructors were not calling the base class copy constructors. The code was doing that on purpose. To make that clear, the code now explicitly calls the default constructor for the base class.

#### PR validation:

The code compiles using a gcc 9 IB.
This is a technical change (explicitly doing what the compiler was implicitly doing) and should not change results.


Dr15Jones:
please test

civanch:
+1
I suspect, that this copy constructor not needed but if it is implemented , then should be correct.

fabiocos:
+1

PR #16366, opened 2016-10-26
Ensure GEANT simulation of exotic long-lived particles that decay outside beampipe

wulsin:
To accomplish this, for any particle with status > 3, set its status to 2.  All such particles are then checked to see if their decay vertex is outside the beampipe; if so it will be passed to GEANT for simulation.  

This fix addresses the problem of a long-lived exotic particle with status code > 3 (e.g., an R-hadron, or chargino), which decays outside the beampipe.  Before this fix, its detector interactions would not be simulated by GEANT, because this module was looking specifically for status values of 1, 2, or 3.  

This fix has been tested for the case of di-chargino pair production, with the chargino ctau set to 1000 cm (decaying to neutralino + pion).  Before the fix, there are no tracks associated to the chargino because no energy deposits have been simulated by GEANT.  After the fix, there are reconstructed tracks for each chargino.

PR #20083, opened 2017-08-08
for neutral exotic WIMPs - send their daughters to Geant4 instead of …

zhangzc11:
…the exotic itself

PR #27732, opened 2019-08-08
Only track charged exotic particles outside CMS beam pipe for abs(pdg id) from 1000000 to 4000000

weishi10141993:
This is to address the issue described in: https://hypernews.cern.ch/HyperNews/CMS/get/simDevelopment/1904.html

The original two functions 'isExotic' and 'isExoticNonDetectable' are now merged into one function 'isChargedExotic'.

Note: Ideally the absolute pdg id upper limit 4000000 could be extended to 9999999 (the largest number in 7-digit, base 10, following PDG MC particle numbering scheme convention) since some other groups might want to track charged exotic particles with pdg id beyond 4000000 in the future. Here I didn't extend it since our neutral signal particle 3000022 is covered by the current upper limit 4000000. But I'll let @civanch comment more.

This PR should be ported to all releases since CMSSW_8_1_X because that's when PR #16366 starts to add the tracking of long lived exotic particles (from my understanding what they really wanted in PR #16366 is to track 'charged' exotic particles, but they didn't specify charge!=0 back then and that caused an issue for our dark photon 3000022). We also need this PR #27732 for 2017 and 2018 central MC sample production.

### DIFF
--- a/SimG4Core/Generators/interface/Generator.h
+++ b/SimG4Core/Generators/interface/Generator.h
@@ -33,8 +33,7 @@ public:
 
 private:
   bool particlePassesPrimaryCuts(const G4ThreeVector &p) const;
-  bool isExotic(int pdgcode) const;
-  bool isExoticNonDetectable(int pdgcode) const;
+  bool isChargedExotic(int pdgcode) const;
   bool IsInTheFilterList(int pdgcode) const;
   void particleAssignDaughters(G4PrimaryParticle *p, HepMC::GenParticle *hp, double length);
   void setGenId(G4PrimaryParticle *p, int id) const { p->SetUserInformation(new GenParticleInfo(id)); }
--- a/SimG4Core/Generators/src/Generator.cc
+++ b/SimG4Core/Generators/src/Generator.cc
@@ -145,7 +145,7 @@ void Generator::HepMC2G4(const HepMC::GenEvent *evt_orig, G4Event *g4evt) {
       // propagated by GEANT
       int status = (*pitr)->status();
       int pdg = (*pitr)->pdg_id();
-      if (status > 3 && isExotic(pdg) && (!(isExoticNonDetectable(pdg)))) {
+      if (status > 3 && isChargedExotic(pdg)) {
         // In Pythia 8, there are many status codes besides 1, 2, 3.
         // By setting the status to 2 for exotic particles, they will be
         // checked: if its decay vertex is outside the beampipe, it will be
@@ -216,7 +216,7 @@ void Generator::HepMC2G4(const HepMC::GenEvent *evt_orig, G4Event *g4evt) {
         if ((!pdgFilterSel && isInTheList) || (pdgFilterSel && !isInTheList)) {
           edm::LogVerbatim("SimG4CoreGenerator")
               << " Skiped GenParticle barcode= " << (*pitr)->barcode() << " PDGid= " << pdg << " status= " << status
-              << " isExotic: " << isExotic(pdg) << " isExoticNotDet: " << isExoticNonDetectable(pdg)
+              << " isChargedExotic: " << isChargedExotic(pdg)
               << " isInTheList: " << isInTheList << " hasDecayVertex: " << hasDecayVertex;
           continue;
         }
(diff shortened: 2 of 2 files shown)


civanch:
@weishi10141993 , I agree, that we may merge two methods but have a problem with this modification: why neutral exotics is excluded ? For today we have a bug, let us fix it but left neutral exotics a possibility to be simulated by G4. When we complete this PR we will back port the fix to CMSW_10_6_X (legacy for run-2) and may be for some other releases. For code format and code check simply run:

scram b code-format

which will format the code automatically in your area, after push again.

SimG4Core/Generators/interface/Generator.h
```diff
@@ -33,8 +33,7 @@ class Generator {
 
 private:
   bool particlePassesPrimaryCuts(const G4ThreeVector &p) const;
-  bool isExotic(int pdgcode) const;
-  bool isExoticNonDetectable(int pdgcode) const;
+  bool isChargedExotic(int pdgcode) const;
```
civanch: @weishi10141993 , rename this to isExotic(int pdgcode)

SimG4Core/Generators/src/Generator.cc
```diff
@@ -495,21 +495,14 @@ bool Generator::particlePassesPrimaryCuts(const G4ThreeVector &p) const {
...
-             ? true
-             : false;
-}
-
-bool Generator::isExoticNonDetectable(int pdgcode) const {
+bool Generator::isChargedExotic(int pdgcode) const {
   int pdgid = std::abs(pdgcode);
   HepPDT::ParticleID pid(pdgcode);
   int charge = pid.threeCharge();
-  return ((charge == 0) && (pdgid >= 1000000 && pdgid < 1000040))  // SUSY
```
civanch: @weishi10141993 , here we need to have some logic allowing neutral be selected.

SimG4Core/Generators/interface/Generator.h
weishi10141993: I didn't quite understand why/how the tracking of neutral extotic particle works at Geant4. What kind of interactions will be actually simulated since they are not charged and have no QCD interaction?

SimG4Core/Generators/src/Generator.cc
weishi10141993: Some other group (PR #20083) specifically do not want to select neutral SUSY particles. They use !isExoticNonDetectable to exclude SUSY particles from what PR #16366 specified.

Do you mean we should allow to select neutral particles from 1000040 to 4000000; and for 1000000 to 1000040, only select charged?

jalimena:
bringing this to the attention of @lowette  and @mcitron

civanch:
@weishi10141993 , @jalimena , @lowette , @mcitron ,
Specific of this class is that HepMC has not only conventional particles known to Geant4 but exotics, which may have different properties. We update this code from time to time due to new request coming from exotics analysis. In order to make logic of the code more transparent we introduced two methods:

bool isExotic(pdg);

bool (isExoticNonDetectable(pdg));

Initially I was trend to agree with this PR but after thinking come to following conclusion: we introduced the logic to handle HepMC particles basing on pdg code, status code, vertex position. We have satisfied all previous requests for exotic studies. When a new  use case appears for me the best strategy is to preserve previous functionality but allow new use case be properly simulated. So, I disagree this this PR. What we should do: make an upgrade which allows proper addition of displaced muons but not changing anything for other exotics, which were implemented in past.

SimG4Core/Generators/interface/Generator.h
civanch: @weishi10141993 , it may decay into charged secondaries.

jalimena:
Hi @civanch , indeed, it would be good to preserve the functionality used by other analyses, but change what needs to be changed for this use case. Hopefully we can accomodate everyone.

weishi10141993:
Hi, @civanch, it’s definitely ok for us to keep the same structure and make minor changes for our case. I can open another PR. 
But even before rushing for a quick fix, I wonder, as I also asked in HyperNews, why Geant4 simulate prompt muon simtracks while our dark photon is Long-lived and decays outside the beam pipe? From the current code, Geant4 should be able to track these Long-lived neutral exotics decaying to charged particles and there should be no problem for our case. 
Are there any possible reasons you could think of?

PR #27738, opened 2019-08-09
Addition of pixelClusterTagInfos to AOD and miniAOD

zucchett:
#### PR description:

This pull request defines a new TagInfo object (PixelClusterTagInfo), produced from the RECO format starting from the siPixelCluster collection, propagated to AOD, and embedded in miniAOD. The TagInfo contains a very compressed information of the pixel clusters geometrically matched to an AK4 jet, consisting of a char for each 4 pixel layers times 6 jet-cluster association cones, plus four unsigned integers for the collected charge. This information is added only to jets with pT > 100 GeV, and is critical for novel studies of b-tagging at high pT without relying on tracks, whose reconstruction efficiency decreases in very energetic jets. The pixel cluster information can be provided to a standalone tagger or (in the future) integrated in combined b-tagging algorithms. Preliminary studies performed on very limited samples estimate that a pixel-based tagger provides a better performance than the DeepCSV for jet pT > 1.2 TeV.

Latest presentations:
https://indico.cern.ch/event/828712/contributions/3468910/attachments/1869375/3075401/Sommerhalder_Zucchetta.pdf
https://indico.cern.ch/event/830302/contributions/3483708/attachments/1871179/3079451/190628_Zucchetta.pdf
https://indico.cern.ch/event/833434/contributions/3495708/attachments/1877898/3093199/190710_Zucchetta.pdf

#### PR validation:

The basic checks have been successfully run (code-checks, runTheMatrix). The only exception is represented by two jobs in runTheMatrix that run on already-produced AOD (136.7611* and 136.8311*) that fail because the needed collections are not found. This is understandable because neither the basic siPixelCluster nor the pixelClusterTagInfo collections are present in AODs by default; the latter collection is added to AODs with this PR.
EDIT: Era modifiers have been introduced to exclude pixelClusterTagInfos from 80X and 94X miniAOD productions, avoiding the crash.

@msommerh as developer, and BTV conveners @mverzett , @kskovpen, @XavierAtCERN, @smoortga are also involved and interested in these developments.


slava77:
> The only exception is represented by two jobs in runTheMatrix that run on already-produced AOD (136.7611* and 136.8311*) that fail because the needed collections are not found.

this should be fixed.
These re-mini workflows are using special era/process modifiers. Please use them to appropriately disable or perhaps instead produce "dummy" collections.

slava77:
@cmsbuild please test

slava77:
>  Commits 63

@zucchett 
please also check if you can squash some intermediate commits to compactify the history a bit.

slava77:
@cmsbuild please test

DataFormats/BTauReco/interface/PixelClusterTagInfo.h
```diff
@@ -0,0 +1,87 @@
...
+    float z = 0;
+    int charge = 0;
+    int layer = 0;
+  };
+
+  struct PixelClusterData {
+    char L1_R004 = 0;
+    char L2_R004 = 0;
+    char L3_R004 = 0;
+    char L4_R004 = 0;
```
slava77: if this is to be used for integers, better use ```int8_t```

DataFormats/BTauReco/interface/PixelClusterTagInfo.h
```diff
@@ -0,0 +1,87 @@
...
+
+    char L1_R006 = 0;
+    char L2_R006 = 0;
+    char L3_R006 = 0;
+    char L4_R006 = 0;
+
+    char L1_R008 = 0;
+    char L2_R008 = 0;
+    char L3_R008 = 0;
+    char L4_R008 = 0;
```
slava77: at least the layer items better be in an array

zucchett: Ok, i implemented both changes in the latest commit.

perrotta:
please test

RecoBTag/PixelCluster/plugins/PixelClusterTagInfoProducer.cc
```diff
@@ -0,0 +1,162 @@
...
+    edm::RefToBase<reco::Jet> jetRef = collectionJets->refAt(j);  // Get jet RefToBase
+
+    reco::PixelClusterData data = {};  // Initialize new Data to 0
+    reco::PixelClusterTagInfo tagInfo = {};
+
+    for (auto cluIt = clusters.begin(); cluIt != clusters.end(); ++cluIt) {
+      TVector3 c3(cluIt->x - firstPV->x(), cluIt->y - firstPV->y(), cluIt->z - firstPV->z());
+      TVector3 j3(jetRef->px(), jetRef->py(), jetRef->pz());
+      float dR = j3.DeltaR(c3);
+      float sC = 12. * 2. / (jetRef->pt());
```
perrotta: Magic numbers?

RecoBTag/PixelCluster/python/pixelCluster_EventSetup_cff.py
```diff
@@ -0,0 +1,4 @@
+import FWCore.ParameterSet.Config as cms
```
perrotta: Why this config file is needed?

RecoBTag/PixelCluster/python/pixelCluster_cff.py
```diff
@@ -0,0 +1,8 @@
+import FWCore.ParameterSet.Config as cms
+
+from RecoBTag.PixelCluster.pixelClusterTagInfos_cfi import *
+
+pixelClusterTask = cms.Task(
```
perrotta: Where is this task used?
If not used, this config file is also useless

RecoBTag/Configuration/python/RecoBTag_cff.py
```diff
@@ -11,6 +11,7 @@
 from RecoBTag.TensorFlow.pfDeepDoubleX_cff import *
 from RecoBTag.MXNet.pfDeepBoostedJet_cff import *
 from RecoVertex.AdaptiveVertexFinder.inclusiveVertexing_cff import *
+from RecoBTag.PixelCluster.pixelCluster_cff import *
```
perrotta: If you import here just directly "RecoBTag.PixelCluster.pixelClusterTagInfos_cfi", there is no need of this include of "RecoBTag.PixelCluster.pixelCluster_cff"

DataFormats/BTauReco/interface/PixelClusterTagInfo.h
```diff
@@ -0,0 +1,60 @@
...
+  struct PixelClusterProperties {
+    float x = 0;
+    float y = 0;
+    float z = 0;
+    int charge = 0;
+    int layer = 0;
+  };
+
+  struct PixelClusterData {
+    int8_t R004[4] = {};
```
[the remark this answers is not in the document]
zucchett: I like the option to pass the size of the array in the constructor, but in this case we can either use the new operator or switch to std::vectors. Which one would you prefer?

RecoBTag/Configuration/python/RecoBTag_EventContent_cff.py
```diff
@@ -32,6 +32,8 @@
         'keep *_pfDeepCSVJetTags_*_*',
         # DeepCMVA
         'keep *_pfDeepCMVAJetTags_*_*',
+        # Pixel Cluster
+        'keep *_pixelClusterTagInfos*_*_*',
```
slava77: we do not allow wildcards in the module names in the event content

PhysicsTools/PatAlgos/python/slimming/MicroEventContent_cff.py
```diff
@@ -84,6 +84,8 @@
 	'keep recoTracks_displacedStandAloneMuons__*',
         # L1 prefiring weights
         'keep *_prefiringweight_*_*',
+        # Pixel Cluster
+        'keep *_pixelClusterTagInfos*_*_*',
```
slava77: here again please replace the module name wildcards

DataFormats/BTauReco/interface/PixelClusterTagInfo.h
```diff
@@ -0,0 +1,60 @@
...
+    int8_t R006[4] = {};
+    int8_t R008[4] = {};
+    int8_t R010[4] = {};
+    int8_t R016[4] = {};
+    int8_t RVAR[4] = {};
+    unsigned int RVWT[4] = {};
```
perrotta: CMS code rules require lowercase initials for class member variables: "r004", ..., "rVWT" (or "rvwt", or whatever you think it is most appropriate)

DataFormats/BTauReco/interface/PixelClusterTagInfo.h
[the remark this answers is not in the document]
perrotta: @zucchett , I honestly don't have preferences, provided memory is dealt with correctly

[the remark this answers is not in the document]
slava77: ```std::array```  provides appropriate support for fixed size arrays

RecoBTag/PixelCluster/plugins/PixelClusterTagInfoProducer.cc
```diff
@@ -0,0 +1,162 @@
...
+
+  // Retrieve tracker topology from geometry
+  edm::ESHandle<TrackerTopology> tTopoH;
+  iSetup.get<TrackerTopologyRcd>().get(tTopoH);
+  const TrackerTopology* tTopo = tTopoH.product();
+
+  std::vector<reco::PixelClusterProperties> clusters;
+
+  // Get vector of detunit ids, and fill a vector of PixelClusterProperties in the loop
+  for (edmNew::DetSetVector<SiPixelCluster>::const_iterator detUnit = collectionClusters->begin();
```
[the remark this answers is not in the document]
zucchett: It's certainly possible to use range loops, but since there are two nested loops we have to either:
1) make a copy of the DetSet<SiPixelCluster> object, slowing the code, or
2) have a not very readable code consisting of a chain of:
collectionClusters->data(d)->data(u) or even more complex syntax like
(*(*collectionClusters)[d])[u]

This is why we opted (and we still have the preference) for iterator loops in first place.

RecoBTag/PixelCluster/plugins/PixelClusterTagInfoProducer.cc
```diff
@@ -0,0 +1,162 @@
...
+      PixelBarrelName pbn(detid, tTopo, m_isPhase1);
+      layer = pbn.layerName();
+    } else if (m_addFPIX && subid == 2) {  // pixel forward
+      PixelEndcapName pen(detid, tTopo, m_isPhase1);
+      layer = pen.diskName();
+    }
+    if (layer == 0 || layer > m_nLayers)
+      continue;
+
+    for (edmNew::DetSet<SiPixelCluster>::const_iterator clustIt = detUnit->begin(); clustIt != detUnit->end();
```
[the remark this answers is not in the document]
zucchett: See also the comment above.

RecoBTag/PixelCluster/plugins/PixelClusterTagInfoProducer.cc
```diff
@@ -0,0 +1,162 @@
...
+  for (unsigned int j = 0, nj = collectionJets->size(); j < nj; j++) {
+    if (collectionJets->at(j).pt() < m_minJetPt)
+      continue;
+
+    edm::RefToBase<reco::Jet> jetRef = collectionJets->refAt(j);  // Get jet RefToBase
+
+    reco::PixelClusterData data = {};  // Initialize new Data to 0
+    reco::PixelClusterTagInfo tagInfo = {};
+
+    for (auto cluIt = clusters.begin(); cluIt != clusters.end(); ++cluIt) {
```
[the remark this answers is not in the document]
zucchett: This one can certainly become a range based loop.

RecoBTag/PixelCluster/plugins/PixelClusterTagInfoProducer.cc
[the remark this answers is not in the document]
perrotta:
```C++
  for (auto & detUnit :  collectionClusters) {
```
wouldn't work?

RecoBTag/PixelCluster/plugins/PixelClusterTagInfoProducer.cc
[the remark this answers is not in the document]
perrotta:
```C++
    for (auto& clustIt : detUnit) {
```
wouldn't work as well?

RecoBTag/PixelCluster/plugins/PixelClusterTagInfoProducer.cc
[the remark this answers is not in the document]
zucchett: Yes, but to make it work i have to first convert the edm::Handle<edmNew::DetSetVector<SiPixelCluster> >
to a
const edmNew::DetSetVector<SiPixelCluster>&

zucchett:
The latest commit includes all the comments received so far.

DataFormats/BTauReco/interface/PixelClusterTagInfo.h
```diff
@@ -0,0 +1,69 @@
...
+    // method to set the PixelClusterData
+    void setData(const PixelClusterData& data_) { pixelClusters = data_; }
+
+    // method to get the PixelClusterData struct
+    const PixelClusterData& data() const { return pixelClusters; }
+
+  private:
+    PixelClusterData pixelClusters;
+
+    edm::RefToBase<Jet> jetRef;
```
perrotta: In CMSSW standards a trailing underscore indicates class members. Therefore, I'd rater rename
- ```jetRef``` -> ```jetRef_```
- ```pixelClusters``` -> ```pixelClusters_```

and above L48 and L54
```C++
    void setJetRef(const edm::RefToBase<Jet>& ref) { jetRef_ = ref; }
```
```C++
    void setData(const PixelClusterData& data) { pixelClusters_ = data; }
```

RecoBTag/PixelCluster/interface/PixelClusterTagInfoProducer.h
```diff
@@ -0,0 +1,89 @@
...
+  edm::EDGetTokenT<edm::View<reco::Jet> > m_jets;
+  edm::EDGetTokenT<reco::VertexCollection> m_vertices;
+  edm::EDGetTokenT<edmNew::DetSetVector<SiPixelCluster> > m_pixelhit;
+  bool m_isPhase1;
+  bool m_addFPIX;
+  int m_minADC;
+  double m_minJetPt;
+  double m_maxJetEta;
+  double m_hadronMass;
+  double hadronMass;
```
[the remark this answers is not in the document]
perrotta: By the way, there is no reason to have it "double", as it is only meant to build a "float"

RecoBTag/PixelCluster/interface/PixelClusterTagInfoProducer.h
```diff
@@ -0,0 +1,89 @@
...
+#include "Geometry/TrackerGeometryBuilder/interface/TrackerGeometry.h"
+#include "Geometry/Records/interface/TrackerDigiGeometryRecord.h"
+#include "Geometry/CommonDetUnit/interface/GeomDetType.h"
+#include "Geometry/CommonTopologies/interface/PixelTopology.h"
+
+// ROOT
+#include "TVector3.h"
+#include "TLorentzVector.h"
+
+class PixelClusterTagInfoProducer : public edm::stream::EDProducer<> {
```
perrotta: Once you make all class member constant, I think this producer can even become a global producer

zucchett: Thanks Andrea, i made the producer global, and merged the .h in the .cc. The other modifications have been applied, too. I just pushed the changes.

RecoBTag/PixelCluster/plugins/PixelClusterTagInfoProducer.cc
```diff
@@ -22,7 +22,89 @@
 //
 //
 
-#include "RecoBTag/PixelCluster/interface/PixelClusterTagInfoProducer.h"
+#ifndef RecoBTag_PixelCluster_PixelClusterTagInfoProducer_h
+#define RecoBTag_PixelCluster_PixelClusterTagInfoProducer_h
```
perrotta: L25 and L26 not needed, since this file is not to be included anywhere

perrotta (commented):
Thank you Alberto!
Just remove these two useless lines, and we can restart the tests

RecoBTag/PixelCluster/plugins/PixelClusterTagInfoProducer.cc
zucchett: Ok, done!

perrotta:
please test

DataFormats/BTauReco/interface/PixelClusterTagInfo.h
```diff
@@ -0,0 +1,69 @@
...
+  struct PixelClusterProperties {
+    float x = 0;
+    float y = 0;
+    float z = 0;
+    int charge = 0;
+    int layer = 0;
+  };
+
+  struct PixelClusterData {
+    std::vector<int8_t> r004;
```
slava77: it looks like this is used as an array of size 4.
Why was std::array not used?

slava77: ah, nm. there is also an explicit support for phase-0, aiming to save a few chars.

slava77:
I was looking for some CPU/memory/storage measurements.
Something on PU\~50.
Did I miss it in this thread, or was it not done yet?

PhysicsTools/PatAlgos/python/slimming/miniAOD_tools.py
```diff
@@ -424,6 +424,32 @@ def miniAOD_customizeCommon(process):
...
+    run2_miniAOD_80XLegacy.toReplaceWith(pfBTaggingTask, pfBTaggingTaskNoPixelCluster)
+    run2_miniAOD_80XLegacy.toModify(process.patJets, addTagInfos = cms.bool(False) )
+    run2_miniAOD_80XLegacy.toModify(process.slimmedJetsNoDeepFlavour, dropTagInfos = cms.string('1') )
+    run2_miniAOD_80XLegacy.toModify(process.updatedPatJetsSlimmedDeepFlavour, addTagInfos = cms.bool(False) )
+    run2_miniAOD_80XLegacy.toModify(process.updatedPatJetsTransientCorrectedSlimmedDeepFlavour, addTagInfos = cms.bool(False) )
+    
+    from Configuration.Eras.Modifier_run2_miniAOD_94XFall17_cff import run2_miniAOD_94XFall17
+    run2_miniAOD_94XFall17.toReplaceWith(pfBTaggingTask, pfBTaggingTaskNoPixelCluster)
+    run2_miniAOD_94XFall17.toModify(process.patJets, addTagInfos = cms.bool(False) )
+    run2_miniAOD_94XFall17.toModify(process.slimmedJetsNoDeepFlavour, dropTagInfos = cms.string('1') )
```
slava77: here and other parts of this file/pr
please drop type specifications for all parameters which already exist.

slava77:
we usually do not create a new package for just one plugin/algorithm.
Can we use an existing package.
@fabiocos I see that you created a PR for the cms-bot to accept this package.
Was there a policy change?

zucchett:
Creating a new package was motivated by the fact that the new tagger being developed is very different from all other approaches to b-tagging, and therefore cannot be associated to any existing package. It's also likely that in the near future the tagger will be expanded and new tools and plugins will be added, thus requiring the creation of the new package anyway.

DataFormats/BTauReco/src/classes_def.xml
```diff
@@ -465,4 +465,10 @@
   <class name="reco::DeepBoostedJetTagInfoRefVector"/>
   <class name="edm::Wrapper<reco::DeepBoostedJetTagInfoCollection>" persistent="false"/>
 
+  <class name="reco::PixelClusterData"/>
+  <class name="reco::PixelClusterProperties"/>
+  <class name="reco::PixelClusterTagInfo"/>
+  <class name="reco::PixelClusterTagInfoCollection"/>
```
slava77: all non-templated classes should have a class version for a more appropriate schema evolution tracking.

RecoBTag/PixelCluster/plugins/PixelClusterTagInfoProducer.cc
```diff
@@ -0,0 +1,262 @@
...
+#include "Geometry/TrackerGeometryBuilder/interface/PixelGeomDetUnit.h"
+#include "Geometry/TrackerGeometryBuilder/interface/PixelGeomDetType.h"
+#include "Geometry/TrackerGeometryBuilder/interface/TrackerGeometry.h"
+#include "Geometry/Records/interface/TrackerDigiGeometryRecord.h"
+#include "Geometry/CommonDetUnit/interface/GeomDetType.h"
+#include "Geometry/CommonTopologies/interface/PixelTopology.h"
+
+// ROOT
+#include "TVector3.h"
+#include "TLorentzVector.h"
```
slava77: please switch to non-TObject vector variants.
See e.g.  in DataFormats/Math/interface

RecoBTag/PixelCluster/plugins/PixelClusterTagInfoProducer.cc
```diff
@@ -0,0 +1,262 @@
...
+PixelClusterTagInfoProducer::PixelClusterTagInfoProducer(const edm::ParameterSet& iConfig)
+    : m_jets(consumes<edm::View<reco::Jet> >(iConfig.getParameter<edm::InputTag>("jets"))),
+      m_vertices(consumes<reco::VertexCollection>(iConfig.getParameter<edm::InputTag>("vertices"))),
+      m_pixelhit(consumes<edmNew::DetSetVector<SiPixelCluster> >(iConfig.getParameter<edm::InputTag>("pixelhit"))),
+      m_isPhase1(iConfig.getParameter<bool>("isPhase1")),
+      m_addFPIX(iConfig.getParameter<bool>("addForward")),
+      m_minADC(iConfig.getParameter<int>("minAdcCount")),
+      m_minJetPt((float)iConfig.getParameter<double>("minJetPtCut")),
+      m_maxJetEta((float)iConfig.getParameter<double>("maxJetEtaCut")),
+      m_hadronMass((float)iConfig.getParameter<double>("hadronMass")),
```
slava77: were these casts really needed?

RecoBTag/PixelCluster/plugins/PixelClusterTagInfoProducer.cc
```diff
@@ -0,0 +1,262 @@
...
+
+  // Retrieve tracker topology from geometry
+  edm::ESHandle<TrackerTopology> tTopoH;
+  iSetup.get<TrackerTopologyRcd>().get(tTopoH);
+  const TrackerTopology* tTopo = tTopoH.product();
+
+  std::vector<reco::PixelClusterProperties> clusters;
+
+  // Get vector of detunit ids, and fill a vector of PixelClusterProperties in the loop
+  for (auto& detUnit : collectionClusters) {
```
slava77: ```auto const&```

RecoBTag/PixelCluster/plugins/PixelClusterTagInfoProducer.cc
```diff
@@ -0,0 +1,262 @@
...
+      PixelBarrelName pbn(detId, tTopo, m_isPhase1);
+      layer = pbn.layerName();
+    } else if (m_addFPIX && subid == 2) {  // pixel forward
+      PixelEndcapName pen(detId, tTopo, m_isPhase1);
+      layer = pen.diskName();
+    }
+    if (layer == 0 || layer > m_nLayers)
+      continue;
+
+    for (auto& clUnit : detUnit) {
```
slava77: ```auto const&```

RecoBTag/PixelCluster/plugins/PixelClusterTagInfoProducer.cc
```diff
@@ -0,0 +1,262 @@
...
+  for (unsigned int j = 0, nj = collectionJets->size(); j < nj; j++) {
+    if (collectionJets->at(j).pt() < m_minJetPt)
+      continue;
+
+    edm::RefToBase<reco::Jet> jetRef = collectionJets->refAt(j);  // Get jet RefToBase
+
+    reco::PixelClusterData data(m_nLayers);
+    reco::PixelClusterTagInfo tagInfo;
+
+    for (auto& cluster : clusters) {
```
slava77: ```auto const&```

RecoBTag/PixelCluster/python/pixelClusterTagInfos_cfi.py
```diff
@@ -0,0 +1,14 @@
+import FWCore.ParameterSet.Config as cms
+
+pixelClusterTagInfos = cms.EDProducer("PixelClusterTagInfoProducer",
```
slava77: this file is redundant, mostly.
```isPhase1``` should be set appropriately via an era

File: PhysicsTools/PatAlgos/python/recoLayer0/bTagging_cff.py [hunk 27]
```diff
@@ -202,6 +204,8 @@
   , 'pfMassIndependentDeepDoubleCvLJetTags:probHcc'                     : [["pfDeepDoubleXTagInfos"], ['pfBoostedDoubleSVAK8TagInfos', "pfImpactParameterAK8TagInfos", 'pfInclusiveSecondaryVertexFinderAK8TagInfos']]
   , 'pfMassIndependentDeepDoubleCvBJetTags:probHbb'                     : [["pfDeepDoubleXTagInfos"], ['pfBoostedDoubleSVAK8TagInfos', "pfImpactParameterAK8TagInfos", 'pfInclusiveSecondaryVertexFinderAK8TagInfos']]
   , 'pfMassIndependentDeepDoubleCvBJetTags:probHcc'                     : [["pfDeepDoubleXTagInfos"], ['pfBoostedDoubleSVAK8TagInfos', "pfImpactParameterAK8TagInfos", 'pfInclusiveSecondaryVertexFinderAK8TagInfos']]
+  # Pixel Cluster tag infos
+  , 'pixelClusterBJetTags'                               : [['pixelClusterTagInfos']]
```
slava77: what does this refer to ? I do not see a ```pixelClusterBJetTags``` defined

PhysicsTools/PatAlgos/python/slimming/miniAOD_tools.py
```diff
@@ -424,6 +424,32 @@ def miniAOD_customizeCommon(process):
     delattr(process, 'selectedUpdatedPatJetsPuppiJetSpecific')
 
     task.add(process.slimmedJetsPuppi)
+
+    # Embed pixelClusterTagInfos in slimmedJets
+    process.patJets.addTagInfos = cms.bool(True)
+    process.patJets.tagInfoSources.append( cms.InputTag("pixelClusterTagInfos") )
```
slava77: shouldn't this be done with the standard btag tools? PhysicsTools/PatAlgos/python/tools/jetTools.py

RecoBTag/Configuration/python/RecoBTag_cff.py
```diff
@@ -35,7 +36,10 @@
     softPFElectronBJetTags,
 
     # overall combined taggers
-    combinedMVAV2BJetTags
+    combinedMVAV2BJetTags,
+    
+    # pixel cluster
+    pixelClusterTagInfos,
```
slava77: BTW, what is the meaning of "legacy" here and why is newly added ```pixelClusterTagInfos``` expected to belong to it?

RecoBTag/Configuration/python/RecoBTag_cff.py
```diff
@@ -79,3 +85,17 @@
...
+# Create b-tagging sequences without pixelClusterTagInfos
+pfBTaggingTaskNoPixelCluster = pfBTaggingTask.copy()
+pfBTaggingTaskNoPixelCluster.remove(pixelClusterTagInfos)
+
+btaggingNoPixelCluster = btagging.copy()
+btaggingNoPixelCluster.remove(pixelClusterTagInfos)
+
+## FastSim modifier (do not run pixelClusterTagInfos because the siPixelCluster input collection is not produced in FastSim)
+from Configuration.Eras.Modifier_fastSim_cff import fastSim
+fastSim.toReplaceWith(btagging, btaggingNoPixelCluster)
```
slava77: wouldn't it be easier to short-circuit the ```pixelClusterTagInfos``` module configuration than manipulate the sequences?
Also, if these \*NoPixelCluster variants are not meant to be exported, better prefix with ```_```

RecoBTag/PixelCluster/plugins/PixelClusterTagInfoProducer.cc
```diff
@@ -0,0 +1,262 @@
...
+    unsigned int subid = detId.subdetId();  // Subdetector type, pix barrel = 1, forward = 2
+
+    // Get the geom-detector
+    const PixelGeomDetUnit* geomDet = dynamic_cast<const PixelGeomDetUnit*>(theTracker.idToDet(detId));
+    const PixelTopology* topol = &(geomDet->specificTopology());
+    int layer = 0;  // The layer index is in range 1-4
+
+    if (subid == 1) {  // pixel barrel
+      PixelBarrelName pbn(detId, tTopo, m_isPhase1);
+      layer = pbn.layerName();
```
slava77: Isn't the TrackerTopology enough?
```pxbLayer``` and ```pxfDisk``` accessors are available .

**Comment by slava77** [comment]
>  and BTV conveners @mverzett , @kskovpen, @XavierAtCERN

also @ferencek 

was this code also reviewed by the BTV group?
At a glance, not everything looks like it follows other btag setup practices.

File: DataFormats/BTauReco/src/classes_def.xml (hunk 20 above)
Comment (zucchett) [reply DataFormats/BTauReco/src/classes_def.xml]: The class version (number 3) has been added to PixelClusterTagInfo and PixelClusterData.

File: DataFormats/BTauReco/interface/PixelClusterTagInfo.h [hunk 32]
```diff
@@ -0,0 +1,69 @@
...
+    PixelClusterTagInfo* clone(void) const override { return new PixelClusterTagInfo(*this); }
+
+    // method to set the jet RefToBase
+    void setJetRef(const edm::RefToBase<Jet>& ref) { jetRef_ = ref; }
+
+    // method to jet the jet RefToBase
+    edm::RefToBase<Jet> jet() const override { return jetRef_; }
+
+    // method to set the PixelClusterData
+    void setData(const PixelClusterData& data) { pixelClusters_ = data; }
```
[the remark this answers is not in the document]
zucchett: We actually use them in the TagInfo Producer, and we would like to keep the possibility to update the jet ref and the data object separately.

RecoBTag/PixelCluster/plugins/PixelClusterTagInfoProducer.cc
```diff
@@ -0,0 +1,262 @@
...
+//
+//
+
+// system include files
+#include <memory>
+
+// user include files
+#include "FWCore/Framework/interface/Frameworkfwd.h"
+#include "FWCore/Framework/interface/stream/EDProducer.h"
+#include "FWCore/Framework/interface/global/EDProducer.h"
```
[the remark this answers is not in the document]
zucchett: The /stream line has been removed.

RecoBTag/PixelCluster/plugins/PixelClusterTagInfoProducer.cc
zucchett: Ok, we replaced both TVector3 with GlobalPoints and reco::deltaR function.

RecoBTag/PixelCluster/plugins/PixelClusterTagInfoProducer.cc
zucchett: They were introduced following a comment from @perrotta asking to make these values floats instead of double.

RecoBTag/PixelCluster/plugins/PixelClusterTagInfoProducer.cc
zucchett: Ok, we adopted the suggestion

RecoBTag/PixelCluster/plugins/PixelClusterTagInfoProducer.cc
zucchett: Same as before

RecoBTag/PixelCluster/plugins/PixelClusterTagInfoProducer.cc
zucchett: Same as before

RecoBTag/PixelCluster/python/pixelClusterTagInfos_cfi.py
zucchett: We added the era modifier from Configuration/Eras/python/Modifier_pixel_2016_cff

PhysicsTools/PatAlgos/python/slimming/miniAOD_tools.py
zucchett: A line in jetTools.py to include pixelClusterTagInfos has been added since the beginning (L595 in jetTools.py)
In principle, it's not necessary to use the jetTool or perform additional updateJetCollection steps: everything we need is already there, we just need to set the existing modules to keep the pixelClusterTagInfos and not discard them.

RecoBTag/Configuration/python/RecoBTag_cff.py
zucchett: The legacy b-tagging sequence has been there a long time before this PR, and my understanding (BTV experts may want to comment) is that the difference with the standard sequence is the use of PF-based taggers vs non-PF. PixelClusters do not rely on PF but on low level information, so it should be the same for both PF and non-PF.

RecoBTag/Configuration/python/RecoBTag_cff.py
zucchett: Not sure what short-circuit would mean here. The *NoPixelCluster are actually exported, for instance in miniAOD_tools.py for Era modifiers. Simply adding the underscore instead of  NoPixelCluster would make the name less clear (in my opinion).

RecoBTag/PixelCluster/plugins/PixelClusterTagInfoProducer.cc
zucchett: The pixel layer is accessed directly from TrackerTopology::layer(), which uses pxbLayer and pxfDisk.

PhysicsTools/PatAlgos/python/slimming/miniAOD_tools.py
zucchett: Type specifications have been dropped from this file (the other cfg being the pixelClusterTagInfos_cfi), except for InputTags.

zucchett:
Tests have been performed on TTbar RelVal samples with simulated PU (/store/relval/CMSSW_10_5_0_pre2/RelValTTbar_13/GEN-SIM-RECO/PU25ns_105X_upgrade2018_realistic_v2-v1) by running the full b-tag sequence without and with the pixelClusterTagInfoProducer:

Peak virtual memory size went from 1966.84 Mbytes to 1967.13 Mbytes (<0.1% increase).

CPU total job increase is from 266.05 to 268.205 s, but even larger variations are observed by repeating the measurement.

Average processing time of the PixelClusterTagInfoProducer with TimeModule over 100 events: 0.0030 s.

PhysicsTools/PatAlgos/python/recoLayer0/bTagging_cff.py
zucchett: This line can be safely deleted.

PhysicsTools/PatAlgos/python/slimming/miniAOD_tools.py
zucchett: Type specifications have been dropped, except for TagInfos and VTagInfos.

RecoBTag/PixelCluster/plugins/PixelClusterTagInfoProducer.cc
slava77: but the target values are already floats. The additional explicit cast does nothing.

zucchett: Ok, i see. I removed the casts in the commit i just pushed.

slava77:
> Average processing time of the PixelClusterTagInfoProducer with TimeModule over 100 events: 0.0030 s.

perhaps an igprof would be a better measure.

zucchett:
IgProf reports, although very lengthy, are here:
https://zucchett.web.cern.ch/zucchett/Pixel/IgProf/igreport_perf_pixel.res (full b-tag sequence)
https://zucchett.web.cern.ch/zucchett/Pixel/IgProf/igreport_perf.res (only PixelClusterTagInfo Producer)
the TimeModule results, filtered only for the PixelClusterTagInfo to be interpreted more easily, are here:
https://zucchett.web.cern.ch/zucchett/Pixel/IgProf/times.txt
reporting the time for each event. The results can also be analyzed in this spreadsheet:
https://docs.google.com/spreadsheets/d/1ZodegFYdsymzVAabkHpDBiYHdx1YxTk7gX7RHtCIFEE/edit?usp=sharing
The ~50% fraction of event with a very small time correspond to the events with no central jets with pT>100 GeV.
pixelClusterTagInfoProducer

slava77:
On 8/22/19 11:13 AM, Alberto Zucchetta wrote:
> IgProf reports, although very lengthy, are here:

please make a browsable version.
Thank you.

zucchett:
I added the browsable versions. For CPU use:
http://zucchett.web.cern.ch/zucchett/cgi-bin/igprof-navigator/igreport_perf
and memory use:
http://zucchett.web.cern.ch/zucchett/cgi-bin/igprof-navigator/igreport_total

PhysicsTools/PatAlgos/python/slimming/miniAOD_tools.py
ferencek: This is somewhat in collision with changes made in ` PhysicsTools/PatAlgos/python/slimming/MicroEventContent_cff.py`. How do you plan to access these TagInfos later on? If you plan to access them directly, then adding `'keep *_pixelClusterTagInfos_*_*'` is fine and you don't need to set `dropTagInfos = '0'` and `addTagInfos = True`. In fact, all of this addition of TagInfos is for nothing because they are later dropped by the output module because there is no explicit keep statement for them. However, if they were kept, there would be a big jump in the event size because you are not only adding pixelClusterTagInfo (note the `tagInfoSources.append( cms.InputTag("pixelClusterTagInfos") )` call) but all other TagInfos.

If you really need to add TagInfo to slimmed jets, keep in mind that they are really stored in a separate collection (the name starts with `recoBaseTagInfosOwned_...`) so in that case you would have to ensure that that particular collection of TagInfos is kept, that you only keep pixel cluster TagInfos (so no appending) and you would also need to implement an appropriate TagInfo accessor method in `pat::Jet` similar to those in https://github.com/cms-sw/cmssw/blob/CMSSW_10_6_2/DataFormats/PatCandidates/interface/Jet.h#L198-L210.

slava77:
>  For CPU use:
> http://zucchett.web.cern.ch/zucchett/cgi-bin/igprof-navigator/igreport_perf

thank you for the link.
Please point me to the config.py or cmsDriver command used to run this.
I see that only one module shows up in the profile.

slava77:
from http://zucchett.web.cern.ch/zucchett/cgi-bin/igprof-navigator/igreport_perf/477
it looks like we need at least an order of magnitude more events processed to have a good measurement.

zucchett:
I copied the config here:
https://zucchett.web.cern.ch/zucchett/Pixel/IgProf/testRun.py
i'm re-running the memory tests with 1000 events.

slava77:
> I copied the config here:
> https://zucchett.web.cern.ch/zucchett/Pixel/IgProf/testRun.py
> i'm re-running the memory tests with 1000 events.

which release is this?
I see that you have ```auto:run2_mc_FULL``` which is nebulously pointing to ```phase1_2021_realistic``` since 1100pre3 (run3 global tag), compared to the ```105X_upgrade2018_realistic_v2``` used to make the test file.

zucchett:
I am currently using a CMSSW_11_0_0_pre5. I'll switch to the `105X_upgrade2018_realistic_v2` to rerun the tests.

RecoBTag/PixelCluster/plugins/PixelClusterTagInfoProducer.cc
```diff
@@ -0,0 +1,253 @@
...
+    unsigned int subid = detId.subdetId();  // Subdetector type, pix barrel = 1, forward = 2
+    if (!(subid == PixelSubdetector::PixelBarrel || (m_addFPIX && subid == PixelSubdetector::PixelEndcap)))
+      continue;
+    unsigned int layer = tTopo->layer(detId);  // The layer index is in range 1-4 or 1-3
+    if (layer == 0 || layer > m_nLayers)
+      continue;
+
+    // Get the geom-detector
+    const PixelGeomDetUnit* geomDet = dynamic_cast<const PixelGeomDetUnit*>(theTracker.idToDet(detId));
+    const PixelTopology* topol = &(geomDet->specificTopology());
```
slava77: it seems like these casts are not needed. IIUC, all methods are virtual and are available in the base classes.
So, the following should supposedly work
```C++
const auto* geomDet = theTracker.idToDet(detId);
const auto* topol = &geomDet->topology();
```

PhysicsTools/PatAlgos/python/slimming/miniAOD_tools.py
zucchett: Thanks Dinko for the suggestion. Indeed, the plan is to use the second method (be able to  get the Tag Infos from slimmedJets). In the latest commit, we drop `*_pixelClusterTagInfos_*_*`, but we now keep `recoBaseTagInfosOwned_slimmedJets_*_*`, replacing the `tagInfoSources.append( cms.InputTag("pixelClusterTagInfos") )` with `tagInfoSources = cms.VInputTag( cms.InputTag("pixelClusterTagInfos") )` in `miniAOD_tools.py`. We also implemented the new accessor method in Jet.h and Jet.cc.

RecoBTag/PixelCluster/plugins/PixelClusterTagInfoProducer.cc
zucchett: Ok, no error is raise during compilation or runtime. We implemented this in the latest commit.

zucchett:
In the latest commits, we implemented Dinko's suggestion (BTV review) to handle and access the tagInfos in miniAODs, and updated the igprof tests with more events and the latest comments from @slava77 . The links are the same:
http://zucchett.web.cern.ch/zucchett/cgi-bin/igprof-navigator/igreport_perf
http://zucchett.web.cern.ch/zucchett/cgi-bin/igprof-navigator/igreport_total

RecoBTag/Configuration/python/RecoBTag_cff.py
perrotta: @zucchett , I think that the suggestion from @slava77 was to apply the modifiers to the pixelClusterTagInfos module, and not to the main reco BTag sequence (or miniAOD).
You can use modifiers to replace pixelClusterTagInfos with some dummy/empty module inside RecoBTag/PixelCluster/python/pixelClusterTagInfos_cfi.py, and you don't have to apply modifiers elsewhere

PhysicsTools/PatAlgos/python/slimming/MicroEventContent_cff.py
```diff
@@ -11,8 +11,8 @@
         'keep *_slimmedTausBoosted_*_*',
         'keep *_slimmedCaloJets_*_*',
         'keep *_slimmedJets_*_*',
-        # drop content created by MINIAOD DeepFlavour production
-        'drop recoBaseTagInfosOwned_slimmedJets_*_*',
+        # keep slimmedJets TagInfos, currently only PixelClusterTagInfo
+        'keep recoBaseTagInfosOwned_slimmedJets_*_*',
```
perrotta: This line is redundant, as it is fully contained in L13, but maybe it can be left as a reminder for the presence of those TagInfos.
@ferencek please anyhow have a look and confirm whether this satisfies your comment.

RecoBTag/Configuration/python/RecoBTag_cff.py
zucchett: Ok, i can short-circuit the PixelCluster producer by setting a very high threshold on jet pT, so an empty collection is produced and does not fail because it never looks for the siPixelCluster collection. I moved the modifiers to pixelClusterTagInfos_cfi.py, and removed them from the b-tagging sequence and partially from miniAOD_tools.py. However in the latter i still have to avoid that the code looks for the pixelClusterTagInfo collection, so some modifiers on the modules in the miniAOD sequnce are still needed.

PhysicsTools/PatAlgos/python/slimming/MicroEventContent_cff.py
ferencek: Thinks look ok to me now. Indeed, L15 is redundant but I agree it probably doesn't hurt to leave it along with L14 as a reminder.

perrotta:
please test

PhysicsTools/PatAlgos/python/slimming/miniAOD_tools.py
```diff
@@ -424,6 +424,25 @@ def miniAOD_customizeCommon(process):
     delattr(process, 'selectedUpdatedPatJetsPuppiJetSpecific')
 
     task.add(process.slimmedJetsPuppi)
+
+    # Embed pixelClusterTagInfos in slimmedJets
+    process.patJets.addTagInfos = True
+    process.patJets.tagInfoSources = cms.VInputTag( cms.InputTag("pixelClusterTagInfos") )
+    process.slimmedJetsNoDeepFlavour.dropTagInfos = '0'
+    process.updatedPatJetsSlimmedDeepFlavour.addTagInfos = True
+    process.updatedPatJetsSlimmedDeepFlavour.tagInfoSources = cms.VInputTag( cms.InputTag("pixelClusterTagInfos") )
```
ferencek: Looking at it more carefully, any changes to `updatedPatJetsSlimmedDeepFlavour` are not needed because TagInfos get added to `updatedPatJetsTransientCorrectedSlimmedDeepFlavour`.

zucchett: Thanks @ferencek , indeed changes to `updatedPatJetsSlimmedDeepFlavour` do not have any effect. They have been removed now also from these lines and the era modifiers, few lines below.

perrotta:
please test

perrotta:
I evaluated the extra size and extra cpu time produced by this PR on a (small) sample of 100 events produced with wf 11024 (2018 TTbar with PU).

- Event size in reco output, 70 B/evt on average:
```
      0.0 ->        69.7         70     NEWO   0.00     recoPixelClusterTagInfos_pixelClusterTagInfos__RECO.
  3525365 ->     3525425         60             0.0     ALL BRANCHES
```

- Event size miniAOD output, also about 70 B/evt on average:
```
      0.0 ->        67.5         67     NEWO   0.09     recoBaseTagInfosOwned_slimmedJets_tagInfos_RECO.
    78137 ->       78211         74             0.1     ALL BRANCHES
```

- CPU timing:
The new pixelClusterTagInfos module adds about 3 ms/evt to the processing time, which represents 0.02% of the total reco + miniAOD processing time (no validation, dqm steps):
```
The same excluding the first 1 events
  delta/mean delta/orJob     original                   new       module name
  ---------- ------------     --------                  ----       ------------
       added      +0.02%         0.00 ms/ev ->         2.95 ms/ev pixelClusterTagInfos
  ---------- ------------     --------                  ----       ------------
Job total:  15.7006 s/ev ==> 15.2512 s/ev
```

RecoBTag/PixelCluster/plugins/PixelClusterTagInfoProducer.cc
```diff
@@ -0,0 +1,253 @@
...
+// For pixel clusters
+#include "DataFormats/SiPixelCluster/interface/SiPixelCluster.h"
+#include "DataFormats/Common/interface/DetSetVectorNew.h"
+#include "DataFormats/Common/interface/Ref.h"
+#include "DataFormats/DetId/interface/DetId.h"
+#include "DataFormats/TrackerCommon/interface/TrackerTopology.h"
+
+// Pixel topology
+#include "DataFormats/SiPixelDetId/interface/PixelSubdetector.h"
+#include "DataFormats/SiPixelDetId/interface/PixelBarrelName.h"
```
perrotta: This include doesn't seem to be needed

RecoBTag/PixelCluster/plugins/PixelClusterTagInfoProducer.cc
```diff
@@ -0,0 +1,253 @@
...
+#include "DataFormats/SiPixelDetId/interface/PixelEndcapName.h"
```
perrotta: This include doesn't seem to be needed

RecoBTag/PixelCluster/plugins/PixelClusterTagInfoProducer.cc
```diff
@@ -0,0 +1,253 @@
...
+#include "DataFormats/SiPixelDetId/interface/PXBDetId.h"
```
perrotta: This include doesn't seem to be needed

RecoBTag/PixelCluster/plugins/PixelClusterTagInfoProducer.cc
```diff
@@ -0,0 +1,253 @@
...
+#include "DataFormats/SiPixelDetId/interface/PXFDetId.h"
```
perrotta: This include doesn't seem to be needed

RecoBTag/PixelCluster/plugins/PixelClusterTagInfoProducer.cc
```diff
@@ -0,0 +1,253 @@
...
+
+// Geometry
+#include "DataFormats/Math/interface/deltaR.h"
+#include "DataFormats/GeometryVector/interface/GlobalPoint.h"
+#include "Geometry/Records/interface/IdealGeometryRecord.h"
```
perrotta: This include doesn't seem to be needed

RecoBTag/PixelCluster/plugins/PixelClusterTagInfoProducer.cc
```diff
@@ -0,0 +1,253 @@
...
+#include "Geometry/TrackerGeometryBuilder/interface/PixelGeomDetUnit.h"
+#include "Geometry/TrackerGeometryBuilder/interface/PixelGeomDetType.h"
```
perrotta: This include doesn't seem to be needed

RecoBTag/PixelCluster/plugins/PixelClusterTagInfoProducer.cc
```diff
@@ -0,0 +1,253 @@
...
+#include "Geometry/TrackerGeometryBuilder/interface/TrackerGeometry.h"
+#include "Geometry/Records/interface/TrackerDigiGeometryRecord.h"
+#include "Geometry/CommonDetUnit/interface/GeomDetType.h"
```
perrotta: This include doesn't seem to be needed

RecoBTag/PixelCluster/plugins/PixelClusterTagInfoProducer.cc
```diff
@@ -0,0 +1,253 @@
...
+#include "Geometry/CommonTopologies/interface/PixelTopology.h"
```
perrotta: This include doesn't seem to be needed

RecoBTag/PixelCluster/plugins/PixelClusterTagInfoProducer.cc
```diff
@@ -0,0 +1,253 @@
...
+
+  std::vector<reco::PixelClusterProperties> clusters;
+
+  // Get vector of detunit ids, and fill a vector of PixelClusterProperties in the loop
+  for (auto const& detUnit : collectionClusters) {
+    if (detUnit.empty())
+      continue;
+    DetId detId = DetId(detUnit.detId());  // Get the Detid object
+    unsigned int detType = detId.det();    // det type, pixel = 1
+    if (detType != 1)
```
perrotta: "1" here appears like a magic number. Please replace L166 and L167 with the single
```C++
    if (detId.det() != DetId::Tracker)
```

RecoBTag/PixelCluster/plugins/PixelClusterTagInfoProducer.cc
```diff
@@ -0,0 +1,253 @@
...
+
+    // Get the geom-detector
+    const auto* geomDet = theTracker.idToDet(detId);
+    const auto* topol = &geomDet->topology();
+
+    for (auto const& clUnit : detUnit) {
+      // get global position of the cluster
+      LocalPoint lp = topol->localPosition(MeasurementPoint(clUnit.x(), clUnit.y()));
+      GlobalPoint clustgp = geomDet->surface().toGlobal(lp);
+      if (m_minADC > 0 and clUnit.charge() < m_minADC)
```
perrotta: Please move at the beginning of the loop: no reason to compute ```lp``` and ```clustgp``` if below the adc threshold

RecoBTag/PixelCluster/plugins/PixelClusterTagInfoProducer.cc
```diff
@@ -0,0 +1,253 @@
...
+
+  // Loop over jets and perform geometrical matching with pixel clusters
+  for (unsigned int j = 0, nj = collectionJets->size(); j < nj; j++) {
+    if (collectionJets->at(j).pt() < m_minJetPt)
+      continue;
+
+    edm::RefToBase<reco::Jet> jetRef = collectionJets->refAt(j);  // Get jet RefToBase
+
+    reco::PixelClusterData data(m_nLayers);
+    reco::PixelClusterTagInfo tagInfo;
```
perrotta: Move this declaration where it is used, i.e. right before line 227

RecoBTag/PixelCluster/plugins/PixelClusterTagInfoProducer.cc
```diff
@@ -0,0 +1,253 @@
...
+
+    for (auto const& cluster : clusters) {
+      GlobalPoint c3(cluster.x, cluster.y, cluster.z);
+      GlobalPoint v3(firstPV->x(), firstPV->y(), firstPV->z());
```
perrotta: ```v3``` can be computed only once outside these two for loops, either before L192 or right after the definition of ```firstPV``` at L136

RecoBTag/PixelCluster/plugins/PixelClusterTagInfoProducer.cc
```diff
@@ -0,0 +1,253 @@
...
+      float dR = reco::deltaR(c3 - v3, jetRef->momentum());
+      float sC = m_hadronMass * 2. / (jetRef->pt());  // 2 mX / pT
```
perrotta: This doesn't need to be recomputed for every cluster: please move outside this loop

RecoBTag/PixelCluster/plugins/PixelClusterTagInfoProducer.cc
```diff
@@ -0,0 +1,253 @@
...
+
+      // Match pixel clusters to jets and fill Data struct
+      if (cluster.layer >= 1 && cluster.layer <= m_nLayers) {
+        int idx(cluster.layer - 1);
+        if (dR < 0.04)
```
perrotta: Some unneeded computations (although admittedly risible in terms of cpu time) could be saved if these conditions are rewritten as:
```C++
        if (dR < 0.16) {
          data.r016[idx]++;
	  if (dR < 0.10) {
	    data.r010[idx]++;
	    if (dR < 0.08) {
	      data.r008[idx]++;
	      if (dR < 0.06) {
		data.r006[idx]++;
		if (dR < 0.04)
		  data.r004[idx]++;
	      }
	    }
	  }
	}
        if (dR < sC) {
          data.rvar[idx]++;
          data.rvwt[idx] += cluster.charge;
	}
```

perrotta (commented):
I tested the following suggestions and the cpu time seems to shrink from 2.95 ms/evt to 2.75 ms/evt , although the difference could be also partially due to the resolution of the method.  In any case applying these updates doesn't hurt

RecoBTag/PixelCluster/plugins/PixelClusterTagInfoProducer.cc
zucchett: Ok, and i can pass the jet ref and the data object to the constructor.

zucchett:
In the latest commit, i implemented all latest @perrotta suggestions.

perrotta:
please test

perrotta:
Thank you @zucchett , PR code is fine now for reco,

Before finally signing it, could you please verify that the output does correspond to what you are expecting? You can check any root file already produced by the automatic tests, for example.

As far as I can see, we're getting:
- recoPixelClusterTagInfos_pixelClusterTagInfos added to reco/aod content
- recoBaseTagInfosOwned_slimmedJets (new one) and recoBaseTagInfosOwned_slimmedJetsPuppi are saved into the miniAOD event content, according to what specified in MicroEventContent_cff.py

In the reco/aod output I can see the vectors in pixelClusters_ with some reasonable content.
Since it is not obvious to me how to access the tagging variables in miniAOD, I'd let you verify that they correctly contain all what is needed

zucchett:
Thanks @perrotta , i verified that i can access the collection in RECO/AOD and the jet tag Infos from slimmedJets in miniAOD, and get consistent results.

In miniAOD, `recoBaseTagInfosOwned_slimmedJets` is correctly added to the event content, but on the other hand, the `recoBaseTagInfosOwned_slimmedJetsPuppi` in miniAOD was not explicitly introduced by us (PixelCluster TagInfos cannot be neither matched nor recalculated for Puppi jets). I believe it was already included because of the pre-existing `'keep *_slimmedJetsPuppi_*_*'` command.

As long as this is ok for you, you can proceed with this PR.

perrotta:
+1
- recoPixelClusterTagInfos_pixelClusterTagInfos produced and added to reco/aod event content in accordance with what planned in the PR descritpion and adapted to the outcome of the review
- new recoBaseTagInfosOwned_slimmedJets propagates this info into the miniAOD event content
- Jenkins tests show that other reco products are not affected
- pixelClusterTagInfos are not included in FastSim, nor in the eras corresponding to past campaigns

santocch:
+1

peruzzim:
of course XPOG is ok with the (very small) addition to MINIAOD event content

kpedro88:
+1

PR #27740, opened 2019-08-09
Fixes for DDShapes to support DD4hep migration

cvuosalo:
#### PR description:

This PR is a piece of #27674, which has been delayed. This PR has fixes for DDShapes classes. Conversions of angles from the degrees returned by DD4hep to the radians expected in CMS code are added. The `DDTruncTubs` class is corrected so that it returns correct values for its parameters.

For now, `DDPseudoTrap` is commented out as the need for this class is still under discussion.

#### PR validation:

This code was tested as described in #27674.

No backport is needed.

### DIFF
--- a/DetectorDescription/DDCMS/interface/DDShapes.h
+++ b/DetectorDescription/DDCMS/interface/DDShapes.h
@@ -155,7 +155,11 @@ namespace cms {
       double alpha2_;
     };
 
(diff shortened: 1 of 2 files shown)


cvuosalo:
@cmsbuild please test

cvuosalo:
+1

fabiocos:
+1

PR #27752, opened 2019-08-12
DD4hep: Tracker FixSupport

vargasa:
#### PR description:

Addresses overlaps on `FixSupport`. Please check: https://github.com/cms-sw/cmssw/commit/754e568586b57b3743e0558297497e4447ba1876

#### PR validation:

`master`
masterb

`master + thisPR`

thisPRb

`Reference`
referenceb

```
Path:
/Tracker_2/TEC_1/TECWheels_1/TECWheelA_0/TECWheelDiskA_1/TECWheelNomexA_1
```

### DIFF
--- a/Geometry/TrackerCommonData/data/tecwheel6.xml
+++ b/Geometry/TrackerCommonData/data/tecwheel6.xml
@@ -19,7 +19,7 @@
         <Tubs name="TECWheel6" rMin="[tecpetal3:PetalContRmin]" rMax="[tecwheel:WheelRmax]" dz="[tecwheel:WheelT]/2" startPhi="0*deg" deltaPhi="360.*deg"/>
         <Tubs name="TECWheelDisk6" rMin="[tecpetal3:PetalContRmin]" rMax="[tecwheel:WheelRmax]" dz="[tecwheel:DiskT]/2" startPhi="0*deg" deltaPhi="360.*deg"/>
         <Tubs name="TECWheelNomex6" rMin="[tecpetal3:PetalContRmin]" rMax="[tecwheel:NomexRmax]" dz="[tecwheel:NomexT]/2" startPhi="0*deg" deltaPhi="360.*deg"/>
-        <Tubs name="TECFixSupport6" rMin="[tecpetal3:PetalContRmin]" rMax="[FixSuppRmax]" dz="[tecwheel:FixSuppT]/2" startPhi="-[FixSuppW]/2" deltaPhi="[FixSuppW]"/>
+        <Tubs name="TECFixSupport6" rMin="[tecpetal3:PetalContRmin]" rMax="[FixSuppRmax]" dz="[tecwheel:FixSuppT]/2" startPhi="-0.103339312*rad" deltaPhi="0.206678624*rad"/>
         <Tubs name="TECOptConnector6" rMin="[tecwheel:OptConnRmin]" rMax="[tecwheel:OptConnRmax]" dz="[tecwheel:OptConnT2]/2" startPhi="-[tecwheel:OptConnW]/2" deltaPhi="[tecwheel:OptConnW]"/>
         <Box name="TECBeamsplitter" dx="0.5*[BeamsplitterHeight]" dy="0.5*[BeamsplitterWidth]" dz="0.5*[BeamsplitterThick]"/>
     </SolidSection>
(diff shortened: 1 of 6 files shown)


civanch:
please test

cvuosalo:
+1

fabiocos:
@vargasa @ianna @cvuosalo this is affecting basically any geometry that we have, there should be a corresponding database update, am I correct?

ianna:
@fabiocos - I think, the number is the same as in the DB payload. So we do not need to update them.

Geometry/TrackerCommonData/data/tecwheel6.xml
```diff
@@ -19,7 +19,7 @@
         <Tubs name="TECWheel6" rMin="[tecpetal3:PetalContRmin]" rMax="[tecwheel:WheelRmax]" dz="[tecwheel:WheelT]/2" startPhi="0*deg" deltaPhi="360.*deg"/>
         <Tubs name="TECWheelDisk6" rMin="[tecpetal3:PetalContRmin]" rMax="[tecwheel:WheelRmax]" dz="[tecwheel:DiskT]/2" startPhi="0*deg" deltaPhi="360.*deg"/>
         <Tubs name="TECWheelNomex6" rMin="[tecpetal3:PetalContRmin]" rMax="[tecwheel:NomexRmax]" dz="[tecwheel:NomexT]/2" startPhi="0*deg" deltaPhi="360.*deg"/>
-        <Tubs name="TECFixSupport6" rMin="[tecpetal3:PetalContRmin]" rMax="[FixSuppRmax]" dz="[tecwheel:FixSuppT]/2" startPhi="-[FixSuppW]/2" deltaPhi="[FixSuppW]"/>
+        <Tubs name="TECFixSupport6" rMin="[tecpetal3:PetalContRmin]" rMax="[FixSuppRmax]" dz="[tecwheel:FixSuppT]/2" startPhi="-0.103339312*rad" deltaPhi="0.206678624*rad"/>
```
fabiocos: @vargasa why can't we have a named constant and rely on the arithmetic on it? Is it a problem of rounding precision?  This value does not seem to be much more mnemonic in degree unfortunately...

Same comment apply for all the remaining instances of this change

vargasa: @fabiocos Please check https://github.com/cms-sw/cmssw/commit/754e568586b57b3743e0558297497e4447ba1876. Respect to the _old_ geometry there's no effective change, the number is the same as the one that is being resolved. Further inquiry is needed regarding why angles were defined using `1/[L]`. That's in the pipeline! This is needed in order to fix the solids definitions as  `dd4hep` and @ianna, and me is expecting angles to be measured in degrees/radians not in `1/[L]`

ianna: @fabiocos - we could not figure out how the FixSuppW suppose to be calculated. My suggestion was that @vargasa would try to verify the dimensions with the engineering drawing. In the meantime he'd use the calculated number.

ianna: @fabiocos and @vargasa - here is what we have in the DB:
```
<Tubs name="tecwheel6:TECFixSupport6" rMin="309.000000000000000000*mm" rMax="331.442799999999976990*mm" dz="5.000000000000000000*mm" startPhi="-5.920906437629368746*deg" deltaPhi="11.841812875258737492*deg"/>
```

fabiocos: @vargasa @ianna thanks for the clarification, so we expect effectively no change in the payload (did you verify with some analyzer? Ok, the numbers you say are stored in the DB are identical, I agree...).
Still the previous definition was odd, but nothing prevents to define another named constant were you set the hardcoded value, may be even in degrees as stored in the DB (this looks easier to interpret).

ianna: @fabiocos - yes, I accessed and stored the payload locally and compared it with one produced locally. They are identical. 
Unfortunately, we can not sensibly describe what the constant should mean prior to comparison with the engineering drawings. For example, why it's -5.9 deg and not -6 deg? Clearly, it's a result of a calculation which needs to be clarified. So far It's a mystery, so I'd rather leave it as is - a mysterious number :-)

fabiocos: @ianna @vargasa Iet's move it forward, but I understand that Andres keeps checking to better understand, am I correct?

fabiocos:
+1

Geometry/TrackerCommonData/data/tecwheel6.xml
ianna: yes, correct. Thanks, @fabiocos

PR #27782, opened 2019-08-15
Reconfigure particle flow cfi

hatakeyamak:
#### PR description:

Reconfigure particleFlow_cfi and make clear the parameters used for PFEGammaFilters, PFMuonAlgo, i.e. different components of PFAlgo. Also used this opportunity to remove obsolete config parameters. 

#### PR validation:

Compiles and run a few matrix tests with success (scenarios for 2017, 2018, and phase2)

This is not a backport PR.

@bendavid @jpata @guitargeek


RecoParticleFlow/PFProducer/interface/PFAlgo.h
```diff
@@ -85,12 +85,7 @@ class PFAlgo {
                             const edm::ValueMap<reco::PhotonRef>& valueMapGedPhotons);
 
   void setPostHFCleaningParameters(bool postHFCleaning,
-                                   double minHFCleaningPt,
-                                   double minSignificance,
-                                   double maxSignificance,
-                                   double minSignificanceReduction,
-                                   double maxDeltaPhiPt,
-                                   double minDeltaMet);
+				   const edm::ParameterSet& pfHFCleaningParams);
```
slava77: there is a preference to reduce dependence on ```ParameterSet``` this is going against that

hatakeyamak: Hmm. What is the reason behind reducing the dependence on ParameterSet?

This PFHFCleaningParameters and also PFBadHcalMitigationParameters are less important, and we don't have to introduce these PSets, but I think it's a good idea to isolate Parameters for PFEGammaFilters and PFMuonAlgo, which I am hoping that EGamma and Muon POGs keep eyes on and manage (factorizing parameter sets to factorize responsibilities).

slava77: I'm using this https://github.com/cms-sw/cmssw/pull/27010#issuecomment-497804713 as a reference.
Perhaps it does not fully apply in this case: there the tool is delivered via the event setup, while here the tool is created by the ED producer directly.

@Dr15Jones 
is it worth to reduce dependence on the ```ParameterSet``` in the helper algorithms used by ED producers/analyzers?
The factory pattern [not the case here, but somewhat related] even enforces this dependence.

hatakeyamak: Okay. I perfer to leave the PR as it is, but let's see what @Dr15Jones would say.

perrotta:
please test

slava77:
@hatakeyamak 
can we introduce proper parameter set validation?
https://twiki.cern.ch/twiki/bin/view/CMSPublic/SWGuideConfigurationValidationAndHelp
This is a recommended way.
I see that this PR is introducing more "existsAs" cases, which are discouraged and may be OK if there is a good reason for an exception.

hatakeyamak:
I will look into it, but will need some time to understand details.

Basically you are suggesting that we introduce fillDescriptions, right?

For parameterSets for PFEGammaFilters, PFMuonAlgo, etc, I am not sure what is the optimal way. Should I try to have fillDescriptions somehow in PFEGammaFilters.cc, PFMuonAlgo.cc, or it had better in PFProducer.cc (the latter would be simpler, but the former may be a better organization conceptually).

RecoParticleFlow/PFProducer/interface/PFAlgo.h
Dr15Jones: In general, the framework group encourages people to reduce dependencies on the ParameterSet class. It can make it harder to reuse code and the calls to `getParameter` always do parsing so are not super efficient.

slava77:
> I will look into it, but will need some time to understand details.
> Basically you are suggesting that we introduce fillDescriptions, right?

yes, please do.
Thank you.

RecoParticleFlow/PFProducer/interface/PFAlgo.h
hatakeyamak: ok.

slava77:
@hatakeyamak 
please clarify on your current plans for this PR.
Thank you.

hatakeyamak:
Plan to make changes you suggested. I slowed down due to grading of exams... Probably I can get it done by sometime next week.

slava77:
> Plan to make changes you suggested. I slowed down due to grading of exams... Probably I can get it done by sometime next week.

just for a weekly check, please clarify on the status.
Thank you.
(if it will take longer, perhaps this PR can be closed and then reopened or replaced later when the changes are made).

hatakeyamak:
ok. closing for now. but will be working on it later today and tomorrow.

Issue #22582, opened 2018-03-12
Huge  log for phase2 workflows 


boudoul:
Greetings,
The phase2 workflows (2023D21, 2023D17 etc...) are still exhibiting very long logs which make testing (and actually scrutinizing the logs ) very complicated . There is below a copy paste of a (small) fraction of the logs  - Looks like it is on thePFTrackProducer:pfTrack .  Could someone investigate if those are harmless and eventually fix or take actions ? I seem to remember a previous open issue on this but I'm not sure (and can't find it actually) .

Reducing the phase2 logs would be very appreciated . Thank you. 

%MSG-w PFTrackTransformer:  PFTrackProducer:pfTrack 12-Mar-2018 01:16:30 CET  Run: 1 Event: 10
KF TRACK Reco track charge = 1, type = 1, Pt = 8.37763, P = 97.0849
	R0 = 0.010947 Z0 = -1.50682
	number of tracker measurements = 0
	number of points total = 4
Traj point id = -1, layer = 0, Eta,Phi = -5.61786,-2.32022, X,Y = -0.00745722,-0.0080141, R,Z = 0.010947,-1.50682, E,Pt = 97.0849,8.37763
Traj point id = -1, layer = 1, Eta,Phi = -3.19199,-0.756139, X,Y = 1.81873,-1.71529, R,Z = 2.5,-30.3697, E,Pt = 97.085,8.37763
Traj point id = -1, layer = -1, Eta,Phi = 0,0, X,Y = 0,0, R,Z = 0,0, E,Pt = 0,0
Traj point id = -1, layer = -1, Eta,Phi = 0,0, X,Y = 0,0, R,Z = 0,0, E,Pt = 0,0
 PROPAGATION TO THE ECAL HAS FAILED
%MSG
%MSG-w PFTrackTransformer:  PFTrackProducer:pfTrack 12-Mar-2018 01:16:30 CET  Run: 1 Event: 10
KF TRACK Reco track charge = 1, type = 1, Pt = 8.37763, P = 97.0849
	R0 = 0.010947 Z0 = -1.50682
	number of tracker measurements = 0
	number of points total = 6
Traj point id = -1, layer = 0, Eta,Phi = -5.61786,-2.32022, X,Y = -0.00745722,-0.0080141, R,Z = 0.010947,-1.50682, E,Pt = 97.0849,8.37763
[...]

PR #27785, opened 2019-08-15
Properly veto HGCal superclusters from entering regular PFBlocks - phase2 scenarios

hatakeyamak:
#### PR description:

Fix vetoes of superclusters from HGCAL going into PFBlocks. As we set up the PF workflow now, we are keeping the endcap HGCAL part separate from the barrel and HF sections, and we don't want superclusters going into regular PFBlocks, but this vetoes was not implemented properly, causing many LogError messages.

This should also address issues reported here:
https://github.com/cms-sw/cmssw/issues/22582

#### PR validation:

Compiles. Also made sure the superclusters from HGCAL are properly removed from PFBlocks in a phase2 matrix test 20434.0 (ttbar D41), while retaining superclusters from barrel.

Not backport.

@bendavid @guitargeek @jpata @Sam-Harper (cannot find github name of newer egamma leaders...)

### DIFF
--- a/RecoParticleFlow/PFProducer/plugins/importers/GSFTrackImporter.cc
+++ b/RecoParticleFlow/PFProducer/plugins/importers/GSFTrackImporter.cc
@@ -48,7 +48,8 @@ void GSFTrackImporter::importToBlock(const edm::Event& e, BlockElementImporterBa
         reco::SuperClusterRef scref = seedref->caloCluster().castTo<reco::SuperClusterRef>();
         if (scref.isNonnull()) {
           // explicitly veto HGCal super clusters
-          if (scref->seed()->seed().det() == DetId::Forward)
+          if (scref->seed()->seed().det() == DetId::HGCalEE || scref->seed()->seed().det() == DetId::HGCalHSi ||
+              scref->seed()->seed().det() == DetId::HGCalHSc || scref->seed()->seed().det() == DetId::Forward)
             continue;
           PFBlockElementSCEqual myEqual(scref);
           auto sc_elem = std::find_if(elems.begin(), SCs_end, myEqual);


perrotta:
please test

hatakeyamak:
I think the change is made as intended, but I don't quite understand some changes in btag and photons for phase2. Thinking... Any inputs will be appreciated.

perrotta:
@hatakeyamak , was there any progress in it?

Looking at the differences in the Phase2 workflows, there are less elements in the PFBlocks , which is exactly what this PR is intended to. Then this propagates into differences in other quantities like secondary vertices, impact parameters, which end up in the differences for photons and btag that you are mentioning. Is it really unexpected, due to the removal of those PFblocks from HGCal?

perrotta:
@hatakeyamak @kpedro88 : do you confirm that the changes observed are all expected?

perrotta:
> > @hatakeyamak , was there any progress in it?
> > Looking at the differences in the Phase2 workflows, there are less elements in the PFBlocks , which is exactly what this PR is intended to. Then this propagates into differences in other quantities like secondary vertices, impact parameters, which end up in the differences for photons and btag that you are mentioning. Is it really unexpected, due to the removal of those PFblocks from HGCal?
> 
> Sorry for slow progress. Something was keeping me occupied for ~1 week, but I am almost done.
> 
> I was hoping to use this opportunity to understand how changes like this propagate to downstream. Just for my education, how/where do you see:
> 
> * there are less elements in the PFBlocks
> * differences in secondary vertices, impact parameters
>   Are you looking at specific plots from automatic DQM/validation plots from matrix/jenkins tests?

Yes, I simply looked at the differences in the "comparison failed" section of the automatic tests results. The "reco comparison differences" looks at the differences between reco quantities in the root outputs.
I must say that even to me those differences look rather reasonable, given the pruning of the PF blocks. But before signing for reco I would like to understand from you whether you think that all those changes make sense, since you expressed doubts on it. I agree that this is a real bug fix, and it must get fixed: however, while this PR is open I would also like to be sure that it does not point to some other bug somewhere else, which could get forgotten would this PR get merged without inspecting.

perrotta:
@hatakeyamak please either confirm that the changes observed in output look reasonable also to you, or just let us know when you think you can be done with your checks (and we'll put on hold this PR till then)

hatakeyamak:
Thanks for your ping @perrotta .

I did some investigation on this for the last few days, and I think I am ready to sign this off. I still cannot fully connect this change to the changes in secondary vertices, but some changes in photons via changes in particleFlowEGamma resulting from intended changes in PFBlocks are clear.

The changes are certainly doing the right thing for PFBlocks, so I think we can sign this off. Sorry that this took long.

slava77:
@cmsbuild please test

to still have the test results available before merging

perrotta:
+1
- Fix implemented as described
- Jenkins tests witness the removal of the unwanted PFBlocks and show some differences in the related Phase2 outputs; those comparisons are endorsed by PF (see https://github.com/cms-sw/cmssw/pull/27785#issuecomment-527492189), while waiting for a more detailed comparison in the relvals

fabiocos:
+1

Issue #27800, opened 2019-08-19
RelMon: more useful DQM-GUI link


jan-kaspar:
In pages like:
https://cms-pdmv.cern.ch/ReleaseMonitoring/CMSSW_11_0_0_pre4_2018gcc8_vs_CMSSW_11_0_0_pre4_2018/DataReport/SingleMuon2018C-v1_106X_dataRun2_v17_RelVal_2018C_319450/41f33f54ab.html
I'd like to propose that the "To the DQM GUI" link points GUI page related to the RelMon page. This would be much more useful than the current static link. Thanks in advance!

kpedro88:
assign dqm

jfernan2:
This issue was partially solved in the PR above

jan-kaspar:
Thanks!

PR #27804, opened 2019-08-20
First attempt to fix DQM GUI urls in RelMon

jfernan2:
#### PR description:

This PR tries to fix issue https://github.com/cms-sw/cmssw/issues/27800
Links points to root Everything workspace instead of dedicated folder to simplify

#### PR validation:

Tested locally

No backport

### DIFF
--- a/Utilities/RelMon/python/directories2html.py
+++ b/Utilities/RelMon/python/directories2html.py
@@ -258,7 +258,7 @@ def get_summary_section(directory,matrix_page=True):
   html= '<div class="span-6">'+\
         '<h3>Summary</h3>'
   html+=get_dir_stats(directory)
-  html+='<a href="%s/%s">To the DQM GUI...</a>' %(server,base_url)
+  html+='<a href="%s/%s/start?runnr=%s;dataset=/%s/%s/DQMIO;sampletype=offline_data;filter=all;referencepos=on-side;referenceshow=all;referencenorm=True;referenceobj1=other:%s:/%s/%s/DQMIO:;referenceobj2=none;referenceobj3=none;referenceobj4=none;search=;striptype=object;stripruns=;stripaxis=run;stripomit=none;workspace=Everything;size=M;focus=;zoom=no;">To the DQM GUI...</a>' %(server,base_url,meta.run1,meta.sample1,meta.release1,meta.run2,meta.sample2,meta.release2)
   html+='</div>'
         
   html+='<div class="span-7 colborder">'+\
--- a/Utilities/RelMon/python/web/templates/directory_summary.html
+++ b/Utilities/RelMon/python/web/templates/directory_summary.html
@@ -22,7 +22,7 @@
             <li><span class="caps">Null: {{ null }}% ({{ nulls }})</span></li>
             <li><span class="caps">Fail: {{ fail }}% ({{ fails }})</span></li>
         </ul>
-        <a href="http://cmsweb.cern.ch/dqm/relval">To the DQM GUI...</a>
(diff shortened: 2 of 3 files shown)


jfernan2:
please test

perrotta:
I wonder why   "Utilities/RelMon" is in the "reconstruction" category instead of "dqm": just a mistake or is there any reason behind?

slava77:
@smuzaffar @mrodozov 
this should technically be tested with the cms-bot
What is the right command to test with the bot using the updated relmon?

mrodozov:
@slava77 testing the python files could be generalize up to "if they compile" (and they does in this case) but for the "what they produce when run" I think a unit test will be required (if there is none, I don't know). Assuming there is no test the proper way to test it would be to ask @jfernan2 what was used to local test it, add it as a unit test and rerun. Apart from that, the python part just ... compiles. If it didn't the scram command would've failed.

jfernan2:
@mrodozov  in my test I just made sure that the piece of python code I am modifying/introducing was working locally (with a print) as expected, producing the desired string for the url link. Not a real unit test.
Nevertheless, the code as it is now before this PR, it is just putting a plain link to the DQM RelVal GUI, which nobody is using since the actual ROOT files compared by RelMon are not loaded

perrotta:
@jfernan2 : I think that this code, and the whole "Utilities/RelMon" package, should belong to dqm rather than reconstruction. If you agree with it, I could just switch "Utilities/RelMon" to dqm in the categories map. Or alternatively, we can can at least share it if you think there are reasons for having it in reco that I am not aware of.
In the meanwhile, I'd just assign this PR to you

perrotta:
assign dqm

slava77:
> @slava77 testing the python files could be generalize up to "if they compile" (and they does in this case) but for the "what they produce when run" I think a unit test will be required (if there is none, I don't know). Assuming there is no test the proper way to test it would be to ask @jfernan2 what was used to local test it, add it as a unit test and rerun. Apart from that, the python part just ... compiles. If it didn't the scram command would've failed.

Just to be sure, what I meant is that the code updated here is used by cms-bot in the run-pr-comparisons jobs
https://github.com/cms-sw/cms-bot/blob/ae1038f46919118a51636f6aeca3aff7dc6869f4/run-pr-comparisons#L71
My question was if there is a way to run the cms-bot comparisons on top of the build using this PR.

jfernan2:
@perrotta fine by us (DQM). I do not know why it is under reco umbrella (historical reasons?), I just learnt it was this way when creating this PR

mrodozov:
> Not a real unit test

it's not exactly a unit test to test a "unit" but that's not unusual to test pieces like this when required. Example is this:
https://github.com/cms-sw/cmssw/tree/master/PhysicsTools/PythonAnalysis/test
which is full of scripts that are testing something. And we refer to it as "unit test" although they are not exactly unit tests

perrotta:
Thak you @jfernan2 : cms-bot PR https://github.com/cms-sw/cms-bot/pull/1190 makes the switch

perrotta:
+1
- (reco pass-through)

jfernan2:
I am sorry @mrodozov but I am not able to create a test file like you propose which mimics the RelMon environment. 
@slava77 in the run-pr-comparison jobs by cms-bot, RelMon is running standalone so that DQM root files are not uploaded to any DQM GUI, except by the bin-by-bin comparison tool which actually uploads them to the development GUI for a real comparison

jfernan2:
+1
This PR is not breaking any RelMon functionality at all: it just replaces a general link to the DQM RelVal GUI by a dedicated url with the files involved in the comparison. In the worst case scenario, non-existing root files in GUI like RelMon standalone tests or malformed url, the link will still work pointing to the RelVal GUI as it was previously

kpedro88:
+1

fabiocos:
@jfernan2 @perrotta @slava77 we already discussed this point in the recent past, as far as I know RelMon is historically under the reconstruction umbrella because it was originally developed by @dpiparo working with @VinInn to monitor and study workflows and in particular RECO performances. But strictly speaking is more a DQM tool than a RECO one IMO

PR #27811, opened 2019-08-20
Fix gcc 9 warnings DQM/DTMonitorClient - Add fallthroughs to clear the warnings

mrodozov:
#### PR description:

Fixes warnings in gcc9 
Substitutes switch statements with less than 5 cases with if conditions

#### PR validation:

Builds without warnings

### DIFF
--- a/DQM/DTMonitorClient/src/DTLocalTriggerLutTest.cc
+++ b/DQM/DTMonitorClient/src/DTLocalTriggerLutTest.cc
@@ -271,19 +271,17 @@ void DTLocalTriggerLutTest::runClientDiagnostic(DQMStore::IBooker& ibooker, DQMS
           int phiNoData = 0;
           int phibNoData = 0;
           for (int stat = 1; stat <= 4; ++stat) {
-            switch (static_cast<int>(phiWhSummary->GetBinContent(sect, stat))) {
-              case 1:
-                phiNoData++;
-              case 2:
-              case 3:
-                phiErr++;
+            int res = static_cast<int>(phiWhSummary->GetBinContent(sect, stat));
+            if (res == 1) {
+              phiNoData++;
+            } else if (res == 3) {
+              phiErr++;
             }
-            switch (static_cast<int>(phibWhSummary->GetBinContent(sect, stat))) {
-              case 1:
-                phibNoData++;
-              case 2:
-              case 3:
-                phibErr++;
+            res = static_cast<int>(phibWhSummary->GetBinContent(sect, stat));
+            if (res == 1) {
+              phibNoData++;
+            } else if (res == 3) {
+              phibErr++;
             }
           }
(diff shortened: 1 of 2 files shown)


mrodozov:
please test

jfernan2:
+1

makortel:
So the earlier fallthrough was a bug?

mrodozov:
It wasn't de facto, if you see the logic of it there is no way both conditions in any of the switch statements to be matching at the same time. It would've become buggy if I have added a break statements, certainly. The idea of it was to iterate without breaking so all possibilities can be checked

DQM/DTMonitorClient/src/DTLocalTriggerLutTest.cc
```diff
@@ -271,19 +271,17 @@ void DTLocalTriggerLutTest::runClientDiagnostic(DQMStore::IBooker& ibooker, DQMS
...
-                phiNoData++;
-              case 2:
-              case 3:
-                phiErr++;
+            int res = static_cast<int>(phiWhSummary->GetBinContent(sect, stat));
+            if (res == 1) {
+              phiNoData++;
+            } else if (res == 3) {
+              phiErr++;
             }
```
makortel: @mrodozov 
Umm, but isn't replacing the switch with if-else essentially equivalent to adding breaks? I mean, if `res` is
* 1: earlier the coded did `phiNoData++; phiErr++;`, and now only `phiNoData++;`
* 2: earlier `phiErr++;`, now nothing
* 3: earlier and now `phiErr++;`

(and no, I don't know which behavior is the correct one, if either)

mrodozov: Hm, never written switch statements without breaks before, I was assuming it wouldn't go into a case "condition" if it doesn't match but it will. So now I don't know if that code was intentionally left without breaks, adding breaks will lead to the if else condition in the update indeed, better ask the authors again.

mrodozov:
assign dqm
again :) would you please clarify on this

VinInn:
can we stop messing with perfectly valid C++ syntax (switch passthrough)
There is no alternative and to me it is perfectly readable and understandable what is happening there.
The fact that google programmers are not able to master it does not mean that we shall not use it....

please allow passthrough in switch

v.

jfernan2:
-1
I checked with authors and indeed the passthrough was intended

Dr15Jones:
@VinInn see https://github.com/cms-sw/cmssw/pull/27799#issuecomment-523053774

davidlange6:
google programmers?

mrodozov:
note that I'm not removing the unused cases because I'm not assuming anymore whats to remain or be removed.

mrodozov:
please test

jfernan2:
+1

PR #27821, opened 2019-08-22
Remove dead assignments in a few sparse reco packages

perrotta:
#### PR description:

Get rid of a few dead assignments and initialization in several sparse reco packages.
Issues were reported by the static analyzer

The obsolete PFPhotonClusters class (which was the target of one of those static analyzer reported issues) is also removed.

#### PR validation:

It builds and the short matrix passes

No changes are expected in outputs whatsoever. The only real bug fix (which was pointed out by the dead assignment report) is in EcalTBWeightUncalibRecHitProducer.cc: if I'm not wrong. that code is not in use since a while already (@argiro)


perrotta:
please test

RecoParticleFlow/PFClusterTools/src/PFPhotonClusters.cc
```diff
@@ -157,11 +157,9 @@ void PFPhotonClusters::FillClusterShape() {
...
         continue;
 
-      //if(std::abs(dphi)<=2 && std::abs(deta)<=2){
       EBDetId EBidSeed = EBDetId(idseed_.rawId());
       EBDetId EBid = EBDetId(id.rawId());
-      int ind1 = EBidSeed.ieta() - EBid.ieta();
-      int ind2 = EBidSeed.iphi() - EBid.iphi();
+      int ind1;
       if (EBidSeed.ieta() * EBid.ieta() > 0) {
         ind1 = EBid.ieta() - EBidSeed.ieta();
```
perrotta: Sometime these warnings raised by the static analyzer are able to pinpoints small mistakes in the code.
In this case, I have the impression that here it should be
```C++
        ind1 = EBidSeed.ieta() - EBid.ieta();
```
and not its opposite.

Of course, if the exact distribution of energy inside the ```e5x5_``` matrix is not needed, this doesn't matter. But if that shape is needed anywhere, this could be a bug.

@Sam-Harper , could you please have a glance and tell me?

perrotta:
@namapane @fcavallo @battibass FYI

RecoParticleFlow/PFClusterTools/src/PFPhotonClusters.cc
Sam-Harper: likely yes. However this is dead code. Can we delete this class?  How best, you make this PR and then I submit the deletion on top of it?

Best,
Sam

perrotta: Thank you @Sam-Harper : cleaning is always welcome!
I can even do it since now here, if you agree, and then after this PR is merged you can check whether there are other obsolete classes that can also get cleaned. I see that the header of this class has to be removed also from RecoParticleFlow/PFProducer/src/PFEGammaAlgo.cc (where it does nothing, and therefore it can be safely removed)

Sam-Harper: >I can even do it since now here, if you agree,

even better, more than happy for you to go ahead and remove. 

> and then after this PR is merged you can check whether there are other obsolete classes that can also get cleaned. 

There almost certainly are. And while I really really want to do this, I'm currently forbidden from contributing to e/gamma code  until I finish writing my sections of the e/gamma paper. Have to eat my vegetables before I can have desert.

perrotta:
please test

perrotta:
@ptcox FYI just in case (but the updates in RecoLocalMuon/CSCRecHitD/src/CSCRecHitDBuilder.cc are rather straightforward)

perrotta:
+1
- Fixes for issues pinpointed by the static analyzer are all rather straightforward and innocuous, and I'd merge them even without waiting for an explicit confirmation from the responsible of local reco in CSC (@ptcox)  and DT (@namapane @fcavallo @battibass)
- Thanks from the feedback received from EGamma, the obsolete PFPhotonClusters class is removed as not used/needed, thus avoiding fixing it
- I have the impression that also the ```EcalTBWeightUncalibRecHitProducer``` could also be removed, but without having an official endorsement from the ECal contact (@argiro) I'd rather limit myself to simply fixing the issue as proposed in this PR
- Jekins tests pass and show no differences, as expected

fabiocos:
+1

PR #27823, opened 2019-08-22
Simulation of APV dynamic gain

EmyrClement:
#### PR description:

Include a simulation of the APV dynamic gain in TIB and TOB, aiming to improve description of 2016 data before the VFP setting was changed part way through the year.  This improvement is targeting the 2016 UL campaign.

- A set of APV baseline distributions, which represent the charge accumulated on a strip over several bunch crossings, are provided as an input
  - These are stored as text files in SimTracker/SiStripDigitizer/data/
  - Are read in a similar way to the APV pulse shape
  - These are provided per layer, and in bins of PU and z within a layer, and are "unrolled" as one long array of numbers in the text files.
- Each time a charge is deposited on a strip, the APV response is simulated.  This is linear for low APV baselines (little charge remaining on strip from previous interactions), but becomes non-linear for high APV baselines.
  - The end effect reduces the charge that is effectively deposited/read out from a strip
- The dynamic gain simulation is currently turned off in this PR

#### PR validation:

- Validation is still ongoing, but an improved description of the shapes of the cluster charge distribution and cluster multiplicity is seen with these changes.


mmusich:
Hi @EmyrClement can you please squash your commits?
Many of the ASCII files are in invalid state (meaning that they are present in one of the commits but missing in the final one), and this is polluting the repository.
Thanks

SimTracker/SiStripDigitizer/plugins/SiStripDigitizerAlgorithm.h
```diff
@@ -170,6 +184,45 @@ class SiStripDigitizerAlgorithm {
...
+  edm::FileInPath apvBaselinesFile_tob6_;
+
+  unsigned int nTruePU_;
+  unsigned int apvBaselines_nBinsPerBaseline_;
+  double apvBaselines_minBaseline_;
+  double apvBaselines_maxBaseline_;
+  std::vector<double> apvBaselines_puBinEdges_;
+  std::vector<double> apvBaselines_zBinEdges_;
+
+  std::vector<std::vector<std::vector<TH1F>>> apvBaselineHistograms_tib_;
```
makortel: I wonder if there would be any better way to organize the data than triple-vector of TH1F.

Actually, how much memory does this and the corresponding `...tob_`  consume? By quick estimate based on the text file sizes I'd guess the total to be the order of 300 kB (assuming the unnecessary double-vector members are cleaned up, and neglecting the overheads of TH1F). Is the number of z and PU bins expected to increase in the future from the current 1 (all files have a single line; btw the according to config there are 0 bins...)?

With the current approach the memory need gets multiplied by the number of EDM streams (i.e. concurrent events). One possible way to make the memory need independent of the number of EDM streams would be to provide the data to the `SiStripDigitizerAlgorithm` via EventSetup (e.g. an ESProducer that reads the data in and produces an ESProduct with the data and interface to return a properly distributed random number using a given `CLHEP::HepRandomEngine`.

civanch:
@EmyrClement, @makortel, two points:
1) data files are large, may be should be defined as externals?
2) correct me if I am wrong, but vector of objects better to fill in the following way:
   vec.reserve(n);
   for(...) { vec.emplace_back(MyClass(...)); }

If you do vec.resize(n) then emplace_back or push_back should not be used.

makortel:
> 1. data files are large, may be should be defined as externals?

I agree.

> 2. correct me if I am wrong, but vector of objects better to fill in the following way:
>    vec.reserve(n);
>    for(...) { vec.emplace_back(MyClass(...)); }
> 
> If you do vec.resize(n) then emplace_back or push_back should not be used.

I think that is exactly what I suggested.

SimTracker/SiStripDigitizer/plugins/SiStripDigitizerAlgorithm.cc
```diff
@@ -305,6 +430,79 @@ void SiStripDigitizerAlgorithm::digitize(edm::DetSet<SiStripDigi>& outdigi,
...
+
+    // Simulate APV response for each strip
+    if (SubDet == 3 || SubDet == 5) {
+      for (int strip = 0; strip < numStrips; ++strip) {
+        if (detAmpl[strip] > 0) {
+          // Convert charge from electrons to fC
+          double stripCharge = detAmpl[strip] * apv_fCPerElectron;
+
+          // Get APV baseline
+          double baselineV = apvBaselineDistribution->GetRandom();
```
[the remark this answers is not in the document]
EmyrClement: I'm not sure, but perhaps something similar to this may be useful:

https://github.com/cms-sw/cmssw/blob/07d3399c66e34a7aa14773466b70643db096a7f7/Mixing/Base/src/PileUp.cc#L49-L63

Or something like std::discrete_distribution could be used, but I don't know if this could use CLHEP::HepRandomEngine.

SimTracker/SiStripDigitizer/plugins/SiStripDigitizerAlgorithm.cc
```diff
@@ -305,6 +430,79 @@ void SiStripDigitizerAlgorithm::digitize(edm::DetSet<SiStripDigi>& outdigi,
...
+    // Get index in apv baseline distributions corresponding to z of detSet and PU
+    const StripTopology* topol = dynamic_cast<const StripTopology*>(&(det->specificTopology()));
+    LocalPoint localPos = topol->localPosition(0);
+    GlobalPoint globalPos = det->surface().toGlobal(Local3DPoint(localPos.x(), localPos.y(), localPos.z()));
+    float detSet_z = fabs(globalPos.z());
+    std::vector<double>::iterator high =
+        std::upper_bound(apvBaselines_zBinEdges_.begin(), apvBaselines_zBinEdges_.end(), detSet_z);
+    unsigned int detSet_zBin = high - apvBaselines_zBinEdges_.begin() - 1;
+    high = std::upper_bound(apvBaselines_puBinEdges_.begin(), apvBaselines_puBinEdges_.end(), nTruePU_);
+    unsigned int puBin = high - apvBaselines_puBinEdges_.begin() - 1;
```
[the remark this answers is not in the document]
EmyrClement: This could happen if the configuration is not set up correctly.  I can catch this and throw an exception here, or earlier on.

SimTracker/SiStripDigitizer/plugins/SiStripDigitizerAlgorithm.h
EmyrClement: In the current text files, there are 4 bins in z, and 26 bins in PU (apologies, I did not include the correct configuration for these, I will include them).  So the memory estimate of 300kb is an underestimate.

Can you provide a similar example of how to provide these data via EventSetup?

davidlange6:
> On Aug 23, 2019, at 4:36 PM, Matti Kortelainen <notifications@github.com> wrote:
> 
> 	• data files are large, may be should be defined as externals?
> I agree.

rather they "should" be in the conditions database...

SimTracker/SiStripDigitizer/plugins/SiStripDigitizerAlgorithm.cc
```diff
@@ -66,7 +73,35 @@ SiStripDigitizerAlgorithm::SiStripDigitizerAlgorithm(const edm::ParameterSet& co
       theSiNoiseAdder(new SiGaussianTailNoiseAdder(theThreshold)),
       theSiDigitalConverter(new SiTrivialDigitalConverter(theElectronPerADC, PreMixing_)),
       theSiZeroSuppress(new SiStripFedZeroSuppression(theFedAlgo)),
-      APVProbabilityFile(conf.getParameter<edm::FileInPath>("APVProbabilityFile")) {
+      APVProbabilityFile(conf.getParameter<edm::FileInPath>("APVProbabilityFile")),
+      includeAPVSimulation_(conf.getParameter<bool>("includeAPVSimulation")),
+      apvBaselinesFile_tib1_(conf.getParameter<edm::FileInPath>("apvBaselinesFile_tib1")),
+      apvBaselinesFile_tib2_(conf.getParameter<edm::FileInPath>("apvBaselinesFile_tib2")),
```
davidlange6: the use of arrays and loops would save a lot of typing here (and elsewhere in this pr)

mmusich:
> rather they "should" be in the conditions database...

@makortel @davidlange6  this whole feature is not completely developed yet.  
I agree with you that the best solution would be to code a dedicated `CondFormat` to store these parameters (I can work to this in parallel), but it would be important to have something that "works" for medium sized dedicated relval production to tune the parameters of the simulation (and try to be in time for the 2016 Ultra-Legacy production). 
As of now the feature is switched off and there is no overhead on the standard production.
@srimanob @franzoni

davidlange6:
sure - just reminding people that the proper solution for data files full of numbers is to put them in the conditions database and not the release...

srimanob:
Dear @makortel @civanch @davidlange6

I am not sure I get a clear conclusion in this PR. This is PR to push out for RelVals to prepare for 2016 UL. It's not the final development. Do you think this is good to go (assume code is working, data file/external is arranged properly)?

Thanks All.

@kpedro88 @fabiocos Do you think we should try to discuss before the release meeting, so we avoid spending much time on the development? Thx.

kpedro88:
@srimanob to me this PR looks far from ready. Pushing unfinished code always causes problems down the line. Since we are still working on validating 2018 UL, I do not think there is an extreme rush yet.

SimTracker/SiStripDigitizer/plugins/SiStripDigitizerAlgorithm.cc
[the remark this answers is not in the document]
makortel: I suppose that would work. By quick look `CLHEP::HepRandomEngine` does not satisfy [*UniformRandomBitGenerator*](https://en.cppreference.com/w/cpp/named_req/UniformRandomBitGenerator) requirement out of the box (needed for `std::discrete_distribution::operator()()`), but a thin wrapper might be possible.

(personally I'd really like if TH1 could be avoided as it has member data unnecessary for this case, leading to memory bloat, and it has caused thread-safety problems earlier within MixingModule)

SimTracker/SiStripDigitizer/plugins/SiStripDigitizerAlgorithm.h
makortel: Already the plain `sizeof(TH1F)` gives 1000 bytes, TID and TOB together make 10, so `4*26*10*1000 = 0.99 MiB` of TH1F overhead alone (per EDM stream).

Documentation for ESProducers is here https://twiki.cern.ch/twiki/bin/view/CMSPublic/SWGuideEventSetupHowTos
Here is an example of the creation of an ESProducer
https://github.com/cms-sw/cmssw/blob/master/RecoTracker/FinalTrackSelectors/plugins/LwtnnESProducer.cc
Here is a minimum what it required to do
https://github.com/makortel/cmssw/commit/ce7239c6fa1
I'm sure @mmusich can suggest a record to use (presumably something whose IOV is the whole job)

SimTracker/SiStripDigitizer/plugins/SiStripDigitizerAlgorithm.cc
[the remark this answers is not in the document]
makortel: An exception as early as possible would be the best solution.

SimTracker/SiStripDigitizer/plugins/SiStripDigitizerAlgorithm.cc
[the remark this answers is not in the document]
EmyrClement: Actually, CLHEP::RandGeneral performs the same task, so I expect this would work without an additional wrapper.  I missed this first time looking at through CLHEP.

SimTracker/SiStripDigitizer/plugins/SiStripDigitizerAlgorithm.h
pieterdavid: If we use a single vector, with a [`PhysicsTools::Calibration::HistogramF3D`](https://github.com/cms-sw/cmssw/blob/master/CondFormats/PhysicsToolsObjects/interface/Histogram3D.h) (pu,z,V0) for each layer, the size of the value arrays (without overhead) should be about `10(layers)*4(z)*26(PU)*82(V0)*4 = 333.125kB` (we would need to implement the random sampling from a (pu,z) row of it, but from the other discussion we should do that anyway). I could write an ESProducer for this (IIRC this is also the starting point for adding it to the database, right?).
As far as I know the only MC-specific SiStrip record we have is the SiStripApvGainSimRcd, we could add this there (it's somewhat related), but can the database put two objects of different type in the same record with the same global tag? Otherwise we need to add a new record anyway.

mmusich: > we could add this there (it's somewhat related), but can the database put two objects of different type in the same record with the same global tag? 

The same record cannot hold two different C++ object abstractions. 
`SiStripApvGainSimRcd` is mapped onto `SiStripApvGain`

https://github.com/cms-sw/cmssw/blob/eb4b5d2150c7e93458cd13d13540513a99a88471/CondCore/SiStripPlugins/plugins/plugin.cc#L33

 which stores the information of the gain for each apv using four vectors, 
  * A vector<unsigned int> (v_detids) stores the detId.
  * A vector<float> (v_gains) stores the value of the gain (more than one per detId).
  * Two vector<unsigned int> (v_ibegin and v_iend) store the correspondence of the v_detids
   * and the ranges of values in v_gain. 

https://github.com/cms-sw/cmssw/blob/master/CondFormats/SiStripObjects/interface/SiStripApvGain.h#L15.

Honestly I think that abusing `SiStripApvGain` for this purpose would be (a part from complicated) ugly as hell.

pieterdavid: ok, thanks for clarifying - shall I add a new record then? How about `SiStripApvSimulationParametersRcd`, with the `SiStripApvSimulationParameters` C++ type?

mmusich: fine by me.

EmyrClement: @pieterdavid another alternative to ```std::vector<std::vector<std::vector<TH1F>>>``` is ```std::vector<std::vector<std::vector<CLHEP::RandGeneral*>>>```, which would make the random sampling more straightforward.  This appears to work, but let me know if there is a preference for using  ```PhysicsTools::Calibration::HistogramF3D```

EmyrClement:
I updated the PR following the proposed changes in #27941

mmusich:
Can please someone test this together with  #27941?
As feature is inactive and there is not yet a payload in DB no changes are expected

fabiocos:
please test with cms-sw/cmssw#27941

civanch:
+1

civanch:
+1

fabiocos:
+1

PR #27665, opened 2019-08-01
Phase2-hcx214 Add a new scenario and workflow with latest Phase2 configuration including HFNose

bsunanda:
#### PR description:

Add a new scenario (2026D47) and its corresponding workflow 22434 with latest version of Phase2 geometry together with HFNose

#### PR validation:

Tested the workflow 22434.0

#### if this PR is a backport please specify the original PR:

No back porting is needed

PR #27824, opened 2019-08-22
Add Skewed Phase 2 Inner Tracker geometry (T16).

ghugo83:
**PR description:**

This PR adds Skewed Phase 2 Inner Tracker geometry (T16), and its associated workflows.
The skewed Inner Tracker description is accessible at: http://cms-tklayout.web.cern.ch/cms-tklayout/layouts-work/repository-git-dev/OT616_200_IT613/layoutpixel.html 

Several ladders are rotated in TBPX around CMX_Z axis ('skewed'). This is to allow space for cooling pipes, as it will be in the detector to be installed.
There is no update in terms of Material Budget here, focus is geometry-only.

Geometry scenario is: 2026D48.
Workflows are: 228xx (no PU), 230xx (PU).

Tracking performance should be compared with T15 (2026D46 scenario).

NB:
- Please merge PR #27665 first.
Present PR is created on top of 27665, to avoid having shifts in workflow numbering (no need to leave free spots for D47).
I (might) have very trivial conflict to solve once 27665 is merged into master.

- **IMPORTANT: This only introduces the skewed geometry description. There might be adjustments to do on CMSSW side to deal with the skewed ladders.** 
Got runs through the 4 steps without extra warning/error, which is already a good sign, but obviously does not mean at all everything is OK...
On CMSSW side, I checked that the sim **geometry construction** is OK.
First feedback from @migliore is that DIGI step should be fine, this should be looked at in details though.
Possible tunings needed on top of this PR!

**PR validation:**
Following was done for validation:
- Solved all overlaps on CMSSW with Fireworks and Geant4 tools.
- Checked sim **geometry construction** is OK on CMSSW side.
- Even if the active geometry is different from T15, logically there should be no DetId diff (based on how the DetIds are computed in CMSSW). Checked that this is the case indeed.
This allows to reuse the fakeConditions from T15.
- Checked no shift in workflows numbering.
- Checked that workflows with D48 run smoothly with no extra error / warning.

FYI: @emiglior @pwittich @bsunanda @fabiocos @kpedro88 @mmusich @jalimena @boudoul


Configuration/Geometry/README.md
```diff
@@ -37,5 +39,7 @@ Several detector combinations have been generated:
 * D44 = T14+C6+M3+I7+O2+F3
 * D45 = T15+C8+M3+I10+O3+F2
 * D46 = T15+C9+M3+I10+O3+F2
+* D47 = T15+C10+M3+I10+O3+F3
+* D48 = T16+C10+M3+I10+O3+F3
```
kpedro88: I don't think a tracker test workflow needs to include the HF nose. I would make either D45 or D46 the basis for D48.

ghugo83: Are C10 and F3 not the latest descriptions available?
In that case, why not having the latest Tracker description with them, and compare D48 versus D47?
It would avoid having latest subdetectorA with old subdetectorB and latest subdetectorB with old subdetectorA. Which would imply that when one wants to access the latest CMS geometry, it does not exist?

kpedro88: C10 and F3 are the latest version of the HF nose. This is a detector R&D effort that is not yet an approved part of the Phase 2 upgrade project. Combining tracker tests with this will just lead to confusion and inconsistencies.

ghugo83: Ha I didn't know C10 and F3 were purely in an hypothetical phase for now.
Ok I will simply revert my last commit then.

kpedro88:
please test workflow 22834.0

cvuosalo:
+1

mmusich:
@kpedro88 as stated in the PR description:
> Even if the active geometry is different from T15, logically there should be no DetId diff (based on how the DetIds are computed in CMSSW). Checked that this is the case indeed.
This allows to reuse the fakeConditions from T15.

This means that using `auto:phase2_realistic_T15` is OK.

Configuration/PyReleaseValidation/python/upgradeWorkflowComponents.py
```diff
@@ -368,6 +370,13 @@
         'Era' : 'Phase2C10_timing_layer_bar',
         'ScenToRun' : ['GenSimHLBeamSpotFull','DigiFullTrigger','RecoFullGlobal', 'HARVESTFullGlobal'],
     },
+    '2026D48' : {
+        'Geom' : 'Extended2026D48',
+        'HLTmenu': '@fake2',
+        'GT' : 'auto:phase2_realistic_T15',
```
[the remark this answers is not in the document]
mmusich: @kpedro88 there is no need for it at the moment. As the list of `DetId`s between `T15` and `T16` is the same, this choice is technically sound.
I cannot exclude a later update in case one discovers there is a better choice set of parameters but this will be checked during the validation procedure.

kpedro88:
@prebello, @chayanit, @pgunnell, @zhenhu please sign

chayanit:
+1

kpedro88:
+1

PR #27827, opened 2019-08-22
Fix gcc9 warnings L1Trigger/GlobalCaloTrigger

mrodozov:
#### PR description:

Fix warnings in L1Trigger/GlobalCaloTrigger

#### PR validation:

Builds without warnings. Added [[fallthrough]] for every "missing" break.

### DIFF
--- a/L1Trigger/GlobalCaloTrigger/test/testElectronSorter.cpp
+++ b/L1Trigger/GlobalCaloTrigger/test/testElectronSorter.cpp
@@ -305,6 +305,7 @@ void convertToGct(EmInputCandVec candidates) {
         } else {
           eta = 5;
         }
+        [[fallthrough]];
       case 3:
         phi = 0;
         if (region == 0) {
@@ -320,6 +321,7 @@ void convertToGct(EmInputCandVec candidates) {
         } else {
           eta = 3;
         }
+        [[fallthrough]];
       case 5:
         phi = 0;
         if (region == 0) {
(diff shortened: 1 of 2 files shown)


mrodozov:
please test

L1Trigger/GlobalCaloTrigger/test/testElectronSorter.cpp
```diff
@@ -305,6 +305,7 @@ void convertToGct(EmInputCandVec candidates) {
         } else {
           eta = 5;
         }
+        [[fallthrough]];
```
fabiocos: @mrodozov why ```[[fallthrough]]``` instead of ```break``` as in all the other cases? I would naively imagine it was just forgotten, looking at the code.
@rekovic could you please check?

L1Trigger/GlobalCaloTrigger/test/testElectronSorter.cpp
```diff
@@ -320,6 +321,7 @@ void convertToGct(EmInputCandVec candidates) {
         } else {
           eta = 3;
         }
+        [[fallthrough]];
```
fabiocos: same here...

mrodozov: Because with the current state the logic is to fallthrough and I was adding it where it was falling anyway. I was assuming breaks were forgotten, too, and then I've been told not to assume since the fallthrough were intentional. Some of them weren't but most were. Where they are forgotten I've added them like in here:
https://github.com/cms-sw/cmssw/pull/27856
but in other places I've been told otherwise:
https://github.com/cms-sw/cmssw/pull/27811#issuecomment-523469488
Solution was to keep the existing logic until told to change it to break by responsible's.

fabiocos: @mrodozov thanks, I see the reasoning behind, but I would like nevertheless to ask @rekovic to check and confirm that this is the intended behaviour of the code

fabiocos:
+1

fabiocos:
merge

PR #27833, opened 2019-08-23
HGCal and jets validation plots harvesting

clelange:
#### PR description:

This PR

- cleans up some commented out code in `DQMOffline/PFTau`
- improves the code style in `DQMOffline/PFTau`
- adds `hgcalPFJetValidation`, i.e. some jet-level validation plots for HGCal
- adds a harvesting script (`Validation/HGCalValidation/scripts/harvestHGCalValidationPlots.py`) for 

#### PR validation:

I produced a couple of Pytha8PtGun gluon samples and the plots came out as expected.

CC @rovere

### DIFF
--- a/DQMOffline/PFTau/interface/Benchmark.h
+++ b/DQMOffline/PFTau/interface/Benchmark.h
@@ -51,23 +51,23 @@ public:
     return (pt > ptMin_ && pt < ptMax_ && eta > etaMin_ && eta < etaMax_ && phi > phiMin_ && phi < phiMax_);
   }
 
-  virtual void setDirectory(TDirectory* dir);
+  virtual void setDirectory(TDirectory *dir);
 
   /// write to the TFile, in plain ROOT mode. No need to call this function in DQM mode
   void write();
 
 protected:
   /// book a 1D histogram, either with DQM or plain root depending if DQM_ has been initialized in a child analyzer or not.
-  //TH1F* book1D(const char* histname, const char* title,
+  //TH1F *book1D(const char *histname, const char *title,
   /// book a 1D histogram, either through IBooker or plain root
-  TH1F* book1D(DQMStore::IBooker& b, const char* histname, const char* title, int nbins, float xmin, float xmax);
+  TH1F *book1D(DQMStore::IBooker &b, const char *histname, const char *title, int nbins, float xmin, float xmax);
 
   /// book a 2D histogram, either with DQM or plain root depending if DQM_ has been initialized in a child analyzer or not.
-  //TH2F* book2D(const char* histname, const char* title,
+  //TH2F *book2D(const char *histname, const char *title,
   /// book a 2D histogram, either through IBooker or plain root
-  TH2F* book2D(DQMStore::IBooker& b,
-               const char* histname,
-               const char* title,
+  TH2F *book2D(DQMStore::IBooker &b,
+               const char *histname,
+               const char *title,
                int nbinsx,
                float xmin,
                float xmax,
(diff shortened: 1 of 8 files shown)


clelange:
FYI @apsallid

civanch:
please test

jfernan2:
@clelange are you now working as HGCAL DQM Validation developer? If so, please add your self to the corresponding e-group in:
https://twiki.cern.ch/twiki/bin/view/CMS/DQMContacts#HGCAL

jfernan2:
@clelange the PR is producing changes across several workflows, adding and modifying histograms. Can you please check in the following link if the behaviour is as expected/desired?
https://cmssdt.cern.ch/SDT/jenkins-artifacts/baseLineComparisons/CMSSW_11_0_X_2019-08-23-1100+5b3c7b/33194/dqm-histo-comparison-summary.html

civanch:
+1

clelange:
> @clelange are you now working as HGCAL DQM Validation developer? If so, please add yourself to the corresponding e-group in:
> https://twiki.cern.ch/twiki/bin/view/CMS/DQMContacts#HGCAL

@jfernan2 No, @apsallid is still the HGCal Validation contact, see https://indico.cern.ch/event/795608/sessions/302178/attachments/1809338/2954349/HGCAL_week_1_intro.pdf#page=4

Regarding the changes: I had a look at the histograms, the changes are tiny and basically only visible for packedGenParticlesValidation and slimmedElectronValidation. I think this is OK, but since the first commit of this PR that introduces the differences is based on work by @rovere I hope he or @apsallid can confirm that this is fine.

rovere:
@jfernan2 These are my observations after carefully checking the page with the comparison between before/after this PR:

* 4.53 has no visible differences (side-by-side plots have identical stat-box entries). The tool reports 8.
* 5.1 has no visible differences (plots are empty). The tool reports 4
* 9.0 same comments as 4.53. The tool reports 16.
* 25.0 same comments as 4.53. The tool reports 16.
* 135.4 same comments as 4.53. the tool reports 4.

At this point, I clearly got bored and did not check any other comparison.
I only verified that the **added** plots are indeed what they are supposed to be.

I have a few suggestions/comments:

* the comparison tool and its display integrated into the PR-bot is a nice addition...
* ..yet if the tool has too many false positive, the general attitude could become to not trust the report itself, which would be a pity.
* Can you please comment on how the comparison is made?
* Remove anything related to the TimerService, since that's non-reproducible almost by definition
* Add to the report (i.e. in the comparison file uploaded to the DQMGUI) the histogram with, e.g., the difference between Base_vs_PR, so that the user can immediately spot where the problems are.
* Exclude empty histograms if they both are.

fioriNTU:
@rovere @jfernan2 , I also had a quick look to the comparison. The plots added by HGCAL are produced in only four workflows, and I guess they are ok https://tinyurl.com/y6s6dlp2

The rest of the differences is due to a change of bin number is some histograms. As an example this histogram https://tinyurl.com/y3xdlrne in the base has 59 bins in the x-axis in the PR has 60. Even if it is empty a difference is found since the comparison is done bin by bin. I guess we can "+1" here

jfernan2:
@rovere thanks for your comments and feedback.
As @fioriNTU pointed out, this PR introduced changes in the number of bins for some histograms so the comparison tool is reflecting so in the rightmost link: same stats box and same entries with the new binning. The leftmost link has the original histogram which is changing and the cente one the new histogram which has changed. Empty histograms are discarded unless they change somehow. In a nutshell: it is a full bin-by-bin comparison withh all the consequences.
We believe this kind of comparison gives the full picture. Of course it can be tuned further: removing anything related to TimerService is nasty and we have tried in the past, for the moment is left out so that it can prove that the tool is working in PRs with no real changes at all.

jfernan2:
+1
DQM Changes seem to be those expected for this PR

fabiocos:
+1

schneiml:
@rovere A quick check on the reported difference in `ParticleFlow/packedGenParticlesValidation/CompWithGenParticles/delta_et_VS_et_` in wf 9.0 shows that it is not a false positive, that plot grew one more row of bins in this PR. 

Changes in binning are potentially quite significant changes, and the tool is perfectly correct to report this change. We probably would not have noticed otherwise.

rovere:
Ciao Marcel,
I understand. I was suggesting to present it in a cleaner way since a
change in empty or non-used bins is not easy to spot by eye.


Ciao and thanks,
-- Marco.

___________
Marco Rovere
Marco.Rovere@cern.ch
CERN EP-CMG | room 40 3-A28 | tel +41227671209 (71209)

PR #22761, opened 2018-03-27
CommonTools-Statistics: Fix bug found from clang warning -Wunsequenced

gartung:
The value of j is undefined because j might be incremented before access.

In file included from /build/cmsbld/jenkins/workspace/build-any-ib/w/tmp/BUILDROOT/c6796210c53b02f1870cdf07f5d562c4/opt/cmssw/slc6_amd64_gcc630/cms/cmssw/CMSSW_10_1_CLANG_X_2018-03-26-2300/src/CommonTools/Statistics/src/SequentialCombinationGenerator.cc:1:
  /build/cmsbld/jenkins/workspace/build-any-ib/w/tmp/BUILDROOT/c6796210c53b02f1870cdf07f5d562c4/opt/cmssw/slc6_amd64_gcc630/cms/cmssw/CMSSW_10_1_CLANG_X_2018-03-26-2300/src/CommonTools/Statistics/interface/SequentialCombinationGenerator.h:105:52: warning: unsequenced modification and access to 'j' [-Wunsequenced]
       for (int j=0;j<p[i];cnew[ip+j]=cnew2[j]=sss[j++]);
                                  ~                ^

PR #27843, opened 2019-08-23
Fix gcc9 warnings in CommonTools

mrodozov:
#### PR description:
Fixes gcc9 warning
#### PR validation:
Builds without warning


mrodozov:
please test

perrotta:
I suspect this PR is revealing some logic error in the code.
Case 2 is fully equivalent to

```C++
  case 2:
      int i = std::min(pmax, n - 1); 
      if (i >= pmin) {
        done = (i >= n - i && n - i >= pmin);
        if (done) {
          p[0] = i;
          p[1] = n - i;
        }
        return done;
      }
      [[fallthrough]];
```

First of all, this shouldn't fall through, I guess, or at least the ```return done;``` statement should be brought out of the for loop (either if block)..
Secondly, I suspect there is some error in the logic (given the simpler equivalent implementation that doesn't need a loop).

On the other hand, this SequentialPartitionGenerator code is only called by SequentialCombinationGenerator in CMSSW, and that SequentialCombinationGenerator is never accessed elsewhere and it was never touched since moving to github in 2013 (I did not look at the history before it). If so, I'd say it can be safer to just remove this dead wood instead of "fixing" to something which seems a bugged behaviour anyhow.

By the way: who is maintaining this ```CommonTools/Statistics``` package in CMSSW? Isn't it wholly obsolete?

perrotta:
@santocch @fabiocos : these gc99 warnings are quite often pointing to some real issue in the code.
Do you know who is responsible of that ```CommonTools/Statistics``` package, and can tell us whether  the SequentialPartitionGenerator has to be fixed, removed, or what?

santocch:
I dug a little through the old CVS repository. Here is what I found:

- Here is the last [CVS status](http://cmscvs.web.cern.ch/cmscvs/cgi/viewvc.cgi/cvsroot/CMSSW/CommonTools/Statistics/interface/) of the CommonTools/Statistics package... it looks like the whole package was imported from ORCA in 2006 by Thomas Speer (not anymore in CMS)
- After the first import in CMSSW nothing really important changed... minor code update and few bug fixes... people involved were (from the developer list in .admin)


>Developers
--
27 | walten : Wolfgang Waltenberger : Wolfgang.Waltenberger@cern.ch
28 | speer : Thomas Speer : speer@cern.ch
29 | vanlaer : Pascal Vanlaer : pvanlaer@ulb.ac.be
30 | rpw : Rick Wilkinson : Rick.Wilkinson@cern.ch
31 |  
32 | >Administrators
33 | speer : Thomas Speer : speer@cern.ch
34 | walten : Wolfgang Waltenberger : Wolfgang.Waltenberger@cern.ch
35 | vanlaer : Pascal Vanlaer : pvanlaer@ulb.ac.be

Practically none of them looks still involved in CMSSW stuff (I guess the only one still in CMS is Pascal Vanlaer). 

My conclusion is I agree with Perrotta... let's remove it.

perrotta:
Thank you @santocch 

Some further check:
- Even in private code in github the only references for these modules are from people who linked the whole CommonTools/Statistics  package, but there is no hint of any usage of either [SequentialPartitionGenerator](https://github.com/search?l=C%2B%2B&p=1&q=SequentialPartitionGenerator+language%3AC%2B%2B&type=Code) or [SequentialCombinationGenerator](https://github.com/search?l=C%2B%2B&q=SequentialCombinationGenerator+language%3AC%2B%2B&type=Code)
- The only actual usage of a SequentialPartitionGenerator is from a user who fully [rewrites it](https://github.com/search?l=Python&q=SequentialPartitionGenerator+language%3AC%2B%2B&type=Code) by himself in python, therefore not linking these modules in CMSSW
- One and half year ago @gartung with PR #22761 already fixed a bug in SequentialCombinationGenerator.h: that bug was discovered after a clang warning, it was supposed to affect the results but it remained unnoticed both before and after the fix was merged in CMSSW_10_2_X and following cycles, thus suggesting that nobody was using it already at that time.

Therefore, also in order to avoid having to maintain it further in the future, I would suggest to ```git rm```:
- CommonTools/Statistics/interface/SequentialCombinationGenerator.h
- CommonTools/Statistics/interface/SequentialPartitionGenerator.h
- CommonTools/Statistics/src/SequentialCombinationGenerator.cc
- CommonTools/Statistics/src/SequentialPartitionGenerator.cc

@mrodozov can you please take care of it by modifying this PR (and its title)? Or do you prefer to just close this one and either me or @santocch will submit another PR with the removal?

fabiocos:
@perrotta @santocch thanks for the detailed check about these classes. I agree that if nobody is using them since a while, it does not make much sense to keep unsupported and unused code, better to just remove it. @mrodozov could you please update this PR according to the conclusion of this discussion?

mrodozov:
please test

perrotta:
+1
- Removal of te unused classes SequentialPartitionGenerator and SequentialCombinationGenerator seems well justified

fabiocos:
@santocch this is along your comments, do you agree? In case please sign it

santocch:
+1

fabiocos:
+1

PR #27844, opened 2019-08-23
Fix gcc 9 warnings in DataFormats

mrodozov:
#### PR description:

Fixes warnings in DataFormats gcc9 IB

#### PR validation:

Builds without warnings. Some warnings were from other packages (PR still open)

### DIFF
--- a/DataFormats/EgammaCandidates/src/Photon.cc
+++ b/DataFormats/EgammaCandidates/src/Photon.cc
@@ -90,6 +90,7 @@ void Photon::setCorrectedEnergy(P4type type, float newEnergy, float delta_e, boo
     case regression1:
       eCorrections_.regression1Energy = newEnergy;
       eCorrections_.regression1EnergyError = delta_e;
+      [[fallthrough]];
     case regression2:
       eCorrections_.regression2Energy = newEnergy;
       eCorrections_.regression2EnergyError = delta_e;
@@ -150,6 +151,7 @@ void Photon::setP4(P4type type, const LorentzVector& p4, float error, bool setTo
     case regression1:
       eCorrections_.regression1P4 = p4;
       eCorrections_.regression1EnergyError = error;
+      [[fallthrough]];
     case regression2:
       eCorrections_.regression2P4 = p4;
       eCorrections_.regression2EnergyError = error;
--- a/DataFormats/EgammaCandidates/src/Photon.cc
+++ b/DataFormats/EgammaCandidates/src/Photon.cc
@@ -90,7 +90,7 @@ void Photon::setCorrectedEnergy(P4type type, float newEnergy, float delta_e, boo
     case regression1:
       eCorrections_.regression1Energy = newEnergy;
       eCorrections_.regression1EnergyError = delta_e;
-      [[fallthrough]];
+      break;
     case regression2:
       eCorrections_.regression2Energy = newEnergy;
       eCorrections_.regression2EnergyError = delta_e;
@@ -151,7 +151,7 @@ void Photon::setP4(P4type type, const LorentzVector& p4, float error, bool setTo
     case regression1:
       eCorrections_.regression1P4 = p4;
       eCorrections_.regression1EnergyError = error;
-      [[fallthrough]];
+      break;
     case regression2:
       eCorrections_.regression2P4 = p4;
       eCorrections_.regression2EnergyError = error;
--- a/DataFormats/EgammaCandidates/src/Photon.cc
+++ b/DataFormats/EgammaCandidates/src/Photon.cc
@@ -90,7 +90,7 @@ void Photon::setCorrectedEnergy(P4type type, float newEnergy, float delta_e, boo
     case regression1:
       eCorrections_.regression1Energy = newEnergy;
       eCorrections_.regression1EnergyError = delta_e;
-      break;
+      [[fallthrough]];
     case regression2:
       eCorrections_.regression2Energy = newEnergy;
       eCorrections_.regression2EnergyError = delta_e;
@@ -151,7 +151,7 @@ void Photon::setP4(P4type type, const LorentzVector& p4, float error, bool setTo
     case regression1:
       eCorrections_.regression1P4 = p4;
       eCorrections_.regression1EnergyError = error;
-      break;
+      [[fallthrough]];
     case regression2:
       eCorrections_.regression2P4 = p4;
       eCorrections_.regression2EnergyError = error;
--- a/DataFormats/HcalCalibObjects/src/HFRecalibration.cc
+++ b/DataFormats/HcalCalibObjects/src/HFRecalibration.cc
@@ -35,7 +35,7 @@ double HFRecalibration::getCorr(int ieta, int depth, double lumi) {
   switch (depth) {
     case 1:
       reCalFactor = (1 + HFParsAB[0][0][ieta] * sqrt(lumi) + HFParsAB[0][1][ieta] * lumi);
-      [[fallthrough]];
+      break;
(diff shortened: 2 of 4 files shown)


mrodozov:
please test

DataFormats/HcalCalibObjects/src/HFRecalibration.cc
```diff
@@ -35,7 +35,7 @@ double HFRecalibration::getCorr(int ieta, int depth, double lumi) {
   switch (depth) {
     case 1:
       reCalFactor = (1 + HFParsAB[0][0][ieta] * sqrt(lumi) + HFParsAB[0][1][ieta] * lumi);
-
+      [[fallthrough]];
```
perrotta: I suspect this wasn't originally supposed to fall through: @abdoulline @igv4321 ?
(maybe a few others in this PR...)

DataFormats/TrackerCommon/src/TrackerTopology.cc
```diff
@@ -396,6 +396,7 @@ SiStripModuleGeometry TrackerTopology::moduleGeometry(const DetId &id) const {
         case 3:
           return SiStripModuleGeometry::W3A;
       }
+      [[fallthrough]];
```
mmusich: I think this should never fall through in real life, though, I would rather break here, after returning an `SiStripModuleGeometry::UNKNOWNGEOMETRY;`

DataFormats/EgammaCandidates/src/Photon.cc
```diff
@@ -90,6 +90,7 @@ void Photon::setCorrectedEnergy(P4type type, float newEnergy, float delta_e, boo
     case regression1:
       eCorrections_.regression1Energy = newEnergy;
       eCorrections_.regression1EnergyError = delta_e;
+      [[fallthrough]];
```
perrotta: Shouldn't it rather break here? @Sam-Harper

DataFormats/EgammaCandidates/src/Photon.cc
```diff
@@ -150,6 +151,7 @@ void Photon::setP4(P4type type, const LorentzVector& p4, float error, bool setTo
     case regression1:
       eCorrections_.regression1P4 = p4;
       eCorrections_.regression1EnergyError = error;
+      [[fallthrough]];
```
perrotta: Shouldn't it rather break here? @Sam-Harper

perrotta:
Once could profit and further remove unneeded ```break``` statements in ```Photon::getCorrectedEnergy``` and ```Photon::getCorrectedEnergyError```

DataFormats/HcalCalibObjects/src/HFRecalibration.cc
abdoulline: @perrotta 
I suppose there had to be "break" originally (apparently forgotten by author), as HF depths are treated separately (method parameter list has "depth" parameter). 
So fall though  doesn't seem to be natural here...

DataFormats/EgammaCandidates/src/Photon.cc
Sam-Harper: yes, this is bug, thanks for pointing this out. Luckily I dont think we use regression1 and its regression2 that is our main correction.  Likewise for comment below

DataFormats/EgammaCandidates/src/Photon.cc
Sam-Harper: yes, see above comment, thanks for spotting!

perrotta:
@mrodozov , all updates have been addressed separately, and it looks like that the correct fix is here in all cases to replace the fallthrough with an apparently forgotten ```break``` statement.
Will you take care of updating this PR accordingly? Or should I do it?

mrodozov:
please test
I understood all four should got to breaks, but not the Photon::* methods reference :/

perrotta:
Thank you @mrodozov, that's correct

The extra update that I suggested (admittently only for the aesthetics) was to remove the ```break;``` statements right after all ```return```'s in both ```Photon::getCorrectedEnergy()``` and ```Photon::getCorrectedEnergyError()``` methods; or alternatively add a ```break;``` also after https://github.com/cms-sw/cmssw/blob/4e417b1ec1d15cda8a076843fbb74910779bd568/DataFormats/EgammaCandidates/src/Photon.cc#L113, just for symmetry,

I won't insist on it, though

perrotta:
@Sam-Harper 
Comparisons show that now all regression2 energy corrections (and errors) are 0, see e.g. RunSinglePh2016B wf136p731:
image

Looking closer at the code, e.g. the two if-blocks starting in https://cmssdt.cern.ch/lxr/source/RecoEgamma/EgammaPhotonAlgos/src/PhotonEnergyCorrector.cc#0141, it seems now to me that probably it was done intentionally to set regression1 and regression2 at the same time. This would be rather convoluted (and error prone), but if so then:
- it was intentional to fall-through at lines L93 and L154 of Photon.cc
- it was not a bug, and therefore the original ```[[fallthrough]];``` directive should be re-stated to replace the ```break;``` statement that I suggested instead as a possible bug fix.

Could you please check and confirm?
If so, I'd ask @mrodozov to revert the changes at  L93 and L154 of Photon.cc: thank you for the patience!

perrotta:
@mrodozov , having checked the code further I'd rather bypass the confirmation of @Sam-Harper (which is however still welcome), and restore your original ```[[fallthrough]]``` directive instead of ```break``` in
- https://github.com/cms-sw/cmssw/pull/27844/files#diff-f552305352bbeeeb4d755958752a43f1R93
- https://github.com/cms-sw/cmssw/pull/27844/files#diff-f552305352bbeeeb4d755958752a43f1R154

At the same time, I'd suggest EGamma people to update their code in a more intuitive way so that you don't have to set ```regression1``` if you actually want to set ```regression2```

Sam-Harper:
Urgh. Yes the photon code is awful and should be burnt. I had wanted to do so but the UL took over sadly.  Sorry for the mix up.

perrotta:
@mrodozov (just in case you missed it, as I see that you modified other similar PR and not this one):
once you revert L93 and L154 of DataFormats/EgammaCandidates/src/Photon.cc to ```[[fallthrough]]```, as it was in your oiginal submission, this PR can get signed by reco

mrodozov:
please test
I missed it indeed

perrotta:
@mrodozov , sorry for noticing only now while going to sign this PR, but the real fix for DataFormats/TrackerCommon/src/TrackerTopology.cc would be replacing the original ```[[fallthrough]];``` (now ```break:```) with:
```C++
  return SiStripModuleGeometry::UNKNOWNGEOMETRY;
```
as also suggested by @mmusich in https://github.com/cms-sw/cmssw/pull/27844#discussion_r317208245
Otherwise, in case ```tidRing(id)``` is not included between 1 and 3 (which should never happen, anyhow) the method doesn't return

mrodozov:
please test

perrotta:
code-checks

perrotta:
+1
- [[fallthrough]] directive inserted in DataFormats/EgammaCandidates/src/Photon.cc, to replicate the identical behaviour as before
- fallthrough replaced by a break in DataFormats/HcalCalibObjects/src/HFRecalibration.cc (bug fix): this could affect HF aging scenarios in Phase2
- fallthrough replaced by a return of UNKNOWNGEOMETRY in DataFormats/TrackerCommon/src/TrackerTopology.cc (logical bug fix, but quite likely never hit): 

pohsun:
+1

fabiocos:
+1

PR #27849, opened 2019-08-23
python 3 fix for GlobalTag.py

Dr15Jones:
#### PR description:

python 3 does not allow comparison of a string to a None. This was occuring in the call to sorted. Replacing the None with '' allows the comparison to function and the final results are the
same.

#### PR validation:

The code was tested in the CMSSW_11_0_DEVEL_X IB which uses python 3 using the configuration generated for the 20661.97 workflow (which was failing in the IB). The resultant configuration for the PoolDBESSource is the same as for the original python 2 case.

### DIFF
--- a/Configuration/AlCa/python/GlobalTag.py
+++ b/Configuration/AlCa/python/GlobalTag.py
@@ -79,7 +79,7 @@ def GlobalTag(essource = None, globaltag = None, conditions = None):
                 for entry in autoKey[1:]:
                   entry = entry.split(',')
                   record     = entry[1]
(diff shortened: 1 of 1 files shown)


Dr15Jones:
please test

pohsun:
+1

mrodozov:
@Dr15Jones yes. I ran it to check the cmsdist PR with the code fixing the relvals. It worked as expected, and we built full IB with it to see the output of the new warning (see the cmsdist PR). 
The test is fine

PR #27850, opened 2019-08-23
Fix gcc9 warning in HeavyFlavorAnalysis/RecoDecay

mrodozov:
#### PR description:

Fixes a warning in HeavyFlavorAnalysis/RecoDecay

#### PR validation:

Builds without warning

### DIFF
--- a/HeavyFlavorAnalysis/RecoDecay/src/BPHPlusMinusCandidate.cc
+++ b/HeavyFlavorAnalysis/RecoDecay/src/BPHPlusMinusCandidate.cc
@@ -73,6 +73,7 @@ void BPHPlusMinusCandidate::add(
                                           << "already containing same sign particle, add rejected";
         return;
       }
+      [[fallthrough]];
     case 0:
       addK(name, daug, searchList, mass, sigma);
   }


mrodozov:
please test

HeavyFlavorAnalysis/RecoDecay/src/BPHPlusMinusCandidate.cc
```diff
@@ -73,6 +73,7 @@ void BPHPlusMinusCandidate::add(
                                           << "already containing same sign particle, add rejected";
         return;
       }
+      [[fallthrough]];
```
fabiocos: @santocch @mrodozov would not a break be more correct? The cases look mutually exclusive to me

fabiocos:
+1

fabiocos:
merge

HeavyFlavorAnalysis/RecoDecay/src/BPHPlusMinusCandidate.cc
santocch: - First call: no particles in the collection... so the routine just add the particle regardless of its charge...
   - Second call: check the charge (it has to be the opposite of the first added particle). If it's opposite then OK fallthrough --> add the particle. If it's the same charge then it's not OK and exit from the routine
   - Third call: too many particles --> EXIT

The routine's logic looks ok like it is now...

santocch:
+1

PR #27852, opened 2019-08-23
Fix gcc9 warning in MuonAnalysis/MuonAssociators

mrodozov:
#### PR description:

Fixes warning in MuonAnalysis/MuonAssociators

#### PR validation:

Builds without warning

### DIFF
--- a/MuonAnalysis/MuonAssociators/src/MatcherUsingTracksAlgorithm.cc
+++ b/MuonAnalysis/MuonAssociators/src/MatcherUsingTracksAlgorithm.cc
@@ -182,6 +182,7 @@ bool MatcherUsingTracksAlgorithm::match(const reco::Candidate &c1,
         }
       }
     }
+      [[fallthrough]];
     case ByPropagatingSrc: {
       FreeTrajectoryState start = startingState(c1, whichTrack1_, whichState1_);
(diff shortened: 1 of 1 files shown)


mrodozov:
please test

MuonAnalysis/MuonAssociators/src/MatcherUsingTracksAlgorithm.cc
```diff
@@ -182,6 +182,7 @@ bool MatcherUsingTracksAlgorithm::match(const reco::Candidate &c1,
         }
       }
     }
+      [[fallthrough]];
```
fabiocos: @santocch @mrodozov a ```break``` would get the same effect here I would say, anyway it does not seem incorrect to me

fabiocos:
+1

fabiocos:
merge

MuonAnalysis/MuonAssociators/src/MatcherUsingTracksAlgorithm.cc
santocch: I agree. The logic would be identical... I'm not sure which one would be better or what difference would a break introduce... probably none so let's keep this one.

santocch:
+1

PR #27870, opened 2019-08-26
Fix 5 warnings in PhysicsTools/MVATrainer

mrodozov:
#### PR description:
Fixes gcc9 warnings

#### PR validation:

Builds without warnings.

### DIFF
--- a/PhysicsTools/MVATrainer/src/LeastSquares.cc
+++ b/PhysicsTools/MVATrainer/src/LeastSquares.cc
@@ -291,6 +291,7 @@ namespace PhysicsTools {
             break;
           } else
             pos = (Position)(pos + 1);
+          [[fallthrough]];
         case POS_WEIGHTS:
           loadVector(elem, n + 1, weights);
           break;
--- a/PhysicsTools/MVATrainer/src/MVATrainer.cc
+++ b/PhysicsTools/MVATrainer/src/MVATrainer.cc
@@ -521,8 +521,8 @@ namespace PhysicsTools {
 
     static struct NameExpect {
       const char *tag;
-      bool mandatory;
-      DOMElement **elem;
+      bool mandatory = false;
+      DOMElement **elem = nullptr;
     } const expect[] = {{"input", true, &xmlInput},
                         {"config", true, &xmlConfig},
                         {"output", true, &xmlOutput},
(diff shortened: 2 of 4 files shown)


mrodozov:
please test

guitargeek:
H! Don't you think it would be better to just delete that package? I did some background research on that package some time ago, and it's even older than TMVA. IMHO there is not point in supporting that code and spending time fixing bugs here. Everyone is now either using TMVA or the Python data science ecosystem to train their MVAs.

The only remaining use of that package is within the TopQuarkAnalysis subsystem, where a few tweaks are needed to remove the dependence [1]. It's probably okay to remove the bits of code there that depend on the MVATrainer, as I doubt the TopQuarkAnalysis subsystem is used (seems to date back even to Tevatron times).

What do you think?

[1] https://github.com/cms-sw/cmssw/compare/master...guitargeek:MVATrainer#diff-e90179e3c3ae80c89460b684461cc835

mrodozov:
would you confirm that's the difference 
https://github.com/cms-sw/cmssw/compare/master...guitargeek:MVATrainer ?
because the other link doesn't show any difference (github has its issues) 
Then I can rework the PR

guitargeek:
Yes you are right. Thanks, I updated the link in my post.

Alright then, I will do a PR with the removal of the package and refer to your gcc warnings PR to exemplify how leaving the MVATrainer in the release can be annoying because it raises new warnings for every major gcc version bump. Can you please leave this PR open until the release managers decided if they want to remove the package or not? Do you agree with this procedure?

mrodozov:
> Can you please leave this PR open until the release managers decided if they want to remove the package or not? Do you agree with this procedure?

Yes it fine with me.

PhysicsTools/MVATrainer/src/LeastSquares.cc
```diff
@@ -291,6 +291,7 @@ namespace PhysicsTools {
             break;
           } else
             pos = (Position)(pos + 1);
+          [[fallthrough]];
```
fabiocos: @santocch I do not see why this should not be a ```break```, could you please cross check?

fabiocos:
hold

this PR becomes obsolete if #27888 is integrated

PR #27884, opened 2019-08-27
Update DTDigi so that it can be accomodate Phase2 TDC units.

namapane:
#### PR description:
Update DTDigi to be able to accomodate the different TDC unit of Phase2 readout (in addition to the current unit, which remains the default).

DTDigiSimLink updated accordingly. Also added an option to create digis in the new units in the simulation (switched off by default).

There is NO change in the default behaviour of any class.

This update was requested by DT DPG (@fcavallo @battibass ). For more details:
https://indico.cern.ch/event/839945/#sc-4-10-dt-digi-format-for-tdc

#### PR validation:
Tested comparing comparing of raw2digi before and after the change with both phase-1 and phase-2 digis.

Note: the correspondig PR for 11_X is #27890.

### DIFF
--- a/DataFormats/DTDigi/interface/DTDigi.h
+++ b/DataFormats/DTDigi/interface/DTDigi.h
@@ -4,8 +4,8 @@
 /** \class DTDigi
  *
  * Digi for Drift Tubes.
- * It can be initialized/set with a time in ns or a TDC count in 25/32 ns 
- * units.
+ * It can be initialized/set with a time in ns or a TDC count in 
+ * the specified base (ie number of counts/BX)
  *  
  *
  * \author N. Amapane, G. Cerminara, M. Pelliccioni - INFN Torino
@@ -21,12 +21,12 @@ public:
   
   /// Construct from the wire#, the TDC counts and the digi number.
   /// number should identify uniquely multiple digis in the same cell.
-  explicit DTDigi (int wire, int nTDC, int number=0);
-
+  explicit DTDigi (int wire, int nTDC, int number=0, int base=32);
+  
   // Construct from the wire#, the time (ns) and the digi number.
-  // time is converted in TDC counts (1 TDC = 25./32. ns)
+  // time is converted in TDC counts (1 TDC = 25./base. ns)
   // number should identify uniquely multiple digis in the same cell.
-  explicit DTDigi (int wire, double tdrift, int number=0);
+  explicit DTDigi (int wire, double tdrift, int number=0, int base=32);
 
   // Construct from channel and counts.
   //  explicit DTDigi (ChannelType channel, int nTDC);
@@ -50,13 +50,13 @@ public:
   double time() const;
 
   /// Get raw TDC count
-  uint32_t countsTDC() const;
+  int32_t countsTDC() const;
 
-  /// Set with a time in ns
-  void setTime(double time);  
+  /// Get the TDC unit value in ns
+  double tdcUnit() const;
 
-  /// Set with a TDC count
-  void setCountsTDC (int nTDC);
+  /// Get the TDC base (counts per BX)
+  int tdcBase() const;
 
   /// Print content of digi
   void print() const;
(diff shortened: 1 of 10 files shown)


civanch:
@namapane, it is necessary making 11_X PR first and this should be a backport to legacy.

boudoul:
Hi @namapane , is this something needed for the next MWGR ?

fcavallo:
Hi. it would be very nice indeed to have it for MWGR4. We are reading out data from the DT slice test and this data format (has the description says) has just the purpose to handle the reading of new TDC.

namapane:
I submitted #27890 for 11_X with exactly the same modifications on top of 11_X code. I guess this PR should now be closed and a new one made after #27890 is approved?

davidlange6:
this presumably breaks reading of any DTDigi stored in a root file. is that correct? Which data tiers does that affect?

namapane:
None: digis are NOT written in any tier (only the raw data re, both for real data and sim).

slava77:
On 8/28/19 7:19 AM, Nicola Amapane wrote:
> None: digis are NOT written in any tier (only the raw data re, both for
> real data and sim).
> 

what about sim digis in SIM  and digis in general in FEVT ?

namapane:
My understanding is that digis are never stored anyhwere. For sim we run digi2raw and store the raw instead. We do not store digis in FEVT since we have the raw anyhow. I'm pretty sure about this and would be surprised otherwise, but you can easily prove me wrong by pointing to any root file which actually contains a persistent DTDigi collection...

fcavallo:
I confirm: FEVT only contain raw data and reco collections for segments and rechits
(however I understood that this PR should be overwritten by 27890, based on 11_X..?)

namapane:
Thanks Francesca. 
Indeed I would suggest to move any discussion to #27890.

civanch:
please test

civanch:
+1

fabiocos:
@namapane @fcavallo look at FEVTDEBUGHLT, used in many RelVals (just run step2 of 10824.0):

```
14:23 cmsdev25 537> edmDumpEventContent step2.root | grep DTDigi
MuonDigiCollection<DTLayerId,DTDigi>    "hltMuonDTDigis"            ""                "HLT"     
MuonDigiCollection<DTLayerId,DTDigi>    "simMuonDTDigis"            ""                "HLT"     
MuonDigiCollection<DTLayerId,DTDigiSimLink>    "simMuonDTDigis"            ""                "HLT"     
vector<DTuROSFEDData>                 "hltMuonDTDigis"            ""                "HLT"     
```

The first 3 entries are affected by this PR in my understanding.

Looking at the output of step3 (RECOSIM):

```
14:25 cmsdev25 540> edmDumpEventContent step3.root | grep DTDigi
MuonDigiCollection<DTLayerId,DTDigiSimLink>    "simMuonDTDigis"            ""                "HLT"     

```

So my conclusion is that this PR modifies the event content wrt what we have for UL so far, at least in some data tiers. Do you agree?

namapane:
The latter are not DTDigis but DTDigiSimLink. What is the piurpose of keeping the links without the actual digis, maybe to enable an indirect association of rechits to simtracks? Does anybody know?
Regarding FEVTDEBUGHLT, is that used only for relvals? The first three collections are indeed affected; I was not aware of this special data tier. I suppose we may want to be able to read relvals across different (pre)releases in 10_6_X, but this looks like a quite special use case (I actually wonder what is the use case for storing digis in this tier since they are not needed anywhere else). 
If this rather special use case is a showstopper we can keep the change for 11X and skip the backport to 10_6_X, but that would be a pity since that is the release where studies for DT phase2 readout are taking place.

fabiocos:
@namapane 10_6_X is not simply the release for phase2 DT studies but it is first of all the release for Run2 UL, and breaking the format there looks an undesired feature to me.

namapane:
That's reasonable of course. I guess we could figure out workarounds for keeping SimLinks backward compatible, but I think that would be too much of a mess, so I propose DT phase2 studies to merge this update in a private recipe until 11_X, and skip integration into 10_6_X. @facvallo @battibass I hope that is fine for you. If so I can close the PR.

fcavallo:
Keeping the new DTDigi format within a private version of the code was indeed the result of a common agreement on this question. Thank you @namapane   for your valuable contribution and @fabiocos for the careful checks. Yes please: close this PR.

PR #26816, opened 2019-05-17
Remove outdated tau code

steggema:
#### PR description:

This removes a lot of outdated and unused tau code, in particular the full package RecoTauTag/TauTagTools and the CaloTau reconstruction. The goal is do make tau development easier and to reduce the maintenance burden. No changes are expected in any of the workflows.

#### PR validation:

Matrix tests were run locally as well as the tau reconstruction sequence itself to validate that there are no unwanted effects. The changes were reviewed by the RECO contacts @mbluj @swozniewski, also to confirm that there are no customers outside of CMSSW to the best of our knowledge.

@rmanzoni @roger-wolf

PR #27888, opened 2019-08-28
Removing the PhysicsTools/MVATrainer package

guitargeek:
#### PR description:

This PR aims to provide an alternative to the recently opened PR https://github.com/cms-sw/cmssw/pull/27870.

After seeing https://github.com/cms-sw/cmssw/pull/27870 it occurred to me that it might be reasonable to remove the PhysicsTools/MVATrainer package so nobody has to spend time supporting and fixing compiler warnings in this code in the future.

The MVATrainer [1] is a framework that wraps around TMVA to train MVAs directly on EDM data and interfacing with the CondDB. It's over 10 years old. Today, people either use TVMA outside of CMSSW on NanoAOD or private "ntuples" or go for machine learning in the python ecosystem.

The only code where the MVATrainer is referenced to in CMSSW is located in the TopQuarkAnalysis subsystem, which seems to be abandoned itself so I suppose removing the code where the MVATrainer is referenced there does not do any harm.

What do you think? Who is responsible for this package now? Is it maybe still used by someone?

[1] https://indico.cern.ch/event/25702/contributions/575840/attachments/451161/625567/plenary-b-tag.pdf

#### PR validation:

No backport intended.


guitargeek:
I figured out it was also used for the Taus in the past (not only BTV), but that code got already removed here https://github.com/cms-sw/cmssw/pull/26816.

ferencek:
As far as I am aware, this package is no longer being used in the BTV group. Whether it is still being actively used in some private code, it is hard to say. Looking at search results https://github.com/search?q=MVATrainer+language%3AC%2B%2B&type=Code there sure is a lot of private code that at least at some point relied on it.

Sooner or later this package had to go and now is probably a good time to remove it to avoid generating extra maintenance work for the core team.

slava77:
let's wait about a week for possible complaints on removal and then merge it.

slava77:
> let's wait about a week for possible complaints on removal and then merge it.

actually, based on the search suggested by @ferencek  https://github.com/search?q=MVATrainer+language%3AC%2B%2B&type=Code 
we can not remove this code without a replacement.

guitargeek:
I think the search results look scarier than they are. First you can restrict the search a bit by looking for "PhysicsTools/MVATrainer" https://github.com/search?p=96&q=PhysicsTools%2FMVATrainer+language%3AC%2B%2B&type=Code.

Fair enough, there is still a lot left. I randomly sampled a bit through the search result and at least half of the repositories where actually "forks" - albeit not in the GitHub sense - of CMSSW. Most of the others were not touched in the last 5 years except for a few outliers not touched in at least one year. It would be cool to sort the search result by "last active", but  I don't think that works in the web interface.

slava77:
hold

Some emails were sent to the reco, AT, and ML HNs.
I asked for feedback by Sept 10.
Depending on the feedback we get we can decide then.
Until then, this PR is on hold.

fabiocos:
@slava77 do we have any news about the inquiry?

Dr15Jones:
+1
for Utilities/ReleaseScripts

santocch:
+1

PR #27890, opened 2019-08-28
Update DTDigi so that it can be accomodate Phase2 TDC units.

namapane:
#### PR description:

Update DTDigi to be able to accomodate the different TDC unit of Phase2 readout (in addition to the current unit, which remains the default).

DTDigiSimLink updated accordingly. Also added an option to create digis in the new units in the simulation (switched off by default).

There is NO change in the default behaviour of any class.

This update was requested by DT DPG (@fcavallo @battibass ). For more details:
https://indico.cern.ch/event/839945/#sc-4-10-dt-digi-format-for-tdc

#### PR validation:

Tested comparing comparing of raw2digi before and after the change with both phase-1 and phase-2 digis.

Note: corresponding update for 10_X is #27884

### DIFF
--- a/DataFormats/DTDigi/interface/DTDigi.h
+++ b/DataFormats/DTDigi/interface/DTDigi.h
@@ -4,8 +4,8 @@
 /** \class DTDigi
  *
  * Digi for Drift Tubes.
- * It can be initialized/set with a time in ns or a TDC count in 25/32 ns 
- * units.
+ * It can be initialized/set with a time in ns or a TDC count in
+ * the specified base (ie number of counts/BX).
  *  
  *
  * \author N. Amapane, G. Cerminara, M. Pelliccioni - INFN Torino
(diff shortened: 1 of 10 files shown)


civanch:
please test

civanch:
+1

kpedro88:
@namapane have you tested the ability to read old instances of these data formats with this PR? The member variables of the data formats are changed, so an I/O rule might be necessary.

namapane:
@kpedro88: see discussion in #27884 (which is the same changes, for the 10_6_X series). Digis are transient and are remade from RAW when needed. No perisistent instance of digis exists to be read, in no data tier. So there is no requirement of bitwise backward-compatibility. I tested a reconstruction chain started on RAW with Phase1 TDC units, as mentioned in the PR description, and it yields identical digis as without the change, so the updated digis are fully backward compatible.

kpedro88:
+1

PR #27819, opened 2019-08-22
HGCAL: 1st version of hadronic reconstruction 

gouskos:
#### PR description:
- add the first version of the hadronic shower reconstruction. NB. Parameters need further tuning [work in progress]
- Introduce the "Out-In" step for the Trackster creation. Currently enabled by default for both EM and HAD reconstruction. Compatibility angles between Layer clusters can be tuned separately for the "In-Out" and "Out-In" steps. 
- Use at least 2 cells to form a Layer cluster in the Silicon part [more robust against noise, less combinatorial background]. For the Scintillator part the requirement is relaxed to 1 cell

#### PR validation:
First validation of the changes can be found in the following links:
- Aug 12: https://indico.cern.ch/event/838150/contributions/3526956/attachments/1892407/3121242/lg-hgccmg-hgchadreco-20190812.pdf
- Aug 19: https://indico.cern.ch/event/842095/contributions/3534224/attachments/1894454/3125220/lg-hgccmg-hgchadreco-20190819.pdf

PR #27917, opened 2019-09-02
Add energy regression and particle identification to RecoHGCal/TICL (reopened)

riga:
#### PR description

This PR adds energy regression and particle identification for trackster objects to the TrackstersProducer of the RecoHGCal/TICL package. Both the regressed energy and ID probabilities for different types of particles are the outputs of a single neural network, stored as a TensorFlow model. The first version of this model is proposed in

cms-data/RecoHGCal-TICL#1.

This PR is a clean version of #27634. The changes since the last review comments can be found [here](https://github.com/riga/cmssw/compare/c75d2de...e9fb179).

#### Performance

Tested with workflow 20634.0 (`TTbar_14TeV_TuneCUETP8M1_2026D41PU` with 200 PU).

The CPU runtime for inference scales linearly with the batch size, i.e., the number of tracksters, as expected for the `no_threads` TensorFlow session. The average runtime per trackster is ~0.6 ms (see slope below). The time includes trackster selection, preprocessing and filling of input data, the actual inference, and handling of results (== `PatternRecognitionbyCA::energyRegressionAndID`).

runtime

The memory consumption, obtained using the `SimpleMemoryCheck` with and without running the regression and ID, is 178 MB for a maximum batch size of ~750 tracksters.

To get an estimate of the memory used per event, I queried `VmRSS` of `/proc/self/status` ([link](https://stackoverflow.com/a/64166/2408501)) at the beginning and at the end of `PatternRecognitionbyCA::energyRegressionAndID`, and also before and after loading the graph and setting up the session.

- Load graph: + 3.6 MB
- Load session: +7.2 MB
- `energyRegressionAndID`:
memory

### DIFF
--- a/DataFormats/HGCalReco/interface/Trackster.h
+++ b/DataFormats/HGCalReco/interface/Trackster.h
@@ -28,6 +28,30 @@ namespace ticl {
 
     edm::ProductID seedID;
     int seedIndex;
+
+    // regressed energy
+    float regressed_energy;
+
+    // types considered by the particle identification
+    enum ParticleType {
+      photon = 0,
+      electron,
+      muon,
+      neutral_pion,
+      charged_hadron,
+      neutral_hadron,
+      ambiguous,
+      unknown,
+    };
+
+    // trackster ID probabilities
+    std::array<float, 8> id_probabilities;
+
+    // convenience method to return the ID probability for a certain particle type
+    inline float id_probability(ParticleType type) {
+      // probabilities are stored in the same order as defined in the ParticleType enum
+      return id_probabilities[(int)type];
+    }
   };
 }  // namespace ticl
 #endif
--- a/DataFormats/HGCalReco/interface/Trackster.h
+++ b/DataFormats/HGCalReco/interface/Trackster.h
@@ -48,7 +48,7 @@ namespace ticl {
     std::array<float, 8> id_probabilities;
 
     // convenience method to return the ID probability for a certain particle type
-    inline float id_probability(ParticleType type) {
+    inline float id_probability(ParticleType type) const {
       // probabilities are stored in the same order as defined in the ParticleType enum
       return id_probabilities[(int)type];
     }
--- a/DataFormats/HGCalReco/interface/Trackster.h
+++ b/DataFormats/HGCalReco/interface/Trackster.h
@@ -33,7 +33,7 @@ namespace ticl {
     float regressed_energy;
 
     // types considered by the particle identification
-    enum ParticleType {
+    enum class ParticleType {
       photon = 0,
       electron,
       muon,
--- a/RecoHGCal/TICL/interface/GlobalCache.h
+++ b/RecoHGCal/TICL/interface/GlobalCache.h
@@ -7,15 +7,17 @@
 #include "PhysicsTools/TensorFlow/interface/TensorFlow.h"
 
 namespace ticl {
-  // base struct across ticl for objects hold in the edm::GlobalCache by plugins
-  struct CacheBase {
+  // base class across ticl for objects hold in the edm::GlobalCache by plugins
+  class CacheBase {
+  public:
     CacheBase(const edm::ParameterSet& params) {}
 
     virtual ~CacheBase() {}
   };
 
   // data structure hold by TrackstersProducer to store the TF graph for energy regression and ID
-  struct TrackstersCache : CacheBase {
+  class TrackstersCache : public CacheBase {
+  public:
     TrackstersCache(const edm::ParameterSet& params) : CacheBase(params), eidGraphDef(nullptr) {}
 
     ~TrackstersCache() override {}
--- a/RecoHGCal/TICL/interface/GlobalCache.h
+++ b/RecoHGCal/TICL/interface/GlobalCache.h
@@ -1,0 +1,27 @@
+// Definition of objects placed in the edm::GlobalCache.
+
+#ifndef RecoHGCal_TICL_GlobalCache_H__
+#define RecoHGCal_TICL_GlobalCache_H__
+
+#include "FWCore/ParameterSet/interface/ParameterSetDescription.h"
+#include "PhysicsTools/TensorFlow/interface/TensorFlow.h"
+
+namespace ticl {
+  // base struct across ticl for objects hold in the edm::GlobalCache by plugins
+  struct CacheBase {
+    CacheBase(const edm::ParameterSet& params) {}
+
+    virtual ~CacheBase() {}
+  };
+
+  // data structure hold by TrackstersProducer to store the TF graph for energy regression and ID
+  struct TrackstersCache : CacheBase {
+    TrackstersCache(const edm::ParameterSet& params) : CacheBase(params), eidGraphDef(nullptr) {}
+
+    ~TrackstersCache() override {}
+
+    std::atomic<tensorflow::GraphDef*> eidGraphDef;
+  };
+}  // namespace ticl
+
+#endif  // RecoHGCal_TICL_GlobalCache_H__
--- a/RecoHGCal/TICL/plugins/PatternRecognitionbyCA.cc
+++ b/RecoHGCal/TICL/plugins/PatternRecognitionbyCA.cc
@@ -14,23 +14,22 @@ PatternRecognitionbyCA::PatternRecognitionbyCA(const edm::ParameterSet &conf, co
     : PatternRecognitionAlgoBase(conf, cache),
       out_in_dfs_(conf.getParameter<bool>("out_in_dfs")),
       max_out_in_hops_(conf.getParameter<int>("max_out_in_hops")),
+      theGraph_(std::make_unique<HGCGraph>()),
+      min_cos_theta_(conf.getParameter<double>("min_cos_theta")),
+      min_cos_pointing_(conf.getParameter<double>("min_cos_pointing")),
+      missing_layers_(conf.getParameter<int>("missing_layers")),
+      min_clusters_per_ntuplet_(conf.getParameter<int>("min_clusters_per_ntuplet")),
+      max_delta_time_(conf.getParameter<double>("max_delta_time")),
+      eidInputName_(conf.getParameter<std::string>("eid_input_name")),
+      eidOutputNameEnergy_(conf.getParameter<std::string>("eid_output_name_energy")),
+      eidOutputNameId_(conf.getParameter<std::string>("eid_output_name_id")),
+      eidMinClusterEnergy_(conf.getParameter<double>("eid_min_cluster_energy")),
+      eidNLayers_(conf.getParameter<int>("eid_n_layers")),
+      eidNClusters_(conf.getParameter<int>("eid_n_clusters")),
       eidSession_(nullptr) {
-  theGraph_ = std::make_unique<HGCGraph>();
-  min_cos_theta_ = conf.getParameter<double>("min_cos_theta");
-  min_cos_pointing_ = conf.getParameter<double>("min_cos_pointing");
-  missing_layers_ = conf.getParameter<int>("missing_layers");
-  min_clusters_per_ntuplet_ = conf.getParameter<int>("min_clusters_per_ntuplet");
-  max_delta_time_ = conf.getParameter<double>("max_delta_time");
-  eidInputName_ = conf.getParameter<std::string>("eid_input_name");
-  eidOutputNameEnergy_ = conf.getParameter<std::string>("eid_output_name_energy");
-  eidOutputNameId_ = conf.getParameter<std::string>("eid_output_name_id");
-  eidMinClusterEnergy_ = conf.getParameter<double>("eid_min_cluster_energy");
-  eidNLayers_ = conf.getParameter<int>("eid_n_layers");
-  eidNClusters_ = conf.getParameter<int>("eid_n_clusters");
-
   // mount the tensorflow graph onto the session when set
   const TrackstersCache *trackstersCache = dynamic_cast<const TrackstersCache *>(cache);
-  if (trackstersCache->eidGraphDef != nullptr) {
+  if (trackstersCache != nullptr && trackstersCache->eidGraphDef != nullptr) {
     eidSession_ = tensorflow::createSession(trackstersCache->eidGraphDef);
   }
 }
@@ -225,8 +224,8 @@ void PatternRecognitionbyCA::energyRegressionAndID(const std::vector<reco::CaloC
     for (int j = 0; j < eidNLayers_; j++) {
       for (int k = seenClusters[j]; k < eidNClusters_; k++) {
         float *features = &input.tensor<float, 4>()(i, j, k, 0);
-        for (int l = 0; l < eidNFeatures_; l++, features++) {
-          *features = 0.f;
+        for (int l = 0; l < eidNFeatures_; l++) {
+          *(features++) = 0.f;
         }
       }
     }
--- a/RecoHGCal/TICL/plugins/PatternRecognitionbyCA.cc
+++ b/RecoHGCal/TICL/plugins/PatternRecognitionbyCA.cc
@@ -10,16 +10,29 @@
 
 using namespace ticl;
 
-PatternRecognitionbyCA::PatternRecognitionbyCA(const edm::ParameterSet &conf)
-    : PatternRecognitionAlgoBase(conf),
+PatternRecognitionbyCA::PatternRecognitionbyCA(const edm::ParameterSet &conf, const CacheBase *cache)
+    : PatternRecognitionAlgoBase(conf, cache),
       out_in_dfs_(conf.getParameter<bool>("out_in_dfs")),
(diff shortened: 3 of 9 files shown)


perrotta:
please test workflow 20493.52 with cms-sw/cmsdist#5120, cms-data/RecoHGCal-TICL#1

DataFormats/HGCalReco/interface/Trackster.h
```diff
@@ -28,6 +28,30 @@ namespace ticl {
...
+      neutral_hadron,
+      ambiguous,
+      unknown,
+    };
+
+    // trackster ID probabilities
+    std::array<float, 8> id_probabilities;
+
+    // convenience method to return the ID probability for a certain particle type
+    inline float id_probability(ParticleType type) {
```
VinInn: add ```const```

RecoHGCal/TICL/plugins/TrackstersProducer.cc
```diff
@@ -46,8 +53,29 @@ class TrackstersProducer : public edm::stream::EDProducer<> {
...
+  if (!graphPath.empty()) {
+    graphPath = edm::FileInPath(graphPath).fullPath();
+    cache->eidGraphDef = tensorflow::loadGraphDef(graphPath);
+  }
+
+  return cache;
+}
+
+void TrackstersProducer::globalEndJob(TrackstersCache* cache) {
+  if (cache->eidGraphDef != nullptr) {
```
VinInn: no need to protect a ```delete```

RecoHGCal/TICL/plugins/PatternRecognitionbyCA.cc
```diff
@@ -10,13 +10,26 @@
...
+  eidOutputNameId_ = conf.getParameter<std::string>("eid_output_name_id");
+  eidMinClusterEnergy_ = conf.getParameter<double>("eid_min_cluster_energy");
+  eidNLayers_ = conf.getParameter<int>("eid_n_layers");
+  eidNClusters_ = conf.getParameter<int>("eid_n_clusters");
+
+  // mount the tensorflow graph onto the session when set
+  const TrackstersCache *trackstersCache = dynamic_cast<const TrackstersCache *>(cache);
+  if (trackstersCache->eidGraphDef != nullptr) {
+    eidSession_ = tensorflow::createSession(trackstersCache->eidGraphDef);
+  }
```
VinInn: else ???

RecoHGCal/TICL/plugins/PatternRecognitionbyCA.cc
```diff
@@ -10,13 +10,26 @@
...
+  eidInputName_ = conf.getParameter<std::string>("eid_input_name");
+  eidOutputNameEnergy_ = conf.getParameter<std::string>("eid_output_name_energy");
+  eidOutputNameId_ = conf.getParameter<std::string>("eid_output_name_id");
+  eidMinClusterEnergy_ = conf.getParameter<double>("eid_min_cluster_energy");
+  eidNLayers_ = conf.getParameter<int>("eid_n_layers");
+  eidNClusters_ = conf.getParameter<int>("eid_n_clusters");
+
+  // mount the tensorflow graph onto the session when set
+  const TrackstersCache *trackstersCache = dynamic_cast<const TrackstersCache *>(cache);
+  if (trackstersCache->eidGraphDef != nullptr) {
```
VinInn: actually one has to test if the dynamic_cast failed...

RecoHGCal/TICL/plugins/PatternRecognitionbyCA.cc
```diff
@@ -10,13 +10,26 @@
...
-PatternRecognitionbyCA::PatternRecognitionbyCA(const edm::ParameterSet &conf) : PatternRecognitionAlgoBase(conf) {
+PatternRecognitionbyCA::PatternRecognitionbyCA(const edm::ParameterSet &conf, const CacheBase *cache)
+    : PatternRecognitionAlgoBase(conf, cache), eidSession_(nullptr) {
   theGraph_ = std::make_unique<HGCGraph>();
   min_cos_theta_ = conf.getParameter<double>("min_cos_theta");
   min_cos_pointing_ = conf.getParameter<double>("min_cos_pointing");
   missing_layers_ = conf.getParameter<int>("missing_layers");
   min_clusters_per_ntuplet_ = conf.getParameter<int>("min_clusters_per_ntuplet");
   max_delta_time_ = conf.getParameter<double>("max_delta_time");
+  eidInputName_ = conf.getParameter<std::string>("eid_input_name");
```
VinInn: all configurable should be declared ```const``` and initialized as
```theGraph_(std::make_unique<HGCGraph>()),``` etc

File: RecoHGCal/TICL/plugins/PatternRecognitionbyCA.cc [hunk 6]
```diff
@@ -94,4 +107,151 @@ void PatternRecognitionbyCA::makeTracksters(const PatternRecognitionAlgoBase::In
...
+        seenClusters[j]++;
+      }
+    }
+
+    // zero-fill features of empty clusters in each layer (6)
+    for (int j = 0; j < eidNLayers_; j++) {
+      for (int k = seenClusters[j]; k < eidNClusters_; k++) {
+        float *features = &input.tensor<float, 4>()(i, j, k, 0);
+        for (int l = 0; l < eidNFeatures_; l++, features++) {
+          *features = 0.f;
```
VinInn: why not
``` *(features++)``` as everywhere else?

File: DataFormats/HGCalReco/interface/Trackster.h (hunk 1 above)
Comment (riga) [reply DataFormats/HGCalReco/interface/Trackster.h]: ✔︎

File: RecoHGCal/TICL/plugins/TrackstersProducer.cc (hunk 2 above)
Comment (riga) [reply RecoHGCal/TICL/plugins/TrackstersProducer.cc]: ✔︎

File: RecoHGCal/TICL/plugins/PatternRecognitionbyCA.cc (hunk 3 above)
Comment (riga) [reply RecoHGCal/TICL/plugins/PatternRecognitionbyCA.cc]: When the `eidGraphDef` in the cache is not set (which happens when no graph path was set) the `eidSession_` should not be set as well and the inference is not performed.

File: RecoHGCal/TICL/plugins/PatternRecognitionbyCA.cc (hunk 4 above)
Comment (riga) [reply RecoHGCal/TICL/plugins/PatternRecognitionbyCA.cc]: ✔︎

File: RecoHGCal/TICL/plugins/PatternRecognitionbyCA.cc (hunk 6 above)
Comment (riga) [reply RecoHGCal/TICL/plugins/PatternRecognitionbyCA.cc]: ✔︎

File: DataFormats/HGCalReco/interface/Trackster.h [hunk 7]
```diff
@@ -28,6 +28,30 @@ namespace ticl {
 
     edm::ProductID seedID;
     int seedIndex;
+
+    // regressed energy
+    float regressed_energy;
+
+    // types considered by the particle identification
+    enum ParticleType {
```
kpedro88: always better to use enum class

kpedro88:
please test workflow 20493.52 with cms-sw/cmsdist#5120, cms-data/RecoHGCal-TICL#1

DataFormats/HGCalReco/interface/Trackster.h
```diff
@@ -28,6 +28,30 @@ namespace ticl {
...
+      unknown,
+    };
+
+    // trackster ID probabilities
+    std::array<float, 8> id_probabilities;
+
+    // convenience method to return the ID probability for a certain particle type
+    inline float id_probability(ParticleType type) const {
+      // probabilities are stored in the same order as defined in the ParticleType enum
+      return id_probabilities[(int)type];
```
slava77: non-trivial methods are not really expected for plain structs in data formats.
Perhaps it's time to make this a normal class with private/encapsulated data members

RecoHGCal/TICL/interface/GlobalCache.h
```diff
@@ -0,0 +1,27 @@
...
+  // base struct across ticl for objects hold in the edm::GlobalCache by plugins
+  struct CacheBase {
+    CacheBase(const edm::ParameterSet& params) {}
+
+    virtual ~CacheBase() {}
+  };
+
+  // data structure hold by TrackstersProducer to store the TF graph for energy regression and ID
+  struct TrackstersCache : CacheBase {
+    TrackstersCache(const edm::ParameterSet& params) : CacheBase(params), eidGraphDef(nullptr) {}
```
slava77: time to switch to classes?

RecoHGCal/TICL/plugins/PatternRecognitionbyCA.cc
slava77: how does this distinguish a bug upstream from a desired silent do nothing behavior?
The code should not be allowed to silently skip when it was expected to do something.

RecoHGCal/TICL/plugins/PatternRecognitionbyCA.cc
```diff
@@ -94,4 +107,151 @@ void PatternRecognitionbyCA::makeTracksters(const PatternRecognitionAlgoBase::In
...
+  // i -> batch / trackster
+  // j -> layer
+  // k -> cluster
+  // l -> feature
+
+  // set default values per trackster, determine if the cluster energy threshold is passed,
+  // and store indices of hard tracksters
+  std::vector<int> tracksterIndices;
+  for (int i = 0; i < (int)tracksters.size(); i++) {
+    // set default values (1)
```
slava77: it looks like range loops should work well here and in the other loops in the scope

RecoHGCal/TICL/plugins/TrackstersProducer.cc
```diff
@@ -21,19 +21,26 @@
...
+#include "RecoHGCal/TICL/interface/GlobalCache.h"
 #include "PatternRecognitionbyCA.h"
 #include "PatternRecognitionbyMultiClusters.h"
 
+#include "PhysicsTools/TensorFlow/interface/TensorFlow.h"
+
 using namespace ticl;
 
-class TrackstersProducer : public edm::stream::EDProducer<> {
+class TrackstersProducer : public edm::stream::EDProducer<edm::GlobalCache<TrackstersCache>> {
```
slava77: this patterns seems to make the ticl::TrackstersCache -> ticl::CacheBase hierarchy useless.
There is no way to select a different algo without changing the producer as well.
Shouldn't the  CacheBase be removed?

felicepantaleo:
@riga can you please fix conflicts and check performance with the latest IB that will have #27819 integrated?

kpedro88:
@riga this needs to be rebased

riga:
@perrotta I will update the runtime numbers in the PR description with the current numbers, and also add the updated memory consumption with #27819 integrated.

@felicepantaleo Sure, I can check again, however, I don't think that the ~0.5ms per trackster will change, but rather just the number of tracksters.

@kpedro88 I just rebased and will address the final comments.

RecoHGCal/TICL/plugins/TrackstersProducer.cc
riga: I interpreted the hard-coded initializer

```cpp
myAlgo_(std::make_unique<PatternRecognitionbyCA>(ps, cache))
```

as a temporary decision, given that `myAlgo_` refers to a generic `PatternRecognitionAlgoBase` ptr.

Regarding the TrackstersCache -> CacheBase structure, my intention here was that I didn't want to force all possible (future) algos derived from that base class to explicitly expect a specialized `TrackstersCache`.

RecoHGCal/TICL/plugins/PatternRecognitionbyCA.cc
riga: 1st comment: You're right. The entire NN inference part was intended to be optional, but I think we should make it mandatory and throw an exception at this point.

2nd comment: I actually need the index a few lines below to store the positions of selected tracksters. But I will change the loops in the scope where possible.

DataFormats/HGCalReco/interface/Trackster.h
riga: I agree. But making all members private and adding getters/setters would result in a lot of additional changes. Maybe we should create another PR for this?

RecoHGCal/TICL/interface/GlobalCache.h
riga: ✔︎

kpedro88:
please test workflow 20493.52 with cms-sw/cmsdist#5120, cms-data/RecoHGCal-TICL#1

rovere:
Hi all,
this PR and its previous incarnation have been around for nearly 2 months.
It would be nice, now that almost all review comments have been addressed, to have the PR integrated ASAP so that pending PRs (already ready) could be submitted.
Thanks,
--Marco.

kpedro88:
+upgrade

perrotta:
> The memory consumption, obtained using the SimpleMemoryCheck with and without running the
> regression and ID, is 178 MB for a maximum batch size of ~750 tracksters.

@riga 178 MB increase is not negligible. On the other hand, SimpleMemoryCheck does not tell us the memory consumption event per event, and if it just corresponds to the maximum batch size it can still be acceptable. Could you please measure the average memory consumption in your sample, or even provide a plot, similar to the one that you already attach for timing, with the memory vs size?

rovere:
@perrotta @slava77 
Can we somehow envisage an automatic procedure to have such metrics? This would cut the integration time.

riga:
@perrotta I added the memory consumption per event to the PR description. In short:

- Load graph: + 3.6 MB
- Load session: +7.2 MB
- Batch of ~750: +9.1 MB

However, I only measured the currently used memory before and after the inference method (but inside it to include its stack). Is there a common tool somewhere to obtain the peak memory inside a certain scope?

fabiocos:
@rovere let me remind that for a fast overall check each PR test embeds SimpleMemoryCheck (as well as the Timing service) global report into the JobReport.xml . More sophisticated analyses needs an ad hoc study

perrotta:
> @perrotta I added the memory consumption per event to the PR description. In short:
> 
> * Load graph: + 3.6 MB
> * Load session: +7.2 MB
> * Batch of ~750: +9.1 MB
> 
> However, I only measured the currently used memory before and after the inference method (but inside it to include its stack). Is there a common tool somewhere to obtain the peak memory inside a certain scope?

Thank you @riga 
In the reco twiki we suggest using IgProf for profilin the memory:

RecoHGCal/TICL/plugins/PatternRecognitionbyCA.cc
perrotta: Please make const all all configurable data members

DataFormats/HGCalReco/interface/Trackster.h
perrotta: Ok to postpone to another PR
Issue #28049 has been created to keep track of it

perrotta:
@riga the comment https://github.com/cms-sw/cmssw/pull/27917#pullrequestreview-282669289 from @VinInn (where he suggests  to make const all all configurable data members of the PatternRecognitionbyCA class) was not implemented.
Please either apply or comment on it.
After that I think that the code is ready to go.

rovere:
> @rovere let me remind that for a fast overall check each PR test embeds SimpleMemoryCheck (as well as the Timing service) global report into the JobReport.xml . More sophisticated analyses needs an ad hoc study

Thanks @fabiocos  for the reminder. It would then be useful to check those numbers even if, IIUC, there's no before/after PR JobReport.xml. Maybe that's something one could envisage for a limited number of workflows so that the direct comparison becomes easier.

riga:
@perrotta I made all members `const`. Also, I ran igprof and browsed through the results. This is the output for the added `energyRegressionAndID` method:

igprof

Fyi: Running igprof-navigator on an lxplus machine (in a CMSSW env) returned an empty page in my browser and the process complains about a "SQLite header and source version mismatch". In the end, I used

```shell
docker run --rm -ti -p 8080:8080 -v `pwd`/thefile.sql3:/thefile.sql3 igprof/igprof igprof-navigator /thefile.sql3
```

Maybe this can go on the Twiki page you pointed me to in case other people have the same issue.

RecoHGCal/TICL/plugins/PatternRecognitionbyCA.cc
riga: ✔︎

perrotta:
please test workflow 20493.52 with cms-sw/cmsdist#5120, cms-data/RecoHGCal-TICL#1

perrotta:
+1
- Energy regression and particle id implemented for HGCal trackster objects
- When enabled, cpu and memory consumption shown to be acceptable even at 200 PU
- TrackstersProducer left configured to skip regression and id: therefore no change is expected in the default config: jenkins tests pass and show no differences

fabiocos:
@kpedro88 could you please check this PR?

kpedro88:
+upgrade

fabiocos:
+1

Issue #27889, opened 2019-08-28
features for UL re-miniAOD in 10_6_X


slava77:
This is to keep track of features needed for the UL re-miniAOD.
The 2017 will likely happen first.
The target start date is still unclear (it will probably be around October 2019).

As of Aug 7  (after #31072 was merged) the re-miniAOD definitions aiming for UL re-miniAOD should use a  ```run2_miniAOD_UL```  modifier, if the configuration is to be controlled by a modifier. 

[as of Dec 9]
- [x] #27882 "Disabled PCA inputs for deepTau v2". Target: all re-miniAOD. (merged Sep 19, 2019)
    - the change to deepTau2017v2p1 is already applied in UL 2016 and 2018 by default
    - controlled by run2_miniAOD_devel
- [x] ~#28845 "Add customizer for PUPPI tune v13" (merged Feb 27, 2020)~
    - this is implemented with a customize method; superseded by v14
- [x] #29601 PUPPI tune v14 (merged May 14, 2020)
- [x] #29045 Adding parameters needed for PUjetID 94X and 102X (merged Mar 3, 2020)
    - bugfix for run2_miniAOD_devel -based setup in #29377 (merged April 2, 2020)
    - controlled by run2_miniAOD_devel
- [x] #29330 Fix MET unclustered energy computation (merged April 1, 2020)
    - controlled by run2_miniAOD_devel
- [x] #29392 Fix MET Significance, Covariance Matrix and unclustered energy lepton subtraction (merged Apr 21, 2020)
    - controlled by run2_miniAOD_devel
- [x] #29521 Keeping sumPt from MET Significance calculation  (merged Apr 21, 2020)
    - enabled, no era modifiers
- [x] #29062 MXNet-based implementation of the ParticleNet tagger  (merged Apr 18, 2020)
    - controlled by run2_miniAOD_devel
- [x] #29324 inverse beta variable in pat::Muons (merged Apr 17, 2020)
    - enabled, no era modifiers
- [x] #29529 Update of the lostTrack collection for MUO  (merged Apr 23, 2020)
- [x] #29537 fix sign of pat::Muon::dxy (merged May 1, 2020)
- [x] #29771 PPS rereco-in-miniAOD (merged May 25, 2020)
    - controlled by run2_miniAOD_devel
- [x] #29739 Tau againstElectronDeadECAL (merged May 27, 2020 )
    - controlled by run2_miniAOD_devel
[...]

PR #27929, opened 2019-09-04
Simplify PUPPI Photon producer and change PUPPI isolation algo

ahinzmann:
#### PR description:

Simplify PUPPI MET computation, by removing (over-complicated) PUPPI photon algorithm and replace it by a simple option in PUPPI itself (avoid any kind of matching etc.).
The PR is supposed to keep tuning and physics of puppi (for jets) and puppiForMET untouched, however puppiNoLeptons (used for lepton isolation) is modified to include the same protection for high pT photons as puppiForMET.

#### PR validation:

Checked on 3000 QCD events that the high-pT photon candidates get the weight increased to 1 in old and new implementation.
Small differences arise when particle pt is close to the pT threshold of 20 within packedCandidate precision.

@lathomas @zdemirag @abenecke @nhanvtran @violatingcp


slava77:
@cmsbuild please test

ahinzmann:
Added commit to recover lepton isolation quantities.
In practice there are three flavors of PUPPI particles:
puppi -> jets
puppiNoLep -> lepton isolation
puppiForMET -> MET

slava77:
@cmsbuild please test

slava77:
@ahinzmann 
was this already shown in a JME meeting?
please add a link to the presentation in the PR description if available.
Thank you.

CommonTools/PileupAlgos/src/PuppiContainer.cc
```diff
@@ -322,8 +324,13 @@ std::vector<double> const &PuppiContainer::puppiWeights() {
...
     if (pWeight * fPFParticles[i0].pt() < fPuppiAlgo[pPupId].neutralPt(fNPV) && rParticle.id == 0)
       pWeight = 0;  //threshold cut on the neutral Pt
-    if ((fPtMax > 0) && (rParticle.id == 0))
-      pWeight = min(max(pWeight, fPFParticles[i0].pt() / fPtMax), 1.);
+    // Protect high pT photons (important for gamma to hadronic recoil balance)
+    if ((fPtMaxPhotons > 0) && (rParticle.pdgId == 22) && (std::abs(fPFParticles[i0].eta()) < fEtaMaxPhotons) &&
+        (fPFParticles[i0].pt() > fPtMaxPhotons))
+      pWeight = 1.;
+    // Protect high pT neutrals
+    if ((fPtMaxNeutrals > 0) && (rParticle.id == 0))
```
slava77: it may be more clear to say here ```else if ```.
It's a bit more trivial than to figure out that the pWeight set to 1 above will not be changed if it inters this condition.

PhysicsTools/PatAlgos/python/slimming/puppiForMET_cff.py
```diff
@@ -21,52 +18,34 @@ def makePuppies( process ):
...
-    process.load('CommonTools.PileupAlgos.PhotonPuppi_cff')
-    task.add(process.puppiPhoton)
-    addToProcessAndTask('puppiForMET', process.puppiPhoton.clone(), process, task)
-    #Line below replaces reference linking wiht delta R matching because the puppi references after merging are not consistent with those of the original PF collection
-    process.puppiForMET.useRefs          = False
-    #Line below points puppi MET to puppi no lepton which increases the response
-    process.puppiForMET.puppiCandName    = 'puppiMerged'
-
+    addToProcessAndTask('puppiNoLepPhoton', process.puppiNoLep.clone(), process, task)
+    process.puppiNoLepPhoton.PtMaxPhotons = cms.double(20.)
```
slava77: please drop type specifications for all parameters which already exist.

PhysicsTools/PatAlgos/python/slimming/puppiForMET_cff.py
```diff
@@ -21,52 +18,34 @@ def makePuppies( process ):
...
-    setupPuppiPhotonMiniAOD(process)
-    task.add(process.egmPhotonIDTask)
-    #Line below replaces reference linking wiht delta R matching because the puppi references after merging are not consistent with those of the original packed candidate collection
-    process.puppiForMET.useRefs          = False
-    #Line below points puppi MET to puppi no lepton which increases the response
-    process.puppiForMET.puppiCandName    = 'puppiMerged'
+    process.puppiNoLep.candName = cms.InputTag('pfNoLepPUPPI')
+    process.puppiNoLep.useWeightsNoLep = True
+    addToProcessAndTask('puppiNoLepPhoton', process.puppiNoLep.clone(), process, task)
+    process.puppiNoLepPhoton.PtMaxPhotons = cms.double(20.)
```
slava77: please drop type specifications for all parameters which already exist.

PhysicsTools/PatAlgos/python/slimming/puppiForMET_cff.py
```diff
@@ -21,52 +18,34 @@ def makePuppies( process ):
...
-    task.add(process.puppiMerged)
-    process.load('CommonTools.PileupAlgos.PhotonPuppi_cff')
-    task.add(process.puppiPhoton)
-    addToProcessAndTask('puppiForMET', process.puppiPhoton.clone(), process, task)
-    #Line below replaces reference linking wiht delta R matching because the puppi references after merging are not consistent with those of the original PF collection
-    process.puppiForMET.useRefs          = False
-    #Line below points puppi MET to puppi no lepton which increases the response
-    process.puppiForMET.puppiCandName    = 'puppiMerged'
-
+    addToProcessAndTask('puppiNoLepPhoton', process.puppiNoLep.clone(), process, task)
```
slava77: would a puppiNoLepModPhoton or puppiNoLepModNeutrals or Mod -> Safe or Protect be a bit more intuitive?

slava77 (commented):
This PR is apparently introducing a third instance of PuppiProducer.
This is a pretty heavy producer (still between 5 and 10% of miniAOD processing time), compared to PuppiPhoton, which is about a factor of 10 faster.
Simple addition of another clone of PuppiProducer needs some better justification than just the apparent complexity of PuppiPhoton due to the CPU cost increase.

Can the current solution be replaced by puppiNoLep with some appropriate rescaling of weights?

If somehow I missed something in the settings and the newly added puppiNoLepPhoton is somehow much faster than puppiNoLep, please provide some measurement in support.

slava77:
@ahinzmann 
please clarify on your plans to update this PR or respond to the review comments
https://github.com/cms-sw/cmssw/pull/27929#pullrequestreview-284435773
Thank you.

ahinzmann:
I'll come back to this after a workshop in about a week. Will compare the computation times then.

slava77:
I made some measurements using re-miniAOD of 2017 94X MC (I used 10K events [or max of whta's available] in RelValH125GGgluonfusion, RelValZEE, and RelValQCD_FlatPt_15_3000HS with pileup).
I see about 4% increase in CPU and essentially no change in the physics outputs.

This does not change my conclusions from the visual inspection in https://github.com/cms-sw/cmssw/pull/27929#pullrequestreview-284435773.

ahinzmann:
Running only makePuppiesFromMiniAOD, the extra instance of PuppiProducer with useExistingWeights is still slower than than the PuppiPhoton. On 1000 ttbar events the average time per event is 2% slower. The effect on the whole MiniAOD should be less, still there is no gain yet, apart from reducing the amount of code.

slava77:
@cmsbuild please test

slava77:
> apart from reducing the amount of code.

is it possible to remove the PuppiPhoton, if the CPU increase was acceptable?
Or do we still need to maintain PuppiPhoton for some future developments?

ahinzmann:
if the CPU budget is ok, we could remove PuppiPhoton

ahinzmann:
To have both CPU improvement and simplification, this is now actually changing the physics output for puppiNoLepton for lepton-isolation, protecting high-pT photons just like in puppiForMET. Impact on isolation performance should be negligible, but need to check this.

slava77:
@cmsbuild please test

ahinzmann:
The 3 changes in TTbar simulation are expected for puppiNoLepton. The changes in jets and MET in data are not expected, not clear where they come from.

ahinzmann:
Were the comparisons including any other code than in this PR? Apart from the min(max()) -> std::clamp, I do not see any change that could affect PUPPI jets in data?

CommonTools/PileupAlgos/src/PuppiContainer.cc
slava77: considering the comment in https://github.com/cms-sw/cmssw/pull/27929#issuecomment-535917262
I'm not 100% sure now that this if -> else if is logically identical.

ahinzmann: the "if else" should recover the same logic that PuppiPhoton had (so without the else it was actually wrong). However, all of this not should no affect puppi (jets), but only puppiNoLeptons

slava77:
@cmsbuild please test

ahinzmann:
So now jets and MET are untouched as expected. Small changes are observed for lepton isolation in samples enriched with high-pT photons due to the addition of the protection for high-pT photons.

ahinzmann:
@ishvetso What is your take on the change of the puppi electron isolation, due to changing the weight of PF-photons with pT>20 GeV to 1? Do you think this causes harm to the performance?
It looks like it leaves the isolated electrons mostly untouched, but increases those with intermediate isolation values to very high ones.

slava77:
@ahinzmann 
there is now a conflict with CommonTools/PileupAlgos/plugins/PuppiProducer.cc
please update this PR 

@SohamBhattacharya and @afiqaize please take a note of this PR for the effects on the puppi no-lep photon isolation.
@ishvetso @ikrav please check https://github.com/cms-sw/cmssw/pull/27929#issuecomment-536993851 
Thank you.

slava77:
@cmsbuild please test

fabiocos:
please test

slava77:
> 
> @SohamBhattacharya and @afiqaize please take a note of this PR for the effects on the puppi no-lep photon isolation.
> @ishvetso @ikrav please check [#27929 (comment)](https://github.com/cms-sw/cmssw/pull/27929#issuecomment-536993851)
> Thank you.

ikrav:
I have not been working with PUPPI isolation for quite some time, and not sure if Ivan has either.
So I don’t have an immediate answer. It may be a good idea to get in touch the present 
EGM ID conveners.

slava77:
> I have not been working with PUPPI isolation for quite some time, and not sure if Ivan has either. So I don’t have an immediate answer. It may be a good idea to get in touch the present EGM ID conveners.
> […](#)
> On Oct 14, 2019, at 7:11 PM, Slava Krutelyov ***@***.***> wrote: @SohamBhattacharya and @afiqaize please take a note of this PR for the effects on the puppi no-lep photon isolation. @ishvetso @ikrav please check #27929 (comment) Thank you. — You are receiving this because you were mentioned. Reply to this email directly, view it on GitHub, or unsubscribe.

@jainshilpi @lsoffi 
please check and advise on the situation with the  puppi no-lep photon isolation.
Thank you.

slava77:
@cmsbuild please test

to refresh the tests

slava77:
+1

for https://github.com/cms-sw/cmssw/pull/27929	71d320d 
- jenkins tests pass and comparisons with the baseline show differences related to puppiNoLep photons.
- local tests show somewhat expected and acceptable (eventually, after checking with EGM) behavior
    - the most visible changes are in slimmedElectrons puppiNoLep photon isolation in photon-reach sample https://github.com/cms-sw/cmssw/pull/27929#issuecomment-536959025. This is a result of pf photons getting now a larger weight in puppiNoLep. 
     - some visible changes also appear in the deepTau ID variables; the effect appears to be rather small based on ten-tau (RelValTenTau_15_500 from 1100pre7) sample there is a rather tiny improvement in rejection vs electron and small degradation in rejection vs jets (below)
    - technical costs are reduced by about 0.3% in CPU time wrt miniAOD job, by removing puppiMerged and changing puppiForMET

deepTau2017v2p1:VSe
all_sign1088VSorig_RelValTenTau_15_500_2018c_patTaus_slimmedTaus__PAT_obj___tauIDs__72__second
VSeVVTight
all_sign1088VSorig_RelValTenTau_15_500_2018c_patTaus_slimmedTaus__PAT_obj___tauIDs__91__second


VSjet
all_sign1088VSorig_RelValTenTau_15_500_2018c_patTaus_slimmedTaus__PAT_obj___tauIDs__73__second
VSjetVVTight
all_sign1088VSorig_RelValTenTau_15_500_2018c_patTaus_slimmedTaus__PAT_obj___tauIDs__92__second

@mbluj @swozniewski

mbluj:
I think that differences in deepTau scores are small enough to be accepted. However, it can be an issue for analyses if there is a plan to backport this development to releases used for ongoing Run-2 analyses (mainly 102X).

fabiocos:
@santocch as far as I am concerned this PR could technically be integrated in master, I am not sure we really want to backport it (up to proponents to say). In any case please check
@peruzzim @fgolf FYI for possible downstream impact on nanoAOD

ahinzmann:
We don't want this change in 10_2.
We only plan to use it for Run3 and possibly UL.
Possibly a back port to 10_6 is desired (but not now) once we actually finalize and change the tune of PUPPI possibly a re-MiniAOD of UL.

slava77:
Regarding a backport,
my understanding of this PR is that it's purpose is to simplify the software use by avoiding using a "complicated" ```PuppiPhoton``` algorithm.
As such, a backport should not be needed, unless it becomes a part of a future development which has to be backported. (perhaps in line with the re-mini https://github.com/cms-sw/cmssw/pull/27929#issuecomment-547823344)

peruzzim:
My understanding, after chatting with a few people:

- this change does not affect NanoAOD, where (at the moment) we do not rerun PUPPI in any release;
- also DeepTau scores in NanoAOD are, therefore, not affected;
- it would affect MiniAOD output in production if backported, because there is a technical change in the algorithm;
- there is a planned further update of the algorithm (the PUPPI tune) that is 
expected to come maybe in time for the final re-miniAOD of the UL. This also connects, I think, with the switch to PUPPI jets for Run-3, for which we will have to discuss what to leave as "legacy" in the Run-2 UL datasets.

On the basis of this, I would propose that:

- this gets merged in master, as an algorithm update (that has the positive effect of also simplifying the maintenance of this module);
- we treat the PUPPI output as the result of a new algorithm, and specifically:
- as it seems impractical to backport this under the protection of the `run2_miniAOD_devel` modifier, we consider the backport to 10_6_X only when the PUPPI tune also gets updated, and will be switched "by release version" for UL re-miniAOD. Please note this in #27889!

peruzzim:
I would also recommend, for clarity, to change the title of this PR to make it clear that it's an update of the algorithm and not just a technical simplification of the code.

santocch:
+1

fabiocos:
+1

fabiocos:
the title now seems clear enough to explain an update in the algorithm

PR #27935, opened 2019-09-05
[HGCal] dEdx weights and plots

apsallid:
#### PR description:

In this PR we (@rovere, @felicepantaleo) work in the following two items: 
1. Extend the SimTracker/TrackerMaterialAnalysis package to provide support for HGCal apart from Tracker.
2. Every time a new HGCal geometry is produced, we calculate the dEdx weights (more info e.g [here](https://indico.cern.ch/event/810242/contributions/3377477/attachments/1823243/2982972/dEdxWeightsandHitCalib_03April2018.pdf) ). The basic script that performs the calculation needs an input txt file that is produced by the SimTracker/TrackerMaterialAnalysis package.

Until now, these items were on a personal branch, however the frequent procedure of the rechit calibration every time a new version of the HGCal geometry appears, dictates that we put all code accessible by everyone and write specific instructions (will be [hgcal website](http://hgcal.web.cern.ch/hgcal/HitCalibration/hitCalibration/)).

#### PR validation:

Both the above tasks are done with the goal not to mess with the Tracker workflow. That's why, apart from the HGCal workflow, the Tracker one was run. No conflicts were observed. 

This is not a backport.


civanch:
please test

SimTracker/TrackerMaterialAnalysis/plugins/MaterialAccountingGroup.cc
```diff
@@ -30,7 +30,17 @@ MaterialAccountingGroup::MaterialAccountingGroup(const std::string& name, const
...
       m_file(nullptr) {
   // retrieve the elements from DDD
-  DDSpecificsMatchesValueFilter filter{DDValue("TrackingMaterialGroup", name)};
+  DDValue namevalue;
+  if (TString(name.c_str()).Contains("Tracker")) {
+    namevalue = DDValue("TrackingMaterialGroup", name);
+  } else if (TString(name.c_str()).Contains("HGCal")) {
+    namevalue = DDValue("Volume", name);
+  } else {
+    std::cout << name << std::endl;
```
civanch: cout and cerr should be substituted by MessageLogger

felicepantaleo:
please test

civanch:
+1

fabiocos:
@smuzaffar the data file in the test subdirectory is quite large, 176 kB, but not larger than many xml geometry descriptions. @apsallid are you planning to have more of these files? BTW why is it put in test and not in data?

apsallid:
@fabiocos Really sorry what file are you referring to? We do not use any xml file for HGCal. This should be used by the Tracker.

rovere:
I share @apsallid's arguments.

fabiocos:
as the jupiter notebook is ultimately an ascii file I can imagine it is compressed in a quite efficient way within git (@smuzaffar), so at the end of the day the impact on the repository size is probably not that big, as in the case of large xml files. I see the advantage of keeping things in a same place without packages outside CMSSW, although many communities have their own maintained repositories for this, but of course along this argument one might merge many other things. As far as I can see this would be the second jupiter notebook within the repository. 

Ok for me to merge it, although if it is a utility may be the test subdirectory is not the best place where to store it

fabiocos:
+1

PR #27947, opened 2019-09-06
ME0TriggerPseudoBuilder  mistook GEM strip pitch unit as radiu and this commit is to fix it

tahuang1991:
#### PR description:
pitch() from GEM geometry returns the strip pitch in cm and was supposed to be converted into radius before it is compared with deltaphi. JiaFu found this bug and  this PR is to fix this bug in  L1Trigger/ME0Trigger/src/ME0TriggerPseudoBuilder.cc

This PR is independent from Sven's previous PR to update Me0 trigger built from GEM trigger pads.   This PR is to fix a old bug in  ME0 pseudo-triggers  which is actually used by EMTF .

#### PR validation:
I ran test with step 20034.0 and everything seems right .   
Summary from test 20034.0:
1 1 1 1 tests passed, 0 0 0 0 failed

FYI: @jshlee @dildick  @rekovic @jiafulow

### DIFF
--- a/L1Trigger/ME0Trigger/src/ME0TriggerPseudoBuilder.cc
+++ b/L1Trigger/ME0Trigger/src/ME0TriggerPseudoBuilder.cc
@@ -106,7 +106,7 @@ ME0TriggerDigi ME0TriggerPseudoBuilder::segmentConversion(const ME0Segment segme
   float centreOfStrip = istrip + 0.25 + phiposition2 * 0.5;
   GlobalPoint gp_digi = etapart->toGlobal(etapart->centreOfStrip(centreOfStrip));
 
-  float strippitch_rad = strippitch/gp.perp(); //unit in rad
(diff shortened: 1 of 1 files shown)


kpedro88:
please test

dildick:
@tahuang1991 While you are making this change, I suggest that you hook up the corresponding producer to the `me0TriggerPseudoDigiTask` task as `me0TriggerConvertedSegments` - to avoid a naming clash. So the new task would look like:

```
me0TriggerPseudoDigiTask = cms.Task(
    simMuonME0PseudoReDigisCoarse,
    me0RecHitsCoarse,
    me0TriggerPseudoDigis,
    me0TriggerConvertedSegments
)
```

rekovic:
please test

kpedro88:
+upgrade

rekovic:
+1

fabiocos:
+1

PR #26592, opened 2019-05-01
Implement core support for concurrent IOVs

wddgit:
PR description:

This implements core support for concurrent IOVs. More work needs to be done before CMSSW is actually running concurrent IOVs. Outside of core tests, there are not any ESSources that have been modified to actually support this kind of concurrency. The core will continue to support and properly run existing ESSources and ESProducers, but the IOVs will not be processed concurrently.

The only changes outside the core are to CondDBESSource, FWLiteESSource, and FWLiteESRecordWriterAnalyzer. These changes are the minimal changes necessary to work with the revised core code. They do not migrate these modules to support concurrent IOVs.

There are two independent ways to control the number of concurrent IOVs:

First one can set the number of IOVs allowed to be processed at the same time for each EventSetupRecord as follows. In the configuration:
```
process.options = dict(
   ...
   eventSetup = dict(
        numberOfConcurrentIOVs = 4,
        forceNumberOfConcurrentIOVs = dict(
            SomeRecord = 4
        )
    )
)
```

In the C++ definition of an EventSetupRecord type:
```
  static constexpr bool allowConcurrentIOVs_ = false;
```

numberOfConcurrentIOVs sets the number of IOVs for all types of EventSetupRecord (defaults to 1). The boolean defined in the C++ class overrides this for a specific EventSetupRecord type (by default this is defined true in the base class). The value in forceNumberOfConcurrentIOVs overrides both numberOfConcurrentIOVs and the value in the C++ class for specified EventSetupRecord types (by default this parameter set is empty).

The second way of controlling this relates to ESSources. Existing ESSources often rely on the time ordering of calls to setIntervalFor and the associated functions that produce data for that interval. In the past the produce function always referred to the immediately preceding setIntervalFor function call. But with concurrent IOVs, this can no longer be true. The Framework will enforce the old time ordering of function calls for existing ESSources. As ESSources are modified to support concurrent IOVs, the following function should be added to ESSource C++ class definition:
```
    bool isConcurrentFinder() const override { return true; }
```

In many but not all cases, existing ESProducers will work in concurrent IOV mode without modification. Most ESSources are going to require work to migrate. We are already looking at CondDBESSource (PoolDBESSource) to see what changes are needed there.

Note that the two mechanisms described above are independent of each other.

If numberOfConcurrentIOVs is set to greater than one and nothing else is done, things might fail horribly. It will take a record by record review of the associated/dependent ESSource and ESProducer code to determine what will and will not function properly with more than one concurrent IOV.

[...]

PR #27948, opened 2019-09-06
Core support for concurrent IOVs in CondDBESSource

wddgit:
#### PR description:

Add to Core support needed for CondDBESSource so that
it can be upgraded for concurrent IOVs. These are requirements
I found as I actually implemented the code to migrate CondDBESSource
to support concurrent IOVs. They are in addition to the changes recently
merged with PR #26592. There are 6 separate commits, each a separate
feature. The first 3 will only be needed by CondDBESSource. The second
two are extra checks for pathological cases, useful in general. The last a
new unit test.

1. Notify a DataProxy when a new IOV actually begins.
2. During job initialization explicitly inform how many concurrent
IOVs will be allowed for each record.
3. Include setIntervalFor calls in the global mutex already used for
DataProxy calls producing data.
4. More carefully reset resources in the LuminosityBlockProcessingStatus
object.
5. Don't allow the number of concurrent IOVs to be greater than the number
of streams or concurrent lumis.
6. Add a new unit test for concurrent IOVs particularly covering new features
listed above.

Note the PR migrating CondDBESSource is ready and I intend to submit
it immediately after this PR is merged.

#### PR validation:

New unit test. Note that until the future PR implementing concurrentIOVs
for CondDBESSource is submitted, this should not cause any noticeable
change in behavior.

Not a backport.

### DIFF
--- a/FWCore/Framework/interface/EventSetupRecordImpl.h
+++ b/FWCore/Framework/interface/EventSetupRecordImpl.h
@@ -122,7 +122,7 @@ namespace edm {
       void clearProxies();
 
       ///Set the cache identifier and validity interval when starting a new IOV
-      void initializeForNewIOV(unsigned long long iCacheIdentifier, ValidityInterval const&);
+      void initializeForNewIOV(unsigned long long iCacheIdentifier, ValidityInterval const&, bool hasFinder);
 
       /**Set the validity interval in a thread safe way. This is used when the
          IOV is already in use and the end of the IOV needs to be updated but
--- a/FWCore/Framework/interface/EventSetupRecordImpl.h
+++ b/FWCore/Framework/interface/EventSetupRecordImpl.h
@@ -122,6 +122,10 @@ namespace edm {
       void clearProxies();
 
       ///Set the cache identifier and validity interval when starting a new IOV
+      ///In addition, also notify the DataProxy's a new IOV is starting.
+      ///(As a performance optimization, we only notify the DataProxy's if hasFinder
+      ///is true. At the current time, the CondDBESSource DataProxy's are the only
+      ///ones who need to know about this and they always have finders).
       void initializeForNewIOV(unsigned long long iCacheIdentifier, ValidityInterval const&, bool hasFinder);
 
       /**Set the validity interval in a thread safe way. This is used when the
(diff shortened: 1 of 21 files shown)


wddgit:
please test

wddgit:
abort test

wddgit:
please test

FWCore/Framework/interface/EventSetupRecordImpl.h
```diff
@@ -122,7 +122,7 @@ namespace edm {
       void clearProxies();
 
       ///Set the cache identifier and validity interval when starting a new IOV
```
Dr15Jones: please update the comment to say what `hasFinder` does.

FWCore/Framework/src/NumberOfConcurrentIOVs.h
```diff
@@ -38,6 +38,8 @@ namespace edm {
 
       void initialize(ParameterSet const* eventSetupPset);
 
+      void initialize(unsigned int nStreams, unsigned int nConcurrentLumis);
```
Dr15Jones: Having three functions named `initialize` with no documentation as to what they are used for does not seem like a good idea to me.

FWCore/Framework/src/NumberOfConcurrentIOVs.cc
```diff
@@ -35,6 +35,13 @@ namespace edm {
       }
     }
 
+    void NumberOfConcurrentIOVs::initialize(unsigned int nStreams, unsigned int nConcurrentLumis) {
+      maxConcurrentIOVs_ = nStreams;
+      if (maxConcurrentIOVs_ > nConcurrentLumis) {
+        maxConcurrentIOVs_ = nConcurrentLumis;
```
Dr15Jones: Should we print a message saying the max was changed?

Dr15Jones (changes requested):

FWCore/Framework/src/NumberOfConcurrentIOVs.cc
wddgit: I added a LogInfo message when the number of streams or concurrent lumis restricts the number of concurrent IOVs below what was configured.

wddgit:
please test

I responded to all the comments and pushed in commits with fixes/improvements. Thanks for the review.

FWCore/Framework/src/NumberOfConcurrentIOVs.h
wddgit: I renamed the functions. That is better, thanks.

FWCore/Framework/interface/EventSetupRecordImpl.h
wddgit: I added a comment.

Dr15Jones:
+1

PR #27957, opened 2019-09-09
Update GE2/1 LUTs for GE2/1-ME2/1 integrated local trigger

dildick:
#### PR description:

This PR updates a few LUTs for GE2/1-ME2/1 in the Phase-2 CSC local trigger. GE2/1 geometry changed some time ago from 12 to 8 eta partitions. The out-of-data LUTs caused significant drops in efficiency for the GE2/1-ME2/1 integrated trigger stubs.

#### PR validation:

I tested this PR with a Phase-2 PU0 single muon sample to make sure that ALCTs were properly matched to GEM pads.


rekovic:
please test

dildick:
@rekovic Do you have any more comments?

rekovic:
+1

fabiocos:
@dildick @rekovic as far as I understand this is supposed to affect *only* Phase2, a not previous setups, am I correct? The PR test does not show any change, I may guess because of the limited statistics

dildick:
@fabiocos Correct. This only affects Phase-2.

I think there would have to be muons flying exactly into those regions of ME2/1 where the LCT efficiency is reduced due to a lower high-voltage to see an effect. Because that is where Cathode LCTs are typically lost. If I'm not mistaken, this reduced high-voltage region is about 10-15% of ME21/, corresponding to a delta eta between 0.1 and 0.2. The tests are limited in statistics. I tested these changes on several thousand muons with |eta| between 1.6 and 2.4.

fabiocos:
+1

PR #27958, opened 2019-09-09
Fix mix-up between the physical and the trigger chamber number in upgrade CSC motherboard

dildick:
#### PR description:

This PR fixes a bug in the initialization of the GEMCoPadProcessor. In the constructor of the base class `CSCGEMMotherboard` the variable `chamber` is used to reset the processor. This needs to be  `theChamber` instead which refers to the physical chamber number, not the trigger chamber number. Without the fix the upgrade motherboards do not produce coincidence pads.

#### PR validation:

I tested this fix with workflow 22034.0.

@tahuang1991

### DIFF
--- a/L1Trigger/CSCTriggerPrimitives/interface/GEMCoPadProcessor.h
+++ b/L1Trigger/CSCTriggerPrimitives/interface/GEMCoPadProcessor.h
@@ -17,7 +17,7 @@
 class GEMCoPadProcessor {
 public:
   /** Normal constructor. */
-  GEMCoPadProcessor(unsigned endcap, unsigned station, unsigned chamber, const edm::ParameterSet& copad);
+  GEMCoPadProcessor(unsigned region, unsigned station, unsigned chamber, const edm::ParameterSet& copad);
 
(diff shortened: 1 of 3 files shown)


rekovic:
please test

rekovic:
+1

rekovic:
Hi @dildick 
We will need a backport of this to 10_6_1_patch2, that is to say a PR that can be used with `cms-l1t-offline:phase2-l1t-integration-CMSSW_10_6_1_patch2`.  Thanks.

fabiocos:
@dildick @rekovic are the changes in the DQM expected? They affect GEM plots, so I would naively they are coherent with this PR, but I would like to ensure that they are really as expected.

dildick:
@rekovic I'll provide back-ports of all recent PRs.

@fabiocos I do see a difference in the GEM copad validation plots for station 1. It's hard to judge from the GEM DQM plots since the plots are not very clear themselves. Before submitting the PR I also tested the copad production on a Phase-2 PU0 single muon gun.

fabiocos:
+1

PR #27960, opened 2019-09-10
Comments to parameters definition for HFStripFilter

toropin:
#### PR description:

In this PR only comments were added to the definition of parameters for HFStripFilter in the
HFPhase1Reconstructor_cfi.py. 
The decision to add comments was made at the following meeting to make the meaning of parameters more clear:
https://indico.cern.ch/event/845652/contributions/3555228/attachments/1903523/3142982/HFStripFilter_Update_6Sep2019.pdf

#### PR validation:

After these changes the code was compiled and all functionality was checked one more time.


perrotta:
@toropin, updates should be integrated first in the master release, and only then backported to the old ones (as 10_6_X here): these comments will get lost if they only remain in an old cmssw release

Moreover, the more natural way of adding these comments is to include them already in the HFStripFilter::fillDescription() method: please also do so.

PR #27961, opened 2019-09-10
Support concurrent LuminosityBlocks in HLTrigReport

Dr15Jones:
#### PR description:
Switched to using a LuminosityBlockCache to allow HLTrigReport to not be sensitive to the ordering of completion of a LuminosityBlock. This allows concurrent LuminosityBlocks to be processed properly.

Support for concurrent Runs would require further work.

Added unit tests to be sure the behavior remains the same.

This change is needed to allow workflows containing this module to be able to use concurrent LuminosityBlocks. This includes the new workflow which is testing such concurrency, 1631.181.
#### PR validation:

The new unit test passes.

### DIFF
--- a/HLTrigger/HLTanalyzers/test/test_hltrigreport_lumi_job.py
+++ b/HLTrigger/HLTanalyzers/test/test_hltrigreport_lumi_job.py
@@ -1,0 +1,4 @@
+from test_hltrigreport_base_cfg import process
+
+process.hlTrigReport.resetBy = "lumi"
+process.hlTrigReport.reportBy = "job"
--- a/HLTrigger/HLTanalyzers/test/test_hltrigreport_run_job.py
+++ b/HLTrigger/HLTanalyzers/test/test_hltrigreport_run_job.py
@@ -1,0 +1,4 @@
+from test_hltrigreport_base_cfg import process
+
+process.hlTrigReport.resetBy = "run"
+process.hlTrigReport.reportBy = "job"
--- a/HLTrigger/HLTanalyzers/plugins/HLTrigReport.cc
+++ b/HLTrigger/HLTanalyzers/plugins/HLTrigReport.cc
@@ -44,11 +44,8 @@ HLTrigReport::ReportEvery HLTrigReport::decode(const std::string& value) {
 //
 // constructors and destructor
 //
-HLTrigReport::HLTrigReport(const edm::ParameterSet& iConfig)
-    : hlTriggerResults_(iConfig.getParameter<edm::InputTag>("HLTriggerResults")),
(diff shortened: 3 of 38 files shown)


Dr15Jones:
please test

fabiocos:
@Dr15Jones running 1361.181 with "-t 4" I still see:

```
%MSG-s ModulesSynchingOnLumis:  AfterModConstruction 11-Sep-2019 10:50:36 CEST  pre-events
The following modules require synchronizing on LuminosityBlock boundaries:
  HLTrigReport hltTrigReport
%MSG
```

Do you confirm that this is the expected behaviour?

HLTrigger/HLTanalyzers/test/test_hltrigreport_run_job.py
```diff
@@ -0,0 +1,4 @@
+from test_hltrigreport_base_cfg import process
+
+process.hlTrigReport.resetBy = "run"
+process.hlTrigReport.reportBy = "job"
```
fwyzard: does it really make sense to reset every run and report only at the end of the job ?

HLTrigger/HLTanalyzers/test/test_hltrigreport_lumi_job.py
```diff
@@ -0,0 +1,4 @@
+from test_hltrigreport_base_cfg import process
+
+process.hlTrigReport.resetBy = "lumi"
+process.hlTrigReport.reportBy = "job"
```
fwyzard: does it really make sense to reset every lumi and report only at the end of the job ?

Dr15Jones: It would give you the last lumi of the job.
Again, it was what the original coe supported so I kept the same behavior.

HLTrigger/HLTanalyzers/test/test_hltrigreport_run_job.py
Dr15Jones: It would give you the last run of the job.
It was what the original code supported so I did the same.

fwyzard: OK... IIRC the possibility was just an accident in the original code, not an intended feature.
Of course, having it doesn't harm, as long as it didn't require extra work to reimplement.

Dr15Jones: > Of course, having it doesn't harm, as long as it didn't require extra work to reimplement.

It actually does require additional logic in order to implement. But it is there and working.

fabiocos:
+1

PR #27963, opened 2019-09-10
Explicitly define move related ReferenceCounted methods

Dr15Jones:
#### PR description:

The clang compiler warned that the '= default' version were not begin generated by the compiler. Fixed via explicit versions.

#### PR validation:

Compiled using CMSSW_11_0_CLANG_X_2019-09-09-2300 and the clang compiler warnings are gone.


Dr15Jones:
code-checks

Dr15Jones:
please test

PR #27981, opened 2019-09-11
MVAComputerContainer::find returns pointer instead of reference

Dr15Jones:
#### PR description:

PhysicsTools::MVATrainerContainer inherits from MVAComputerContainer and it was returning a reference to null in its overriden find method. This was leading to unit test crashes under clang since clang (correctly) optimized away a check for null on the address of the returned reference.

All uses of find were updated to the new interface.

#### PR validation:

The original code caused the unit test in TopQuarkAnalysis/TopEventSelection to crash in CMSSW_11_0_CLANG_X_2019-09-10-2300 when optimizations were used. With this change the unit tests run without crashing.

### DIFF
--- a/CondFormats/PhysicsToolsObjects/interface/MVAComputer.h
+++ b/CondFormats/PhysicsToolsObjects/interface/MVAComputer.h
@@ -255,7 +255,7 @@ namespace PhysicsTools {
       virtual ~MVAComputerContainer() {}
 
       MVAComputer &add(const std::string &label);
-      virtual const MVAComputer &find(const std::string &label) const;
+      virtual const MVAComputer *find(const std::string &label) const;
       virtual bool contains(const std::string &label) const;
 
       // cacheId stuff to detect changes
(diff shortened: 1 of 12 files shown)


slava77:
@Dr15Jones 
it looks like we are going to go ahead with #27888

Dr15Jones:
@slava77 I had completely forgotten about that one. Oh well.

perrotta:
please test

Dr15Jones:
@perrotta removing `MVATrainerContainer` does make this pull request unnecessary. It was `MVATrainerContainer` which 'broke' the interface to the base class `MVAComputerContainer::find` method by being able to return a reference to a nullptr. The base class implemenation does not have that fault.

perrotta:
> @perrotta removing `MVATrainerContainer` does make this pull request unnecessary. It was `MVATrainerContainer` which 'broke' the interface to the base class `MVAComputerContainer::find` method by being able to return a reference to a nullptr. The base class implemenation does not have that fault.

Thank you @Dr15Jones 
I just wanted to be sure

Dr15Jones:
@perrotta it was good you asked since I hadn't done the full analysis until this morning.

PR #27983, opened 2019-09-12
Squash all Patatrack developments on top of CMSSW_11_3_0_pre5

fwyzard:
This PR includes in a single commit all the Patatrack developments from the past years.

There are few ongoing issues, but this is good enough to start making the actual PRs.

I would ask all affected L2's to have a look at the changes in their packages, and suggest how they would like them to be split and organised.

Of course, any comments on the code itself are welcome !


makortel:
I'm fixing the clang warnings.

slava77:
> However, I would ask all affected L2's to have a look at the changes in theiir packages, and suggest how they would like them to be split and organised.

My guess for a good order would be to be maximally orthogonal
- the core
- data formats and producers split by DPG/POG areas with individual test workflows available (sub-splitting may be needed for larger cases) fully factorized cases
- cross-area cases which do not fit the above

Some simpler cases which could establish a common coding pattern better go first (after the core) so that we can settle on acceptable approaches in an easier way.

Somewhat of a guess, at least for the reco-related code, I'd put about 1K lines of non-boilerplate code per week as an expected rate of integration.

CUDADataFormats/TrackingRecHit/interface/TrackingRecHit2DSOAView.h
```diff
@@ -0,0 +1,101 @@
+#ifndef CUDADataFormats_TrackingRecHit_interface_TrackingRecHit2DSOAView_h
+#define CUDADataFormats_TrackingRecHit_interface_TrackingRecHit2DSOAView_h
+
+#include <cuda_runtime.h>
+#include <cuda/api_wrappers.h>
+
+#include "CUDADataFormats/SiPixelCluster/interface/gpuClusteringConstants.h"
+#include "HeterogeneousCore/CUDAUtilities/interface/HistoContainer.h"
+#include "HeterogeneousCore/CUDAUtilities/interface/cudaCompat.h"
+#include "Geometry/TrackerGeometryBuilder/interface/phase1PixelTopology.h"
```
ianna: this one should be moved to ```Geometry/CommonTopology``` package

fwyzard: @ianna what file do you mean ?
`CUDADataFormats/TrackingRecHit/interface/TrackingRecHit2DSOAView.h` or `Geometry/TrackerGeometryBuilder/interface/phase1PixelTopology.h` ?

I would be hesitant to move the first one, because of the dependency on CUDA.

We can certainly move the second one - but it's not a new file introduced by this PR, it is already present in release.

Configuration/PyReleaseValidation/python/upgradeWorkflowComponents.py
```diff
@@ -145,6 +145,34 @@
     'suffix' : '_pixelTrackingOnly',
     'offset' : 0.5,
 }
+upgradeSteps['pixelTrackingOnlyNtupleFit'] = {
```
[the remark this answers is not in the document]
fwyzard: I know, but I'm no longer surprised by the level of support that the Patatrack effort is receiving.

I.e., none whatesover.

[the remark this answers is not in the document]
kpedro88: @fwyzard my proposal was intended as a way to prevent exactly this occurrence. Nobody from the patatrack side seemed to be interested. I assumed if it was important that someone would make a PR.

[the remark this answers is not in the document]
fwyzard: OK, thank I will try to make a PR once #28281 is merged.

CUDADataFormats/TrackingRecHit/interface/TrackingRecHit2DSOAView.h
fwyzard: @ianna could you remind me what you meant ?

Configuration/PyReleaseValidation/python/upgradeWorkflowComponents.py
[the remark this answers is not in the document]
fwyzard: @kpedro88 I see that #28281 was merged. Do you know if it entered pre11 ?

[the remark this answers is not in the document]
fabiocos: @fwyzard no, this came after pre11 https://cmssdt.cern.ch/SDT/ReleaseNotes/CMSSW_11/CMSSW_11_0_0_pre11.html 
it will be in pre12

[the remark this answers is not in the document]
fwyzard: Understood, thanks.

fwyzard:
Is there a way to ask @cmsbot to test this on top of plain CMSSW_11_0_0_pre11 instead of the current CMSSW_11_0_X and the latest IBs ?

fwyzard:
@smuzaffar there seems something wrong with the result of the code checks ?
Looks like https://cmssdt.cern.ch/SDT/code-checks/cms-sw-PR-27983/12618/code-checks.patch includes also some unrelated diff ?

smuzaffar:
@fwyzard , can you please resolve the conflicts?

fwyzard:
Honestly, no, I cannot - they will be resolved once pre12 is out and we rebase the Patatrack development on top of pre12.
Building releases ourselves is complicated and time consuming enough, I am definitely not going to build IBs...

fwyzard:
`scram b code-checks` [reports](https://cmssdt.cern.ch/SDT/code-checks/cms-sw-PR-27983/12961/code-checks.patch), among others
```patch
diff --git a/CUDADataFormats/TrackingRecHit/interface/TrackingRecHit2DHeterogeneous.h b/CUDADataFormats/TrackingRecHit/interface/TrackingRecHit2DHeterogeneous.h
index 1b27558efe3..b1d51f5b457 100644
--- a/CUDADataFormats/TrackingRecHit/interface/TrackingRecHit2DHeterogeneous.h
+++ b/CUDADataFormats/TrackingRecHit/interface/TrackingRecHit2DHeterogeneous.h
@@ -91,7 +91,7 @@ TrackingRecHit2DHeterogeneous<Traits>::TrackingRecHit2DHeterogeneous(uint32_t nH
         (std::is_same<Traits, cudaCompat::GPUTraits>::value) {
       cudautils::copyAsync(m_view, view, stream);
     } else {
-      m_view.reset(view.release());
+      m_view = std::move(view);
     }
     return;
   }
@@ -138,7 +138,7 @@ TrackingRecHit2DHeterogeneous<Traits>::TrackingRecHit2DHeterogeneous(uint32_t nH
       (std::is_same<Traits, cudaCompat::GPUTraits>::value) {
     cudautils::copyAsync(m_view, view, stream);
   } else {
-    m_view.reset(view.release());
+    m_view = std::move(view);
   }
 }
 
```

However, the suggested version does not actually compile:
```
>> Compiling  /data/user/fwyzard/patatrack/CMSSW_11_0_0_pre12/src/RecoLocalTracker/SiPixelRecHits/plugins/PixelRecHits.cu 
/data/user/fwyzard/patatrack/CMSSW_11_0_0_pre12/src/CUDADataFormats/TrackingRecHit/interface/TrackingRecHit2DHeterogeneous.h(94): error: no operator "=" matches these operands
            operand types are: cudaCompat::GPUTraits::unique_ptr<TrackingRecHit2DSOAView> = cudautils::host::unique_ptr<TrackingRecHit2DSOAView>
          detected during instantiation of "TrackingRecHit2DHeterogeneous<Traits>::TrackingRecHit2DHeterogeneous(uint32_t, const pixelCPEforGPU::ParamsOnGPU *, const uint32_t *, cudaStream_t) [with Traits=cudaCompat::GPUTraits]" 
/data/user/fwyzard/patatrack/CMSSW_11_0_0_pre12/src/RecoLocalTracker/SiPixelRecHits/plugins/PixelRecHits.cu(43): here

/data/user/fwyzard/patatrack/CMSSW_11_0_0_pre12/src/CUDADataFormats/TrackingRecHit/interface/TrackingRecHit2DHeterogeneous.h(141): error: no operator "=" matches these operands
            operand types are: cudaCompat::GPUTraits::unique_ptr<TrackingRecHit2DSOAView> = cudautils::host::unique_ptr<TrackingRecHit2DSOAView>
          detected during instantiation of "TrackingRecHit2DHeterogeneous<Traits>::TrackingRecHit2DHeterogeneous(uint32_t, const pixelCPEforGPU::ParamsOnGPU *, const uint32_t *, cudaStream_t) [with Traits=cudaCompat::GPUTraits]" 
/data/user/fwyzard/patatrack/CMSSW_11_0_0_pre12/src/RecoLocalTracker/SiPixelRecHits/plugins/PixelRecHits.cu(43): here

2 errors detected in the compilation of "/tmp/tmpxft_00024f87_00000000-8_PixelRecHits.compute_70.cpp1.ii".
```

How do I tell `clang-tidy`to back off ?

fwyzard:
Looks like one can use a `// NOLINT` comment for that.

fwyzard:
@cmsbuild please check

fwyzard:
@smuzaffar can you remind how to request @cmsbot to run the GPU tests ?

smuzaffar:
enable gpu
this above message should now additionally run GPU tests too.  Currently we have https://github.com/cms-sw/cms-bot/blob/master/cmssw-pr-test-config#L2 workflows set for GPUs

fwyzard:
@cmsbuild, please test

fwyzard:
@cmsbuild, please test

RecoLocalTracker/SiPixelRecHits/plugins/PixelRecHits.cu
```diff
@@ -0,0 +1,86 @@
...
+#endif
+
+    // assuming full warp of threads is better than a smaller number...
+    if (nHits) {
+      setHitsLayerStart<<<1, 32, 0, stream>>>(clusters_d.clusModuleStart(), cpeParams, hits_d.hitsLayerStart());
+      cudaCheck(cudaGetLastError());
+    }
+
+    if (nHits >= TrackingRecHit2DSOAView::maxHits()) {
+      edm::LogWarning("PixelRecHitGPUKernel")
```
fwyzard: Looks like the MessageLogger is now using c++17 features (namely, `std::string_view`) so we cannot use in a .cu file.

@makortel @VinInn what do you think ?
Should we keep only the kernels in a .cu file, and move the host code to a .cc file ?

HeterogeneousCore/CUDATest/plugins/TestCUDAProducerGPUKernel.cu
```diff
@@ -0,0 +1,131 @@
...
+
+  cudaCheck(cudaMemcpyAsync(d_a.get(), h_a.get(), NUM_VALUES * sizeof(float), cudaMemcpyHostToDevice, stream));
+  cudaCheck(cudaMemcpyAsync(d_b.get(), h_b.get(), NUM_VALUES * sizeof(float), cudaMemcpyHostToDevice, stream));
+
+  int threadsPerBlock{32};
+  int blocksPerGrid = (NUM_VALUES + threadsPerBlock - 1) / threadsPerBlock;
+
+  auto d_c = cudautils::make_device_unique<float[]>(NUM_VALUES, stream);
+  auto current_device = cudautils::currentDevice();
+  edm::LogVerbatim("TestHeterogeneousEDProducerGPU")
```
fwyzard: Looks like the MessageLogger is now using c++17 features (namely, `std::string_view`) so we cannot use in a .cu file.

@makortel @VinInn what do you think ?
Should we keep only the kernels in a .cu file, and move the host code to a .cc file ?

RecoLocalTracker/SiPixelRecHits/plugins/PixelCPEFastESProducer.cc
```diff
@@ -0,0 +1,102 @@
...
+  return std::make_unique<PixelCPEFast>(pset_,
+                                        &iRecord.get(magfieldToken_),
+                                        iRecord.get(pDDToken_),
+                                        iRecord.get(hTTToken_),
+                                        &iRecord.get(lorentzAngleToken_),
+                                        genErrorDBObjectProduct,
+                                        lorentzAngleWidthProduct);
+}
+
+void PixelCPEFastESProducer::fillDescriptions(edm::ConfigurationDescriptions& descriptions) {
```
fwyzard: The existing CPE ESProducers use inheritance to define the configuration in steps, but I am not expert enough in how the original code is structured, so I implemented a stand-alone `fillDescriptions()` method.

To be reviewed by Tracker DPG along with the rest ...

CUDADataFormats/BeamSpot/src/BeamSpotCUDA.cc
```diff
@@ -0,0 +1,9 @@
+#include "CUDADataFormats/BeamSpot/interface/BeamSpotCUDA.h"
+
+#include "HeterogeneousCore/CUDAUtilities/interface/cudaCheck.h"
+#include "HeterogeneousCore/CUDAUtilities/interface/device_unique_ptr.h"
+
+BeamSpotCUDA::BeamSpotCUDA(Data const* data_h, cudaStream_t stream) {
+  data_d_ = cudautils::make_device_unique<Data>(stream);
+  cudaCheck(cudaMemcpyAsync(data_d_.get(), data_h, sizeof(Data), cudaMemcpyHostToDevice, stream));
```
fwyzard: @makortel @VinInn could you remind me why we have the `cudaMemcpyAsync` in the dataformat itself, rather than in the relevant EDProducer ?

RecoLocalTracker/SiPixelRecHits/plugins/PixelRecHits.cu
VinInn: I would say remove the warning

fwyzard: I've coded a possible workaround in https://github.com/cms-patatrack/cmssw/pull/421 .

RecoLocalTracker/SiPixelRecHits/plugins/PixelCPEFastESProducer.cc
mmusich: This class looks a copy paste of https://github.com/cms-sw/cmssw/blob/master/RecoLocalTracker/SiPixelRecHits/plugins/PixelCPEGenericESProducer.cc.
You can get rid of some of these using a call to ` PixelCPEBase::fillPSetDescription(desc);`.
For the algorithm-specific ones, we implemented https://github.com/cms-sw/cmssw/blob/91be707d4e28bfd3b85525a5feae0e36b0f39181/RecoLocalTracker/SiPixelRecHits/src/PixelCPEGeneric.cc#L748
I believe the same should be applied to the `PixelCPEFast` class.

RecoLocalTracker/SiPixelRecHits/src/PixelCPEFast.cc
```diff
@@ -0,0 +1,585 @@
...
+  if (UseErrorsFromTemplates_) {
+    if (!SiPixelGenError::pushfile(*genErrorDBObject_, thePixelGenError_))
+      throw cms::Exception("InvalidCalibrationLoaded")
+          << "ERROR: GenErrors not filled correctly. Check the sqlite file. Using SiPixelTemplateDBObject version "
+          << (*genErrorDBObject_).version();
+  }
+
+  // Rechit errors in case other, more correct, errors are not vailable
+  // This are constants. Maybe there is a more efficienct way to store them.
+  xerr_barrel_l1_ = {0.00115, 0.00120, 0.00088};
```
mmusich: from where does this come from!?

mmusich: answering to myself, from here: https://github.com/cms-sw/cmssw/blob/master/RecoLocalTracker/SiPixelRecHits/src/PixelCPEGeneric.cc#L101. 
Not sure this still applies anyway in both places.

CUDADataFormats/TrackingRecHit/interface/TrackingRecHit2DHeterogeneous.h
```diff
@@ -0,0 +1,150 @@
...
+  if (0 == nHits) {
+    if
+#ifndef __CUDACC__
+        constexpr
+#endif
+        (std::is_same<Traits, cudaCompat::GPUTraits>::value) {
+      cudautils::copyAsync(m_view, view, stream);
+    } else {
+      m_view.reset(view.release());  // NOLINT: std::move() breaks CUDA version
+    }
```
makortel: We could also use older "dialect" of metaprogramming to never expose the "else" branch for the CUDA version.

This makes me realize that the `reset()`+`release()` with the host-pinned and device `unique_ptr's` is dangerous because of the custom deleters (e.g. this kind of code would make the host memory to be released with `cudaFree()` leading to an exception to be thrown there), and could be prevented by not providing such functions.

CUDADataFormats/BeamSpot/src/BeamSpotCUDA.cc
makortel: The division of work between the EDProducer and the data format class is somewhat arbitrary. This way the details of the (device) memory allocations are fully controlled by the data format class, and the object can be constructed+inserted into the event within `Event::emplace()`.

fwyzard:
@cmsbuild, please test

fwyzard:
@smuzaffar , is there a way to ask @cmsbot to run additional runTheMatrix.py workflows on the GPU machine ?

smuzaffar:
@fwyzard , `please test workflow wfnum,fwnum` should allow to run extra tests too. For this PR, GPU tests are not going to finish as we are running them on condor batch nodes where we only have 1 CPU. So compiling will take many hours and might not finish in time.

Configuration/PyReleaseValidation/python/upgradeWorkflowComponents.py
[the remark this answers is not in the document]
fwyzard: The placeholder for the Patatrack workflows is implemented by #28522 .

fwyzard:
> @fwyzard , `please test workflow wfnum,fwnum` should allow to run extra tests too.

Does that apply to the GPU machines, too ?

smuzaffar:
yes

smuzaffar:
@fwyzard , about https://github.com/cms-sw/cmssw/pull/27983#issuecomment-559752617 , I think the problem is the `-std=c++14` flag we pass to `nvcc` . clang-tidy suggestion works with `-std=c++1z`

fwyzard:
> @fwyzard , about [#27983 (comment)](https://github.com/cms-sw/cmssw/pull/27983#issuecomment-559752617) , I think the problem is the `-std=c++14` flag we pass to `nvcc` . clang-tidy suggestion works with `-std=c++1z`

Unfortunately `nvcc` supports only c++14.

fwyzard:
@cmsbuild, please test

smuzaffar:
please ignore the commnet https://github.com/cms-sw/cmssw/pull/27983#issuecomment-563971114 made by bot

fwyzard:
Updated and rebased on top of CMSSW_11_1_0_pre2 .

fwyzard:
@cmsbuild, please test

fwyzard:
enable gpu

fwyzard:
@cmsbuild, please test

fwyzard:
enable gpu

fwyzard:
@cmsbuild, please test

fwyzard:
@cmsbuild, please test

fwyzard:
@cmsbuild, please test

fwyzard:
@cmsbuild, please test

fwyzard:
@cmsbuild, please test

fwyzard:
@cmsbuild, please test

fwyzard:
@cmsbuild, please test

kpedro88:
Code rule violations: https://cmssdt.cern.ch/SDT/jenkins-artifacts/pull-request-integration/PR-c28630/5364/codeRules/cmsCodeRule1.txt
```
Rule1 : https://raw.githubusercontent.com/cms-sw/cmssw/master/Utilities/ReleaseScripts/python/cmsCodeRules/config.py
Search for "using namespace" or "using std::" in header files

/RecoPixelVertexing/PixelTrackFitting/interface/BrokenLine.h
[10]
/RecoPixelVertexing/PixelTriplets/plugins/BrokenLineFitOnGPU.h
[23]
/RecoPixelVertexing/PixelTriplets/plugins/CAHitNtupletGeneratorKernelsImpl.h
[22]
/RecoPixelVertexing/PixelTriplets/plugins/RiemannFitOnGPU.h
[21]
/RecoPixelVertexing/PixelTriplets/plugins/gpuPixelDoublets.h
[10]
/RecoPixelVertexing/PixelVertexFinding/test/VertexFinder_t.h
[46]
```

Header consistency issues: https://cmssdt.cern.ch/SDT/jenkins-artifacts/pull-request-integration/PR-c28630/5364/headers_chks.log
```
src/HeterogeneousCore/MPICore/interface/WrapperHandle.h:76:15: error: template-id 'convert_handle<>' for 'void edm::convert_handle(edm::BasicHandle&&, edm::Handle<edm::WrapperBase>&)' does not match any template declaration
   inline void convert_handle(BasicHandle&& bh, Handle<WrapperBase>& result) {
               ^~~~~~~~~~~~~~
In file included from /cvmfs/cms-ib.cern.ch/week1/slc7_amd64_gcc820/cms/cmssw-patch/CMSSW_11_1_X_2020-03-24-2300/src/FWCore/Framework/interface/Event.h:21,
                 from src/HeterogeneousCore/MPICore/interface/WrapperHandle.h:12:
/cvmfs/cms-ib.cern.ch/week1/slc7_amd64_gcc820/cms/cmssw-patch/CMSSW_11_1_X_2020-03-24-2300/src/DataFormats/Common/interface/ConvertHandle.h:22:13: note: candidate is: 'template<class T> edm::Handle<T> edm::convert_handle(edm::BasicHandle&&)'
   Handle<T> convert_handle(BasicHandle&& bh) noexcept(true) {
             ^~~~~~~~~~~~~~
src/HeterogeneousCore/MPICore/interface/WrapperHandle.h: In member function 'bool edm::Event::getByLabel(const edm::InputTag&, edm::Handle<T>&) const [with PROD = edm::WrapperBase]':
src/HeterogeneousCore/MPICore/interface/WrapperHandle.h:150:41: error: no matching function for call to 'convert_handle(std::remove_reference<edm::BasicHandle&>::type, edm::Handle<edm::WrapperBase>&)'
     convert_handle(std::move(bh), result);  // throws on conversion error
                                         ^
In file included from /cvmfs/cms-ib.cern.ch/week1/slc7_amd64_gcc820/cms/cmssw-patch/CMSSW_11_1_X_2020-03-24-2300/src/FWCore/Framework/interface/Event.h:21,
                 from src/HeterogeneousCore/MPICore/interface/WrapperHandle.h:12:
/cvmfs/cms-ib.cern.ch/week1/slc7_amd64_gcc820/cms/cmssw-patch/CMSSW_11_1_X_2020-03-24-2300/src/DataFormats/Common/interface/ConvertHandle.h:22:13: note: candidate: 'template<class T> edm::Handle<T> edm::convert_handle(edm::BasicHandle&&)'
   Handle<T> convert_handle(BasicHandle&& bh) noexcept(true) {
             ^~~~~~~~~~~~~~
/cvmfs/cms-ib.cern.ch/week1/slc7_amd64_gcc820/cms/cmssw-patch/CMSSW_11_1_X_2020-03-24-2300/src/DataFormats/Common/interface/ConvertHandle.h:22:13: note:   template argument deduction/substitution failed:
src/HeterogeneousCore/MPICore/interface/WrapperHandle.h:150:41: note:   candidate expects 1 argument, 2 provided
     convert_handle(std::move(bh), result);  // throws on conversion error
                                         ^
src/HeterogeneousCore/MPICore/interface/WrapperHandle.h: In member function 'bool edm::Event::getByLabel(const string&, const string&, edm::Handle<T>&) const [with PROD = edm::WrapperBase; std::__cxx11::string = std::__cxx11::basic_string<char>]':
src/HeterogeneousCore/MPICore/interface/WrapperHandle.h:165:41: error: no matching function for call to 'convert_handle(std::remove_reference<edm::BasicHandle&>::type, edm::Handle<edm::WrapperBase>&)'
     convert_handle(std::move(bh), result);  // throws on conversion error
                                         ^
In file included from /cvmfs/cms-ib.cern.ch/week1/slc7_amd64_gcc820/cms/cmssw-patch/CMSSW_11_1_X_2020-03-24-2300/src/FWCore/Framework/interface/Event.h:21,
                 from src/HeterogeneousCore/MPICore/interface/WrapperHandle.h:12:
/cvmfs/cms-ib.cern.ch/week1/slc7_amd64_gcc820/cms/cmssw-patch/CMSSW_11_1_X_2020-03-24-2300/src/DataFormats/Common/interface/ConvertHandle.h:22:13: note: candidate: 'template<class T> edm::Handle<T> edm::convert_handle(edm::BasicHandle&&)'
   Handle<T> convert_handle(BasicHandle&& bh) noexcept(true) {
             ^~~~~~~~~~~~~~
/cvmfs/cms-ib.cern.ch/week1/slc7_amd64_gcc820/cms/cmssw-patch/CMSSW_11_1_X_2020-03-24-2300/src/DataFormats/Common/interface/ConvertHandle.h:22:13: note:   template argument deduction/substitution failed:
src/HeterogeneousCore/MPICore/interface/WrapperHandle.h:165:41: note:   candidate expects 1 argument, 2 provided
     convert_handle(std::move(bh), result);  // throws on conversion error
                                         ^
src/HeterogeneousCore/MPICore/interface/WrapperHandle.h: In member function 'bool edm::Event::getByToken(edm::EDGetToken, edm::Handle<T>&) const [with PROD = edm::WrapperBase]':
src/HeterogeneousCore/MPICore/interface/WrapperHandle.h:177:41: error: no matching function for call to 'convert_handle(std::remove_reference<edm::BasicHandle&>::type, edm::Handle<edm::WrapperBase>&)'
     convert_handle(std::move(bh), result);  // throws on conversion error
                                         ^
In file included from /cvmfs/cms-ib.cern.ch/week1/slc7_amd64_gcc820/cms/cmssw-patch/CMSSW_11_1_X_2020-03-24-2300/src/FWCore/Framework/interface/Event.h:21,
                 from src/HeterogeneousCore/MPICore/interface/WrapperHandle.h:12:
/cvmfs/cms-ib.cern.ch/week1/slc7_amd64_gcc820/cms/cmssw-patch/CMSSW_11_1_X_2020-03-24-2300/src/DataFormats/Common/interface/ConvertHandle.h:22:13: note: candidate: 'template<class T> edm::Handle<T> edm::convert_handle(edm::BasicHandle&&)'
   Handle<T> convert_handle(BasicHandle&& bh) noexcept(true) {
             ^~~~~~~~~~~~~~
/cvmfs/cms-ib.cern.ch/week1/slc7_amd64_gcc820/cms/cmssw-patch/CMSSW_11_1_X_2020-03-24-2300/src/DataFormats/Common/interface/ConvertHandle.h:22:13: note:   template argument deduction/substitution failed:
src/HeterogeneousCore/MPICore/interface/WrapperHandle.h:177:41: note:   candidate expects 1 argument, 2 provided
     convert_handle(std::move(bh), result);  // throws on conversion error
                                         ^
src/HeterogeneousCore/MPICore/interface/WrapperHandle.h: In member function 'bool edm::Event::getByToken(edm::EDGetTokenT<T>, edm::Handle<T>&) const [with PROD = edm::WrapperBase]':
src/HeterogeneousCore/MPICore/interface/WrapperHandle.h:189:41: error: no matching function for call to 'convert_handle(std::remove_reference<edm::BasicHandle&>::type, edm::Handle<edm::WrapperBase>&)'
     convert_handle(std::move(bh), result);  // throws on conversion error
                                         ^
In file included from /cvmfs/cms-ib.cern.ch/week1/slc7_amd64_gcc820/cms/cmssw-patch/CMSSW_11_1_X_2020-03-24-2300/src/FWCore/Framework/interface/Event.h:21,
                 from src/HeterogeneousCore/MPICore/interface/WrapperHandle.h:12:
/cvmfs/cms-ib.cern.ch/week1/slc7_amd64_gcc820/cms/cmssw-patch/CMSSW_11_1_X_2020-03-24-2300/src/DataFormats/Common/interface/ConvertHandle.h:22:13: note: candidate: 'template<class T> edm::Handle<T> edm::convert_handle(edm::BasicHandle&&)'
   Handle<T> convert_handle(BasicHandle&& bh) noexcept(true) {
             ^~~~~~~~~~~~~~
/cvmfs/cms-ib.cern.ch/week1/slc7_amd64_gcc820/cms/cmssw-patch/CMSSW_11_1_X_2020-03-24-2300/src/DataFormats/Common/interface/ConvertHandle.h:22:13: note:   template argument deduction/substitution failed:
src/HeterogeneousCore/MPICore/interface/WrapperHandle.h:189:41: note:   candidate expects 1 argument, 2 provided
     convert_handle(std::move(bh), result);  // throws on conversion error
                                         ^
gmake: *** [config/SCRAM/GMake/Makefile.rules:1753: tmp/slc7_amd64_gcc820/src/HeterogeneousCore/MPICore/src/HeterogeneousCoreMPICore/WrapperHandle.h.chk_header] Error 1
```

There are many minor fluctuations in the comparisons. Are these expected/understood?

fwyzard:
Rebased to
  - include the latest ECAL developments
  - apply more code formatting fixes
  - remove the MPI-related changes

fwyzard:
The code-checks errors are actually the merge conflicts - will have to see how to deal with those.

fwyzard:
@cmsbuild, please test

fwyzard:
@cmsbuild, please test

CUDADataFormats/Track/test/BuildFile.xml
```diff
@@ -0,0 +1,13 @@
+<use name="HeterogeneousCore/CUDAUtilities"/>
+
+<bin file="TrajectoryStateSOA_t.cpp" name="cpuTrajectoryStateSOA_t">
+  <use name="eigen"/>
+  <flags CXXFLAGS="-g -DGPU_DEBUG"/>
+</bin>
+
+<bin file="TrajectoryStateSOA_t.cu" name="gpuTrajectoryStateSOA_t">
```
smuzaffar, line 8: @fwyzard , can you please add cuda-gcc-support protection [a] for cuda released build products 

```
<iftool name="cuda-gcc-support">
....
....
</iftool>
```

fwyzard: To be honest, I'm getting to the point where I will simply stop caring about the integration in CMSSW.

VinInn: I think the build system should figure out by itself.

smuzaffar: thats a good point @VinInn

fwyzard: @smuzaffar , on architectures / compiler versions that do not support CUDA, like GCC 9, can we simply not build the CUDA externals, and automatically disable everything that depends on it ?

@makortel if we actually do this, would it be a reason to split the CUDA code into seprate libraries and plugins ?

smuzaffar: @fwyzard , we should keep on build cuda external otherwise how would we know if cuda support a GCC vet rsion or not :-)
But I agree that if cuda does not support our gcc version then we should disable the cuda based  build products in cmssw. This means we should have these products build in a way that they are optional . So cmssw should not fail if they do not exist.

VinInn: How all this will fit in "Heterogeneity"? In principle our ultimate goal is to have a single source.
The "cpu" version shall compile, link, be tested with any kind of compiler we wish to use.

fwyzard: > we should keep on building cuda external otherwise how would we know if cuda support a GCC vet rsion or not :-)

Yes, well, what I meant is "if CUDA is not supported, building the external could fail".

However I realise that we do not really want to do that - for example, we can build llvm/clang with CUDA support as long as we have the libraries, even if GCC does not work with it.

So it's not 100% clear that `<use name="cuda"/>` should mean the CUDA libraries or the CUDA compiler.

fwyzard: >  The "cpu" version shall compile, link, be tested with any kind of compiler we wish to use.

Yes, absolutely.

makortel: > So it's not 100% clear that `<use name="cuda"/>` should mean the CUDA libraries or the CUDA compiler.

In principle the need of the CUDA compiler could be figured out if there are any `.cu` source files, right?

fwyzard: Currently, yes.

If we use Alpaka or similar, we may end up with `.cc` files that need to be compiled by NVCC (or with `.cu` files that include `.cc` files).

fwyzard:
In the [Unit Tests](https://cmssdt.cern.ch/SDT/jenkins-artifacts/pull-request-integration/PR-c28630/5554/unitTests.log) we have some errors like
```
----- Begin Fatal Exception 06-Apr-2020 20:56:25 CEST-----------------------
An exception of category 'LogicError' occurred while
   [1] Constructing module: class=SiPixelClusterProducer label='siPixelClustersPreSplitting@cpu'
Exception Message:
Duplicate Product Identifier 
The Framework requires a unique branch name for each product
which consists of four parts: a friendly class name, module label,
product instance name, and process name. A product has been
registered with a duplicate branch name. The most common way
to fix this error is to modify the product instance name in
one of the offending 'produces' function calls. Another fix
would be to delete one of them if they are for the same product.

    friendly class name = SiPixelClusteredmNewDetSetVector
    module label = siPixelClustersPreSplitting
    product instance name = 
    process name = LHEembeddingCLEAN

The following additional information is not used as part of
the unique branch identifier.

    branch types = Event  Event
    class name = edmNew::DetSetVector<SiPixelCluster>

Note that if the four parts of the branch name are the same,
then this error will occur even if the branch types differ!

```

@makortel could you have a look what is causing it ?

CUDADataFormats/Track/test/BuildFile.xml
makortel: > @smuzaffar , on architectures / compiler versions that do not support CUDA, like GCC 9, can we simply not build the CUDA externals, and automatically disable everything that depends on it ?
> 
> @makortel if we actually do this, would it be a reason to split the CUDA code into seprate libraries and plugins ?

From the framework perspective not really (not sure if the split would make the life easier for the build system, ok libraries could make sense anyway). As outlined in https://github.com/cms-sw/cmssw/issues/28576#issuecomment-605681672, **if** we want a portable configuration, the framework needs to be able to construct all EDModules  and ESProducers for all the platforms the configuration supports. I.e. the CUDA EDProducers need to exist even when we can't build CUDA, implying that the pieces of code depending on CUDA needs to be protected with `#ifdef`s.

(not my favorite solution, but this is where we are today)

makortel: > If we use Alpaka or similar, we may end up with `.cc` files that need to be compiled by NVCC (or with `.cu` files that include `.cc` files).

Right, but then the build system should be otherwise aware (either by explicit rules in the `BuildFile.xml`, or by built-in logic) whether to build for the CUDA backend or not.

makortel:
> could you have a look what is causing it ?

I'll take a look.

makortel:
The other unit test failure
```
----- Begin Fatal Exception 06-Apr-2020 20:11:27 CEST-----------------------
   [0] Processing the python configuration file named /data/cmsbld/jenkins/workspace/ib-run-pr-tests/CMSSW_11_1_X_2020-04-06-1100/src/SLHCUpgradeSimulations/Geometry/test/phase2_digi_reco_pixelntuple_cfg.py
Exception Message:
TypeError: ElectronPerADCGain does not already exist, so it can only be set to a cms.EDProducer or cms.EDAlias

At:
  /cvmfs/cms-ib.cern.ch/week1/slc7_amd64_gcc820/cms/cmssw-patch/CMSSW_11_1_X_2020-04-06-1100/python/FWCore/ParameterSet/Modules.py(283): __addParameter
  /cvmfs/cms-ib.cern.ch/week1/slc7_amd64_gcc820/cms/cmssw-patch/CMSSW_11_1_X_2020-04-06-1100/python/FWCore/ParameterSet/Modules.py(315): __setattr__
  /data/cmsbld/jenkins/workspace/ib-run-pr-tests/CMSSW_11_1_X_2020-04-06-1100/src/SLHCUpgradeSimulations/Geometry/test/phase2_digi_reco_pixelntuple_cfg.py(102): <module>

```
is caused by
https://github.com/cms-sw/cmssw/blob/67e85b5bcf41005db3a2cbf5318c3febbe569e5e/SLHCUpgradeSimulations/Geometry/test/phase2_digi_reco_pixelntuple_cfg.py#L102
and should get fixed with
```diff
- process.siPixelClustersPreSplitting.ElectronPerADCGain  = cms.double(135.)
+ process.siPixelClustersPreSplitting.cpu.ElectronPerADCGain  = 135.
```

CUDADataFormats/Track/test/BuildFile.xml
fwyzard: > the framework needs to be able to construct all EDModules and ESProducers for all the platforms the configuration supports. I.e. the CUDA EDProducers need to exist even when we can't build CUDA, implying that the pieces of code depending on CUDA needs to be protected with `#ifdef`s.

For what is worth, I really don't like this approach.

At this point, I think that the requirement to have unique hashes for the configuration is doing more harm to the process than (any ?) good.

fwyzard:
Thanks, I'll add this.

fwyzard:
@andrius-k @jfernan2 @kmaeshima @fioriNTU @schneiml
could you explain why the various `TestDQMOfflineConfiguration` tests are returning **ERRORS** ?

I don't see any errors in [the logs](https://cmssdt.cern.ch/SDT/jenkins-artifacts/pull-request-integration/PR-c28630/5554/unitTests.log).

schneiml:
@fwyzard Most likely the tests timed out. The chunks should be small enough to pass in the PR tests, but they might be quite close to time limit. In the normal case, each of the tests will clearly say after all of the output if it passed or what the problem was. In this case, they were killed before that.

fwyzard:
I see - OK, then I will ignore them for the time being.

CUDADataFormats/Track/test/BuildFile.xml
makortel: > > the framework needs to be able to construct all EDModules and ESProducers for all the platforms the configuration supports. I.e. the CUDA EDProducers need to exist even when we can't build CUDA, implying that the pieces of code depending on CUDA needs to be protected with `#ifdef`s.
> 
> For what is worth, I really don't like this approach.
> 
> At this point, I think that the requirement to have unique hashes for the configuration is doing more harm to the process than (any ?) good.

I replied in https://github.com/cms-sw/cmssw/issues/28576#issuecomment-610461957.

We can certainly discuss this subject in a Core Software meeting (today or later).

makortel:
> > could you have a look what is causing it ?
> 
> I'll take a look.

It turned out to be a genuine bug in SwitchProducer. I think I have a fix for it now, I'll prepare proper tests for it next before making a PR.

fwyzard:
@cmsbuild, please test

fwyzard:
@smuzaffar can we add the workflows `10824.511` and `10824.512` to the GPU tests ?

fwyzard:
@cmsbuild, please test

smuzaffar:
> @smuzaffar can we add the workflows `10824.511` and `10824.512` to the GPU tests ?

done

fwyzard:
enable gpu

fwyzard:
@cmsbuild, please test

RecoLocalTracker/SiPixelRecHits/plugins/SiPixelRecHitConverter.cc
```diff
@@ -185,9 +189,34 @@ namespace cms {
...
+      numberOfClusters += nclus;
+    }
+    hitsModuleStart[0] = 0;
+    assert(clusInModule.size() > gpuClustering::MaxNumModules);
+    for (int i = 1, n = clusInModule.size(); i < n; ++i)
+      hitsModuleStart[i] = hitsModuleStart[i - 1] + clusInModule[i - 1];
+    assert(numberOfClusters == int(hitsModuleStart[gpuClustering::MaxNumModules]));
+
+    // yes a unique ptr of a unique ptr so edm is happy and the pointer stay still...
+    iEvent.emplace(tHost_, std::move(hmsp));  // hmsp is gone, hitsModuleStart still alive and kicking...
```
fwyzard, line 216: @VinInn I reorganised this a little to fix the merge conflicts and make it slightly more in line with the upstream code.
Could you confirm that this still does what it is supposed to do ?

VinInn: as I said in all my recent talks to DPG this is a huge pain&mess until not merged in official CMSSW...
I will have a look, BUT we really need to consolidate this code with the other 3 producers

VinInn: at first glance looks ok. would need to check the diff generated by the merge conflicts...

fwyzard:
please test

fwyzard:
@cmsbuild, please test

RecoLocalTracker/SiPixelClusterizer/plugins/gpuCalibPixel.h
```diff
@@ -0,0 +1,67 @@
...
+#include "CondFormats/SiPixelObjects/interface/SiPixelGainForHLTonGPU.h"
+#include "HeterogeneousCore/CUDAUtilities/interface/cuda_assert.h"
+
+#include "gpuClusteringConstants.h"
+
+namespace gpuCalibPixel {
+
+  constexpr uint16_t InvId = 9999;  // must be > MaxNumModules
+
+  constexpr float VCaltoElectronGain = 47;         // L2-4: 47 +- 4.7
```
mmusich, line 16: @VinInn, as you maybe know we changed the gain calibration scheme (for Run3 only) in this PR: https://github.com/cms-sw/cmssw/pull/29333 by absorbing the VCal calibration from within the gain calibration DB payloads. 
Since CMSSW_11_1_0_pre7 this code is not compatible anymore with the gain calibration in Global Tag.
This leads to funky pixel cluster charge:

image

**IF** we have to support Run-2 *and* Run-3 production as well these numbers can't stay hard-coded as here. Please advice.

VinInn: my advice would be to change the payload for run< 3 as well.
How the code for run < 3 looks like? please point me to the calibration code for run <3

mmusich: we wanted to avoid changing also the run < 3 payloads because it's a labour intensive procedure involving tens of tags and scores of GTs...
The numbers you have currently in the patatrack branch are OK for Run1 and Run2, while for Run3 we have just:
```python
VCaltoElectronGain = 1, # all gains=1, pedestals=0
VCaltoElectronGain_L1 = 1,
VCaltoElectronOffset = 0,
VCaltoElectronOffset_L1 = 0
```
all the conversion factors are already included in the Run3 gain DB payloads (all gains are multiplied and shifted according to the slope and offsets which were earlier passed as configuration parameters in the clusterizer).
You might also consult this talk: https://indico.cern.ch/event/879474/contributions/3855389/

@dkotlins @IzaakWN @tsusa @tvami FYI.

mmusich: @VinInn 

>  (will have a look, will go in some sort of config, I suspect)

yes it is here: https://github.com/cms-sw/cmssw/blob/master/RecoLocalTracker/SiPixelClusterizer/python/SiPixelClusterizer_cfi.py#L7-L28

VinInn: I will introduce a flag isRun3 setting it using the modifier and take proper action in gpuCalib

VinInn: https://github.com/cms-patatrack/cmssw/pull/492

fwyzard: @mmusich, thanks for the warning.

@VinInn would it make sense to make the four values customisable from python, like in `SiPixelClusterizer_cfi.py ` ?

I don't mind either way, simply trying to think what is more "future proof".

fwyzard: @mmusich, is there a difference in data vs MC ?

mmusich: > @mmusich, is there a difference in data vs MC ?

no.

fwyzard: Thanks.

VinInn: I do not like to have in config values that shall be coherent (this is production not a testbeam)
I prefer a flag   and values (not in DB), hardcoded.

fwyzard:
@cmsbuild, please test

fwyzard:
code-checks

fwyzard:
@cmsbuild, please test

fwyzard:
@cmsbuild, please test

fwyzard:
@cmsbuild, please test

fwyzard:
Apart from the conflict, these two seem counter intuitive:
```diff
diff --git a/CUDADataFormats/SiPixelDigi/src/SiPixelDigiErrorsCUDA.cc b/CUDADataFormats/SiPixelDigi/src/SiPixelDigiErrorsCUDA.cc
index ef229be4b991..075d408a6f6f 100644
--- a/CUDADataFormats/SiPixelDigi/src/SiPixelDigiErrorsCUDA.cc
+++ b/CUDADataFormats/SiPixelDigi/src/SiPixelDigiErrorsCUDA.cc
@@ -38,5 +38,5 @@ SiPixelDigiErrorsCUDA::HostDataError SiPixelDigiErrorsCUDA::dataErrorToHostAsync
   }
   auto err = *error_h;
   err.set_data(data.get());
-  return HostDataError(std::move(err), std::move(data));
+  return HostDataError(err, std::move(data));
 }
```
and
```diff
diff --git a/EventFilter/SiPixelRawToDigi/plugins/SiPixelDigiErrorsSoAFromCUDA.cc b/EventFilter/SiPixelRawToDigi/plugins/SiPixelDigiErrorsSoAFromCUDA.cc
index 8817606043a6..c5b568750ad7 100644
--- a/EventFilter/SiPixelRawToDigi/plugins/SiPixelDigiErrorsSoAFromCUDA.cc
+++ b/EventFilter/SiPixelRawToDigi/plugins/SiPixelDigiErrorsSoAFromCUDA.cc
@@ -52,7 +52,7 @@ void SiPixelDigiErrorsSoAFromCUDA::acquire(const edm::Event& iEvent,
   const auto& gpuDigiErrors = ctx.get(iEvent, digiErrorGetToken_);

   auto tmp = gpuDigiErrors.dataErrorToHostAsync(ctx.stream());
-  error_ = std::move(tmp.first);
+  error_ = tmp.first;
   data_ = std::move(tmp.second);
   formatterErrors_ = &(gpuDigiErrors.formatterErrors());
 }
```
@makortel what do you think ?
Are we missing a move constructor somewhere ?

makortel:
> ```diff
> diff --git a/CUDADataFormats/SiPixelDigi/src/SiPixelDigiErrorsCUDA.cc b/CUDADataFormats/SiPixelDigi/src/SiPixelDigiErrorsCUDA.cc
> index ef229be4b991..075d408a6f6f 100644
> --- a/CUDADataFormats/SiPixelDigi/src/SiPixelDigiErrorsCUDA.cc
> +++ b/CUDADataFormats/SiPixelDigi/src/SiPixelDigiErrorsCUDA.cc
> @@ -38,5 +38,5 @@ SiPixelDigiErrorsCUDA::HostDataError SiPixelDigiErrorsCUDA::dataErrorToHostAsync
>    }
>    auto err = *error_h;
>    err.set_data(data.get());
> -  return HostDataError(std::move(err), std::move(data));
> +  return HostDataError(err, std::move(data));
>  }
> ```

The `err` is `cms::cuda::SimpleVector` that holds two intergers and a raw pointer
https://github.com/cms-sw/cmssw/blob/6d084dd7e479d66be398b8e45a40f50fae63ed6b/HeterogeneousCore/CUDAUtilities/interface/SimpleVector.h#L115-L120
 and has default copy and move operations, so a move would be effectively the same as copy.

> ```diff
> diff --git a/EventFilter/SiPixelRawToDigi/plugins/SiPixelDigiErrorsSoAFromCUDA.cc b/EventFilter/SiPixelRawToDigi/plugins/SiPixelDigiErrorsSoAFromCUDA.cc
> index 8817606043a6..c5b568750ad7 100644
> --- a/EventFilter/SiPixelRawToDigi/plugins/SiPixelDigiErrorsSoAFromCUDA.cc
> +++ b/EventFilter/SiPixelRawToDigi/plugins/SiPixelDigiErrorsSoAFromCUDA.cc
> @@ -52,7 +52,7 @@ void SiPixelDigiErrorsSoAFromCUDA::acquire(const edm::Event& iEvent,
>    const auto& gpuDigiErrors = ctx.get(iEvent, digiErrorGetToken_);
> 
>    auto tmp = gpuDigiErrors.dataErrorToHostAsync(ctx.stream());
> -  error_ = std::move(tmp.first);
> +  error_ = tmp.first;
>    data_ = std::move(tmp.second);
>    formatterErrors_ = &(gpuDigiErrors.formatterErrors());
>  }
> ```

Same happens here, `error_` is `cms::cuda::SimpleVector` (well, this one calls the function above).

It seems to me that these changes are ok, and they do remove a possibly confusing `std::move`.

fwyzard:
@cmsbuild, please test

fwyzard:
@cmsbuild, please test

fwyzard:
Most of the changes have been proposed for integration in:
[#31703](https://github.com/cms-sw/cmssw/pull/31703) Patatrack integration - common data formats (5/N)
[#31704](https://github.com/cms-sw/cmssw/pull/31704) Patatrack integration - calorimeters shared code (6/N)
[#31719](https://github.com/cms-sw/cmssw/pull/31719) Patatrack integration - ECAL local reconstruction (7/N)
[#31720](https://github.com/cms-sw/cmssw/pull/31720) Patatrack integration - HCAL local reconstruction (8/N)
[#31721](https://github.com/cms-sw/cmssw/pull/31721) Patatrack integration - Pixel local reconstruction (9/N)
[#31722](https://github.com/cms-sw/cmssw/pull/31722) Patatrack integration - Pixel track reconstruction (10/N)
[#31723](https://github.com/cms-sw/cmssw/pull/31723) Patatrack integration - Pixel vertex reconstruction (11/N)

fwyzard:
@cmsbuild, please test

fwyzard:
please test

fwyzard:
please test

fwyzard:
please test

fwyzard:
please test

fwyzard:
I think for the first time the Patatrack branch is finally passing all tests !

smuzaffar:
@fwyzard , I have added GPU context in the bot comments and commits statuses. If you do not mind , I would like to re-run the tests for this PR

fwyzard:
of course, go ahead

smuzaffar:
please test

fwyzard:
code-checks

fwyzard:
please test

fwyzard:
please test

fwyzard:
please test

jfernan2:
It looks like many Tracking plots are being modified by this PR getting less entries, I wonder if some aceptance parameter was not updated in the process as happened in https://github.com/cms-sw/cmssw/pull/31721 
Thanks

fwyzard:
Yes, but I'm keeping the change here for the moment.

fwyzard:
please test

fwyzard:
enable gpu

fwyzard:
please test

fwyzard:
please test

fwyzard:
please test

fwyzard:
+1

silviodonato:
@fwyzard can we finally close this PR?

fwyzard:
I think so - most changes are in CMSSW_11_3_0_pre6.

PR #27987, opened 2019-09-13
Adding trigonometric functions into the FormulaEvaluator

abuubaidah89:
Adding trigonometric functions to FormulaEvaluator. This is the first step in modifying JER scale factors application to be jet pT dependent. Latest JER scale factors might need trigonometric functions for parameterization as a function of pT.

### DIFF
--- a/CommonTools/Utils/src/FormulaEvaluator.cc
+++ b/CommonTools/Utils/src/FormulaEvaluator.cc
@@ -607,7 +607,7 @@ namespace {
   const std::string k_TMath__ASinH("TMath::ASinH");
   const std::string k_atanh("atanh");
   const std::string k_TMath__ATanH("TMath::ATanH");
-  
+
   EvaluatorInfo FunctionFinder::createEvaluator(std::string::const_iterator iBegin,
                                                 std::string::const_iterator iEnd) const {
     EvaluatorInfo info;
--- a/CommonTools/Utils/src/FormulaEvaluator.cc
+++ b/CommonTools/Utils/src/FormulaEvaluator.cc
@@ -581,6 +581,32 @@ namespace {
   const std::string k_TMath__Sqrt("TMath::Sqrt");
   const std::string k_abs("abs");
   const std::string k_TMath__Abs("TMath::Abs");
+  const std::string k_cos("cos");
+  const std::string k_sin("sin");
+  const std::string k_tan("tan");
+  const std::string k_acos("acos");
+  const std::string k_asin("asin");
+  const std::string k_atan("atan");
+  const std::string k_atan2("atan2");
+  const std::string k_cosh("cosh");
+  const std::string k_sinh("sinh");
+  const std::string k_tanh("tanh");
+  const std::string k_acosh("acosh");
+  const std::string k_asinh("asinh");
+  const std::string k_atanh("atanh");
+  const std::string k_TMath__Cos("TMath::Cos");
+  const std::string k_TMath__Sin("TMath::Sin");
+  const std::string k_TMath__Tan("TMath::Tan");
+  const std::string k_TMath__ACos("TMath::ACos");
+  const std::string k_TMath__ASin("TMath::ASin");
+  const std::string k_TMath__ATan("TMath::ATan");
+  const std::string k_TMath__ATan2("TMath::ATan2");
+  const std::string k_TMath__CosH("TMath::CosH");
+  const std::string k_TMath__SinH("TMath::SinH");
+  const std::string k_TMath__TanH("TMath::TanH");
+  const std::string k_TMath__ACosH("TMath::ACosH");
+  const std::string k_TMath__ASinH("TMath::ASinH");
+  const std::string k_TMath__ATanH("TMath::ATanH");
 
   EvaluatorInfo FunctionFinder::createEvaluator(std::string::const_iterator iBegin,
                                                 std::string::const_iterator iEnd) const {
@@ -633,7 +659,7 @@ namespace {
     if (info.evaluator.get() != nullptr) {
       return info;
     }
-
+    
     info = checkForSingleArgFunction(
         iBegin, iEnd, m_expressionFinder, k_TMath__Sqrt, [](double iArg) -> double { return std::sqrt(iArg); });
     if (info.evaluator.get() != nullptr) {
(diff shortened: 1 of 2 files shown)


abuubaidah89:
For attention @FHead  @juska

fabiocos:
please test

FHead:
Hi all,

I see the error message for the failed unit test is

`Bla/Blo/src/precompile.h file not found neither in /build/cmsbld/jenkins/workspace/ib-run-pr-tests/CMSSW_11_0_X_2019-09-12-2300 nor in /cvmfs/cms-ib.cern.ch/week1/slc7_amd64_gcc820/cms/cmssw-patch/CMSSW_11_0_X_2019-09-12-2300 nor in /cvmfs/cms-ib.cern.ch/nweek-02593/slc7_amd64_gcc820/cms/cmssw/CMSSW_11_0_X_2019-09-11-2300 `

It seems to be unrelated to the modified code.  Do you know how we can resolve this?


Best Regards,
Yi Chen

perrotta:
please test
(The precompile.h file should be there, let retry)

CommonTools/Utils/src/FormulaEvaluator.cc
```diff
@@ -581,6 +581,32 @@ namespace {
   const std::string k_TMath__Sqrt("TMath::Sqrt");
   const std::string k_abs("abs");
   const std::string k_TMath__Abs("TMath::Abs");
+  const std::string k_cos("cos");
+  const std::string k_sin("sin");
```
slava77: please group TMath and "std" variants in the same way they appear already for other functions.

slava77 (commented):
please add tests covering each of the newly added functions at least once in CommonTools/Utils/test/testFormulaEvaluator.cc

CommonTools/Utils/src/FormulaEvaluator.cc
abuubaidah89: hi @slava77 , i have modified those codes as per your suggestion.

perrotta:
please test

abuubaidah89:
Error at the function acosh. I will take a look at it.

slava77:
@cmsbuild please test

CommonTools/Utils/test/testFormulaEvaluator.cc
```diff
@@ -585,6 +585,214 @@ void testFormulaEvaluator::checkFormulaEvaluator() {
     CPPUNIT_ASSERT(f.evaluate(emptyV, emptyV) == 2);
   }
 
+  {
+    reco::FormulaEvaluator f("cos(0.5)");
+
+    std::vector<double> emptyV;
+
+    CPPUNIT_ASSERT(fabs(f.evaluate(emptyV, emptyV) - std::cos(0.5)) < 1e-9);
```
slava77: similarly complex log, pow, exp unit tests are using equality.
Please switch to it here (for all of the added tests) unless there is a strong reason not to.

FHead: I believe the trigger for this is that the tests for some of the functions failed right above commit https://github.com/cms-sw/cmssw/pull/27987/commits/094b3f6f09194c982daaab0de22e72bfee35db76 .

It is traced down to a difference at the 16th digit after the decimal point for std::acosh(...).  What would you suggest them to do?

Best Regards,
Yi Chen

abuubaidah89: All other functions have passed the equality check in the previous test except for std::acosh(...), and the cause of the failure is as mentioned by Yi Chen. Upon further inspection, the std::atanh(...) have the same issue too.

An example of the issue is as demonstrated below:

    std::cout << std::setprecision(20) << "acosh(2.0) = " << f.evaluate(emptyV, emptyV) << "\n";
    std::cout << std::setprecision(20) << "std::acosh(2.0) = " << std::acosh(2.0) << "\n";
    std::cout << "acosh(2.0) - std::acosh(2.0) = " << fabs(f.evaluate(emptyV, emptyV) - std::acosh(2.0));

The results:

> .....acosh(2.0) = 1.3169578969248165734
> std::acosh(2.0) = 1.3169578969248167954
> acosh(2.0) - std::acosh(2.0) = 2.2204460492503130808e-16

Hence I change the requirement for the test from evaluating equality to the difference (fabs(a-b) < 1e-9...), and for consistency, I did the same for every other trigonometric function added.

What is your suggestion to resolve this? Should I maintain using equality for every other function and only calculate the difference for std::acosh(...) and std::atanh(...)?

abuubaidah89: Note that this only happens for std::acosh() and std::atanh() but no problem for the corresponding TMath functions.

slava77: for all cases where equality works, I prefer that we use equality.
It would be good to understand the origins of the difference in the other case, but for the unit test now an abs seems to be the only solution.

slava77:
@cmsbuild please test

CommonTools/Utils/test/testFormulaEvaluator.cc
slava77: @Dr15Jones 
please check if you have some insight in the numerical difference observed for achosh calls.
Thank you.

Dr15Jones: gcc appears to be calculating the value at compile time, see https://godbolt.org/z/4sADhe, and gets a different value from the call.  While if I do the same with clang, see https://godbolt.org/z/4KmgvL, it calls std::acosh and returns the same value as the call to the modified FormulaEvaluator. 

So this is just a case of two different ways of truncating a double at the last bit.

slava77: should we try to convince gcc here to compute the reference value at run time
or is is reasonable to just leave fabs(delta) here?

Dr15Jones: How about leave the delta but add a comment about gcc computing the value differently at compile time?

abuubaidah89: Thanks @Dr15Jones for the insight. So how do I proceed, @slava77? Comment them out?

slava77: > Thanks @Dr15Jones for the insight. So how do I proceed, @slava77? Add comment above the deltas?

yes, please add a comment above the ```CPPUNIT_ASSERT``` lines using fabs [one comment is enough if these fabs are within 5-10 lines from each other]

slava77:
@cmsbuild please test

slava77:
+1

for https://github.com/cms-sw/cmssw/pull/27987	5d02d8d 
- jenkins tests pass and comparisons with the baseline show no differences (expected)

fabiocos:
+1

fabiocos:
merge

santocch:
+1

Issue #27374, opened 2019-06-27
Fix jets with crazy momenta in miniAOD


perrotta:
BoostedDoubleSVProd is experiencing some crash in nanoAOD productions from miniAOD, see  https://its.cern.ch/jira/browse/CMSCOMPPR-6445

The origin of the crash was identified as due to subjets components that have nan's in their momenta values. A debug of one such case revealed that those subjets with nan's were originating from jets having crazy momenta (|p|=82 TeV in the example pinpointed in #27238).

The fix proposed in #27238 is able to rescue BoostedDoubleSVProd from this issue, but it doesn't solve the original bug, This will have to be identified and fixed when the jets are produced.

A recipe to reproduce could be to run on the following event
```
process.source.eventsToProcess = cms.untracked.VEventRange('303832:918:1081148408-303832:918:1081148408')
process.source.fileNames = filesRelValTTbarPileUpMINIAODSIM
process.source.fileNames = cms.untracked.vstring( '/store/data/Run2017E/DoubleEG/MINIAOD/31Mar2018-v1/00000/5A8BC921-F437-E811-81ED-1866DAEB3370.root')
```
a nanoAOD or any other job which accesses subjets from miniAOD.

Another possible job that hits it is obtained by customizing and run RecoBTag/TensorFlow/test/test_deep_doublex_cfg.py, as explained in https://github.com/cms-sw/cmssw/pull/27238#issuecomment-505643478

This is put to the attention of the jet reco contacts: please @rappoccio @knash @gkasieczka have a look and report here about your findings

perrotta:
assign reconstruction

slava77:
in which jet collection are these jets and does the problem come from miniAOD or from AOD?

andrzejnovak:
> in which jet collection are these jets and does the problem come from miniAOD or from AOD?

https://github.com/cms-sw/cmssw/blob/master/RecoBTag/TensorFlow/test/test_deep_doublex_cfg.py#L48

perrotta:
Curiously enough, in the same miniAOD input file there are also a few CaloJets with about 80/90 TeV energy:
![CaloJets](https://user-images.githubusercontent.com/4069749/60343601-a8670e80-99b4-11e9-9a64-e6a399fcce76.jpg)

slava77:
> > in which jet collection are these jets and does the problem come from miniAOD or from AOD?
> 
[...]

PR #27238, opened 2019-06-18
Fix BoostedDoubleSVProd crashing with nan/inf daughters

andrzejnovak:
#### PR description:

Fix for a problem reported in https://its.cern.ch/jira/browse/CMSCOMPPR-6445. Crashes caused by daughters with nan/inf values. Protection is added.

#### PR validation:

Local test, minor straightforward change

PR #27988, opened 2019-09-13
Truncate pt of PackedCandidate

rappoccio:
#### PR description:

In response to #27374, truncating pt of PackedCandidate in case the 32 bit representation is beyond the 16-bit max (represented by the "minifloat" as inf, as per that standard requirement [here](ftp://ftp.fox-toolkit.org/pub/fasthalffloatconversion.pdf)).  This was giving problems downstream when the 16-bit "infinity" was widened to 32 bits and became inaccurate.

#### PR validation:

This is so deep in the software stack that it is challenging to predict the behavior without ALL of the validations being run.

### DIFF
--- a/DataFormats/PatCandidates/src/PackedCandidate.cc
+++ b/DataFormats/PatCandidates/src/PackedCandidate.cc
@@ -11,7 +11,7 @@ CovarianceParameterization pat::PackedCandidate::covarianceParameterization_;
 std::once_flag pat::PackedCandidate::covariance_load_flag;
 
 void pat::PackedCandidate::pack(bool unpackAfterwards) {
-  float unpackedPt = std::min(p4_.load()->Pt(),MiniFloatConverter::max());
+  float unpackedPt = std::min(p4_.load()->Pt(), MiniFloatConverter::max());
   packedPt_ = MiniFloatConverter::float32to16(unpackedPt);
   packedEta_ = int16_t(std::round(p4_.load()->Eta() / 6.0f * std::numeric_limits<int16_t>::max()));
   packedPhi_ = int16_t(std::round(p4_.load()->Phi() / 3.2f * std::numeric_limits<int16_t>::max()));
--- a/DataFormats/PatCandidates/src/PackedCandidate.cc
+++ b/DataFormats/PatCandidates/src/PackedCandidate.cc
@@ -11,7 +11,7 @@ CovarianceParameterization pat::PackedCandidate::covarianceParameterization_;
 std::once_flag pat::PackedCandidate::covariance_load_flag;
 
 void pat::PackedCandidate::pack(bool unpackAfterwards) {
-  float unpackedPt = std::min(p4_.load()->Pt(), MiniFloatConverter::max());
+  float unpackedPt = std::min<float>(p4_.load()->Pt(), MiniFloatConverter::max());
   packedPt_ = MiniFloatConverter::float32to16(unpackedPt);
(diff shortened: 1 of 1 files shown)


perrotta:
@gpetruc FYI

perrotta:
Thank you @rappoccio 
Could you please run the validation on your sample with problematic events, just to verify that this actually fixes the issue in #27374? Then for a larger scale validation (on "normal" events) we'll rely on the usual RelVals

perrotta:
please test

slava77:
am I guessing correctly that this is going to be backported to 106X?
@perrotta @rappoccio what do you think?

perrotta:
> am I guessing correctly that this is going to be backported to 106X?
> @perrotta @rappoccio what do you think?

That would not satisfy the "no change" rule for production releases, at least for a few very rare cases.
On the other hand, I don't even see the promised backport of the patch #27238, that would have made this fix not necessary: @andrzejnovak , was it submitted and integrated only in some nanoAOD "private" release?

rappoccio:
I can do a backport if desired, but let me know.

perrotta:
@rappoccio could you please verify as in https://github.com/cms-sw/cmssw/issues/27374#issuecomment-516060700 that the packing of the candidates gets fixed with this PR?

rappoccio:
Apologies for the delay. I confirm that this is now working as intended.

Before fix: 

```
Events->Scan("patPackedCandidates_packedPFCandidates__PAT.obj.pt()", "patPackedCandidates_packedPFCandidates__PAT.obj.pt() > 100");
***********************************
*    Row   * Instance * patPacked *
***********************************
*        0 *     1614 *       inf *
***********************************
```

After fix: 

```
Events->Scan("patPackedCandidates_packedPFCandidates__PAT.obj.pt()", "patPackedCandidates_packedPFCandidates__PAT.obj.pt() > 100");
***********************************
*    Row   * Instance * patPacked *
***********************************
*        0 *     1614 *     65504 *
***********************************
```

So this is giving the 16-bit max as intended.

perrotta:
Thank you @rappoccio : everything as expected, but better to check it explicitly

perrotta:
+1
- The fix replaces nan's in miniAOD outputs with some overflow value for nonphysically monster energy jets in PackedCandidate, thus avoiding possible crashes in the code that try to access them.
- Jenkins tests pass and show no differences for non pathological events

perrotta:
One has to decide whether it is worth backporting this to 10_6_X 

Discussing it with Slava we agreed that this could be considered more a bug fix that prevents unwanted crashes in productions (e.g. nanoAOD), rather than a behavior changing PR. Then probably a backport PR should be foreseen. What do you think, @fabiocos?

fabiocos:
@perrotta to my mind this is just a protection for pathological situations, so I do not see any violation of the no-physics-change policy if a backport is done (and I would consider it useful as 10_6_X should serve us for long time). So I would suggest to go ahead wit it

andrzejnovak:
@perrotta also if you backport this, the boosted sv producer protection
should be redundant

rappoccio:
Backport is #28043

slava77:
> @perrotta also if you backport this, the boosted sv producer protection should be redundant

it will actually not be redundant for the cases of running on old miniAOD (e.g. some nanoAOD reprocessing with btags remade). I'm not sure though if we have these cases at the production scale for 106X.

fabiocos:
+1

fabiocos:
merge

santocch:
+1

PR #27985, opened 2019-09-13
DQM tau miniaod validation CMSSW_106X

aspiezia:
#### PR description:
Backport of https://github.com/cms-sw/cmssw/pull/27925
As in the previous pull request:
This PR has been done in order to add some histograms for the tau validation. The histograms are meant to validate the miniAOD part, in particolar for what it concern the DeepTau discriminators that can be checked only at miniAOD level. @rmanzoni @roger-wolf

#### PR validation:

First the changes have been tested with some RelVal samples in order to check that the histograms are created and filled in a sensible way.
Then, the instructions of https://cms-sw.github.io/PRWorkflow.html (see [*]) have been followed.

[*]
cd src
scram b distclean
git cms-checkdeps -a -A
scram b -j 8
scram b runtests
runTheMatrix.py -l limited -i all --ibeos

PR #27925, opened 2019-09-04
TAU miniaod validation/dqm

aspiezia:
#### PR description:
This PR has been done in order to add some histograms for the tau validation. The histograms are meant to validate the miniAOD part, in particolar for what it concern the DeepTau discriminators that can be checked only at miniAOD level. @rmanzoni @roger-wolf 

#### PR validation:
First the changes have been tested with some RelVal samples in order to check that the histograms are created and filled in a sensible way.
Then, the instructions of https://cms-sw.github.io/PRWorkflow.html (see [*]) have been followed.

[*]
cd src
scram b distclean 
git cms-checkdeps -a -A
scram b -j 8
scram b runtests
runTheMatrix.py -l limited -i all --ibeos

PR #27989, opened 2019-09-13
backport of tau validation in 102X

aspiezia:
After the changes in CMSSW_11X (https://github.com/cms-sw/cmssw/pull/27925)
and the backport in CMSSW_106X (https://github.com/cms-sw/cmssw/pull/27985),
we would like also to backport in CMSSW_102X.
@rmanzoni @roger-wolf


jfernan2:
@aspiezia what's the reason behind this backport?

jfernan2:
@aspiezia ìt depends on which UltraLegacy campaigns you plan to validate

rmanzoni:
Hi, 106 we need for sure, if anything for the UL, and Aniello has submitted a pr for that. 102 we were less certain. I think we need to understand if there will be newer 102 samples that will include DeepTau.
Please let us come back to you later today.

rmanzoni:
@jfernan2: we have backports of DeepTau to 10_2_X (adding DeepTau monitor plots is the main goal of @aspiezia's PR), so we thought it would be useful to have it in 10_2_X DQM too.

If the 10_2_X release cycle is not closed I guess it's perhaps ok to have this PR in, even though I agree that this is not critical and the focus is on UL, and hence 10_6_X, and after.

@jfernan2 let us know if there is something that doesn't sound right in this plan.

jfernan2:
@rmanzoni  I do not know if 10_2_X is actually closed, I let @fabiocos to comment but I am afraid there will not be any further production

jfernan2:
please test

fabiocos:
@rmanzoni @aspiezia @jfernan2 10_2_X is a production release, so only motivated backports should enter, it is still active for production of course. The point is to understand whether we have a real production use case where to benefit of this code.

rmanzoni:
@fabiocos at the moment, and for the foreseeable future, we don't plan anything that is not managed centrally. To my knowledge, correct me if I'm wrong, there's no 10_2_X campaign ahead of us. For the UL campaigns, we have a corresponding 10_6_X PR, so that is covered. Perhaps the conclusion is that this backport is not needed?

jfernan2:
+1

jfernan2:
@aspieza I believe this PR can be closed since the backport is not needed. Thanks

aspiezia:
@jfernan2 that is ok for me.

PR #27991, opened 2019-09-13
Removed unimplemented move constructor from StorageAccount

Dr15Jones:
#### PR description:

Although the move constructor was marked '=default' the compiler was unable to generate it. Given all member data of the class as atomics of builtin types, a move constructor would be no
different from a copy constructor anyway.

#### PR validation:

Compiling using clang no longer generates the warning.


Dr15Jones:
please test

Dr15Jones:
+1

fabiocos:
+1

PR #27993, opened 2019-09-13
Removed implicitly dropped move assignment from RectangularEtaPhiTrackingRegion

Dr15Jones:
#### PR description:

The compiler was implictly not implementing the move assignment operator even though it was marked as '=default'. The reason is there is a const member data which prohibits any assignments.

#### PR validation:

Compiling using clang no longer gives a warning.


Dr15Jones:
please test

perrotta:
Is it safe to delete a method which was explicitly included in the class? Well, apparently yes, because all tests run succesfully to conclusion: this demonstrates that it is never used in the current cmssw code.
I wonder however if we could have get the same result, without deleting the copy assignment, by making static the const member of the class.
@Dr15Jones what do you think?

Dr15Jones:
@perrotta the const is not actually fixed at compile time, the main constructor is passed a value used to set that const. Therefore making it static will not work.

perrotta:
+1
- Technical
- Jenkins tests pass

fabiocos:
+1

PR #27994, opened 2019-09-14
stronger edm::propagate_const const-correctness

dan131riley:
#### PR description:

The current edm::propagate_const permits a const smart pointer to contain a non-const object.  This is not allowed by the gcc and llvm implementations of the  proposed std::experimental::propagate_const it is meant to emulate.  This PR modifies edm::propagate_const to match the behavior of the gcc and llvm implementations.

#### PR validation:

"checkdeps -a -A" checked out ~95% of CMSSW, but only one violation of the stricter check was found.  The only effects of this change should be compile time errors, but nevertheless a limited matrix was run, with no unexpected errors (some workflows failed with step 1 DAS errors).

### DIFF
--- a/FWCore/Utilities/interface/propagate_const.h
+++ b/FWCore/Utilities/interface/propagate_const.h
@@ -61,35 +61,39 @@ namespace edm {
     }
 
     // ---------- const member functions ---------------------
-    element_type const* get() const {
-      return to_raw_pointer(m_value);
-    }
+    element_type const* get() const { return to_raw_pointer(m_value); }
     element_type const* operator->() const { return this->get(); }
     element_type const& operator*() const { return *m_value; }
 
     operator element_type const*() const { return this->get(); }
 
     // ---------- member functions ---------------------------
-    element_type* get() {
-      return to_raw_pointer(m_value);
-    }
+    element_type* get() { return to_raw_pointer(m_value); }
     element_type* operator->() { return this->get(); }
     element_type& operator*() { return *m_value; }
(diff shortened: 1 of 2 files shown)


Dr15Jones:
please test

fwyzard:
@dan131riley thanks for the change.

At this point, is there any reason why we don't just do
```c++
#include <experimental/propagate_const>

namespace edm {
  using std::experimental::propagate_const;
}
```
?

dan131riley:
If it ever reaches full std:: I expect we’ll get rid of the edm version entirely, but currently there is some doubt about its prospects—it was proposed in 2014, so it has been passed over for both 2017 and 2020 merge windows.

Dr15Jones:
+1

Dr15Jones:
@fwyzard I think Dan's reply did an excellent job expressing my thoughts. I'm very much against the use of std::experiment items in production code. Compilers are not required to implement them, there is no guarantee they will make it into the standard or even stick around. Even if they do make it into the standard their interface is not guaranteed to remain the same. By having our own versions of classes which work for our case we will always have the code around and we can switch to the std:: version at our convenience if they are adopted.

fwyzard:
@Dr15Jones sure, ok.

I mentioned because - in this occasion - we seem to be copying what GCC and CLANG are providing, so I thought we might as well use it directly.

Thanks for the feedback and for adapting the code to work without C++ 17.

fabiocos:
+1

PR #26232, opened 2019-03-21
Save DQM output per lumisection for UL rereco

schneiml:
#### PR description:

This PR enables per-lumisection saving of *all* DQM output for the UL rereco. This wish came up for machine learning studies targeting per-LS data certification or online monitoring.

In prompt reco, we have per-LS DQM output available due to how Tier0 processes only ~1 LS per job, and the per-job DQMIO output files are available. It would be very nice to have the same granularity of data preserved for the UL rereco, which in addition provides the full run2 dataset processed using the same software.

This is safe to do for most of DQM, since HARVESTING will re-generate per run/job statistics from the per-lumisection snapshots. It is not save for legacy modules that modify MEs at non-standard times or that do not support merging correctly; none of these should exist in RECO/DQM step since the threaded migration, but since there are still a few, I apply the change only to the non-legacy MEs.

#### PR validation:

Validated on a modified `136.85` that crosses a lumi boundary: `runTheMatrix.py -l 136.85 --command '--customise_commands "if process.process == \"reHLT\": process.source.skipEvents=cms.untracked.uint32(3800)\n"' --ibeos -t 8`. Will need some special attention during relval since the DQMIO output file size will increase.

Known problems:
- Various "known unstable" plots change behaviour. These change behaviour on trivial changes such as adding debug output, and might be caused by race conditions.
- The behaviour of per-lumi plots in harvesting changes. This can lead to "double counting" in the existing offline per-lumi harvesting code, and cause subtly incorrect results. Since there is very little such code, I manually adapted the places I could find; however, due to the freedom that harvesting code has, I can not guarantee that these are all cases.

PR #28002, opened 2019-09-16
DQM/Info: fix offline DCS bit summary

schneiml:
#### PR description:

This PR attempts to fix the DCS status plot at `Info/EventInfo/reportSummaryMap`. Since the harvesting semantic changes from #26232 this plot was showing incorrect values, depending on the order that LS where processed in.

The fix is a simple additional `Reset` call to restore the old semantics. However, there are more problems with this module: The `Physics Declared` row seems to not work correctly since a long time, and shows more arbitrary fluctuations with this change. Also, it is not clear if this code will work without the SCAL FED.

Therefore, in the longer term, we should drop this module and replace it with the logically much simpler online version. That, however, might require additional changes to the render plugins in DQMGUI.

This PR also contains some scripts to help debugging such problems in HARVESTING.

#### PR validation:

In the `runTheMatrix` workflows no changes are expected, since the problem only appears when more than one lumisecition is processed. This PR contains a test script that re-runs harvesting from production DQMIO data that was used to test this change.

Not a backport, but we might want a backport for UL rereco.

### DIFF
--- a/DQM/EcalMonitorClient/src/DQWorkerClient.cc
+++ b/DQM/EcalMonitorClient/src/DQWorkerClient.cc
@@ -176,7 +176,6 @@ namespace ecaldqm
       int i = 0;
       while (auto me = meset.second->getME(i)) {
         if (me->getLumiFlag()) {
-          std::cout << "+++ reset " << me->getFullname() << "\n";
           // reset per-lumi histograms in offline harvesting so that they only show
           // data of the current lumisection.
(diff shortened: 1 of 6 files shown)


jfernan2:
please test

jfernan2:
+1

fabiocos:
for what I can test the added scripts look functional

fabiocos:
+1

PR #28005, opened 2019-09-16
improve upgrade PU assignment

kpedro88:
#### PR description:

In #25996, Phase 2 PU workflows were configured to use 200PU automatically. However, this configuration relied on the name of the default dataset, which a) isn't always set and b) no longer necessarily matches the workflow year because of the recent 2023 -> 2026 change in #27488. Instead, the year can be matched to the key directly.

#### PR validation:

Tested runTheMatrix.py to make sure the right PU configuration was assigned to each year.


kpedro88:
please test

chayanit:
+1

kpedro88:
+upgrade

fabiocos:
+1

PR #28010, opened 2019-09-16
Upgrade CondDBESSource to support concurrent IOVs

wddgit:
#### PR description:

The most significant part of this involves replicating
the PayloadProxy object so there is one per IOV and
moving the IOVProxy object out of PayloadProxy into
DataProxyWrapper so there can only be one of those
per record type. When each new IOV actually becomes
active the IOV range gets copied into the PayloadProxy.

Except for a few tests, the number of concurrent IOVs
will still be configured to be 1 everywhere so nothing
will run with concurrent IOVs even after this PR is
merged. I would suggest additional testing and validation
be done before concurrent IOVs are used in production.
And also some thought will be necessary to determine
where the performance gain justifies the extra memory used
by the concurrent IOVs.

#### PR validation:

A new unit test is added that runs CondDBESSource with
multiple IOVs being processed concurrently, reads data
from the database into BeamSpot objects, and compares
that data with a reference file to verify the values
read remain the same.

Not a backport and should not be backported.


wddgit:
please test

FYI @Dr15Jones @makortel

wddgit:
I was wondering how the reviews of this are progressing? I am happy to make changes in response to reviewer comments or answer questions. When can I expect to get some feedback? I'll rebase this on top of other work if necessary... If this fails some test, I would be happy to debug problems if you tell me how to reproduce them.

fabiocos:
@ggovi @christopheralanwest @pohsun this PR seems to be overlooked, could you please check it and comment in case?

fabiocos:
please test

to refresh the test after 3 weeks

ggovi:
@wddgit I must confess that at a first glance I could not fully understand the need/the logic of all these changes. Also, this part was not covered by our last discussion in the AlCa meeting, if I am not mistaken.

wddgit:
Hi @ggovi ,
If you look at the presentation on slide 6, titled "2 PRs" it explains that the plan was two have separate PRs: one PR to implement Core support for concurrent IOVs and a second PR to migrate CondDBESSource to support concurrent IOVs. PR 26592 has since been merged and implements Core support. The only changes in CondCore in PR 26592 were the minimal interface changes that had to go in at the same time as the Core changes. After PR 26592  was merged, CondDBESSource was still only capable of handling one IOV at a time. The presentation included more discussion of this PR on other slides.

The main issue is that if multiple IOVs are running simultaneously, then there must be multiple copies of the data objects in memory at the same time. There is no way around this. One of the design goals that influenced the details of my implementation was an effort to minimize changes to the CondDBESSource code. But significant changes had to be made to replicate those data objects for each simultaneously running IOV. And the rest of it is necessary to make that replication work properly and concurrently.

These changes in CondDBESSource are critical for concurrent IOVs. Without it, Core support for concurrent IOVs will just be a useless unused Framework capability. CondDBESSource is really the source for IOVs in most processes using the EventSetup.

wddgit:
please test

I rebased the PR to pick up a Core bug fix that was recently merged and also because this was just in general based on a commit that is old.

I also fixed a minor problem in the unit test that this PR adds. The problem was completely internal to the unit test. The code changes in the outside the unit test are exactly the same as before I pushed the latest commit doing the rebase.

wddgit:
I have recently also run a reco step3 RelVal test with the number of concurrent IOVs set to 2. The job ran successfully to completion without any obvious problem (like a crash, set fault...). We still need more tests before we increase this parameter value in production jobs. Only one record actually crossed an IOV boundary with the data files used and the results were not validated, but still this is a good sign.

wddgit:
The following plots show the difference this PR will make. They both plot the number of modules running concurrently at each moment in time in a process configured to use 16 threads. They both show reconstruction (step 3) of runTheMatrix.py test 136.849 modified to remove the parts which do not support concurrent lumis (mainly DQM) and with the output module replaced by the asciioutput module. In the first plot the number of IOVs that can run concurrently is restricted to 1. In the second plot, 2 IOVs are allowed to run concurrently. There is a dip between 80 and 100 seconds in the first plot at an IOV boundary. Just one record having an IOV change caused this (BeamSpot related). The second plot shows this dip vanishes.

final1x

final206x

Note the disappearance of the dip is the important thing the plot shows. I wouldn't pay attention to the total running time as the conditions on the machine and statistics were not good enough for that to be a meaningful measurement. One cannot derive a single number that is the efficiency improvement related to this change because  it is so dependent on the use case. For example, increasing the number of threads would significantly increase the efficiency improvement. Or selecting input files with different numbers of events per luminosity block. The plots above both select input from the configuration as follows:
```
lumisToProcess = cms.untracked.VLuminosityBlockRange('315489:7-315489:8')
```
The importance of the changes introduced by this PR will be greater in the future, because we expect the number of threads to increase and also we expect to have more IOV boundaries inside runs.

We are still waiting for comments or approval from reviewers on this PR...

fabiocos:
@wddgit thanks for the documentation.

@ggovi @christopheralanwest how are you planning to proceed here? Do you want more information, is there a discussion planned? Or are there point comments to the code? This PR is now pending since one moth and half, we need to figure out how to move forward.

ggovi:
For what I'am concerned, I sign it. It was probably quite complicate to understand how to adapt the existing code. Personally I would have attempted a strategy more 'on top' of the PayloadProxy - but I've no idea if this would have been possible at all. In any case,  I consider this one a clever piece of work. Thanks @wddgit

ggovi:
+1

fabiocos:
@christopheralanwest @tlampen @pohsun this PR has now been signed by @ggovi , do you have any further comment/doubt about it? Otherwise I would like to integrated it before next pre-release is made (as this is a long pending development)

christopheralanwest:
please test

In workflow 11634.0 there are differences in reconstructed quantities (photon conversions and PF link data). For example, see the plot below:

EgammaV__ConversionValidator_ConversionInfo_nConvAll

I honestly don't see how this could be an effect of this PR, given that MC uses only one IOV. Was there some other known source of irreproducibility in the master branch at the time that the tests were run 11 days ago?

In any case, I'm re-running the tests to see if the differences persist.

fabiocos:
@christopheralanwest I am afraid that the differences that you see are the manifestation of an unrelated problem to be fixed, see https://github.com/cms-sw/cmssw/issues/28072 I did not find for the time being the reason for this, to be looked at in greater detail

christopheralanwest:
@fabiocos Thanks for the explanation about workflow 11634.0.

@wddgit With this PR, is the code implementation for concurrent IOVs completed so that the concurrent IOV functionality can be turned on with python configuration only? Once this PR is included in a release and prior to the use of concurrent IOVs in production, I would like to have a dedicated set of RelVals generated, analogous to those for changes to gcc or ROOT versions. Given the potential for this PR to generate irreproducibilities in situations that are not tested by our usual workflows, I think that a large scale test is needed to ensure that there are no edge cases that are handled incorrectly.

wddgit:
Yes, concurrent IOVs will be ready to be turned on with only configuration changes. I agree it would be good to have the large scale tests you are describing. It has not been tested thoroughly in full blown production executables running many different ESProducers and ESSources. More testing and validation needs to be done before we modify production configurations to run multiple IOVs concurrently. So far the tests include unit tests with test modules. Plus I have run one full reco configuration, but it only crossed an IOV boundary for one record in the data I used. It is hard to find data with IOV boundaries inside runs. We do not have the capability to have multiple runs execute concurrently yet and this makes it hard to create meaningful tests.

One thing that is important to keep in mind is that people are working on features that will create IOV boundaries within runs more often in the future. This will make it easier to test and also will make the new ability added by this PR more valuable in the future.

And while I would be willing to try if asked, I am probably not the best person to be setting up those tests. I am an expert in the details of the Core code, but probably the skills and experience for generating those tests lies elsewhere. I might be the best person to debug problems that are found.

One other comment, there is a bug related to concurrent lumis that I am working on fixing today. It is not a concurrent IOV problem and is already in code merged into CMSSW. It affects LuminosityBlockSummaryCache when run with concurrent lumis. It is independent of this PR, but if you try to test this PR you might hit it.

christopheralanwest:
+1

However, prior to activation of this feature, it seems that a large-scale test is needed.

fabiocos:
@Dr15Jones @wddgit I will integrate this PR first, as requested. I think we need to clarify with PPD (@srimanob) and PdmV (@chayanit @pgunnell) how to best manage a test of this before activating it in production. I am not sure that a relval-like test with a few events crossing a lumi boundary (as we added already in PyReleaseValidation) is useful in this case.

fabiocos:
+1

wddgit:
A couple more comments about turning concurrent IOVs on in production. This can be enabled one record at a time. So if we identified some small set of records whose IOVs changed inside runs and were causing inefficiencies, then we could enable concurrent IOVs only for that small set of records.

And we can also work the other way. We could set the parameter to 2 for all records, except leave it at 1 for the records we know have a problem with concurrent IOVs or use lots of memory or just will never have more than 1 IOV. (And this can be either put in the configuration or hard coded in C++.)

I don't know what the best way forward is.

I checked the IBs after merging. I do not see any problems with the parameter set to 1. So far so good.

CondCore/ESSources/test/TestConcurrentIOVsCondCore_cfg.py
```diff
@@ -0,0 +1,78 @@
...
+        messageLevel = cms.untracked.int32(0),
+        security = cms.untracked.string('')
+    ),
+    DumpStat = cms.untracked.bool(False),
+    ReconnectEachRun = cms.untracked.bool(False),
+    RefreshAlways = cms.untracked.bool(False),
+    RefreshEachRun = cms.untracked.bool(False),
+    RefreshOpenIOVs = cms.untracked.bool(False),
+    connect = cms.string('frontier://FrontierProd/CMS_CONDITIONS'),
+    globaltag = cms.string('110X_dataRun2_v5'),
```
mmusich, line 69: @wddgit @christopheralanwest @ggovi 
I am wondering if it is intentional to hardcode a GT name here. 
Whenever a record present in this GT will be removed from CMSSW, that will cause errors in this unit test (see e.g.: https://github.com/cms-sw/cmssw/pull/31729#issuecomment-708664358).
For this use case, it seems to me that's an overkill. Would not it be sufficient to just access on multi-IOVed per-LS record BeamSpot tag such as in:
```diff
diff --git a/CondCore/ESSources/test/TestConcurrentIOVsCondCore_cfg.py b/CondCore/ESSources/test/TestConcurrentIOVsCondCore_cfg.py
index c6aed33355c..7cfecb62e9b 100644
--- a/CondCore/ESSources/test/TestConcurrentIOVsCondCore_cfg.py
+++ b/CondCore/ESSources/test/TestConcurrentIOVsCondCore_cfg.py
@@ -66,12 +66,15 @@ process.GlobalTag = cms.ESSource("PoolDBESSource",
     RefreshEachRun = cms.untracked.bool(False),
     RefreshOpenIOVs = cms.untracked.bool(False),
     connect = cms.string('frontier://FrontierProd/CMS_CONDITIONS'),
-    globaltag = cms.string('110X_dataRun2_v5'),
+    globaltag = cms.string(''),
     pfnPostfix = cms.untracked.string(''),
     pfnPrefix = cms.untracked.string(''),
-    snapshotTime = cms.string(''),
-    toGet = cms.VPSet()
-)
+    #snapshotTime = cms.string(''),
+    snapshotTime = cms.string('2020-10-10 00:00:00.000'),
+    toGet = cms.VPSet(cms.VPSet(cms.PSet(record = cms.string("BeamSpotObjectsRcd"),
+                                         tag = cms.string("BeamSpotObjects_2017UL_LumiBased_v2")
+                                         )))
+    )
```
?
IFF there are concerns that the tag content might change over time (unlikely, but possible due to accidental overwritings) I believe one can leverage the `snapshotTime` of the `ESSource`.

wddgit: I'll submit a pull request to fix this. Thanks for debugging this. I did not know global tags could become obsolete in this way. As long as your changes lead to the same data being read from the database for run 132598 (selected because I know the values in the database are different for different lumis), then it should work just fine for this unit test. @mmusich Do you prefer the empty snapshotTime string or the non-empty one above?

mmusich: Thanks.
> Do you prefer the empty snapshotTime string or the non-empty one above?

Probably non-empty is safer, but the ultimate decision should come from @ggovi

PR #28028, opened 2019-09-19
Add APV simulation in PreMixingSiStripWorker

pieterdavid:
#### PR description:

This adds the new APV simulation from https://github.com/cms-sw/cmssw/pull/27823 also to the PreMixingSiStripWorker, which is used instead of the SiStripDigitizer when premixing is used.
It goes together with the configuration changes in https://github.com/cms-sw/cmssw/pull/28024 .

#### PR validation:

by @mmusich, see https://github.com/cms-sw/cmssw/pull/28024 : when enabling the APV simulation at the data mixing step (and not when making the premixing pileup library) the cluster charge distribution matches with the non-premixing scenario.

CC: @EmyrClement

### DIFF
--- a/SimTracker/SiStripDigitizer/plugins/PreMixingSiStripWorker.cc
+++ b/SimTracker/SiStripDigitizer/plugins/PreMixingSiStripWorker.cc
@@ -6,6 +6,7 @@
 #include "FWCore/ServiceRegistry/interface/Service.h"
 #include "DataFormats/Common/interface/Handle.h"
 #include "SimGeneral/MixingModule/interface/PileUpEventPrincipal.h"
+#include "SimDataFormats/PileupSummaryInfo/interface/PileupSummaryInfo.h"
 #include "CondFormats/SiStripObjects/interface/SiStripBadStrip.h"
 #include "CalibTracker/Records/interface/SiStripDependentRecords.h"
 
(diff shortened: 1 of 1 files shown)


SimTracker/SiStripDigitizer/plugins/PreMixingSiStripWorker.cc
```diff
@@ -339,6 +357,21 @@ void PreMixingSiStripWorker::put(edm::Event& e,
...
 
+  const bool simulateAPVInThisEvent = includeAPVSimulation_ && (CLHEP::RandFlat::shoot(engine) < fracOfEventsToSimAPV_);
+  float nTruePU = 0.;  // = ps.getTrueNumInteractions();
+  if (simulateAPVInThisEvent) {
+    iSetup.get<TrackerTopologyRcd>().get(tTopo);
+    iSetup.get<SiStripApvSimulationParametersRcd>().get(apvSimulationParametersHandle);
+    const auto it = std::find_if(
+        std::begin(ps), std::end(ps), [](const PileupSummaryInfo& bxps) { return bxps.getBunchCrossing() == 0; });
+    if (it != std::begin(ps)) {
+      nTruePU = it->getTrueNumInteractions();
```
pieterdavid: Since this is the main difference with the SiStripDigitizer implementation: is this the correct way to get the average number of interactions (Poisson parameter) for the in-time bunch-crossing? The original code is here: https://github.com/cms-sw/cmssw/blob/master/SimTracker/SiStripDigitizer/plugins/SiStripDigitizerAlgorithm.cc#L249-L281 but the same object is not available here (or I did not immediately find a way to get it)

mmusich:
For the record, I attach here couple of plots obtained running 200 events using the wfs introduced in #28024 using the release and the changes proposed here:
image

image


(black is 25202.1, red is 250202.2 in the two cases).
The discrepancy between standard mixing and pre-mixing is recovered.

fabiocos:
please test

civanch:
+1

SimTracker/SiStripDigitizer/plugins/PreMixingSiStripWorker.cc
makortel: @pieterdavid In short: yes, it is.

(the `ps` is the `vector<PileupSummaryInfo>` from the premixed pileup event)

fabiocos:
+1

ready to enter together #28024

SimTracker/SiStripDigitizer/plugins/PreMixingSiStripWorker.cc
pieterdavid: @makortel thanks for confirming!

PR #28030, opened 2019-09-19
Default LCT type is ALCTCLCT

dildick:
#### PR description:

This PR changes the default LCT type from `CLCTALCT` (CLCT-centric) to `ALCTCLCT` (ALCT-centric). I also added the LCT type to the print-out operator.

#### PR validation:

I tested it with a single muon gun using `cmsDriver.py SingleMuPt100_pythia8_cfi.py --conditions auto:phase2_realistic -n 10 --era Phase2 --eventcontent FEVTDEBUG -s GEN,SIM --datatier GEN-SIM --beamspot HLLHC --geometry Extended2023D41 --python SingleMuPt100_2023tilted_GenSimFull.py --no_exec --fileout file:step1.root`

FYI @tahuang1991

### DIFF
--- a/DataFormats/CSCDigi/interface/CSCCorrelatedLCTDigi.h
+++ b/DataFormats/CSCDigi/interface/CSCCorrelatedLCTDigi.h
@@ -3,7 +3,7 @@
 
 /**\class CSCCorrelatedLCTDigi
  *
- * Digi for Correlated LCT trigger primitives. 
+ * Digi for Correlated LCT trigger primitives.
  *
  *
  * \author L. Gray, UF
@@ -120,8 +120,8 @@ public:
 
   /// SIMULATION ONLY ////
   enum Type {
+    ALCTCLCT,      // ALCT-centric (default)
     CLCTALCT,      // CLCT-centric
-    ALCTCLCT,      // ALCT-centric
     ALCTCLCTGEM,   // ALCT-CLCT-1 GEM pad
     ALCTCLCT2GEM,  // ALCT-CLCT-2 GEM pads in coincidence
     ALCT2GEM,      // ALCT-2 GEM pads in coincidence
--- a/DataFormats/CSCDigi/src/CSCCorrelatedLCTDigi.cc
+++ b/DataFormats/CSCDigi/src/CSCCorrelatedLCTDigi.cc
@@ -81,10 +81,17 @@ void CSCCorrelatedLCTDigi::print() const {
 }
 
 std::ostream& operator<<(std::ostream& o, const CSCCorrelatedLCTDigi& digi) {
-  return o << "CSC LCT #" << digi.getTrknmb() << ": Valid = " << digi.isValid() << " Quality = " << digi.getQuality()
-           << " MPC Link = " << digi.getMPCLink() << " cscID = " << digi.getCSCID() << "\n"
-           << "  cathode info: Strip = " << digi.getStrip() << " Pattern = " << digi.getPattern()
+  return o << "CSC LCT #" << digi.getTrknmb()
+           << ": Valid = " << digi.isValid()
+           << " Quality = " << digi.getQuality()
+           << " MPC Link = " << digi.getMPCLink()
+           << " cscID = " << digi.getCSCID()
+           << " type = " << digi.getType() << "\n"
+           << "   cathode info: Strip = " << digi.getStrip()
+           << " Pattern = " << digi.getPattern()
            << " Bend = " << ((digi.getBend() == 0) ? 'L' : 'R') << "\n"
-           << "    anode info: Key wire = " << digi.getKeyWG() << " BX = " << digi.getBX() << " bx0 = " << digi.getBX0()
-           << " syncErr = " << digi.getSyncErr() << "\n";
+           << "   anode info: Key wire = " << digi.getKeyWG()
+           << " BX = " << digi.getBX()
+           << " bx0 = " << digi.getBX0()
+           << " syncErr = " << digi.getSyncErr() <<"\n";
 }
--- a/DataFormats/CSCDigi/src/CSCCorrelatedLCTDigi.cc
+++ b/DataFormats/CSCDigi/src/CSCCorrelatedLCTDigi.cc
@@ -81,17 +81,11 @@ void CSCCorrelatedLCTDigi::print() const {
 }
 
 std::ostream& operator<<(std::ostream& o, const CSCCorrelatedLCTDigi& digi) {
-  return o << "CSC LCT #" << digi.getTrknmb()
-           << ": Valid = " << digi.isValid()
-           << " Quality = " << digi.getQuality()
-           << " MPC Link = " << digi.getMPCLink()
-           << " cscID = " << digi.getCSCID()
-           << " type = " << digi.getType() << "\n"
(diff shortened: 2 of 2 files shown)


fabiocos:
@dildick which is the expect impact of this change in accessing old data?

fabiocos:
please test

dildick:
@fabiocos There is no impact on data. The LCT `type` is only used in simulation to understand and improve the performance of the local trigger algorithm.

civanch:
+1

civanch:
@dildick , is there any impact on legacy Run-2 simulation?

davidlange6:
hi
I had a similar concern - this change naively can break lots of things, in particular any data that has been saved persistently. 

Is there an example bit of code where the first item in this list of enums is important?

fabiocos:
@dildick where is this ```Type``` used? To fill the value of the private member ```type_``` I would say, for instance looking at
https://cmssdt.cern.ch/lxr/source/L1Trigger/CSCTriggerPrimitives/src/CSCGEMMotherboard.cc#0189

You say that this member is not used for data, and meaningful only for simulation. But this object is nevertheless stored persistently in the simulation output, is it? So when reading back old simulation digis this PR would change the meaning of that data member, because the same integer value now points to a different type, wrt older releases, and therefore has a different meaning and expected code behaviour associated, am I correct?

ptcox:
Hi,

I'm a bit confused about this. Digis are not persistent, surely? (It was 
rather traumatic for us when they stopped being persistent!)

But the fact that this enum is used to set a 'type' does worry me a bit, 
because clearly some 'output' of the trigger primitive  simulation has 
to feed downstream into the L1 trigger simulation, and it seems likely 
the digis are used to make that transfer - not persistently, but 
transiently within the same job.

  I hope Sven will confirm that the L1 trigger simulation does not make 
use of this 'type'.

Regards,

Tim

Fabio Cossutti wrote on 9/23/19 18:20:

dildick:
No, as far as I know, the L1 trigger emulator does not make use of the LCT type information. This data member was added relatively recently by me to understand how frequently LCTs of a certain type are produced in the CSC local trigger. This is something that we have yet to understand fully. That data member allows precisely that. The type must not be used in the L1-trigger simulation anywhere, as the type is not contained in the online 32-bit LCT dataformat. Consequently, the LCT type is not properly defined when running the LCT unpacker on RAW data. According to Andrew Brinkerhoff and Jia Fu Low, the LCT type is not used any part of the EMTF emulator our offline analysis.

dildick:
So this data member should really only be used in the offline analysis of the CSC local trigger performance.

fabiocos:
@dildick @ptcox in the FEVTDEBUGHLT formst (see the output of step2 in test workflow 10824.0 just as an example) I do see:

```
15:50 cmsdev25 6444> edmDumpEventContent step2.root | grep CSCCorrelatedLCTDigi
MuonDigiCollection<CSCDetId,CSCCorrelatedLCTDigi>    "simCscTriggerPrimitiveDigis"   ""                "HLT"     
MuonDigiCollection<CSCDetId,CSCCorrelatedLCTDigi>    "simCscTriggerPrimitiveDigis"   "MPCSORTED"       "HLT"     
```

so in these output the modified data format is used, right? Ok, this is mostly for test and debug purposes, but we have it. So this PR in my understanding is changing the meaning of  a data type that is persistently written, at least for test/debug purposes. Whether this is used elsewhere or not I do not know, according to Sven it is not (@rekovic @benkrikler FYI). In this limit it could be ok, provided that we know that at some point is meaning is changing. But at this point, following https://twiki.cern.ch/twiki/bin/view/CMSPublic/SWGuideCreatingNewProducts#Class_versioning I wonder whether we should not change the class version number in the corresponding classes_def.xml , as basically the meaning of a persistently stored type is changing @davidlange6 @Dr15Jones

fabiocos:
@dildick @ptcox @Dr15Jones any comment to my suggestion to have a new class version in https://cmssdt.cern.ch/lxr/source/DataFormats/CSCDigi/src/classes_def.xml#0014 for CSCCorrelatedLCTDigi ? The meaning of the type is essentially modified.

davidlange6:
@fabiocos - since the member data are not altered, i don't think there is a way to change the class version (happy to be corrected...)

what is the actual benefit of this change? Variables of type Type are somewhere initialized to 0 (eg, the first element)?

fabiocos:
@davidlange6 IMHO it depends whether the redefinition of an enum type used t define a data member is considered to alter the data member or not, if not of course the class version bump is not an option. Still the meaning of the member is altered in a set of data tiers that are persistently stored, although only for test and debugging purposes

davidlange6:
if root considered this, then the cmssw code would not compile, failing with an error that the class version should be updated with a new checksum..

Dr15Jones:
> if root considered this, then the cmssw code would not compile, failing with an error that the class version should be updated with a new checksum..

I agree with David, if our edmCheckClassVersion doesn't say the version number must change then there is no reason to change the version number.

fabiocos:
@davidlange6 @Dr15Jones thank you for your comments, so from the root/edm point of view nothing changes and there is no version update possible. In practice this PR swaps the numerical value associated with ```CSCCorrelatedLCTDigi::CLCTALCT``` and ```CSCCorrelatedLCTDigi::ALCTCLCT ```.
I see that the ```getType()``` method of this class is used nowhere within CMSSW, and the Type itself is just mentioned in https://cmssdt.cern.ch/lxr/source/L1Trigger/CSCTriggerPrimitives/src/CSCMotherboard.cc#0184 and other places in this class. From the discussion I get that the practical effect will be seen only by people studying outside of CMSSW workflows the CSC trigger primitives from the MC emulator: everything will be self-consistent in 11_0_X or in older releases, but looking at a ```FETVDEBUGHLT``` output of RelVals in older releases with the updated code will provide an interpretation of the type stored that is not what originally meant.

Without a version of the DataFormat there is no way to realize this from the data itself. The only argument I get from @dildick and @ptcox is that this information is basically not used but by CSC experts which should be aware of the change. Either we buy this and merge the PR, or I do not see an easy way out.

Retriggering the question of David, what is the actual benefit of this change? Variables of type Type are somewhere initialized to 0 (eg, the first element)?

fabiocos:
This type was introduced back in Summer 2017 https://github.com/cms-sw/cmssw/commit/caf49e53a94f63e2735c3a25002c13a269df7736#diff-412da00fbfe38f474e2cb9c6e0318560

fabiocos:
I had a discussion with @dildick to clarify how to possibly move forward with this PR, and the outcome in my understanding is:

- the change would be convenient for the developers, but strictly speaking not needed;

- anyway further changes in the meaning of data members (the quality word) is foreseen, due to the firmware evolution, but again it would be a change that does not trigger any class version update (from the point of view of root the format does not move);

- the simple-minded solution I proposed to keep track of these changes of meaning of data members in the format itself is to add a version number data member and corresponding setter/getter methods that would allow any user to keep track in next future about the possible changes of meaning of various parts of this format in an independent way on the historical memory and/or external documentation of the code.

dildick:
@fabiocos So let's close this PR for now until we can agree on a solution to track changes in the LCT data format for which ROOT is agnostic.

PR #28031, opened 2019-09-20
Python3: Changes needed to build cmssw with python3 as default

smuzaffar:
These changes are needed in order to build cmssw with python3 set as default python

### DIFF
--- a/CondFormats/Serialization/python/condformats_serialization_generate.py
+++ b/CondFormats/Serialization/python/condformats_serialization_generate.py
@@ -1,4 +1,5 @@
 #!/usr/bin/env python
+from __future__ import print_function
 '''CMS Conditions DB Serialization generator.
 
 Generates the non-intrusive serialization code required for the classes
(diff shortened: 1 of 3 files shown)


smuzaffar:
please test

fabiocos:
@ggovi could you please check the DB part? As far as I can see the main change is the need to specify the decoding of strings (@smuzaffar do you confirm ? This looks a sizeable difference compared to python2)

smuzaffar:
@fabiocos @govvi,  yes we have some files with special characters in comments  ( e.g files like these https://github.com/cms-sw/cmssw/blob/da232d4850c59deb4e0016a4dea376a24acbbd84/CalibFormats/SiPixelObjects/src/PixelBase64.cc#L4 ) so python3 fails to read these types of files in to string. That is why we need encoding. Note that all these special characters are in comments sections so this does not change the generated source file by condformats_serialization_generate.py .

ggovi:
@smuzaffar thanks a lot.

ggovi:
+1

smuzaffar:
@Dr15Jones , any objection on these changes?

fabiocos:
+1

PR #28033, opened 2019-09-20
Fix a discrepancy in Puppi weights when running on MiniAOD w/ useExistingWeights=False

hqucms:
#### PR description:

When running Puppi on MiniAOD w/ `useExistingWeights=False`, in some cases there is a discrepancy between the puppiWeight and the rescaled pT (i.e., rescaled pT != original pT * packCand->puppiWeight()). This is tracked to:

https://github.com/cms-sw/cmssw/blob/fb7c976b0bd6784c61295852158c36675939be56/CommonTools/PileupAlgos/plugins/PuppiProducer.cc#L275-L291

on L281 and L283, `iPuppiMatched` should not be used to access the puppi weight from `lWeights`, as `lWeights` is 1-to-1 with `hPFProduct` -- instead `val` is the correct index. 

As this only affects the puppi weights set for the new packedCandidates, this discrepancy does not affect cases when `useExistingWeights=True`, or running on RECO/AOD inputs.

This PR fixes this discrepancy and also simplifies the logic to compute the Puppi-scaled p4.

@ahinzmann

### DIFF
--- a/CommonTools/PileupAlgos/interface/PuppiContainer.h
+++ b/CommonTools/PileupAlgos/interface/PuppiContainer.h
@@ -20,10 +20,8 @@ public:
   // const std::vector<double> puppiAlpha   () {return fAlpha;}
   const std::vector<double> &puppiAlphasMed() { return fAlphaMed; }
   const std::vector<double> &puppiAlphasRMS() { return fAlphaRMS; }
-  const std::vector<int> &recoToPup() const { return fRecoToPup; }
 
   int puppiNAlgos() { return fNAlgos; }
-  std::vector<PuppiCandidate> const &puppiParticles() const { return fPupParticles; }
 
 protected:
   double goodVar(PuppiCandidate const &iPart, std::vector<PuppiCandidate> const &iParts, int iOpt, const double iRCone);
@@ -46,13 +44,11 @@ protected:
   const std::vector<RecoObj> *fRecoParticles;
   std::vector<PuppiCandidate> fPFParticles;
   std::vector<PuppiCandidate> fChargedPV;
-  std::vector<PuppiCandidate> fPupParticles;
   std::vector<double> fWeights;
   std::vector<double> fVals;
   std::vector<double> fRawAlphas;
   std::vector<double> fAlphaMed;
   std::vector<double> fAlphaRMS;
-  std::vector<int> fRecoToPup;
 
   bool fApplyCHS;
   bool fInvert;
--- a/CommonTools/PileupAlgos/plugins/PuppiProducer.cc
+++ b/CommonTools/PileupAlgos/plugins/PuppiProducer.cc
@@ -200,11 +200,9 @@ void PuppiProducer::produce(edm::Event& iEvent, const edm::EventSetup& iSetup) {
   fPuppiContainer->setNPV(npv);
 
   std::vector<double> lWeights;
-  std::vector<PuppiCandidate> lCandidates;
   if (!fUseExistingWeights) {
     //Compute the weights and get the particles
     lWeights = fPuppiContainer->puppiWeights();
-    lCandidates = fPuppiContainer->puppiParticles();
   } else {
     //Use the existing weights
     int lPackCtr = 0;
@@ -223,12 +221,6 @@ void PuppiProducer::produce(edm::Event& iEvent, const edm::EventSetup& iSetup) {
         }
       }
       lWeights.push_back(curpupweight);
-      PuppiCandidate curjet(curpupweight * lPack->px(),
-                            curpupweight * lPack->py(),
-                            curpupweight * lPack->pz(),
-                            curpupweight * lPack->energy());
-      curjet.set_user_index(lPackCtr);
-      lCandidates.push_back(curjet);
       lPackCtr++;
     }
   }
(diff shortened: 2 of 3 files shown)


ahinzmann:
@hqucms this would be a nice simplification.
I wonder if this also works for puppiNoLeptons and puppiForMET, where the particle list is shorter than the original list of candidates, because leptons are removed before running PUPPI?

hqucms:
@ahinzmann I think nothing should break as the weight list `lWeights` is always equal to the input candidate list, right? The output particle/p4 list (`lCandidates`) is not needed after the simplification.

ahinzmann:
Ok. Just since I got paranoid with this piece of code, could you make a comparison of puppiMET:
1. running makePuppies on AOD (just like in the official MiniAOD sequence)
2. running makePuppiesFromMiniAOD on MiniAOD with useExistingWeights=True
3. running makePuppiesFromMiniAOD on MiniAOD with useExistingWeights=False
I believe 1. and 2. should be part of some validation workflows, but 3. is not.

hqucms:
@ahinzmann Since 1 and 2 are in the PR validation workflow, maybe we can first trigger a PR test to check them.

perrotta:
please test

hqucms:
> Ok. Just since I got paranoid with this piece of code, could you make a comparison of puppiMET and the weights of the puppiForMET particles:
> 
> 1. running makePuppies on AOD (just like in the official MiniAOD sequence)
> 2. running makePuppiesFromMiniAOD on MiniAOD with useExistingWeights=True
> 3. running makePuppiesFromMiniAOD on MiniAOD with useExistingWeights=False
>    I believe 1. and 2. should be part of some validation workflows, but 3. is not.
> 
> In fact it would be nice to have example config for those three cases in the /test folder as well. Currently there is only a test for 2.

@ahinzmann

hqucms:
> Ok. Just since I got paranoid with this piece of code, could you make a comparison of puppiMET and the weights of the puppiForMET particles:
> 
> 1. running makePuppies on AOD (just like in the official MiniAOD sequence)
> 2. running makePuppiesFromMiniAOD on MiniAOD with useExistingWeights=True
> 3. running makePuppiesFromMiniAOD on MiniAOD with useExistingWeights=False
>    I believe 1. and 2. should be part of some validation workflows, but 3. is not.
> 
> In fact it would be nice to have example config for those three cases in the /test folder as well. Currently there is only a test for 2.

@ahinzmann 
I have run the tests on a JetHT 2017F data file for the 3 options and found no change in puppiMET with this PR [0]. I have not figured out how to access the weights for puppiForMET particles though -- it seems that they are not saved in the event content?

[0] https://cernbox.cern.ch/index.php/s/LUlVAKr3vSH47Cp

hqucms:
Regarding the small changes in the b-tagging discriminants in the PR test: they are caused by changes in numerical precision in the calculation of the puppi scaled p4. Before the PR, the candidate p4 is computed with double -> float (when [creating](https://github.com/cms-sw/cmssw/blob/fb7c976b0bd6784c61295852158c36675939be56/CommonTools/PileupAlgos/plugins/PuppiProducer.cc#L83-L86) the [RecoObj](https://github.com/cms-sw/cmssw/blob/fb7c976b0bd6784c61295852158c36675939be56/CommonTools/PileupAlgos/interface/RecoObj.h#L26)) -> [double](https://github.com/cms-sw/cmssw/blob/fb7c976b0bd6784c61295852158c36675939be56/CommonTools/PileupAlgos/plugins/PuppiProducer.cc#L278), now is double -> [double](https://github.com/cms-sw/cmssw/blob/fc452b10088d484e0ee133c4d558bd73e024ee7f/CommonTools/PileupAlgos/plugins/PuppiProducer.cc#L272-L275) without the intermediate conversion. Inserting an intermediate conversion to float can reproduce the old behavior exactly.

ahinzmann:
@hqucms The puppiForMET particles are indeed no saved. But the puppiNoLepton weights (which is a subset of puppiForMET) are stored in the packedCandidates.
Regarding b-tagging, I'd say we should go for the solution without the intermediate conversion.

slava77:
@ahinzmann
please clarify if your comments earlier in the thread were already addressed.
Thank you.

ahinzmann:
Looking at the three sumET plots in [0], useExistingWeights=True and False indeed look almost the same, however aod2mini looks very different from the other two. It doesn't seem caused by this PR, but certainly something that we need to follow up on.
@hqucms You plotted sumET, but in fact it's better to look at MET (simply called ".pt" in slimmedMETsPuppi)

[0] https://cernbox.cern.ch/index.php/s/LUlVAKr3vSH47Cp

slava77:
+1

for https://github.com/cms-sw/cmssw/pull/28033	fc452b1 
- jenkins tests pass and comparisons with the baseline show small differences in puppi-related variables in PAT jets and MET
    - local tests show that slimmedJetsPuppi have changes in their kinematics at numerical precision level (the pt stays the same, but eta changes within delta of 4e-6).

fabiocos:
+1

fabiocos:
merge

ahinzmann:
@hqucms @all
Could we backport this to 10_2 and 10_6? During PUPPI tuning activities we stumbled on the same issue and would like to make sure no JetMET analysis is affected by this. Even if production is not affected by this.

santocch:
+1

hqucms:
> @hqucms @ALL
> Could we backport this to 10_2 and 10_6? During PUPPI tuning activities we stumbled on the same issue and would like to make sure no JetMET analysis is affected by this. Even if production is not affected by this.

@ahinzmann 
Please see https://github.com/cms-sw/cmssw/pull/28169 and https://github.com/cms-sw/cmssw/pull/28170 for the backports. I made a minimal change to fix the weights, leaving out the code simplification part in this PR (which leads to some small changes in e.g., the b-tag discs at numerical precision level, due to the removal of some float--double conversion).

PR #28035, opened 2019-09-20
Simplify jet daughter processing in DeepBoostedJet

hqucms:
#### PR description:

This PR simplifies the treatment of Puppi jet daughters in DeepAK8. In the past, we rely on a postfix string to distinguish whether the AK8 jets have puppi scaled daughters or not, which is quite fragile. Now we unifies it to always first get the original candidate (via the `key()`, inspired by @raggleton's [RekeyJets](https://github.com/UHH2/UHH2/blob/RunII_102X_v1/core/plugins/RekeyJets.cc)) and then scaled it by the puppi weight. 

#### PR validation:

Tested in a boosted TT sample and no change in the discriminator outputs found.

### DIFF
--- a/PhysicsTools/PatAlgos/python/tools/jetTools.py
+++ b/PhysicsTools/PatAlgos/python/tools/jetTools.py
@@ -638,27 +638,15 @@ def setupBTagging(process, jetSource, pfCandidates, explicitJTA, pvSource, svSou
             if btagInfo == 'pfDeepBoostedJetTagInfos':
                 if pfCandidates.value() == 'packedPFCandidates':
                     # case 1: running over jets whose daughters are PackedCandidates (only via updateJetCollection for now)
-                    jetSrcName = jetSource.value().lower()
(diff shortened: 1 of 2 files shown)


slava77:
@cmsbuild please test

perrotta:
@hqucms ; can this be tested in any standard workflow, or do you have to run something manually? Please provide a recipe for testing, in case this is not run in one of the automatically tested workflows.

hqucms:
@perrotta DeepAK8 is run in the standard MiniAOD/NanoAOD workflows now so it should be covered.

perrotta:
+1
- It unifies the jet daughter processing, whether or not those daughters are already scaled by puppi weight; the implementation follows what is in the PR description
- Larger tests are reported in the PR description, where no change in the discriminator outputs are found

santocch:
+1

fabiocos:
+1

PR #28038, opened 2019-09-20
DQM: remove support for References inside CMSSW

schneiml:
#### PR description:

For the modernization of the core DQM infrastrucutre in CMSSW, we decided to drop some rarely used features. This PR removes the first of these: Reference histograms.

The goal is not to remove all of the related code (instead, we will benefit from not having to re-implement it in future), but to remove all usages to see if anything important relies on the references.

The main usage are the Kolmogorov and Chi2 based QTests, which are used in many places. However, since we did not have reasonable reference runs set for a long time now, I suspect that all the output based on these is essentially useless. This may not be true for some validation workflows, but then Validation often uses it's own implementation of comparisons.

Additionally, this PR removes some more old test cases, which happen to be the only usage of some features. 

#### PR validation:

Compiles, and seems to run. However, some references to removed histograms might remain. Also, we need to validate that indeed nobody relies on CMSSW-side references any more.


schneiml:
hold

I still see a problem in the local tests.

schneiml:
unhold

Looks ok now, need to check if any summaries show up empty.

schneiml:
please test

Let's see what happens.

schneiml:
please test

Let's try again.

Martin-Grunewald:
seems stuck, let's rerun tests

Martin-Grunewald:
please test

schneiml:
Ok, seems there are some more things to fix.

schneiml:
please test

now 10024 should work, and hopefully also all the others

schneiml:
The differences are mostly as expected: many quality tests are removed. A few summaries derived from them are affected as well, but we believe these values where not useful anyways. I asked for confirmation by the subsystem experts, waiting for responses.

jfernan2:
+1

Martin-Grunewald:
+1

fabiocos:
@peruzzim @fgolf @santocch please have a look

peruzzim:
Yes, I had a private mail exchange with the author of the PR to understand it better, but then my attention got diverted by other more urgent tasks.
I have not fully understood whether we are using the features removed by this PR for validating nanoAOD, could we please have a few more days to check the workflows after NanoAODv6 is prepared?

fabiocos:
@peruzzim I don't think there is a big urgency to integrate this PR, if it is just a matter of a few more days

peruzzim:
hold

After more discussion with @schneiml, it turns out we are actually using this feature to validate nanoAOD. We have devised a quick workaround and a longer term strategy to modify our validation workflow, I will unhold as soon as the first is implemented.

fioriNTU:
@peruzzim , for my education, could you point me to the code using the internal CMSSW reference for validating nanoAOD? I personally think that, for any possible use case, having a reference outside CMSSW is much more flexible. And, also, why nanoAOD can not be validated with RelMon?

peruzzim:
unhold

I have modified the nanoAOD validation script to avoid using references in CMSSW, by running up to the harvesting included without references, comparing the DQM histograms with reference ones only at the very end, and then filtering the DQM histogram file to only select only the changed histograms and upload them as a separate dataset to DQM.
We can have further discussion in a dedicated setting (meeting) if desired.

peruzzim:
+xpog
for the change to PhysicsTools/NanoAOD/test/dqmQualityTests.xml which removes functionality not supported anymore

santocch:
+1

fabiocos:
@peruzzim @schneiml I understand that further discussions about XPOG DQM needs may follow, but they are not a show-stopper for the integration of this PR, and that a solution in the private XPOG procedure is already in place. So I will move forward with the integration of this PR.

fabiocos:
+1

PR #28039, opened 2019-09-20
Fix problem with fromTrackerSeeds parameter in ElectronSeedProducer and improve LayerMeasurements class

### LINKED ISSUE #27541: Problem with FromTrackerSeeds parameter in ElectronSeedProducer
While looking at the workflow 134.804 step 3 in the UBSAN IB, it found that a nullptr was being dereferenced here

https://github.com/cms-sw/cmssw/blob/3bde531d2713e0745e278e0ccfec03891051be35/RecoEgamma/EgammaElectronAlgos/src/PixelHitMatcher.cc#L73

the variable with the nullptr was `theTracker`.

That value is set here
https://github.com/cms-sw/cmssw/blob/3bde531d2713e0745e278e0ccfec03891051be35/RecoEgamma/EgammaElectronAlgos/src/PixelHitMatcher.cc#L79

which is called from here
https://github.com/cms-sw/cmssw/blob/3bde531d2713e0745e278e0ccfec03891051be35/RecoEgamma/EgammaElectronAlgos/src/ElectronSeedGenerator.cc#L176-L177

where `theMeasurementTracker` is set here
https://github.com/cms-sw/cmssw/blob/3bde531d2713e0745e278e0ccfec03891051be35/RecoEgamma/EgammaElectronAlgos/src/ElectronSeedGenerator.cc#L160-L166

Note that `theMeasurementTracker` is only set if `fromTrackerSeeds_` is `false`. Which comes from the parameter
https://github.com/cms-sw/cmssw/blob/3bde531d2713e0745e278e0ccfec03891051be35/RecoEgamma/EgammaElectronAlgos/src/ElectronSeedGenerator.cc#L63

which comes in from 
https://github.com/cms-sw/cmssw/blob/1bbf498141e29ed4751ec80e7461a444a3a7055a/RecoEgamma/EgammaElectronProducers/plugins/ElectronSeedProducer.cc#L128

Given `LayerMeasurements` has no check that the values passed could be a nullptr, it is unclear why we do not get a segmentation fault when running this code.

guitargeek:
#### PR description:

This PR intends to fix https://github.com/cms-sw/cmssw/issues/27541.

If the `fromTrackerSeeds` parameter in the `ElectronSeedProducer` would be set to `false`, a segfault would occur, as detected in the UBSAN IB. This never happened in practice because  `fromTrackerSeeds = False` is actually not supported and triggers an exception [1].

We understand now that` fromTrackerSeeds` is never supposed to be `False` and therefore we can just remove all branches in the code for this case, fixing both the issue and simplifying the code.

When I tried to fix the issue, I initially didn't know how it could be fixed and did some changes with the goal of making the problem more obvious to me. This involved changing the `LayerMeasurements` class such that there are no pointers anymore which can be `nullptr` and therefore cause trouble when dereferenced. In retrospect, this has nothing to do with the actual issue, but I include this change anyway in the PR as I think it's an improvement.

This caused the interface of `LayerMeasurements` to change a little bit, and while adapting the code I found an unused function in some muon plugins which could be removed. It was apparently replaced by a more efficient function.

[1] https://github.com/cms-sw/cmssw/blob/master/RecoEgamma/EgammaElectronAlgos/src/ElectronSeedGenerator.cc#L275

#### PR validation:

No backports intended.

### DIFF
--- a/RecoEgamma/EgammaElectronAlgos/interface/ElectronSeedGenerator.h
+++ b/RecoEgamma/EgammaElectronAlgos/interface/ElectronSeedGenerator.h
@@ -93,11 +93,11 @@ private:
 
   const bool dynamicphiroad_;
   const bool fromTrackerSeeds_;
-  edm::Handle<std::vector<reco::Vertex> > theVertices;
-  edm::EDGetTokenT<std::vector<reco::Vertex> > verticesTag_;
+  edm::Handle<std::vector<reco::Vertex> > vertices_;
+  const edm::EDGetTokenT<std::vector<reco::Vertex> > verticesTag_;
 
-  edm::Handle<reco::BeamSpot> theBeamSpot;
-  edm::EDGetTokenT<reco::BeamSpot> beamSpotTag_;
+  edm::Handle<reco::BeamSpot> beamSpot_;
+  const edm::EDGetTokenT<reco::BeamSpot> beamSpotTag_;
 
   const float lowPtThreshold_;
   const float highPtThreshold_;
@@ -112,46 +112,43 @@ private:
   const double deltaPhi1Coef1_;
   const double deltaPhi1Coef2_;
 
-  const std::vector<const TrajectorySeedCollection*>* theInitialSeedCollV = nullptr;
+  const std::vector<const TrajectorySeedCollection*>* initialSeedCollectionVector_ = nullptr;
 
-  edm::ESHandle<MagneticField> theMagField;
-  edm::ESHandle<TrackerGeometry> theTrackerGeometry;
-  //edm::ESHandle<GeometricSearchTracker>       theGeomSearchTracker;
+  edm::ESHandle<MagneticField> magField_;
+  edm::ESHandle<TrackerGeometry> trackerGeometry_;
   KFUpdator updator_;
   std::unique_ptr<PropagatorWithMaterial> propagator_;
 
-  const MeasurementTracker* theMeasurementTracker;
-  edm::EDGetTokenT<MeasurementTrackerEvent> theMeasurementTrackerEventTag;
+  const MeasurementTracker* measurementTracker_;
+  const edm::EDGetTokenT<MeasurementTrackerEvent> measurementTrackerEventTag_;
 
-  const NavigationSchool* theNavigationSchool;
+  const NavigationSchool* navigationSchool_;
 
-  const edm::EventSetup* theSetup;
+  const edm::EventSetup* setup_;
 
   PRecHitContainer recHits_;
   PTrajectoryStateOnDet pts_;
 
   // keep cacheIds to get records only when necessary
   unsigned long long cacheIDMagField_;
-  //  unsigned long long cacheIDGeom_;
   unsigned long long cacheIDNavSchool_;
   unsigned long long cacheIDCkfComp_;
   unsigned long long cacheIDTrkGeom_;
 
-  std::string measurementTrackerName_;
+  const std::string measurementTrackerName_;
 
-  bool useRecoVertex_;
+  const bool useRecoVertex_;
 
-  float deltaPhi2B_;
-  float deltaPhi2F_;
+  const float deltaPhi2B_;
+  const float deltaPhi2F_;
 
-  float phiMin2B_;
-  float phiMin2F_;
-  float phiMax2B_;
-  float phiMax2F_;
+  const float phiMin2B_;
+  const float phiMin2F_;
+  const float phiMax2B_;
+  const float phiMax2F_;
 
   PixelHitMatcher electronMatcher_;
   PixelHitMatcher positronMatcher_;
-
 };
 
 #endif  // ElectronSeedGenerator_H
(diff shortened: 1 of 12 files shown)


slava77:
@cmsbuild please test

perrotta:
please test

perrotta:
code-checks

perrotta:
please test

RecoEgamma/EgammaElectronAlgos/src/ElectronSeedGenerator.cc
```diff
@@ -17,188 +17,112 @@
...
+      // new B/F configurables
+      deltaPhi2B_(pset.getParameter<double>("DeltaPhi2B")),
+      deltaPhi2F_(pset.getParameter<double>("DeltaPhi2F")),
+      phiMin2B_(pset.getParameter<double>("PhiMin2B")),
+      phiMin2F_(pset.getParameter<double>("PhiMin2F")),
+      phiMax2B_(pset.getParameter<double>("PhiMax2B")),
+      phiMax2F_(pset.getParameter<double>("PhiMax2F")),
+      electronMatcher_(pset.getParameter<double>("ePhiMin1"),
+                       pset.getParameter<double>("ePhiMax1"),
+                       phiMin2B_,
```
slava77: would std::bind be useful here to be less repetitive in electronMatcher_ and positronMatcher_?

RecoEgamma/EgammaElectronProducers/plugins/ElectronSeedProducer.cc
```diff
@@ -230,16 +230,13 @@ void ElectronSeedProducer::produce(edm::Event& e, const edm::EventSetup& iSetup)
...
-  std::unique_ptr<ElectronSeedCollection> pSeeds(seeds);
-  ElectronSeedCollection::iterator is;
-  for (is = pSeeds->begin(); is != pSeeds->end(); is++) {
-    edm::RefToBase<CaloCluster> caloCluster = is->caloCluster();
-    SuperClusterRef superCluster = caloCluster.castTo<SuperClusterRef>();
-    LogDebug("ElectronSeedProducer") << "new seed with " << (*is).nHits() << " hits"
-                                     << ", charge " << (*is).getCharge() << " and cluster energy "
+  for (auto const& seed : *seeds) {
+    SuperClusterRef superCluster = seed.caloCluster().castTo<SuperClusterRef>();
+    LogDebug("ElectronSeedProducer") << "new seed with " << seed.nHits() << " hits"
```
slava77: should the loop be placed in ifdef ```EDM_ML_DEBUG``` ? ... that's unless it's obvious that the compiler ignores this loop otherwise.

RecoEgamma/EgammaElectronAlgos/interface/ElectronSeedGenerator.h
```diff
@@ -38,9 +37,11 @@
 #include "FWCore/Framework/interface/ESHandle.h"
 #include "FWCore/Framework/interface/EDConsumerBase.h"
 
+#include "TrackingTools/KalmanUpdators/interface/KFUpdator.h"
+
 class PropagatorWithMaterial;
 class KFUpdator;
-class PixelHitMatcher;
+class PixelHitMatcher_;
```
slava77: is this used in this file?

guitargeek: No, that's an artefact of find-replace. Actually all the forward declarations can be removed.

RecoEgamma/EgammaElectronAlgos/src/ElectronSeedGenerator.cc
guitargeek: I thought of that too, but concluded it's a waste of time as it is probably only an intermediate solution. There is nothing that makes the electron and positron matcher really different except for a sign flip in the phi-window. IMHO one could just adapt the matcher class a bit and only use one instance after all. I'm afraid doing that would blow up this PR, however.

RecoEgamma/EgammaElectronProducers/plugins/ElectronSeedProducer.cc
guitargeek: I think it won't be ignored, I will do as you suggest.

guitargeek:
Hi @slava77, I addressed your comments and further did some trivial cleanup of the `ElectronSeedGenerator` class (removing unused member functions and converting private member functions that don't use the class member variables to free anonymous functions to make this explicit). You are right about the redundancy of the electron and positron matcher, maybe that could be streamlined in another PR eventually.

slava77:
@cmsbuild please test

@guitargeek thank you for the quick update

slava77:
+1

for https://github.com/cms-sw/cmssw/pull/28039	20455a4 
- code changes are in line with the PR description and the follow up review; the changes are technical and should not produce any differences in outputs

fabiocos:
@ssekmen @sbein could you please have a look? The fastsim part of this PR looks limited

sbein:
+1

fabiocos:
+1

PR #28042, opened 2019-09-20
DQM: Remove Tags.

schneiml:
#### PR description:

In the olden days, _tags_ could be associated with `MonitorElement`s for whatever purposes.

I fail to see any usage of tags today, so we will not support this in the future, and this PR removes the API as well as the I/O support. 

#### PR validation:
Runs through a simple workflow.

Special care must be taken to make sure that nothing in AlCa breaks; the code compiles fine without any of the get-by-tag APIs, so I suspect that actually none are used, but the SiStrip modules used for AlCa are among the few modules actually tagging MEs.

### DIFF
--- a/DQM/HLXMonitor/src/HLXMonitor.cc
+++ b/DQM/HLXMonitor/src/HLXMonitor.cc
@@ -169,8 +169,6 @@ void HLXMonitor::SetupHists(DQMStore::IBooker &iBooker) {
     Set2Above[iWedge] =
         iBooker.book1D("Set2_Above", "HF+ Wedge " + wedgeNum.str() + ": Above Threshold 2 - Set 2", NBINS, XMIN, XMAX);
     ETSum[iWedge] = iBooker.book1D("ETSum", "HF+ Wedge " + wedgeNum.str() + ": Transverse Energy", NBINS, XMIN, XMAX);
-
-    iBooker.tagContents(monitorName_ + "/HFPlus/Wedge" + tempStreamer.str(), iWedge + 1);
   }
 
   if (NUM_HLX > 17) {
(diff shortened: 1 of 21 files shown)


schneiml:
please test

Dr15Jones:
+1

mmusich:
> Special care must be taken to make sure that nothing in AlCa breaks; the code compiles fine without any of the get-by-tag APIs, so I suspect that actually none are used, but the SiStrip modules used for AlCa are among the few modules actually tagging MEs.

what's the plan for validating this? Should this PR acquire alca signature?

schneiml:
@mmusich first, lets see what the fallout in the PR tests is. I am surprised I did not touch any AlCa code (probably that is a good sign), but it might not hurt to get input from their side.

In general, however, this should not affect any code that does not explicitly call any `*tag*` methods, so we should be fine.

schneiml:
The DQMHistoSizes seems to be rather confused, but that is probably nothing to worry about. The main difference in the comparison is removing all the tag objects (as expected), and an Unexpected change in `EgammaV`. That one will need a follow up.

schneiml:
Checking other comparisons, it looks like `EgammaV/ConversionValidator` is unstable in workflow 11634 in general.

It should probably be fixed/blacklisted.

That means this PR seems fine, unless there are any other concerns.

jfernan2:
+1

fabiocos:
@schneiml @jfernan2 I would say it is worthwhile to cross check with @christopheralanwest and @mmusich whether this is ok from their point of view, following your comments

fabiocos:
+1

No problem was reported from the checks by @mmusich and other Tracker experts

fabiocos:
@smuzaffar is the bot stuck?

PR #28016, opened 2019-09-18
Updating Puppiproducer to include fillDescriptions

alefisico:
#### PR description:

To test PUPPI at HLT for Run3/PhaseII, PuppiProducer has to be updated with the fillDescription function to be recognized in condfb. This PR introduces this function in that producer. This change should be completely transparent to the PUPPI algorithm itself, or any other part of CMSSW.

#### PR validation:

Run `runTheMatrix.py -l limited -i all --ibeos`, and all the test passed. Here is the output of it: https://cernbox.cern.ch/index.php/s/QbCjBhba83NjFCr

PR #28048, opened 2019-09-22
HGCal - bugfix: Fix clamp range in tiling

felicepantaleo:
#### PR description:
The upper limit when using `std::clamp` on the number of bins should be the number of bins minus 1.
When hitting the very last bin, this could lead to SEGFAULT.

@rovere @ZihengChen

### DIFF
--- a/DataFormats/HGCalReco/interface/TICLLayerTile.h
+++ b/DataFormats/HGCalReco/interface/TICLLayerTile.h
@@ -18,7 +18,7 @@ public:
     static_assert(etaRange >= 0.f);
     float r = ticl::constants::nEtaBins / etaRange;
     int etaBin = (std::abs(eta) - ticl::constants::minEta) * r;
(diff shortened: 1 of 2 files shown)


felicepantaleo:
type bugfix

felicepantaleo:
please test workflow 20493.52

perrotta:
It is strange that in wf 11634.0 there are differences in the number of conversion photons
The very same was observed in PR #28016, and both here and there I don't see any possible reason for it, It looks like some non-reproducibility was recently introduced.
Let retry running the tests, just to verify if anything changed

perrotta:
please test workflow 20493.52

perrotta:
+1
- Fixes are correct
- Jenkins tests pass

kpedro88:
+upgrade

fabiocos:
+1

PR #28054, opened 2019-09-23
DDAngular: Minor bug fix of a debug message

cvuosalo:
#### PR description:

Minor debug message bug fix. The message was being incorrectly triggered because the index of an array rather than the array value was being checked. This PR puts in the correct check.

#### PR validation:

Test config was run to show the message no longer inappropriately appears.

No backport.


cvuosalo:
type bugfix

cvuosalo:
@cmsbuild please test

ianna:
+1

fabiocos:
+1

PR #28021, opened 2019-09-18
Python3: Changes needed to run runTheMatrix with python2/3

smuzaffar:
#### PR description:

Changes needed to run runTheMatrix with python3

#### PR validation:

Locally runing of runTheMatix (to dump relval commands) was successful. Order of command-line arguments to cmsDriver.py is not identical (because of change of `dict` behaviour in python3)

- **python2 runTheMatrix.py -n -e -l 137.8 -i all**
```
found a total of  794  workflows:
      of which the following 1 were selected:

137.8  RunEGamma2018C+HLTDR2_2018+RECODR2_2018reHLT_skimEGamma_Offline_L1TEgDQM+RunEGamma2018D+HLTDR2_2018+RECODR2_2018reHLT_skimEGamma_Prompt_L1TEgDQM+HARVEST2018_L1TEgDQM_MULTIRUN [1]: input from: /EGamma/Run2018C-v1/RAW with run [] 
                                           [2]: cmsDriver.py step2  --datatier FEVTDEBUGHLT --conditions auto:run2_hlt_relval -s L1REPACK:Full,HLT:@relval2018 --process reHLT --data  --era Run2_2018 --eventcontent FEVTDEBUGHLT -n 100 
                                           [3]: cmsDriver.py step3  --conditions auto:run2_data -s RAW2DIGI,L1Reco,RECO,SKIM:ZElectron+SinglePhotonJetPlusHOFilter+EXOMONOPOLE,EI,PAT,ALCA:SiStripCalZeroBias+SiStripCalMinBias+SiStripCalSmallBiasScan+TkAlMinBias+EcalESAlign,DQM:@standardDQMFakeHLT+@miniAODDQM+@L1TEgamma --runUnscheduled  --process reRECO --data  --era Run2_2018 --eventcontent RECO,MINIAOD,DQM --hltProcess reHLT --scenario pp --datatier RECO,MINIAOD,DQMIO --customise Configuration/DataProcessing/RecoTLR.customisePostEra_Run2_2018 -n 100 
                                           [4]: input from: /EGamma/Run2018D-v1/RAW with run []
                                           [5]: cmsDriver.py step5  --datatier FEVTDEBUGHLT --conditions auto:run2_hlt_relval -s L1REPACK:Full,HLT:@relval2018 --process reHLT --data  --era Run2_2018 --eventcontent FEVTDEBUGHLT -n 100 
[...]

PR #28057, opened 2019-09-24
[Python3][Devel] testing python3 in devel IBs

smuzaffar:
Same as #28021 but for DEVEL IBs to test https://github.com/cms-sw/cmsdist/pull/5241


Dr15Jones:
One of the failures is from here:
https://github.com/cms-sw/cmssw/blob/master/Configuration/Applications/python/ConfigBuilder.py#L146

Matti says to fix the problem one needs to add `universal_newlines=True` to the `Popen` call.

PR #28062, opened 2019-09-24
Replace boost::array with std::array

davidlange6:
#### PR description:

Replace boost::array with std::array

we do appear to store and use data of type HFPhase1PMTData, which is affected by this PR. both boost::array and std::array are written out as T[N] from what I see. The other CondFormats are not used afaik (but also should be fine in case there is some payload somewhere.)

#### PR validation:

compilation ok, 2018 workflows run ok. They should test all changes

### DIFF
--- a/CondFormats/HcalObjects/interface/HFPhase1PMTData.h
+++ b/CondFormats/HcalObjects/interface/HFPhase1PMTData.h
@@ -12,6 +12,8 @@
 #endif
 #include "CondFormats/HcalObjects/interface/AbsHcalFunctor.h"
 
+#include <array>
+
 class HFPhase1PMTData {
 public:
   // Functor enum for the cut shapes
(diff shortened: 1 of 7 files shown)


davidlange6:
please test

abdoulline:
Attn.  Igor @igv4321

davidlange6:
Indeed... I'm a bit puzzled why this didn't happen to me locally. I must have made a mistake.

CondFormats/HcalObjects/interface/HFPhase1PMTData.h
```diff
@@ -12,6 +12,8 @@
 #endif
```
slava77: should the line above be changed to ```#include "boost/serialization/array.hpp"``` or the full if/else dropped?

davidlange6:
will reopen when I have a chance to look into these issues..

PR #28077, opened 2019-09-26
AOD changes from JME

rappoccio:
#### PR description:

In the past, PUPPI has not been stable enough to go into AOD. It was always rerun at MINIAOD level. However BTV would like PUPPI in AOD so they can run costly algorithms at RECO time. In addition, AK8CHS is no longer recommended so there is no particular need to have it at all, much less in AOD. Finally, the CMS Top Tagger is now truly deprecated and only used in validation / DQM. As such, we have switched to monitory the newer top tagging algorithm.

In AOD:
- Add AK4Puppi
- Switch AK8CHS to AK8Puppi
- Drop CMS Top Tagger (deprecated) 

In MINIAOD:
- Add PFClusterJets p4's to the pat::Jets (in addition to / replacing current calo jets)

#### PR validation:

Workflows run: 

`- runTheMatrix.py -l limited -i all --ibeos`

(NOTE: currently runTheMatrix tests are failing with an unrelated error but these were checked on other reconstruction workflows.)

Details also given for: 
```
1341.0 (RSKKGluon)
25202.0 (TTbar fullsim + reco + validation)
```

For the TTbar fullsim worfklow:

*Adding to AOD*:

```
recoPFJets_ak4PFJetsPuppi__RECO. 3690 1061
recoPFJets_ak8PFJetsPuppi__RECO. 1260.6 487.2
recoPFJets_ak8PFJetsPuppiSoftDrop_SubJets_RECO. 265 255
```

*Removing from AOD*:

```
recoPFJets_ak8PFJetsCHS__RECO. 6833.8 1735.4
recoPFJets_ak8PFJetsCHSSoftDrop_SubJets_RECO. 1315 506
recoPFJets_cmsTopTagPFJetsCHS_caTopSubJets_RECO. 1666.2 593.4
recoBasicJets_cmsTopTagPFJetsCHS__RECO. 417 288.3
recoBasicJets_ak8PFJetsCHSSoftDrop__RECO. 407.2 284.7
```

To compare to other collections (still remaining):

```
recoPFJets_ak4PFJets__RECO. 42550.7 9657.1
recoPFJets_ak4PFJetsCHS__RECO. 31311 7297.7
recoPFJets_pfJetsEI__RECO. 31267.6 7286
```

So overall we're going to save some space with PUPPI AK8 in AOD.

For MINIAOD we get minor changes due to removing CHS userfloats, etc. 

Old version (with AK8CHS):
```
patJets_slimmedJets__RECO. 37241.8 12613.3
patJets_slimmedJetsPuppi__RECO. 22317.5 10986.9
patJets_slimmedJetsAK8__RECO. 11287.5 1899.9
patJets_slimmedJetsAK8PFPuppiSoftDropPacked_SubJets_RECO. 277.8 271.9
```

new version (no AK8CHS):
```
patJets_slimmedJets__RECO. 36523.9 12503.7
patJets_slimmedJetsPuppi__RECO. 22317.5 10986.5
patJets_slimmedJetsAK8__RECO. 10308.9 1730.7
patJets_slimmedJetsAK8PFPuppiSoftDropPacked_SubJets_RECO. 277.8 271.9
```

Changes checked on the boosted top quark (RSKKgluon sample), and the changes are as expected there as well. 


rappoccio:
Please test.

slava77:
@cmsbuild please test

ahinzmann:
@rappoccio: I don't see the PFClusterJets in the diff?

slava77:
@cmsbuild please test

slava77:
> (NOTE: currently runTheMatrix tests are failing with an unrelated error but these were checked on other reconstruction workflows.)

@rappoccio 
what is the meaning of "unrelated" in your comment. The workflows are clearly broken by this PR.

e.g. in 10024.0
```  
 [1] Calling method for module B2GHadronicHLTValidation/'b2gDiJetHLTValidation'
Exception Message:
Looking for a container with elements of type: reco::Jet
Looking for module label: ak8PFJetsCHS
```

Please test the [currently failing] workflows in your local area before updating the PR.

PhysicsTools/PatAlgos/python/slimming/applySubstructure_cff.py
```diff
@@ -131,7 +76,10 @@ def applySubstructure( process, postfix="" ) :
         e.toModify(getattr(process,'nb1AK8PuppiSoftDrop'+postfix), cuts = ['pt > 999999', 'pt > 999999', 'pt > 999999'] )
         e.toModify(getattr(process,'nb2AK8PuppiSoftDrop'+postfix), cuts = ['pt > 999999', 'pt > 999999', 'pt > 999999'] )
 
-
+    from RecoJets.JetProducers.nJettinessAdder_cfi import Njettiness
+    addToProcessAndTask('NjettinessAK8Subjets'+postfix, Njettiness.clone(), process, task)
+    getattr(process,"NjettinessAK8Subjets"+postfix).src = cms.InputTag("ak8PFJetsPuppiSoftDrop"+postfix, "SubJets")
+    getattr(process,"NjettinessAK8Subjets").cone = cms.double(0.8)
```
slava77: this should be available with just ```process.NjettinessAK8Subjets.cone = ``` .. that's unless a ```+postfix``` is missing here

please drop type specifications for all parameters which already exist.
Here it will be ```process.NjettinessAK8Subjets.cone = 0.8``` and ```getattr(process,"NjettinessAK8Subjets"+postfix).src = "ak8PFJetsPuppiSoftDrop"+postfix+":SubJets"```

RecoJets/Configuration/python/RecoPFJets_cff.py
```diff
@@ -113,10 +94,12 @@
                            fixedGridRhoFastjetCentral,
                            fixedGridRhoFastjetCentralChargedPileUp,
                            fixedGridRhoFastjetCentralNeutral,
+                           puppi,
```
slava77: please provide some timing measurements for the HI case; about 100 events of wf 140.56 should suffice.
was puppi validated for the HI data?

IIRC, a different "pileup" (underlying event) subtraction scheme is used for HI physics, at least for somewhat central events.

slava77:
@rappoccio 
according to the tests, the PR is still broken.
Please clarify on the status of fixing it.
Thank you.

rappoccio:
@slava77 sorry, yes, I've been totally swamped this week, I will get to it ASAP (but probably Monday).

rappoccio:
Apologies for the mess (*) and for the delay, but the tests are now working except a few failures due to external files not being found. 


(*) My local tests were working because I apparently forgot the "-a" in "checkdeps -a" ;)

slava77:
@cmsbuild please test

let's try one more time

rappoccio:
The patAlgos runtests are failing because the file it is looking for (of course) doesn't have our changes in place yet. 

```
09-Oct-2019 14:39:04 CEST  Initiating request to open file root://eoscms.cern.ch//eos/cms/store/user/cmsbuild/store/relval/CMSSW_9_2_2/RelValProdTTbar_13/AODSIM/91X_mcRun2_asymptotic_v3-v1/10000/EEB99F74-DA4D-E711-A41C-0025905A48F2.root
09-Oct-2019 14:39:08 CEST  Successfully opened file root://eoscms.cern.ch//eos/cms/store/user/cmsbuild/store/relval/CMSSW_9_2_2/RelValProdTTbar_13/AODSIM/91X_mcRun2_asymptotic_v3-v1/10000/EEB99F74-DA4D-E711-A41C-0025905A48F2.root
Begin processing the 1st record. Run 1, Event 2, LumiSection 1 on stream 0 at 09-Oct-2019 14:39:29.178 CEST
%MSG-w DataSource:  PATTauProducer:patTaus 09-Oct-2019 14:39:57 CEST  Run: 1 Event: 2
The following tau discriminators have not been found in the event:
byVVLooseIsolationMVArun2v1DBdR03oldDMwLT, byVVLooseIsolationMVArun2v1DBnewDMwLT, byVVLooseIsolationMVArun2v1DBoldDMwLT
They will not be embedded into the pat::Tau object.
Note: this message will be printed only at first occurence.
%MSG
----- Begin Fatal Exception 09-Oct-2019 14:39:57 CEST-----------------------
   [0] Processing  Event run: 1 lumi: 1 event: 2 stream: 0
   [1] Running path 'outpath'
   [2] Prefetching for module PoolOutputModule/'out'
   [3] Prefetching for module PATMETProducer/'patMETsAK8PFCHS'
   [4] Prefetching for module CorrectedPFMETProducer/'AK8PFchsForAK8PFCHSType1p2CorMet'
   [5] Calling method for module PFJetMETcorrInputProducer/'AK8PFchsForAK8PFCHSJetMETcorr'
Exception Message:
Looking for type: std::vector<reco::PFJet>
Looking for module label: ak8PFJetsPuppi

   Additional Info:

```

jfernan2:
@rappoccio  is it expected that this PR touched HLT DQM plots related to AK8 jets?
Or is it just a consequence of moving from CHS to Puppi Jets which are read by default as if they were CHS jets?
I see this in MC from Run2 wfs (10k,11k,12k series) but not in data (except MinBias2011 wf 1000):
https://cmssdt.cern.ch/SDT/jenkins-artifacts/baseLineComparisons/CMSSW_11_0_X_2019-10-09-1100+48c390/33731/dqm-histo-comparison-summary.html

rappoccio:
@jfernan2 yes, this will touch all the DQM plots that looked at ak8PFJetsCHS, which is now (proposed to be) dropped. The new ones will be ak8PFJetsPuppi, so all of the downstream code needs to be adjusted. The DQM histograms for CHS will not agree with those for PUPPI.

jfernan2:
+1

PhysicsTools/PatAlgos/test/patTuple_addBTagging_cfg.py
```diff
@@ -126,31 +126,31 @@
 )
 process.patJetsAK4PF.addTagInfos = True
 
-# uncomment the following lines to add ak8PFJetsCHS with new b-tags to your PAT output
+# uncomment the following lines to add ak8PFJetsPuppi with new b-tags to your PAT output
```
slava77: it looks like the comment is obsolete now
or perhaps it should say the reverse "comment out .. to remove "?

PhysicsTools/PatAlgos/test/patTuple_addBTagging_cfg.py
```diff
@@ -126,31 +126,31 @@
...
    algo = 'AK',
    rParam = 0.8,
    btagDiscriminators = ['pfBoostedDoubleSecondaryVertexAK8BJetTags']
 )
 process.patJetsAK8PFCHS.addTagInfos = True
 
-# uncomment the following lines to add subjets of ak8PFJetsCHSSoftDrop with new b-tags to your PAT output
+# uncomment the following lines to add subjets of ak8PFJetsPuppiSoftDrop with new b-tags to your PAT output
 from pdb import set_trace
 addJetCollection(
```
slava77: it looks like the comment is obsolete now

RecoJets/Configuration/python/RecoPFJets_cff.py
ahinzmann: @FHead confirmed that we don't use PUPPI for HI RECO, so we can remove this here.

slava77 (commented):
I see that there are quite a few places where JME configs create AK8 modules with "PFCHS" in the names, which are reconfigured to use Puppi as inputs.
This better be cleaned up.

I guess we are not quite ready yet to commit to a generic name (e.g. PUS for "pileup subtraction") to decouple the utility name of the collections from the algorithm used to make the inputs.
Or are we?

slava77:
> I see that there are quite a few places where JME configs create AK8 modules with "PFCHS" in the names, which are reconfigured to use Puppi as inputs.

at a glance, though, the obvious places are only in the /test files. Perhaps this has a limited scope then.

slava77:
> The patAlgos runtests are failing because the file it is looking for (of course) doesn't have our changes in place yet.


does this issue affect any other workflows?
If it's restricted to just a few cfg in /test, I think that it may be easier to insert temporary edits to the tests to still use the CHS jets. These can be updated later in a follow up PR.

We should not leave unit tests broken.

slava77:
> In the past, PUPPI has not been stable enough to go into AOD. It was always rerun at MINIAOD level. However BTV would like PUPPI in AOD so they can run costly algorithms at RECO time.

was the PAT step updated to remove PUPPI?
This PR will be complete only after a standard miniAOD running on new AOD which was made to run PUPPI will not rerun puppi and everything downstream from it.

PhysicsTools/PatAlgos/python/slimming/miniAOD_tools.py
```diff
@@ -371,7 +371,8 @@ def miniAOD_customizeCommon(process):
         process.makePatTausTask, _makePatTausTaskWithTauReReco
         )
     
-    # Adding puppi jets
+    # puppi jets : Updated 2019 by S. Rappoccio. PUPPI now being placed into RECO.
+    # If this is a RE-RECO or otherwise, this will pick up the later PUPPI, including tunes.
```
slava77: was this meant to say "re-miniAOD"?

slava77:
@rappoccio 
please clarify on the status of this PR.
Thank you.

rappoccio:
@slava77 sorry for the delay, I'm trying to get to it this week.

rappoccio:
Rebasing. Do not test yet.

PhysicsTools/PatAlgos/test/patTuple_addBTagging_cfg.py
rappoccio: As you recommend below I will just leave the puppi out of the tests for the time being until things are integrated.

PhysicsTools/PatAlgos/test/patTuple_addBTagging_cfg.py
rappoccio: As you recommend below I will just leave the puppi out of the tests for the time being until things are integrated.

PhysicsTools/PatAlgos/test/patTuple_addBTagging_cfg.py
```diff
@@ -126,31 +126,31 @@
 )
 process.patJetsAK4PF.addTagInfos = True
 
-# uncomment the following lines to add ak8PFJetsCHS with new b-tags to your PAT output
+# uncomment the following lines to add ak8PFJetsPuppi with new b-tags to your PAT output
 addJetCollection(
    process,
    labelName = 'AK8PFCHS',
```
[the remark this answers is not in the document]
rappoccio: As you recommend below I will just leave the puppi out of the tests for the time being until things are integrated.

PhysicsTools/PatAlgos/python/slimming/miniAOD_tools.py
rappoccio: Yes sorry.

rappoccio (commented):
As you recommend I will just leave the puppi out of the tests for the time being until things are integrated.

rappoccio:
I have backed out the changes to the PAT tests so they will continue to function for the time being. I have also fixed the comments.

slava77:
@cmsbuild please test

DQMOffline/Trigger/python/JetMonitor_cff.py
```diff
@@ -234,7 +234,7 @@
 # HLT_AK8PFJet80
```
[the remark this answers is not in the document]
jfernan2: I understand this improvement will go to a different PR, please let me know in case not.

jfernan2:
+1
The bulk of changes come from the switch AK8CHS to AK8Puppi in HLT/BTV and HLT/JME which are monitoring paths related to AK8 Jets.

rappoccio:
@slava77 Sorry, missed this. MINIAOD will currently not re-run PUPPI. I'm not sure from your comment if this is what you want.

slava77:
> @slava77 Sorry, missed this. MINIAOD will currently not re-run PUPPI. I'm not sure from your comment if this is what you want.

ah, good to know. 
How does it work, btw? Packed candidates consume puppi to save the puppi weights.

rappoccio:
Puppi is run upstream during RECO in this PR, [here](https://github.com/cms-sw/cmssw/blob/52054d8ef87f435f546ff515fe1639283708a996/RecoJets/Configuration/python/RecoPFJets_cff.py#L35).

slava77:
> Puppi is run upstream during RECO in this PR, [here](https://github.com/cms-sw/cmssw/blob/52054d8ef87f435f546ff515fe1639283708a996/RecoJets/Configuration/python/RecoPFJets_cff.py#L35).

but RECO is not running in a miniAOD process.
Did I miss that the puppi collections are now saved in AOD?

rappoccio:
Right, that's the idea. Instead of running at MINIAOD level, we are running puppi at RECO level with this PR.

slava77:
so, the puppi collection is saved now in AOD?
otherwise I'm still missing how packed candidates can read from puppi  collection when reading AOD file (standard miniAOD processing).

rappoccio:
Ahh, I see the question. I think you're right. We should keep the puppi collection in RECO/AOD or I think unscheduled mode may re-run puppi for MINIAOD.

peruzzim:
Were these important items discussed in a RECO/AT meeting? May I suggest we have a comprehensive discussion there, at the first occasion?

slava77:
> Were these important items discussed in a RECO/AT meeting? May I suggest we have a comprehensive discussion there, at the first occasion?

sort of.
This started in connection with the proposal on Sep 20 https://indico.cern.ch/event/849068/contributions/3567832/attachments/1911759/3159499/JetMet_RECO_update.pdf
I was really puzzled back then how we managed to run puppi in reco consistently with miniAOD (and not rerunning it in miniAOD).

The other question I had then was about the daughters of the puppi jets. What do we store for these jets; are the daughters accessible?

PhysicsTools/PatAlgos/python/slimming/applySubstructure_cff.py
slava77: https://github.com/cms-sw/cmssw/pull/28077/files#r331037650 needs to be addressed

RecoBTag/TensorFlow/test/test_deep_doubleb_cfg_AOD.py
```diff
@@ -62,7 +62,7 @@
    process,
    #labelName = 'AK8PFCHS',
    labelName = 'AK8PFPuppi',
-   #jetSource = cms.InputTag('ak8PFJetsCHS'),
+   #jetSource = cms.InputTag('ak8PFJetsPuppi'),
```
slava77: this was marked as resolved at the point when modifications in other test files (PhysicsTools/PatAlgos/test/patTuple_addBTagging_cfg.py PhysicsTools/PatAlgos/test/patTuple_addJets_cfg.py) were removed/reverted , however the file is still modified, even though in a more harmless way.
Was it intentional to keep?

RecoJets/Configuration/python/RecoPFJets_cff.py
slava77: an update here (drop puppi from recoPFJetsHITask) is still pending

slava77 (commented):
while the issue of having puppi both in RECO and PAT is being understood, here are a few comments that remained not addressed from the last review iteration

RecoBTag/TensorFlow/test/test_deep_doubleb_cfg_AOD.py
rappoccio: Yes, having this as the "test" case is desired.

rappoccio:
Please hold off on testing, I've had to commit to address comments but need to change releases to test.

rappoccio:
@slava77 the configuration is complaining if I don't add the type in the "applySubstructure" file. Going to have to revert that.

PhysicsTools/PatAlgos/python/slimming/applySubstructure_cff.py
slava77: please confirm that here on L82 there should be no postfix.
if the plain ```NjettinessAK8Subjets``` should be here, I'm curious where it is created.

slava77:
> @slava77 the configuration is complaining if I don't add the type in the "applySubstructure" file. Going to have to revert that.

Which parameter was it?
was it about ```.cone = cms.double(0.8)``` ? I can't find that parameter in the NjettinessAdder. Am I missing something or is this a fake parameter?

rappoccio:
Hi, Slava,

Ahh, you're right, the postfix is indeed the issue there. I will fix and start the tests.

PhysicsTools/PatAlgos/python/slimming/applySubstructure_cff.py
```diff
@@ -131,7 +76,10 @@ def applySubstructure( process, postfix="" ) :
         e.toModify(getattr(process,'nb1AK8PuppiSoftDrop'+postfix), cuts = ['pt > 999999', 'pt > 999999', 'pt > 999999'] )
         e.toModify(getattr(process,'nb2AK8PuppiSoftDrop'+postfix), cuts = ['pt > 999999', 'pt > 999999', 'pt > 999999'] )
 
-
+    from RecoJets.JetProducers.nJettinessAdder_cfi import Njettiness
+    addToProcessAndTask('NjettinessAK8Subjets'+postfix, Njettiness.clone(), process, task)
+    getattr(process,"NjettinessAK8Subjets"+postfix).src = cms.InputTag("ak8PFJetsPuppiSoftDrop"+postfix, "SubJets")
+    getattr(process,"NjettinessAK8Subjets"+postfix).cone = cms.double(0.8)
```
slava77: I'm guessing from the commit comments that simple ```cone = 0.8```
did not work.
May I rephrase my earlier question: where is this parameter actually used?

slava77: after some digging, I see that there was a "cone" parameter in this module config until  5 years ago. 
The suggestion to drop the type specs correctly pointed to a buggy code.

rappoccio: Correct. This should be "R0". I'm working on it now.

rappoccio:
OK, this has been fixed. My tests were successful.

PhysicsTools/PatAlgos/python/slimming/applySubstructure_cff.py
```diff
@@ -131,7 +76,10 @@ def applySubstructure( process, postfix="" ) :
         e.toModify(getattr(process,'nb1AK8PuppiSoftDrop'+postfix), cuts = ['pt > 999999', 'pt > 999999', 'pt > 999999'] )
         e.toModify(getattr(process,'nb2AK8PuppiSoftDrop'+postfix), cuts = ['pt > 999999', 'pt > 999999', 'pt > 999999'] )
 
-
+    from RecoJets.JetProducers.nJettinessAdder_cfi import Njettiness
+    addToProcessAndTask('NjettinessAK8Subjets'+postfix, Njettiness.clone(), process, task)
+    getattr(process,"NjettinessAK8Subjets"+postfix).src = cms.InputTag("ak8PFJetsPuppiSoftDrop"+postfix, "SubJets")
+    getattr(process,"NjettinessAK8Subjets"+postfix).R0 = 0.8
```
slava77: this parameter is already set to 0.8
https://github.com/cms-sw/cmssw/blob/783cecb12c181a0b36f30a144fe1d48bd0dd0008/RecoJets/JetProducers/python/nJettinessAdder_cfi.py#L9
what is the reason for the explicit assignment here?

rappoccio: Just for backwards "compatibility". I guess it can go if you want.

ahinzmann:
Hi,
since we don't get to discuss this during a meeting this week, let me bring it up here.
Shouldn't we also add the PUPPI weights also to AOD, such that they don't need to be recomputed for MiniAOD, i.e.
keep "recoCandidateedmPtredmValueMap_puppi_*_*"
@rappoccio How much would this increase size? Would this avoid triggering a rerun of puppi for MiniAOD?

slava77:
> In principle we can add it, but we should do it in a separate PR, I think.

if PUPPI is to be moved up to reco, it should be done consistently.

I have one pending question for which I still did not get an answer.
Jets are persisted including their daughters (about 40% of the jet collection size on file).
Normally it should be possible to get all the details about the jet from these daughters. 
E.g. the ak4PFJetsCHS are pointing to the standard PF candidate collection.
In this PR the PUPPI jets daughters are not saved in the event but the "dau" data member is still filling up space.
Is this OK for the users? 
Can we wipe out the pointers if the pointed to data is not stored?

rappoccio:
Regarding the daughter access: Admittedly it's not an ideal solution. In principle we can add them, especially since we're about to drop the Candidates attached to the Track Jets. 

Would that be okay?

slava77:
> Regarding the daughter access: Admittedly it's not an ideal solution. In principle we can add them, especially since we're about to drop the Candidates attached to the Track Jets.
> 
> Would that be okay?

minimally, "adding them" sounds like an addition of a full clone of PF candidates.
This will not work.
I'm guessing from the context that this can instead be a collection of RecoChargedRefCandidate or similar.

Using the approach from miniAOD seems more practical.
IIUC, there the daughters point to the standard packed cands.
A similar approach here would add the puppi weight to the PF candidates so that the puppi jet daughters can point to these PF candidates.
Although a less invasive approach would be to set CandidatePtr daughters for the puppi jets to be the parent PF candidates and to also save the puppi weights.

BTW, in the version of this PR as of 52054d8, it looks like everything related to PUPPI was remade in the miniAOD step (including even ak4PFJetsPuppi).
This gives little motivation to save anything new PUPPI related in the AOD step until the workflows can be decoupled.

slava77:
@cmsbuild please test

rappoccio:
Closing this because in the review the scope needs to be broadened to add PUPPI to PFCandidates directly. Will make a new PR.

PR #28080, opened 2019-09-27
Python2to3: cmp to key conversion in list sort substitute

mrodozov:
#### PR description:
Fixes python3 implementation of list sorting using a comparison function
as described here: 
https://stackoverflow.com/questions/2531952/how-to-use-a-custom-comparison-function-in-python-3
#### PR validation:
Works both with py2 and py3

### DIFF
--- a/FWCore/Concurrency/scripts/edmStreamStallGrapher.py
+++ b/FWCore/Concurrency/scripts/edmStreamStallGrapher.py
@@ -422,8 +422,6 @@ def printStalledModulesInOrder(stalledModules):
         t.sort(reverse=True)
         priorities.append((name,sum(t),t))
 
-    def sumSort(i,j):
-        return cmp(i[1],j[1])
     priorities.sort(key=lambda a: a[1], reverse=True)
 
     nameColumn = "Stalled Module"
--- a/FWCore/Concurrency/scripts/edmStreamStallGrapher.py
+++ b/FWCore/Concurrency/scripts/edmStreamStallGrapher.py
@@ -6,7 +6,6 @@ from operator import attrgetter,itemgetter
 import sys
 from collections import defaultdict
 import six
-from functools import cmp_to_key
 #----------------------------------------------
 def printHelp():
     s = '''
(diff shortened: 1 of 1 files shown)


mrodozov:
please test

Dr15Jones:
How was this tested ?

mrodozov:
This test 
TestFWCoreConcurrencyDriver 
was failing in 
https://github.com/cms-sw/cmsdist/pull/5241 
and using the PY3 IB I tried this suggestion, tried also with py2 IB and also worked.

FWCore/Concurrency/scripts/edmStreamStallGrapher.py
```diff
@@ -424,7 +424,7 @@ def printStalledModulesInOrder(stalledModules):
 
     def sumSort(i,j):
         return cmp(i[1],j[1])
-    priorities.sort(cmp=sumSort, reverse=True)
+    priorities.sort(key=cmp_to_key(sumSort), reverse=True)
```
smuzaffar: following should work for both py2 and 3
```
priorities = sorted(priorities , key=lambda a: a[1], reverse=True)
```

Dr15Jones:
```suggestion
    priorities.sort(key=lambda a: a[1], reverse=True)
```
And then delete the function `sumSort`. This allows changing the (possibly large) list inplace which is probably more efficient.

mrodozov:
please test

FWCore/Concurrency/scripts/edmStreamStallGrapher.py
```diff
@@ -6,7 +6,7 @@
 import sys
 from collections import defaultdict
 import six
-
+from functools import cmp_to_key
```
Dr15Jones: The import does not appear to be needed

mrodozov:
please test

Dr15Jones:
+1

fabiocos:
+1

PR #28083, opened 2019-09-27
Py2to3: use builtin int for py2 - py3 compatibility

mrodozov:
#### PR description:
Change long with int:
https://docs.python.org/3/whatsnew/3.0.html#integers
first answer on SO about that 
Not sure if int is acceptable the way we are running now. 
If not I can put version check condition

#### PR validation:
Runs with both py2 and 3

### DIFF
--- a/DataFormats/FWLite/python/__init__.py
+++ b/DataFormats/FWLite/python/__init__.py
@@ -512,7 +512,10 @@ class Events:
         """Jumps to event entryIndex"""
         if self._veryFirstTime:
             self._createFWLiteEvent()
-        return self._event.to ( int(entryIndex) )
+        if sys.version_info[0] > 2:
+            return self._event.to ( int(entryIndex) )
+        else:
+            return self._event.to ( long(entryIndex) )
 
         
     def toBegin (self):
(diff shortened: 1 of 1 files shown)


mrodozov:
please test

Dr15Jones:
Given the entry number can be very large, I would guess this will not work always with python 2

mrodozov:
@Dr15Jones  
it's updated now

Dr15Jones:
@davidlange6 @ericvaandering suggestions?

Dr15Jones:
please test

ericvaandering:
That looks correct to me. Do you have a "standard comment" which could be used to flag all of this noise so that it can be simplified when python2 is a bad memory?

Alternatively if you are using py-future, this is cleaner: https://python-future.org/compatible_idioms.html#long-integers

Then when you are truly converted you remove `from builtins import int  `

ericvaandering:
A search for builtins shows other code already using it. I'd suggest that in this case.

davidlange6:
yes, I also support using builtins here. will be easier to clean up in N years when we can drop python2.

Dr15Jones:
-1
Please change to using builtins package.

mrodozov:
@Dr15Jones changed.

Dr15Jones:
please test

Dr15Jones:
+1
please change the pull requests title since it no longer reflects what was done.

fabiocos:
+1

@Dr15Jones I updated the title to reflect what has been the finally adopted solution

PR #28088, opened 2019-09-29
Run3-Sim47 Allow saving tracks with momentum above a certain threshold

bsunanda:
#### PR description:

These changes allow saving some of the Sim tracks entering in the HGCal region and have kinetic energies above a threshold. By default the thresholds are set high - one need to configure it to allow tracks at lower energies

#### PR validation:

Tested with cfg files in SimG4CMS/Calo/test/python area and also with the standard run matrix

Nothing special

### DIFF
--- a/SimG4CMS/Calo/interface/CaloTrkProcessing.h
+++ b/SimG4CMS/Calo/interface/CaloTrkProcessing.h
@@ -64,7 +64,6 @@ private:
   double eMin_, eMinFine_, eMinFinePhoton_;
   int lastTrackID_;
   std::vector<Detector> detectors_, fineDetectors_;
-  const SimTrackManager* m_trackManager;
 };
 
 #endif
--- a/SimG4CMS/Calo/interface/CaloTrkProcessing.h
+++ b/SimG4CMS/Calo/interface/CaloTrkProcessing.h
@@ -38,13 +38,6 @@ public:
   void fillHits(edm::PCaloHitContainer&, const std::string&) override {}
 
 private:
-  void update(const BeginOfEvent* evt) override;
-  void update(const G4Step*) override;
-  std::vector<std::string> getNames(G4String, const DDsvalues_type&);
-  std::vector<double> getNumbers(G4String, const DDsvalues_type&);
-  int isItCalo(const G4VTouchable*);
-  int isItInside(const G4VTouchable*, int, int);
-
   struct Detector {
     Detector() {}
     std::string name;
@@ -55,15 +48,22 @@ private:
     std::vector<int> fromLevels;
   };
 
+  void update(const BeginOfEvent* evt) override;
+  void update(const G4Step*) override;
+  std::vector<std::string> getNames(G4String, const DDsvalues_type&);
+  std::vector<double> getNumbers(G4String, const DDsvalues_type&);
+  int isItCalo(const G4VTouchable*, const std::vector<Detector>&);
+  int isItInside(const G4VTouchable*, int, int);
+
   // Utilities to get detector levels during a step
   int detLevels(const G4VTouchable*) const;
   G4LogicalVolume* detLV(const G4VTouchable*, int) const;
   void detectorLevel(const G4VTouchable*, int&, int*, G4String*) const;
 
-  bool testBeam, putHistory;
-  double eMin;
-  int lastTrackID;
-  std::vector<Detector> detectors;
+  bool testBeam_, putHistory_, doFineCalo_;
+  double eMin_, eMinFine_, eMinFinePhoton_;
+  int lastTrackID_;
+  std::vector<Detector> detectors_, fineDetectors_;
   const SimTrackManager* m_trackManager;
 };
 
--- a/SimG4CMS/Calo/interface/CaloTrkProcessing.h
+++ b/SimG4CMS/Calo/interface/CaloTrkProcessing.h
@@ -50,8 +50,8 @@ private:
 
   void update(const BeginOfEvent* evt) override;
   void update(const G4Step*) override;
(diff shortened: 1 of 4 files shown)


bsunanda:
@cmsbuild Please test

bsunanda:
@cmsbuild Please test

SimG4CMS/Calo/src/CaloTrkProcessing.cc
```diff
@@ -52,45 +55,61 @@ CaloTrkProcessing::CaloTrkProcessing(const std::string& name,
   DDsvalues_type sv(fv.mergedSpecifics());
 
   G4String value = "Calorimeter";
-  std::vector<std::string> caloNames = getNames(value, sv);
+  std::vector<std::string> caloNames = getNames(value, sv, false);
```
kpedro88: better to capture these return values as `const std::vector<std::string>&` (similarly for all other calls to `getNames()` or `getNumbers()` in this file)

SimG4Core/Application/python/g4SimHits_cfi.py
```diff
@@ -559,3 +562,5 @@
 phase2_timing.toModify( g4SimHits.ECalSD,
                              StoreLayerTimeSim = cms.untracked.bool(True),
                              TimeSliceUnit = cms.double(0.001) )
+from Configuration.Eras.Modifier_phase2_common_cff import phase2_common
+phase2_common.toModify( g4SimHits.CaloTrkProcessing, DoFineCalo = True)
```
kpedro88: Do we want this enabled by default for Phase2? Is there a change in processing time or event content size?

kpedro88:
@lgray fyi

bsunanda:
@cmsbuild Please test

SimG4Core/Application/python/g4SimHits_cfi.py
bsunanda: Initially I thought of having very high threshold and have it as default for phase2. But thinking again, I see your point - it may be better to have reasonable threshold and remove this from default so that only interested people will use this

kpedro88:
+upgrade

SimG4CMS/Calo/src/CaloTrkProcessing.cc
```diff
@@ -30,17 +30,20 @@ CaloTrkProcessing::CaloTrkProcessing(const std::string& name,
...
-                              << " History flag " << putHistory;
-
+  testBeam_ = m_p.getParameter<bool>("TestBeam");
+  eMin_ = m_p.getParameter<double>("EminTrack") * MeV;
+  putHistory_ = m_p.getParameter<bool>("PutHistory");
+  doFineCalo_ = m_p.getParameter<bool>("DoFineCalo");
+  eMinFine_ = m_p.getParameter<double>("EminFineTrack") * MeV;
+  eMinFinePhoton_ = m_p.getParameter<double>("EminFinePhoton") * MeV;
+#ifdef EDM_ML_DEBUG
+  edm::LogVerbatim("CaloSim") << "CaloTrkProcessing: Initailised with TestBeam = " << testBeam_ << " Emin = " << eMin_
```
civanch: @bsunanda , this printout better not protect by ifdef

SimG4Core/Application/python/g4SimHits_cfi.py
civanch: @bsunanda , @kpedro88 , I agree that this is for detector study and not for production.

civanch:
@bsunanda , some comments:
1) better not protect printouts by ifdef in constructor of a class, from my point of view ifdef is useful in run time
2) the method you choose is correct, however, from my point of view, it can be much simpler. At constructor you access a pointer to G4Region named "HGCalRegion", in run time check that pre-stepPoint is inside this region - 1 line of code, you know for 100% that you are inside sensitive part of HGCal and may apply the filter. So, struct "detector" is not needed.

bsunanda:
@cmsbuild Please test

SimG4CMS/Calo/interface/CaloTrkProcessing.h
```diff
@@ -55,15 +48,22 @@ class CaloTrkProcessing : public SensitiveCaloDetector,
...
 
-  bool testBeam, putHistory;
-  double eMin;
-  int lastTrackID;
-  std::vector<Detector> detectors;
+  bool testBeam_, putHistory_, doFineCalo_;
+  double eMin_, eMinFine_, eMinFinePhoton_;
+  int lastTrackID_;
+  std::vector<Detector> detectors_, fineDetectors_;
   const SimTrackManager* m_trackManager;
```
ianna: ```trackManager_``` for consistency?

ianna (changes requested):

bsunanda:
@cmsbuild Please test

civanch:
@bsunanda , sorry, I do not raise this question before. Is it optimal introduce MCtruth into sensitive detector code? I mean that until now MCtruth was in TrackingAction/StackingAction - it is more natural place, because there is sorting particles at production - we create a new MCtruth object depending on energy of a particle at production. Inside the SD code it is more easy identify the moment when a particle hit HGCal but to handle particles produced inside HGCal user actions are better.

ianna:
+1

civanch:
+1
let us left this variant and later may be make MC truth in user actions.

fabiocos:
@kpedro88 I think your remarks have been addressed, and you had already signed this code. Latest changes look straightforward to me, I merge the PR, please check it and comment in case.

fabiocos:
+1

fabiocos:
merge

kpedro88:
+upgrade

PR #28091, opened 2019-09-30
HLTTauDQM: moved some warnings behind a verbose flag

slehti:
#### PR description:

HLTTauDQM was filling logs with warnings. Those warnings are now moved behind a verbose parameter, false by default. 

#### PR validation:

Tested that the Verbose flag disables/enables the warning printouts with false/true.

### DIFF
--- a/DQMOffline/Trigger/src/HLTTauDQMPath.cc
+++ b/DQMOffline/Trigger/src/HLTTauDQMPath.cc
@@ -203,9 +203,10 @@ namespace {
     if (pset.existsAs<int>(parameterName))
       return pset.getParameter<int>(parameterName);
     else {
-      if(verbose)
-      edm::LogWarning("HLTTauDQMOfflineSource") << "No parameter '" << parameterName << "' in configuration of filter "
-                                                << filterName << " pset " << pset.dump() << std::endl;
+      if (verbose)
+        edm::LogWarning("HLTTauDQMOfflineSource")
+            << "No parameter '" << parameterName << "' in configuration of filter " << filterName << " pset "
+            << pset.dump() << std::endl;
       return 0;
     }
   }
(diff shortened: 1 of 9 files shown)


jfernan2:
please test

jfernan2:
+1

DQMOffline/Trigger/src/HLTTauDQMPath.cc
```diff
@@ -197,13 +197,16 @@ namespace {
...
-                       const std::string& parameterName) {
+                       const std::string& parameterName,
+                       bool verbose) {
     const edm::ParameterSet& pset = HLTCP.modulePSet(filterName);
     if (pset.existsAs<int>(parameterName))
       return pset.getParameter<int>(parameterName);
     else {
-      edm::LogWarning("HLTTauDQMOfflineSource") << "No parameter '" << parameterName << "' in configuration of filter "
-                                                << filterName << " pset " << pset.dump() << std::endl;
+      if (verbose)
```
fabiocos: @slehti @jfernan2 sorry, isn't this exercise a complex way to downgrade a LogWarning to a LogInfo?

fabiocos: @slehti @jfernan2 ping

slehti: I did not want to change any LogWarnings into LogInfo, only to have a way to silence some warnings. And to have them available when needed. We need to fix a long standing problem in many RelVal workflows coming from HLTTau DQM modules: warnings are flooding the logs (since a long time).

fabiocos:
@slehti @jfernan2 sorry, I would like to understand the strategy behind this PR: are we trying to solve the problem of an incorrect input to the module, or just to hide it under the carpet? ```LogWarning``` is a quite high level message, meaning there is something preventing the code to properly work. Effectively switching off with a ```verbose``` flag is not different than turning it to a ```LogInfo``` that is normally ignored by the usual MessageLogger settings that we use, and that is an equivalent but simpler solution. 

As far as I can see this PR is not helping in running correctly the module. At this point why not just switch it off, if at the end it will anyway produce empty histograms?

DQMOffline/Trigger/src/HLTTauDQMPath.cc
slehti: Timeline: as soon as possible.

jfernan2: Then I believe @fabiocos is right: we should not sweep under the carpet

fabiocos:
hold

fabiocos:
@slehti @jfernan2 what is the plan for this issue? Are we expecting a real solution at some point? I do not see much the point of keeping this PR open in this way.

PR #28092, opened 2019-09-30
DQM: Merge MonitorElement and ConcurrentMonitorElement

schneiml:
#### PR description:
Moving forward, the separation between `ConcurrentMonitorElement` and `MonitorElement` is rather poitnless and annoying, given that *all* DQM will need to use `ConcurrentMonitorElement` semantics in the future.

This PR introduces a Frankenstein-ME that uses parts of the new ME implementation [1] (namely, the locked `MonitorElementData::Value` type) in the current ME code. The result is that the usual interactions with the ME are now locked and thread-safe, similar to the `ConcurrentMonitorElement`. This means we can now drop the `ConcurrentMonitorElement` and instead use `MonitorElement*`. Of course there are still plenty of non-thread-safe interactions in the `MonitorElement`, but these should not be used.

[1] https://github.com/schneiml/cmssw/blob/dqm-new-dqmstore-on-CMSSW_11_0_0_pre5/DQMServices/Core/src/MonitorElement.cc
#### PR validation:
No output changes expected, and none observed.
However, some of the tests where probably incorrect -- the code was almost certainly uncompilable yet passed the tests.


schneiml:
please test

just to see if anything blows up so far.

rovere:
@schneiml 
Ciao Marcel, at some point CMS/DMQ has to take a decision related to DQMGUI. This Pr is fully destructive and the DQMGUI will not even compile against it.
We can still build it against older versions, which is what we do, but a clear road-map and plan have to be put in place.

schneiml:
@rovere Ciao Marco, yes, we'll need to have this discussion in a bit broader scope at some point.

However, I think it is fine to remove the DQMGUI support on DQMStore's side, since we changed/need to change the inner workings of DQM dramatically on CMSSW side (first threaded mode, in the future decentralized DQMStore). IIRC DQMGUI currently uses a CMSSW7 version of the DQMStore, so it has missed a lot of changes already (does the version it uses even have threaded mode?).

Even if we stick with the old DQMGUI, it is probably perfectly fine to maintain it's version of the DQMStore separately. After all, all that we care about is file IO (and network IO, which is arguably a special case of file IO), and the file formats need to be stable for other reasons anyways. Also, no changes to the file formats are planned at the moment (except maybe minor changes to DQMIO).

schneiml:
Ok, now the big changes are also in. I'd have liked to check the previous state first but it seems illegal to add new deprecations; so whatever, let's hope for the best and debug later in case...

@fabiocos @smuzaffar what would be the procedure to deprecate APIs that are currently in use?I sprinkled `DQM_DEPRECATED` across things we don't want to support any more, but of course there are still some usages, so that increases the warning count... For now I just disabled `DQM_DEPRECATED`, but we might want to find a better solution.

Also rebased to master, as there were merge conflicts. Part of the reason to make the PR now is that it became impractical to maintain this change in a private branch, with some `ConcurrentME` users under active development.

schneiml:
please test

Seems like it did not blow up immediately in a local test. Let's see how it goes.

schneiml:
Hmm, these tests seem to take long. I hope there is a timeout in case of deadlocks, these changes have a pretty good chance of deadlocks as a failure mode...

DQM/BeamMonitor/plugins/BeamConditionsMonitor.h
```diff
@@ -11,49 +11,64 @@
...
-    typedef dqm::reco::MonitorElement MonitorElement;
-    // MonitorElements
-    ConcurrentMonitorElement h_x0_lumi;
-    ConcurrentMonitorElement h_y0_lumi;
-  };
-};  // namespace beamcond
-
-class BeamConditionsMonitor : public DQMGlobalEDAnalyzer<beamcond::RunCache, edm::LuminosityBlockCache<void>> {
+
+class BeamConditionsMonitor : public edm::EDAnalyzer {
```
fwyzard: Isn't the use of `edm::EDAnalyzer` deprecated since years ?

DQM/HLTEvF/plugins/LumiMonitor.cc
```diff
@@ -31,12 +30,12 @@ namespace {
   };
 
   struct Histograms {
-    ConcurrentMonitorElement numberOfPixelClustersVsLS;
-    ConcurrentMonitorElement numberOfPixelClustersVsLumi;
-    ConcurrentMonitorElement lumiVsLS;
-    ConcurrentMonitorElement puVsLS;
-    ConcurrentMonitorElement pixelLumiVsLS;
-    ConcurrentMonitorElement pixelLumiVsLumi;
+    dqm::reco::MonitorElement const* numberOfPixelClustersVsLS;
```
fwyzard: what is the `dqm::reco` namespace ?

DQM/BeamMonitor/plugins/BeamConditionsMonitor.h
schneiml: It is, I just reverted this module back to it's previous state becaue I was not really sure what the correct semantics in multi-threaded running would be.

It's a rather pointless exercise to try to fix it, it is only supposed to run online anyways. And actually, as far as I see, it is not running at all anywhere right now. I might just delete it entirely.

DQM/HLTEvF/plugins/LumiMonitor.cc
schneiml: Introduced here, with explanation: https://github.com/cms-sw/cmssw/pull/27476

It might not be strictly required to use the fully qualified name everywhere, but in the global analyzers you will see them more than usual (since the declarations are outside the class scope).

fwyzard:
@schneiml how is the concurrency implemented for the "new ME" ?

schneiml:
@fwyzard It's a very simple `std::mutex` for each histogram, and all operations that touch the histogram need to take the lock first. "Escape hatches" like `getTH1` exist and are unsafe, but I want to make them non-`const` to mark them (not yet done).

Switching to a more fancy locking primitive like you used in the `ConcurrentME` is an option, but first it has to work. My concern there was that the thing currently lives in `DataFormats` (not needed for now, but in the future), so further restrictions might apply.

In the longer term, `DQMEDAnalyzer` will go back to `edm::stream` but share the MEs across all streams, and the global DQMStore will be replaced by per-module ones that put edm products into runs and lumis.

schneiml:
please abort

pretty sure we got stuck somewhere, I'll try to reproduce that locally

schneiml:
please test

The last two WF I saw in the log run fine for me, so give it another try.

schneiml:
I fiddled around a bit and found a way to prevent (const) access to `getTH1` from the `DQMGlobalEDAnalyzer` context, by making it const only in the `dqm::legacy::MonitorElement` (which is now used everywhere except `DQMGlobalEDAnalyzer`. It's not beautiful, but seems to work.

The inheritance between the different MEs is actually the other way round compared to the "new DQMStore" prototype. This should not really matter for anything, but arguably what I used here makes more sense. Maybe we should hide more methods inside the `dqm::legacy` namespace, I only put the most obvious offenders there for now as a proof of concept.

I am afraid me pushing messed up the tests again, which where running for 6h. I'd like to see them finish at some point; judging by the log they seem not stuck, just taking long.

schneiml:
please test

Let's see if there are results once I am back next week.

andrius-k:
please test

fwyzard:
> @fwyzard It's a very simple `std::mutex` for each histogram, and all operations that touch the histogram need to take the lock first. "Escape hatches" like `getTH1` exist and are unsafe, but I want to make them non-`const` to mark them (not yet done).
> 
> Switching to a more fancy locking primitive like you used in the `ConcurrentME` is an option, but first it has to work.

Actually, I would say that it has to work without any negative impact on the HLT performance (at least), before it can be merged.

> My concern there was that the thing currently lives in `DataFormats` (not needed for now, but in the future), so further restrictions might apply.

Speaking specifically about DataFormats, to me it doesn't sound like a good idea to have `mutex`es or to use the `const`/`mutable` idiom to indicate thread-safety.
The framework group may disagree, or they may have a better proposal.

> In the longer term, `DQMEDAnalyzer` will go back to `edm::stream` but share the MEs across all streams, and the global DQMStore will be replaced by per-module ones that put edm products into runs and lumis.

makortel:
> Switching to a more fancy locking primitive like you used in the `ConcurrentME` is an option, but first it has to work.

(Out of curiosity) How exactly is the `ConcurrentME` locking mechanism more fancy than the mutex used in `MonitorElement` (from `MonitorElementData`)?

> Speaking specifically about DataFormats, to me it doesn't sound like a good idea to have `mutex`es or to use the `const`/`mutable` idiom to indicate thread-safety.
> The framework group may disagree, or they may have a better proposal.

The `MonitorElementData` in DataFormats should not have a `std::mutex` member. At the point of inserting a product into Event/Lumi/Run the filling is finished, and the histogram/scalar can be treated as  `const` object, right? So why would the mutex be needed in the DataFormat?

Methods mutating the visible state of an object should not be marked as `const`, even if they are thread safe. I.e. `MonitorElement::Fill()` etc should stay as non-`const`. We have the `CMS_THREAD_SAFE` macro that expands to `[[cms::thread_safe]]` attribute (except on selected compilers) that can be used to mark thread-safe non-`const` methods (and we should do that more often).

fwyzard:
> (Out of curiosity) How exactly is the ConcurrentME locking mechanism more
> fancy than the mutex used in MonitorElement (from MonitorElementData)?
>

It's not fancier, it's a tbb::spinlock.

It is more lightweight than a mutex and faster for very low contention, as
is the case for the few DQM histograms filled in the HLT.

>

schneiml:
@fwyzard That sounds like a very reasonable choice.

I'd like to know from core side (@Dr15Jones or is there some other expert?) if we can use `tbb::spinlock` in data formats, or if there is some cute trick to keep the lock in the correct place without attempting to serialize it. Or if this matters at all for a non-persistent product (which should be enough for the beginning, *unless* that means the `DQMRootOutputModule` would not see it any more because the framework is smart).

schneiml:
@makortel the `CMS_THREAD_SAFE` is something to look at. I thought we started to follow the `const == thread-safe` convention in some places, but I agree that it is rather confusing esp. when mixed with the old API.

About keeping the lock outside the dataformat, that is just rather hard. The product owns the `TH1` (otherwise ownership turns into a nightmare), so it makes sense to also have it protect it.

The obvious choice would be to move the lock into the `MonitorElement`, which wraps around the `MonitorElementData` wherever it is handled by DQM code, *but* multiple `MonitorElement`s can share the same MEData (we cannot make the DQM code share MEs without breaking the API). This could be fixed with another layer of indirection or by having the lock on a pointer next to the MEData, both of which sounds rather complicated and inefficient.

We *could* make a second MEData type that is used for the "mutable" use case and embeds all the fields plus the lock, with conversions in both ways. That also sounds complicated and I know no elegant way to implement it, but it would convey some useful information in the type system.

makortel:
> I thought we started to follow the `const == thread-safe` convention in some places

`const` has never meant just "thread safe", it also includes a promise of not mutating the visible state. (and yes, I have been confused on the exact meaning as well)

> This could be fixed with another layer of indirection or by having the lock on a pointer next to the MEData, both of which sounds rather complicated and inefficient.
> ...
> We could make a second MEData type that is used for the "mutable" use case and embeds all the fields plus the lock, with conversions in both ways. That also sounds complicated and I know no elegant way to implement it

We're essentially talking about `MonitorElementData::Value`, right? (which is essentially a home-made union/variant between `uint_64, double, std::string, std::unique_ptr<TH1>`) I don't really see how either way would be very complicated to do.

It really boils down to whether event products (persistent or transient) should be allowed to mutate their visible state (through `const` methods). Allowing that would open a loophole through which any consumer could change the visible state of such products (ok, the updates would be thread safe, but the results would be non-deterministic).

fwyzard:
>
> multiple MonitorElements can share the same MEData


mhm, does that mean that multiple modules can fill the same ME ?

schneiml:
> mhm, does that mean that multiple modules can fill the same ME ?

Yes. That is the requirement to go back to `edm::stream`, and get rid of the `one`-modules.

schneiml:
> Ah, you mean different stream-instances of the same module (i.e. a single module in the python configuration), not entirely different modules, right ?

Yes exactly. Cross-class filling is also possible in harvesting (you can get some existing MEs and modify them), but in that case we do copy-on-write to keep ownership clear (that is the plan at least).

fwyzard:
> Yes exactly. Cross-class filling is also possible in harvesting (...) but in that case we do copy-on-write to keep ownership clear

OK, so a ME is supposed to be `const` once it has been put into the Event, and it can only be modified by the module that produces it - correct ?

schneiml:
@makortel I don't fully get your `MEData::Value` comment. Yes, We could replace that with a polymorphic type (via template or pointer+inheritance) that can either have the lock (for the mutable type) or not (for the dataformat), but it still adds quite a bit of complexity. Is there an elegant pattern to have `Value` as a template parameter yet have `MEData<LockedValue>` and `MEData<ConstantValue>` have a common base class to use for the pointers? (Perferrably without writing pages and pages of constructors and accessor methods.)

schneiml:
@fwyzard It gets a bit more complictated. Please make sure to not confuse whatever is implemented ATM with what we are discussing here.

What is implemented here (and in our prototype branch) is:
- MEs are non-const after booking.
- MEs have a few const, thread-safe mutators. The default ME type to be used in subsystem code is `ME const*` (but this can't be enforced). `Fill` can be called on const MEs.
- Products are `const MEData`s. `MEData`s don't have any mutators, but do contain a `mutable` TH1. They are supposed to be immutable, as required by EDM. Don't touch unless you know what you are doing.
- In Harvesting, we wrap effectively-const MEDatas into non-const MEs, and anything that might potentially modify the ME will trigger a `clone` in the background. We need to do this lazily to perserve the old API, yet not completely blow up memory consumption.

What we discussed here is:
- Use different MEData for products and MEs-that-are-not-yet-products. The product MEDatas then can be "really" const, with no locks and `mutable`s in them.
- In Harvesting, we still need MEs backed by product MEDatas, and we'd need to transparently clone them into mutable MEDatas.
- In booking, we'd book mutable MEDatas and turn them to products (changing type in process) at endLumi/endRun.

There is some preliminary documentation here (lots of text):
https://github.com/schneiml/cmssw/tree/dqm-new-dqmstore-on-CMSSW_11_0_0_pre5/DQMServices

makortel:
@schneiml For example create a separate type for the Event/Lumi/Run along
```cpp
struct PMonitorElementData {
  using Scalar = MonitorElementData::Scalar;
  using Key = MonitorElementData::Key;
  struct Value {
    Scalar scalar_;
    edm::propagate_const<std::unique_ptr<TH1>> object_;
  };

  Key key_;
  Value value_;
};
```
and move the contents between `MonitorElementData::Value` and `PMonitorElementData::Value` as needed. Or make `MonitorElementData::Value` to hold equivalent of `PMonitorElementData::Value` and the `std::mutex`.

(did I read the diff correctly that the "move MonitorElementData to Event/Lumi/Run" part is not part of this PR?)

schneiml:
> (did I read the diff correctly that the "move MonitorElementData to Event/Lumi/Run" part is not part of this PR?)

Yes, for now my plan is to first do all the required API changes on top of the old `edm::Service` DQMStore, and then have a PR later that swaps the implementation.

I'll think about your suggestion, not sure if I fully get it atm.

fwyzard:
> Please make sure to not confuse whatever is implemented ATM with what we are discussing here.

If I did, sorry for the confusion !

schneiml:
@makortel Coming back to `CMS_THREAD_SAFE`, what exactly does this do? I see it mostly applied to `static` variables, with the convention that statics need to be `CMS_THREAD_SAFE` or const. I can't find any traces of documentation, and the PR introducing it [1] is not too explicit either.

The point here is, `MonitorElement` will always have (logically [2]) non-thread safe methods, which need to be publicly available in booking (and potentially harvesting). By the `const == thread-safe` convention we can get this compile-time checked (booking returns non-const MEs, all long-lived pointers are const), and it would be nice to have some other mechanism that allows the same safety. It could once again be implemented using inheritance, but that again adds some complexity.

[1] https://github.com/cms-sw/cmssw/pull/1711
[2] We can make everything data-race free, except maybe `getTH1` and friends which we could get rid off with some effort, but things such as setting options on the histogram will always be _logically_ racy when called from multiple threads.

makortel:
> Coming back to `CMS_THREAD_SAFE`, what exactly does this do? I see it mostly applied to `static` variables, with the convention that statics need to be `CMS_THREAD_SAFE` or const. I can't find any traces of documentation, and the PR introducing it [1] is not too explicit either.

It translates to `[[cms::thread_safe]]` (custom) attribute (except on selected compilers)
https://github.com/cms-sw/cmssw/blob/28ee89b235caa54439092c0de66c224a576b2923/FWCore/Utilities/interface/thread_safety_macros.h#L3-L4
The `cms::thread_safe` attribute documents that the non-`const` object/method is thread safe, and is used by the static analyzer.

I agree that the macro and the attribute should be properly documented to facilitate their wider use.

> The point here is, `MonitorElement` will always have (logically [2]) non-thread safe methods, which need to be publicly available in booking (and potentially harvesting). By the `const == thread-safe` convention we can get this compile-time checked (booking returns non-const MEs, all long-lived pointers are const), and it would be nice to have some other mechanism that allows the same safety.

We understand. We still think that it would be much better to not break the C++ convention by misusing `const` qualifier for methods that mutate the visible state. For example TBB's concurrent containers mark thread-safe mutating methods as non-`const`, and thread-unsafe methods with `unsafe_` prefix on the method name.

We could think of extending the meaning `cms::thread_safe` attribute to mean "mutating and thread safe" (so in a sense being in between "const == non-mutating and thread safe" and "non-const == all bets are off") such that for `cms::thread_safe`-marked objects only `const` and `cms::thread_safe` methods are allowed to be called. This convention could be enforced with the static analyzer. (e.g. ATLAS is doing something like this)

How would you enforce that the long-lived pointers are stored as pointers-to-`const`?

schneiml:
> We could think of extending the meaning cms::thread_safe attribute to mean "mutating and thread safe" (so in a sense being in between "const == non-mutating and thread safe" and "non-const == all bets are off") such that for cms::thread_safe-marked objects only const and cms::thread_safe methods are allowed to be called. This convention could be enforced with the static analyzer. (e.g. ATLAS is doing something like this)

That sounds good. These static analyzer tests are those that run in PR tests by default, I assume.

> How would you enforce that the long-lived pointers are stored as pointers-to-const?

Mostly by convention. `MonitorElement` is a class-local typedef anyways in most places, so that could include the `const` by default (I hope that works -- not done like that so far). That should handle it for all the existing code, for new code, we just have to watch (or build static analyzer rules...)

makortel:
> > We could think of extending the meaning cms::thread_safe attribute to mean "mutating and thread safe" (so in a sense being in between "const == non-mutating and thread safe" and "non-const == all bets are off") such that for cms::thread_safe-marked objects only const and cms::thread_safe methods are allowed to be called. This convention could be enforced with the static analyzer. (e.g. ATLAS is doing something like this)
> 
> That sounds good. These static analyzer tests are those that run in PR tests by default, I assume.

(don't hold your breath, this topic is really more complex than my paragraph above suggests)

> > How would you enforce that the long-lived pointers are stored as pointers-to-const?
> 
> Mostly by convention. `MonitorElement` is a class-local typedef anyways in most places, so that could include the `const` by default (I hope that works -- not done like that so far). That should handle it for all the existing code, for new code, we just have to watch (or build static analyzer rules...)

@Dr15Jones and I had a long discussion on this topic.

First a few questions: the `bookHistograms()` continues to be called at beginRun time, right? Will it be the per-stream-beginRun (i.e. from `beginRun()` method of a `edm::stream::EDProducer`)? Will the DQMStore that is in Global/LumiCache (sorry, I've forgotten this detail of the plan) of an EDProducer serialize/protect the `bookHistograms()` calls from the stream instances of the EDProducer? What would happen if one of the streams proceeds to call `accumulate()` while some/rest are still waiting for `bookHistograms()` to be called (i.e. the underlying `MonitorElement` object would get calls to thread-safe and to thread-unsafe functions)?

What would you think about the following
* The `MonitorElement` returned by `book1D()` etc. would have fully thread safe public interface (`const` for non-mutating functions, non-`const` for mutating)
  * therefore user code can only call thread safe functions
* The `book1D()` etc. would take additional (optional) lambda that makes all the necessary modifications on the TH1 (see example below) by taking the TH1 as an argument
  * Alternatively or in addition the `MonitorElement` could a method (e.g.) `mutate()` that takes a similar lambda and locks the mutex internally
  * A nice property of having the lambda as part of `book1D()` is that for the second (and subsequent) calls to the `book1D()` the lambda can be ignored as the customizations have already been done

Note that this kind of an approach would have a straightforward migration path: introduce overloads with the lambda argument, move calling code to use those overloads (e.g. package by package), remove thread-unsafe pieces of the public interface of `MonitorElement`.

A random example: code that now looks like this https://github.com/cms-sw/cmssw/blob/217e02f21874875155545dfa82866c43b8aa735a/DQMOffline/EGamma/src/ElectronDqmAnalyzerBase.cc#L93-L108
would look like
```cpp
return iBooker.book1D(newName(name), title, nchX, lowX, highX, [&](TH1F* h) { // or auto h
  if (h->GetSumw2N() == 0)
    h->Sumw2();
  if (!titleX.empty()) {
    h->GetXaxis()->SetTitle(titleX.c_str());
  }
  if (!titleY.empty()) {
    h->GetYaxis()->SetTitle(titleY.c_str());
  }
  if (TString(option) != "") {
    h->SetOption(option);
  }
  if (bookStatOverflowFlag_) {
    h->StatOverflows(kTRUE);
  }
});
```
or with `mutate()`
```cpp
auto me = iBooker.book1D(newName(name), title, nchX, lowX, highX);
me->mutate([&](TH1F* h) { // or auto h
  if (h->GetSumw2N() == 0)
    h->Sumw2();
  if (!titleX.empty()) {
    h->GetXaxis()->SetTitle(titleX.c_str());
  }
  if (!titleY.empty()) {
    h->GetYaxis()->SetTitle(titleY.c_str());
  }
  if (TString(option) != "") {
    h->SetOption(option);
  }
  if (bookStatOverflowFlag_) {
    h->StatOverflows(kTRUE);
  }
});
return me;
```

If you choose an approach like this, the core team is willing to help migrating the client code.

fwyzard:
Hi @makortel ,
in your proposal:,

>     * The `MonitorElement` returned by `book1D()` etc. would have fully thread safe public interface (`const` for non-mutating functions, non-`const` for mutating)
>     * The `book1D()` etc. would take additional (optional) lambda that makes all the necessary modifications on the TH1 (see example below) by taking the TH1 as an argument
>       * Alternatively or in addition the `MonitorElement` could a method (e.g.) `mutate()` that takes a similar lambda and locks the mutex internally

  - does the `MonitorElement` still uses a lock in the functions that are part of its public interface ?

  - do you propose to remove the `getTH1F()` from the public interface, and expose the underlying root object only through the lambda ?

makortel:
@fwyzard 

> > ```
> > * The `MonitorElement` returned by `book1D()` etc. would have fully thread safe public interface (`const` for non-mutating functions, non-`const` for mutating)
> > * The `book1D()` etc. would take additional (optional) lambda that makes all the necessary modifications on the TH1 (see example below) by taking the TH1 as an argument
> >   * Alternatively or in addition the `MonitorElement` could a method (e.g.) `mutate()` that takes a similar lambda and locks the mutex internally
> > ```
> 
> * does the `MonitorElement` still uses a lock in the functions that are part of its public interface ?

Yes, because those functions would have to be internally thread safe.

(I want to note that there is a way to avoid the locks, but that would require larger interface changes towards the DQM analyzer modules, and my impression from all the discussion so far is that such changes are not preferred)

> * do you propose to remove the `getTH1F()` from the public interface, and expose the underlying root object only through the lambda ?

Yes, exactly.

schneiml:
Hi @makortel,

> First a few questions: the bookHistograms() continues to be called at beginRun time, right? ...

Yes -- the plan is to call `bookHistograms` *on each stream*, and de-duplicating the booked MEs in the (shared between streams) DQMStore. Actually, each stream instance will get its own ME instances, which potentially share the underlying MEData -- that is important in case streams switch lumis at different times, and we want to get per-lumi results: Then the `ME*`'s in the instance need to redirect to new histograms at different times. However, the `ME ` can *neither* assume to be accessed only from a single thread, since `DQMGlobalEDAnalyzer` will give the same ME object to all threads. So both `ME` and `MEData` need to have thread-safe interactions by default.

> What would happen if one of the streams proceeds to call accumulate() while some/rest are still waiting for bookHistograms() to be called (i.e. the underlying MonitorElement object would get calls to thread-safe and to thread-unsafe functions)?

That is actually a problem that I did not really have in view. I'd just enforce that all booking needs to be done before any events can be processed.

> What would you think about the following ...

That is a nice idea, probably the only thing that can really work. However, it does not solve all problems, and it has the big problem of "API changes"...

> The MonitorElement returned by book1D() etc. would have fully thread safe public interface (const for non-mutating functions, non-const for mutating) 

The problem with this is that if we are serious about reproduceability, we cannot have any `get` in this interface, and it would logically need to be write-only: as soon as reads can be reordered with writes, these reads depend on execution order and are not reproducible.

This is actually the core problem here: making all the interactions (except `getTH1` and friends) "thread safe" (as in, "data race free"), is easy and even enforced by having to take the lock for any interaction (even reads, ofc). This is the mode that I am currently moving towards. 

What we really need is a separation of data (bin content) and metadata (axis,  etc.): Metadata can be read but not written in processing, data can be written but not read. (Actually, even `setBinContent` and friends are unsafe in multi-threaded context due to lost update problems, but we could argue these away as "logic bugs".) And then we have to talk about harvesting and `edm::one` users, where this does and should not apply.

> ... lambda ...

This sounds good, except for the need for a migration. But the migration should also be straightforward, limited to moving around code in the booking call. The hard part (mostly for the migration path) is what to do with the modules that actually do rely on doing things that would be illegal in the future. Under this category that we vaguely call "legacy code" are "real" legacy `EDAnalyzer`s, *lots* of modules cheating the rules of the "threaded migration" and still doing things in `{begin|end}{Lumi|Run}` (even talking to `edm::Service<DQMStore>`, while formally "migrated"), explicit `edm::one` users, and harvesting, which is a bigger mess entirely, including code shared between reco and harvesting.

We could make it an opt-in, providing all the unsafe things for the legacy code, with questionable behaviour -- pretty much the idea of the `dqm::legacy` namespace. My plan here was to couple this to the DQM sequences cleanup, and more aggressively remove legacy modules if nobody claims responsibility for maintaining them. Then, we have a chance to get the main production sequences clean of unmigrated code.

So overall I guess I agree, but things are (as always) more complicated. Today I reworked the `ME` ecosystem to have the lock outside the DataFormat and be mostly const-correct, and I like the way it looks so far (It compiles, but I have not run anything yet): https://github.com/schneiml/cmssw/blob/dqm-new-monitorelement/DQMServices/Core/interface/MonitorElement.h#L47-L156

PS:
> (don't hold your breath, this topic is really more complex than my paragraph above suggests)

That matches my experience...

makortel:
Hi @schneiml, thanks for your detailed reply.

> > What would happen if one of the streams proceeds to call accumulate() while some/rest are still waiting for bookHistograms() to be called (i.e. the underlying MonitorElement object would get calls to thread-safe and to thread-unsafe functions)?
> 
> That is actually a problem that I did not really have in view. I'd just enforce that all booking needs to be done before any events can be processed.

I'm sorry to tell that the framework does not support the concept of blocking waits across the instances of a stream module. Attempts to do so can easily lead to deadlocks (similar to the dining philosophers problem).

> > What would you think about the following ...
> 
> That is a nice idea, probably the only thing that can really work. However, it does not solve all problems, and it has the big problem of "API changes"...

Could you be more specific (e.g. with some examples) what problems remain unsolved?

> What we really need is a separation of data (bin content) and metadata (axis, etc.): Metadata can be read but not written in processing, data can be written but not read. (Actually, even `setBinContent` and friends are unsafe in multi-threaded context due to lost update problems, but we could argue these away as "logic bugs".) 

I fully agree.

> And then we have to talk about harvesting and `edm::one` users, where this does and should not apply.

Do you know if there are any legitimate `edm::one` modules that are supposed to run in prompt reco (or any offline processing) that watch lumis (or runs)?

(well, actually already `edm::one` with per-lumi saving of histograms has the same performance issues as `edm::one` watching lumis...)

> > ... lambda ...
> 
> This sounds good, except for the need for a migration. But the migration should also be straightforward, limited to moving around code in the booking call. The hard part (mostly for the migration path) is what to do with the modules that actually do rely on doing things that would be illegal in the future. 

I want to remind that we are offering assistance for the migration of the client code for the new API.

> We could make it an opt-in, providing all the unsafe things for the legacy code, with questionable behaviour -- pretty much the idea of the `dqm::legacy` namespace. My plan here was to couple this to the DQM sequences cleanup, and more aggressively remove legacy modules if nobody claims responsibility for maintaining them. Then, we have a chance to get the main production sequences clean of unmigrated code.

Sounds good to me.

> So overall I guess I agree, but things are (as always) more complicated. Today I reworked the `ME` ecosystem to have the lock outside the DataFormat and be mostly const-correct, and I like the way it looks so far (It compiles, but I have not run anything yet): https://github.com/schneiml/cmssw/blob/dqm-new-monitorelement/DQMServices/Core/interface/MonitorElement.h#L47-L156

Thanks for sharing, by quick look it looks pretty good except the `clone` in `accessMut()` would be better as `unique_ptr<MutableMonitorElememntData>`). (do we know if `TH1::Clone()` is really thread safe?)

Seeing the `frozen_` and `mutable_` made us think a bit further. Theoretically, breaking the API now in one place may open a window of opportunity to clean up the API elsewhere as well so that the implementation would be cleaner and easier to maintain for the next decade. We are wondering, what if the `MonitorElement` class would not be used by both the `IBooker` and `IGetter`? Instead, one could separate the class used to fill histograms from the class used to read histograms, and thus simplify thread-safety issues and make the separate concepts more explicit to the users. (if we understood correctly, then then `frozen_` and `mutable_` would not be needed)

fwyzard:
Just to inject more entropy in an already complicated issue... what about looking at ROOT 7 experimental RHist ?
I don't mean we to adopt it today - but maybe to choose interfaces that could eventually be adapted to use it ?

schneiml:
@fwyzard That is worth considering if we go towards bigger changes. I don't know what `RHist` looks like atm, but it might have grown one or the other feature that I requested last year.

The rough reasoning to completely ignore it so far is:
- If we keep harvesting as is, there is absolutely no chance to get rid of `getTH1` *in general*. We can eliminate it in multi-threaded reco, but not in harvesting. So we'd need to convert on demand.
- Switching to a new histogram might make multi-threaded filling more efficient, at the cost of subtly changing behavior. Seeing all the options that people set (and also the not necessarily sane default values that people rely on), this sounds like a lot of well hidden traps that need expert review.
- Performance in filling is not really the concern. Typically (IIRC) the jobs where the `Fill` calls become relevant perf-wise are also IO limited. (Not sure about HLT use though, I mostly remember those things from the PixelPhase1 days.)

Now, if we do go towards bigger changes in harvesting (which I'd really like to see, but also which I am really scared of -- also re what @makortel is saying), the pay off might be better by getting rid of explicit ROOT dependences in the API entirely.

schneiml:
@makortel more comments:

> I'm sorry to tell that the framework does not support the concept of blocking waits across the instances of a stream module. Attempts to do so can easily lead to deadlocks (similar to the dining philosophers problem).

Oh. This could be a problem, then. Probably not in practice, since what the booking sets up is *probably* idempotent (so things are logically sound) and with all the locking, it should also be technically safe to change histogram configuration while filling, but that would be pretty ugly.

> Could you be more specific (e.g. with some examples) what problems remain unsolved?

This was meant as an intro to the following item.

> Do you know if there are any legitimate edm::one modules that are supposed to run in prompt reco (or any offline processing) that watch lumis (or runs)?

I have no evidence about this. I *think* the explicit `edm::one` should be primarily used for niche things like online DQM, stand-alone tools, etc., but there might be some stuff hiding on the offline production sequences that heavily relies on `one` behaviour. The reason that I cannot exclude this is that even in the `edm::stream` days after the threaded migration, we had a bunch of `edm::EDAnalyzer` legacy modules accessing DQMStore left on the prod sequences. Some of these might have turned `one` by now.

> I want to remind that we are offering assistance for the migration of the client code for the new API.

I acknowledge, and this changes the equation a lot. From my point of view, any non-trivial API changes are basically impossible. With some extra help, we might be able to fix some things properly. 

> (do we know if TH1::Clone() is really thread safe?)

That is actually an interesting question. I *really* hope so.

> frozen_ and mutable_

I actually had similar thoughts while writing this code (sort-of for the second time, the older "new DQMStore" prototype already had that copy-on-write logic).

The only reason to have this rather intricate contraption is to support the old harvesting API, pretending to modify things in place while actually creating new products. 

So, the first obvious thing would be to only have this logic in the harvesting ME, where we can get away with less threading related trouble. I decided to keep the ME implementation in one piece mostly because the reco ME is a subset of the harvesting ME and the locking is a good idea even in harvesting. It also makes live easier (I hope) in the messy cases that sit somewhere in between, like online. Having the full API fully safe is not that hard and should cause the fewest unexpected problems.

The second obvious thing is that we could get rid of this rather complicated and surprising c-o-w contraption by *just* changing the harvesting API. This is on the one hand very attractive -- the harvesting API that we have to provide is giant, ugly and tightly bound to ROOT, all for a rather small number of modules, many of which stopped doing anything useful years ago. I'd even go as far a banning custom code in harvesting *entirely*. The problem is, among these few modules are some of the more complicated and least understood that I know of, doing rather important things, such as the producing the certification summaries and the AlCa data. I don't know which modules (or rather which code paths within modules) are really important and how well these are understood by the experts (that is, if they could be re-written from a specification if needed).

For a random example, consider `SiPixelActionExecutor`:
- It concerns the Phase0 Pixel detector, now history, but still required for processing old data. The interest in the Tracker group to work on this code is probably limited. 
- There are many code paths that run conditionally. I'd not be surprised if some have not run in years.
- The code uses lots of interfaces on IGetter as well as ROOT objects, and is not really easy to understand. One could try to re-write it from a definition (like I did in the Phase1 days), but the resulting output would of course not be 1-to-1 compatible.

Anecdotally, one such module exists for every non-trivial subsystem of CMS. Pushing the subsystem developers to change these won't be very popular, and I don't feel confident changing this code myself.

For an even more "dumb" problem, many subsystems came up with smart ways to share code between harvesting and reco, which was easy when everything used the same ME. This elegant design now bites us, the borders between reco and harvesting code are blurry. Typically this could be fixed by templating out the ME type, but then even if we perform these changes, subsystem developers might not be comfortable with template code suddenly showing up in their modules.

schneiml:
please test

This it the new version with a `tbb::spin_lock` on the ME side, rathe than the MEData side. It seems to run in my local test, lets see if there is reasonable output.

schneiml:
Ok, Looks like it needs another rebase...

schneiml:
Looking at all the reported warnings,

```
In file included from /build/cmsbld/jenkins/workspace/ib-run-pr-tests/CMSSW_11_0_X_2019-10-17-2300/src/DQMServices/Core/interface/DQMStore.h:27:
/build/cmsbld/jenkins/workspace/ib-run-pr-tests/CMSSW_11_0_X_2019-10-17-2300/src/DQMServices/Core/interface/MonitorElement.h:483:18: warning: 'dqm::legacy::MonitorElement::getTH1' hides overloaded virtual function [-Woverloaded-virtual]
    virtual TH1 *getTH1() const {
                 ^
/build/cmsbld/jenkins/workspace/ib-run-pr-tests/CMSSW_11_0_X_2019-10-17-2300/src/DQMServices/Core/interface/MonitorElement.h:386:18: note: hidden overloaded virtual function 'dqm::impl::MonitorElement::getTH1' declared here: different qualifiers (unqualified vs 'const')
    virtual TH1 *getTH1();
```

This is exactly the intended effect, as far as I understand. Is there a way to silence this warning?

makortel:
@schneiml 

> Is there a way to silence this warning?

Add `using dqm::reco::MonitorElement::getTH1;` etc. into the definition of `dqm::legacy::MonitorElement` class.

schneiml:
please test

Rebased to yesterdays IB. Now it has chances to compile at least. Not sure about the warnings.

DQMServices/Core/interface/MonitorElement.h
```diff
@@ -15,58 +23,140 @@
...
+        return mut->access();
+      }  // else
+      auto frozen = frozen_.load();
+      if (frozen) {
+        // in case of an immutable object read from edm products, create an
+        // access object without lock.
+        return Access{std::unique_lock<dqmmutex>(), frozen->key_, frozen->value_};
+      }
+      // else
+      assert(!"Attempting to access MonitorElement not backed by any data!");
```
Dr15Jones: Make this an exception instead of an assert and the compiler warning will go away. You are getting the warning since the compiler knows assert can be turned off with CPP macros.

schneiml: Ah yes. I saw that one but forgot.

schneiml:
please test

now we got actual chances of success.

schneiml:
Ok, looks like I changed behaviour w.r.t scalars. I remember changing some logic in one place, maybe that piece was actually required.

I'll try to fix that, and then see if the non-scalar changes remain.

schneiml:
@makortel @Dr15Jones regarding making the API safer for booking (which will not happen in this PR, but we had the discussion here), I had another idea:

We could make the booking calls return a non-ME type that allows an implicit conversion to `ME*` (via `operator ME*()`). This should be drop in compatible for the code that does not use special features, and for the code that does, it could look similar to the "fluent interface builder"-pattern: each `set*()` returns `this`. The builder value would hold a `unique_lock` (or similar) like the `AccessMut` type does in this PR, and could completely replace it. Accepting a lambda for TH1-manipulations would be an option as well. The value then decays into a `ME*` releasing the lock, and the resulting ME does not have the `set*` methods but allows getting back the builder (or rather `Mutator`, or whatever) to make them accessible.

I am not exactly sure how `unique_lock` and returning `this` would interact, but I think it could be made to work. I wanted to build a prototype, but given that changing the tons of methods in MonitorElement always takes a lot of time, I thought I'd ask first what you think...

The reason for me to not provide the lambda-option right now is that it *also* requires duplicating the ME API into a second type, with methods that don't take the lock (the alternative would be reentrant locks, which are probably not great speed wise). They could go into the `AccessMut` helper, and then this approach looks very similar to what I just outlined except for the implicit conversion.

fwyzard:
Do you mean something like
```c++
MonitorElement* me = ibooker->book(...)->setSomeThing(...)->setSomeOtherThing(...);
```
where `book()` returns a `MonitorElementMutator*` with the `set...` methods ?

I don't think you can do it by pointer,  since you cannot define `operator MonitorElement*()` for a `MonitorElementMutator*` pointer - only for the `MonitorElementMutator` type.

You could work around it adding a final method like
```c++
MonitorElement* me = ibooker->book(...)->setSomeThing(...)->setSomeOtherThing(...)->finalise();
```
where `finalise()` also makes the mutator object invalid.

schneiml:
please test

Fixed an obvious bug involving scalars, that should fix the majority of differences. Some will remain, but lets see what is left with this fix.

schneiml:
@fwyzard Yes, that is exactly what I was thinking about.

Having the `finalise()` is not really an option, since that would mean it is no longer drop-in compatible. Unless we hide everything behind a set of new (renamed) booking APIs, which I guess would be possible, but I don't really like it (we need less APIs, not more...).

My idea was to throw around the `MEMutator` by value. It should be pretty light (only a pointer + the `unique_lock`) and all the methods could be inlined, so I think it should work. Though it would probably require a lot of wrestling with move-semantics, each method needs to return a move-ref and might need to have "ref-qualifiers for member functions" (just learned about that now).

fwyzard:
Ah, ok, then something more like
```c++
MonitorElement* me = ibooker->book(...).setSomeThing(...).setSomeOtherThing(...);
```

But I don't understand why you consider adding `finalise()` a change in API, while adding `setSomeThing(...)` and `setSomeOtherThing(...)` is fine ?

Apart form that, how do you prevent users from holding the `MEMutator` by value ?
I.e.
```c++
auto me = ibooker->book(...).setSomeThing(...).setSomeOtherThing(...);
```

schneiml:
@fwyzard `finalise()` would be an API break for *everybody*, even those 90% (unsubstantiated guess) of DQM code that don't use any `setSomeThing(...)`. Just changing the `setSomeThing` affects only the code that does use these methods, which is still a lot, but quite a bit less. 

Re holding the value: I can't stop that, and it would be perfectly fine, e.g. if you have whatever concerns about method chaining (a very real thing might be e. g. setting a lot of bin labels in a loop). The lock is in the value, and we are safe.

Now, of course, if you hold the `Mutator` in a member variable, things will deadlock very quickly, but I'd consider that better than a memory corruption (which you can still easily obtain by using APIs in illegal ways). (I do hope we have some sort of deadlock detection in our infrastructure, since people will probably find ways to abuse the API that we can't even imagine.)

One real concern that I see might be that people hold the `Mutator` in `auto` variables for longer than needed, which would make the `bookHistograms` method calls effectively synchronized against each other. That would be rather inefficient, but still, should be safe.

schneiml:
wow, that comparison is so clean that I am seriously wondering if the comparison tool is broken...

I hope/assume this is not the case, so looks like this is ready. It will need another rebase once #28038 is in (hopefully any hour now), and then maybe another for #28099. Hope the world does not move too much in the mean time...

fwyzard:
I see that I am confused.

> Re holding the value: I can't stop that, and it would be perfectly fine, e.g. if you have whatever concerns about method chaining (a very real thing might be e. g. setting a lot of bin labels in a loop). The lock is in the value, and we are safe.

What happens when multiple streams get each its own copy of the `mutator` ?
The mutex cannot be in the `mutator` object, because each stream would get a different one, making it pretty useless.
So the mutex has to be in the underlying `MonitorElement`, and as you pointed out ...
 
> Now, of course, if you hold the `Mutator` in a member variable, things will deadlock very quickly, but I'd consider that better than a memory corruption (which you can still easily obtain by using APIs in illegal ways).

... that makes it very easy for a DQM module to deadlock.

I would argue that the goal of a synchronisation API should be to make deadlocks *hard*, not *easy* ?

> (I do hope we have some sort of deadlock detection in our infrastructure, since people will probably find ways to abuse the API that we can't even imagine.)

I don't think so - the job would simply get stuck. 

> One real concern that I see might be that people hold the `Mutator` in `auto` variables for longer than needed, which would make the `bookHistograms` method calls effectively synchronized against each other. That would be rather inefficient, but still, should be safe.

True, I hadn't though of that.

By the way, what happens if one stream books an histogram, and starts filling it, and then an other stream re-books the same histogram, and tries to do some customisation like adjusting some axes ?
How do we prevent that from happening ?

Dr15Jones:
> By the way, what happens if one stream books an histogram, and starts filling it, and then an other stream re-books the same histogram, and tries to do some customisation like adjusting some axes ?
How do we prevent that from happening ?

The API Matti and I proposed, via the lambda, avoids that since the lambda only runs at the first histogram construction request.

schneiml:
Re `TH1`: I'd follow the policy to discourage direct ROOT dependence as much as possible. I think this is in the spirit of the original design, and might pay of in the future e.g. regarding ROOT7. 

If we go ahead and embrace using `TH1` and friends directly, that would make things a bit easier, but it might bite in the future. I'd be very much interested in the opinion of the CMSSW core team on that.

Re deadlocks: The "stupid" case of a deadlock can very easily happen here when a locking method is called from a locked context. This one is statically detectable and happens deterministically, so it is not very dangerous but also pretty impossible to systematically avoid. The second "stupid" case would be leaking a `Mutator`/`access` object, that might not be detected statically but it would also happen deterministically. The "complicated" case where multiple locks are involved should be excluded by design.

So it is not so much about detecting deadlocks that happen in production (that would be... bad), but rather about aborting the PR tests if they take excessively long (I am surprised I never triggered that so far -- as far as I know).

schneiml:
@Dr15Jones  Re "lambda as a booking parameter": I like that idea for this specific reason, but there are two, minor pragmatic problems:
- Changing the booking API, at all. It is just rather painful because there are so many methods involved. Now at least the `ConcurrentBooker` is gone...
- It does not handle the harvesting or `endRun` etc. (for one modules) cases. So we need another mechanism anyways.

The benefit of running the code only once is nice, but not worth that much given that most booking code will be idempotent anyways, just by nature. Relying on that would be a case of offloading this problem to the user, though...

makortel:
> So it is not so much about detecting deadlocks that happen in production (that would be... bad), but rather about aborting the PR tests if they take excessively long (I am surprised I never triggered that so far -- as far as I know).

We do abort PR tests if they take too long (I don't remember the threshold on the top of my head though). But PR tests taking long are not a proof of a deadlock (there can be many reasons/bugs that could cause jobs to run for a too long time). Nor is PR tests succeeding in time a proof of no deadlocks.

schneiml:
@fwyzard apart from what @Dr15Jones said -- the problem you point out was what triggered this discussion.

Re locks/mutator: The `mutex` is in the ME datastructure (somewhere, precise place to be discussed), the `unique_lock` (aka `lock_guard`) in the `Mutator`.

schneiml:
@makortel @fwyzard Re dead locks, again: That is why I want to point out that there are at least two types of deadlocks: those that happen deterministically, and are a bug like any other (even more harmless than a nullpointer access, since the nullpointer access might silently succeed, while taking a lock twice on the same callstack will *always* deadlock).

The other type, that happens only in certain interleavings and that requires more than one lock to happen, however, we need to avoid like the plague because they are impossible to catch once they get a chance to happen...

fwyzard:
If you do allow this

> the resulting ME does not have the `set*` methods but allows getting back the builder (or rather `Mutator`, or whatever) to make them accessible.

then the "complicated" case

> The "complicated" case where multiple locks are involved should be excluded by design.

is very easy to produce:
```c++
MonitorElement* me1 = ibooker.book(...);
MonitorElement* me2 = ibooker.book(...);

if (...) {
  me1->mutate().setSomething();
  me2->mutate().setSomething();
}
if (...) {
  me2->mutate().setSomething();
  me1->mutate().setSomething();
}
```
will potentially try to acquire the `me1` and `me2` locks in different order at different times, leading to a nondeterministic deadlock.

In fact, it could also happen if one does not allow the `MonitorElement` to recover the mutator:
```c++
MonitorElement* me1;
MonitorElement* me2;
if (...) {
  auto mut1 = ibooker.book(...);
  auto mut2 = ibooker.book(...);
  // set stuff
  me1 = mut1;
  me2 = mut2;
} else {
  auto mut2 = ibooker.book(...);
  auto mut1 = ibooker.book(...);
  // set stuff
  me1 = mut1;
  me2 = mut2;
}
```

If the condition evaluates to something different for different streams, we have a potential deadlock.

schneiml:
@fwyzard Re "time of booking": We could do that, but it involves another large transform in the existing DQM code.

The problem is, the booking code not just books all the histogram (it would be fine to do that only once), but also populates all the `ME*` member variables. So unless we refactor all 500+ DQM modules to do that initialization in some other way, we need to run the booking code on each stream instance independently.

And changing the way a few methods are called is easy, compared to changing how the subsystem code manages state. There are entire class hierarchies dedicated to this in some subsystems.

fwyzard:
> @fwyzard in your first code block, I don't see what the problem is -- there are never two locks taken at the same time. However your concern is very valid, and the second example demonstrates it. I guess that is a point in favour of the "lambda only" solution.

The problem arises when different streams run the code: the first stream could be in the first `if` block , while a second could enter the second `if` block.

makortel:
> > The benefit of running the code only once is nice, but not worth that much given that most booking code will be idempotent anyways, just by nature. Relying on that would be a case of offloading this problem to the user, though...
> 
> About this... do the `stream` modules have access to the `global` run and lumi transitions ?

They do with `{Run,LuminosityBlock}Cache`, `{Run,LuminosityBlock}SummaryCache`, and `{Begin,End}{Run,LuminosityBlock}Producer` extensions
https://twiki.cern.ch/twiki/bin/view/CMSPublic/FWMultithreadedFrameworkStreamModuleInterface

fwyzard:
> And changing the way a few methods are called is easy, compared to changing how the subsystem code manages state. There are entire class hierarchies dedicated to this in some subsystems.

Very true :-(

schneiml:
@fwyzard Re the example, I still don't see it: Both threads take a lock, release it, then take another, then release it. This is exactly the situation that we have in this PR now: Each operation is atomic and takes a lock. To get a deadlock, the `Mutator` object needs to live longer than a single line of code so there are two locks taken on the same thread (like in the second example).

I really hope I am not missing something here, because this is exactly what happens now...

fwyzard:
I think you are right, there are only temporaries involved, and they would be destroyed at the end of the statement.

It can still be a problem if the mutators are held on to:
```c++
if (...) {
  auto mut1 = me1->mutate();
  auto mut2 = me2->mutate();
  mut1.setSomething();
  mut2.setSomething();
}
if (...) {
  auto mut2 = me2->mutate();
  auto mut1 = me1->mutate();
  mut1.setSomething();
  mut2.setSomething();
}```

**Comment by schneiml** [comment]
@fwyzard yep, that is the dangerous case, and it is problematically easy to get it if we allow the `Mutator` temporaries.  

With lambdas it would be a bit harder to get that (it would require nested lambdas, and we should be able to spot those in code review), but then, it might still be required in cases (for sure if we want to also cover the harvesting-like operations).

**Comment by schneiml** [comment]
please test

rebased after #28038, with some history cleanup. I hope I did not screw up anything.

**Comment by schneiml** [comment]
Rebased to CMSSW_11_0_X_2019-10-24-2300 and fixed a number of issues.

There was only one merge conflict related to `oneDQMEDAnalyzer`, which is now deleted, but there where some compilation problems:
- `HGCalValidation/plugins/TICLPFValidation.cc` probably didn't exist before.
- Many modules had compilation errors due to the `dqm::reco::ME` vs. `dqm::legacy::ME` separation that this PR introduces. This is expected, but the problems where *not* spotted by the last round of tests which *did* include the change. I strongly suspect a bug in the PR test procedure here. (Not all dependencies spotted?)
- I commented some `softReset` calls rather than fixing them, since #28247 will delete them anyways. We have to decide in which order to put these PRs in; I'd prefer this one first unless there are unexpected problems now.

**Comment by schneiml** [comment]
please test

let's see what happens. As mentioned, I suspect that the previous round of tests actually failed to spot some compilation errors.

**Comment by schneiml** [comment]
please test

maybe now...

**Comment by schneiml** [comment]
+1

This looks good to me now. Further changes are of course expected (the booking cleanup discussed here, and more changes to actually use `MonitorElementData` completely), but I think we should move forward with this PR to prevent it from getting more conflicts.

@christopheralanwest @Dr15Jones @fwyzard @rekovic @benkrikler Cold you please have a look and eventually sign?

File: DataFormats/Histograms/interface/MonitorElementCollection.h [hunk 4]
```diff
@@ -0,0 +1,229 @@
...
+// For now, no additional (meta-)data is needed apart from the MEs themselves.
+// The framework will take care of tracking the plugin and LS/run that the MEs
+// belong to.
+
+// Only to hold the mergeProduct placeholder for now.
+class MonitorElementCollection {
+  std::vector<std::unique_ptr<const MonitorElementData>> data_;
+
+public:
+  void push_back(std::unique_ptr<const MonitorElementData>&& value) {
```
Dr15Jones: I think it is cleaner to do `push_back(std::unique_ptr<const MonitorElementData> value)` as that will match what other standard containers of `std::unique_ptr` will do.

DataFormats/Histograms/src/classes_def.xml
```diff
@@ -73,4 +73,10 @@
 
  <class name="DQMToken" ClassVersion="0"/>
  <class name="edm::Wrapper<DQMToken>" persistent="false"/>
+ <class name="MonitorElementCollection" persistent="true"/>
```
Dr15Jones: `persistent="true"` is the default so no reason to explicitly say it.

Dr15Jones: This class should declare the `ClassVersion="3"` and track the version hash.

DataFormats/Histograms/src/classes_def.xml
```diff
@@ -73,4 +73,10 @@
...
+ <class name="edm::Wrapper<MonitorElementCollection>" persistent="true"/>
+ <class name="MonitorElementData" persistent="true"/>
```
Dr15Jones: This class should declare the ClassVersion="3" and track the version hash.

DataFormats/Histograms/src/classes_def.xml
```diff
@@ -73,4 +73,10 @@
...
+ <class name="edm::Wrapper<MonitorElementData>" persistent="true"/>
+ <class name="vector<const MonitorElementData*>" persistent="true"/>
```
Dr15Jones: Shouldn't this be `<class name="vector<std::unique_ptr<const MonitorElementData>>" >` ?

Dr15Jones (changes requested):

DataFormats/Histograms/src/classes_def.xml
schneiml: Indeed, I changed the actual `MECollection` a lot for this PR but did not update any of this. Will have a look tomorrow.

schneiml:
please test

Rebased and addressed @Dr15Jones comments. Rather than setting up the DataFormats properly, I keep them `persistent=false` for now, since we won't have them as products in the near future.

Dr15Jones:
+1
For DataFormats/Histograms

fwyzard:
Can somebody profile the impact of these changes on the `FastTimerService`, `ThroughputService`, and the other DQM modules running inside the HLT, possibly *before* they are merged ?

schneiml:
@fwyzard I'd do that, if somebody provides me with reasonable instructions on how to run a realistic HLT-like config. Ideally outside P5, and from files available to the general public...

schneiml:
Rebased onto `CMSSW_11_0_X_2019-10-30-2300` and migrated the new/modified HGCAL validation modules on the way. Compiles locally, hope everything is still fine.

fwyzard:
@fwyzard wrote:
> Can somebody profile the impact of these changes on the `FastTimerService`, `ThroughputService`, and the other DQM modules running inside the HLT, possibly _before_ they are merged ?

@schneiml wrote:
> @fwyzard I'd do that, if somebody provides me with reasonable instructions on how to run a realistic HLT-like config. Ideally outside P5, and from files available to the general public...

<br/>
@Martin-Grunewald @gennai @Sam-Harper can you help @schneiml check the timing ?

Martin-Grunewald:
For example:
https://twiki.cern.ch/twiki/bin/view/CMS/TriggerStudiesTiming
https://twiki.cern.ch/twiki/bin/view/CMS/TimingStudiesRecipes2018

schneiml:
@Martin-Grunewald That gets me somewhere but not all the way there.

https://twiki.cern.ch/twiki/bin/view/CMS/TimingStudiesRecipes2018 Links to https://twiki.cern.ch/twiki/bin/viewauth/CMS/HLTCpuTimingReports2018 for 
> A guide on the most recent tests with data can be taken from the 2018 Timing Reports twiki. In particular, notice the setup used. 

Though I mostly see reports there.

https://twiki.cern.ch/twiki/pub/CMS/HLTCpuTimingReports2018/setupUsed_HLTv3.5V10.txt is something I can reproduce, and it also seems to run on CMSSW11 (though, should I change the global tag? If yes, to what?)

I cannot access the files that are recommended as input, looks like the paths are local to the TSG machines. https://twiki.cern.ch/twiki/bin/view/CMS/TriggerStudiesTiming hints that I can get data from EOS/xrootd, but which data should I use? Which data is on disk in sufficient amounts and representative for a timing study like this?

Also, how do I get the input files into the configuration? `--input myroot.root` or some other mechanism? How do I choose the number of threads? How many threads should I use? (I guess as many as possible, given we worry about locking here).

I'd appreciate a transcript that I can copy-paste into a `cmsdev` machine shell that sets up a configuration and runs it. I am afraid if I guess my way through I will not end up with a realistic setup.

schneiml:
+1 

This is getting increasingly urgent, with the end of CMSSW11 cycle approaching. As far as I see, there are no specific concerns, apart from the timing checks mentioned by @fwyzard, that could also be addressed once we have this PR merged.

fwyzard:
On Fri, 1 Nov 2019, 10:48 Marcel Schneider, <notifications@github.com>
wrote:

> +1
>
> This is getting increasingly urgent, with the end of CMSSW11 cycle
> approaching. As far as I see, there are no specific concerns, apart from
> the timing checks mentioned by @fwyzard <https://github.com/fwyzard>,
> that could also be addressed once we have this PR merged.
>

I disagree. The fast execution was one of the goals that guided the
implementation of the ConcurrentMonitorElement; replacing it with a
different solution without checking the impact on that same performance is
not acceptable IMHO.

And the "could be addressed later" argument simply did not work too many
times to be considered.

>

schneiml:
@fwyzard the implementation should be mostly equivalent, apart from one more level of indirection.

>  "could be addressed later" 

I'd like to address it today, if I could.

Dr15Jones:
+1

fabiocos:
@gennai @Sam-Harper following the comment https://github.com/cms-sw/cmssw/pull/28092#issuecomment-548462396 could you please address the question of @schneiml or point to someone able to do that? The point raised by @fwyzard is relevant for the integration of this code, and as 11_0_X needs to be used for productions we cannot afford to use it just to validate choices a posteriori.

Sam-Harper:
so I'm following up with our experts and hope to have a timing recipe for you soon. 

In general though @schneiml , we would prefer if it could be run on our dedicated timing machines vocms003, vocms004 as this allows us an apples to apples comparison. Would this be a problem for you do to so. If not, could you email me your cern username, I'll add you to the  people who can access then.

fioriNTU:
@Sam-Harper thank you very much for "activating" your experts so quickly! However, since this measurement has never be done in the context of the DQM core team, and moving on with this PR is quite urgent, it would be possible to have this test ran by the HLT expert himself? 

I do not know how long it can take, but I am sure you can take 1/10 of the time with respect to us. Do you think it is feasible?

Sam-Harper:
@fioriNTU , this is not normally our policy to do this for developers otherwise we quickly get overwhelmed. 

However in this particular case we are also having to commission the timing in 11_0_X as a reference (still iterating with experts), and therefore is it not any significant extra work to also run the timing for this PR at the same time.  So we will do so.

Sam-Harper:
So incase folks are wondering what happened, the test was run, realised it was inadequate for this type of change, re-run again. Those results have been looked at and furtther tests were needed which are on going.

schneiml:
@Sam-Harper thanks for the update!

It looks like this sill merges cleanly after #28297, so we might get away without another rebase (unless #28247 goes first).

Sam-Harper:
Okay timing is okay

image

image

No real difference within the jitter.

Sam-Harper:
btw the first results, I got a 10ms increase but it looks like it was just a random upset as re-running it multiple times we converged on the above result.

fioriNTU:
Thank you very much @Sam-Harper !

fabiocos:
@Sam-Harper @fwyzard @Martin-Grunewald are there other concerns/comments by HLT experts?
@christopheralanwest @tlampen @rekovic could you please check and comment in case?

fwyzard:
From my side, I'm happy with the result of the test Sam has done.

Martin-Grunewald:
+1

schneiml:
Thanks @Sam-Harper !

Btw, if you have reasonably easy instructions how to run these things now I am still interested (on the longer term), to do some profiling and see how much impact DQM has at all in these configurations. In Offline, it is typically very little, which is why I think that many speed hacks in the DQM code are out of place...

fabiocos:
+1

fabiocos:
@christopheralanwest @tlampen @rekovic please check this and comment in case for a possible further iteration, the changes in your are looks consequential to the heart of the PR

fabiocos:
merge

christopheralanwest:
+1

PR #28096, opened 2019-10-01
Enable pt-dependent JER-SF and phi-dependent JEC application in PAT

ahinzmann:
#### PR description:

Enable pt-dependent jet resolution scale factors in PAT. This is needed to apply Autumn18_V7 JER-SF, released last week.
Enable phi-dependent jet energy scale in PAT. This will be needed for 2018 JEC to deal with the HEM issue.
This PR has no impact with the current global tags in any release. However, it will enable using a new global tag (currently in the makings) with pt-dependent JER-SFs and later with phi-dependent JECs.

A test config to apply resolution smearing from a db file has been added under /test.
Here is an example db-file containing pt-dependent JER-SFs:
https://github.com/cms-jet/JRDatabase/blob/master/SQLiteFiles/Autumn18_V7_MC.db

#### PR validation:

Test config runs after inclusion of this PR. It was checked that this does not change the memory consumption, compared to using the old code with Autumn18_V1_MC.db (no pt-dependence).

### DIFF
--- a/PhysicsTools/PatUtils/test/runJERsmearingOnMiniAOD.py
+++ b/PhysicsTools/PatUtils/test/runJERsmearingOnMiniAOD.py
@@ -9,12 +9,14 @@ from Configuration.StandardSequences.Eras import eras
 
 process = cms.Process('PAT2',eras.Run2_2016)
 
+### Example how to check timing
 #process.Timing = cms.Service("Timing",
 #  summaryOnly = cms.untracked.bool(False),
 #  useJobReport = cms.untracked.bool(True)
 #)
 
-process.SimpleMemoryCheck = cms.Service("SimpleMemoryCheck",ignoreTotal = cms.untracked.int32(1) ) 
+### Example how to check memory
+#process.SimpleMemoryCheck = cms.Service("SimpleMemoryCheck",ignoreTotal = cms.untracked.int32(1) )
 
 # import of standard configurations
 process.load('Configuration.StandardSequences.Services_cff')
@@ -35,7 +37,7 @@ process.maxEvents = cms.untracked.PSet(
 # Input source
 process.source = cms.Source("PoolSource",
     fileNames = cms.untracked.vstring(
-    'file:////afs/cern.ch/user/h/hinzmann/workspace/tmp/102D73A7-5B87-E611-936E-0CC47A1E0472.root',
+    '/store/relval/CMSSW_11_0_0_pre6/RelValTTbar_13/MINIAODSIM/PU25ns_110X_upgrade2018_realistic_v3-v1/20000/F38B9A9F-4B4B-3D4B-8C9D-9A9B945194EF.root',
     ),
     secondaryFileNames = cms.untracked.vstring(),
     skipEvents = cms.untracked.uint32(145)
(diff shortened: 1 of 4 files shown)


slava77:
@cmsbuild please test

PhysicsTools/PatUtils/test/runJERsmearingOnMiniAOD.py
```diff
@@ -0,0 +1,128 @@
...
+process.load('Configuration.StandardSequences.FrontierConditions_GlobalTag_cff')
+
+process.maxEvents = cms.untracked.PSet(
+    input = cms.untracked.int32(1000)
+)
+
+# Input source
+process.source = cms.Source("PoolSource",
+    fileNames = cms.untracked.vstring(
+    'file:////afs/cern.ch/user/h/hinzmann/workspace/tmp/102D73A7-5B87-E611-936E-0CC47A1E0472.root',
```
slava77: private files from AFS should not be used in the released source code.
Please change to something available centrally.

PhysicsTools/PatUtils/test/runJERsmearingOnMiniAOD.py
```diff
@@ -0,0 +1,128 @@
...
+                ),
+
+            # Scale factors
+            cms.PSet(
+                record = cms.string('JetResolutionScaleFactorRcd'),
+                tag    = cms.string('JR_Autumn18_V7_MC_SF_AK4PFchs'),
+                label  = cms.untracked.string('AK4PFchs')
+                ),
+            ),
+        connect = cms.string('sqlite:Autumn18_V7_MC.db')
```
slava77: is this an extremely temporary solution?

Shouldn't the payloads be uploaded already to the DB, or is the expectation that this mechanism will always be done with a local db file?

Is this test file really needed in the release?
The SmearedPATJetProducer is apparently used in the standard workflows. So, to enable the new corrections one would ideally just need to run the standard PAT with a candidate GT.

ahinzmann: The main reason to provide this config was, to be able to test this before the global tag is available. It will come in a couple of days.
The SmearedPATJetProducer is only used for MET uncertainties in standard workflows, thus not the most direct test of it.
If the file is nevertheless not desired I can remove it?

slava77:
@cmsbuild please test

PhysicsTools/PatUtils/test/runJERsmearingOnMiniAOD.py
slava77: I think that a commented out variant provided as an example is OK

peruzzim:
Can we please proceed more quickly with the preparation of this PR? It's needed for NanoAODv6 (and in general for analysis). @ahinzmann have you implemented the requested changes?

ahinzmann:
yes, all changes are in

slava77:
+1

for https://github.com/cms-sw/cmssw/pull/28096  89b308a 
- jenkins tests pass and comparisons with the baseline show no differences in the PR tests by itself (the new features in JECs are not exercised)

santocch:
+1

fabiocos:
+1

PR #28099, opened 2019-10-01
DQM: Restructure the edm::one DQMEDAnalyzer base classes

schneiml:
#### PR description:

This is another set of changes extracted out of the "new DQMStore" [1] project.

Here, we apply some aesthetic changes to the old `one::DQMEDAnalyzer`, to make it more consistent with the `DQMGlobalEDAnalyzer`, as well as a number of batch renames: DQM modules are no longer allowed to override `endRun` or `endLuminosityBlock`, instead `dqmEnd*` variations are provided (like we already had in some other places).

As a side effect, `DQMEDAnalyzer` and `DQMOneEDAnalyzer` are made to no longer share any code; this was not really required, and the implementations are still identical, but we might have them diverge soon. On that occasion, I also fixed the production of DQMTokens: they are not really used for anything and it was fine that they were not produced correctly, but now things are closer to how the product-based DQM will look like.

[1] https://github.com/schneiml/cmssw/tree/dqm-new-dqmstore-on-CMSSW_11_0_0_pre5

#### PR validation:

There should be no functional changes, no output changes are expected.
The bin-to-bin comparison shows unexpected changes that can be explained by old bugs being fixed. See https://github.com/cms-sw/cmssw/pull/28099#issuecomment-542278728 .


schneiml:
@cmsbuild ping

Does that work or does one need super powers for this?

schneiml:
please test

This should be fine, assuming there is still output at all.

schneiml:
Rebased to CMSSW_11_0_X_2019-10-13-2300 and removed the method that caused warnings (it was dead code anyways).

schneiml:
please test

local compilation failed with obscure errors, probably just a unclean work area.

schneiml:
Ok, it looks like several plots that used to be empty now have entries again, possibly related to the per-lumi saving logic. This is not expected, but also not necessarily a bad thing. Requires further investigation.

schneiml:
A plausible explanation is that in the past, `endRun`/`endLumi` were called at the wrong time (or maybe not at all), so the changes done to MEs in these methods where lost. By renaming all, now plots filled in these methods work again.

This is controversial, since we did not really support modifying MEs in `endRun`/`endLumi` since the "threaded migration": With `edm::stream` based modules, each module sees partial statistics and certain things done in `endRun` etc. won't work correctly. After the `edm::one` migration, this was not a problem any more, but since we plan to move back to `edm::stream`, we'd appreciate if people would not rely on things happening in `endRun`. For modules that explicitly use `DQMOneEDAnalyzer` however, relying on `endRun` is fine and safe.

In conclusion, this PR accidentally fixes some bugs introduced at an unknown time in the past. However, we reserve the right to break some of these things in the future, since they might be impossible to support correctly.

schneiml:
+1

fwyzard:
> Ok, it looks like several plots that used to be empty now have entries again, possibly related to the per-lumi saving logic. This is not expected, but also not necessarily a bad thing. Requires further investigation.

@schneiml , could you point out what modules and plots are involved ?

Martin-Grunewald:
+1

schneiml:
@fwyzard Large chunks are
- METRate: https://cmssdt.cern.ch/dxr/CMSSW/source/DQMOffline/JetMET/src/METAnalyzer.cc#1031
- hActiveUnits: https://cmssdt.cern.ch/dxr/CMSSW/source/DQM/DTMonitorModule/src/DTDCSByLumiTask.cc#112
- AlcaBeamMonitor: https://cmssdt.cern.ch/dxr/CMSSW/source/DQM/BeamMonitor/plugins/AlcaBeamMonitor.cc

plus three random plots from DQMDCSInfoClient (known broken for a while), SiStripMonitorHardware/FEDHistograms, and ESIntegrityTask (as far as I can see).

fwyzard:
Thanks - nothing HLT related then :-)

rekovic:
+1

schneiml:
@santocch @peruzzim @christopheralanwest @tlampen Could you have a look at this and eventually provide the missing signatures?

This PR should not cause any problematic changes.

peruzzim:
+xpog

for the minor change to PhysicsTools/NanoAOD/plugins/NanoAODDQM.cc

fabiocos:
@santocch @peruzzim @christopheralanwest @tlampen the changes in your area look modest and technical. I would like to close this PR integration before building next pre-release, please have a look, I will move forward

christopheralanwest:
+1

fabiocos:
code-checks

fabiocos:
+1

fabiocos:
merge

santocch:
+1

PR #28110, opened 2019-10-03
Reconfigure particle flow config files

hatakeyamak:
#### PR description:

Reconfigure particleFlow_cfi and make clear the parameters used for PFEGammaFilters, PFMuonAlgo, i.e. different components of PFAlgo. Also used this opportunity to remove obsolete config parameters.

This is a follow-up of
https://github.com/cms-sw/cmssw/pull/27782
and fillDescriptions are introduced, and reduced ParameterSets use a little bit wrt PR27782

#### PR validation:

Compiles and run a few matrix tests with success (scenarios for 2017, 2018, and phase2)

This is not a backport PR.

@bendavid @jpata @guitargeek


hatakeyamak:
Actually, this will fail the jenkins tests that runs HLT. I am not familiar with what to do to update HLTrigger/Configuration/python/HLT_GRun_cff.py reflecting the changes here, and suggestions will be appreciated.

slava77:
@cmsbuild please test

RecoParticleFlow/PFProducer/plugins/importers/GeneralTracksImporter.cc
```diff
@@ -21,7 +21,9 @@ class GeneralTracksImporter : public BlockElementImporterBase {
         cleanBadConvBrems_(
             conf.existsAs<bool>("cleanBadConvertedBrems") ? conf.getParameter<bool>("cleanBadConvertedBrems") : false),
         debug_(conf.getUntrackedParameter<bool>("debug", false)) {
-    pfmu_ = std::unique_ptr<PFMuonAlgo>(new PFMuonAlgo(conf));
+    bool postMuonCleaning =
+        conf.existsAs<bool>("postMuonCleaning") ? conf.getParameter<bool>("postMuonCleaning") : false;
+    pfmu_ = std::unique_ptr<PFMuonAlgo>(new PFMuonAlgo(conf, postMuonCleaning));
```
Martin-Grunewald: Why is this existsAs needed here (and in some other locations) if the parameter is part of fillDescriptions ayway?

hatakeyamak: This comes from PFBlockProducer.cc (not PFProducer.cc) for which we haven't introduced fillDescriptions yet...

hatakeyamak:
As I wrote earlier, I wasn't sure what to update HLT configs, so these failures were sort of expected. I will appreciate advice on this.

Dr15Jones:
> As I wrote earlier, I wasn't sure what to update HLT configs, so these failures were sort of expected. I will appreciate advice on this.

If the HLT module uses a different module label, it is possible to have different `fillDescriptions` descriptions for different module labels.

Martin-Grunewald:
Is `bCalibSecondary` gone for good?
In that case put a customisation into HLTrigger/Configuration/python/customizeHLTforCMSSW.py
to kill this parameter in any instance of the producer. This would be temporary until the HLT menu in
ConfDB is updated with your new cfi from fillDescriptions. Newly added parameters should be already taken care of fillDescriptions. In any case it would be good if there is only ONE set of top-level py parameters for the producer, as this is what ConfDB can handle.

hatakeyamak:
Hi Martin. Thank you. Yes, bCalibSecondary should be gone for good (as far as I can tell, it was not used for quite some time). There are several others that are removed because they were not used or moved into sub-psets.

I am updating HLTrigger/Configuration/python/customizeHLTforCMSSW.py, following your suggestions.

hatakeyamak:
Wait. Need to check one more thing.

hatakeyamak:
Updated  HLTrigger/Configuration/python/customizeHLTforCMSSW.py based on comments from @Martin-Grunewald and made some cleanups.

Tested with matrix tests - 10024.0, 10824.0, 20434.0, 11634.0, 140.56, 158.0 and looks ok to me.

slava77:
@cmsbuild please test

hatakeyamak:
The failures might be due to some database access issue rahter than problem of this PR. I re-ran some failed tests locally, and they ran ok. Can the tests be re-triggered?

perrotta:
@cmsbuild please test

hatakeyamak:
Thank you @perrotta

hatakeyamak:
Just checking in to see how this PR looks for reviewers.

fabiocos:
@Martin-Grunewald @perrotta @slava77 latest test was technically positive, could you please have a look at this PR?

Martin-Grunewald:
+1

RecoParticleFlow/PFProducer/plugins/PFProducer.cc
```diff
@@ -333,3 +309,233 @@ void PFProducer::produce(Event& iEvent, const EventSetup& iSetup) {
...
+
+  desc.add<bool>("useEGammaElectrons", true);
+  desc.add<edm::InputTag>("egammaElectrons", edm::InputTag("mvaElectrons"));
+
+  desc.add<bool>("useEGammaFilters", true);
+  desc.add<bool>("useProtectionsForJetMET", true);  // Propagated to PFEGammaFilters
+
+  // For PFEGammaFilters
+  {
+    edm::ParameterSetDescription psd_PFEgammaFilters;
```
slava77: it looks like this PSet  description fully belongs to the PFEgammaFilters (not used anywhere else by the PFProducer). This is a pattern usually implies that the defaults better be defined by ```PFEgammaFilters``` 

A suggested approach for this for a general case is https://twiki.cern.ch/twiki/bin/view/CMSPublic/SWGuideConfigurationValidationAndHelp#A_Plugin_Module_Using_Another_He
Here, since the factory pattern is not used for the helper class, a simpler solution is to define in ```PFEgammaFilters```
```c++
static void
fillPSetDescription(edm::ParameterSetDescription& iDesc) {
    iDesc.add<double>("electron_iso_pt", 10.0);
 ...
}
```

RecoParticleFlow/PFProducer/plugins/PFProducer.cc
```diff
@@ -333,3 +309,233 @@ void PFProducer::produce(Event& iEvent, const EventSetup& iSetup) {
...
+    psd_PFMuonAlgo.add<double>("metFactorForRejection", 4.0);
+    psd_PFMuonAlgo.add<double>("metFactorForHighEta", 25.0);
+    psd_PFMuonAlgo.add<double>("ptFactorForHighEta", 2.0);
+    psd_PFMuonAlgo.add<double>("metFactorForFakes", 4.0);
+    psd_PFMuonAlgo.add<double>("minMomentumForPunchThrough", 100.0);
+    psd_PFMuonAlgo.add<double>("minEnergyForPunchThrough", 100.0);
+    psd_PFMuonAlgo.add<double>("punchThroughFactor", 3.0);
+    psd_PFMuonAlgo.add<double>("punchThroughMETFactor", 4.0);
+    psd_PFMuonAlgo.add<double>("cosmicRejectionDistance", 1.0);
+    desc.add<edm::ParameterSetDescription>("PFMuonAlgoParameters", psd_PFMuonAlgo);
```
slava77: similarly here, the defaults can be defined in ```PFMuonAlgo::fillPSetDescription```

RecoParticleFlow/PFProducer/plugins/PFProducer.cc
```diff
@@ -333,3 +309,233 @@ void PFProducer::produce(Event& iEvent, const EventSetup& iSetup) {
...
+
+  desc.add<double>("goodPixelTrackDeadHcal_minEta", 2.3);    // abs(trackRef->eta()) > X
+  desc.add<double>("goodPixelTrackDeadHcal_maxPt", 50.0);    // trackRef->ptError()/trackRef->pt() < X
+  desc.add<double>("goodPixelTrackDeadHcal_ptErrRel", 1.0);  // trackRef->ptError()/trackRef->pt() < X
+  desc.add<double>("goodPixelTrackDeadHcal_chi2n", 2);       // trackRef->normalizedChi2() < X
+  desc.add<int>(
+      "goodPixelTrackDeadHcal_maxLost3Hit",
+      0);  // max missing outer hits for a track with 3 valid pixel layers (can set to -1 to reject all these tracks)
+  desc.add<int>("goodPixelTrackDeadHcal_maxLost4Hit",
+                1);  // max missing outer hits for a track with >= 4 valid pixel layers
```
slava77: due to clang-format driven line breaks, it's more practical to put the comment above the line.
Anyways, these seem to fall under comments that better be known to the parameters with ```setComment```

RecoParticleFlow/PFProducer/plugins/PFProducer.cc
```diff
@@ -333,3 +309,233 @@ void PFProducer::produce(Event& iEvent, const EventSetup& iSetup) {
...
+                                           edm::InputTag("particleFlowClusterHF", "Cleaned"),
+                                       });
+
+  // calibration parameters for HF:
+  desc.add<bool>("calibHF_use", false);
+  desc.add<std::vector<double>>("calibHF_eta_step", {0.0, 2.9, 3.0, 3.2, 4.2, 4.4, 4.6, 4.8, 5.2, 5.4});
+  desc.add<std::vector<double>>("calibHF_a_EMonly", {1., 1., 1., 1., 1., 1., 1., 1., 1., 1.});
+  desc.add<std::vector<double>>("calibHF_a_EMHAD", {1., 1., 1., 1., 1., 1., 1., 1., 1., 1.});
+  desc.add<std::vector<double>>("calibHF_b_HADonly", {1., 1., 1., 1., 1., 1., 1., 1., 1., 1.});
+  desc.add<std::vector<double>>("calibHF_b_EMHAD", {1., 1., 1., 1., 1., 1., 1., 1., 1., 1.});
```
slava77: these are all different from what was in particleFlow_cfi.py before.
Is this intentional?
is this hardcoded calib machinery still needed?

RecoParticleFlow/PFProducer/plugins/PFProducer.cc
```diff
@@ -333,3 +309,233 @@ void PFProducer::produce(Event& iEvent, const EventSetup& iSetup) {
...
+
+  descriptions.add("particleFlow", desc);
```
slava77: it looks like only _cff file should be exposed to users for a consistent use.
For this case, since there were known use cases of ```particleFlow_cfi```, it is better to avoid confusion and name this differencely, e.g. use the default label ```pfProducer``` or ```particleFlowDefault```

RecoParticleFlow/PFProducer/src/PFAlgo.cc
```diff
@@ -23,7 +23,9 @@ PFAlgo::PFAlgo(double nSigmaECAL,
       calibration_(calibration),
       thepfEnergyCalibrationHF_(thepfEnergyCalibrationHF),
       connector_() {
-  pfmu_ = std::make_unique<PFMuonAlgo>(pset);
+  const edm::ParameterSet pfMuonAlgoParams = pset.getParameter<edm::ParameterSet>("PFMuonAlgoParameters");
+  bool postMuonCleaning = pset.existsAs<bool>("postMuonCleaning") ? pset.getParameter<bool>("postMuonCleaning") : false;
```
slava77: isn't ```PFAlgo``` used only by the ```PFProducer```?
If so, the ```postMuonCleaning``` was already added via the fillDescriptions there

slava77 (commented):
sorry for some delay in the review; I've accumulated some backlog

RecoParticleFlow/PFProducer/plugins/PFProducer.cc
hatakeyamak: These are not, but I didn't want to completely remove them. So, parameters are just set like these.

RecoParticleFlow/PFProducer/plugins/PFProducer.cc
hatakeyamak: Now, particleFlow_cfi.py is being removed. I think I can use e.g. particleFlowDefault, but I am not so sure why it's preferred.

RecoParticleFlow/PFProducer/src/PFAlgo.cc
hatakeyamak: Hmm. PFAlgo's fillDescriptions just has this parameter then? I am not sure I like the idea, but if it's really preferred, I think I can do it.

hatakeyamak:
Thanks @slava77 I will try to work on these suggestions later today or tomorrow, but I had a few questions which I posted above..

RecoParticleFlow/PFProducer/src/PFAlgo.cc
slava77: I thought this was the case from 
https://github.com/cms-sw/cmssw/pull/28110/files#diff-24b7d20f1967c466ee0593961eb6c237R325

RecoParticleFlow/PFProducer/plugins/PFProducer.cc
slava77: As in the current state of the PR, the particleFlow_cfi.py is not removed, it is instead replaced by the auto-generated particleFlow_cfi.py file which will be incomplete if anyone tries to use it.
the main reason to rename this auto-generated file is to avoid possible confusion for someone who had it in their old configuration (outside of the official CMSSW code tree).

RecoParticleFlow/PFProducer/src/PFAlgo.cc
hatakeyamak: now i see what you mean. thank you.

RecoParticleFlow/PFProducer/plugins/PFProducer.cc
Martin-Grunewald: @slava77 What do you mean here? For ConfDB parsing we'd need a cfi file with at least all top-level parameters...

slava77: @Martin-Grunewald 
I expect that you will use particleFlow_cff, the new standard reference file.
Perhaps I need to be otherwise educated on how you pick a file for parsing

Martin-Grunewald: @slava77 
If there is a cfi it picks that, rather than a cff, by default. What is in the cff and not cfi?

slava77: era-related modifications are in the cff

hatakeyamak: Era-dependent modifications used to be in cfi, but i moved them to cff, because now cfi is auto-generated, and I didn't know how to keep the era-related modifications in cfi which is now auto-generated.

slava77: @Martin-Grunewald 
to expand on my note in https://github.com/cms-sw/cmssw/pull/28110#discussion_r336455010: please clarify if there is a specific rule in the HLT confDB parsing which requires using the auto-generated cfi file (which is incomplete for offline use case) rather than the partly manually made cff file (which is fully usable for offline configs).
In other words, is there anything that would prevent you from using the particleFlow_cff?

Martin-Grunewald: @slava77 
Currently ConfDB parsing uses cfi by default if there is one, but I can tell it to use a cff. In this case, does the cfi simply contain less parameters, with the rest added by hand elsewhere or by fillDescriptions of other plugings/code?

slava77: in this case it should be the same parameters, just the values can change depending on the era/modifier, which is defined/applied in the cff.

Is there some magic memory from the past in the cfi file discovery mechanism? If we rename this one to particleFlowDefault_cfi (as proposed above), would ConfDb still be able to find it?

Martin-Grunewald: Yes, sure, it looks at any cfi it finds. Also, it is really more about the parameters themselves, not so much their precise values which may need adjustment for hlt anyway. So the name of the file could remain as is.

hatakeyamak: ok. thx for clarification.

slava77:
> Thanks @slava77 I will try to work on these suggestions later today or tomorrow, but I had a few questions which I posted above..

I just wanted to check if all questions on what to do were addressed.
@hatakeyamak please clarify.
Thank you.

hatakeyamak:
Yes, at least conceptually. I am just working on implementing suggestions now. I might ask for some help while implementing them.

RecoParticleFlow/PFProducer/plugins/PFProducer.cc
hatakeyamak: I have to admit I am still not sure how to do this and need help.
We don't really use "factory", so I am not sure how this recipe will fit it into this use case.

Some specific questions are:
+ PFEGammaFilters may be considered as a helper, but it's not a plugin module right now. We need to make it as a plugin?
+ Do we need to introduce a new "factory" that creates PFEGammaFilters as a new plugin?
I think I need a little more guidance to get this implemented.

hatakeyamak: Maybe a little more in detail, by looking at:
https://twiki.cern.ch/twiki/bin/view/CMSPublic/SWGuideConfigurationValidationAndHelp#A_Plugin_Module_Using_Another_He
I probably understand steps 1,2,3, but steps 4,5,6 are unclear (which files these lines need to go in?) to me right now.

slava77: the link to the twiki was given for a more complete reference for a general case.
For this specific case, please follow the suggested code snippets:

add in PFEgammaFilters.h,cc: (as already posted above)
```c++
static void PFEgammaFilters::fillPSetDescription(edm::ParameterSetDescription& iDesc) {
    iDesc.add<double>("electron_iso_pt", 10.0);
 ...
}
```
then in this file replace L349-L409 with 
```c++
PFEgammaFilters::fillPSetDescription(psd_PFEgammaFilters);
```

hatakeyamak:
Thank you @slava77. I think I implemented all suggested changes.

I also introduced fillPSetDescription for PFCandConnector with the same spirit as we used for PFEGammaFilters and PFMuonAlgo.

In PFMuonAlgo, we cannot get rid of getParameter with default values, as it is used not only by PFProducer/PFAlgo but by others as well, but I synchronized default values between getParameter's in constructor and fillPSetDescriptions, and I think it should have no consequence to reco/HLT outputs.

slava77:
@cmsbuild please test

hatakeyamak:
Comparisons look good to me.

fabiocos:
@hatakeyamak there is a conflict introduced by the integration of #28200 which, looking at the code, seems simple to solve keeping your formulation (unless you want to make an update)

hatakeyamak:
Thank you @fabiocos . I think I solved the conflict.
Does this PR look good now, reviewers?

slava77:
@cmsbuild please test

slava77:
+1

for https://github.com/cms-sw/cmssw/pull/28110	6756673 
    - I only inspected the value changes from the code diffs, additionally relying on lack of differences in the results

hatakeyamak:
Thanks @slava77.
Looks like this is still waiting for HLT signature. @fwyzard @Martin-Grunewald

Martin-Grunewald:
+1

fabiocos:
+1

Issue #28072, opened 2019-09-26
`EgammaV/ConversionValidator` DQM ouptut is unstable in workflow 11634


schneiml:
While validating  #28042 I noticed that many plots in the `EgammaV/ConversionValidator` hierarchy are unstable in the PR tests. This seems to only affect workflow 11634, while the same plots seem stable in WF 11624 (which *only* exists in 10_6, while 11634 is *only* in 11?).

We could blacklist it, but this might point to a deeper problem somewhere else.

```
# voms-proxy-init required
WF=11634
ME=EgammaV/ConversionValidator/ConversionInfo/convEtaAss
curl -E /tmp/x509up_u$(id -u) https://cmsweb.cern.ch/dqm/dev/data/json/samples | python -m json.tool  | grep -o '"[^"]*'$WF'.*"' | tr -d '"' > datasets
echo "Got $(wc -l datasets)."
for d in $(cat datasets); do 
  echo -n $d ';'
  curl -E /tmp/x509up_u$(id -u) https://cmsweb.cern.ch/dqm/dev/jsonfairy/archive/1$d/$ME ; echo
done > mes
echo "ME did not change in (at least) $(grep -c unsupported mes) tests"
echo "ME did change in $(grep -c title mes) tests"
```

Says:
```
ME did not change in (at least) 703 tests
ME did change in 36 tests
```
for wf 11624, but 
```
ME did not change in (at least) 101 tests
ME did change in 54 tests
``` 
for wf 11634.

Dr15Jones:
assign dqm, reconstruction

fabiocos:
I believe that the first appearance of the issue in a merged PR is for #28011 , so I would say that the problem is present already in 11_0_0_pre9

jfernan2:
Hi, I believe this issue has been solved

slava77:
> Hi, I believe this issue has been solved

@jfernan2 
do you happen to know the origin of the fix?

slava77:
+1

this is apparently resolved, likely not related to reco in the end

smuzaffar:
this has been resolved

PR #28114, opened 2019-10-04
Added a tool to list the contents of DQMIO metadata table

andrius-k:
#### PR description:

A a tool to list the contents of DQMIO metadata table.

#### PR validation:

Tool validated locally.

### DIFF
--- a/DQMServices/Components/scripts/dqmiodumpindices.py
+++ b/DQMServices/Components/scripts/dqmiodumpindices.py
@@ -1,0 +1,36 @@
+#!/usr/bin/env python
+from __future__ import print_function
+
+import uproot
+import argparse
+
+from prettytable import PrettyTable
+from collections import defaultdict
+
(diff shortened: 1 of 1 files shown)


andrius-k:
please test

andrius-k:
please test

fabiocos:
@andrius-k the DQM difference in the test is clearly unrelated to this PR, the script runs smoothly on the output of wf 136.85 but fails on the output of 1325.7 (nanoAOD).

andrius-k:
please test

@fabiocos Thanks for spotting this! The crash happened because I didn't take into account that ME type can be unknown. This is now fixed. And yes, the differences are clearly unrelated to this PR.

andrius-k:
+1

fabiocos:
+1

the code has been now protected against failure in wf 1325.7

schneiml:
@fabiocos, @andrius-k the comparison noise here is another case of #28072.

PR #28116, opened 2019-10-04
TICL: Create PF candidates as output of TICL

steggema:
#### PR description:

This PR prepares the integration of TICL into CMS Particle Flow by creating PFCandidates that can subsequently be integrated in the central CMS PF collection. It does so with the help of an intermediate data format, the TICLCandidate, by converting the TICL output (a list of tracksters) into TICLCandidates, and subsequently the TICLCandidates into regular PFCandidates. It also adds DQM plots for the such-produced PFCandidates.

Presentations were given in a RECO/AT meeting, https://indico.cern.ch/event/841640/contributions/3539177/attachments/1896768/3129567/2019_08_23_reco.pdf, and in an HGCAL DPG meeting, https://indico.cern.ch/event/849004/contributions/3567562/attachments/1910534/3156782/2019_09_18_hgcal_dpg.pdf

#### PR validation:

The workflow has been validated by enabling TICL as defined in RecoHGCal/TICL/python/ticl_iterations.py and checking the DQM output.

#### if this PR is a backport please specify the original PR: -

 @rovere @felicepantaleo @bendavid @hatakeyamak  (HGCal RECO and PF)


felicepantaleo:
please test workflow 20493.52

rovere:
Not sure what went wrong with the header-chek. I checked on lxplus the same IB used to test this PR and the offending header is there at its place.
I also checked on the most recent IB I could setup and, again, there's not missing header from TensorFlow.

kpedro88:
please test workflow 20493.52

steggema:
It seems like the tests I ran locally did not properly cover the hadronic iterations; I will update within the next day.

felicepantaleo:
please test workflow 20493.52

kpedro88:
@smuzaffar do you understand why this PR is failing the header check?

smuzaffar:
because of https://github.com/cms-sw/cmssw/blob/master/RecoHGCal/TICL/interface/GlobalCache.h#L7 , I think this should be removed.

rovere:
@smuzaffar could you elaborate a bit more? The include is valid and has been there before this PR. Is the header-check something new related to cms-bot PR testing?

davidlange6:
I guess the error is to indicate that Reco* packages are not intended to depend on PhysicsTools ones



On 8 Oct 2019, at 08:49, Marco Rovere <notifications@github.com> wrote:

﻿

@smuzaffar<https://github.com/smuzaffar> could you elaborate a bit more? The include is valid and has been there before this PR. Is the header-check something new related to cms-bot PR testing?

kpedro88:
@rovere looking at the usage in that file, it just has a pointer to a class from the tensorflow external. Rather than including the PhysicsTools interface, which is not used, in the BuildFile.xml you can put `<use name="tensorflow-cc" />`. It's also possible that RecoHGCal/TICL should have a BuildFile.xml; currently just the plugins folder has one.

smuzaffar:
@kpedro88 , this class is only used by `RecoHGCal/TICL/plugins` [a] , so beter to move it to `plugins` subdirectory (where we already have `tensorflow-cc` dependency)

[a]
RecoHGCal/TICL/interface/GlobalCache.h 
  - RecoHGCal/TICL/plugins/PatternRecognitionbyMultiClusters.cc 
  - RecoHGCal/TICL/plugins/PatternRecognitionbyCA.cc 
  - RecoHGCal/TICL/plugins/TrackstersProducer.cc RecoHGCal/TICL/plugins/HGCGraph.cc

smuzaffar:
looks like we do not need `RecoHGCal/TICL/interface`, all header should move to plugins directory

steggema:
I confirm that either adding <use name="tensorflow-cc" /> to the newly create BuildFile.xml or moving all previously existing headers in the interface directory to the plugins directory addresses the issue. I'll do the latter since it seems to be the preferred solution.

(My guess why the check was triggered is that I added a top-level BuildFile.xml in this PR; not sure the result of the check makes sense but it's not important for this PR.)

smuzaffar:
you are right @steggema , adding the top level BuildFile instructs scram to build a shared library RecoHGCalTICL and this cause scram to also test headers files using the dependency list of top level BuildFile. Please move the headers in plugins directlry and drop top level BuildFile too. It is only needed when you want to build a public shared library

rovere:
> you are right @steggema , adding the top level BuildFile instructs scram to build a shared library RecoHGCalTICL and this cause scram to also test headers files using the dependency list of top level BuildFile. Please move the headers in plugins directlry and drop top level BuildFile too. It is only needed when you want to build a public shared library

Ok, that makes sense, now.

perrotta:
please test workflow 20493.52

jfernan2:
+1
Usual inestability on 11634.0

DataFormats/HGCalReco/interface/TICLCandidate.h
```diff
@@ -0,0 +1,68 @@
...
+    TICLCandidate(const edm::Ptr<Trackster>& trackster)
+        : LeafCandidate(), time_(0.f), timeError_(-1.f), rawEnergy_(0.f), tracksters_({trackster}) {}
+
+    inline float time() const { return time_; }
+    inline float timeError() const { return timeError_; }
+
+    void setTime(float time) { time_ = time; };
+    void setTimeError(float timeError) { timeError_ = timeError; }
+
+    // inline const reco::TrackRef trackRef() const { return trackRef_; }
```
kpedro88: delete commented-out code

DataFormats/HGCalReco/interface/TICLCandidate.h
```diff
@@ -0,0 +1,68 @@
...
+
+    inline float rawEnergy() const { return rawEnergy_; }
+    void setRawEnergy(float rawEnergy) { rawEnergy_ = rawEnergy; }
+
+    inline const std::vector<edm::Ptr<Trackster> > tracksters() const { return tracksters_; };
+
+    void setTracksters(const std::vector<edm::Ptr<Trackster> >& tracksters) { tracksters_ = tracksters; }
+    void addTrackster(const edm::Ptr<Trackster>& trackster) { tracksters_.push_back(trackster); }
+    // convenience method to return the ID probability for a certain particle type
+    inline float id_probability(ParticleType type) const {
```
kpedro88: since there are other ParticleType enums defined elsewhere in CMSSW, probably best to reference this one explicitly as `ticl::Trackster::ParticleType`

DataFormats/HGCalReco/interface/TICLCandidate.h
```diff
@@ -0,0 +1,68 @@
...
+#include "DataFormats/Candidate/interface/LeafCandidate.h"
+#include "DataFormats/Common/interface/Ref.h"
+#include "DataFormats/HGCalReco/interface/Trackster.h"
+#include "DataFormats/Math/interface/Point3D.h"
+#include "DataFormats/TrackReco/interface/Track.h"
+#include "DataFormats/TrackReco/interface/TrackFwd.h"
+
+// A TICLCandidate is a lightweight physics object made from one or multiple Tracksters.
+
+namespace ticl {
```
kpedro88: at some point, I thought we had agreed to use the `ticl` namespace only for classes that don't have `TICL` in the name

RecoHGCal/TICL/plugins/TICLCandidateFromTrackstersProducer.cc
```diff
@@ -0,0 +1,143 @@
...
+  std::unique_ptr<TracksterMomentumPluginBase> momentum_algo_;
+  std::unique_ptr<TracksterTrackPluginBase> track_algo_;
+  float min_particle_prob_;
+};
+DEFINE_FWK_MODULE(TICLCandidateFromTrackstersProducer);
+
+namespace {
+  int pdg_id_from_idx(size_t i) {
+    switch (i) {
+      case 0:
```
kpedro88: use the actual enums for these case values if possible

RecoHGCal/TICL/plugins/TracksterGeneralTrackPlugin.cc
```diff
@@ -0,0 +1,42 @@
...
+  public:
+    typedef edm::Ptr<reco::Track> TrackPtr;
+    explicit TracksterRecoTrackPlugin(const edm::ParameterSet&, edm::ConsumesCollector&& iC);
+    void setTrack(const std::vector<const Trackster*>& tracksters,
+                  std::vector<ticl::TICLCandidate>& ticl_cands,
+                  edm::Event& event) const override;
+  };
+
+  TracksterRecoTrackPlugin::TracksterRecoTrackPlugin(const edm::ParameterSet& ps, edm::ConsumesCollector&& ic)
+      : TracksterTrackPluginBase(ps, std::move(ic)) {  // ,
```
kpedro88: delete commented-out character

RecoHGCal/TICL/plugins/TracksterP4FromEnergySumPlugin.cc
```diff
@@ -0,0 +1,97 @@
...
+  std::tuple<TracksterMomentumPluginBase::LorentzVector, float> TracksterP4FromEnergySum::calcP4(
+      const ticl::Trackster& trackster,
+      const reco::Vertex& vertex,
+      const std::vector<reco::CaloCluster>& calo_clusters) const {
+    std::array<double, 3> barycentre{{0., 0., 0.}};
+    double energy = 0.;
+    size_t counter = 0;
+
+    for (auto idx : trackster.vertices) {
+      auto fraction = 1.f / trackster.vertex_multiplicity[counter++];
```
kpedro88: check for possible divide by zero

RecoHGCal/TICL/plugins/TracksterP4FromEnergySumPlugin.cc
```diff
@@ -0,0 +1,97 @@
...
+      auto weight = calo_clusters[idx].energy() * fraction;
+      energy += weight;
+      barycentre[0] += calo_clusters[idx].x() * weight;
+      barycentre[1] += calo_clusters[idx].y() * weight;
+      barycentre[2] += calo_clusters[idx].z() * weight;
+    }
+    std::transform(
+        std::begin(barycentre), std::end(barycentre), std::begin(barycentre), [&energy](double val) -> double {
+          return val / energy;
```
kpedro88: check for possible divide by zero

DataFormats/HGCalReco/interface/TICLCandidate.h
fwyzard: I guess that'ss why there is a `typdef` few lines above: `typedef Trackster::ParticleType ParticleType;`

kpedro88: Ah, I missed it, thanks.

DataFormats/HGCalReco/interface/TICLCandidate.h
steggema: Done.

DataFormats/HGCalReco/interface/TICLCandidate.h
steggema: Removed ticl namespace.

RecoHGCal/TICL/plugins/TracksterGeneralTrackPlugin.cc
steggema: Done.

RecoHGCal/TICL/plugins/TICLCandidateFromTrackstersProducer.cc
steggema: Done and fully agreed that this is better.

RecoHGCal/TICL/plugins/TracksterP4FromEnergySumPlugin.cc
steggema: Check added.

RecoHGCal/TICL/plugins/TracksterP4FromEnergySumPlugin.cc
steggema: Check added.

felicepantaleo:
please test workflow 20493.52

jfernan2:
+1

kpedro88:
+upgrade

rovere:
@perrotta @slava77 
A kind reminder.

perrotta:
> you are right @steggema , adding the top level BuildFile instructs scram to build a shared library RecoHGCalTICL and this cause scram to also test headers files using the dependency list of top level BuildFile. Please move the headers in plugins directory and drop top level BuildFile too. It is only needed when you want to build a public shared library

@steggema , I am belatedly late in jumping in this PR. Looking at the thread I see this comment from @smuzaffar which suggests to move all headers into the ```\plugin``` directory, and drop the top level BuildFile too.
As of now, I see that still ```TracksterMomentumPluginBase``` and ```TracksterTrackPluginBase``` sit in the ```\interface``` (and ```\src```) directories, although being only used by the new plugins here:  this prevents removing the top level Buildfile.

I see that this PR does not fail the headers check any more, and therefore it should be considered as technically fixed. However, I would like to get confirmation from @smuzaffar that he is fine with the current arrangements. or if it would still be preferable to move to ```\plugin``` whatever is not needed to build a public shared library.

RecoHGCal/TICL/src/TracksterMomentumPluginBase.cc
```diff
@@ -0,0 +1,4 @@
+#include "RecoHGCal/TICL/interface/TracksterMomentumPluginBase.h"
+#include "FWCore/Framework/interface/MakerMacros.h"
+
+EDM_REGISTER_PLUGINFACTORY(TracksterMomentumPluginFactory, "TracksterMomentumPluginFactory");
```
smuzaffar: this is wrong, we shoud never register a plugin from a shared library

RecoHGCal/TICL/src/TracksterTrackPluginBase.cc
```diff
@@ -0,0 +1,4 @@
+#include "RecoHGCal/TICL/interface/TracksterTrackPluginBase.h"
+#include "FWCore/Framework/interface/MakerMacros.h"
+
+EDM_REGISTER_PLUGINFACTORY(TracksterTrackPluginFactory, "TracksterTrackPluginFactory");
```
smuzaffar: this is wrong, we shoud never register a plugin from a shared library

smuzaffar:
@steggema and @perrotta , there is no need to have these interface/ src/ files.  Currently these are registering edm plugins which will cause duplicate plugins if this library is linked by any other package.
Please move these to plugins/ and drop the top level BuildFile.

RecoHGCal/TICL/src/TracksterMomentumPluginBase.cc
kpedro88: @smuzaffar this is registering a factory, not a plugin. Doesn't this need to be in a shared library to be accessible?

smuzaffar: ah did not realize that @kpedro88 . Looking at cmssw , yes plugin factories are ok to be in shared libs.

RecoHGCal/TICL/src/TracksterTrackPluginBase.cc
smuzaffar: ignore the above comment

steggema:
@perrotta From the updated responses, I understand that there's no action required from my side. Please let me know if I misunderstood.

rovere:
From the last time I checked, pre11 is due today and it would be a pity if this Pr is not part of it.
Also, could someone confirm if [this](https://twiki.cern.ch/twiki/bin/viewauth/CMS/CMSSW_11_0_0) is still the confirmed schedule for `CMSSW_11`

perrotta:
> @perrotta From the updated responses, I understand that there's no action required from my side. Please let me know if I misunderstood.

Yes, as far as I understand @smuzaffar does not ask any more to move everything into the plugin directory and remove the top level Buildfile.
I just checked with him because that comment wasn't answered at the time.

smuzaffar:
true, PR is its current state looks good. No need to move tol level buildfile or interface src files. headers file checks look good.

perrotta:
The footprint on CPU time and output size of this PR is quite limited, at least for the non PU workflow 20493.52

- The addition to the AOD output is limited to the two new branches, with an overall contribution to the total size which is of the order of permil
```
      0.0 ->       199.0        199     NEWO   0.06     recoPFCandidates_pfTICLProducer__RECO.
      0.0 ->       255.1        255     NEWO   0.07     TICLCandidates_ticlCandidateFromTrackstersProducer__RECO.
   351284 ->      351737        454             0.1     ALL BRANCHES
```

- Nothing is added, for the moment, to the minAOD output

- The CPU consumptions of the new modules is minimal:
```
  delta/mean delta/orJob     original                   new       module name
  ---------- ------------     --------                  ----       ------------
       added      +0.00%         0.00 ms/ev ->         0.02 ms/ev ticlPFValidation
       added      +0.00%         0.00 ms/ev ->         0.03 ms/ev pfTICLProducer
       added      +0.00%         0.00 ms/ev ->         0.02 ms/ev ticlCandidateFromTrackstersProducer
  ---------- ------------     --------                  ----       ------------
Job total:  1.00444 s/ev ==> 1.00693 s/ev
```

perrotta:
+1
- TICLCandidate's are created as intermediate step tpowards PF candidates
- The implementation corresponds to what anticipated in the PR description, and also takes into account the following review
- Jenkins tests show that "normal" Phase2 workflows are not affected, while for workflows in which TICL reconstruction is activated the expected products show up in output

fabiocos:
+1

PR #28118, opened 2019-10-05
Muon FSR recovery for miniaod and nanoaod 

arizzi:
#### PR description:
This PR adds the producers for FSR recovery needed for next round of NanoAOD.
As the producer are generic enough to run also on top of MiniAOD they have been included as discussed with muon POG (https://its.cern.ch/jira/browse/CMSMUONS-300) outside the PhysicsTools/NanoAOD package

The nanoaod enabling code is not included in this PR and will be integrated once this enters cmssw.
A version of this PR with the nanoaod code (actually just some configuration) has been tested here:
https://github.com/cms-nanoAOD/cmssw/pull/413

#### PR validation:

Tests have been performed in the nanoaod PR linked above, no standard workflow is currently affected (inclusion during miniaod production step has not yet being discussed in XPOG)

@NTrevisani @Jorieger @peruzzim

### DIFF
--- a/MuonAnalysis/MuonAssociators/plugins/MuonFSRAssociator.cc
+++ b/MuonAnalysis/MuonAssociators/plugins/MuonFSRAssociator.cc
@@ -42,7 +42,7 @@ public:
     desc.add<edm::InputTag>("photons")->setComment("FSR photon collection to associate with muons");
     desc.add<edm::InputTag>("muons")->setComment("collection of muons to associate with FSR photons");
 
-    descriptions.add("MuonFSRAssociator", desc);
+    descriptions.addWithDefaultLabel(desc);
   }
   ~MuonFSRAssociator() override {}
 
--- a/MuonAnalysis/MuonAssociators/plugins/MuonFSRAssociator.cc
+++ b/MuonAnalysis/MuonAssociators/plugins/MuonFSRAssociator.cc
@@ -25,42 +25,36 @@
 //
 
 class MuonFSRAssociator : public edm::global::EDProducer<> {
-
 public:
-  
-  explicit MuonFSRAssociator(const edm::ParameterSet &iConfig):
+  explicit MuonFSRAssociator(const edm::ParameterSet& iConfig)
+      :
 
-    photons_ {consumes<pat::GenericParticleCollection>(iConfig.getParameter<edm::InputTag>("photons"))},
-    muons_   {consumes<edm::View<reco::Muon>>(iConfig.getParameter<edm::InputTag>("muons"))}
+        photons_{consumes<pat::GenericParticleCollection>(iConfig.getParameter<edm::InputTag>("photons"))},
+        muons_{consumes<edm::View<reco::Muon>>(iConfig.getParameter<edm::InputTag>("muons"))}
 
   {
-      
-      produces<edm::Association<std::vector<pat::GenericParticle>>>();
-      produces<edm::ValueMap<int>>("fsrIndex");
+    produces<edm::Association<std::vector<pat::GenericParticle>>>();
+    produces<edm::ValueMap<int>>("fsrIndex");
+  }
 
-    }
-
-   static void fillDescriptions(edm::ConfigurationDescriptions& descriptions) {
+  static void fillDescriptions(edm::ConfigurationDescriptions& descriptions) {
     edm::ParameterSetDescription desc;
     desc.add<edm::InputTag>("photons")->setComment("FSR photon collection to associate with muons");
     desc.add<edm::InputTag>("muons")->setComment("collection of muons to associate with FSR photons");
 
     descriptions.add("MuonFSRAssociator", desc);
-  } 
+  }
   ~MuonFSRAssociator() override {}
-  
+
 private:
-  
   void produce(edm::StreamID, edm::Event&, const edm::EventSetup&) const override;
 
-// ----------member data ---------------------------
+  // ----------member data ---------------------------
   const edm::EDGetTokenT<pat::GenericParticleCollection> photons_;
   const edm::EDGetTokenT<edm::View<reco::Muon>> muons_;
-
 };
 
 void MuonFSRAssociator::produce(edm::StreamID streamID, edm::Event& iEvent, const edm::EventSetup& iSetup) const {
-
   using namespace std;
 
   edm::Handle<pat::GenericParticleCollection> photons;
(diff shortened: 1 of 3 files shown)


peruzzim:
please test

peruzzim:
This PR will be needed for NanoAODv6, and the test failure seems unrelated. Could we please proceed with its review?

slava77:
@cmsbuild please test

slava77:
> This PR will be needed for NanoAODv6, and the test failure seems unrelated. Could we please proceed with its review?

@peruzzim 
this PR does not modify the Nano packages.
Does this enter via some dependencies into the nano workflows?

arizzi:
@slava77 see the PR description, this one adds two producers, the PR with nanoaod changes will be integrated first via the cms-nanoaod github repo (see https://github.com/cms-nanoAOD/cmssw/pull/416)

peruzzim:
this PR was tested successfully together with the new nanoAOD code

MuonAnalysis/MuonAssociators/plugins/MuonFSRAssociator.cc
```diff
@@ -0,0 +1,89 @@
...
+    produces<edm::Association<std::vector<pat::GenericParticle>>>();
+    produces<edm::ValueMap<int>>("fsrIndex");
+  }
+
+  static void fillDescriptions(edm::ConfigurationDescriptions& descriptions) {
+    edm::ParameterSetDescription desc;
+    desc.add<edm::InputTag>("photons")->setComment("FSR photon collection to associate with muons");
+    desc.add<edm::InputTag>("muons")->setComment("collection of muons to associate with FSR photons");
+
+    descriptions.add("MuonFSRAssociator", desc);
```
slava77: CMS style for a default module name is to start from a lower case ```muonFSRAssociator```
or can just use ```descriptions.addWithDefaultLabel(desc);```

MuonAnalysis/MuonAssociators/plugins/MuonFSRProducer.cc
```diff
@@ -0,0 +1,217 @@
...
+            "packedCandidate collection");
+    desc.add<edm::InputTag>("muons", edm::InputTag("slimmedMuons"))
+        ->setComment("collection of muons to correct for FSR ");
+    desc.add<double>("muonPtMin", 20.)->setComment("minimum pt of the muon to look for a near photon");
+    desc.add<double>("muonEtaMax", 2.4)->setComment("max eta of the muon to look for a near photon");
+    desc.add<double>("photonPtMin", 2.0)->setComment("minimum photon Pt");
+    desc.add<double>("deltaROverEt2Max", 0.05)->setComment("max ratio of deltsR(mu,photon) over et2 of the photon");
+    desc.add<double>("isolation", 2.0)->setComment("relative isolation cut");
+
+    descriptions.add("MuonFSRProducer", desc);
```
slava77: ```descriptions.addWithDefaultLabel(desc);```

slava77 (commented):
this review includes only very cursory check of the code in  MuonAnalysis; it's not in the reco category. This will not hold the reco signature.

slava77:
+1

for https://github.com/cms-sw/cmssw/pull/28118	46f1bfa 
- reco code is affected only in the DataFormats/PatCandidates/src/classes_def_objects.xml to add a few persisted data dictionary declarations; these apparently match the produced data in the new producers added to the MuonAnalysis/MuonAssociators
- jenkins tests pass

fabiocos:
@santocch could you please have a look?
@arizzi @peruzzim could you please consider @slava77 suggestion? Other than that I think the code may be integrated (I will take previous reco signature as good).

arizzi:
done

fabiocos:
@slava77 your suggestion has been implemented, providing a successful run of the nanoAOD code (see #28122 and I did a private cross check). As you had already signed the PR and nothing apart your suggestion has been modified, I will consider this as ok and integrate it

@peruzzim @santocch FYI

fabiocos:
+1

fabiocos:
merge

MuonAnalysis/MuonAssociators/plugins/MuonFSRProducer.cc
```diff
@@ -0,0 +1,217 @@
...
+        }
+      }
+      if (!closest)
+        continue;
+
+      // Check that is not in footprint of an electron
+      pat::PackedCandidateRef pfcandRef = pat::PackedCandidateRef(pfcands, iter_pf - pfcands->begin());
+
+      for (auto electrons_iter = electrons->begin(); electrons_iter != electrons->end(); ++electrons_iter) {
+        for (auto itr = electrons_iter->associatedPackedPFCandidates().begin();
```
Dr15Jones: This does not work. `associatedPackedPFCandidates()` returns a temporary object

https://github.com/cms-sw/cmssw/blob/a41ba6cb802a727726f33a31be291ca441534016/DataFormats/PatCandidates/interface/Electron.h#L260

so the calls to `electrons_iter->associatedPackedPFCandidates().begin()` and `electrons_iter->associatedPackedPFCandidates().begin()` are actually returning iterators to two different containers (both of which are gone during the iteration of the loop).

This was found by the address sanitizer
https://cmssdt.cern.ch/SDT/cgi-bin/logreader/slc7_amd64_gcc820/CMSSW_11_0_ASAN_X_2019-10-18-2300/pyRelValMatrixLogs/run/250202.118_ProdTTbar_13_pmx25ns+ProdTTbar_13UP18+DIGIPRMXUP18_PROD_PU25+RECOPRMXUP18PROD_PU25+NANOEDMMC2018_PROD/step4_ProdTTbar_13_pmx25ns+ProdTTbar_13UP18+DIGIPRMXUP18_PROD_PU25+RECOPRMXUP18PROD_PU25+NANOEDMMC2018_PROD.log#/

arizzi: shouldn't it loop forever then?

Dr15Jones: It's actually undefined behavior. I'm guessing that the two temporaries just happen to be made at the same point of the stack. This causes a luck accident that it just happens to work for now.

arizzi: ok then we may experience random crashes in large production.
@peruzzim I'm not sure if the nanov6 campaign was already submitted, but I'd suggest to fix this (just having a local copy of the vector)

Dr15Jones: I think a possibly better solution would be to switch to using ranged for loop
```C++
for(auto const& cand: electrons_iter->associatedPackedPFCandidates()) {
```
The language guarantees that the temporary container returned from the call to `electrons_iter->associatedPackedPFCandidates()` will be kept around for the lifetime of the for loop.

peruzzim: implemented this solution in #28221

jfernan2:
+1

PR #28123, opened 2019-10-07
[Python3] fix MakeBuildSet to work with both py2/3

smuzaffar:
#### PR description:

This fixes the MakeBuildSet script for python3. This is used for checking fwlite build set dependencies

#### PR validation:

Locally run it with both python2 and python3.

### DIFF
--- a/Utilities/ReleaseScripts/scripts/MakeBuildSet
+++ b/Utilities/ReleaseScripts/scripts/MakeBuildSet
@@ -1,4 +1,5 @@
 #! /usr/bin/env python
+from __future__ import print_function
 #pylint: disable-msg=W0403
 """
 MakeBuildSet - utilizes dependency discovery mechanism for Partial Releases.
@@ -38,7 +39,7 @@ def usage():
     """
     # Require minimum arguments and use options only to
     # override the defaults
-    print """
+    print ("""
 USAGE:
     """ + programName + """ -d <dir> [OPTIONS] [<package>]+
 OPTIONS:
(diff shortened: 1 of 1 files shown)


smuzaffar:
please test
although there is nothing in normal PR tests which will make use of this change. It only runs during IB tests

smuzaffar:
+core

Utilities/ReleaseScripts/scripts/MakeBuildSet
```diff
@@ -330,9 +331,8 @@ pitems = []
 if outputFormat != "all":
     outputFormats = [outputFormat]
 for item in outputFormats:
-    pkg = myPackages[item].keys()
-    pkg.sort()
+    pkg = sorted(myPackages[item].keys())
```
fabiocos: @smuzaffar for my understanding, is this change really needed or is it just for smarter coding? The ```list.sort()``` should still be available in python3 as far as I can see https://docs.python.org/3/howto/sorting.html

fabiocos:
+1

davidlange6:
yes, but .keys() is no longer a list in python3. It is an iterator instead

fabiocos:
@davidlange6 ok, you are right, so it cannot be used in that context, thanks

Utilities/ReleaseScripts/scripts/MakeBuildSet
smuzaffar: I could have used `pkg=list(myPackages[item].keys()); pkg.sort()` but I think current change should be faster if not slower.

PR #28129, opened 2019-10-08
Phase2-hgx223 Update the validity finding code

bsunanda:
#### PR description:

Correct the code which decides the valid detid of silicon cells

#### PR validation:

Use standard runTheMatrix code to check the changes

Nothing special

### DIFF
--- a/Geometry/HGCalCommonData/src/HGCalDDDConstants.cc
+++ b/Geometry/HGCalCommonData/src/HGCalDDDConstants.cc
@@ -568,15 +568,8 @@ std::pair<float, float> HGCalDDDConstants::locateCell(int cell, int lay, int typ
   return std::make_pair(x, y);
 }
 
-std::pair<float, float> HGCalDDDConstants::locateCell(int lay,
-                                                      int waferU,
-                                                      int waferV,
-                                                      int cellU,
-                                                      int cellV,
-                                                      bool reco,
-                                                      bool all,
-                                                      bool debug) const {
-
+std::pair<float, float> HGCalDDDConstants::locateCell(
+    int lay, int waferU, int waferV, int cellU, int cellV, bool reco, bool all, bool debug) const {
   float x(0), y(0);
   int indx = HGCalWaferIndex::waferIndex(lay, waferU, waferV);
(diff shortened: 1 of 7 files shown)


bsunanda:
@cmsbuild Please test

Geometry/HGCalCommonData/src/HGCalDDDConstants.cc
```diff
@@ -1495,6 +1473,39 @@ bool HGCalDDDConstants::waferInLayerTest(int wafer, int lay, bool full) const {
   return in;
 }
 
+std::pair<double, double> HGCalDDDConstants::waferPosition(int lay,
+                                                           int waferU,
+                                                           int waferV,
+                                                           bool reco,
+                                                           bool
+#ifdef EDM_ML_DEBUG
```
kpedro88: is this really necessary?

kpedro88:
are comparison changes expected here? if so, what causes the changes?

Geometry/HGCalCommonData/src/HGCalDDDConstants.cc
fabiocos: it looks a similar situation to another update of a few days ago, where we preferred at the end not to have the argument list conditional to a debug flag

bsunanda:
Earlier the validity of a cell was incomplete. So some silicon cells were declared valid though they were not. This affects cells at the edges in the so called virtual wafers. So some difference is expected.

Geometry/HGCalCommonData/src/HGCalDDDConstants.cc
bsunanda: Since debug is enclosed within ifdef, clang complains of unused variable

kpedro88: Is the debug flag really needed in addition to the ifdef?

bsunanda: I can remove ifdef part for this code. "debug" is needed for some special debugging

bsunanda:
@cmsbuild Please test

cvuosalo:
+1

fabiocos:
+1

PR #28141, opened 2019-10-09
Run3-hcx231 Preparing to remove accessing DDDCompactView from SimG4CMS/Calo code (Step 1)

bsunanda:
#### PR description:

Access spectra parameters to be used for simulation in the Geometry step and make it available for SIM so that SIM will not deal with DDDCompactView.

#### PR validation:

Use standard runTheMatrix.py

Nothing special

### DIFF
--- a/CondFormats/GeometryObjects/src/classes_def.xml
+++ b/CondFormats/GeometryObjects/src/classes_def.xml
@@ -166,7 +166,7 @@
    <field name="mode_"            mapping="blob"/>
    <field name="copiesInLayers_"  mapping="blob"/>
  </class>
- <class name="HcalSimulationParameters" class_version="0"> 
+ <class name="HcalSimulationParameters" class_version="1"> 
    <field name="attenuationLength_"     mapping="blob"/>
    <field name="lambdaLimits_"          mapping="blob"/>
    <field name="shortFiberLength_"      mapping="blob"/>
--- a/CondFormats/GeometryObjects/src/classes_def.xml
+++ b/CondFormats/GeometryObjects/src/classes_def.xml
@@ -167,14 +167,30 @@
    <field name="copiesInLayers_"  mapping="blob"/>
  </class>
  <class name="HcalSimulationParameters" class_version="0"> 
-   <field name="attenuationLength_"  mapping="blob"/>
-   <field name="lambdaLimits_"       mapping="blob"/>
-   <field name="shortFiberLength_"   mapping="blob"/>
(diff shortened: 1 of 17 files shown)


bsunanda:
@cmsbuild Please test

CondFormats/GeometryObjects/src/classes_def.xml
```diff
@@ -167,14 +167,30 @@
    <field name="copiesInLayers_"  mapping="blob"/>
  </class>
  <class name="HcalSimulationParameters" class_version="0">
```
ianna: shouldn't it be a new version?

Geometry/HcalCommonData/interface/CaloDDDSimulationConstants.h
```diff
@@ -0,0 +1,26 @@
...
+ */
+
+#include "CondFormats/GeometryObjects/interface/CaloSimulationParameters.h"
+
+class CaloDDDSimulationConstants {
+public:
+  CaloDDDSimulationConstants(const CaloSimulationParameters* cps);
+  ~CaloDDDSimulationConstants();
+
+  const CaloSimulationParameters* claosimpar() const { return calospar_; }
```
cvuosalo:
```suggestion
  const CaloSimulationParameters* caloSimPar() const { return calospar_; }
```

cvuosalo:
The title of the PR would be clearer if it were "Preparing to remove..." rather than "Trying to remove...". None of the files in this PR are from SimG4CMS.

CondFormats/GeometryObjects/src/classes_def.xml
bsunanda: Changed

Geometry/HcalCommonData/interface/CaloDDDSimulationConstants.h
bsunanda: Done

CondFormats/GeometryObjects/src/classes_def.xml
Dr15Jones: FYI, a class_version == 0 tells ROOT not to store that object.

bsunanda: Should I change it back to 0 then?

cvuosalo:
+1

ggovi:
+1

fabiocos:
+1

PR #28151, opened 2019-10-10
DAQ: updating module parameters and a new unit test

smorovic:
#### PR description:
Changes:
- DAQ modules: making some parameters tracked because they always need to be specified and updated by generated Python code in the online environment
- Cleanup: obsolete parameters in modules and test scripts, removal of obsolete scripts
- New unit test which runs startBU.py and startFU.py. This verifies that BU-FU file based protocol works correctly (using fake BU generator). DQM histogram and HLTRates pseudo-stream producers are added to the FU script.
- Removed separate adler32 check in DAQ service (for algorithm mostly obsolete except with very old raw files). Now the check is toggled using common veifyChecksum parameter
- adding DAQTestHistograms module for both unit tests and upcoming performance testing (will be able to produce enough histograms to emulate fastHadd CPU usage).

#### PR validation:
New unit test checks that parameter changes and new DQM test module execute correctly.


mommsen:
please test

smorovic:
Since the test succeeds on my setup, I removed log redirection to capture all output in what cmsbuild gives back. With this we should be able to see where it fails.

smorovic:
Not sure what is the problem. Writer outputs LS1 and LS2 data together with End of lumi file markers. While reader process  sees raw files, it doesn't see the marker and just keeps waiting in a loop (until cmsbuild terminates it). It is difficult to debug further as it works locally, so I will remove it from automatic unit tests but leave the test script which can be executed manually.

mommsen:
please test
Errors seem to be unrelated to the code changes.

mommsen:
+1

perrotta:
+1
- Modifications correspond to what declared in the PR description
- They are however all DAQ related
- Jenkins tests pass and show no differences in reco quantities

EventFilter/Utilities/plugins/BuildFile.xml
```diff
@@ -8,8 +8,10 @@
   <use   name="FWCore/Sources"/>
   <use   name="FWCore/Utilities"/>
   <use   name="IOPool/Streamer"/>
+  <use   name="DQMServices/Core"/>
```
fabiocos: @smorovic the addition of the DQM-related plugin adds this dependency that misses in the whole rest of the EventFilter area. Would not be possible to locate that plugin into a DQM package? At present the most obvious location I may see without creating a new package just for one class is ```DQM/HLTEvF```.

Are  you expecting to have more DQM-oriented code in this package, or within EventFilter?

fabiocos: @jfernan2 FYI

smorovic: @fabiocos, I agree, it does not have to go to EventFilter/Utilities. DQM/HLTEvF to me seems as a good location, and there is no dependence on the rest of the EventFilter code.

We don't foresee to add more EventFilter-specific DQM code. Intention for this module is to  generate a lot of objects (and bins) to produce high CPU load by fastHadd in performance tests on DAQ hosts.

fabiocos:
please test

fabiocos:
@smorovic thank you, I will consider previous signatures as valid, since this change looks very self-contained. Of course we need the DQM approval as well (as this is effectively a DQM module I believe it is also appropriate they are aware of that)

mommsen:
+1

perrotta:
+1
- As in https://github.com/cms-sw/cmssw/pull/28151#issuecomment-542412167, since nothing relevant for reco changed since then

jfernan2:
+1

Martin-Grunewald:
+1

fabiocos:
+1

smorovic:
hi @fwyzard,

thanks for pointing it out, I didn't know this detail. Seeing that there is also getUntrackedParameter without default parameter, it could be used to create error in case parameter is missing.

Problem I wanted to solve in the first place is that ConfDB GUI doesn't add empty untracked string by default to the configuration when adding new module (cases where empty string is default in fillDescriptions). It does add the parameter for the tracked string as well as non-string untracked types or untracked string with non-empty defaults. This behavior can be seen when EvFDaqDirector is added, and I found one other example:  ESProducer HcalTopologyIdealEP. Example of GUI adding the untracked string parameter with non-empty default is with ThroughputService.

Since I wanted, in particular, to make "fileBrokerHost" in EvFDaqDirector (untracked empty string by default) always present when creating the configuration, options without using tracked variable are to either change ConfDB (GUI) behavior, or workaround by not using empty string as default.

Martin-Grunewald:
Well, in the GUI simply add the empty string "" as a parameter value in the instances of the new module by hand (GUI editing).

smorovic:
That also seems OK. If default value is not provided in the fillDescriptions, cmsRun will fail due to this parameter missing from the configuration, so the maintainer of the menu will have to set it by hand in the GUI. At least it will prevent running without the parameter (which will be needed by a RCMS python snippet).
If that sounds fine, I can change to this with a new PR.

smorovic:
Hi,
I reverted to untracked in #28359.

PR #28155, opened 2019-10-10
Allow ConsumesCollector to be copied

wddgit:
#### PR description:

Allow ConsumesCollector to be copied. Add copy constructor and copy assignment operator.

This makes ConsumesCollector and ProducesCollector consistent. After experience with ConsumesCollector over time we decided that not having these functions caused more difficulties than the protection it added. The original reason not to include them was to try to make it harder to save a ConsumesCollector as member data and use it late, but it is anyway possible to save using the move constructor and lack of copy functions made ConsumesCollector a little difficult to use.

#### PR validation:

New functions used in a unit test. Outside of that nothing uses these new functions so nothing should be changed.

### DIFF
--- a/FWCore/Framework/interface/ConsumesCollector.h
+++ b/FWCore/Framework/interface/ConsumesCollector.h
@@ -39,9 +39,9 @@ namespace edm {
   class ConsumesCollector {
   public:
     ConsumesCollector() = delete;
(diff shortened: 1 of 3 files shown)


wddgit:
please test

Dr15Jones:
+1

fabiocos:
+1

PR #28180, opened 2019-10-15
Fix a scaling issue in MagneticField/VolumeBasedEngine

amadio:
#### PR description:

In commit 0f77253518f774b8ed2c627a202d378b57a34df6, the MagGeometry was made thread-safe by using an `std::atomic<>` for caching the last volume used. This means that all threads, despite working on unrelated events, share a common cached volume, which is not suitable most of the time. The atomic also creates problems for concurrency, since whenever a thread caches a volume, the  CPU cacheline is invalidated, forcing all other threads to fetch the value of the cached volume from memory every time, consequently suffering from higher latency. Since that volume is not suitable, it's likely that each thread will then replace it with a new value, which is not good for the other threads, and so on, significantly slowing down the whole simulation, since the magnetic field is called many times for each track to integrate its trajectory.

This problem has been profiled and benchmarked in Intel's VTune Amplifier against cmssw branch `CMSSW_11_0_GEANT4_X_2019-10-10-2300`. The results are shown below with labels `before` and `after` the changes from this pull request. In all cases, hyper-threading was disabled, and the threads were pinned using `taskset` with an appropriate number of CPUs. The machine where all runs were performed is a dual-socket [Intel Xeon E5-2698 v3](https://ark.intel.com/content/www/us/en/ark/products/81060/intel-xeon-processor-e5-2698-v3-40m-cache-2-30-ghz.html) with 64GB of RAM. We simulate minimum bias events at E<sub>cm</sub> = 13TeV, and vary the total number of events and number of threads in each analysis.  The `cmsDriver` command used to create the input file is shown below:
```
cmsDriver.py MinBias_13TeV_cfi --era Run2_2018 -s GEN,SIM --pileup=NoPileUp
--conditions auto:phase1_2018_realistic --geometry DB:Extended
--eventcontent=RAWSIM --datatier GEN-SIM --dirout=./ -n $nevents --nThreads $nthreads
--fileout file:scaling.root --mc --no_exec cmsRun MinBias_13TeV_cfi_GEN_SIM.py
```

#### PR validation:

There is no functional change introduced by this pull request. It is purely a performance optimization. The only thing that is different is that some barrel properties are computed only once at initialization rather than at every call to the magnetic field, as that seemed like a sensible thing to do as extra optimization.

Here is a scaling plot showing scaling before and after the changes:
<p align="center">
</p>

A throughput table before and after the changes is shown below. Throughput is calculated as (1024 events) / Δt, where Δt = (t₂ - t₁), t₁ is the start time for processing the first event, and t₂ is the end of the job (i.e. estimate of total time minus initialization time).

|Threads|Throughput (Events/s) [before]|Throughput (Events/s) [after]|Percent Change|
|-|-|-|-|
1|0.290|0.289|-0.34|
2|0.609|0.594|-2.46|
4|1.166|1.206|+3.43|
8|1.903|1.950|+2.47|
12|2.767|2.868|+3.65|
16|3.368|3.390|+0.65|
24|4.675|5.120|+9.52|
32|4.471|6.059|+35.52|

Comparison of hotspots analysis at 32 threads in Intel VTune Amplifier (before - after):
<p align="center">

</p>

Comparison of micro-architecture analysis at 32 threads in Intel VTune Amplifier (before - after):
screenshot

Bottom-up micro-architecture comparison of before and after for 32 threads in Intel VTune Amplifier:
screenshot

Best regards,
—Guilherme

### DIFF
--- a/MagneticField/VolumeBasedEngine/interface/MagGeometry.h
+++ b/MagneticField/VolumeBasedEngine/interface/MagGeometry.h
@@ -61,8 +61,6 @@ private:
 
   bool inBarrel(float R, float Z) const;
 
-  mutable std::atomic<MagVolume const*> lastVolume;  // Cache last volume found
-
   std::vector<MagBLayer const*> theBLayers;
   std::vector<MagESector const*> theESectors;
 
--- a/MagneticField/VolumeBasedEngine/interface/MagGeometry.h
+++ b/MagneticField/VolumeBasedEngine/interface/MagGeometry.h
@@ -59,7 +59,7 @@ private:
   // Linear search (for debug purposes only)
   MagVolume const* findVolume1(const GlobalPoint& gp, double tolerance = 0.) const;
 
-  bool inBarrel(const GlobalPoint& gp) const;
+  bool inBarrel(float R, float Z) const;
 
   mutable std::atomic<MagVolume const*> lastVolume;  // Cache last volume found
 
@@ -75,5 +75,6 @@ private:
 
   bool cacheLastVolume;
   int geometryVersion;
+  float R1, R2, Z0, Z1, Z2;
 };
 #endif
--- a/MagneticField/VolumeBasedEngine/src/MagGeometry.cc
+++ b/MagneticField/VolumeBasedEngine/src/MagGeometry.cc
@@ -68,6 +68,31 @@ MagGeometry::MagGeometry(int geomVersion,
   int nEBins = theESectors.size();
   if (nEBins > 0)
     theEndcapBinFinder = new PeriodicBinFinderInPhi<float>(theESectors.front()->minPhi() + Geom::pi() / nEBins, nEBins);
+
+  // Compute barrel dimensions based on geometry version
+  switch (geomVersion >= 120812 ? 0 : (geomVersion >= 90812 ? 1 : 2)) {
+    case 0:
+      R1 = 172.400f;
+      R2 = 308.735f;
+      Z0 = 350.000f;
+      Z1 = 633.290f;
+      Z2 = 662.010f;
+      break;
+    case 1:
+      R1 = 172.400f;
+      R2 = 308.755f;
+      Z0 = 350.000f;
+      Z1 = 633.890f;
+      Z2 = 662.010f;
+      break;
+    case 2:
+      R1 = 172.400f;
+      R2 = 308.755f;
+      Z0 = 350.000f;
+      Z1 = 633.290f;
+      Z2 = 661.010f;
+      break;
+  }
 }
 
 MagGeometry::~MagGeometry() {
@@ -110,7 +135,11 @@ MagVolume const* MagGeometry::findVolume1(const GlobalPoint& gp, double toleranc
   MagVolume6Faces const* found = nullptr;
 
   int errCnt = 0;
-  if (inBarrel(gp)) {  // Barrel
+
+  float R = gp.perp();
+  float Z = fabs(gp.z());
+
+  if (inBarrel(R, Z)) {  // Barrel
     for (vector<MagVolume6Faces const*>::const_iterator v = theBVolumes.begin(); v != theBVolumes.end(); ++v) {
       if ((*v) == nullptr) {  //FIXME: remove this check
         cout << endl << "***ERROR: MagGeometry::findVolume: MagVolume for barrel not set" << endl;
@@ -155,8 +184,11 @@ MagVolume const* MagGeometry::findVolume(const GlobalPoint& gp, double tolerance
   }
 
   MagVolume const* result = nullptr;
-  if (inBarrel(gp)) {  // Barrel
-    double R = gp.perp();
+
+  float R = gp.perp();
+  float Z = fabs(gp.z());
+
+  if (inBarrel(R, Z)) {  // Barrel
     int bin = theBarrelBinFinder->binIndex(R);
 
     // Search up to 3 layers inwards. This may happen for very thin layers.
@@ -198,16 +230,7 @@ MagVolume const* MagGeometry::findVolume(const GlobalPoint& gp, double tolerance
   return result;
 }
 
-bool MagGeometry::inBarrel(const GlobalPoint& gp) const {
-  float Z = fabs(gp.z());
-  float R = gp.perp();
-
+bool MagGeometry::inBarrel(float R, float Z) const {
   // FIXME: Get these dimensions from the builder.
-  if (geometryVersion >= 120812) {
-    return (Z < 350. || (R > 172.4 && Z < 633.29) || (R > 308.735 && Z < 662.01));
-  } else if (geometryVersion >= 90812) {  // FIXME no longer supported
-    return (Z < 350. || (R > 172.4 && Z < 633.89) || (R > 308.755 && Z < 662.01));
-  } else {  // versions 71212, 90322
-    return (Z < 350. || (R > 172.4 && Z < 633.29) || (R > 308.755 && Z < 661.01));
-  }
+  return ((Z < Z0) || (Z < Z1 && R > R1) || (Z < Z2 && R > R2));
 }
--- a/MagneticField/VolumeBasedEngine/src/MagGeometry.cc
+++ b/MagneticField/VolumeBasedEngine/src/MagGeometry.cc
@@ -39,8 +39,7 @@ MagGeometry::MagGeometry(int geomVersion,
                          const std::vector<MagESector const*>& tes,
                          const std::vector<MagVolume6Faces const*>& tbv,
                          const std::vector<MagVolume6Faces const*>& tev)
-    : lastVolume(nullptr),
-      theBLayers(tbl),
+    : theBLayers(tbl),
       theESectors(tes),
       theBVolumes(tbv),
       theEVolumes(tev),
@@ -177,13 +176,12 @@ MagVolume const* MagGeometry::findVolume1(const GlobalPoint& gp, double toleranc
 
 // Use hierarchical structure for fast lookup.
 MagVolume const* MagGeometry::findVolume(const GlobalPoint& gp, double tolerance) const {
-  // Check volume cache
-  auto lastVolumeCheck = lastVolume.load(std::memory_order_acquire);
-  if (lastVolumeCheck != nullptr && lastVolumeCheck->inside(gp)) {
-    return lastVolumeCheck;
-  }
+  static thread_local MagVolume const* lastVolume = nullptr;
 
-  MagVolume const* result = nullptr;
+  MagVolume const* result = lastVolume;
+
+  if (result && result->inside(gp))
+    return result;
 
   float R = gp.perp();
   float Z = fabs(gp.z());
@@ -225,7 +223,7 @@ MagVolume const* MagGeometry::findVolume(const GlobalPoint& gp, double tolerance
   }
 
   if (cacheLastVolume)
-    lastVolume.store(result, std::memory_order_release);
+    lastVolume = result;
 
   return result;
 }
--- a/MagneticField/GeomBuilder/test/stubs/MagGeometryExerciser.cc
+++ b/MagneticField/GeomBuilder/test/stubs/MagGeometryExerciser.cc
@@ -65,7 +65,7 @@ bool MagGeometryExerciser::testFindVolume(const GlobalPoint& gp) {
 
   if (vol == 0) {
     cout << "Test ERROR: No volume found! Global point: " << gp << " , GP Z = " << gp.z() << ", GP perp = " << gp.perp()
-         << " isBarrel: " << theGeometry->inBarrel(gp) << endl;
+         << " isBarrel: " << theGeometry->inBarrel(gp.perp(), fabs(gp.z())) << endl;
 
     // Try with a linear search
     vol = (MagVolume6Faces const*)theGeometry->findVolume1(gp, tolerance);


amadio:
@civanch Please review. Thank you.

civanch:
@amadio , this is a very good fix. You need to fix code-format issues and may be remove "cout", which is left in the code.

amadio:
@civanch, Thank you. I wondered if it was ok to remove, since it can be a performance issue, and it's easy to add back if needed, but decided to make only the minimal changes. I will fix formatting and make other improvements like removing the print out calls.

perrotta:
please test

namapane:
> 
> 
> @civanch, Thank you. I wondered if it was ok to remove, since it can be a performance issue, and it's easy to add back if needed, but decided to make only the minimal changes. I will fix formatting and make other improvements like removing the print out calls.

@amadio, great job. Regarding the couts: they occur only in case of errors so should not be a real performance issue, and would better be replaced with LogError than removed.

MagneticField/VolumeBasedEngine/src/MagGeometry.cc
```diff
@@ -148,15 +176,17 @@ MagVolume const* MagGeometry::findVolume1(const GlobalPoint& gp, double toleranc
...
-  }
+  static thread_local MagVolume const* lastVolume = nullptr;
+
+  MagVolume const* result = lastVolume;
 
-  MagVolume const* result = nullptr;
-  if (inBarrel(gp)) {  // Barrel
-    double R = gp.perp();
+  if (result && result->inside(gp))
+    return result;
```
slava77: generally, MagGeometry is not a singleton. So, using a static here will lead to incorrect behavior for the odd cases where MagGeometry is not unique.
It's perhaps safe enough to add a configurable to choose between atomic and thread_local way of caching.

@Dr15Jones @makortel is there another way to make ES products have per-stream/thread caches?

Dr15Jones: > is there another way to make ES products have per-stream/thread caches?

There is not. Please avoid thread_local as well since we've had problems with some systems based on the the total number of thread_local being used in an application.

Dr15Jones: If one wants to sped up searches, returning the last hit volume as part of the API and having the caller pass it back it is the best thread-safe way to do it.

namapane: @slava77, you are right that MagGeometry is not guaranteed to be unique, but we don't have real use cases where two instances would be present. At least it should not happen in any of the standard configurations, right?
@Dr15Jones: the first problem with returning the volume to the API is that volume search is a detail of the implementation of this specific field provider. Volume caching is an internal optimization specific of this provider. The general MF interface does not know about volumes and we have maps that are not volume-based; in particular we also serve a fast simplified parametrization of the tracker that is not based on volumes, cf. https://github.com/cms-sw/cmssw/blob/CMSSW_10_3_X/Configuration/StandardSequences/python/MagneticField_cff.py
The second problem that I see is that having to pass back the pointer and let the caller handle it will require substantial modification in CMSSW, and I can imagine that in some cases it will be hard to reuse the pointer without some refactoring (imagine a call to the field within a class method, or if the field is called in several parts of some functional operation)  so that practically it will lead to dropping this optimization in many cases.

MagneticField/VolumeBasedEngine/src/MagGeometry.cc
```diff
@@ -110,7 +134,11 @@ MagVolume const* MagGeometry::findVolume1(const GlobalPoint& gp, double toleranc
   MagVolume6Faces const* found = nullptr;
 
   int errCnt = 0;
-  if (inBarrel(gp)) {  // Barrel
+
+  float R = gp.perp();
+  float Z = fabs(gp.z());
```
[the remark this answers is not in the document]
amadio: In this case I think it's clear what the variables mean, as they're just the coordinate name, but I can rename to `r` and `z`, or `rho` and `z` since we are talking about cylindrical coordinates.

MagneticField/VolumeBasedEngine/src/MagGeometry.cc
amadio: I agree with both @Dr15Jones and @namapane that the best solution would be to let the caller handle the caches, but that requires some extra refactoring throughout other parts of the code. I made my changes considering that, even if not a singleton, realistically only a single instance of the class exists. The thread local cache does have a cost, but is probably a good compromise given the circumstances. Note that the main magnetic field has the same problem and we need to fix it as well. See, e.g., https://github.com/cms-sw/cmssw/blob/03d4c04f6c7c596ddad1a8d4193d322a7dcd9e0c/SimG4Core/MagneticField/interface/Field.h#L19-L20 and https://github.com/cms-sw/cmssw/blob/master/SimG4Core/MagneticField/src/Field.cc#L20-L39

MagneticField/VolumeBasedEngine/interface/MagGeometry.h
```diff
@@ -75,5 +73,6 @@ class MagGeometry {
 
   bool cacheLastVolume;
   int geometryVersion;
+  float R1, R2, Z0, Z1, Z2;
```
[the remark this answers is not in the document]
amadio: I tried to follow the style of surrounding code, but will change the names. These are barrel properties, and I made them uppercase to indicate they are "constants" (i.e. they don't change after construction).

amadio:
As for the printouts, I can see a small performance increase when I remove them. The branch is still there at runtime even when the `if` is almost always `false`. For a function called millions of times like this, it might make sense to remove or use `#ifdef` to not pay this cost.

namapane:
> 
> 
> As for the printouts, I can see a small performance increase when I remove them. The branch is still there at runtime even when the `if` is almost always `false`. For a function called millions of times like this, it might make sense to remove or use `#ifdef` to not pay this cost.

Which couts exactly are you referring to in this case (which line numbers)?

PS: as for the previous comment, I was actually argumenting **against** letting the caller handle the caches.

civanch:
@amadio , for SIM we do not need static thread local cache, because Field object is unique per thread, so any cache value/pointer are simply class members. I am personally struggle against overuse of thread local storage in Geant4 and we slowly reduce it size. I do not know about RECO field, is it really shared between threads?

amadio:
@civanch Can we then make the field in this PR also a unique instance for each thread? We may have some hurdles (like the vectors taken by reference at construction, then deleting their members in the destructor, which would prevent reusing the vectors for all threads), but it might be a good way to avoid the performance issue with `std::atomic` without needing to add a `thread_local` volume cache.

civanch:
@amadio , for SIM definitely, concerning RECO we need @namapane comment.

Dr15Jones:
>  I do not know about RECO field, is it really shared between threads?

Yes, all EventSetup objects are shared by all threads.

civanch:
@Dr15Jones , normally we should share static data or static objects with const methods, concrete field class, which access such static objects in case of SIM is unique per thread and cache may be implement there.

Dr15Jones:
One way to approach this would be to to add a new interface to the MagneticField class which allows for external caching (this can be done in a type agnostic way). Then have a 'local' MagneticField class which is a wrapper around the shared MagneticField and the cache which is passed into the shared MagneticField.

civanch:
I mean the same.

Dr15Jones:
@civanch 
Shared non-const static objects are against the CMSSW rules as they are inherently non-thread safe (plus other problems with long term maintenance).

civanch:
Sure, I agree.

namapane:
> 
> 
> One way to approach this would be to to add a new interface to the MagneticField class which allows for external caching (this can be done in a type agnostic way). Then have a 'local' MagneticField class which is a wrapper around the shared MagneticField and the cache which is passed into the shared MagneticField.

Uhm, I have a hard time figuring out what you mean here. Can you provide a more concrete example? 
My point is: in the reco code, the user gets  a pointer to a MF "engine" which in the most general case does not use volumes at all. There's just one specific type of engine which internally uses volumes, it's not the only one. The user code should not normally know whether caching is necessary or not. It is not clear to me how you would get around this?

namapane:
> 
> 
> @civanch Can we then make the field in this PR also a unique instance for each thread? We may have some hurdles (like the vectors taken by reference at construction, then deleting their members in the destructor, which would prevent reusing the vectors for all threads), but it might be a good way to avoid the performance issue with `std::atomic` without needing to add a `thread_local` volume cache.



> 
> 
> > I do not know about RECO field, is it really shared between threads?
> 
> Yes, all EventSetup objects are shared by all threads.

Indeed; and besides, a full volume-based map takes some memory (something like ~60 MB of raw data alone) which would be a waste to replicate in each thread, even if that was possible.

civanch:
We all agree that there is a static data as a field and static object in event setup to access this data. 

I like a proposal to change the interface allowing to keep cache in external classes, which calls this static object. For me, it is not a good solution if a static object has cache implemented as thread_local or atomic. In SIM we already instantiate field class per thread following Geant4 interface. The interface to the CMS field may be changed to be maximally effective. It is a very local place in SIM, we can do whatever reasonable modification.

namapane:
@civanch I assume you mean that in SIM the G4 field interface is instantiated per-thread, but that is just a wrapper of the actual field engine that is common to all threads (otherwise we would not be discussing this issue, which was observed in SIM if I understand correctly).
This said I do not understand what is the proposal. In RECO the usage of the field is ubiquitous, we can sure make changes to accomodate the need for SIM but I don't think we can substantially change the interface for RECO. (But maybe I'm missing something on what you have in mind)

Dr15Jones:
@civanch 
>We all agree that there is a static data as a field and static object in event setup to access this data.

I assume by static you do not mean C++ `static` (which I completetly disagree with) but with _immutable_, i.e. unchanging (which I do agree with).

civanch:
@Dr15Jones , yes immutable object with field data.
@namapane , I mean, that ideally cache should be in wrapper object and not in field engine.

namapane:
> 
> 
> @Dr15Jones , yes immutable object with field data.
> @namapane , I mean, that ideally cache should be in wrapper object and not in field engine.

That does not really work because this "volume" thing is an implementation detail of *this specific* engine. It's only pertinent to the specific implementation of one specific field engine. This specific engine can be optimized using a cache, other engines know nothing about volumes.
We can make the cache be part of the package interface, but that's weird because it makes the interface dependent from a detail of one specific implementation, which is not the only one.

civanch:
@namapane , then probably we should merge this PR and have a homework to think about better solution. I suspect, that SIM wrapper may be slightly optimized.

What I do not know: how many engines we have? All are used?

namapane:
@ civanch , I definitely agree the SIM wrapper could be optimized and the MF interface can also possibly be adapted to such optimization if needed, but we should be careful with changes that require a lot of recoding in reco.
Regarding your question, besides the volume-based map engine we have a number of parametrizations; one (ParabolicParametrizedMagneticField) a fast and simplified model served by default the MF standard sequences for the use of some part of tracker tracking where the full accuracy is not necessary; another one (OAEParametrizedMagneticField) is an accurate parametrization of the inner region and is actually used as a component to the volume-based map; then we have one extremely accurate but extremely slow parametrization (PolyFit2DParametrizedMagneticField) that is optional and used only for very specific studies. In addition we also have a constant-B map, that is used for aything with magnet off (rare but not unheard of).

slava77:
IIUC, the proposal was to add an interface with caching.
E.g. ```MagneticField::inTesla(const GlobalPoint& gp, void* cache) const = {inTesla(gp);} ```
This can be overloaded for the VolumeBasedMagneticField where it will pass .
I guess that the pointer to the MagGeometry call; in fact, MagGeometry can even be changed slightly to reuse the same ```::findVolume(gp, chache)``` method for both its current interthread-ineffective internal caching and for this new interface.

The new ```inTesla(gp, cache)``` can be adiabatically added where this call optimization becomes useful/important.

Dr15Jones:
@slava77 
The interface would probably be
```C++
MagneticField::inTesla(const GlobalPoint& gp, void const*& cache) const
```
As that allows the cache address to be modified. One could even do
```C++
MagneticField::inTesla(const GlobalPoint& gp, std::any& cache) const
```
however the overhead of using `std::any` seems pretty high when I used compiler explorer to investigate.

In either case, it could be good to use a wrapper anyway to hide the cache
```C++
namespace local {
class MagneticField {
    public:
    MagneticField( ::MagneticField const& field) : field_(field), cache_(nullptr) {}

   float inTesla(  const GlobalPoint& gp) const {
       return field_.inTesla(gp, cache_);
    }
    private:
        ::MagneticField const& field_;
        void const* cache_;
};
}
```
Then we modify the client code to use `local::MagneticField` rather than `MagneticField` directly.

namapane:
I didnt't know that void* was an option; it used to be somehow deprecated at some point.
So if I understand correctly it will be the responsibility of the caller to create the local::MagneticField out of the ESHandle<MagneticField> retrieved from the EventSetup, otherwise the code will keep working, but with no cache, right?
A question is who is going to implement this for RECO; a quick and not fully filtered search:
https://cmssdt.cern.ch/dxr/CMSSW/search?q=type-ref%3AMagneticField+&case=true&redirect=true

It should be noted that caching is essentially irrelevant in the tracker region, where a parametrization is used anyhow. So for tracker tracking it should not matter (unless for any reason there are extrapolations outside the tracker region where the parametrization is defined). Also, tracker tracking uses both the full map and a fast simple parametrization for some operations where accuracy is not important; users of the latter does not need to be updated. The main place where caching is critical is for propagators that are supposed to be used outside the tracker regions. My impression is that a switch to local caching would be better guided by timing studies to determine where the cache is really necessary than changing it everywhere. Another possibility would be to do some studies to track the fraction of cache hits for each user by instrumenting the code. Timing studies would be anyhow necessary to check that we don't degrade performance for tracking for any reason.
This is probably a nice optimization task, but I think it would require some time, maybe it makes sense to merge this PR in the meanwhile.

slava77:
> maybe it makes sense to merge this PR in the meanwhile.

based on my review and comments from @Dr15Jones , the current state of the PR is not for merging.

Dr15Jones:
I would not actually use a `void *` in the user facing API, instead I'd used a new type which expresses the intent
```C++
class MagneticFieldCache {
public:
    MagneticFieldCache(): cache_{nullptr} {}
   template<typename T> void set(T const* iCache) {cache_ = iCache;}
   template<typename T> T const* get() const {return reinterpret_cast<T const*>(iCache); }
private:
    void const* cache_;
};
```
The `local::MagneticField` class would then hold a `MagneticFieldCache` and we'd use
```C++
MagneticField::inTesla(const GlobalPoint& gp, MagneticFieldCache& cache) const
```
This is just as efficent as using a `void*` but allows for better maintenance in case we we a more sophisticated caching mechanism in the future.

amadio:
Given the discussion so far, I'm not sure what's the best course of action for this pull request. If some version of it is intended to be merged after modifications, please let me know. Otherwise, please feel free to close this and let me know if/how this will be handled by CMS. Thank you.

namapane:
What about a mixed approach where we add the interface for optional external caching, but keep internal caching as a default. So SIM will be able to use this optimization, and more use cases could be progressively added for RECO after proper profiling, without having to change and profile everything in CMSSW at the same time?

slava77:
> What about a mixed approach where we add the interface for optional external caching, but keep internal caching as a default. So SIM will be able to use this optimization, and more use cases could be progressively added for RECO after proper profiling, without having to change and profile everything in CMSSW at the same time?

Yes.
I had the same in mind in https://github.com/cms-sw/cmssw/pull/28180#issuecomment-544140956 with the new interface available for "adiabatic" replacements.

namapane:
Makes sense. Compared to your comment I think that Dr15Jones's proposal of 
`MagneticField::inTesla(const GlobalPoint& gp, MagneticFieldCache& cache)` const
won't allow to reuse the same  MagGeometry::findVolume(gp, chache) for both internal and external cache, but it will be straightforward to move the internal cache up to `MagGeometry::fieldInTesla(const GlobalPoint& gp)` and add a  separate `MagGeometry::fieldInTesla(const GlobalPoint& gp, MagneticFieldCache& cache)` since these very simple methods.
I need to do some work on the subsystem in the next weeks and I could try to add this interface as well in the PR if you wish. It will take some time (couple of weeks I hope) so if you need this to proceed quicker it may be better to develop it independently.

slava77:
>  I need to do some work on the subsystem in the next weeks and I could try to add this interface as well in the PR if you wish. It will take some time (couple of weeks I hope) so if you need this to proceed quicker it may be better to develop it independently.

The timeline sounds good to me.

IIUC, the plan is to keep the current (before this PR) default caching code, which remains effective in a single thread; for the effective multi-thread use we will have a new interface method added with a cache reference or pointer as an argument. The new method can be used to replace the old one (with default internal caching strategy) in all affected callers to enable local caching.

slava77:
-1

Following the preceding discussion, the solution proposed in this PR needs to be implemented in a different way.
I added #28284 to keep track of the progress.

fabiocos:
@amadio @namapane given the outcome of the discussion, what is the plan? #28284 is keeping track of the issue, and a different implementation is needed, but it will not come immediately. Are you planning to update at a later stage this PR, or do you prefer to close it and come back with a fresh new one later one?

fabiocos:
hold

namapane:
@fabiocos, I personally have a preference for a fresh PR instead of patching this one; however commit 6d4970c from this PR could be integrated directly if that proceeds quickly, being an independent optimization. If that will take long, better moving it as well to a new PR.
This is of course just my preference, if @amadio prefers doing otherwise (e.g. implementing the agreed solution himself in this PR) I have no objection.

fabiocos:
@amadio what is the status of this PR at this point?

amadio:
In  the last simulation meeting where I presented this along with some other performance analysis of simulation, it was not clear to me from the discussions that the optimization mentioned above is safe to go in as is, since there may be cases when the geometry changes (multiple runs in same process). Therefore, I will close this pull request and leave this to be handled in a new pull request, as suggested above, by someone from CMS with more knowledge of other areas that will be affected by these changes (e.g. reconstruction).

namapane:
I am coming back to the proposal of extending the API for external caching discussed here. I've been looking more closely to this and would like to implement it in the next few weeks. Before doing this, I have some questions:
- @Dr15Jones, @slava77, do you still consider the mechanism drafted in the comment above
(https://github.com/cms-sw/cmssw/pull/28180#issuecomment-544264803) to be the way to go? 
-I have a concern on the validity of the external cache. Are we guaranteed that there is no scenario where the map is re-created while users still keep the cache, which would then essentially become a dangling pointer? In principle the map can be re-created at IOV boundaries, e.g. if the magnet current changes. I'm not sure if this is a real concern - this does not happen of course for SIM and for reco of MC, but couldn't it happen (at least in principle) for HLT? Is there a guarantee that all possible user modules are also re-created in any scenario where the map is re-created?
- In any case, @amadio, @civanch, is there still interest in implementing external caching in SIM if this API is made available? 
Thanks,
  N.

VinInn:
``` since there may be cases when the geometry changes (multiple runs in same process).```
Can somebody seriously show an example of this working?
My understanding that with mechanism such as "era" we cannot run "multiple geometries" in the same process.

I would much prefer  to simply say "do not even think to change geometry in a process"

**Comment by namapane** [comment]
Ciao Vincenzo, I'm not sure if your question is intended to me - as that is in fact the same question I am asking, ie., are we guaranteed that there is no use case where the external cache mechanism drafted here https://github.com/cms-sw/cmssw/pull/28180#issuecomment-544264803 will produce a dangling pointer? 
If somebody more expert than me says there is none, I am happy and will go on with impementing the proposal. I just am not sure myself.

**Comment by namapane** [comment]
Anyhow let's take, as a purely tecnhical example, this test:
https://github.com/cms-sw/cmssw/blob/master/CondTools/RunInfo/test/test_runinfo_cfg.py
it creates a fictional job with three runs with three runs which different current values. In principle the map should be re-created for each run given that the current changes, isn't it? 
My concern is if clients (ESProducers) that may hold a cache are also re-created at the same time. Otherwise the cache will become invalid.
This may not be a realistic use case; I just want to figure out if this is something we have to bother about, or not, before implementing this caching mechanism.

**Comment by slava77** [comment]
>     * -I have a concern on the validity of the external cache. Are we guaranteed that there is no scenario where the map is re-created while users still keep the cache, which would then essentially become a dangling pointer? In principle the map can be re-created at IOV boundaries

I suppose the proposed caching class can keep an ESWatcher https://twiki.cern.ch/twiki/bin/view/CMSPublic/SWGuideHowToGetDataFromES#Seeing_that_a_Record_has_changed

**Comment by Dr15Jones** [comment]
> I suppose the proposed caching class can keep an ESWatcher

I'd actually suggest using the lower level value `EventSetupRecord::cacheIdentifier()`. Anytime an IOV changes for a `EventSetupRecord` the value returned by `cacheIdentifier()` changes. It is what `ESWatcher` uses internally.

**Comment by namapane** [comment]
Uhm, this is going to add some complexity and, I suppose, cost to this machinery. I'm not proficient in this kind of magic but I suppose it's one more object to percolate and one more check to be made (for each call?) in a place where @amadio has found that even removing some "if (debug)" was bringing an overall improvement given the overall number of calls.
I wonder if we should rather not reconsider other options, e.g. the initial suggestion of a simple configurable to choose between atomic and thread_local way of caching (cf. @slava77 's comment https://github.com/cms-sw/cmssw/pull/28180#discussion_r336228760). 
Thread_local could be set on for SIM only, where there's no odd case to use more than one instance of the MF geometry at a time. 
Reco/HLT, which is less MF-intensive and likely suffers less from atomic, could stay with the current  behavior. Not too bad IMHO, especially since I doubt developers will jump on implementing external caching for reco anyhow.
BTW this would have the nice advantage to have been already coded and profiled in detail by @amadio.
What do others think?

**Comment by namapane** [comment]
Uhm, this is going to add some complexity and, I suppose, cost to this machinery. I'm not proficient in this kind of magic but I suppose it's one more object to percolate and one more check to be made (for each call?) in a place where @amadio has found that even removing some "if (debug)" was bringing an overall improvement given the overall number of calls.
I wonder if we should rather not reconsider other options, e.g. the initial suggestion of a simple configurable to choose between atomic and thread_local way of caching (cf. @slava77 's comment https://github.com/cms-sw/cmssw/pull/28180#discussion_r336228760). 
Thread_local could be set on for SIM only, where there's no odd case to use more than one instance of the MF geometry at a time. 
Reco/HLT, which is less MF-intensive and likely suffers less from atomic, could stay with the current  behavior. Not too bad IMHO, especially since I doubt developers will jump on implementing external caching for reco anyhow.
BTW this would have the nice advantage to have been already coded and profiled in detail by @amadio.
What do others think?

**Comment by VinInn** [comment]
I cannot agree with @namapane . the caching of mag field should return the current field immediately.
This is what everybody else does: why not cms? btw what wrong with thread_local even in reco?
for sure more efficient that a trashing useless atomic

**Comment by slava77** [comment]
> The cacheIdentifier would not need to be checked each call. It only needs to be checked when the MagenticField is pulled from the EventSetup. In fact, I'd only create the cache on the stack and therefore no worry about changes from event to event.

right, we can leave it up to the user to figure out not to cache across eons.

**Comment by namapane** [comment]
So you mean to forget about the above suggestion for adding ESWatcher/cacheIdentifier()?
Regarding @VinInn's comment - I lack competence to tell what's wrong with thread_local; for sure if it is an option it is a very appealing one, because (a) it does perform as shown in great detail by @amadio; (b) it is very simple to code and maintain; (c) it has already been implemented. 
Just my 2 cents...

**Comment by Dr15Jones** [comment]
thread_local does not work with IOV transitions just like function local statics do not unless you are also do a test each call for a change.

**Comment by slava77** [comment]
On 4/21/20 10:11 AM, Nicola Amapane wrote:
> So you mean to forget about the above suggestion for adding
> ESWatcher/cacheIdentifier()?

right,
that's what I meant with the previous comment

**Comment by namapane** [comment]
> 
> 
> thread_local does not work with IOV transitions just like function local statics do not unless you are also do a test each call for a change.

fine, but you and Slava just explained that we don't care about IOV transitions because we just "leave it up to the user" to avoid messing up with cache validity. 
So what I conclude is the reverse of your argument: IOV transitions are not a problem for either external caching and thread_local.

**Comment by namapane** [comment]
(BTW, when the map is re-created, a local thread_local cache can be easily be re-initialized to null.So I do not agreee that a test for a change of IOV would be needed)

**Comment by Dr15Jones** [comment]
> So what I conclude is the reverse of your argument: IOV transitions are not a problem for either external caching and thread_local.

'Leaving it up to the user' is not exactly what we were implying. If the cache is made on the stack then it gets set the first time every time so no need to do a 'check'. If you use the `cacheIdentifier()` or `ESWatcher` then the user is using the tools we tell everyone to use if they are caching something from the EventSetup (which is exactly what the original code change was saying to do).

Using a `thread_local` makes it impossible to ever handle IOVs correctly without the local code constantly checking to see if something changed.

**Comment by Dr15Jones** [comment]
(BTW, when the map is re-created, a local thread_local cache can easily be re-initialized to null. So I do not agree that a test for a change of IOV would be needed)
How? That would work for a 'static' but you do not know when a thread changes from one IOV to another.

Plus I can't see how a thread_local would work with our ability to run concurrent IOVs (where one thread can jump between different IOVs)

**Comment by makortel** [comment]
> btw what's wrong with thread_local even in reco?

E.g. we've had problems by going over the TLS size, and apparently are still having occasional problems with ARM and POWER (already when loading shared objects). In that view it would be the best to minimize the use of `thread_local`.

**Comment by namapane** [comment]
> 
> 
> (BTW, when the map is re-created, a local thread_local cache can easily be re-initialized to null. So I do not agree that a test for a change of IOV would be needed)
> How? That would work for a 'static' but you do not know when a thread changes from one IOV to another.
> 
> Plus I can't see how a thread_local would work with our ability to run concurrent IOVs (where one thread can jump between different IOVs)

Frankly I do not understand what you mean. The problem which I brought up with IOV is for cases where the MF geometry needs to be re-created from scratch. Then the cache pointer points to something that has been deleted. If the cache is within my MF geometry, I can initialize it to null when I create the new geometry. There's no other possible issue with IOV boundaries if that does not involve deleting and re-creating the MF geometry from scratch. Or maybe I am missing what you are thinking of.
Also regarding concurrent IOV, I wonder how that would work in this case. Do you mean we can have a case where one thread may end up needing several different MFGeometries at the same time and jump from one to the other? I have a hard time figuring out how this can happen. It would be useful to see a working example.

**Comment by Dr15Jones** [comment]
>  I can initialize it to null when I create the new geometry.
How? Say you have 4 threads. On thread 1 you are creating the new geometry so you can get access to the thread_local for thread 1. But you do not have access to the thread_local for threads 2-4.

**Comment by namapane** [comment]
that means only a one-call cache miss for threads 2-4, who cares? It does not lead to an invalid cache and a crash.
Except for this concurrent IOV issue, I agree, but that I fail to understand...

**Comment by makortel** [comment]
> Also regarding concurrent IOV, I wonder how that would work in this case. Do you mean we can have a case where one thread may end up needing several different MFGeometries at the same time and jump from one to the other? I have a hard time figuring out how this can happen. It would be useful to see a working example.

We are deploying framework's "concurrent lumis" feature, which means that the framework can process events from multiple LuminosityBlocks concurrently (in practice 2 is likely sufficient). The events are still limited to the same IOV of all EventSetup records.

Concurrent IOVs allows events from different IOVs to be processed at the same time. When that happens, we have multiple copies of the relevant EventSetup products (one for each IOV in flight). Any shared global state between these products prevents this feature to be used for that EventSetup record.

**Comment by makortel** [comment]
> that means only a one-call cache miss for threads 2-4, who cares? It does not lead to an invalid cache and a crash.

This PR has
```cpp
static thread_local MagVolume const* lastVolume = nullptr;
MagVolume const* result = lastVolume;
if (result && result->inside(gp))
  return result;
```
If the `MagVolume` pointed by `lastVolume` has been deleted, the `result->inside()` call leads to crash (in the best case).

namapane:
because it was made static in the PR (not sure why), but it could well be a non-static thread-local. 
Just as the current cache, which is a non-static atomic. 
Then when the MF object which is shared between threads 1-4 is re-created, it is re-initialized in eg thread 1, and needs to be passed again to all other threads.
But maybe I am missing something.

makortel:
If `lastVolume` would not be function-`static`, it would be a normal local variable and get destructed at the end of scope (i.e. there would be no caching).

civanch:
Hi all,

I am a bit confused and may be misunderstand something. In my view, when MF is created it is using volume structures. However, when we instantiate field in OscarProducer we do not care about volumes. We instantiate make access to CMS MF in absolute coordinats. At each step of a charged particle in each thread Geant4 checks its MF cache and if a change of track point is more than 1 mm (it is a parameter, which may be optimized) MF is called, in the interface we provide global coordinates and get back field vector. We do not need extra cache inside MF service, all what we need is to remove atomic overhead. So, our main requirement: do not have this double cache which not speed-up but slow down.

civanch:
To say more, inside SIM we may extend our cache in order less frequently call MF service. Also Geant4 do not provide volume information. Potentially we can but I am not sure if this is needed, because we instantiate field as a global field, not field per volume. I saw in the code some tails of implementation of field per volume but it seems that this was an attempt long ago to do this but never used.

namapane:
@civanch, the underlying global MF service that you are calling uses its own volume structures internally, that are not exposed to the user. Each time SIM misses the 1 mm cache it needs to call the MF, which internally needs to make a volume search. That is a costly operation which is very often removed by the internal volume cache. MF volumes are typically of O(1m) dimensions, so by making 1 mm step you have a huge number of cache hits in the internal volume cache. if you are not convinced I can show how to disable the volume cache, you will easily find out how SIM timing degrades.

namapane:
> 
> 
> To say more, inside SIM we may extend our cache in order less frequently call MF service. Also Geant4 do not provide volume information. Potentially we can but I am not sure if this is needed, because we instantiate field as a global field, not field per volume. I saw in the code some tails of implementation of field per volume but it seems that this was an attempt long ago to do this but never used.

I think there is a misunderstanding here. MF volumes are a detail of the implementation of MF and are completely unrelated to G4 geometry. What you have seen is probably something within CMS SIM code, not the internal implementation of the MF.

namapane:
@makortel @Dr15Jones OK I think I now understand the issue of not being able to reset the cache in multiple threads at IOV changes. But still, that applies only to RECO, not to SIM where there's only one instance of MF per process, in any possible scenario. 
So the initial proposal of a flag to choose between thread-local and the current caching would still do the job for SIM in a simple way. RECO could go on with the current caching (which would be the case anyhow, because it is very unlikely that an external caching mechanism will be implemented in e.g. tracking).
This said, if you think this is not an acceptable solution and we must implement (optional) external caching, what I can do is to prepare a PR where I rearrange the current cache one level up in the call hierarchy as anticipated in https://github.com/cms-sw/cmssw/pull/28180#issuecomment-545431737, including the minor optimizations that @amadio put in this PR (edited according to @slava77 's comments), plus few others along the same lines. 
Then I think it would be more appropriate for a framework expert to take care of implementing the external caching mechanism, implementing it in SIM, and testing/profiling it, since I'm not sure I grasp all the details there.
Let me know what you think.

amadio:
I agree with @namapane about removing the cache. I tried it when working on the code proposed here, and saw that simulation times got much worse. That's when I decided to try the thread local cache. The thread local cache is based on the idea that it prevents threads integrating trajectories for different particles from invalidating the cache for each other. The atomic also creates extra work to keep caches coherent between threads, which is not a huge problem for low number of threads, but degrades performance as the number of threads increases. I don't know the details of how the code is used in the framework other than for simulation, but it might be possible to make the cache per-instance, to solve the problem in RECO, or add a method to set the cached volume to null, which could be called when the geometry gets updated. Although I agree that caching a level up would be better, if making that work is too difficult, invalidating the cache when the geometry changes might be an ok solution.

slava77:
> @makortel @Dr15Jones OK I think I now understand the issue of not being able to reset the cache in multiple threads at IOV changes. But still, that applies only to RECO, not to SIM where there's only one instance of MF per process, in any possible scenario.
> So the initial proposal of a flag to choose between thread-local and the current caching would still do the job for SIM in a simple way. RECO could go on with the current caching (which would be the case anyhow, because it is very unlikely that an external caching mechanism will be implemented in e.g. tracking).

IIRC, it was mentioned a few times that a thread_local is not an option due to limited availability of thread local storage.


> This said, if you think this is not an acceptable solution and we must implement (optional) external caching, what I can do is to prepare a PR where I rearrange the current cache one level up in the call hierarchy as anticipated in [#28180 (comment)](https://github.com/cms-sw/cmssw/pull/28180#issuecomment-545431737), including the minor optimizations that @amadio put in this PR (edited according to @slava77 's comments), plus few others along the same lines.
> Then I think it would be more appropriate for a framework expert to take care of implementing the external caching mechanism, implementing it in SIM, and testing/profiling it, since I'm not sure I grasp all the details there.
> Let me know what you think.

civanch:
Thanks for explantations. Now I understand, that external cache seems not natural. For the case of SIM, we have no information about internal structure of volumes for MF. It would be very strange if SIM handle external cache with something, which we do not understand and cannot test. Whatever cache should be fully handled by MF. Naively I would think, that this may be done via one thread_local variable. It should be not an object in TLS but one pointer to an object or struct. There is not need to put extended information inside TLS.

namapane:
> > 
> 
> IIRC, it was mentioned a few times that a thread_local is not an option due to limited availability of thread local storage.

I understand, but we are talking of just 8 bytes (one pointer) per thread: whatever the limit on TLS is, that should be a negligible addition. 
Are we really at a level where 8 bytes per thread make a difference in a job?

I do realize (eventually...) that resetting it to null across different threads at IOV boundaries would be a problem (@amadio: adding a method to set the cached volume to null as you suggested has the problem, I suppose, that we have no proper place to put the call so that it has effect in all relevant threads). But since MF in SIM never needs to be re-created during a job, this is not an issue in SIM, which is where this optimization matters, and we could activate it only there.

In any case, if you agree I can make a PR with the small refactoring+minor optimizations mentioned above, which can then be used as a basis to develop both a prototype of external caching or with configurable choice between thread_local and the current caching. They could then be evaluated for their performance and side effects to e.g. TLS limits.

Dr15Jones:
> I understand, but we are talking of just 8 bytes (one pointer) per thread: whatever the limit on TLS is, that should be a negligible addition.

My understanding is the limitation on TLS is not on the size of the data, the limitation is the total number of shared libraries with at least 1 TLS. I also believe we live near that boundary right now and we have no control of TLS on third-party libraries (e.g. Geant uses a huge number of them spread across many shared libraries).

makortel:
To me it looks like it should be fairly straightforward to pass a cache-to-MagneticField-wrapper from here
https://github.com/cms-sw/cmssw/blob/ee0a2618b506a8e0bcd204d0abfe0e1771268a5a/SimG4Core/Application/src/RunManagerMTWorker.cc#L290-L293
via
https://github.com/cms-sw/cmssw/blob/ee0a2618b506a8e0bcd204d0abfe0e1771268a5a/SimG4Core/MagneticField/src/FieldBuilder.cc#L25
to
https://github.com/cms-sw/cmssw/blob/ee0a2618b506a8e0bcd204d0abfe0e1771268a5a/SimG4Core/MagneticField/src/Field.cc#L25

civanch:
@Dr15Jones , you are write about uncontrolled use of TLS by Geant4, especially in first Geant4 MT release 10.0. However, in the new Geant4 10.6 TLS usage substantially reduced (it is reduced in each new release). So, an addition of one pointer in CMS MF is a negligible value compared with saving introduced in the recent Geant4.

@makortel , I agree, that we can put cache in place you mention but I cannot understand why it is needed : MF may have its own cache without Geant4 interface.

makortel:
> I agree, that we can put cache in place you mention but I cannot understand why it is needed : MF may have its own cache without Geant4 interface.

The current caching inside the MagneticField (`std::atomic` shared across threads) scales poorly (cache trashing), and a `thread_local` caching inside the MagneticField has other problems.

civanch:
@makortel , OK, please, propose modification of G4/MF interface we will use whatever cache object MF needed.

Dr15Jones:
The idea is add the ability to cache values to the `MagneticField` interface

Add the following to  MagneticField/Engine/interface/MagneticField.h

```C++
class MagneticFieldCache {
public:
    MagneticFieldCache(): cache_{nullptr} {}
   template<typename T> void set(T const* iCache) {cache_ = iCache;}
   template<typename T> T const* get() const {return reinterpret_cast<T const*>(iCache); }

   void reset() { cache_ = nullptr;}
private:
    void const* cache_;
};
```

Change the interface in MagneticField to be
```C++
float inTesla(const GlobalPoint& gp, MagneticFieldCache& cache) const;

 GlobalVector inKGauss(const GlobalPoint& gp, MagneticFieldCache& cache) const;

  /// Field value ad specified global point, in 1/Gev
  GlobalVector inInverseGeV(const GlobalPoint& gp, MagneticFieldCache& cache) const 

GlobalVector inTeslaUnchecked(const GlobalPoint& gp, MagneticFieldCache& cache) const;
```

Then create a new header file MagneticField/Engine/interface/localMagneticField.h
```C++
namespace local {
class MagneticField {
    public:
    MagneticField(): field_(nullptr), cache_(nullptr) {}
    MagneticField( ::MagneticField const* field) : field_(field), cache_(nullptr) {}

  /// Field value ad specified global point, in KGauss
  GlobalVector inKGauss(const GlobalPoint& gp) const { return inTesla(gp) * 10.F; }

  /// Field value ad specified global point, in 1/Gev
  GlobalVector inInverseGeV(const GlobalPoint& gp) const { return inTesla(gp) * 2.99792458e-3F; }

  /// True if the point is within the region where the concrete field
  // engine is defined.
  bool isDefined(const GlobalPoint& /*gp*/) const { field_->isDefined(); }

  /// Optional implementation that derived classes can implement to provide faster query
  /// by skipping the check to isDefined.
  GlobalVector inTeslaUnchecked(const GlobalPoint& gp) const {
    return field_->inTeslaUnched(gp);
  }

  /// The nominal field value for this map in kGauss
  int nominalValue() const {
    return field_->nominalValue();
  }

   float inTesla(  const GlobalPoint& gp) const {
       return field_->inTesla(gp, cache_);
    }
    
    void reset(::MagneticField const* iField) {
      field_ =iField;
      cache_ = nullptr;
    }
    private:
        ::MagneticField const* field_;
        MagneticFieldCache cache_;
};
}
```

Then replace the use of `::MagneticField` with `local::MagneticField`.

If you'd rather be less intrusive, just add a new interface to `::MagneticField`
```C++
float inTeslaFast(const GlobalPoint& gp, MagneticFieldCache& cache) const;
```
and then change `local::MagneticField` to call `inTelsaFast`. Then in places where MagneticField lookup is a bottleneck (like the simulation) use `local::MagneticField` rather than `::MagneticField`.

namapane:
@Dr15Jones, I have been assuming we all agree that external caching will be an optional alternative to the current caching, and that we do not want to remove internal caching all of a sudden from everywhere in CMSSW to leave it up to developers to add the external cache instead. 

From your draft code it seems that you are instead thinking to leave only the external cache as an option. I think this is not realistic. Implementing external caching everywhere in CMS will be a significant task, I don't see how this can happen in a realistic release timeline. We don't want to release an update that makes HLT slower all of a sudden, and tell developers they can fix it by adding external caching where relevant, for example.

Also given the way the MF usage is ubiquitous in RECO and HLT and fragmented e.g. in different tracking classes, I expect the required changes to be quite invasive. I am also not sure it will always be easy to percolate the cache (or the local::MF pointer) to all contexts which benefit of a common cache.

So I think we should not all of a sudden remove the current caching, just (at most) add one version with inTesla(const GlobalPoint& gp, MagneticFieldCache& cache) that uses the external cache, instead of the internal one that is used in inTesla(const GlobalPoint& gp). (We don't need to call it inTeslaFast, which would BTW be a misleading name)

makortel:
I agree trying to migrate all code using `MagneticField` in a single PR would not be feasible (*), but on the other hand leaving the "trashing caching" does not sound too future proof either. How about
* we add a new `inTesla()` overload with the cache, and leave the "trashing caching" on (as @namapane suggested above)
* we migrate SIM to use the cache (should be straightforward by initial inspection, I can volunteer to do that)
* we start a migration process for the rest. It will take some time, but the scope (*) does not look too bad (especially when comparing to various other migrations that we are doing or have done)
   * I'd think code following the recommended practices should be fairly straightforward to migrate from `MagneticField` to `local::MagneticField`, and the difficult ones might need to be re-worked eventually anyway for the EventSetup-consumes.
   * core team can help in the migration

?

(*) `git grep "ESHandle<MagneticField>"` shows 286 files in 130 packages

VinInn:
My proposal

```
namespace {
  std::atomic<int> instanceCounter(0);
  thread_local int localInstance=0;
  thread_local MagVolume const* lastVolume = nullptr;
}

MagGeometry::MagGeometry(...) : 
  me_(++instanceCounter(),....    // with "const int me_;"
{....}

MagVolume const* MagGeometry::findVolume(const GlobalPoint& gp, double tolerance) const {

if (me_!=localInstance) {
  localInstance=me_;
  lastVolume = nullptr;
}

```

namapane:
@markotel, I agree some migration could be done afterwards, but I don't think it makes sense to blindly migrate everything without some supporting profiling. 
@amadio has done a remarkabke profiling job for SIM and measured its was necessary, but mostly because of the fact that different SIM threads invalidate the cache for each other, than for the actual use of atomic. RECO has very different usage patterns, is much less intensive in the use of MF, and I am not sure that fragmenting the cache in 286 different different contexts will not reduce the advantage of the cache, even compared to the advantage of it being non-trashing. 
I personally would not embark in a large migration without even knowing how much "trashing" is the trashing... 

@VinInn if I understand correctly your proposal solves the initialization at IOV boundary with
the access to an atomic being only on construction (thus  irrelevant in terms of "trashing") and the cost being only 2x8 bytes in TLS and one more int==int check per call.
No migration needed, issue can be closed in a couple of hours.,, looks brillant to me :-)
I do understand that thread_local has "other problems" although I still did not grasp (even after some time googling) what they really are: total storage size? number of variables? number of libraries with at least 1 TLS? Are we far from the limit now that TLS in Geant4 10.6 has been substantially reduced as explained by @civanch? 
I may be a bit old-school but I tend to prefer a simple solution like this one unless I see some hard numbers on how 2x8 bytes in one more library with TLS are significantly worse than migrating 286 contexts in 130 packages.
Just my 2 cents...

civanch:
Hi all,

I think, @VinInn proposal is the best. If there are still questions, then 
may be it is possible to measure TLS usage for SIM and RECO of 2021 WFs?

makortel:
Vincenzo's proposal (https://github.com/cms-sw/cmssw/pull/28180#issuecomment-618174565) would indeed work with IOV transitions and multiple `VolumeBasedMagneticField` objects within a `cmsRun` job (e.g. concurrent IOVs).

Core would nevertheless prefer to avoid `thread_local`, but if reco and sim would like to proceed with Vincenzo's proposal we we can tolerate it as a one-shot workaround than may not become a general solution (and would prefer to have this documented in comments near the place). A very small number number of cases not following the rules can be tolerated, but a large number of such cases will lead to huge problems.

The arguments against are really about keeping our code as future proof as we can
* As mentioned before, we have had (and are having again in POWER and ARM) problems with the number of shared objects with thread-local storage
* We have experienced the situation where the stack and the thread-local storage collided, debugging that was a nightmare
* Therefore the use of thread-local storage does not really scale with the number of uses (practical upper limit will be reached relatively early, and with strange symptoms at best), whereas a local cache would scale orders of magnitude further with the number of uses

slava77:
> Core would nevertheless prefer to avoid `thread_local`, but if reco and sim would like to proceed

I still did not see a significant cost in coding of the implementation with external cache (assuming that the current atomic internal cache can remain concurrently).
So, considering the repeated concerns about using `tread_local`, it seems to me that the plan from October as outlined in #28284 is still valid.

namapane:
I still did not see a significant cost in adding 2x8 bytes TLS, so considered all other concerns related to the external cache, I call myself out.  :-)

I am now preparing a PR with the other minor improvements discussed above, eg the pre-computing of barrel boundaries proposed by @amadio (it gives a 0.4% overall speedup for non-cached MF calls in callgrind; I think I could add a few more of the same kind with very limited effort).
One question, in the discussion it was suggested to shield the if (debug) {...} couts with #ifdefs because even if they equal to if (false) they add a measurable overhead in SIM. Are there objections on just adding #ifdef MF_DEBUG and maybe change the enclosed couts to LogDebug?

BTW I realized moving the current atomic cache one level up in the call hierarchy as I initially suggested (see above) is maybe not a good idea since MagGeometry::findVolume is used directly by SteppingHelixPropagator. I'm not sure that really benefits from the cache. This should be checked before considering moving the cache up to MagGeometry::fieldInTesla.

VinInn:
Strictly speaking in cmssw we need just ONE global thread_local  (a small integer) that identifies in which framework stream one is. FW can set it just before invoking a module (together many other activities it does at that point...).
And this thread_local variable can be located in the main program and accessed through a simple function.
Then every body can build is preferred container indexed using such an identifier.

namapane:
So as anticipated I made #29556 with minor optimizations from this one (revisited) + couts cleanup discussed above.
I think it would be interesting to run some profiling of SIM with @Vinn's proposal while the discussion evolves. I can implement it in my cmssw repo on top of the above PR if anybody from SIM is willing to give it a try.

VinInn:
maybe we should move the discussion to an "issue"

civanch:
@namapane , whatever try you will propose as a PR we will try to test your PR on CPU efficiency promptly.

@VinInn , I am trying convince  my Geant4 collègues to remove maximally thread_local varriables but in fact this clean-up is going slowly. May be some quantitative analysis can help.

namapane:
@civanch , I pushed a branch in my clone:
https://github.com/namapane/cmssw/tree/MF_TLScache
Not sure it is appropriate to make a PR for it, given the ongoing discussion. 
But if that make it technically easier for you to pick it, and if there is no objection, I can do.

Some technicalities:
- It is made on top of #29556 (ie includes it)
- Passes the MF regression-tests
- I added a test that creates several different IOVs in the same job and verified that instance counters behave in the correct way, at least when IOV changes happen seqientially in a single thread. 

I don't know how to test IOV transition across multiple threads. I tried to set several events per run (ie repeating the regression test once for each event) so that different events in the same run and IOV can be processed in parallel. I think this is not happening, at least according to the output:

```
Begin processing the 1st record. Run 300001, Event 1, LumiSection 1 on stream 0 at 25-Apr-2020 11:40:17.085 CEST
Begin processing the 2nd record. Run 300001, Event 2, LumiSection 1 on stream 0 at 25-Apr-2020 11:40:24.426 CEST
Begin processing the 3rd record. Run 300001, Event 3, LumiSection 1 on stream 0 at 25-Apr-2020 11:40:25.546 CEST
Begin processing the 4th record. Run 300001, Event 4, LumiSection 1 on stream 0 at 25-Apr-2020 11:40:26.676 CEST
Begin processing the 5th record. Run 300001, Event 5, LumiSection 1 on stream 0 at 25-Apr-2020 11:40:27.793 CEST
Begin processing the 6th record. Run 300002, Event 1, LumiSection 2 on stream 0 at 25-Apr-2020 11:40:28.927 CEST
Begin processing the 7th record. Run 300002, Event 2, LumiSection 2 on stream 0 at 25-Apr-2020 11:40:35.809 CEST
Begin processing the 8th record. Run 300002, Event 3, LumiSection 2 on stream 0 at 25-Apr-2020 11:40:36.995 CEST
Begin processing the 9th record. Run 300002, Event 4, LumiSection 2 on stream 0 at 25-Apr-2020 11:40:38.185 CEST
...
```

I interpret "on stream 0"  to mean they are all in the same thread (right?)

Do/can I need to instruct the FW to run different events in parallel?

civanch:
@namapane , these days for me would be easier if this would be PR. Can I simply test #29556 ?

namapane:
No, #29556 does not contain the TLS cache, only some minor improvements.
I will make a PR so that you can test it.

slava77:
> Strictly speaking in cmssw we need just ONE global thread_local (a small integer) that identifies in which thread one is.
> This thread_local variable can be located in the main program and accessed through a simple function.
> it can be set by the framework before a module is called or just autonomously in the function itself (if (0==threadId) threadId = ++atomicThreadCounter;)
> Then every body can build is preferred container indexed using such an identifier. (if the FW provides a global function telling how many threads there are even better)

would ```std::this_thread::get_id()``` work?
If this doesn't cycle up to infinity due to task start/stop cycles,
then even a map (or a vector?) of caches will work.

namapane:
@civanch, done: it's #29561.

makortel:
@namapane 

> I interpret "on stream 0" to mean they are all in the same thread (right?)

Strictly speaking no. It means that there is one concurrent event in flight (or one concurrent "EDM stream" processing events). The default is to use one thread (which is likely what you observe), although technically multiple threads can be used also with a single stream.

> Do I need to/can I instruct the FW to run different events in parallel?

The number of threads to use can be set at command line with `cmsRun -n <NT> config.py`. Then the framework will use the same number of streams (concurrent events) as the number of threads. The two (threads and streams) can be also controlled separately in the configuration with
```python
process.options.numberOfThreads = <NT>
process.options.numberOfStreams = <NS> # default is 0, which means to use the number of threads
```

If you want to try out concurrent lumis, that can be enabled with
```python
process.options.numberOfConcurrentLuminosityBlocks = 2
```

makortel:
@Dr15Jones did a little test of crafting a virtual interface and an implementation that does a `thread_local` caching in a shared library, and compared that to a local caching (code [here](https://github.com/makortel/tls_test)). The test confirms the observation mentioned e.g. in [this blog post](http://david-grs.github.io/tls_performance_overhead_cost_linux/), that is in conjunction with shared libraries accessing `thread_local` is slower than accessing a regular variable. In Chris' test the `thread_local` access was 1.5-2x slower than accessing a "local cache" (in case all accesses are cache hits).

Design-wise a local cache suits better with expressing parallelism in terms of tasks than a per-thread cache (implemented in whatever way). For example, if a job uses less streams than threads and has algorithm(s) using e.g. `tbb::parallel_for`, some threads are necessarily running tasks from different events. With a per-thread cache, it can be that a thread processes a task from event 1 (starts likely with a cache miss), then goes to process a task from event 2 (starts likely with a cache miss), then goes to process a task from event 3 (starts likely with a cache miss) etc. With a local cache, the cache can be copied for each task from the code creating the task, so tasks using the same volume as the caller start their work with a cache hit.

Our memory needs typically grow on the number of streams (concurrent events) instead of number of threads. Even if today we use the same number of threads and streams, it may be that in the future we need ot (or can) use less streams than threads. In such a case per-thread objects would consume more memory (or resources in general) than per-stream objects.

In general our experience so far is that getting per-thread work/storage to work properly with tasks is tricky (Geant4 integration being an extreme example). Also, local work/storage is easier to reason about, and therefore likely easier to maintain in the long term, than per-thread work/storage.

From the discussion in https://github.com/cms-sw/cmssw/pull/29561#discussion_r415135136 it seems that we are looking for a general pattern rather than a one-off workaround (https://github.com/cms-sw/cmssw/pull/28180#issuecomment-618582522), so core advices against using a per-thread cache.

slava77:
> From the discussion in [#29561 (comment)](https://github.com/cms-sw/cmssw/pull/29561#discussion_r415135136) it seems that we are looking for a general pattern rather than a one-off workaround ([#28180 (comment)](https://github.com/cms-sw/cmssw/pull/28180#issuecomment-618582522)), so core advices against using a per-thread cache.

do I read this correctly as we are back to #28284,
namely, no to ```thread_local``` or other variants like array indexed by ```"int"(this_thread::get_id())```?

makortel:
> > From the discussion in [#29561 (comment)](https://github.com/cms-sw/cmssw/pull/29561#discussion_r415135136) it seems that we are looking for a general pattern rather than a one-off workaround ([#28180 (comment)](https://github.com/cms-sw/cmssw/pull/28180#issuecomment-618582522)), so core advices against using a per-thread cache.
> 
> do I read this correctly as we are back to #28284,
> namely, no to `thread_local` or other variants like array indexed by `"int"(this_thread::get_id())`?

Yes, the "local cache" is the solution preferred by core as the most future proof solution (in terms of both performance and maintenance).

namapane:
@markotel,
1. The overhead of TLS is obviously there; it has has been measured by @amadio to be 0.34% in a REAL WORLD test of SIM (first row of the table). Thus, negligible. 
In a callgrind test of non-cached MF calls, I see an overhead of about 2.3% (MF calls only). This is the worst-case scenario - a job doing MF calls only, cache-miss only-, and looks quite negligible as well.

2. Your argument of the cache miss at start of the event is, frankly, nonsense.
The start of the event is likely to be in the tracker region, where we use a parametrization, for which we don't even look at the cache. For whatever happens next, or for whatever other scenario that does not start from the tracker, it is extremely unlikely that two events will make their first MF call that is not in the same region into the same volume, be it in SIM, RECO or HLT. There's simply no realistic scenario where this can happen.
Besides, the real advantage of the cache is that it saves volume search dozens, of hundreds of times when stepping in a meters-long volume. Saving one more or one less search is pretty much irrelevant. 
On the other hand, as I already pointed out, fragmenting the cache into hundreds of contexts may  well result in more cache misses if some disconnected contexts work e.g. sequentially on the same objects. 

3. Regarding memory growing with the number of streams, are we speaking of an issue of 16 bytes per stream?? Or did I miss something?

4. Finally, I disagree on the fact that a very invasive design that has to be implemented in 286 contexts in 130 packages is easier to maintain. It will make a simple thing as calling the MF a very complicated matter for the average user.

namapane:
To elaborate: the advantage of the cache is essentially within a single "task" when that task has to call the MF a very large number of times in a limited region. I can't imagine any scenario where the first call of a task is not a cache miss; not only across events, but even for tasks repeades within the same event (e.g. tracking different particles in parallel).

makortel:
@namapane 
To repeat from https://github.com/cms-sw/cmssw/pull/28180#issuecomment-618582522: for a one-shot use in the MagneticField the `thread_local` can be tolerated, but for a general pattern we advice against it.

namapane:
Understood; but Slava's comment https://github.com/cms-sw/cmssw/pull/28180#issuecomment-620186033 was not about general patterns, on which we all agree; it was on this specific use case. In this context your answer reads as a suggestion of rejecting #29561.

makortel:
Sorry that I was unclear about the context. I've understood the discussion here such that a one-shot solution would be on the table. But then I got confused in https://github.com/cms-sw/cmssw/pull/29561#discussion_r415135136 by the discussion about "precedent" and what exactly that would imply. Aiming to not spread the discussion too much I thought this PR (which is very close of really being an issue now) would be better for comments concerning general use beyond the #29561 .

VinInn:
Fine all this.

Can the Core team make a concrete proposal for a code pattern that avoid percolation?
I think it is responsibility of the core team to provide such a solution w/o dumping the burden to the end users.

makortel:
I'm assuming the `MagneticField` is already percolated everywhere it is needed. The "local cache" proposal would then only change
```cpp
// from
void function(const MagneticField& mf);
// to
void function(const local::MagneticField& mf);
```
(although for intra-module concurrent constructs like `tbb::parallel_for` a copy of the `local::MagneticField` would need to be passed as it itself should not be accessed concurrently).

The code getting the `MagneticField` from the `EventSetup` would change
```cpp
// from
const MagneticField& mf = eventSetup.getData(magneticFieldToken_);
// to
local::MagneticField mf(&eventSetup.getData(magneticFieldToken_));
```

Core team can help in the migration.

VinInn:
>I'm assuming the MagneticField is already percolated everywhere it is needed. 
and it was a pain: at the limit of non scaling. (and there are still places where it is somehow locally cached).

makortel:
> (and there are still places where it is somehow locally cached).

But even there the `local::MagneticField` should work. The interface supports updating the cache so a helper keeping the `MagneticField const*` as a member can be updated in the same manner

```cpp
struct Helper {
  void update(EventSetup const& eventSetup) {
    mf = &eventSetup.getData(mfToken);
  }

  MagneticField const *mf;
};

// to
struct Helper {
  void update(EventSetup const& eventSetup) {
    mf.reset(&eventSetup.getData(mfToken));
  }

  local::MagneticField mf;
};
// and "->" to "." for mf access
```

slava77:
> I'm assuming the `MagneticField` is already percolated everywhere it is needed. The "local cache" proposal would then only change

please clarify if this proposal requires that every user of MF to change to it, or is it possible to be done gradually, while both interfaces are supported.
From the discussion last fall (at least) I was still thinking that both local and global caching can coexist.

makortel:
> > I'm assuming the `MagneticField` is already percolated everywhere it is needed. The "local cache" proposal would then only change
> 
> please clarify if this proposal requires that every user of MF to change to it, or is it possible to be done gradually, while both interfaces are supported.
> From the discussion last fall (at least) I was still thinking that both local and global caching can coexist.

Correct, the "local cache" and the current "shared `std::atomic` cache" can coexist and a migration can be done gradually.

smuzaffar:
thanks,
For one of the sim warnings, I do not know what should be the correct fix so I had opened a issue
https://github.com/cms-sw/cmssw/issues/29569https://github.com/cms-sw/cmssw/issues/29569

namapane:
> The code getting the `MagneticField` from the `EventSetup` would change
> 
> ```c++
> // from
> const MagneticField& mf = eventSetup.getData(magneticFieldToken_);
> // to
> local::MagneticField mf(&eventSetup.getData(magneticFieldToken_));
> ```
> 

Doesn't the caller have to own the cache? At least, in the original proposal (https://github.com/cms-sw/cmssw/pull/28180#issuecomment-617817229)
the local::MagneticField(::MagneticField const* field) seems to be cache-less.

makortel:
> > The code getting the `MagneticField` from the `EventSetup` would change
> > ```c++
> > // from
> > const MagneticField& mf = eventSetup.getData(magneticFieldToken_);
> > // to
> > local::MagneticField mf(&eventSetup.getData(magneticFieldToken_));
> > ```
> 
> Doesn't the caller have to own the cache? At least, in the original proposal ([#28180 (comment)](https://github.com/cms-sw/cmssw/pull/28180#issuecomment-617817229))
> the local::MagneticField(::MagneticField const* field) seems to be cache-less.

Yes, the caller (or `local::MagneticField` to be precise) owns the cache (`MagneticFieldCache`). In the example above it is alive until the scope ends. It can also be made a member of a helper class (or even EDModule, but I'd expect that to be unnecessary, in most cases at least).

makortel:
I implemented the "local cache" approach in #29631.

PR #28188, opened 2019-10-16
use python from env instead of /usr/bin/python

smuzaffar:
Remove the hard coded `/usr/bin/python` and pick up python from PATH. This will make sure that we use the python provided by cmssw.


smuzaffar:
please test

qliphy:
+1

fabiocos:
+1

@Martin-Grunewald @jfernan2 @andrius-k these are simple technical fixes that I will merge in next IB, please have a look

jfernan2:
+1

Martin-Grunewald:
+1

PR #28192, opened 2019-10-16
Run3-fwk03 Avoid crash in making payloads

bsunanda:
#### PR description:

Modify DDCoreToDDXMLOutput code to avoid crash in presence of HGCal geometry

#### PR validation:

Tested using a cfg file in DetectorDescription/OfflineDBLoader/test area

Nothing special

### DIFF
--- a/DetectorDescription/OfflineDBLoader/src/DDCoreToDDXMLOutput.cc
+++ b/DetectorDescription/OfflineDBLoader/src/DDCoreToDDXMLOutput.cc
@@ -271,9 +271,9 @@ void DDCoreToDDXMLOutput::solid(const DDSolid& solid, std::ostream& xos) {
       std::vector<double> zs = rs.zscaleVec();
 
       xos << "<ExtrudedPolygon name=\"" << rs.toString() << "\"";
-      for (unsigned int i : x)
+      for (unsigned int i = 0; i < x.size(); ++i)
         xos << " <XYPoint x=\"" << x[i] << "*mm\" y=\"" << y[i] << "*mm\"/>\n";
-      for (unsigned int k : z)
+      for (unsigned int k = 0; k < z.size(); ++k)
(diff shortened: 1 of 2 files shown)


bsunanda:
@cmsbuild Please test

DetectorDescription/OfflineDBLoader/src/DDCoreToDDXMLOutput.cc
```diff
@@ -271,9 +271,9 @@ void DDCoreToDDXMLOutput::solid(const DDSolid& solid, std::ostream& xos) {
       std::vector<double> zs = rs.zscaleVec();
 
       xos << "<ExtrudedPolygon name=\"" << rs.toString() << "\"";
-      for (unsigned int i : x)
+      for (unsigned int i = 0; i < x.size(); ++i)
```
fabiocos: @bsunanda sorry, why does HGCal geometry imply a difference between these two syntaxes?

bsunanda: HGCal geometry uses extruded polygon and with that in the scenario, the code used to crash in the first syntax and produces the correct xml file in the second.

cvuosalo: The old code was assigning `i = x[0]`, then `i = x[1]`, etc. It would use `i` as the index in `x[i]` in line 275. That reference would probably be way out of bounds and could have caused the crash.

cvuosalo:
+1

DetectorDescription/OfflineDBLoader/src/DDCoreToDDXMLOutput.cc
fabiocos: ok, sorry, a trivial confusion between array index and element. Printing i instead of x[i] the range loop would be ok, but you need the index also for the other variable

fabiocos:
+1

PR #28195, opened 2019-10-17
clean cms tag collector / cvs services

davidlange6:
#### PR description:

I had noticed that we still have cms tag collector scripts in CMSSW. I cleaned up those and decided to remove stuff that depends on CMSSW in CVS. All should be quite obsolete.


davidlange6:
please test

probably the tip of the iceberg of what can be deleted in this package...

smuzaffar:
+core

fabiocos:
+1

@davidlange6 thanks, I agree this is forgotten stuff, the ```addpkg``` equivalent for git is in out scripts since quite long time

PR #28200, opened 2019-10-17
make track quality requirement in GeneralTracksImporter and PFTrackAl…

jsalfeld:
…goTools configurable

#### PR description:

Track Quality requirements in the PFBlockProducer are hardcoded. This PR is intended add configurability.

#### PR validation:

I ran OnLine_HLT_GRun.py with the new input for different track quality requirements in "process.hltParticleFlowBlock" PFBlockProducer. Locally I did a print out to check if bools are set correctly and the string correctly passed.

### DIFF
--- a/RecoParticleFlow/PFProducer/plugins/importers/GeneralTracksImporter.cc
+++ b/RecoParticleFlow/PFProducer/plugins/importers/GeneralTracksImporter.cc
@@ -15,8 +15,9 @@ public:
       : BlockElementImporterBase(conf, sumes),
         src_(sumes.consumes<reco::PFRecTrackCollection>(conf.getParameter<edm::InputTag>("source"))),
         muons_(sumes.consumes<reco::MuonCollection>(conf.getParameter<edm::InputTag>("muonSrc"))),
-	trackQuality_(
-	  (conf.existsAs<std::string>("trackQuality")) ? conf.getParameter<std::string>("trackQuality") : trackQuality_ = "highPurity"),
+        trackQuality_((conf.existsAs<std::string>("trackQuality"))
+                          ? reco::TrackBase::qualityByName(conf.getParameter<std::string>("trackQuality"))
+                          : reco::TrackBase::qualityByName("highPurity")),
         DPtovPtCut_(conf.getParameter<std::vector<double> >("DPtOverPtCuts_byTrackAlgo")),
         NHitCut_(conf.getParameter<std::vector<unsigned> >("NHitCuts_byTrackAlgo")),
         useIterTracking_(conf.getParameter<bool>("useIterativeTracking")),
@@ -33,7 +34,7 @@ private:
 
   edm::EDGetTokenT<reco::PFRecTrackCollection> src_;
   edm::EDGetTokenT<reco::MuonCollection> muons_;
-  std::string trackQuality_;
+  const reco::TrackBase::TrackQuality trackQuality_;
   const std::vector<double> DPtovPtCut_;
   const std::vector<unsigned> NHitCut_;
   const bool useIterTracking_, cleanBadConvBrems_, debug_;
@@ -65,7 +66,8 @@ void GeneralTracksImporter::importToBlock(const edm::Event& e, BlockElementImpor
         if (trkel->trackType(reco::PFBlockElement::T_FROM_GAMMACONV) && cRef.empty() && dvRef.isNull() &&
             v0Ref.isNull()) {
           // if the Pt resolution is bad we kill this element
-          if (!PFTrackAlgoTools::goodPtResolution(trkel->trackRef(), DPtovPtCut_, NHitCut_, useIterTracking_, debug_, trackQuality_)) {
+          if (!PFTrackAlgoTools::goodPtResolution(
+                  trkel->trackRef(), DPtovPtCut_, NHitCut_, useIterTracking_, trackQuality_, debug_)) {
             itr = elems.erase(itr);
             continue;
           }
@@ -117,7 +119,8 @@ void GeneralTracksImporter::importToBlock(const edm::Event& e, BlockElementImpor
                               (pfmu_->hasValidTrack(muonref, false) && PFMuonAlgo::isMuon(muonref)));
     }
     if (thisIsAPotentialMuon ||
-        PFTrackAlgoTools::goodPtResolution(pftrackref->trackRef(), DPtovPtCut_, NHitCut_, useIterTracking_, debug_, trackQuality_)) {
+        PFTrackAlgoTools::goodPtResolution(
+            pftrackref->trackRef(), DPtovPtCut_, NHitCut_, useIterTracking_, trackQuality_, debug_)) {
       trkElem = new reco::PFBlockElementTrack(pftrackref);
       if (thisIsAPotentialMuon && debug_) {
         std::cout << "Potential Muon P " << pftrackref->trackRef()->p() << " pt " << pftrackref->trackRef()->p()
--- a/RecoParticleFlow/PFProducer/plugins/importers/GeneralTracksImporter.cc
+++ b/RecoParticleFlow/PFProducer/plugins/importers/GeneralTracksImporter.cc
@@ -1,3 +1,4 @@
+#include "FWCore/MessageLogger/interface/MessageLogger.h"
 #include "RecoParticleFlow/PFProducer/interface/BlockElementImporterBase.h"
 #include "DataFormats/ParticleFlowReco/interface/PFBlockElementTrack.h"
 #include "DataFormats/ParticleFlowReco/interface/PFRecTrackFwd.h"
@@ -22,8 +23,7 @@ public:
         NHitCut_(conf.getParameter<std::vector<unsigned> >("NHitCuts_byTrackAlgo")),
         useIterTracking_(conf.getParameter<bool>("useIterativeTracking")),
         cleanBadConvBrems_(
-            conf.existsAs<bool>("cleanBadConvertedBrems") ? conf.getParameter<bool>("cleanBadConvertedBrems") : false),
-        debug_(conf.getUntrackedParameter<bool>("debug", false)) {
+            conf.existsAs<bool>("cleanBadConvertedBrems") ? conf.getParameter<bool>("cleanBadConvertedBrems") : false) {
     pfmu_ = std::unique_ptr<PFMuonAlgo>(new PFMuonAlgo(conf));
   }
 
@@ -37,7 +37,7 @@ private:
   const reco::TrackBase::TrackQuality trackQuality_;
   const std::vector<double> DPtovPtCut_;
   const std::vector<unsigned> NHitCut_;
-  const bool useIterTracking_, cleanBadConvBrems_, debug_;
+  const bool useIterTracking_, cleanBadConvBrems_;
   std::unique_ptr<PFMuonAlgo> pfmu_;
 };
 
@@ -67,7 +67,7 @@ void GeneralTracksImporter::importToBlock(const edm::Event& e, BlockElementImpor
             v0Ref.isNull()) {
           // if the Pt resolution is bad we kill this element
           if (!PFTrackAlgoTools::goodPtResolution(
-                  trkel->trackRef(), DPtovPtCut_, NHitCut_, useIterTracking_, trackQuality_, debug_)) {
+                  trkel->trackRef(), DPtovPtCut_, NHitCut_, useIterTracking_, trackQuality_)) {
             itr = elems.erase(itr);
             continue;
           }
@@ -118,13 +118,12 @@ void GeneralTracksImporter::importToBlock(const edm::Event& e, BlockElementImpor
       thisIsAPotentialMuon = ((pfmu_->hasValidTrack(muonref, true) && PFMuonAlgo::isLooseMuon(muonref)) ||
                               (pfmu_->hasValidTrack(muonref, false) && PFMuonAlgo::isMuon(muonref)));
     }
-    if (thisIsAPotentialMuon ||
-        PFTrackAlgoTools::goodPtResolution(
-            pftrackref->trackRef(), DPtovPtCut_, NHitCut_, useIterTracking_, trackQuality_, debug_)) {
+    if (thisIsAPotentialMuon || PFTrackAlgoTools::goodPtResolution(
+                                    pftrackref->trackRef(), DPtovPtCut_, NHitCut_, useIterTracking_, trackQuality_)) {
       trkElem = new reco::PFBlockElementTrack(pftrackref);
-      if (thisIsAPotentialMuon && debug_) {
-        std::cout << "Potential Muon P " << pftrackref->trackRef()->p() << " pt " << pftrackref->trackRef()->p()
-                  << std::endl;
+      if (thisIsAPotentialMuon) {
+        LogDebug("GeneralTracksImporter")
+            << "Potential Muon P " << pftrackref->trackRef()->p() << " pt " << pftrackref->trackRef()->p() << std::endl;
       }
       if (muId != -1)
         trkElem->setMuonRef(muonref);
--- a/RecoParticleFlow/PFProducer/plugins/importers/GeneralTracksImporter.cc
+++ b/RecoParticleFlow/PFProducer/plugins/importers/GeneralTracksImporter.cc
@@ -17,7 +17,7 @@ public:
         muons_(sumes.consumes<reco::MuonCollection>(conf.getParameter<edm::InputTag>("muonSrc"))),
         trackQuality_((conf.existsAs<std::string>("trackQuality"))
                           ? reco::TrackBase::qualityByName(conf.getParameter<std::string>("trackQuality"))
-                          : reco::TrackBase::qualityByName("highPurity")),
+                          : reco::TrackBase::highPurity),
         DPtovPtCut_(conf.getParameter<std::vector<double> >("DPtOverPtCuts_byTrackAlgo")),
         NHitCut_(conf.getParameter<std::vector<unsigned> >("NHitCuts_byTrackAlgo")),
         useIterTracking_(conf.getParameter<bool>("useIterativeTracking")),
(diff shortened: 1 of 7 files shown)


RecoParticleFlow/PFProducer/plugins/importers/GeneralTracksImporter.cc
```diff
@@ -31,10 +33,10 @@ class GeneralTracksImporter : public BlockElementImporterBase {
 
   edm::EDGetTokenT<reco::PFRecTrackCollection> src_;
   edm::EDGetTokenT<reco::MuonCollection> muons_;
+  std::string trackQuality_;
```
fwyzard: could you make this `const` ?

RecoParticleFlow/PFTracking/interface/PFTrackAlgoTools.h
```diff
@@ -18,5 +18,6 @@ namespace PFTrackAlgoTools {
                         const std::vector<double>& DPtovPtCut,
                         const std::vector<unsigned>& NHitCut,
                         bool useIterTracking,
-                        bool debug = false);
+                        bool debug = false,
+			std::string trackQuality = "highPurity");
```
fwyzard: I would suggest to swap the order, and have `trackQuality` before `debug`.

RecoParticleFlow/PFProducer/plugins/importers/GeneralTracksImporter.cc
```diff
@@ -15,6 +15,8 @@ class GeneralTracksImporter : public BlockElementImporterBase {
       : BlockElementImporterBase(conf, sumes),
         src_(sumes.consumes<reco::PFRecTrackCollection>(conf.getParameter<edm::InputTag>("source"))),
         muons_(sumes.consumes<reco::MuonCollection>(conf.getParameter<edm::InputTag>("muonSrc"))),
+	trackQuality_(
+	  (conf.existsAs<std::string>("trackQuality")) ? conf.getParameter<std::string>("trackQuality") : trackQuality_ = "highPurity"),
```
makortel:
```suggestion
	  (conf.existsAs<std::string>("trackQuality")) ? conf.getParameter<std::string>("trackQuality") : "highPurity"),
```

RecoParticleFlow/PFTracking/interface/PFTrackAlgoTools.h
makortel: Also
```suggestion
			const std::string& trackQuality = "highPurity");
```

RecoParticleFlow/PFTracking/src/PFTrackAlgoTools.cc
```diff
@@ -237,9 +237,11 @@ namespace PFTrackAlgoTools {
...
                         const std::vector<unsigned>& NHitCut,
                         bool useIterTracking,
-                        bool debug) {
-    //recheck that the track is high purity!
-    if (!trackref->quality(reco::TrackBase::highPurity))
+                        bool debug,
+			std::string trackQuality) {
+    
+    //check quality of tracks
+    if(!trackref->quality(reco::TrackBase::qualityByName(trackQuality)))
```
makortel: Actually, it would be better to do the string-to-enum conversion (`reco::TrackBase::qualityByName()`) once in the constructor (and possibly throw an exception if the returned value is `undefQuality`), and pass the enum to this function.

RecoParticleFlow/PFTracking/interface/PFTrackAlgoTools.h
jsalfeld: The change of order would complicate backward compatibility and introduce changes to where this function ("goodPtResolution") is used without trackQuality variable as input explicitly.

Let me know if this is ok (in principle I agree that having "debug" as last variable).

As far as I can tell, the affected classes are these ones:
RecoParticleFlow/PFTracking/plugins/HGCalTrackCollectionProducer.cc
RecoParticleFlow/PFProducer/plugins/importers/GeneralTracksImporterWithVeto.cc

hatakeyamak:
Perhaps this is beyond the scope of this PR, but it may be nice if we can move these config parameter setting for PFBlockProducer to fillDescription-based, as I am doing for PFProducer on https://github.com/cms-sw/cmssw/pull/28110
(https://twiki.cern.ch/twiki/bin/view/CMSPublic/SWGuideConfigurationValidationAndHelp)
which allow us to validate input parameters. Just wanted to mention this to see if this is something RECO conveners would prefer and Jakob would be willing to work on.

RecoParticleFlow/PFTracking/interface/PFTrackAlgoTools.h
jsalfeld: @makortel is your suggested change ready to go into the commit or needs testing?
@fwyzard let me know if the other classes should be changed to accommodate the suggested change of order

RecoParticleFlow/PFTracking/src/PFTrackAlgoTools.cc
makortel: Changing the `trackQuality` parameter here and `GeneralTracksImporter::trackQuality_` to `TrackBase::TrackQuality` enum? Yes, please do (and of course it needs to be eventually tested).

slava77:
@cmsbuild please test

RecoParticleFlow/PFProducer/plugins/importers/GeneralTracksImporter.cc
```diff
@@ -15,6 +15,9 @@ class GeneralTracksImporter : public BlockElementImporterBase {
       : BlockElementImporterBase(conf, sumes),
         src_(sumes.consumes<reco::PFRecTrackCollection>(conf.getParameter<edm::InputTag>("source"))),
         muons_(sumes.consumes<reco::MuonCollection>(conf.getParameter<edm::InputTag>("muonSrc"))),
+        trackQuality_((conf.existsAs<std::string>("trackQuality"))
+                          ? reco::TrackBase::qualityByName(conf.getParameter<std::string>("trackQuality"))
+                          : reco::TrackBase::qualityByName("highPurity")),
```
makortel: Here and elsewhere, this could be just
```suggestion
                          : reco::TrackBase::highPurity),
```

RecoParticleFlow/PFTracking/src/PFTrackAlgoTools.cc
```diff
@@ -237,9 +237,10 @@ namespace PFTrackAlgoTools {
                         const std::vector<double>& DPtovPtCut,
                         const std::vector<unsigned>& NHitCut,
                         bool useIterTracking,
+                        reco::TrackBase::TrackQuality trackQuality,
                         bool debug) {
```
perrotta: I notice that the ```debug``` option is not needed any more by this method. It was introduced by commit https://github.com/cms-sw/cmssw/commit/5eafe51907989c8245f37f5fe46a7d8ce9a0c135#diff-05a1b01a408107e62fa1938501541abd  , but then made immediately after useless when moving all ```cout``` comment lines to ```LogDebug``` with https://github.com/cms-sw/cmssw/commit/4c64ac90711ec6a8eb9ac9e8c4a763035219887d#diff-05a1b01a408107e62fa1938501541abd

You can profit of this PR to remove that method parameter

RecoParticleFlow/PFProducer/plugins/importers/GeneralTracksImporterWithVeto.cc
```diff
@@ -123,7 +128,8 @@ void GeneralTracksImporterWithVeto::importToBlock(const edm::Event& e,
                               (pfmu_->hasValidTrack(muonref, false) && PFMuonAlgo::isMuon(muonref)));
     }
     if (thisIsAPotentialMuon ||
-        PFTrackAlgoTools::goodPtResolution(pftrackref->trackRef(), DPtovPtCut_, NHitCut_, useIterTracking_, debug_)) {
+        PFTrackAlgoTools::goodPtResolution(
+            pftrackref->trackRef(), DPtovPtCut_, NHitCut_, useIterTracking_, trackQuality_, debug_)) {
       trkElem = new reco::PFBlockElementTrack(pftrackref);
       if (thisIsAPotentialMuon && debug_) {
         std::cout << "Potential Muon P " << pftrackref->trackRef()->p() << " pt " << pftrackref->trackRef()->p()
```
perrotta: Also in this class the ```debug_``` parameter is not needed: just replace this line with a corresponding ```LogDebug``` (the same as already done in ```PFTrackAlgoTools.cc```) and remove all references to the ```debug_``` parameter in the code.

RecoParticleFlow/PFProducer/plugins/importers/GeneralTracksImporter.cc
```diff
@@ -15,6 +15,9 @@ class GeneralTracksImporter : public BlockElementImporterBase {
       : BlockElementImporterBase(conf, sumes),
         src_(sumes.consumes<reco::PFRecTrackCollection>(conf.getParameter<edm::InputTag>("source"))),
         muons_(sumes.consumes<reco::MuonCollection>(conf.getParameter<edm::InputTag>("muonSrc"))),
+        trackQuality_((conf.existsAs<std::string>("trackQuality"))
```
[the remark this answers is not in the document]
jsalfeld: @perrotta, @hatakeyamak, conf.existsAs is used throughout the PF code, may I suggest to move this proposal to another future PR?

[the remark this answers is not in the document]
hatakeyamak: That's ok for me.

RecoParticleFlow/PFTracking/interface/PFTrackAlgoTools.h
```diff
@@ -18,5 +18,6 @@ namespace PFTrackAlgoTools {
                         const std::vector<double>& DPtovPtCut,
                         const std::vector<unsigned>& NHitCut,
                         bool useIterTracking,
+                        const reco::TrackBase::TrackQuality trackQuality = reco::TrackBase::highPurity,
```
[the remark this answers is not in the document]
jsalfeld: I checked, that various inputs in the configuration are correctly converted into the proper enums.
I can remove the default value, since I included the default value in the input of the two other plugins making use of this function. Hence for the 3 plugins I adjusted, there should be no difference wrt the previous version.

RecoParticleFlow/PFTracking/plugins/HGCalTrackCollectionProducer.cc
```diff
@@ -53,6 +53,7 @@ class HGCalTrackCollectionProducer : public edm::stream::EDProducer<> {
   // variables needed for copied goodPtResolution function
   // need to go back and figure out sensible values
   bool debug_;
```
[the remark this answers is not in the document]
jsalfeld: @perotta done in commit below

RecoParticleFlow/PFTracking/src/PFTrackAlgoTools.cc
jsalfeld: @perotta done in commit below

perrotta:
please test

perrotta:
@jsalfeld shouldn't you update also some python config with the tunable "trackQuality" parameter?
Of course, for now that parameter will remain set to the default highPurity threshold, but it will be possible to configure it in some forthcoming update. Moreover, if the python config gets updated since now it will be picked in the next parsing of the HLT menu, and become available in there.

jsalfeld:
@fwyzard @Sam-Harper @gennai would you have a comment on @perrotta comment above?

hatakeyamak:
If I understand correctly, this new config parameter can be added to:
https://github.com/cms-sw/cmssw/blob/master/RecoParticleFlow/PFProducer/python/particleFlowBlock_cfi.py#L41-L50
but when it's not there, the default value is used, and I thought that's the behavior we want for now.

perrotta:
> If I understand correctly, this new config parameter can be added to:
> https://github.com/cms-sw/cmssw/blob/master/RecoParticleFlow/PFProducer/python/particleFlowBlock_cfi.py#L41-L50
> but when it's not there, the default value is used, and I thought that's the behavior we want for now. So, PR is ok as it is on this point?

Yes, at the bare minimum I would add (already in this PR) the new parameter, with its default value set, in both 
- RecoParticleFlow/PFProducer/python/particleFlowBlock_cfi.py
- RecoParticleFlow/PFTracking/python/hgcalTrackCollection_cfi.py

Moreover, if I am not wrong, the ```maxDPtOPt``` parameter in the blob that you pointed out is redundant (say useless), as it is correctly already defined in RecoParticleFlow/​PFProducer/​python/​particleFlow_cfi.py: if so, please remove it.

Then, the HGCalTrackCollectionProducer is simple enough that a ```fillDescriptions()``` method could be even added in this PR, thus being able to get rid of the ```existsAs()``` checks in there. I don't think this would be a complicate addition here, but I'm fine if you postpone it to a following PR.

Finally, the real completion of the work would be to implement ```fillDescriptions()``` in the PFBlockProducer as a whole. I agree that this is a more complex and delicate task that could be better postponed in a future and dedicated PR.

@Martin-Grunewald I don't understand how the parsing of the parameters of the PFBlockProducer is done. Shouldn't they be taken from the RecoParticleFlow/PFProducer/python/particleFlowBlock_cfi.py?
If I look at the parameters in the HLT menu, I see that both the ```maxDPtOPt``` (and this is in principle ok, see above) and the ```cleanBadConvertedBrems``` (which should be set at its defalult "false" value, instead) are missing. So: where is the configuration for the HLT menu parsed from?

Sam-Harper:
> @fwyzard @Sam-Harper @gennai would you have a comment on @perrotta comment above?

If we wish to use any functionality in the HLT of the new features, my understanding is that we need a fillDescriptions() or at least the parameter set in the modules python file. Otherwise we cant access it and it would likely cause us problems. I understand the reluctance of the fillDescptions but could you not just at it to the python config? The one exception to this is if the parameter is part of a PSet although even in this case , it would be nice.

jsalfeld:
I added the "trackQuality" parameter with default parameter to the files:
RecoParticleFlow/PFProducer/python/particleFlowBlock_cfi.py
RecoParticleFlow/PFTracking/python/hgcalTrackCollection_cfi.py
From the latter I removed the redundant "debug" parameter, as was removed from the Producer in the previous commit.

perrotta:
> I added the "trackQuality" parameter with default parameter to the files:
> RecoParticleFlow/PFProducer/python/particleFlowBlock_cfi.py
> RecoParticleFlow/PFTracking/python/hgcalTrackCollection_cfi.py
> From the latter I removed the redundant "debug" parameter, as was removed from the Producer in the previous commit.

Thank you @jsalfeld !
Please evaluate also the possibility of removing the ```maxDPtOPt``` parameter from the configuration of the particleFlowBlock prooducer. After that, all other requests that were listed in my previous [comment](https://github.com/cms-sw/cmssw/pull/28200#issuecomment-546675536) could even be postponed to some following PR

jsalfeld:
@perrotta, thanks a lot.

Let me know if I understand correctly: For maxDPtOPt:
> Please evaluate also the possibility of removing the maxDPtOPt parameter from the configuration of the particleFlowBlock prooducer.

My understanding is that the module python:

> RecoParticleFlow/​PFProducer/​python/​particleFlow_cfi.py

defines input for the PFProducer RecoParticleFlow/PFProducer/plugins/PFProducer.cc. Is it correct? As such they seem to be meant as being different. After digesting, I see that the PFProducer passes the parameters to PFAlgo.cc 
https://github.com/cms-sw/cmssw/blob/master/RecoParticleFlow/PFProducer/plugins/PFProducer.cc#L120

which in turn passes them to PFMuonAlgo.cc

https://github.com/cms-sw/cmssw/blob/master/RecoParticleFlow/PFProducer/src/PFAlgo.cc#L26

Only here (for muons) the maxDPtOPt parameter is used. The same is true for other parameters like "trackQuality" and "minPixelHits"...

perrotta:
Indeed that's also my impression, Because of that, I guess that

perrotta:
Indeed that's also my impression. Because of it, I guess that

https://github.com/cms-sw/cmssw/blob/master/RecoParticleFlow/PFProducer/python/particleFlow_cfi.py#L122

is the correct place where to define it, while in

https://github.com/cms-sw/cmssw/blob/master/RecoParticleFlow/PFProducer/python/particleFlowBlock_cfi.py#L46

it is redundant and can be removed (just to clean up the config)

jsalfeld:
@perrotta , but I think
https://github.com/cms-sw/cmssw/blob/master/RecoParticleFlow/PFProducer/python/particleFlowBlock_cfi.py#L46

Is being used here (via GeneralTracksImporter and ultimately PFTrackAlgoTools.cc):
https://github.com/cms-sw/cmssw/blob/master/RecoParticleFlow/PFTracking/src/PFTrackAlgoTools.cc#L259

Whereas 
https://github.com/cms-sw/cmssw/blob/master/RecoParticleFlow/PFProducer/python/particleFlow_cfi.py#L122

is being used here (via PFProducer.cc and ultimately PFMuonAlgo.cc):
https://github.com/cms-sw/cmssw/blob/master/RecoParticleFlow/PFProducer/src/PFMuonAlgo.cc#L492

Isn't it?

perrotta:
Yes, but the DPtovPtCut_ in GeneralTracksImporter is taken from the configuration parameter "DPtOverPtCuts_byTrackAlgo", see https://github.com/cms-sw/cmssw/blob/master/RecoParticleFlow/PFProducer/plugins/importers/GeneralTracksImporter.cc#L18.

There is no "maxDPtOPt" parameter ever read out from the config in GeneralTracksImporter

jsalfeld:
yes indeed, will do asap

perrotta:
please test

hatakeyamak:
> yes indeed, will do asap. I assume these things will be fixed once the fillDescription is used and the cfi`s are produced automatically.

Thanks for sorting this out.

hatakeyamak:
Comparisons look good to me.

perrotta:
+1
- The trackQuality to compare with is now made configurable in the mentioned classes
- Some cleaning of the code and python configurations is also implemented

fabiocos:
+1

Martin-Grunewald:
@perrotta 
Just to confirm: PFBlockProducer is currently parsed from ```RecoParticleFlow/PFProducer/python/particleFlowBlock_cfi.py```

PR #28205, opened 2019-10-17
Run3-hcx232 Attempt to remove access of DDCompactView from SimG4CMS/Calo code (Step 2)

bsunanda:
#### PR description:

Attempt to remove access of DDCompactView from SimG4CMS/Calo code (Step 2)

#### PR validation:

Use standard runTheMatrix.py code

Nothing special

### DIFF
--- a/Geometry/HcalCommonData/python/caloSimulationParameters_cff.py
+++ b/Geometry/HcalCommonData/python/caloSimulationParameters_cff.py
@@ -1,0 +1,9 @@
+import FWCore.ParameterSet.Config as cms
+
+from Geometry.HcalCommonData.caloSimulationParameters_cfi import *
+
+from Configuration.Eras.Modifier_dd4hep_cff import dd4hep
+
+dd4hep.toModify(caloSimulationParameters,
+                fromDD4Hep = cms.bool(True)
+)
--- a/Configuration/Eras/python/Modifier_dd4hep_cff.py
+++ b/Configuration/Eras/python/Modifier_dd4hep_cff.py
@@ -1,0 +1,4 @@
+import FWCore.ParameterSet.Config as cms
+
+dd4hep =  cms.Modifier()
+
--- a/Configuration/Geometry/python/GeometryExtended2015CastorMeasured_cff.py
+++ b/Configuration/Geometry/python/GeometryExtended2015CastorMeasured_cff.py
@@ -6,7 +6,5 @@ import FWCore.ParameterSet.Config as cms
 # Ideal geometry, needed for simulation
 from Geometry.CMSCommonData.cmsExtendedGeometry2015CastorMeasuredXML_cfi import *
 from Geometry.TrackerNumberingBuilder.trackerNumberingGeometry_cfi import *
-from Geometry.HcalCommonData.hcalParameters_cfi import *
-from Geometry.HcalCommonData.hcalSimulationParameters_cfi import *
-from Geometry.HcalCommonData.hcalDDDSimConstants_cfi import *
+from Geometry.HcalCommonData.hcalDDDSimConstants_cff import *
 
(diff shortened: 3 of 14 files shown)


bsunanda:
@cmsbuild Please test

Geometry/HcalCommonData/python/caloSimulationParameters_cff.py
```diff
@@ -0,0 +1,9 @@
+import FWCore.ParameterSet.Config as cms
+
+from Geometry.HcalCommonData.caloSimulationParameters_cfi import *
+
+from Configuration.Eras.Modifier_dd4hep_cff import dd4hep
+
+dd4hep.toModify(caloSimulationParameters,
```
fabiocos: @bsunanda where is this modifier supposed to be used?

fabiocos:
@bsunanda do you understand the reason of the differences in the DQM comparison?

Geometry/HcalCommonData/python/caloSimulationParameters_cff.py
bsunanda: This is an advance statement. Hope we shall use this once we have a full dd4hep application.

bsunanda:
No - I don't. The affected code only decides creation of SimTracks in calorimeter region and even there it should be the same as before which I checked

cvuosalo:
+1

civanch:
@bsunanda , this PR is fine to me but there is a question: do we really need a collection of logical volumes per calorimeter to be stored in SD classes? Can we avoid using these vectors?

civanch:
+1
let us remove DDCompactView first, later we may return to navigation between Geant4 objects

kpedro88:
@bsunanda we need to understand the comparison changes before this PR can be approved

bsunanda:
I understand the differences are in Phase2 workflows. I do not understand the variables where the differences occur.

bsunanda:
@kpedro88 Any idea of the variables which show the difference?

kpedro88:
I don't see a clear indicator in the changed plots. Can you just compare the values of any modified member variables in `CaloTrkProcessing`?

bsunanda:
I printed out all the variables in the CaloTrkProcessing - there is no difference.

kpedro88:
Huh, then I'm stumped. Let's just try retesting it again in case there was some transient issue.

kpedro88:
please test

kpedro88:
the changes in phase 2 workflows are consistent between this test and the previous test, so it seems to be a real effect.

bsunanda:
@kpedro88 Could you let me know which workflows see the differences

kpedro88:
@bsunanda you can look here: https://cmssdt.cern.ch/SDT/jenkins-artifacts/baseLineComparisons/CMSSW_11_0_X_2019-10-23-1100+0d63f3/33930/validateJR.html

The phase 2 workflows with apparently real differences are:
20034.0
20434.0
22034.0

bsunanda:
@kpedro88 I checked the access of the constants and initialization of CaloTrkProcessing in the old and new scenario. They are identical. So I doubt the code causes the differences observed. This code decides some extra SimTracks to be saved (tracks entering calorimeter).

kpedro88:
@bsunanda a change in the SimTracks saved could potentially impact the HGCal cheated reco, if it changes the SimClusters...

bsunanda:
@kpedro88 I understand that. But I cannot find a cause for this. The logic is unchanged and the input parameters are the same. Can it be simply statistics issue?

kpedro88:
I looked through the code again, very carefully, and I didn't notice anything inconsistent or out of place. There is a slim possibility that the organization of stack memory was slightly altered, just enough to shift a few uninitialized values somewhere else... but this is very difficult to track down, and also wouldn't really be the fault of this PR. So I'm going to approve it.

kpedro88:
+upgrade

fabiocos:
+operations

the addition of a new modifier is in preparation for future use, it looks meaningful for the intermediate migration period

fabiocos:
+1

@bsunanda @kpedro88 I understand there is no apparent simple explanation for the observed differences, I think that a careful check on larger statistics will be needed

PR #28133, opened 2019-10-09
Cleanup obsolete part from Egamma PixelHitMatcher class 

guitargeek:
#### PR description:

Hello,

in a follow up to https://github.com/cms-sw/cmssw/pull/28039 I quickly had a look into how to reduce the redundancy of the two `PixelHitMatchers` in the `ElectronSeedGenerator`, as anticipated in this comment [1]. The first thing I realized was that a big chunk of the PixelHitMatcher is obsolete, as indicated by this over 6 years old comment [2]. This includes so much code that it's probably a good to have a PR that removes this code before thinking of any other changes.

As a little side effect, I moved the related ElectronSeedProducer plugin into a single cc file and (more unrelated) moved the only remaining class from EgammaElectronProducers/interface to the plugins directory because only the plugins depend on this class and therefore a separate EgammaElectronProducers library is not needed.

[1] https://github.com/cms-sw/cmssw/pull/28039#discussion_r329598269
[2] https://github.com/cms-sw/cmssw/blob/master/RecoEgamma/EgammaElectronAlgos/src/PixelHitMatcher.cc#L249

#### PR validation:

Code compiles, local matrix tests pass.

#### if this PR is a backport please specify the original PR:

No backport intended.

PR #28213, opened 2019-10-18
Remove manually generated cfi file for ElectronSeedProducer

guitargeek:
#### PR description:

The `ecalDrivenElectronSeeds_cfi.py` file redundantly redefines all the configuration values that are already set in the `fillDescriptions()` method of the `ElectronSeedProducer`. It would be better to avoid this to make the `ElectronSeedProducer` more maintainable. The entries in `fillDescriptions` have been reshuffled and commented to match the previous easy-to-read structure in the cfi file.

On this occasion, the `fromTrackerSeeds` parameter was finally removed completely, as it could not be disabled anymore anyway (the code would have segfaulted has reported in https://github.com/cms-sw/cmssw/issues/27541) and the standard tracker seeds should always be used now. That finishes what https://github.com/cms-sw/cmssw/pull/28039 and https://github.com/cms-sw/cmssw/pull/28133 started.

Another update is that the HoE values are not stored anymore in some unused vectors. I think this must be a remnant from removing the HoE values in the ElectronSeed class [1] two years ago.

Unrelated to this, there is a separate commit that moves the forward declaration and implementation of `GsfElectronAlgoHeavyObjectCache` into the `GsfElectronAlgo` class. It was always a bit strange to me why many other helper classes for GsfElectronAlgo (that are also much longer) don't have their own files while the HeavyObjectCache (which really doesn't make sense to use in another context) does not. 

[1] https://github.com/cms-sw/cmssw/blob/master/DataFormats/EgammaReco/interface/ElectronSeed.h#L16

#### PR validation:

No backport intended.


slava77:
@cmsbuild please test

RecoEgamma/EgammaElectronProducers/plugins/ElectronSeedProducer.cc
```diff
@@ -283,78 +256,93 @@ void ElectronSeedProducer::fillDescriptions(edm::ConfigurationDescriptions& desc
...
+    psd0.add<edm::InputTag>("initialSeeds", {""});  //keep for be compatibility
+    psd0.add<std::vector<edm::InputTag>>("initialSeedsVector", {});
+    psd0.add<bool>("preFilteredSeeds", false);
+    psd0.add<bool>("useRecoVertex", false);
+    psd0.add<edm::InputTag>("vertices", {"offlinePrimaryVerticesWithBS"});
+    psd0.add<edm::InputTag>("beamSpot", {"offlineBeamSpot"});
+    psd0.add<bool>("dynamicPhiRoad", true);
+
+    // specify where to get the hits from
+    psd0.add<edm::InputTag>("measurementTrackerEvent", {"MeasurementTrackerEvent"});
```
slava77: is this parameter still used?

RecoEgamma/EgammaElectronProducers/plugins/ElectronSeedProducer.cc
```diff
@@ -283,78 +256,93 @@ void ElectronSeedProducer::fillDescriptions(edm::ConfigurationDescriptions& desc
...
+    psd0.add<double>("hOverEConeSize", 0.15);
+    psd0.add<double>("maxHOverEBarrel", 0.15);
+    psd0.add<double>("maxHOverEEndcaps", 0.15);
+    psd0.add<double>("maxHBarrel", 0.0);
+    psd0.add<double>("maxHEndcaps", 0.0);
+
+    // H/E rechits
+    psd0.add<edm::InputTag>("hcalRecHits", {"hbhereco"});  // OBSOLETE
+    psd0.add<double>("hOverEHBMinE", 0.7);                 // OBSOLETE
+    psd0.add<double>("hOverEHFMinE", 0.8);                 // OBSOLETE
```
slava77: better remove these, if they are obsolete

slava77 (commented):
nice work grouping the parameters in the ElectronSeedProducer.

RecoEgamma/EgammaElectronProducers/plugins/ElectronSeedProducer.cc
guitargeek: Yes, this one is still used. Not in the current configuration thought, but at least the plugin could theoretically be configured such that it is actually used in a meaningful way.

RecoEgamma/EgammaElectronProducers/plugins/ElectronSeedProducer.cc
guitargeek: Thanks, that should  definitely go away. I will push an update after local tests.

slava77:
@cmsbuild please test

slava77:
+1

for https://github.com/cms-sw/cmssw/pull/28213	6df92ce 
- essentially technical code changes are in line with the PR description and the follow up review

fabiocos:
@chayanit @pgunnell please have a look, this is just changing a commented line. I will move forward with integration

chayanit:
+1

fabiocos:
+1

PR #28230, opened 2019-10-21
Bug fix for concurrent IOVs

wddgit:
#### PR description:

Only affects corner case with nConcurrentIOVs set to
more than one and the appendToDataLabel parameter set
to a non-empty value. The symptom is an assert failure,
when the assert should pass. The assert check didn't take
into account that the parameter could change the DataKey
labels.

#### PR validation:

Existing tests should be sufficient. The failure was in code
that is itself only a sanity check.

### DIFF
--- a/FWCore/Framework/interface/DataProxyProvider.h
+++ b/FWCore/Framework/interface/DataProxyProvider.h
@@ -78,7 +78,8 @@ namespace edm {
 
         EventSetupRecordKey const& recordKey() const;
 
-        void insert(std::vector<std::pair<DataKey, std::shared_ptr<DataProxy>>>&&);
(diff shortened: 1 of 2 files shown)


wddgit:
please test

Dr15Jones:
How did you test that what failed now works?

wddgit:
The runTheMatrix step3 reco test I was using to make plots for the CHEP talk hits the corner case and fails with the assert after I modified it to run with 2 concurrent IOVs. It runs successfully to completion with the fix.

Dr15Jones:
please test

Dr15Jones:
+1

fabiocos:
+1

PR #28236, opened 2019-10-22
Phase2-hgx226 Correct wafer size for the HGCal V11 version

bsunanda:
#### PR description:

Correct wafer size for the v11 version. It is also wrong for v9 and v10. But the corrected size is not added in the old workflows

#### PR validation:

Tested with standard matrix test

Nothing special


bsunanda:
@cmsbuild Please test

fabiocos:
@bsunanda @kpedro88 as far as I can see only scenarios not previously used for production are modified, and this should be acceptable IMO. Are the new xml files provided for private studies and/or reference?

bsunanda:
V11 is the current default for HGCal DPG and we shall improve from that version. So the changes in the V11 wafer size is extremely useful. The changes for v9 and v10 are made in v9c and v10c - they are not in any workflow - but may be one can make specific scenarios for some comparisons for change in version - they will be indeed some private scenarios to my current understanding

kpedro88:
+upgrade
changes isolated to D46 as expected

cvuosalo:
+1

fabiocos:
+1

PR #28250, opened 2019-10-23
Phase2-hcx235 Check HFNose geometry and also add the possibility to run SIM+Valid for HGCal

bsunanda:
#### PR description:

Check HFNose geometry and also add the possibility to run SIM+Valid for HGCal

#### PR validation:

Used standard tests in runTheMatrix

Nothing special


bsunanda:
@cmsbuild Please test

jfernan2:
+1
Only changes in unstable 11634 wf, Test files mainly, not real sequences added to configuration so no extra plots yet

cvuosalo:
+1

civanch:
+1

BigProducts/Simulation/BuildFile.xml
```diff
@@ -20,6 +20,7 @@
 <use name="SimG4Core/DD4hepGeometry"/>
 <use name="Validation/HcalHits"/>
 <use name="Validation/EcalHits"/>
+<use name="Validation/HGCalValidation"/>
```
fabiocos: @bsunanda this package contains not only classes related to simulation but also to reconstruction validation, I do not think this makes a huge difference but I think it is the first case to be included in this BigProduct definition. @Dr15Jones @smuzaffar this falls in the ```core``` domain, what's you opinion here?

bsunanda: I often run sim validation along with simulation. If this is not a part of biglib, I have to make a separate validation package for SIM only

smuzaffar: as Validation/HGCalValidation depends on geant4 ( https://github.com/cms-sw/cmssw/blob/master/Validation/HGCalValidation/BuildFile.xml#L27 ) , so it is correct to add it to biglib othewise cmsRun will crash due to loading of geant4 shared and static libs.

smuzaffar: so if I understand correctly, Validation/HGCalValidation (shared lib or plugin)  was not used  by any test/workflow that is why we are not getting crashes?
@bsunanda , do you know what workflow one should run to get this loaded now?

bsunanda: This is not in any standard workflow - if one includes any SimHit validation in the workflow (for example the cfg file Validation/HGCalValidation/test/python/testHFNoseSimHitStudy_cfg.py it needs the library in biglib). It is like Validation/HcalHits package. In Hcal, we have at least 3 packages - for SimHit, Digi, RecHit. Here for HGCal and HFNose we have only 1.

smuzaffar:
+core

bsunanda:
@fabiocos @kpedro88 Could you please merge this PR so that I can proceed with the next step?

fabiocos:
+1

PR #28262, opened 2019-10-24
Merged the MeasurementEstimators into the PixelHitMatcher

guitargeek:
#### PR description:

After my last PR https://github.com/cms-sw/cmssw/pull/28213 definitely removed the support in the electron pixel hit matching for not using the standard tracker seeds, there is still some code left to remove which is now obsolete. Much of it could be found in the classes `BarrelMeasurementEstimator` and `ForwardMeasurementEstimator`, which unnecessarily inherit from `MeasurementEstimator` and only one of their respective member functions is actually used. Whether a track extrapolation falls into the matching window in the next layer.

I propose to integrate the remaining functionality of these estimator classes after cleaning them up as private functors of the `PixelHitMatcher` class. Sorry for having multiple of these small cleanup PRs, but I don't want to risk any differences in the comparisons by doing to much at once and possibly making mistakes.

#### PR validation:

No backport intended.


slava77:
@cmsbuild please test

slava77:
code-checks

slava77:
>     * File  RecoEgamma/EgammaElectronAlgos/interface/PixelHitMatcher.h  modified in PR(s): #28039

I see that #28039 was merged 23 days ago.
I thought that the purpose of the checks was to see conflicts with open PRs.
@smuzaffar @fabiocos please check the logic of these checks.

smuzaffar:
thanks @slava77 for pointing this, I will check.

RecoEgamma/EgammaElectronAlgos/interface/PixelHitMatcher.h
```diff
@@ -22,22 +22,25 @@
...
-
 #include "DataFormats/TrajectorySeed/interface/TrajectorySeedCollection.h"
 #include "DataFormats/EgammaReco/interface/ElectronSeed.h"
-
 #include "RecoTracker/TransientTrackingRecHit/interface/TSiPixelRecHit.h"
 #include "RecoTracker/TkDetLayers/interface/GeometricSearchTracker.h"
 #include "TrackingTools/MeasurementDet/interface/LayerMeasurements.h"
+#include "TrackingTools/TrajectoryState/interface/TrajectoryStateOnSurface.h"
+#include "TrackingTools/DetLayers/interface/MeasurementEstimator.h"
+#include "RecoTracker/TransientTrackingRecHit/interface/TSiPixelRecHit.h"
```
slava77: this is a duplicate from L28

RecoEgamma/EgammaElectronAlgos/interface/PixelHitMatcher.h
```diff
@@ -22,22 +22,25 @@
...
+#include "DataFormats/GeometrySurface/interface/BoundPlane.h"
+#include "TrackingTools/DetLayers/interface/MeasurementEstimator.h"
```
slava77: dup of L32

RecoEgamma/EgammaElectronAlgos/interface/PixelHitMatcher.h
```diff
@@ -22,22 +22,25 @@
...
+#include "TrackingTools/TrajectoryState/interface/TrajectoryStateOnSurface.h"
```
slava77: also a dup

RecoEgamma/EgammaElectronAlgos/src/PixelHitMatcher.cc
```diff
@@ -8,13 +8,83 @@
...
+#include "DataFormats/GeometryVector/interface/VectorUtil.h"
+#include "TrackingTools/TransientTrackingRecHit/interface/TransientTrackingRecHit.h"
+#include "DataFormats/Math/interface/normalizedPhi.h"
 
 #include <typeinfo>
 #include <bitset>
 
+namespace std {
+  template <>
+  struct hash<std::pair<const GeomDet *, GlobalPoint> > {
```
slava77: this is now too inconspicuous.
Would it work in this context to move this hash to the anonymous namespace and add that as a 3rd template parameter in the PixelHitMatcher::operator unordered_map?

RecoEgamma/Examples/plugins/ElectronSeedAnalyzer.cc
```diff
@@ -28,8 +28,6 @@
 #include "RecoEgamma/EgammaElectronAlgos/interface/FTSFromVertexToPointFactory.h"
 #include "TrackingTools/TrajectoryState/interface/FreeTrajectoryState.h"
 #include "TrackingTools/MaterialEffects/interface/PropagatorWithMaterial.h"
-#include "RecoEgamma/EgammaElectronAlgos/interface/BarrelMeasurementEstimator.h"
-#include "RecoEgamma/EgammaElectronAlgos/interface/ForwardMeasurementEstimator.h"
 #include "DataFormats/GeometryCommonDetAlgo/interface/PerpendicularBoundPlaneBuilder.h"
 #include "TrackingTools/MaterialEffects/interface/PropagatorWithMaterial.h"
```
slava77: old case here of a dup of L30

RecoEgamma/EgammaElectronAlgos/src/PixelHitMatcher.cc
```diff
@@ -8,13 +8,83 @@
...
+#include "TrackingTools/TrajectoryParametrization/interface/GlobalTrajectoryParameters.h"
+#include "TrackingTools/DetLayers/interface/rangesIntersect.h"
+#include "DataFormats/GeometryVector/interface/VectorUtil.h"
+#include "DataFormats/Math/interface/normalizedPhi.h"
+#include "RecoEgamma/EgammaElectronAlgos/interface/ElectronUtilities.h"
+#include "TrackingTools/TrajectoryParametrization/interface/GlobalTrajectoryParameters.h"
+#include "TrackingTools/DetLayers/interface/rangesIntersect.h"
+#include "DataFormats/GeometryVector/interface/VectorUtil.h"
+#include "TrackingTools/TransientTrackingRecHit/interface/TransientTrackingRecHit.h"
+#include "DataFormats/Math/interface/normalizedPhi.h"
```
slava77: dup of L11-15

RecoEgamma/EgammaElectronAlgos/src/PixelHitMatcher.cc
```diff
@@ -8,13 +8,83 @@
 #include "DataFormats/DetId/interface/DetId.h"
 #include "DataFormats/GeometryCommonDetAlgo/interface/PerpendicularBoundPlaneBuilder.h"
 #include "FWCore/MessageLogger/interface/MessageLogger.h"
+#include "RecoEgamma/EgammaElectronAlgos/interface/ElectronUtilities.h"
```
slava77: dup of L2

RecoEgamma/EgammaElectronAlgos/interface/PixelHitMatcher.h
```diff
@@ -22,22 +22,25 @@
...
+#include "DataFormats/GeometrySurface/interface/BoundPlane.h"
```
slava77: dup

RecoEgamma/EgammaElectronAlgos/interface/PixelHitMatcher.h
```diff
@@ -22,22 +22,25 @@
...
 #include "DataFormats/TrajectorySeed/interface/TrajectorySeedCollection.h"
 #include "DataFormats/EgammaReco/interface/ElectronSeed.h"
-
 #include "RecoTracker/TransientTrackingRecHit/interface/TSiPixelRecHit.h"
 #include "RecoTracker/TkDetLayers/interface/GeometricSearchTracker.h"
 #include "TrackingTools/MeasurementDet/interface/LayerMeasurements.h"
+#include "TrackingTools/TrajectoryState/interface/TrajectoryStateOnSurface.h"
+#include "TrackingTools/DetLayers/interface/MeasurementEstimator.h"
+#include "RecoTracker/TransientTrackingRecHit/interface/TSiPixelRecHit.h"
+#include "DataFormats/GeometrySurface/interface/BoundPlane.h"
```
slava77: not used in this file, move to .cc

slava77 (commented):
please also run an include checker; I doubt that I noticed all replicated includes

guitargeek:
Hi Slava, I now put the hash functor and a using statement for the map type in the new file `RecoEgamma/EgammaElectronAlgos/interface/utils.h`, because I didn't feel it belongs into one of the other more reco specific files as it's just a datastructure. Are you ok with the names? I know `utils.h` does not say much itself, but I think with the context of the full path that should be ok.

slava77:
> Hi Slava, I now put the hash functor and a using statement for the map type in the new file `RecoEgamma/EgammaElectronAlgos/interface/utils.h`, because I didn't feel it belongs into one of the other more reco specific files as it's just a datastructure. Are you ok with the names? I know `utils.h` does not say much itself, but I think with the context of the full path that should be ok.

was there a problem using a method from the anonymous namespace? It's only needed in the unordered_map in .cc and the std::unordered_map has a separate parameter for the hash method

guitargeek:
That would have worked, but since I wanted to reuse the code for the `TrajSeedMatcher`. So it had to go in some header file because I wanted to inline it for best performance, and so far I thought anonymous namespaces in header files is not really good practice because I read it in the core guidelines:
https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#Rs-unnamed
And indeed, the compiler would give me a warning if I would use the anonymous namespace there:

```
scratch/rembser/cmssw/dev/CMSSW_11_0_0_pre11/src/RecoEgamma/EgammaElectronAlgos/interface/TrajSeedMatcher.h:48:7: warning: 'TrajSeedMatcher' has a field 'TrajSeedMatcher::trajStateFromPointPosChargeCache_' whose type uses the anonymous namespace [-Wsubobject-linkage]
```

RecoEgamma/EgammaElectronAlgos/src/PixelHitMatcher.cc
```diff
@@ -173,7 +217,7 @@ std::vector<SeedWithInfo> PixelHitMatcher::operator()(const std::vector<const Tr
           auto idx2 = std::distance(hits.first, it2);
           const DetId id2 = it2->geographicalId();
           const GeomDet *geomdet2 = it2->det();
-          const std::pair<const GeomDet *, GlobalPoint> det_key(geomdet2, hit1Pos);
+          const auto det_key = std::make_pair(geomdet2->gdetIndex(), hit1Pos);
```
slava77: the logic of the map filling is different now.
Is the gdetIndex unique?

slava77:
> That would have worked, but I wanted to reuse the code for the `TrajSeedMatcher`

I see, a feature creep

RecoEgamma/EgammaElectronAlgos/src/PixelHitMatcher.cc
guitargeek: I honestly didn't really check, I just did this for the sake of unifying with the HLT code which does that. I trust Sams modification there. It's certainly better than C-casting the GeomDet [1] to a `long long` and hoping for the best!

[1] https://github.com/cms-sw/cmssw/blob/master/Geometry/CommonDetUnit/interface/GeomDet.h#L102

RecoEgamma/EgammaElectronAlgos/interface/TrajSeedMatcher.h
```diff
@@ -290,8 +272,8 @@ class TrajSeedMatcher {
   std::unordered_map<int, TrajectoryStateOnSurface> trajStateFromVtxPosChargeCache_;
   std::unordered_map<int, TrajectoryStateOnSurface> trajStateFromVtxNegChargeCache_;
```
slava77: change these too?

guitargeek: These are different, as the key is just `int` and not `pair<int, GlobalPoint const&>`.

slava77:
On 11/1/19 9:01 AM, Slava Krutelyov wrote:
> *@slava77* commented on this pull request.
> 
> ------------------------------------------------------------------------
> 
> In RecoEgamma/EgammaElectronAlgos/interface/TrajSeedMatcher.h

> 
>> @@ -290,8 +272,8 @@ class TrajSeedMatcher {
>    std::unordered_map<int, TrajectoryStateOnSurface> trajStateFromVtxPosChargeCache_;
>    std::unordered_map<int, TrajectoryStateOnSurface> trajStateFromVtxNegChargeCache_;
> 
> change these too?

nm, wrong type => not applicable
>

slava77:
@cmsbuild please test

slava77:
+1

for https://github.com/cms-sw/cmssw/pull/28262	50c5ef8 
- essentially all technical code changes are in line with the PR description and the follow up review

fabiocos:
+1

smuzaffar:
@slava77 , @fabiocos ( referring to your cmment: https://github.com/cms-sw/cmssw/pull/28262#issuecomment-548139480 ) we had missed few github web hooks that is why for some PRs the https://github.com/cms-sw/cms-prs/blob/master/cms-sw/cmssw/.other/files_changed_by_prs.json was not updated. This has been fixed now and just to make sure that we have not missed any thing we force update it every 4 hours.

PR #28269, opened 2019-10-25
[GCC9] fix implicit-fallthrough warning for TrackingTools/TrackAssociator

smuzaffar:
This should fix the GCC9 compilation warning

### DIFF
--- a/TrackingTools/TrackAssociator/src/TrackDetMatchInfo.cc
+++ b/TrackingTools/TrackAssociator/src/TrackDetMatchInfo.cc
@@ -353,7 +353,7 @@ DetId TrackDetMatchInfo::findMaxDeposition(EnergyType type) {
           id = (*hit)->id();
         }
       }
-    }
+    } break;
     default:
       throw cms::Exception("FatalError")
           << "Maximal energy deposition: unkown or not implemented energy type requested, type:" << type;
--- a/TrackingTools/TrackAssociator/src/TrackDetMatchInfo.cc
+++ b/TrackingTools/TrackAssociator/src/TrackDetMatchInfo.cc
@@ -329,7 +329,8 @@ DetId TrackDetMatchInfo::findMaxDeposition(EnergyType type) {
     } break;
     case TowerTotal:
     case TowerEcal:
-    case TowerHcal: {
+    case TowerHcal:
+    case TowerHO: {
       for (std::vector<const CaloTower*>::const_iterator hit = towers.begin(); hit != towers.end(); hit++) {
         double energy = 0;
         switch (type) {


smuzaffar:
please test

perrotta:
The GCC9 compilation warning hit here a real bug, which doesn't see, to be correctly fixed as done here.

At first, I wondered why the bug was never hit. It looks like that the ```TrackDetMatchInfo::findMaxDeposition(...)``` method is only called in CMSSW with arguments (see https://cmssdt.cern.ch/lxr/search?%21v=CMSSW_11_0_X_2019-10-24-2300&_filestring=&_string=findMaxDeposition):
- ```TrackDetMatchInfo::EcalRecHits```
- ```TrackDetMatchInfo::HcalRecHits```
- ```TrackDetMatchInfo::HORecHits```

Therefore, by chance the bug is currently never encountered.

I think that the simplest correct fix should be to replace L330 to L360 with
```C++
    case TowerTotal:
    case TowerEcal:
    case TowerHcal:
    case TowerHO: {
      for (std::vector<const CaloTower*>::const_iterator hit = towers.begin(); hit != towers.end(); hit++) {
        double energy = 0;
        switch (type) {
          case TowerTotal:
            energy = (*hit)->energy();
            break;
          case TowerEcal:
            energy = (*hit)->emEnergy();
            break;
          case TowerHcal:
            energy = (*hit)->hadEnergy();
            break;
          case TowerHO:
            energy = (*hit)->energy();
            break;
          default:
            throw cms::Exception("FatalError") << "Unknown calo tower energy type: " << type;
        }
        if (energy > maxEnergy) {
          maxEnergy = energy;
          id = (*hit)->id();
        }
      }
    } break;
    default:
      throw cms::Exception("FatalError")
          << "Maximal energy deposition: unkown or not implemented energy type requested, type:" << type;
  }
```

(i.e. adding another line with
```C++
      case TowerHO:
```
after current L332, besides keeping the additional ```break``` statement as now in this PR)

Adding Tracking contacts to have a look and get confirmation of what I'm saying or comment otherwise: @mtosi @JanFSchulte

smuzaffar:
@mtosi @JanFSchulte , do you agree with the suggestion made by @perrotta here https://github.com/cms-sw/cmssw/pull/28269#issuecomment-546429812 ? I can update the PR if no objections

mtosi:
ciao, sorry the delay
+1 (of course)

smuzaffar:
please test

perrotta:
+1
- Bug fixed in a (so far) unaccessed part of TrackDetMatchInfo::findMaxDeposition(...)

fabiocos:
+1

PR #28271, opened 2019-10-25
[GCC9] fix warning by adding the missing break for RecoParticleFlow/PFProducer

smuzaffar:
This should fix the GCC9 compilation warning, added missing break

### DIFF
--- a/RecoParticleFlow/PFProducer/plugins/PFEGammaProducer.cc
+++ b/RecoParticleFlow/PFProducer/plugins/PFEGammaProducer.cc
@@ -197,7 +197,7 @@ void PFEGammaProducer::produce(edm::Event& iEvent, const edm::EventSetup& iSetup
       switch (elements[0].type()) {
         case reco::PFBlockElement::SC:
           edm::LogError("PFEGammaProducer") << "PFBLOCKALGO BUG!!!! Found a SuperCluster in a block by itself!";
-          [[fallthrough]];
+          break;
         case reco::PFBlockElement::PS1:
         case reco::PFBlockElement::PS2:
         case reco::PFBlockElement::ECAL:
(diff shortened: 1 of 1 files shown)


smuzaffar:
please test

RecoParticleFlow/PFProducer/plugins/PFEGammaProducer.cc
```diff
@@ -197,6 +197,7 @@ void PFEGammaProducer::produce(edm::Event& iEvent, const edm::EventSetup& iSetup
       switch (elements[0].type()) {
         case reco::PFBlockElement::SC:
           edm::LogError("PFEGammaProducer") << "PFBLOCKALGO BUG!!!! Found a SuperCluster in a block by itself!";
+          [[fallthrough]];
```
perrotta: I would rather ```break``` here: @hatakeyamak @bendavid do you agree ?

smuzaffar: @hatakeyamak @bendavid do you agree?

smuzaffar: ping

lgray: break will yield the correct behavior still since the SuperCluster block element is alone in the block to begin with. Probably better to do that.

smuzaffar:
please test

RecoParticleFlow/PFProducer/plugins/PFEGammaProducer.cc
hatakeyamak: Oops. Somehow, I didn't notice this ping from this PR until now.
The "break" as suggested by @perrotta and @lgray sounds reasonable to me.

perrotta:
+1
- Bug fix, in a piece of code evidently never hit in any of the current workflows
- Jenkins tests pass: no differences in outputs

fabiocos:
+1

PR #28274, opened 2019-10-25
Fix for gcc9 warning in HLTrigger/JSONMonitoring

mrodozov:
#### PR description:
Fixes
 https://cmssdt.cern.ch/SDT/cgi-bin/buildlogs/slc7_amd64_gcc900/CMSSW_11_0_X_2019-10-24-2300/HLTrigger/JSONMonitoring

#### PR validation:
Compile without warning

### DIFF
--- a/HLTrigger/JSONMonitoring/plugins/L1TriggerJSONMonitoring.cc
+++ b/HLTrigger/JSONMonitoring/plugins/L1TriggerJSONMonitoring.cc
@@ -41,7 +41,7 @@ struct L1TriggerJSONMonitoringData {
     std::vector<unsigned int> l1tAcceptCalibration;  // number of "calibration" events accepted by each L1 trigger
     std::vector<unsigned int> l1tAcceptRandom;       // number of "random" events accepted by each L1 trigger
     std::vector<unsigned int> tcdsAccept;  // number of "physics", "calibration", "random" and other event types
-    int prescaleIndex;                     // prescale column index
+    int prescaleIndex = {};                // prescale column index
   };
 
   // variables initialised for each run
--- a/HLTrigger/JSONMonitoring/plugins/L1TriggerJSONMonitoring.cc
+++ b/HLTrigger/JSONMonitoring/plugins/L1TriggerJSONMonitoring.cc
@@ -33,8 +33,10 @@
 #include "CondFormats/DataRecord/interface/L1TUtmTriggerMenuRcd.h"
 
 struct L1TriggerJSONMonitoringData {
-  // variables accumulated event by event in each stream
+  // special values for prescale index checks
   static constexpr const int kPrescaleUndefined = -2;
+  static constexpr const int kPrescaleConflict = -1;
+  // variables accumulated event by event in each stream
   struct stream {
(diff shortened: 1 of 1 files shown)


mrodozov:
please test

HLTrigger/JSONMonitoring/plugins/L1TriggerJSONMonitoring.cc
```diff
@@ -41,7 +41,7 @@ struct L1TriggerJSONMonitoringData {
     std::vector<unsigned int> l1tAcceptCalibration;  // number of "calibration" events accepted by each L1 trigger
     std::vector<unsigned int> l1tAcceptRandom;       // number of "random" events accepted by each L1 trigger
     std::vector<unsigned int> tcdsAccept;  // number of "physics", "calibration", "random" and other event types
-    int prescaleIndex;                     // prescale column index
+    int prescaleIndex = {};                // prescale column index
```
fwyzard: This is supposed to be initialised here, since it is initialised before its first use.

If there is any suspicion that it is used uninitialised, then it should be initialised to `kPrescaleUndefined`, not `0`.
Then, it would make sense to initialise also:
```
stream.processed = 0;
lumisection.prescaleIndex = kPrescaleUndefined;
```

fwyzard (changes requested):

HLTrigger/JSONMonitoring/plugins/L1TriggerJSONMonitoring.cc
mrodozov: kPrescaleUndefined was undefined in that scope (which I blindly tried and now it doesn't compile). would you suggest the change please, I'll commit it :/

mrodozov:
please test

HLTrigger/JSONMonitoring/plugins/L1TriggerJSONMonitoring.cc
```diff
@@ -34,14 +34,15 @@
 
 struct L1TriggerJSONMonitoringData {
   // variables accumulated event by event in each stream
+  static constexpr const int kPrescaleUndefined = -2;
```
fwyzard: please do not duplicate this from the `L1TriggerJSONMonitoring` class

fwyzard: maybe you can move the whole block
```c++
  // special values for prescale index checks
  static constexpr const int kPrescaleUndefined = -2;
  static constexpr const int kPrescaleConflict = -1;
```
above the comment on line 36
```
// variables accumulated event by event in each stream
```

and modify `L1TriggerJSONMonitoring` to reuse it from `L1TriggerJSONMonitoringData` ?

fwyzard (changes requested):

mrodozov:
please test
lets see if this works :)

fwyzard:
> lets see if this works :)

thanks for the changes !

fabiocos:
@Martin-Grunewald is this fix ok to be integrated?

Martin-Grunewald:
+1

fabiocos:
+1

PR #28280, opened 2019-10-25
Cleanup old and obsolete configurations related to MF

namapane:
#### PR description:

Cleanup of test configurations for obsolete field maps that are no longer supported since a long time. 
Made all .pys refer to the proper StandardSequences. 
No functional changes expected.
Note: some of the modified test .pys are quite obviously obsolete (eg those in DPGAnalysis/Skims seems to have been last updated at the time of CRAFT): please note the goal here is not to fix/update them, just to eradicate every reference in CMSSW to maps declared obsolete so that they are not used as an example elsewhere.
Also, some cleanup of tests configurations under MagneticField, and revert  one xml file that was erroneusly removed in an earlier cleanup (useful only for reference).

#### PR validation:

Checked the formal correctness of each modified .py (a good fraction of the modified .py files seems to be obsolete and does not run out of the box).
Checked functionality of modified .py files under MagneticField.


perrotta:
please test

cvuosalo:
+1

kpedro88:
+upgrade

perrotta:
@namapane , the configuration MagneticFieldVolumes_1103l.xml was removed more that two years ago, and apparently never used since then.
Out of curiosity, which is the reason to recover it now in the release master?
The corresponding test is also still commented out in https://github.com/cms-sw/cmssw/pull/28280/files#diff-528eee0e3ac815194d51f6b8164914c1R26

namapane:
@perrotta, because this geometry is still in use (for the 2T and 4Tworking points only), we just don't notice because it is taken by default from the DB. It was a mistake to remove the corresponding xml file since it is a geometry that is still supported. 
It is relevant to have the xml back in the context of the migration to DD4hep.
Regarding the link you posted,  it points to displayMFGeom_cfg.py which is supposed to be used for visualization.  You can visualize only one geometry at a time, and the file by default shows the geometry used in 3.8T maps, which is more relevant. (Visualization does not work for me at the moment BTW, but this is a different issue).

DPGAnalysis/Skims/python/CSCSkim_BFieldStudies_MC_ToscaMap090322_cfg.py
```diff
@@ -27,9 +27,7 @@
 #------------------------------------------
 process.load("Configuration/StandardSequences/Geometry_cff")
 
-process.load("Configuration.StandardSequences.MagneticField_38T_UpdatedMap_cff")
-# trick to make it work with newnew magfield (not in 229)
-process.VolumeBasedMagneticFieldESProducer.version='grid_1103l_090322_3_8t'
+process.load("Configuration.StandardSequences.MagneticField_cff")
```
fabiocos: @namapane for my understanding, this replacement is not corresponding to the intended configuration described in the title, so I wonder whether it would not be better just to get rid of this (not the only case)

namapane: @fabiocos, this set of .pys is quite obviously a relic of CRAFT and it is apparent that they have not been used since. They are also no longer running as far as I can tell. Anyhow the MF configuration file they point to (MagneticField_38T_UpdatedMap_cff) already collapses to the standard sequence, cf:
https://cmssdt.cern.ch/dxr/CMSSW/source/Configuration/StandardSequences/python/MagneticField_38T_UpdatedMap_cff.py
So the inconsistency with the file name is already present, independently of this PR.
Indeed these cfg.py should be considered for removal. As I noted in the PR description, such a cleanup is outside the scope of this PR because I have no elements myself to decide what has to go and what has to be kept maybe for historical reasons. I would prefer such a removal, if deemed appropriate, to happen independently, at least because of the number of signatures involved in approving this PR.

fabiocos:
+operations

the update of StandardSequences looks coherent with the purpose of the PR

perrotta:
+1
- Modifications in the reco (test) code are consequent to the removal of the obsolete BField configurations
- Jenkins tests pass

perrotta:
Maybe some further cleaning should be applied to 
- CondFormats/​MFObjects/​test/​writeMagFieldConfigDB.py
- SimPPS/PPSPixelDigiProducer/test/FromGun2DigiAnal_.py
(ok for moving to another PR with it, to speed up)

fabiocos:
@civanch @ssekmen @sbein @chayanit @pgunnell @qliphy @agrohsje @alberto-sanchez @rekovic this proposal looks ok to me, unless you have objections I wold like to move forward with this cleaning that should be practically transparent

ssekmen:
+1

namapane:
@perrotta, writeMagFieldConfigDB.py is used to create db objects for the DB and I would have a slight preference to keep history of past configurations there. Since this .py is not as a cfi that can be loaded elsewhere it can make no harm, and it can also not be misused as an example (I think). But if you insist I can clean it up in a future PR.

Regarding SimPPS/PPSPixelDigiProducer/test/FromGun2DigiAnal_.py: this contains both the correct standard sequence (L103) and an hardcoded one (L139).
I do not know which one would be taken in the job (the hardcoded one has no label so it cannot be accessed separately from the default one, I think). Also I think the hardcoded one cannot be working since MagneticField/GeomBuilder/data/MagneticFieldVolumes_1103l.xml had been removed some time ago. 
However, before a clean-up it would be better to know what was the  intended purpose of this hardcoded configuration. @mundim, you appear to be the author: can you please clarify why there is this configuration block 
https://github.com/cms-sw/cmssw/blob/master/SimPPS/PPSPixelDigiProducer/test/FromGun2DigiAnal_.py#L118-L201
besides the standard MF configuration at L103?

In any case, I would prefer to proceed with further cleanups separately (I need to make other PRs for MagneticField and I could add them there).

civanch:
+1

fabiocos:
+1

fabiocos:
merge

@rekovic @santocch @agrohsje @qliphy please have a look

mundim:
> @perrotta, writeMagFieldConfigDB.py is used to create db objects for the DB and I would have a slight preference to keep history of past configurations there. Since this .py is not as a cfi that can be loaded elsewhere it can make no harm, and it can also not be misused as an example (I think). But if you insist I can clean it up in a future PR.
> 
> Regarding SimPPS/PPSPixelDigiProducer/test/FromGun2DigiAnal_.py: this contains both the correct standard sequence (L103) and an hardcoded one (L139).
> I do not know which one would be taken in the job (the hardcoded one has no label so it cannot be accessed separately from the default one, I think). Also I think the hardcoded one cannot be working since MagneticField/GeomBuilder/data/MagneticFieldVolumes_1103l.xml had been removed some time ago.
> However, before a clean-up it would be better to know what was the intended purpose of this hardcoded configuration. @mundim, you appear to be the author: can you please clarify why there is this configuration block
> https://github.com/cms-sw/cmssw/blob/master/SimPPS/PPSPixelDigiProducer/test/FromGun2DigiAnal_.py#L118-L201
> besides the standard MF configuration at L103?
> 
> In any case, I would prefer to proceed with further cleanups separately (I need to make other PRs for MagneticField and I could add them there).

Hello. I'm sorry, but I was unable to read of reply to the messages these days. So, in fact, I'm not the developer, but managing the port of this code from TOTEM collaboration to CMS. I had a more detailed look into this config and I would just delete it in spite of the usefulness to have it available for digi tests, but there is another (simpler) one. I apologize to eventually have made you loose you time.

santocch:
+1

namapane:
@mundim, thanks for checking. Do you mean removing L118-L201, or removing the entire file?

mundim:
> @mundim, thanks for checking. Do you mean removing L118-L201, or removing the entire file?

Once again, sorry for the delay. I could not reply promptly and then I just forgot about it.
I meant removing the entire file (it needs more changes to move from TOTEM framework to CMS and this scaped the attention of the former developer as well as mine).

PR #28282, opened 2019-10-25
Many classes in EgammaElectronAlgos turned into pure functions

guitargeek:
#### PR description:

This is one of the occasional PRs to make the EGamma algos follow a bit more a functional paradigm for better multithread-friendlieness and code readability.

I propose to convert the member functions of all the classes in `EgammaElectronAlgos` that don't have member variables into free functions that should all live in the same namespce (I suggest here the name `electronAlgos`, I think one namespace for each former class would be a bit excessive, but I don't think CMS has any guidelines for that so that's maybe a thing to discuss). Functions that were previously private are simply not declared in the public header file.

All these functions were then adapted such that they are pure, i.e. they don't mutate their inputs and always have an output of a defined type. This means that when the GsfElectrons are created, they are actually only mutated in the GsfElectronAlgo itself and never in the various helper functions that it calls.

Optimally, the formerly private functions should live in the anonymous namespace, but I was a bit reluctant to do that because it would cause huge chunks of code to shift indentation without actual changes. Is there maybe a way to put functions in the anonymous namespace without enclosing them with `namespace { }` and therefore causing an additional layer of indentation?

This PR does not clash with https://github.com/cms-sw/cmssw/pull/28262.

#### PR validation:

No backport intended.

### DIFF
--- a/RecoEgamma/EgammaElectronAlgos/interface/EgAmbiguityTools.h
+++ b/RecoEgamma/EgammaElectronAlgos/interface/EgAmbiguityTools.h
@@ -7,7 +7,7 @@
 #include "DataFormats/GsfTrackReco/interface/GsfTrackFwd.h"
 #include "DataFormats/EcalRecHit/interface/EcalRecHitCollections.h"
 
-namespace EgAmbiguityTools {
+namespace electronAlgos {
   // for clusters
   float sharedEnergy(reco::CaloCluster const& clu1,
                      reco::CaloCluster const& clu2,
@@ -26,6 +26,6 @@ namespace EgAmbiguityTools {
   bool isBetter(reco::GsfElectron const&, reco::GsfElectron const&);
   bool isInnerMost(reco::GsfElectron const&, reco::GsfElectron const&);
 
-}  // namespace EgAmbiguityTools
+}  // namespace electronAlgos
 
 #endif
--- a/RecoEgamma/EgammaElectronAlgos/interface/EgAmbiguityTools.h
+++ b/RecoEgamma/EgammaElectronAlgos/interface/EgAmbiguityTools.h
@@ -24,7 +24,7 @@ namespace egamma {
 
   // electrons comparison
   bool isBetterElectron(reco::GsfElectron const&, reco::GsfElectron const&);
-  bool isInnerMostElectron(reco::GsfElectron const&, reco::GsfElectron const&);
+  bool isInnermostElectron(reco::GsfElectron const&, reco::GsfElectron const&);
 
 }  // namespace egamma
 
--- a/RecoEgamma/EgammaElectronAlgos/interface/EgAmbiguityTools.h
+++ b/RecoEgamma/EgammaElectronAlgos/interface/EgAmbiguityTools.h
@@ -7,7 +7,7 @@
 #include "DataFormats/GsfTrackReco/interface/GsfTrackFwd.h"
 #include "DataFormats/EcalRecHit/interface/EcalRecHitCollections.h"
 
-namespace electronAlgos {
+namespace egamma {
   // for clusters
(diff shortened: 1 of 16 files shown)


perrotta:
please test

RecoEgamma/EgammaElectronAlgos/interface/EgAmbiguityTools.h
```diff
@@ -7,7 +7,7 @@
 #include "DataFormats/GsfTrackReco/interface/GsfTrackFwd.h"
 #include "DataFormats/EcalRecHit/interface/EcalRecHitCollections.h"
 
-namespace EgAmbiguityTools {
+namespace electronAlgos {
```
slava77: CMSSW style for namespaces is to use lower case. This change seems like an opportunity to conform to it.

slava77:
> Optimally, the formerly private functions should live in the anonymous namespace, but I was a bit reluctant to do that because it would cause huge chunks of code to shift indentation without actual changes. Is there maybe a way to put functions in the anonymous namespace without enclosing them with `namespace { }` and therefore causing an additional layer of indentation?

please check that the symbols for these methods are local in your build of the library

guitargeek:
Hi Slava, I checked the symbols with
```
readelf -Ws $CMSSW_BASE/lib/slc7_amd64_gcc820/libRecoEgammaEgammaElectronAlgos.so
```
Unfortunately, the formerly private functions that are just not declared in the header file at this moment still show up with entries like this:
```
   435: 0000000000054e50    69 FUNC    GLOBAL DEFAULT   12 _Z13fEtaEndcapBadd
```
On the other hand, if I put them in an anonymous namespace, they don't show up in the list at all, which is very nice.

Here is the worst thing about the current PR I noticed now: If I create in some other package a `cc` file that tries to use one of the formerly private functions, it can! It just needs to declare them before:
```C++
double fEtaBarrelBad(double scEta);
double a = fEtaBarrelBad(1.0);
```
will compile and link with no problem. So maybe there is no way around the anonymous namespace if we want to get this right?

slava77:
> will compile and link with no problem. So maybe there is no way around the anonymous namespace if we want to get this right?

apparently.

slava77:
@cmsbuild please test

RecoEgamma/EgammaElectronAlgos/interface/EgAmbiguityTools.h
slava77: this is still pending

guitargeek: Sorry I did not realize this was a request, I thought it was just a statement! I actually thought that just the first letter has to be lower case, but then I remember it wrong. So it should be all lower case and no spaces? That means `electronalgos`?

slava77: would ```electron```  look OK for you, it's shorter?
perhaps even ```egamma``` can work unambiguously.

guitargeek: Good suggestions! I will go for `egamma`, because some functions don't actually take electrons but other egamma objects like superclusters. Also, I think this sets a good precedent because IMHO the concept of each POG having it's namespace makes sense to me.

guitargeek:
Hi @slava77, finally I went for your proposal of taking the `egamma` namespace and adapted some of the function names to make clear they are dealing with electrons.

slava77:
@cmsbuild please test

slava77:
+1

for https://github.com/cms-sw/cmssw/pull/28282	68d3b63 

fabiocos:
+1

PR #28286, opened 2019-10-28
[PY3] unit test fixed for Aligenmnet

smuzaffar:
These changes are needed to fix unit tests for Alignment for python3
- cPickle in python3 now comes from _pickle
- set time out to 3600 s (1h) instead of sys.maxsize which is like forever
- ConfigParser in python is renamed to configparser

### DIFF
--- a/Alignment/CommonAlignment/scripts/tkal_create_file_lists.py
+++ b/Alignment/CommonAlignment/scripts/tkal_create_file_lists.py
@@ -12,7 +12,10 @@ import math
 import bisect
 import random
 import signal
-import cPickle
+if sys.version_info[0]>2:
+  import _pickle as cPickle
+else:
+  import cPickle
 import difflib
 import argparse
 import functools
@@ -363,14 +366,14 @@ class FileListCreator(object):
         for d in self._datasets: print_msg("\t"+d)
         print_msg("This may take a while...")
 
-        result = pool.map_async(get_events_per_dataset, self._datasets).get(sys.maxsize)
+        result = pool.map_async(get_events_per_dataset, self._datasets).get(3600)
         self._events_in_dataset = sum(result)
 
-        result = pool.map_async(get_max_run, self._datasets).get(sys.maxsize)
+        result = pool.map_async(get_max_run, self._datasets).get(3600)
         self._max_run = max(result)
 
-        result = sum(pool.map_async(get_file_info, self._datasets).get(sys.maxint), [])
-        files = pool.map_async(_make_file_info, result).get(sys.maxint)
+        result = sum(pool.map_async(get_file_info, self._datasets).get(3600), [])
+        files = pool.map_async(_make_file_info, result).get(3600)
         self._file_info = sorted(fileinfo for fileinfo in files)
 
         self.rereco = any(len(fileinfo.runs)>1 for fileinfo in self._file_info)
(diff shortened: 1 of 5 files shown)


smuzaffar:
please test

Alignment/CommonAlignment/scripts/tkal_create_file_lists.py
```diff
@@ -363,14 +366,14 @@ def _request_dataset_information(self):
         for d in self._datasets: print_msg("\t"+d)
         print_msg("This may take a while...")
 
-        result = pool.map_async(get_events_per_dataset, self._datasets).get(sys.maxsize)
+        result = pool.map_async(get_events_per_dataset, self._datasets).get(3600)
```
fabiocos: @smuzaffar ```sys.maxint``` has gone in python3, but isn't ```sys.maxsize``` still available, according to https://docs.python.org/3/library/sys.html ? What is the rationale behind this hardcoded value?

davidlange6: ints are gone in python3. so maxsize is available but does not give the same answer as in python2. 

The error is
OverflowError: timestamp too large to convert to C _PyTime_t (which instead is a 64-bit int in C)

but perhaps 2147483647 would conserve the old behavior?

smuzaffar: @fabiocos , sys.maxint and sys.maxsize are used here to set a timeout (i.e. 9223372036854775807 sec).  There is no need to set such long timeouts.

fabiocos: @davidlange6 according to the documentation ```sys.maxsize``` is now a variable of type ```Py_ssize_t``` so not what is desired here, I agree

fabiocos: @smuzaffar this argument is more related to the functionality of the unit test I would say, waiting for an infinite time is not what is desired

fabiocos:
@christopheralanwest @tlampen @pohsun the changes look technical, please check, I would like to integrate this after pre11 is built

tlampen:
+1

davidlange6:
if 3600 is 3600 seconds as a max limit for this command, then it seems sufficiently large to allow a set of das queries to run.

smuzaffar:
yes it is in seconds. I leave it to alca developers to set it to some reasonable time. For IBs, we kill unit test after 30 mins anyway :-)

mmusich:
> yes it is in seconds. I leave it to alca developers to set it to some reasonable time. For IBs, we kill unit test after 30 mins anyway :-)

This file (which is tested in a unit test) is also used for the actual aggregation of the data for running the alignment. Maybe waiting for an infinite amount of time is not OK for a unit test, but on the other hand I am not sure if the script is supposed to run for > 1h in real life applications. @adewit @connorpa. 
Maybe that parameter can be made configurable and set to 1h for the unit test.

davidlange6:
presumably one can set this to a big enough number for all usecases - however i guess cms actually doesn't want scripts running hours worth of das queries either....

fabiocos:
@mmusich @connorpa @adewit @christopheralanwest can we clarify what is the real life need for this timeout? For the unit test purpose the solution by @smuzaffar is clearly ok, in case we have different use cases we should clarify a meaningful value, possibly before integrating this PR

mmusich:
> can we clarify what is the real life need for this timeout?

The feedback I got in the meanwhile is that it usually takes few minutes to run, so 1h timeout sounds about reasonable also for the offline use-case.

fabiocos:
+1

PR #28288, opened 2019-10-28
[PY3] run dablooms tests for python2 only

smuzaffar:
PyDabloom is only avaiolable for python2. Building it for python3 fails. Looks like this project is not moving to python3 any time soon, there is no development since Feb 2016 ( https://github.com/bitly/dablooms ). We use dablooms for Utilities/StaticAnalyzers only ( https://github.com/cms-sw/cmssw/blob/master/Utilities/StaticAnalyzers/BuildFile.xml#L2 ). @gartung can you please check if we can drop this dependency?

For now, I have disabled this unit test for python3 IBs.

### DIFF
--- a/PhysicsTools/PythonAnalysis/test/BuildFile.xml
+++ b/PhysicsTools/PythonAnalysis/test/BuildFile.xml
@@ -40,4 +40,4 @@
 <iftool name="py2-cx-oracle">
   <test name="testCxOracle" command="python -c 'import cx_Oracle'"/>
 </iftool>
-<test name="testPYDablooms" command="python -c 'import pydablooms'"/>
+<test name="testPYDablooms" command="$(LOCALTOP)/src/PhysicsTools/PythonAnalysis/test/test-pydablooms"/>
--- a/PhysicsTools/PythonAnalysis/test/test-pydablooms
+++ b/PhysicsTools/PythonAnalysis/test/test-pydablooms
@@ -1,0 +1,8 @@
+#!/usr/bin/env python
+from __future__ import print_function
+import sys
+if sys.version_info[0]>2:
+  print("Dabloom is not available for Python 3")
+else:
+  import pydablooms
+  


smuzaffar:
please test

gartung:
The package is used to generate a bloom filter that is used by the static analyzer to check for safe class names. Removing the dependency would require removing the safe class name filter. @Dr15Jones is it OK to remove the safe class name filter?

Dr15Jones:
> @Dr15Jones is it OK to remove the safe class name filter?

seems like a bad idea. Guess we need some sort of resolution for dealing with the filter in python 3.

gartung:
I think I have found a suitable replacement
https://pybloomfiltermmap3.readthedocs.io/en/latest/
It has a python2 implementation as well that can be packaged.

gartung:
That package does not include linkable library. It does mention compiling a pyx file with Cython, but I don't see how that could be linked into the StaticAnayzer library.

davidlange6:
we need a c++ callable function or both python and c++ callable functions?

gartung:
Both. The python script generates the bloom.bin file and it is loaded in a c++ callable function.

davidlange6:
is the overhead of just recreating the bloom.bin content at run time initialization high? 
eg here is an easy to integrate c++ library https://www.partow.net/programming/bloomfilter/index.html
(I can't find the classes.txt inputs that creates the bloom.bin in the python script)

gartung:
So the package can be installed with pip for python2 or python3. Is there a way to call a python function from inside a c++ library?

gartung:
Classes.txt is generated by the class name dumper in a separate run.

gartung:
It's a header only implementation.

davidlange6:
i mean to avoid producing bloom.bin altogether and to build the bloom filter on the fly in CmsSupport.cpp where currently bloom.bin is read in.

gartung:
I will work on updating the generation and usage of bloom.bin using the Partow bloom filter implementation. This will replace py-dabloom package with a partow-bloom(sp?) package and the corresponding tool file.

gartung:
The generation step has to be separate from the using step.

gartung:
The generation step is done in the same Jenkins job where it is used. I don't think bloom.bin in the repo has been update in a long time.

davidlange6:
why? Too resource intensive?

gartung:
It is used with every invocation of the static analyzer which is per file.

davidlange6:
and there is a major factor saved between reading classes.txt and bloom.bin I guess. Got it.

gartung:
At the time it was implemented I was maintaining the classes.txt in the repo. Giulio though the bloom.bin was a better way to store it. This was before the Jenkins job was set up to generate and use classes.txt(bloom.bin) in the same job.

fabiocos:
@smuzaffar @Dr15Jones @davidlange6 @gartung this PR is not directly addressing the replacement discussed in the thread, but it is implementing a simple workaround for the short term to the the unit test working. I wonder whether we would prefer to leave it broken waiting for a new solution, or track it with an issue and integrate this workaround. What could be the expected timescale?

gartung:
I would propose removing the test completely. That is what I have in the PR I am working on. I will submit a cmsdist PR with the new package to be used by Utilities/StaticAnalyzers.

smuzaffar:
closing in favor of #28315

PR #28277, opened 2019-10-25
Fix gcc 9 warning in L1Trigger/GlobalTriggerAnalyzer

mrodozov:
#### PR description:

Fixes gcc9 warning in L1Trigger/GlobalTriggerAnalyzer

#### PR validation:

Compiles without the warning
I could've put only 

`return ~((~(unsigned)0) << length);`

decided to be more expresive

PR #28291, opened 2019-10-28
Fix undefined behavior in L1GtPatternWriter

Dr15Jones:
#### PR description:

Bit shifting a negative number is undef. The starting number is now an unsigned value instead. This fixes a gcc 9 warning.

#### PR validation:

The warning is no longer present in gcc9 builds.
Using compiler explorer to test the change, the same value is returned for both gcc and clang after the change.
https://godbolt.org/z/oHP_EL


Dr15Jones:
please test

Dr15Jones:
Solves same problem as #28277

fabiocos:
@rekovic @benkrikler the change looks technical and straightforward, please have a look, I will integrate it after pre11 is built

fabiocos:
+1

fabiocos:
merge

PR #28297, opened 2019-10-28
DQM: Remove DQMServices/Diagnostic

schneiml:
#### PR description:

This looks like an old variation of Historic DQM. Since the new HDQM
managed by the DQM group since more than a year lives outside CMSSW,
and Tracker HDQM did for a long time before, I assume this is unused.

I can't tell if anybody is using maybe some of the funcitonality in that
package, but removing it is probably the best way to find out.

The `SiStripHistoricInfoClient` was removed as it depends on `DQMServices/Diagnostic` but does not add much in terms of functionality in its main functionality. There is a large amount of code in `test/` in that package, but if it is all related to this historic DQM it is probably no longer needed either.

Same goes for some modules in `DQMOffline/CalibTracker`.

#### PR validation:
None; assuming package is unused.


jfernan2:
please test

schneiml:
I completely overlooked `SiStripHistoricInfoClient`, which is a bunch of very thin wrappers (not really doing anything) around `DQMServices/Diagnostic`. I assume it is safe to remove with the same reasoning as before.

schneiml:
please test

hope I got all the dependent files now, it's not that many after all...

schneiml:
+1

No confirmation from Tracker yet, in case they are still using specific modules we'll need to bring them back. `DQMServices/Diaggnostic` will be removed either ways.

mmusich:
> No confirmation from Tracker yet, 

given that a message was sent yesterday should not be so surprising :)

schneiml:
@mmusich yep, and I intentionally phrased that as "information", not "question", assuming we'd move forward today anyways...

mmusich:
>  assuming we'd move forward today anyways...

the underlying technology is interesting. Unfortunately it was an effort discontinued since Run 1 (some information on the workflow is here: https://twiki.cern.ch/twiki/bin/viewauth/CMS/DQMHistoryClient).

at any rate it looks to me there is nothing currently used, but I would like @arossi83 and @sroychow to confirm.

schneiml:
> the underlying technology is interesting.

I agree, and that is actually a common theme in many of these "Remove ..." PRs. Many of those things are good ideas an could be very useful, but ended up being not used. I am very pragmatic here: If (almost) nobody uses it, it is thrown out, since we barely afford maintaining the commonly used things in the long term. But many good ideas hide in this old code...

fabiocos:
@arossi83 @sroychow could you please confirm that this code may be safely removed?

sroychow:
From my side it is okay to go ahead with this PR since nothing is used currently.

fabiocos:
+1

PR #28299, opened 2019-10-29
Take python exceptions from builtins

guitargeek:
#### PR description:

Today I wanted to use the `edmDumpEventContent` script with python3, which can be easily done by automatically updating the code with `python-modernize` except for one hiccup: the `import exceptions` line in *PhysicsTools/PythonAnalysis/python/cmstools.py*. Following the information in this post [1], the exceptions can be just taken from the builtins instead of the exceptions module in python 2 as well, which makes the code python 3 compatible. There was also another spot in CMSSW where the former `exceptions` module was imported, so I changed this one too.

[1] https://stackoverflow.com/questions/27030933/python-2s-exceptions-module-is-missing-in-python3-where-did-its-contents-go

#### PR validation:

I can use `edmDumpEventContent` with python 3 when I take this PR and run `python-modernize -n -w .` on top of it.

No backport intended.

### DIFF
--- a/PhysicsTools/PythonAnalysis/python/cmstools.py
+++ b/PhysicsTools/PythonAnalysis/python/cmstools.py
@@ -8,7 +8,6 @@ from __future__ import print_function
 from builtins import range
 import re
 import ROOT
-import exceptions
 import six
 import sys
 ### define tab completion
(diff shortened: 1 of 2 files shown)


perrotta:
please test

slava77:
+1

for https://github.com/cms-sw/cmssw/pull/28299	6c10fd7 
- technical; as described. The only reco file affected is a test file RecoLocalCalo/EcalRecProducers/test/compare_csv.py

santocch:
+1

fabiocos:
@guitargeek thanks for this update, I need to check several other scripts not covered by regular tests in IB

fabiocos:
+1

PR #28304, opened 2019-10-29
Strip Clusterizer from Zero Suppressed Readout

VinInn:
see

https://indico.cern.ch/event/841666/contributions/3583433/attachments/1918082/3171950/SpeedingStripCluster.pdf
 and

https://indico.cern.ch/event/841668/contributions/3612166/attachments/1934026/3204280/SpeedingStripClusterUpdate.pdf

to run it just
```
from RecoLocalTracker.SiStripClusterizer.SiStripClustersFromZSRawFacility_cfi import SiStripClustersFromZSRawFacility as _siStripClusters
process.siStripClusters = _siStripClusters.clone()
```


slava77:
@pieterdavid @alesaggio 
please check the code and clarify if the changes are OK for the DPG or if some changes need to be made

jfernan2:
@VinInn Despite I would not expect differences in DQM histograms due to the minimal changes in DQM//SiStripMonitorTrack  and Validation/TrackerRecHits, the bin by bin test reflects changes in number of entries and rounding of mean/RMS values for some workflows, and in folders not directly related to Tracking/SiStrip like recoTau, Egamma, JetMET, Muon and PF.
The changes are not significant but it would be nice to understand them. Is there anything else touched beyond the improvement of the Strip Cluster Production (for HLT)?

pieterdavid (commented):
There are two kinds of changes in this PR:
- changes to speed up the existing clusterizer: these should be purely technical, so the differences in the comparisons need to be understood (could it be the fact that instead of dividing by the gain, `1./gain` is stored as a float and multiplied by? The result is truncated, so I'd be surprised, but that's the only change that at first sight could change the output), otherwise this should be fine from the DPG side.
- a new producer that makes clusters directly from the raw buffers (for the two 8-bit zero-suppressed modes with the top two bits truncated). This is not included in any workflow yet and needs further testing and validation before it will be, so it can be merged before everything is final to facilitate that. I added a few more detailed comments and questions for this.

RecoLocalTracker/SiStripClusterizer/plugins/ClustersFromZSRawAlgorithm.h
```diff
@@ -0,0 +1,94 @@
...
+public:
+  using Det = StripClusterizerAlgorithm::Det;
+
+  template <typename OUT, bool CLUS_CHARGE_CUT = false, bool NOISE_BY_STRIP_CUT = false>
+  void clustersFromZS(
+      uint8_t const* data, int offset, int lenght, uint16_t stripOffset, Det const& det, OUT& out) const {
+    constexpr int clusterNoiseCutFactor =
+        4;                                  // 10^2/5^2 : 10 because noise is store *10; 5 because os a 5 sigma S/N cut
+    constexpr int stripNoiseCutFactor = 5;  // 10/2
+    constexpr float clusterChargeCut = 1200.0f;  // tuned by VI
```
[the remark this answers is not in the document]
VinInn: Configurable only if somebody proves that it may be needed to be changed at runtime for the same release in different workflows

RecoLocalTracker/SiStripClusterizer/plugins/ClustersFromZSRawAlgorithm.h
```diff
@@ -0,0 +1,94 @@
...
+
+#include "RecoLocalTracker/SiStripClusterizer/interface/StripClusterizerAlgorithm.h"
+#include "DataFormats/SiStripCluster/interface/SiStripClusterTools.h"
+
+// produce SiStripCluster direclty out of those in ZeroSupressed Raw data
+class ClustersFromZSRawAlgorithm final : public StripClusterizerAlgorithm {
+public:
+  using Det = StripClusterizerAlgorithm::Det;
+
+  template <typename OUT, bool CLUS_CHARGE_CUT = false, bool NOISE_BY_STRIP_CUT = false>
```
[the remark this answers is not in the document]
VinInn: oops, was not intended.
My proposal IS to cut on CCC at 1200.
the clusterizer will be indeed faster (10 even 20% less clusters)
The S/N >5 is enabled hardwired by default (line 47)

VinInn:
@jfernan2 , there are few changes here and there. I do not exclude possible differences due to float->int->float transformations

RecoLocalTracker/SiStripClusterizer/plugins/ClustersFromZSRawAlgorithm.h
```diff
@@ -0,0 +1,94 @@
+#ifndef RecoLocalTrackerSiStripClusterizerClustersFromZSRawAlgorithm_H
+#define RecoLocalTrackerSiStripClusterizerClustersFromZSRawAlgorithm_H
+
+#include "RecoLocalTracker/SiStripClusterizer/interface/StripClusterizerAlgorithm.h"
+#include "DataFormats/SiStripCluster/interface/SiStripClusterTools.h"
+
+// produce SiStripCluster direclty out of those in ZeroSupressed Raw data
+class ClustersFromZSRawAlgorithm final : public StripClusterizerAlgorithm {
```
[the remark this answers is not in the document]
VinInn: I have no problem with that (it is irrelevant from a compilation/linking point of view) provided that central code reviewers agree.

RecoLocalTracker/SiStripClusterizer/plugins/ClustersFromZSRawAlgorithm.h
```diff
@@ -0,0 +1,94 @@
...
+      int clusSize = 0;
+      //
+      auto saveCluster = [&]() {
+        // save cluster
+        // do not cut if extendable
+        endStrip = firstStrip + clusSize;
+        if (extend)
+          out.back().extend(adc, adc + clusSize);
+        else {
+          if (endStrip % 128 == 0 || clusterNoiseCutFactor * sumRaw * sumRaw > noise2) {
```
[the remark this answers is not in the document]
pieterdavid: Is this to apply the cut only after merging clusters that go across APV boundaries (which is not implemented yet, as far as I understand)? Shouldn't the cut in that case also be disabled for clusters that begin at the first strip of an APV?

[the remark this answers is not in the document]
VinInn: Why?  clusters that begin at the first strip of an APV are merged with the previous one on the fly.
In any case in reco we cut on CCC at >1400 with is higher than the S/N cut at 5.

mmusich:
adding @robervalwalsh in the watchers as per his request.

VinInn:
for reason I do not understand I do not receive notification by e-mail on this PR

RecoLocalTracker/SiStripClusterizer/plugins/ClustersFromZSRawAlgorithm.h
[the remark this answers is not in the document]
mmusich: @VinInn what is the gain in making it not configurable? We cannot be certain there won't ever be again a problem with CCC.

RecoLocalTracker/SiStripClusterizer/plugins/ClustersFromZSRawAlgorithm.h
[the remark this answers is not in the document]
mmusich: @slava77 @perrotta please comment.

RecoLocalTracker/SiStripClusterizer/plugins/ClustersFromZSRawAlgorithm.h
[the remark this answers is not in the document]
pieterdavid: I see, sorry, I missed that.

RecoLocalTracker/SiStripClusterizer/plugins/ClustersFromZSRawAlgorithm.h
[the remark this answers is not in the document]
VinInn: is faster and in my opinion less error prone.
Im personally against a long list of configurable about internal details of algorithms that only experts understand.
If something goes wrong a new release will be needed and at that point one can do whatever needed.

Anyhow this code will belong to TRK-DPG. Up to you to modify as it better suites the use cases that you will decide to support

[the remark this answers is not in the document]
VinInn: Let me be more precise:
1200 is the lowest CCC that makes sense (see my first presentation): it roughly corresponds to the 5 S/<N> cut
(as a matter of fact we completely forgot to study the effect of S/<N> for the mitigation of the "HIP" issue in 2016).
The only alternative (besides rising to the default tracking one) is to remove it.
And this is trivially achievable using two versions of the algorithm as apply-CCC is a template argument.

jfernan2:
+1
from DQM side, changes understood as differences due to float->int->float transformations

RecoLocalTracker/SiStripClusterizer/plugins/ClustersFromZSRawProducer.cc
```diff
@@ -0,0 +1,300 @@
...
+    std::unique_ptr<sistrip::FEDBuffer> buffer;
+
+    // Retrieve FED raw data for given FED
+    const FEDRawData& rawData = rawColl.FEDData(fedId);
+
+    // Check on FEDRawData pointer
+    if
+      UNLIKELY(!rawData.data()) {
+        if (edm::isDebugEnabled()) {
+          edm::LogWarning("ClustersFromZSRawProducer") << " NULL pointer to FEDRawData for FED id " << fedId;
```
slava77: ClustersFromRawProducer is using some standardized category names. 
it seems practical here to also use the same.
OTOH, if this logging is not essential for some downstream runtime analysis and just for debugging, perhaps it's better to change these two lines to one LogDebug

RecoLocalTracker/SiStripClusterizer/plugins/ClustersFromZSRawProducer.cc
```diff
@@ -0,0 +1,300 @@
...
+      produces<edmNew::DetSetVector<SiStripCluster>>();
+    }
+
+    static void fillDescriptions(edm::ConfigurationDescriptions& descriptions) {
+      edm::ParameterSetDescription desc;
+
+      desc.add<edm::InputTag>("ProductLabel", edm::InputTag("rawDataCollector"));
+      desc.add<bool>("DoAPVEmulatorCheck", false);
+
+      descriptions.add("SiStripClustersFromZSRawFacility", desc);
```
slava77: is the name "Facility" a historical reference to the on-demand case?
A default module instance is generally preferable ```siStripClusterizerFromZSRaw```

RecoLocalTracker/SiStripClusterizer/plugins/ClustersFromZSRawProducer.cc
```diff
@@ -0,0 +1,300 @@
...
+
+    void produce(edm::Event& ev, const edm::EventSetup& es) override {
+      initialize(es);
+
+      // get raw data
+      edm::Handle<FEDRawDataCollection> rawData;
+      ev.getByToken(productToken_, rawData);
+
+      auto output = std::make_unique<edmNew::DetSetVector<SiStripCluster>>();
+      assert(!output->onDemand());
```
slava77: isn't this superfluous given the line above?

RecoLocalTracker/SiStripClusterizer/plugins/ClustersFromZSRawProducer.cc
```diff
@@ -0,0 +1,300 @@
...
+    }  // end loop over dets
+  }
+
+  void ClusterFiller::fill(StripClusterizerAlgorithm::Det const& det,
+                           StripClusterizerAlgorithm::output_t::FastFiller& record) {
+    incReady();
+
+    auto idet = record.id();
+
+    COUT << "filling " << idet << std::endl;
```
slava77: idet is not used other than this cout, perhaps just replace it here with ```record.id()```

RecoLocalTracker/SiStripClusterizer/plugins/ClustersFromZSRawProducer.cc
```diff
@@ -0,0 +1,300 @@
...
+
+      // check channel
+      const uint8_t fedCh = conn->fedCh();
+
+      if
+        UNLIKELY(!buffer->channelGood(fedCh, doAPVEmulatorCheck)) {
+          if (edm::isDebugEnabled()) {
+            std::ostringstream ss;
+            ss << "Problem unpacking channel " << fedCh << " on FED " << fedId;
+            edm::LogWarning("ClustersFromZSRawProducer") << ss.str();
```
slava77: ```sistrip::mlRawToCluster_``` category?
also, the intermediate ostringstream instance looks unnecessary

RecoLocalTracker/SiStripClusterizer/plugins/ClustersFromZSRawProducer.cc
```diff
@@ -0,0 +1,300 @@
...
+    }  // end loop over conn
+
+    incAct();
+
+    if (!record.empty())
+      incNoZ();
+
+    //    COUT << "filled " << record.size() << std::endl;
+    //    for (auto const& cl : record)
+    //      COUT << cl.firstStrip() << ',' << cl.amplitudes().size() << std::endl;
```
slava77: is this commented out code needed?

RecoLocalTracker/SiStripClusterizer/src/StripClusterizerAlgorithm.cc
```diff
@@ -155,42 +156,70 @@ void StripClusterizerAlgorithm::initialize(const edm::EventSetup& es) {
...
+}
+
+void StripClusterizerAlgorithm::fillDets() {
+  dets.resize(detIds.size());
+  for (int i = 0, ni = detIds.size(); i < ni; ++i) {
+    auto id = detIds[i];
+    auto& det = dets[i];
+    det.ind = i;
+    det.detId = id;
+    det.noiseRange = noiseHandle->getRangeByPos(indices[det.ind].ni);
```
slava77: ```const atuo& index = indices[det.ind]``` and used it later may be more readable

RecoLocalTracker/SiStripClusterizer/src/StripClusterizerAlgorithm.cc
```diff
@@ -155,42 +156,70 @@ void StripClusterizerAlgorithm::initialize(const edm::EventSetup& es) {
...
+    for (int ic = 0; ic < s; ++ic)
+      det.m_weight[ic] = 1.f / (*(gainRange.first + ic));
+    // fill average noise"
+    for (int ic = 0; ic < s; ++ic) {
+      auto fs = 128 * ic;
+      auto ns = 0;
+      int an = 0;
+      for (auto is = fs; is < fs + 128; ++is) {
+        auto n = det.rawNoise(is);
+        if (n > 0 && n < 500) {
```
slava77: please clarify why 500 is the limit

RecoLocalTracker/SiStripClusterizer/interface/StripClusterizerAlgorithm.h
```diff
@@ -24,25 +24,28 @@ class StripClusterizerAlgorithm {
   // state of detID
   struct Det {
     bool valid() const { return ind != invalidI; }
-    float noise(const uint16_t& strip) const { return SiStripNoises::getNoise(strip, noiseRange); }
-    float gain(const uint16_t& strip) const { return SiStripGain::getStripGain(strip, gainRange); }
-    bool bad(const uint16_t& strip) const { return quality->IsStripBad(qualityRange, strip); }
+    uint16_t rawNoise(const uint16_t strip) const { return SiStripNoises::getRawNoise(strip, noiseRange); }
+    float noise(const uint16_t strip) const { return SiStripNoises::getNoise(strip, noiseRange); }
+    float weight(const uint16_t strip) const { return m_weight[strip / 128]; }
```
slava77: shouldn't this be ```aveWeight``` to denote that this is per channel?
Well, looking at
https://github.com/cms-sw/cmssw/blob/4c0913fabe8081975cee9b0683d6906338a7da3f/CondFormats/SiStripObjects/interface/SiStripApvGain.h#L78-L81
it was already the case in a less explicit way.
So, this name change will make it more explicit.

RecoLocalTracker/SubCollectionProducers/src/HITrackClusterRemover.cc
```diff
@@ -319,8 +319,7 @@ void HITrackClusterRemover::process(OmniClusterRef const &ocluster, SiStripDetId
     return;
   if (!fromTrack) {
     int clusCharge = 0;
-    for (std::vector<uint8_t>::const_iterator iAmp = cluster->amplitudes().begin(); iAmp != cluster->amplitudes().end();
-         ++iAmp) {
+    for (auto iAmp = cluster->amplitudes().begin(); iAmp != cluster->amplitudes().end(); ++iAmp) {
       clusCharge += *iAmp;
```
slava77: ```for (auto amp : cluster->amplitudes()) clusCharge += amp;``` or even std::accumulate

slava77 (commented):
sorry, some other priorities/PRs delayed review of this.

RecoLocalTracker/SiStripClusterizer/plugins/ClustersFromZSRawAlgorithm.h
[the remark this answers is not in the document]
slava77: it's already in the plugins directory, which means the interface exposure is limited. So, I have no strong feeling about merging. Perhaps it's even worth thinking of simply using ClustersFromZSRawAlgorithm as one of the options in the ClustersFromRawProducer

CondFormats/SiStripObjects/interface/FedChannelConnection.h
```diff
@@ -226,4 +227,14 @@ inline void FedChannelConnection::fedCh(uint16_t& fed_ch) { fedCh_ = fed_ch; }
 inline void FedChannelConnection::fedCrate(uint16_t& fed_crate) { fedCrate_ = fed_crate; }
 inline void FedChannelConnection::fedSlot(uint16_t& fed_slot) { fedSlot_ = fed_slot; }
 
+inline uint16_t FedChannelConnection::apvPairNumber() const {
+  if (apv0_ == 32 || apv1_ == 33) {
+    return 0;
+  } else if (apv0_ == 34 || apv1_ == 35) {
+    return 1;
+  } else {  // if (apv0_ == 36 || apv1_ == 37) {
+    return nApvPairs_ - 1;
```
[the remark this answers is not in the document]
VinInn: better be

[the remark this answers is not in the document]
VinInn: this is production code, not debugging of the database.

RecoLocalTracker/SiStripClusterizer/plugins/ClustersFromZSRawProducer.cc
VinInn: whatever. It is historical

RecoLocalTracker/SiStripClusterizer/plugins/ClustersFromZSRawProducer.cc
```diff
@@ -0,0 +1,300 @@
...
+
+    // construct FEDBuffer
+    try {
+      buffer.reset(new sistrip::FEDBuffer(rawData.data(), rawData.size()));
+      if
+        UNLIKELY(!buffer->doChecks(false))
+      throw cms::Exception("FEDBuffer") << "FED Buffer check fails for FED ID" << fedId << ".";
+    } catch (const cms::Exception& e) {
+      if (edm::isDebugEnabled()) {
+        edm::LogWarning("ClustersFromZSRawProducer")
```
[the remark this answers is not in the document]
VinInn: no clue, old code

RecoLocalTracker/SiStripClusterizer/plugins/ClustersFromZSRawProducer.cc
```diff
@@ -0,0 +1,300 @@
...
+            << "Exception caught when creating FEDBuffer object for FED " << fedId << ": " << e.what();
+      }
+      return std::unique_ptr<sistrip::FEDBuffer>();
+    }
+
+    /*
+    // dump of FEDRawData to stdout
```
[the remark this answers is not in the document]
VinInn: is debug, can be protected by #ifdef

RecoLocalTracker/SiStripClusterizer/plugins/ClustersFromZSRawProducer.cc
VinInn: most probably 20 year old. not me to change it

RecoLocalTracker/SiStripClusterizer/plugins/ClustersFromZSRawProducer.cc
VinInn: all assert are by definition superfluous

RecoLocalTracker/SiStripClusterizer/plugins/ClustersFromZSRawProducer.cc
VinInn: the compiler will take care of it

RecoLocalTracker/SiStripClusterizer/plugins/ClustersFromZSRawProducer.cc
VinInn: old code.

RecoLocalTracker/SiStripClusterizer/src/StripClusterizerAlgorithm.cc
VinInn: maybe

RecoLocalTracker/SubCollectionProducers/src/HITrackClusterRemover.cc
VinInn: this code is there since 20 years, not me to change

VinInn:
I think that this PR is too controversial. I touches a lot of old code that works as is.
In any case I started this with a wrong idea of how the ZS was working.

In my opinion what needs to be done is more profound: the ZS FED code has to be changed to produce individual cluster and one has to implement a new ZSreformatter that produces clusters as simple SoA of reduced content, no copy of the digi. if spitting is required to be done on the fly.

I leave all this for some R&D in the tracker DPG.

closing, sorry for the noise and all the time wasted by you all.

slava77:
> closing, sorry for the noise and all the time wasted by you all.

IIUC, some parts of this PR are speeding up the current production version of the code. It may be useful to still have these in.
Thank you.

pieterdavid:
> IIUC, some parts of this PR are speeding up the current production version of the code. It may be useful to still have these in.

Indeed, @vischia and I will have a closer look at the differences and submit these changes as a new PR

PR #28305, opened 2019-10-29
call close with the option to shutdown workers

dan131riley:
#### PR description:

On close, ROOT defaults to detaching "remote sessions", which apparently includes workers on the local node.  This has been observed to occupy resources on the build servers.  This PR calls `TProof::Close("S")`, which should tell the workers to shutdown rather than detach.

#### PR validation:

Ran the standard tests and observed that no worker processes were left behind.  This PR should have no impact on anything outside the scope of the TFWLiteSelector unit tests.


Dr15Jones:
please test

Dr15Jones:
+1

fabiocos:
+1

PR #28312, opened 2019-10-30
Make cond::persistency::KeyList const thread-safe

Dr15Jones:
#### PR description:

KeyList change
- Avoid need for updating an internal cache to avoid thread-safety issues. This was accomplished by avoiding the _load()_ then _get()_ call chain and instead have specialized functions which combine both.
- internal keys are only set by CondDBESSource Proxy.
- All uses can not call `const` interfaces.

SessionImpl change
- Need to use a mutex since the `SessionImpl` can be called concurrently from both CondDBESSource and from a user of the KeyList obtained from the EventSetup.

#### PR validation:

Ran workflow 25.0 from the IB using 4 threads and the jobs completed successfully.

### DIFF
--- a/CondCore/CondDB/interface/KeyList.h
+++ b/CondCore/CondDB/interface/KeyList.h
@@ -54,7 +54,7 @@ namespace cond {
       /** Retrieve the item associated with the key directly from the DB.
           The function is non-const since it is not thread-safe. */
       template <typename T>
-      std::shared_ptr<T> getUsingKey(unsigned long long key) {
+      std::shared_ptr<T> getUsingKey(unsigned long long key) const {
         auto item = loadFromDB(key);
         return deserialize<T>(item.first, item.second.first, item.second.second);
       }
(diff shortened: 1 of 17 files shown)


Dr15Jones:
please test

fabiocos:
@ggovi @christopheralanwest @pohsun @tlampen could you please have a look?

tlampen:
+1

ggovi:
+1

fabiocos:
+1

@rekovic changes to the L1T domain are restricted to one file only, and consequential to the update implied by this PR. Please have a look, I will merge the development to move forward

fabiocos:
@Dr15Jones @wddgit this PR and #28010 are unfortunately in conflict, regardless the order in which they are merged. What is your preference? In principle I am ok in merging this, for #28010 I think we are technically ok, but we need to clarify the validation strategy. 

Please let me know how do you prefer to move

Dr15Jones:
@fabiocos  #28010 is more important to the framework so I'd prefer that to be merged first.

fabiocos:
@Dr15Jones as expected this now needs to be rebased, although it does not look like very complex

fabiocos:
@Dr15Jones in order to fix the conflict: I assume you want to keep your change, removing the ```load``` method in favour of a ```setKey``` one, and keep the addition of ```init```. Please confirm

Dr15Jones:
@fabiocos I'm at CHEP so haven't had a chance yet to try to do the needed rebase. I do intend to fix the merge conflict once I get a chance.

Dr15Jones:
please test

Dr15Jones:
This was rebased on master and making the changes as suggested by @fabiocos

fabiocos:
@ggovi @christopheralanwest @tlampen could you please have a final look at this PR, after it has been rebased?

pohsun:
+1

fabiocos:
@ggovi as this was a rebase, I will consider your previous signature as valid, please comment in case. I will integrate this PR in next IB

fabiocos:
+1

fabiocos:
merge

PR #28314, opened 2019-10-30
python unittest cleanup: make them run in parallel

smuzaffar:
Clean up python unit tests
- Move all tests from `imports.sh` in to BuildFile so that they can run in parallel
- Drop separate `Theano` tests files and make use of `testTheano.sh` to run all three `Theano` tests
- removed unused `imports.py`


smuzaffar:
@fabiocos , can we get this in IB? This just make sure that we run these unit tests in parallel

santocch:
+1

fabiocos:
+1

@smuzaffar we should be able to see this in next IB

PR #28316, opened 2019-10-30
DQM: Restructure QTests

schneiml:
#### PR description:

This PR aims to clean up the way QTests are handled in DQM. Rather than spreading the logic all over the place, it is now all in a single module.
- `QualityTester` becomes a plugin in `DQMServices/Components`, handling all QTest things.
- The `DQMServices/ClientConfig` package disappears without replacement.
- Some useful parts of `ClientConfig` are rescued over to `Components`, most prominently `DQMGenericClient`.
- Some modules that used to use the XML parsing helpers from `ClientConfig` are now stubbed out (`SiPixelMonitorClient`/`SiStripMonitorClient`/`SiStripCommissioningClients`). Subsystem contacted to see if they need to be/can be ported.
- The XML parsing is now done with `boost::property_tree` rather than `xerces`. Much less code, however more strict (some of the QTest configuration XML used in production was malformed!) and probably worse error reporting.
- The QTest implementation remain in `DQMServices/Core` `QTest.[h|cc]`. However, the *only* place that uses them now is `QualityTester`.
- `QReport`s (results) remain available via the `MonitorElement`, but don't contain pointers to `QCriterion`s (the actual tests) any more.
- `QualityTester` is now a `DQMEDHarvester`, rather than a legacy module.
- The ordering `QualityTester` vs. other harvesters is not enforced by dependencies. This may need to be fixed.

#### PR validation:

Runs a simple workflow. Opening this PR for more validation, while some functionality is still missing (XML parsing for tracker clients).

Need detailed comparisons over many workflows to see if all QTests are still applied in the same way There might be XML parsing differences, and the pattern matching is different in the new code (now using POSIX `fnmatch` rather than `classlib`).


schneiml:
please test

... to see what happens.

For actual integration, I should probably factor out all the configuration changes to get the number of changed files down a bit.

schneiml:
All the failures where (for now) caused by a missing closing tag in `DQM/L1TMonitorClient/data/L1EmulatorCsctfQualityTests.xml`.

schneiml:
please test

maybe now we can get all the way to a comparison.

schneiml:
Looks like a bunch of QTests went missing with these changes -- not too surprising, I'll look at the details tomorrow. At least, the summaries seem to *mostly* still work, apart from some changes due to the missing (or maybe changed?) QTests.

schneiml:
The problem here is that most configs set the QTests to run on `endRun` (*only*), but the `DQMEDHarvester` does not run on `endRun`. So only those few that also enable endJob get produced.

I am looking for a good solution; running all Qtests in `endJob` is not good since then things depend on order of execution of the modules.

schneiml:
Now things should be better. I updated the `DQMEDHarvester` base class using the respective one form the "new DQMStore" branch, which means that also harvesting in general could be broken, but we get some better use of dependencies and more callbacks, including `dqmEndRun` (actually, I just added that one...).

The other big source of missing QTests was that we indeed want `*` to match `/` as well.

Overall, order of execution is a big mess with qtests. By running them in `endRun` we can be sure they are done by the time harvesting reads them in `endJob`, but then we might want to have Qtests on MEs booked in `endJob`, so we need to run the Qtests again. Actually, `MuonRecoTest` [1] books MEs and then expects the QTests on them to be populated immediatly. Not sure if that ever worked, it sure won't now.

In `endRun` we could enforce order by consuming and producing Tokens, but for `endJob`, we are doomed to have order encoded in sequences, afaik.

[1] https://cmssdt.cern.ch/dxr/CMSSW/source/DQMOffline/Muon/src/MuonRecoTest.cc#57

schneiml:
please test

... hope it all builds fine, the Harvesting code causes a number of compiler warnings...

schneiml:
Nothing to see here, except some new, untested code for `SiPixelActionExecutor` (which is responsible for many of the missing plots). Somehow screwed up my dev environment, saving changes here...

schneiml:
please test

rebased again, the `SiPixelActionExecutor` seems to work again, and let's see if not running qtests on endJob by default is closer to the old behaviour.

schneiml:
The latest update brings an (untested, not sure which sequence would use it) config parser  for `SiStripActionExecutor` and a bugfix that should resolve another pile of comparison failures.

schneiml:
Ok, after some long debugging, things got a bit more clear: Since the new QTests are only applied at discrete points in time (rather than "as needed", as the old implementation somehow did), we get into trouble with module ordering. Issue #28354 should explain the problem.

To work around that, I now moved the computations of `DQMGenericClient` to `endRun`, *and* enforce correct order in `endRun` using `DQMToken`s. No idea how much fallout that causes. 

Not let's do a round of tests and see how many differences remain (or which new ones appear).

Also, `SiStripConfigParser` is still untested, it seems to not run in any offline sequences.

schneiml:
please test

let's see what happens now.

schneiml:
please test

really, it should pass code-checks now.

schneiml:
The bad news: the `SiStripMonitorClient` bad modules list mechanism is broken.

The good news: it looks like *only* the bad modules list mechanism is broken.

Actually, I am mildly puzzled. The mechanism in general was working well before, only a few QTests misfiring due to mir-read configuration caused too many bad modules. Now, it seems we have too little bad modules, but the thing that really changed was the `DQMGenericClient`? or maybe the ordering dependency. I'll investigate.

schneiml:
please test

Now, commit a9a0dbc should fix the `-8` in many different workflows, which are related to some muon plots booked in harvesting. Commit 29c77f3 should fix the 100's-1000's of missing SiStrip BadModuleList entries in some workflows, except apparently it does not, due to unexpected behaviour of EDM reported in issue #28354. Let's see what happens.

schneiml:
please abort

found a way that actually works, coming soon

schneiml:
please test

now we got chances for a pretty clean result.

schneiml:
And indeed, things look much better. The *changes* that we now see need further investigation. The additional elements are as far as I remember QTests in Tracking which seem to be quite normal, not sure why they did not run in the past, and I don't think I care too much.

schneiml:
Ok, `SiStripCertificationInfo` is pretty much unfixable until we learn more about #28354. I tried moving it back by `consumesMany<DQMToken, edm::InRun>` like I did for `siStripOfflineAnalyser`, but it gets scheduled right *before* `siStripOfflineAnalyser`, where it probably should go *after* it to work correctly. I am not sure if there currently *is* a way to schedule things manually, and if there is, I have no idea how to do it.

Note that implementing #28364 will *not* fix this problem, here we need manual scheduling of legacy modules. For all I know, this is not a new problem; it just happened to work *by chance* in the past. Even when I go back to legacy `EDAnalyzer` for the two modules in question, EDM now orders them exactly the wrong way:
```
++++++ starting: global end run for module: label = 'siStripQTester' id = 9
++++++ finished: global end run for module: label = 'siStripQTester' id = 9
++++++ starting: global end run for module: label = 'siStripCertificationInfo' id = 65
++++++ finished: global end run for module: label = 'siStripCertificationInfo' id = 65
++++++ starting: global end run for module: label = 'siStripOfflineAnalyser' id = 10
++++++ finished: global end run for module: label = 'siStripOfflineAnalyser' id = 10
```

Before, they where just ordered by module ID (it seems), which would be perfectly correct.

schneiml:
superseded by #28379 (+ #28375, + another one not opened yet)

PR #28317, opened 2019-10-30
[PY3] Set PYTHONHOME to python2 to fix PyDevParameterSet unittest

smuzaffar:
This shoudl fix the PY3 IB unit test for  FWCore/PyDevParameterSet. For PY3 IBs , PyDevParameterSet is linked to python2 but `python` executable points to `python3` this make pybind11 to fail as it tries to look for python2 shared libs under python3/lib path. 

This change will only set  PYTHONHOME to python2 path for PY3 IBs.

### DIFF
--- a/FWCore/PyDevParameterSet/test/BuildFile.xml
+++ b/FWCore/PyDevParameterSet/test/BuildFile.xml
@@ -4,4 +4,7 @@
   <use   name="FWCore/ParameterSet"/>
   <use   name="FWCore/PyDevParameterSet"/>
   <use   name="FWCore/Utilities"/>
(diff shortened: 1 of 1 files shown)


smuzaffar:
@fabiocos , can we get this in IB. It is just to cleanup the unit tests and let them run parallel

smuzaffar:
+core

fabiocos:
@smuzaffar this looks as a temporary workaround right?

fabiocos:
+1

to be verified in next IB

smuzaffar:
correct, this is just to workaround for py3. Once we full move to py3 then the package PyDevParameterSet will be removed

PR #28326, opened 2019-10-31
drop EI from AOD data tier

slava77:
EI products take almost 2% of AOD.
These were introduced before we had miniAOD as a standard reference.

@rappoccio @arizzi @gpetruc @peruzzim @fgolf 
in case you know of any use case from your current/past PAT/miniAOD knowledge base.

@mbluj @swozniewski pfTausDiscriminationByDecayModeFinding and pfTausDiscriminationByIsolation are kept only via CommonTools/ParticleFlow/python/EITopPAG_EventContent_cff.py
Is there any use case for these products?

### DIFF
--- a/Configuration/EventContent/python/EventContent_cff.py
+++ b/Configuration/EventContent/python/EventContent_cff.py
@@ -491,7 +491,6 @@ AODEventContent.outputCommands.extend(EvtScalersAOD.outputCommands)
 AODEventContent.outputCommands.extend(OnlineMetaDataContent.outputCommands)
 AODEventContent.outputCommands.extend(TcdsEventContent.outputCommands)
 AODEventContent.outputCommands.extend(CommonEventContent.outputCommands)
-AODEventContent.outputCommands.extend(EITopPAGEventContent.outputCommands)
 ctpps_2016.toModify(AODEventContent, outputCommands = AODEventContent.outputCommands + RecoCTPPSAOD.outputCommands)
 
 RAWAODEventContent.outputCommands.extend(AODEventContent.outputCommands)


slava77:
@cmsbuild please test

mbluj:
> @mbluj @swozniewski pfTausDiscriminationByDecayModeFinding and pfTausDiscriminationByIsolation are kept only via CommonTools/ParticleFlow/python/EITopPAG_EventContent_cff.py
> Is there any use case for these products?

These products looks as copies of "regular" tau reco products (probably with modified inputs) which are stored via [RecoTauTag/Configuration/python/RecoTauTag_EventContent_cff.py](https://github.com/cms-sw/cmssw/blob/master/RecoTauTag/Configuration/python/RecoTauTag_EventContent_cff.py). I am not aware of any use case of these "EI taus".

fabiocos:
@rappoccio @arizzi @gpetruc @peruzzim @fgolf before I integrate this PR could some of you comment?

fabiocos:
@slava77 in the meanwhile I have integrated #28328 that is related to this PR

slava77:
@ahinzmann @knash @lathomas 
tagging you, since there is also a somewhat related private thread.
Note that the EI collections can be remade from AOD (#28328)

ahinzmann:
@slava77 We confirmed with our TOP-PAG jet contact that no analysis in TOP is using this EI in AOD at the moment.

slava77:
> @slava77 We confirmed with our TOP-PAG jet contact that no analysis in TOP is using this EI in AOD at the moment.

Thank you.
AFAIK, TOP DQM was even the only customer for this in the physics DQM pieces.
IIRC, the setup itself was based in part on TQAF selections.
So, it sounds safe to drop it then.
@fabiocos please consider including this in pre12.
Thank you.

fabiocos:
+operations

no negative reaction from experts inquired

fabiocos:
+1

PR #28328, opened 2019-10-31
fixes to allow running EI from AOD

slava77:
- update PFCandidateWithSrcPtrFactory to make a candidate using the earliest but still available pointer
- bugfixes to pfTaus (EI) to make the right temporary products and use a correct input 
    - IIRC, these were made "so that it runs" initially when EI setup was introduced
- load a BTV computer  to the known ES

This now allows to run
```
cmsDriver.py remakeEI --conditions auto:phase1_2023_realistic --era Run3 -s EI\
 --eventcontent AODSIM --datatier AODSIM --filein file:aod.root --fileout file:aodEI.root -n -1
```
The executed set of modules looks relevant.

The PR is somewhat related to #28326 (in a way that with this PR merging of #28326 is much more straightforward)


slava77:
@cmsbuild please test

slava77:
+1

for https://github.com/cms-sw/cmssw/pull/28328	31458df  
- the default EI setup, where the EI is running together with the RECO step, which is tested in the jenkins tests came back with no differences

santocch:
+1

fabiocos:
+1

PR #28330, opened 2019-11-01
HCAL: implementation of TDC simulation functionality for Phase1 HB/HE

lwang046:
#### PR description:
1,Update of the HB/HE TDC simulation
    The current version of HcalTDC.cc was developed but obsolete and never used in simulation production. The algorithm in it is rewritten to provide TDC timing according to the specifications of QIE11 running at p5, and the timing method is now used when simulating hcal electronics to get the TDC time
2, Modification of MC Digis packing (PackerHelp.h and HCalDigiToRawuHTR.cc)
    Convert the data format according to uHTR specifications for HB 
3, Making TDC current cut a configurable (HcalElectronicsSim.cc and some other files)
    Passed a new variable to the constructor to allow for tuning of TDC current threshold from python configuration files.

#### PR validation:

(1) pion-gun results are identical w/wo the TDC branch in 11_0_0_pre11: [results here](https://cms-cpt-software.web.cern.ch/cms-cpt-software/General/Validation/SVSuite/HCAL/calo_scan_single_pi/Run3/11_0_0_pre11_run3_TDC_vs_11_0_0_pre11_run3_SinglePi/)

(2) TDC has been tested privately and yields expected results; eg. page 5 [here](https://indico.cern.ch/event/856715/contributions/3606082/attachments/1929084/3194540/TDC_in_Simulation.pdf)

(3) runTheMatrix.py -s test is OK.

### DIFF
--- a/EventFilter/HcalRawToDigi/plugins/PackerHelp.h
+++ b/EventFilter/HcalRawToDigi/plugins/PackerHelp.h
@@ -129,7 +129,7 @@ namespace QIE11HeaderSpec0 {
   static const int MASK_FLAVOR = 0x7;
   static const int OFFSET_HEADER_BIT = 15;
   static const int MASK_HEADER_BIT = 0x1;
-}  // namespace QIE11HeaderSpec
+}  // namespace QIE11HeaderSpec0
 
 namespace QIE11HeaderSpec3 {
   static const int OFFSET_FIBERCHAN = 0;
@@ -144,7 +144,7 @@ namespace QIE11HeaderSpec3 {
   static const int MASK_FLAVOR = 0x7;
   static const int OFFSET_HEADER_BIT = 15;
   static const int MASK_HEADER_BIT = 0x1;
-}
+}  // namespace QIE11HeaderSpec3
 
 namespace TPHeaderSpec {
   static const int OFFSET_TOWER = 0;
@@ -418,14 +418,18 @@ public:
 
     int fiber = eid.fiberIndex();
     int fiberchan = eid.fiberChanId();
-  // capacitor id for the first sample
+    // capacitor id for the first sample
 
     // Flavor is added
 
     int flavor = qiedf[0].flavor();
     int capid0;
-    if (qiedf.samples() == 0) capid0=0;
-    if (flavor == 3) capid0 = qiedf[1].capid(); else capid0 = qiedf[0].capid();
+    if (qiedf.samples() == 0)
+      capid0 = 0;
+    if (flavor == 3)
+      capid0 = qiedf[1].capid();
+    else
+      capid0 = qiedf[0].capid();
 
     // adding only cases for flavor 3
 
@@ -625,37 +629,42 @@ public:
 
 // converts HE QIE digies to HB data format
 
-  QIE11DataFrame convertHB2HE(QIE11DataFrame qiehe) {
-    QIE11DataFrame qiehb = qiehe;
-    int adc_, tdc_, capid_;
-    bool soi_;
-    int is=0;
+QIE11DataFrame convertHB2HE(QIE11DataFrame qiehe) {
+  QIE11DataFrame qiehb = qiehe;
+  int adc_, tdc_, capid_;
+  bool soi_;
+  int is = 0;
 
-    //  flavor for HB digies is hardcoded here
-    static const int hbflavor = 3;
+  //  flavor for HB digies is hardcoded here
+  static const int hbflavor = 3;
 
-    //  iterator over samples
-    for (edm::DataFrame::const_iterator it = qiehe.begin(); it != qiehe.end(); ++it) {
-      if (it == qiehe.begin()) continue;
-      adc_ = qiehe[is].adc();
-      tdc_ = qiehe[is].tdc();
-      soi_ = qiehe[is].soi();
+  //  iterator over samples
+  for (edm::DataFrame::const_iterator it = qiehe.begin(); it != qiehe.end(); ++it) {
+    if (it == qiehe.begin())
+      continue;
+    adc_ = qiehe[is].adc();
+    tdc_ = qiehe[is].tdc();
+    soi_ = qiehe[is].soi();
 
-      if(tdc_>=0&&tdc_<=24) tdc_=0;
-      else if(tdc_>=25&&tdc_<=49) tdc_=1;
-      else if(tdc_==63) tdc_=2;
-      else if(tdc_==62) tdc_=3;
+    if (tdc_ >= 0 && tdc_ <= 24)
+      tdc_ = 0;
+    else if (tdc_ >= 25 && tdc_ <= 49)
+      tdc_ = 1;
+    else if (tdc_ == 63)
+      tdc_ = 2;
+    else if (tdc_ == 62)
+      tdc_ = 3;
 
-      qiehb.setSample(is,adc_,tdc_,soi_);
-      is++;
-    };
+    qiehb.setSample(is, adc_, tdc_, soi_);
+    is++;
+  };
 
-    // puting flavor is safe here because flavor is stored in the same bits for all flavors
-    qiehb.setFlavor(hbflavor);
-    capid_ = qiehe[0].capid();
-    qiehb.setCapid0(capid_);
+  // puting flavor is safe here because flavor is stored in the same bits for all flavors
+  qiehb.setFlavor(hbflavor);
+  capid_ = qiehe[0].capid();
+  qiehb.setCapid0(capid_);
 
-    return qiehb;
-  }
+  return qiehb;
+}
 
 #endif
--- a/EventFilter/HcalRawToDigi/plugins/PackerHelp.h
+++ b/EventFilter/HcalRawToDigi/plugins/PackerHelp.h
@@ -420,35 +420,26 @@ public:
     int fiberchan = eid.fiberChanId();
     // capacitor id for the first sample
 
-    // Flavor is added
-
     int flavor = qiedf[0].flavor();
     int capid0;
     if (qiedf.samples() == 0)
       capid0 = 0;
-    if (flavor == 3)
-      capid0 = qiedf[1].capid();
-    else
+    else if (flavor == 3) {
+      header |= (fiberchan & QIE11HeaderSpec3::MASK_FIBERCHAN) << QIE11HeaderSpec3::OFFSET_FIBERCHAN;
+      header |= (fiber & QIE11HeaderSpec3::MASK_FIBER) << QIE11HeaderSpec3::OFFSET_FIBER;
+      header |= (0x0 & QIE11HeaderSpec3::MASK_MP) << QIE11HeaderSpec3::OFFSET_MP;
+      header |= (0x0 & QIE11HeaderSpec3::MASK_LINKERROR) << QIE11HeaderSpec3::OFFSET_LINKERROR;
+      header |= (flavor & QIE11HeaderSpec3::MASK_FLAVOR) << QIE11HeaderSpec3::OFFSET_FLAVOR;  //flavor
+      header |= (0x1 & QIE11HeaderSpec3::MASK_HEADER_BIT) << QIE11HeaderSpec3::OFFSET_HEADER_BIT;
+    } else {
       capid0 = qiedf[0].capid();
-
-    // adding only cases for flavor 3
-
-    switch (flavor) {
-      case 0:
-        header |= (fiberchan & QIE11HeaderSpec0::MASK_FIBERCHAN) << QIE11HeaderSpec0::OFFSET_FIBERCHAN;
-        header |= (fiber & QIE11HeaderSpec0::MASK_FIBER) << QIE11HeaderSpec0::OFFSET_FIBER;
-        header |= (capid0 & QIE11HeaderSpec0::MASK_CAPID) << QIE11HeaderSpec0::OFFSET_CAPID;
-        header |= (0x0 & QIE11HeaderSpec0::MASK_FIBERERR) << QIE11HeaderSpec0::OFFSET_FIBERERR;
-        header |= (flavor & QIE11HeaderSpec0::MASK_FLAVOR) << QIE11HeaderSpec0::OFFSET_FLAVOR;  //flavor
-        header |= (0x1 & QIE11HeaderSpec0::MASK_HEADER_BIT) << QIE11HeaderSpec0::OFFSET_HEADER_BIT;
-      case 3:
-        header |= (fiberchan & QIE11HeaderSpec3::MASK_FIBERCHAN) << QIE11HeaderSpec3::OFFSET_FIBERCHAN;
-        header |= (fiber & QIE11HeaderSpec3::MASK_FIBER) << QIE11HeaderSpec3::OFFSET_FIBER;
-        header |= (0x0 & QIE11HeaderSpec3::MASK_MP) << QIE11HeaderSpec3::OFFSET_MP;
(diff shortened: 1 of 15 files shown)


abdoulline:
Long, you might want to edit the title to smth. more self-explanatory, e.g.:

"HCAL: implementation of TDC functionality for Phase1 HB/HE"

abdoulline:
ups..   TDC ->  TDC simulaton...

slava77:
@cmsbuild please test

abdoulline:
@perrotta, @civanch, @slava77
Dear colleagues, would be good to have it in pre12.

EventFilter/HcalRawToDigi/plugins/PackerHelp.h
```diff
@@ -401,14 +418,37 @@ class UHTRpacker {
...
+    // adding only cases for flavor 3
+
+    switch (flavor) {
+      case 0:
+        header |= (fiberchan & QIE11HeaderSpec0::MASK_FIBERCHAN) << QIE11HeaderSpec0::OFFSET_FIBERCHAN;
+        header |= (fiber & QIE11HeaderSpec0::MASK_FIBER) << QIE11HeaderSpec0::OFFSET_FIBER;
+        header |= (capid0 & QIE11HeaderSpec0::MASK_CAPID) << QIE11HeaderSpec0::OFFSET_CAPID;
+        header |= (0x0 & QIE11HeaderSpec0::MASK_FIBERERR) << QIE11HeaderSpec0::OFFSET_FIBERERR;
+        header |= (flavor & QIE11HeaderSpec0::MASK_FLAVOR) << QIE11HeaderSpec0::OFFSET_FLAVOR;  //flavor
+        header |= (0x1 & QIE11HeaderSpec0::MASK_HEADER_BIT) << QIE11HeaderSpec0::OFFSET_HEADER_BIT;
```
perrotta: Shouldn't it ```break``` here?

perrotta: Above ```flavor == 0``` is the "default": shouldn't it be the same also here?
Wouldn't be more convenient to include these into the if block above (from L429)?

EventFilter/HcalRawToDigi/plugins/PackerHelp.h
```diff
@@ -401,14 +418,37 @@ class UHTRpacker {
...
-    header |= (0x1 & QIE11HeaderSpec::MASK_HEADER_BIT) << QIE11HeaderSpec::OFFSET_HEADER_BIT;
+    // capacitor id for the first sample
+
+    // Flavor is added
+
+    int flavor = qiedf[0].flavor();
+    int capid0;
+    if (qiedf.samples() == 0)
+      capid0 = 0;
+    if (flavor == 3)
```
perrotta: ```else if```?

EventFilter/HcalRawToDigi/plugins/PackerHelp.h
```diff
@@ -587,4 +627,44 @@ class UHTRpacker {
   };
 };
 
+// converts HE QIE digies to HB data format
+
+QIE11DataFrame convertHB2HE(QIE11DataFrame qiehe) {
+  QIE11DataFrame qiehb = qiehe;
+  int adc_, tdc_, capid_;
```
perrotta: Normally a trailing underscore is reserved to class members, not for local variables of a class method

EventFilter/HcalRawToDigi/plugins/PackerHelp.h
lwang046: Indeed it would be more convenient to include these lines in the if block, such modifications made in commit 768d39c

EventFilter/HcalRawToDigi/plugins/PackerHelp.h
lwang046: renamed these local variables in commit 768d39c

EventFilter/HcalRawToDigi/plugins/PackerHelp.h
lwang046: `else if` is more logically correct in this case

EventFilter/HcalRawToDigi/plugins/PackerHelp.h
```diff
@@ -401,14 +418,28 @@ class UHTRpacker {
...
-    header |= (0x0 & QIE11HeaderSpec::MASK_FLAVOR) << QIE11HeaderSpec::OFFSET_FLAVOR;  //flavor
-    header |= (0x1 & QIE11HeaderSpec::MASK_HEADER_BIT) << QIE11HeaderSpec::OFFSET_HEADER_BIT;
+    // capacitor id for the first sample
+
+    int flavor = qiedf[0].flavor();
+    int capid0;
+    if (qiedf.samples() == 0)
+      capid0 = 0;
+    else if (flavor == 3) {
+      header |= (fiberchan & QIE11HeaderSpec3::MASK_FIBERCHAN) << QIE11HeaderSpec3::OFFSET_FIBERCHAN;
```
perrotta: cupid0 assignment forgotten here

SimCalorimetry/HcalSimAlgos/src/HcalTDC.cc
```diff
@@ -2,34 +2,24 @@
...
-  // start with a loop over 10 samples
   bool hasTDCValues = true;
   if (lf.preciseSize() == 0)
     hasTDCValues = false;
 
-  // "AC" hysteresis from Tom:  There cannot be a second transition
-  // within the QIE within 3ns
-  int const rising_ac_hysteresis(5);
-  int lastRisingEdge(0);
+  //std::cout << "TDC threshold: " << TDC_Threshold << std::endl;
```
perrotta: remove commented out code, if not needed

SimCalorimetry/HcalSimAlgos/src/HcalTDC.cc
```diff
@@ -46,82 +36,60 @@ void HcalTDC::timing(const CaloSamples& lf, Digi& digi, CLHEP::HepRandomEngine*
...
-          TDC_RisingEdge = i - preciseBegin;
-          lastRisingEdge = 0;
-          // std::cout << " rising ";
-        }
-        // std::cout << '\n';
+      for (int i = preciseBegin; i < preciseEnd; ++i) {  //find the TDC time value in each TS
 
-        //This is the place for hysteresis code.  Need to to arm the
-        //tdc by setting the risingReady or fallingReady to true.
+        //std::cout << " preciseBin: " << i << " preciseAt(i): " << lf.preciseAt(i) << std::endl;
```
perrotta: remove commented out code, if not needed

SimCalorimetry/HcalSimAlgos/src/HcalTDC.cc
```diff
@@ -46,82 +36,60 @@ void HcalTDC::timing(const CaloSamples& lf, Digi& digi, CLHEP::HepRandomEngine*
...
+		<< " capid: " << digi[ibin].capid() 
+		<< " risingEdge: " << TDC_RisingEdge 
+		<< " packedTDC: " << packedTDC 
+		<< " tdc: " << digi[ibin].tdc() 
+		<< " flavor(HE=0/HB=3): " << digi.flavor() 
+		<< " SubDet(HE=2/HB=1): " << HcalDetId(digi.detid()).subdet() 
+		<< std::endl;
+        }
+    }*/
+    // if(HcalDetId(digi.detid()).ieta()==1&&HcalDetId(digi.detid()).iphi()==1) std::cout<<digi<<std::endl;
```
perrotta: remove commented out code, if not needed

EventFilter/HcalRawToDigi/plugins/PackerHelp.h
lwang046: This is intended, since capid is not stored in the header of HB.

perrotta: Uhm...  Before commit https://github.com/cms-sw/cmssw/pull/28330/commits/768d39c706ed0c293ec0b0337b94112631730e2d ```capid0``` was defined also in case of ```flavor == 3```, see previous L430 of EventFilter/HcalRawToDigi/plugins/PackerHelp.h. Was it a bug, then?

In view of this latest statement, then you could simplify the code (and not remain with unassigned variables in it). In EventFilter/HcalRawToDigi/plugins/PackerHelp.h please:
- Move the definitions of ```int capid0``` and ```int fiberErr``` at L345-355 inside the ```if``` clause after the ```else``` at L361
- Remove the useless ```if (qiedf.samples() == 0)``` part at L425-426m and move the definition of ```capid0``` together with the assignment at L435

EventFilter/HcalRawToDigi/plugins/PackerHelp.h
```diff
@@ -587,4 +618,44 @@ class UHTRpacker {
...
+    else if (tdc == 62)
+      tdc = 3;
+
+    qiehb.setSample(is, adc, tdc, soi);
+    is++;
+  };
+
+  // puting flavor is safe here because flavor is stored in the same bits for all flavors
+  qiehb.setFlavor(hbflavor);
+  capid = qiehe[0].capid();
```
perrotta: Please define ```capid``` here, and not at the beginning of the method (L625):
```C++
int capid = qiehe[0].capid();
```

EventFilter/HcalRawToDigi/plugins/PackerHelp.h
```diff
@@ -587,4 +618,44 @@ class UHTRpacker {
...
+    tdc = qiehe[is].tdc();
+    soi = qiehe[is].soi();
+
+    if (tdc >= 0 && tdc <= 24)
+      tdc = 0;
+    else if (tdc >= 25 && tdc <= 49)
+      tdc = 1;
+    else if (tdc == 63)
+      tdc = 2;
+    else if (tdc == 62)
```
perrotta: Aren't other values possible for ```qiehe[is].tdc()``` other than:
- 0<=tdc<=49
- 62
- 63 ?

(Even in case of errors, misconfigurations, etc...)

lwang046: According to the original QIE11 specifications, 0-49 for time information, 50-64 for special codes such as modes, lock, errors and special current condition. But HB has only 2 bits available for TDC, not all these information can be stored, so only the ones interested for physics are selected here. Special code 62 corresponds to the case when the current discriminator starts high. Special code 63 is when the TDC current discriminator starts low and remains low (nothing happened). @abdoulline I thought MC simulation is an ideal case where we don't have  such misconfiguration errors?

abdoulline:
Long, here is the packing of _whatever_ you've put in HcalTDC.cc, 
so no other codes possible in MC...

EventFilter/HcalRawToDigi/plugins/PackerHelp.h
perrotta: I wonder whether it is ever possible that ```QIE11DataFrame::setSample(is, adc, tdc, soi)``` is called with a value of ```tdc``` different from 0-3, as it could be the case if codes 50-64 are returned, and what would happen in that case.
If there is not such a possibility the current implementation is fine, although it is hard to understand it by just looking at the code (and it may be error prone, I'd say...)

lwang046: This convertHB method is currently only used to convert MC QIE11, where only 0-49, 62 and 63 are assigned to the tdc, with their meanings mentioned above.

EventFilter/HcalRawToDigi/plugins/PackerHelp.h
lwang046: yeah capid0 is declared a bit early but only needs to be used for flavor==0 case. changes made according to suggestion for simplification.

perrotta:
please test

civanch:
+1

perrotta:
DQM comparisons in the jenkins automatic tests show differences, which are quite likely explained by the increased TDC threshold set. E.g. for wf 10824.0 HEM entries lower from 14952 to 772, with a completely different time distribution:

image

image

@lwang046 can you confirm that plots as above (as well al all the other different ones visible in the test outputs) are in fact expected? And, just for my own education, could you please explain what do they describe, and what exactly determines those differences?

perrotta:
> @perrotta
> Andrea,
> (1) TDC simulation shouldn't affect anything in RECO at this stage, as my Run3 particle gun results showed with the original branch ((1) in the header of PR).
> (2) While HCAL DQM (primarily data-oriented, which we're not using for regular RelVals MC campaigns) looks at TDC, which **was not** simulated previously (wo this PR) at all, so previously in TDC-dedicated bits there was (_undeleted_) MIX information, while only now we start putting meaningful TDC information in there.
> So, please, ignore what was TDC MC timing before...

In fact, tests do not show any difference in reco outputs. I was only puzzled by the ones showing up in DQM. Thank you Salavat for explaining their origin

abdoulline:
...Clarification:  Mixing *continues* to use TDC bits, but now (with this PR) at the end of Digitization (after Mixing) comes TDC simulation, which fills these bits with regular TDC info before packing (Digi2Raw)...

perrotta:
+1
- "Meaningful" TDC bits added to HCAL simulation, and their content correspond to what expected by the authors
- Reco outputs are not affected at this stage: jenkins tests pass and show no differences in reco

fabiocos:
+1

perrotta:
@lwang046 @abdoulline 

Since the merging yesterday of this PR, tests in other pull requests show differences in the DQM outputs for HCal TDC's. See for example the test results of #28363, where workflows with and without pileup have this issue.

As you already pointed out earlier in this thread, this does not affect reco outputs. However, it seems to point to some non-reproducibility of the TDC results provided by this PR, which should be better understood.

Could you please have a look and provide an explanation (and possibly also a fix) for such a non-reproducibility?

fabiocos:
@perrotta it seems we arrive at the same conclusion

abdoulline:
@perrotta 
Could you, please, elaborate a bit on "non-reproducibility" of HCAL TDC ?...

fabiocos:
@abdoulline changes in LETDCTime histograms for DigiTask in a geometry PR for DD4help... https://cmssdt.cern.ch/SDT/jenkins-artifacts/baseLineComparisons/CMSSW_11_0_X_2019-11-07-1100+cfae32/34131/10024.0_TTbar_13+TTbar_13TeV_TuneCUETP8M1_2017_GenSimFull+DigiFull_2017+RecoFull_2017+HARVESTFull_2017+ALCAFull_2017+NanoFull_2017/Hcal_DigiTask.html

abdoulline:
So, this is 10024.0  2017...     
Well, as I've mentioned previously, premixing is using TDC bits in QIE11DataFrame up until TDC simulation is done specifically for HB/HE _after premixing_   
HCAL ElectronicsSimulation (DIGI) is using CLHEP::HepRandomEngine, but I don't see right away the reasons of any TDC-plots irresproducibility...

abdoulline:
Just for laying down the observations: 
the issue affects **only** one HE (HEP17) pilot module of 2017, which was Phase1 QIE11 (TDC is simulated specifically for QIE11), unlike the rest of HE, which was legacy Phase0 QIE8...  

I must admit I don't understand the origin of issue,  it might be related to DQM deliberate back-incompatibility (unlike DQMOffline) and it's primary data "orientation" (we're not using it for MC RelVals).  
And it looks harmless, albeit obviously _annoying_ in PR's DQM tests... 

A fairly quick workaround could be that we can _switch off_ TDC simulation for all eras, 
but  >= Run3  by setting TDC current cut=-999  by default and then **not** run TDC simulation in this case, while set it to actual value only/specifically for Run3 (Phase2 inherits it) in the config.file.

PR #28331, opened 2019-11-01
mark mutable atomic_value_ptr members as CMS_THREAD_SAFE in [V]ParamererSetEntry

slava77:
based on what's apparently false-positive reports of thread unsafety by the static analyzer.

I expect that this PR will substantially reduce the total count of warnings (IIUC, this count is replicated from every include and use case of [V]ParameterSetEntry)

### DIFF
--- a/FWCore/ParameterSet/interface/ParameterSetEntry.h
+++ b/FWCore/ParameterSet/interface/ParameterSetEntry.h
@@ -8,6 +8,7 @@
   */
 
 #include "FWCore/Utilities/interface/atomic_value_ptr.h"
+#include "FWCore/Utilities/interface/thread_safety_macros.h"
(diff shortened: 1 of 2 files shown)


slava77:
@cmsbuild please test

slava77:
@Dr15Jones 
are there concerns about this change?

Dr15Jones:
+1

fabiocos:
+1

slava77:
looking at the static analyzer results between CMSSW_11_0_X_2019-11-10-0000 and CMSSW_11_0_X_2019-11-10-2300, I conclude that this PR reduced the number of issues in the category "mutable member if accessed via const pointer " from 80.7K to 67.1K (17%).

PR #28332, opened 2019-11-01
Simplify inheritance hirachy of EcalClusterFunctions

guitargeek:
#### PR description:

On my egamma refactors, I frequently encounter these plugins inheriting from `EcalClusterFunctionBaseClass` and it's always a bit complicated to understand what is going on.

For every plugin (let's say the `EcalClusterCrackCorrection`, there is another intermediate class defined (`EcalClusterCrackCorrectionBaseClass`) which is private to the plugins directory and is only used such that the `EcalClusterCrackCorrection` can inherit from it. This pattern is repeated for all the plugins. I would suggest to simplify this by having the plugins all inherit from `EcalClusterFunctionBaseClass` directly. Like this, the `RecoEcal/EgammaCoreTools/plugins` directory looks radically simpler and it would really help for understanding the code and future refactors. For example, one can now just do `git grep EcalClusterFunctionBaseClass` to find all the plugins that inherit from it. This PR doesn't make any actual changes to the code, it just reshuffles some lines.

#### PR validation:

CMSSW complies and local matrix tests pass.

No backport intended.


slava77:
@ferriff it looks like you are the original developer/designer of these tools in RecoEcal/EgammaCoreTools
please check that this update is still OK wrt the original design ideas or if those are not needed anymore.
@argiro please confirm that this is fine for the ECAL DPG.
Thank you.

guitargeek:
I assumed the idea was to maybe have at some point separate factories for alternative functions that serve the same purpose, like crack correction or energy correction, is that correct? But since these don't exist, this inheritances were just complicated boilerplate at this point.

slava77:
@cmsbuild please test

VinInn:
is this inheritance needed at all?

guitargeek:
> is this inheritance needed at all?

@VinInn Right now it is technically used because many plugins in Egamma use the `EcalClusterFunctionFactory` to create these helper functions. Whether this is the good way to use these helper functions is another question. In this PR I only wanted to address the "internal" inheritance hierarchy in the plugins directory for now.

perrotta:
please test

perrotta:
+1
- The simplification achieved by eliminating the intermediate layer of base classes is welcome, and blessed (in private communications) by the ECAL contacts
- Additional simplification and cleanings also applied during the review
- Plugins functionalities are not affected

fabiocos:
+1

PR #24356, opened 2018-08-21
GEM strip definition change

jshlee:
To make all strip/channel definitions unify with HW it was suggested that we change all definitions for strips and channels to start from 0.

Bug in gem digitizer which was causing low efficiency (https://indico.cern.ch/event/748857/contributions/3097829/attachments/1699357/2736224/180809_GEMSimpleModel_bug_fix.pdf)

@dildick @kpedro88 @watson-ij @slowmoyang

PR #28335, opened 2019-11-01
Redefine (in)valid GEM digis and pads

dildick:
#### PR description:

PR #24356 reset the begin pad as 0 instead of 1. This broke a function in the GEM-CSC trigger. To fix it, I redefine when strips, pads and clusters are valid. Default strips/pads are now initialized with bx -99 and strip/pad number 2^16-1. 

#### PR validation:

I checked that it runs with WF 22034. 

N/A

### DIFF
--- a/DataFormats/GEMDigi/interface/GEMDigi.h
+++ b/DataFormats/GEMDigi/interface/GEMDigi.h
@@ -4,7 +4,7 @@
 /** \class GEMDigi
  *
  * Digi for GEM
- *  
+ *
  * \author Vadim Khotilovich
  *
  */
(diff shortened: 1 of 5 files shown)


fabiocos:
please test

kpedro88:
+upgrade

civanch:
+1

fabiocos:
@dildick are we sure this does not interfere with existing datasets? I mean that these values cannot be assumed by old good data, and therefore their state of being good is unchanged?

dildick:
@fabiocos Correct. GEM BX value is 0 for in-time hits, while the GEM strip number is maximum 768.

fabiocos:
+1

PR #28339, opened 2019-11-04
[TEST ONLY] New DQMEDHarvester

schneiml:
This is a single commit out of #28316 to isolate the changes cause by it.

This is not supposed to be merged.


schneiml:
please test

schneiml:
Ok, no changes here, so all the things going on in #28316 are actually cause by the qtest changes.

PR #28346, opened 2019-11-05
Added consumes to DD related ESProducers

Dr15Jones:
#### PR description:

Also use ESGetToken to retrieve data.

#### PR validation:

The unit tests succeed, even when using #28223.

### DIFF
--- a/DetectorDescription/DDCMS/plugins/DDCompactViewESProducer.cc
+++ b/DetectorDescription/DDCMS/plugins/DDCompactViewESProducer.cc
@@ -44,13 +44,12 @@ public:
   ReturnType produce(const IdealGeometryRecord&);
 
 private:
-  const string m_label;
+  const edm::ESGetToken<DDDetector, IdealGeometryRecord> m_detToken;
 };
 
 DDCompactViewESProducer::DDCompactViewESProducer(const edm::ParameterSet& iConfig)
-    : m_label(iConfig.getParameter<std::string>("appendToDataLabel")) {
-  setWhatProduced(this);
-}
+    : m_detToken(setWhatProduced(this).consumes<DDDetector>(
+          edm::ESInputTag("", iConfig.getParameter<std::string>("appendToDataLabel")))) {}
 
 DDCompactViewESProducer::~DDCompactViewESProducer() {}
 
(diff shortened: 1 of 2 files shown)


Dr15Jones:
please test

Dr15Jones:
One important note,  #28223 depends on this pull request to fix a unit test failure but this pull request does NOT depend on #28223 and can be integrated at anytime before the other pull request.

Dr15Jones:
@ianna @cvuosalo please review as this is the last pull requested needed to finalize a major code campaign for the Framework group.

fabiocos:
+1

@ianna @cvuosalo the code looks ok to me, please check and in case comment for a further iteration. As I want to get ready for a next pre-release I would like to have the development of which this code is part finalized in time

fabiocos:
merge

cvuosalo:
@Dr15Jones @namapane I am just seeing this PR now. I think it is related to the reason the magnetic field geometry builder stopped working. I'm guessing that the changes made in this PR to ` DetectorDescription/DDCMS/plugins/DDCompactViewESProducer.cc` need also to be made to  `DetectorDescription/DDCMS/plugins/DDCompactViewMFESProducer.cc` to fix the mf geometry builder.

namapane:
@cvuosalo, there's one thing I do not understand here, why do we need an explicit ESProducer here: 
https://github.com/cms-sw/cmssw/blob/5d038bc896ae144f071cd0ab83e07c36b3970757/MagneticField/GeomBuilder/test/python/testMagGeometry.py#L84
while we have no explicit DDCompactViewESProducer set up in any of the .py for the non-dd4hep versions of the MF?
And also, would this DDCompactViewMFESProducer snippet have to go in the cfi included in the standard sequences or not? I also have to figure out if that is needed for the dd4hep producer from DB, which would not require a  DDCompactViewMFESProducer, I suppose, since it must create iteslf the  DDCompactView from the fileBlob.

cvuosalo:
@namapane I'm not sure why it is needed now. The way this program gets its products has been changed over the last several months by @Dr15Jones and @makortel so I'm not sure anymore how it should work. I can tell you that originally, before I created the DDCompactViewMFESProducer, the Python config generated errors saying that an ESProducer was needed. You could try commenting out the DDCompactViewMFESProducer to see if it's still needed. Maybe there is a better way now to write the Python config. You could ask @Dr15Jones or @ianna when she returns.

fabiocos:
@cvuosalo the problem you are mentioning isn't clear to me, I integrated this PR as it looked technically straightforward and ok. Could you please open an issue and describe the problem, so as we can cleanly keep track of it?

namapane:
@fabiocos well this is part of the plan of migrating to dd4hep. At the moment we only have a static producer for dd4hep that is not used in production; to complete the migration we have to implement the version from DB and while doing so we have to get record dependencies right. It is not clear to me how to do this but that's in my to-do list (proceeding slowly because of teaching...)

namapane:
I fixed the problem in #28642; as suggested by Carl DDCompactViewMFESProducer had to be updated.

cvuosalo:
+1

PR #28348, opened 2019-11-05
Add a possibility to display a GDML file

ianna:
#### PR description:

A GDML file produced by G4 GDML writer can be viewed in cmsShow. For example:

```
cmsShow -c g4Geom.fwc --sim-geom-file=cmsG4Geometry.gdml --tgeo-name=world-volume
```
Screen Shot 2019-11-05 at 12 06 08

@civanch - FYI

### DIFF
--- a/Fireworks/Core/src/FWGeometryTableViewManager.cc
+++ b/Fireworks/Core/src/FWGeometryTableViewManager.cc
@@ -94,12 +94,13 @@ void FWGeometryTableViewManager::setGeoManagerFromFile() {
   TFile* file = FWGeometry::findFile(m_fileName.c_str());
   fwLog(fwlog::kInfo) << "Geometry table file: " << m_fileName.c_str() << std::endl;
   try {
-    if (!file)
-      throw std::runtime_error("No root file.");
-
-    file->ls();
-
-    s_geoManager = (TGeoManager*)file->Get(m_TGeoName.c_str());
+    if (!file) {
+      // Try it as a GDML file
+      s_geoManager = TGeoManager::Import(m_fileName.c_str(), m_TGeoName.c_str());
+    } else {
+      file->ls();
+      s_geoManager = (TGeoManager*)file->Get(m_TGeoName.c_str());
+    }
     if (!s_geoManager)
       throw std::runtime_error("Can't find TGeoManager object in selected file.");
 


ianna:
please test

civanch:
@ianna , looks great.

Fireworks/Core/src/FWGeometryTableViewManager.cc
```diff
@@ -94,12 +94,13 @@ void FWGeometryTableViewManager::setGeoManagerFromFile() {
...
   try {
-    if (!file)
-      throw std::runtime_error("No root file.");
-
-    file->ls();
-
-    s_geoManager = (TGeoManager*)file->Get(m_TGeoName.c_str());
+    if (!file) {
+      // Try it as a GDML file
+      s_geoManager = TGeoManager::Import(m_fileName.c_str(), m_TGeoName.c_str());
```
fabiocos: @ianna as far as I can see ```TGeoManager``` has already a protection against odd files, and in any case the final check here on the pointer not being null should be enough...

ianna:
FWGeometry::findFile returns a nullptr if it’s not a root file.

On 6 Nov 2019, at 17:33, Fabio Cossutti <notifications@github.com<mailto:notifications@github.com>> wrote:


@fabiocos commented on this pull request.

In Fireworks/Core/src/FWGeometryTableViewManager.cc<https://github.com/cms-sw/cmssw/pull/28348#discussion_r343198654>:

> @@ -94,12 +94,13 @@ void FWGeometryTableViewManager::setGeoManagerFromFile() {
   TFile* file = FWGeometry::findFile(m_fileName.c_str());
   fwLog(fwlog::kInfo) << "Geometry table file: " << m_fileName.c_str() << std::endl;
   try {
-    if (!file)
-      throw std::runtime_error("No root file.");
-
-    file->ls();
-
-    s_geoManager = (TGeoManager*)file->Get(m_TGeoName.c_str());
+    if (!file) {
+      // Try it as a GDML file
+      s_geoManager = TGeoManager::Import(m_fileName.c_str(), m_TGeoName.c_str());


@ianna<https://github.com/ianna> as far as I can see TGeoManager has already a protection against odd files, and in any case the final check here on the pointer not being null should be enough...

alja:
@fabio 
The argument in FWGeometryTableViewManager::setGeoManagerFromFile() can be a root file or GDML file, which is not in root format. The TGeoManager can read the GDML files, but TFile can't be created out of it.

I think Yana's change is OK. But, I would change FWGeometry::findPath() [1] to locate just a file location without creating the TFile object. This will make a consistent search for a geo files in given environment. @yana can you also make that change?

 https://github.com/cms-sw/cmssw/blob/master/Fireworks/Core/src/FWGeometry.cc#L26

fabiocos:
@alja @ianna thanks for your comments, mine was not a doubt but just stating that the code should be safe also because of the TGeoManager built-in protection.

@ianna there is a suggestion for an update (with the wrong address, may be you did not see it) https://github.com/cms-sw/cmssw/pull/28348#issuecomment-550452453 
Although in that piece of code I do not see a real use case now, it does not harm I would say.

ianna:
@fabiocos and @alja - I think, searching for the GDML files in the CMSSW paths is not required since it's not a standard geometry file, but rather an import from another tool for debugging purposes only. Using ```--sim-geom-file``` option is a quick solution, but perhaps we should introduce another command line option ```--import-geom-file```? We ought to discuss it and it would be another PR implementing it.

civanch:
@ianna , @fabiocos , @alja , as far as I know, we do not have any gdml files inside CMSSW data. Production of a new gdml for any version of CMS or a sub-detector takes few minutes. In our practice, we provide these gdml for external use, for example, to study geometry performance by geometry experts of Geant4. I am not sure, if it is required to have an official gdml files supported by CMSSW external data facility. So, for me this PR is fine.

alja:
@ianna Thanks Yana for the explanation. To me PR is fine too.
+1

fabiocos:
@alja apparently your +1 has not been seen by the bot as it is not the first character, anyway your will is clear, thanks

fabiocos:
+1

fabiocos:
merge

PR #28349, opened 2019-11-05
HGCAL digitisation: end-of-life updates, more realism for startup

franzoni:
#### PR description:

+ improvements and extensions to handle of HGCAL end-of-life scenario
. improvements to the ageing mode (tools to toggle on/off different ingredients)
. fix for the treatment of the common noise, set minimum charge collection efficiency to 0

+ tool to apply realistic noise/gain setting also to startup scenarios ( `SLHCUpgradeSimulations/Configuration/python/aging.py:realisticHGCalStartup` )

+ easier to handle also in startup scenarios (reshuffling `HGCalSiNoiseMap::getSiOpCharacteristics`)
   in this PR the startup scenario is not touched, only tools prepared for the next PR

+ HGCSample:
. bits allocation changed to 12 bits reserved for data Data ( see HGC TDR page 161: adc mode only needs 10, ToT instead needs 12); 4 newly reserved for gain (currently only 3 gains used; 12 are technically possible and may be used)
. boolean `toaFired_` removed, one bit reserve to it

#### PR validation:

Occupancies (only pileup, average 200) :

radialprofile_counts_CEE_5_ps_gf

at end-of-life with these improvements have been validated against a simple analytical model which, starting form the sole occupancy of in-time hits w/o noise, predicts the extra occupancy stemming from noise and from out of time pileup. Comparing pred and FS (full-sim) in the last four columns of the table supports the validity of the occupancies form this PR

Screenshot 2019-11-05 at 16 32 16

These updated occupancies and the validation model have  been presented at the HGCAL DPG see [here](https://indico.cern.ch/event/855542/contributions/3602119/).

This PR is not a back-port

@pfs @deguio


franzoni:
please test

franzoni:
please test

kpedro88:
```
----- Begin Fatal Exception 05-Nov-2019 18:48:25 CET-----------------------
   [1] Validating configuration of module: class=HGCalLayerClusterProducer label='hgcalLayerClusters'
Exception Message:
Illegal parameter found in configuration.  The parameter is named:
 'scaleByDoseAlgo'
You could be trying to use a parameter name that is not
allowed for this plugin or it could be misspelled.
```

DataFormats/HGCDigi/bin/HGCSampleTest.cpp
```diff
@@ -0,0 +1,107 @@
...
+  std::cout << "\t + repetitions [int]: " << repetitions << std::endl;
+  // second command line argument (whatever it is) will activate
+  //                  the random choice of values for all inputs
+  bool generateRandomValues = (argc > 2 ? true : false);
+  std::cout << "\t + generateRandomValues [true/false]: " << generateRandomValues << "\n" << std::endl;
+
+  // init static values
+  uint32_t adc = 124;
+  uint32_t gain = 15;
+  uint32_t thrADC = 900;
```
kpedro88: can be const

DataFormats/HGCDigi/bin/HGCSampleTest.cpp
```diff
@@ -0,0 +1,107 @@
...
+  uint32_t toa = 0;
+  bool mode = false;
+  bool toaFired = true;
+
+  // do the trials: time/performance test and exploit randomisation to check
+  unsigned long int u = 0;
+  for (; u < repetitions; u++) {
+    // randomise all inputs, if chosen at the command line
+    if (generateRandomValues) {
+      adc = rand() % 4096;
```
kpedro88: better to use the more modern [C++ random number library](https://en.cppreference.com/w/cpp/numeric/random) rather than `rand()`

SLHCUpgradeSimulations/Configuration/python/aging.py
```diff
@@ -59,11 +59,17 @@ def ageHF(process,turnon):
...
     from SimCalorimetry.HGCalSimProducers.hgcalDigitizer_cfi import HGCal_setEndOfLifeNoise
-    process = HGCal_setEndOfLifeNoise(process)
+    process = HGCal_setEndOfLifeNoise(process,byDose=True,byDoseAlgo=algo)
     return process
 
+def realisticHGCalStartup(process):
+    from SimCalorimetry.HGCalSimProducers.hgcalDigitizer_cfi import HGCal_setRealisticStartupNoise
+    process = HGCal_setRealisticStartupNoise(process)
+    return process
+
```
kpedro88: remove extra space

SimCalorimetry/HGCalSimAlgos/interface/HGCalSiNoiseMap.h
```diff
@@ -56,10 +61,10 @@ class HGCalSiNoiseMap : public HGCalRadiationMap {
   std::vector<double> &getIleakParam() { return ileakParam_; }
   std::vector<std::vector<double> > &getENCsParam() { return encsParam_; }
   std::vector<double> &getLSBPerGain() { return lsbPerGain_; }
-  std::vector<double> &getMaxADCPerGain() { return maxADCPerGain_; }
+  std::vector<double> &getMaxADCPerGain() { return chargeAtFullScaleADCPerGain_; }
 
 private:
-  //
+  //vector of three params, per sensor type: 0:120 [mum], 1:200, 2:300
```
kpedro88: if we know these will always have 3 values, could use `std::array` rather than `std::vector`

DataFormats/HGCDigi/interface/HGCSample.h
```diff
@@ -12,31 +12,48 @@
 
 class HGCSample {
 public:
-  enum HGCSampleMasks { kThreshMask = 0x1, kModeMask = 0x1, kToAMask = 0x3ff, kDataMask = 0xfff };
-  enum HGCSampleShifts { kThreshShift = 31, kModeShift = 30, kToAShift = 13, kDataShift = 0 };
+  enum HGCSampleMasks {
```
kpedro88: it seems like the changes in this class could break backward compatibility for reading old MC samples.

Currently the class version for `HGCSample` is not tracked, but it probably should be: https://github.com/cms-sw/cmssw/blob/92e65f7236037152d597ac5f8f3407ff0a9eb625/DataFormats/HGCDigi/src/classes_def.xml

kpedro88:
@franzoni it seems like the changes here shouldn't have a major impact on the CPU usage of the digitization code with aging enabled (compared to previous versions), but has this been explicitly checked?

rovere:
From extra quick look, the crash could be related to this line:
https://github.com/cms-sw/cmssw/blob/master/RecoLocalCalo/HGCalRecProducers/interface/HGCalCLUEAlgo.h#L98

DataFormats/HGCDigi/bin/HGCSampleTest.cpp
franzoni: tks, addressed

DataFormats/HGCDigi/bin/HGCSampleTest.cpp
franzoni: tks, addressed

SLHCUpgradeSimulations/Configuration/python/aging.py
franzoni: tks, addressed

SimCalorimetry/HGCalSimAlgos/interface/HGCalSiNoiseMap.h
franzoni: tks, addressed

DataFormats/HGCDigi/interface/HGCSample.h
franzoni: > class could break backward compatibility for reading old MC samples.

it does indeed. Especially, kToAShift = 13 having changed to kToAShift = 12 (this is allegedly a fix - shifting by 13 was wasting 1 bit) will break backward compatibility.

I am not aware we are aiming at reading in 11_ release, MC productions done with previous cycles. I gather that tracking the version won't solve the incompatibility, it'll rather provide documentation of it ? I can introduce version tracking if requested, I just don't fully know how to do it (I've poked around a bit.. how does one computed the [ckecsksum](https://github.com/cms-sw/cmssw/blob/master/DataFormats/HGCRecHit/src/classes_def.xml#L3)  for instance ?

franzoni:
thanks @kpedro88 for the detailed review.

> @franzoni it seems like the changes here shouldn't have a major impact on the CPU usage of 
> the digitization code with aging enabled (compared to previous versions), but has this been 
> explicitly checked?

We introduced the `HGCSampleTest.cpp` to keep indeed that under control.
With the commit I've just added, and using `HGCSampleTest.cpp` as a benchmark there's a gain in time, see the **[vanilla pre11]** test versus the **[PR28349]** test below.
The ~10% gain stems from having removed the check wether words being set are larger than the mask of bits available, resulting in ceiling the word to the max value possible; with the current implementation, a word exceeding the max value will just get masked.
This is de facto the same behaviour as was used before this PR (where the ceiling was in one of the method, which was however never used).


**[vanilla pre11]**
```
HGCSampleTest  1000000000
Basic performance tests for HGCSample (pseudo-random seed set according to local time)

num parameters entered: 2
	 + repetitions [int]: 1000000000
	 + generateRandomValues [true/false]: 0


Done 1000000000	1000000000
HGCSampleTest 1000000000  8.55s user 0.04s system 99% cpu 8.640 total

```


**[PR28349]**
```
time HGCSampleTest  1000000000
Basic performance tests for HGCSample (pseudo-random seed set according to local time)

num parameters entered: 2
	 + repetitions [int]: 1000000000
	 + generateRandomValues [true/false]: 0


Done 1000000000	1000000000
HGCSampleTest 1000000000  7.64s user 0.08s system 97% cpu 7.881 total
```

franzoni:
please test

DataFormats/HGCDigi/interface/HGCSample.h
kpedro88: In general, people like to be able to read files from the old productions as long as there hasn't been a new production yet (at least if the geometries have stayed the same).

scram can generate the checksum. Here is an example:
```
cmsrel CMSSW_11_0_0_pre11
cd CMSSW_11_0_0_pre11/src
cmsenv
git cms-addpkg Formats/HGCDigi
```
Edit `DataFormats/HGCDigi/src/classes_def.xml` to change `   <class name="HGCSample"/>` to:
```
   <class name="HGCSample" ClassVersion="2">
    <version ClassVersion="2" checksum="0"/>
   </class>
```
then `scram b` will notice the incorrect checksum and provide an error:
```
error: class 'HGCSample' has a different checksum for ClassVersion 3. Increment ClassVersion to 4 and assign it to checksum 3298262857
Suggestion: You can run 'scram build updateclassversion' to generate src/DataFormats/HGCDigi/src/classes_def.xml.generated with updated ClassVersion
```
Running the suggested command `scram b updateclassversion` will generate a new xml file and will show this message:
```
>> Checking src/DataFormats/HGCDigi/src/classes_def.xml for EDM Class Version update
>> New Class Version File:  src/DataFormats/HGCDigi/src/classes_def.xml.generated
```
Then you can just do:
```
mv DataFormats/HGCDigi/src/classes_def.xml.generated DataFormats/HGCDigi/src/classes_def.xml
```
and then delete the dummy version 2 in the xml (recommended to start with version 3).

In this case, you would do this for the version of the class before your changes, then do it again after your changes (which would become version 4).

Further, it is possible to make the change actually backwards compatible by specifying an IO rule (which tells ROOT how to read an old version into the memory structure of the new version when deserializing). You can search the repo for `ioread` to see examples of this, though it is not always straightforward to make it work.

RecoLocalCalo/HGCalRecProducers/interface/HGCalCLUEAlgo.h
```diff
@@ -97,6 +97,8 @@ class HGCalCLUEAlgo : public HGCalClusteringAlgoBase {
     edm::ParameterSetDescription descNestedNoiseMIP;
     descNestedNoiseMIP.add<bool>("scaleByDose", false);
     iDesc.add<edm::ParameterSetDescription>("scaleByDose", descNestedNoiseMIP);
+    descNestedNoiseMIP.add<unsigned int>("scaleByDoseAlgo", 0);
+    iDesc.add<edm::ParameterSetDescription>("scaleByDoseAlgo", descNestedNoiseMIP);
     descNestedNoiseMIP.add<std::string>("doseMap", "");
     iDesc.add<edm::ParameterSetDescription>("doseMap", descNestedNoiseMIP);
```
slava77: we already have the following in the config 
```python
    scaleByDose = cms.PSet(
      scaleByDose = cms.bool(False)
    ),
    doseMap = cms.PSet(
      scaleByDose = cms.bool(False),
      doseMap = cms.string('')
    ),
    noiseMip = cms.PSet(
      scaleByDose = cms.bool(False),
      doseMap = cms.string(''),
      noise_MIP = cms.double(0.01)
    ),
```
There will now be also ```scaleByDoseAlgo = cms.PSet(scaleByDose = False, scaleByDoseAlgo = 0)``` with ```scaleByDoseAlgo = 0``` added to all the following PSets.
is this nested accumulating content really desired?

DataFormats/HGCDigi/bin/BuildFile.xml
```diff
@@ -0,0 +1,3 @@
+<use name="DataFormats/HGCDigi" />
+<bin   name="HGCSampleTest" file="HGCSampleTest.cpp">
```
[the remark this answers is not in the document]
franzoni: hi @slava77 thanks for your comments.
I gather that moving to /test does not provide an executable which is available in the $PATH.
I found very useful being able to execute at leisure the HGCSampleTest while developing, is there a way to do that ?
Otherwise I'd see good reasons to keep the .cpp in bin.

RecoLocalCalo/HGCalRecProducers/interface/HGCalCLUEAlgo.h
franzoni: thanks @slava77 
I see there are 3 params added to the main pset which are not needed.
I've added a commit which fixes that - at least according to my understanding.
The relvals run.
@deguio (who committed the 3 lines I am removing) @rovere may comment, do you share the logic ?

franzoni:
please test

DataFormats/HGCDigi/bin/BuildFile.xml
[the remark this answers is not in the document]
slava77: @Dr15Jones @davidlange6 
since this is the first /bin in \*DataFormats, I'm curious if this is accidental or by design
(I didn't find anything against in https://twiki.cern.ch/twiki/bin/view/CMSPublic/SWGuideCreatingNewProducts#Restrictions_on_the_package_defi

SimCalorimetry/HGCalSimAlgos/src/HGCalRadiationMap.cc
```diff
@@ -3,7 +3,10 @@
 #include <fstream>
 
 //
-void HGCalRadiationMap::setDoseMap(const std::string& fullpath) { doseMap_ = readDosePars(fullpath); }
+void HGCalRadiationMap::setDoseMap(const std::string& fullpath, const unsigned int& algo) {
```
slava77: better pass ```algo``` by value, it costs more to pass by ref in this case

slava77:
+1

for https://github.com/cms-sw/cmssw/pull/28349	fe44ec5 
- reco changes are only in RecoLocalCalo/HGCalRecProducers/interface/HGCalCLUEAlgo.h and look good (including also a fix to some incorrect filling of PSet introduced in the past devels)
- jenkins tests pass and comparisons with the baseline show only (relevant) changes in the phase-2 wf D46 22034.0 apparently related to the digitization changes upstream of reco

franzoni:
please test

franzoni:
hello @kpedro88 @slava77 
all your comments posted so far have been addressed - there's only one item possibly up for discussion form @kpedro88 [here](https://github.com/cms-sw/cmssw/pull/28349#discussion_r342854148), where I've asked a question/advice.

Let us know if there's more info or actions you'd like. Tkx!

DataFormats/HGCDigi/bin/BuildFile.xml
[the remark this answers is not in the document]
fabiocos: @slava77 IMHO this is just a simple test code as we have many others, and might simply go into a test directory (we have many of them in DataFormats). Discussing in the past with @davidlange6 I think /bin was mostly intended for standalone utilities... Anyway I let others to express themselves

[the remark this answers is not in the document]
franzoni: hello @fabiocos and all
can one get an executable to be run in parallel to developing, if the .cpp moves to the /test directory ?
Perhaps it was me? ,  when I moved this small program to test, no executable appeared in a folder covered by my $PATH .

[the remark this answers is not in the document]
franzoni: I also tried with the test programs of one of the DetId directories, and could not execute it.

DataFormats/HGCDigi/interface/HGCSample.h
kpedro88: @franzoni if you need more clarification here, please let me know

DataFormats/HGCDigi/bin/BuildFile.xml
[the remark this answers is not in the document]
makortel: > can one get an executable to be run in parallel to developing, if the .cpp moves to the /test directory ?
> Perhaps it was me? , when I moved this small program to test, no executable appeared in a folder covered by my $PATH .

The executables from `test` end up in `$CMSSW_BASE/test/$SCRAM_ARCH` and can be run from there.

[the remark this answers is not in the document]
Dr15Jones: In addition, executables in `test` can be run via `scram b runtests` and are run in every IB and pull request.

slava77:
IIUC, the file in /bin should be moved to /test after the most recent comments

franzoni:
Hi @slava77 and all, I'll follow up, ok.

DataFormats/HGCDigi/interface/HGCSample.h
franzoni: >  people like to be able to read files from the old productions as long as there hasn't been a new production yet

indeed. Existing large scale production have been done w/ 10_6 (for the L1T TDR). 
The next round will be with 11_0 ( see https://indico.cern.ch/event/856064/contributions/3602726/attachments/1940953/3218266/19-11-07_News_PPD.pdf ) where also geometry will be updated. 

> Here is an example:

Thanks! I've taken also the liberty of pointing this [documentation](https://twiki.cern.ch/twiki/bin/view/CMSPublic/SWGuideCreatingNewProducts#Class_versioning) to this set of instructions. The twiki paragraph may need updating.

>  it is possible to make the change actually backwards compatible by specifying an IO rule

`ioread` is used to handle root serialisation formats, in the examples I've looked at. Here the backward incompatibility is generated by a different interpretation of the bits in the same variable ( `value_` ), which is not resolved by de-serialising root differently. Backward compatibility would need adding a version number to the content of the class or of `value_` itself, and steer the content of `value_` accordingly. In my opinion this would be an overkill at this stage.
Do we concur ?

franzoni:
> IIUC, the file in /bin should be moved to /test after the most recent comments

@slava77 : done.

franzoni:
please test

DataFormats/HGCDigi/interface/HGCSample.h
kpedro88: The io rule would need to take the input `value_` from a version 3 instance of the product, and then modify it to have the correct bit sequence for version 4. An example of this comes from HcalDetId, which underwent a similar change: https://github.com/cms-sw/cmssw/blob/master/DataFormats/HcalDetId/src/classes_def.xml#L12-L17

But actually, I'm not sure how ROOT will interpret a 10_6 sample (when there was no class version assigned)...

DataFormats/HGCDigi/src/classes_def.xml
```diff
@@ -1,5 +1,9 @@
 <lcgdict>
-   <class name="HGCSample"/>
+   <class name="HGCSample" ClassVersion="11">
+    <version ClassVersion="11" checksum="2120380856"/>
+    <version ClassVersion="10" checksum="3298262857"/>
```
slava77: was the old version really 10?
Was it ever recorded in the persisted data? (IIRC, the default class version defined internally by root is something  either 0 or 1).
IIUC, the "old" version 10 and the checksum for it here have no consequence and can be removed.

New class versions should start from 3
https://twiki.cern.ch/twiki/bin/view/CMSPublic/SWGuideCreatingNewProducts#Class_versioning

slava77 (commented):
DataFormats/HGCDigi/bin/HGCSampleTest.cpp and  DataFormats/HGCDigi/bin/BuildFile.xml are still a part of this PR. They should be removed.

DataFormats/HGCDigi/src/classes_def.xml
kpedro88: The old version was not recorded. I agree the versioning should start from 3 (had overlooked this before). I thought it was helpful to have the checksum for the old version of the class, but maybe it doesn't matter.

franzoni:
> They should be removed.
addressed. Thanks @slava77

DataFormats/HGCDigi/src/classes_def.xml
franzoni: The paragraph https://twiki.cern.ch/twiki/bin/view/CMSPublic/SWGuideCreatingNewProducts#Class_versioning is in many aspects outdated.

Quoting from the twiki: 
>  If a class was previously stored without a class version assigned then one should use 10 as the initial version instead of 3 because ROOT will automatically assign a version number to each unversioned class starting with 3 and increment the version number each time it finds an instance of a class with a different checksum while writing to the file.

we should start from version 10. @kpedro88 had asked I would record the checksum also of the version which is already in CMSSW:
> I thought it was helpful to have the checksum for the old version of the class, but maybe it doesn't matter.

Given this from the twiki:
> .. because ROOT will automatically assign a version number to each unversioned class starting with 3

I'll remove the checksum of the existing dataformat and add the checksum introduced by this PR versioning it to 10.

DataFormats/HGCDigi/src/classes_def.xml
```diff
@@ -1,5 +1,8 @@
 <lcgdict>
-   <class name="HGCSample"/>
+   <class name="HGCSample" ClassVersion="10">
+    <version ClassVersion="10" checksum="2120380856"/>
```
slava77: new class versions should start from 3 https://twiki.cern.ch/twiki/bin/view/CMSPublic/SWGuideCreatingNewProducts#Class_versioning

kpedro88: @slava77 please read https://github.com/cms-sw/cmssw/pull/28349#discussion_r345994831

slava77:
@cmsbuild please test

DataFormats/HGCDigi/src/classes_def.xml
slava77: ok, thank you.

franzoni:
w/ @pfs we've implemented  schema evolution for `HGCSample`
https://root.cern.ch/root/html/io/DataModelEvolution.html
Thanks @kpedro88 for advice!

With the lats 5 commits, `HGCSample` V9 written with releases up to 11_0_0_pre12 can be read also using the V10 `HGCSample` which is introduced by this PR.

**Validation 1: inspection of a few samples**
We've generated and written out 1 event in 11_0_0_pre11, and printed the `HGCSample` from the HGCalUncalibRecHitRecWeightsAlgo.
Read back the same event in  11_0_0_pre11 + this PR, and printed the `HGCSample` from the HGCalUncalibRecHitRecWeightsAlgo - in this case the `HGCSample` have been de-serialised using the rules defined (by this PR) for the schema evolution. This is the outcome for four which have been inspected:
```

+++ V9  RecWeightsAlgo - id: 2148550960 data: 129 getToAValid: 1 raw: 2149326977
10000000000111000010000010000001
         \--> toa  !\--> data
*** V10 RecWeightsAlgo - id: 2148550960 data: 129 getToAValid: 1 raw: 2685276289
10100000000011100001000010000001
  ~       \--> toa  \-->  data


+++ V9  RecWeightsAlgo - id: 2154385067 data: 341 getToAValid: 1 raw: 2149278037
10000000000110110110000101010101
         \--> toa  !\--> data
*** V10 RecWeightsAlgo - id: 2154385067 data: 341 getToAValid: 1 raw: 2685251925
10100000000011011011000101010101
  ~       \--> toa  \-->  data


+++ V9  RecWeightsAlgo - id: 2154398930 data: 41 getToAValid: 0 raw: 2147483689
10000000000000000000000000101001
         \--> toa  !\--> data
*** V10 RecWeightsAlgo - id: 2154398930 data: 41 getToAValid: 0 raw: 2147483689
10000000000000000000000000101001
  ~       \--> toa  \-->  data


*** V9 RecWeightsAlgo - id: 2154399251 data: 54 getToAValid: 0 raw: 2147483702
10000000000000000000000000110110

+++ V10 RecWeightsAlgo - id: 2154399251 data: 54 getToAValid: 0 raw: 2147483702
10000000000000000000000000110110
```

One can see the toa bits shifting (to remove the 13th bit previously left unused) in V10 w.r.t to V9 , and the boolean bit (from the boolean data member suppressed in V10) appearing at bit 30.

**Validation 2: 10 photon gun events**
10 photon gun events generated in 10_6_x and read form 11_0_0_pre11 [with](https://gitlab.cern.ch/psilva/hgcdigiformatvalidation/blob/master/data/10_6_X_to_11_0_0_pre11_withversioning.diff) and [without](https://gitlab.cern.ch/psilva/hgcdigiformatvalidation/blob/master/data/10_6_X_to_11_0_0_pre11_noversioning.diff) the commits of the schema evolution. Out of 347k `HGCSample` , no differences in toa value and toa valid bit.

franzoni:
please test

franzoni:
we have addressed all comments received so far.
Please let us know if more actions are needed on our part.

franzoni:
please test

kpedro88:
There is a very minor change in the HGCHEB RecHits, which propagates to downstream quantities:
HGCHEB RecHits

Is this expected?

DataFormats/HGCDigi/interface/HGCSample.h
```diff
@@ -71,6 +71,36 @@ class HGCSample {
   bool getToAValid() const { return getWord(kToAValidMask, kToAValidShift); }
   uint32_t operator()() { return value_; }
 
+  /**
+     @short Data Model Evolution
+  */
+  static uint32_t convertToNewFormat(uint32_t valueOldForm, bool toaFiredOldForm) {
```
slava77: shouldn't this have a version in the name?
"New" will become obsolete after another update.

slava77:
+1

for https://github.com/cms-sw/cmssw/pull/28349	2d60768 
- updates since the last reco signoff did not affect code in the reco category https://github.com/cms-sw/cmssw/pull/28349/files/fe44ec559fd7ea5b112f2cb2f0c01e0ddb255366..2d60768da0090f003a39dad7a76b7f79e5e3baf8
- jenkins tests pass and comparisons with the baseline are apparently showing the same kind of differences starting from HGCAL hits

franzoni:
> Is this expected?

we have an understanding. Working on final checks before we report here.

DataFormats/HGCDigi/interface/HGCSample.h
franzoni: addressed.

franzoni:
> Is this expected?

It was not expected.

The [2 hits which change amplitude](https://github.com/cms-sw/cmssw/pull/28349#issuecomment-554122677) in the plot (are bit-clamped, hence they both decrease in value) are only in the scintillator section of the HGCAL; they're the result of : 
1) the [feature](https://images.app.goo.gl/S6PxFx4hxgCZV6NT6),  present since pre12, of using  [13 bits in place of the 12 available](https://github.com/cms-sw/cmssw/blob/CMSSW_11_0_0_pre12/SimCalorimetry/HGCalSimProducers/python/hgcalDigitizer_cfi.py#L238) in `HGCSample`
2) having removed the ceiling of the amplitudes in the `HGCSample` itself, to remove the cpu performance penalty ( we discussed this [earlier in this PR](https://github.com/cms-sw/cmssw/pull/28349#issuecomment-550381238l) and given that the HGCAL digitizer is protected against values beyond pow(2, `adcNbits `)
combined.

How @pfs and I propose to solve this, in agreement with @deguio and @rovere (HGCAL DPG) is:

- set the number bits (``adcNbits ``) used for the amplitude of the scintillator section to 12, which is a must
- halve `adcSaturation_fC` (550->275); that won't change number-of-mips/bit currently stored nor the actual saturation value, which is _de factor_ 275 even in the current master (where `HGCSample` implements the ceiling for values above 4095) despite having chosen 550
- ensure that in the scintillator digitizer, whenever the adc value exceeds the 12 bit coverage, the `HGCSample` is set to 4095

Validation: 
We've generated 20 events with relval 21234; 2 hits in the scintillator exceeded the 4095 and got clamped with this PR - we reused these GEN-SIM for incremental tests, in a situation similar to the plot by @kpedro88 .

At first comparison of { 0x DetId , data in adc counts, toavalid } between [2d60768](https://github.com/cms-sw/cmssw/commit/2d60768da0090f003a39dad7a76b7f79e5e3baf8) and IB-CMSSW_11_0_X_2019-11-14-1100.

```
779c779
< 0xa4163e6c 3235 0
---
> 0xa4163e6c 4095 0
831c831
< 0xa418406c 347 0
---
> 0xa418406c 4095 0
```

We found out that the two hits would have 7331 and 4443 counts resp (which are 4096 more than the value . With the three commits which will follow this post, the value 4095 are restored in `HGCSample`.

In a pull request which will follow this one, we'll reassess the usage of bits and choice of least-significant-bit for the scintillator section.

franzoni:
please test

kpedro88:
+upgrade
thanks @franzoni et al. for the careful investigation
this is an opportune time (before the next TDR production) to standardize HGCal digitization practices

slava77:
+1

for https://github.com/cms-sw/cmssw/pull/28349	4c9cb63 
- changes since the last reco signoff for 2d60768 were outside of the reco category
- jenkins tests pass and the comparisons now show no differences

fabiocos:
@civanch could you please check and comment or sign it?

@franzoni thanks for addressing the issue with test code in bin directory. The backward compatibility issue has been addressed, so this code looks ready for integration (apart for possible residual comments by simulation)

civanch:
+1

fabiocos:
+1

PR #28160, opened 2019-10-11
[10.6.X]  Dynamical APV gain simulation (backport)

mmusich:
#### PR description:
Combined back-port of:
   * https://github.com/cms-sw/cmssw/pull/27823: Simulation of APV dynamic gain
   * https://github.com/cms-sw/cmssw/pull/27941: Add SiStripApvSimulationParameters database object and record
   * https://github.com/cms-sw/cmssw/pull/28028: Add APV simulation in PreMixingSiStripWorker 
   * https://github.com/cms-sw/cmssw/pull/28081: Simulate APV dynamic gain in TID and TEC

#### PR validation:
Validated by cherry-picking commits from PRs #28024 introducing special testing wfs in master
```
cmsrel CMSSW_10_6_X_2019-10-09-1100
cd CMSSW_10_6_X_2019-10-09-1100/src/
git cms-init
git cms-merge-topic CMSTrackerDPG:apv_dynamical_gain_simulation_10_6_X
cmsenv
git cms-addpkg Configuration/PyReleaseValidation
git cherry-pick b48f40164749f2e047c5e688e58eff089190db28
git cherry-pick 6d65ed7acf9dd01f15387469eab33d5508dd16b0
git cherry-pick 4fca3de96745d2b2d5e293133799d5a6d7e7e752
git cherry-pick 2ffa8135eaa8b5e832f0e5910e2f06775280f5a9
git cherry-pick 3ca3003f5770544486c53dfc77371de9b70d3caf
git cherry-pick e9ab9299c489e760642b764a1a40cc795ec99fd5
git cherry-pick 2f88d522865f4243ad0ff0bc591865db660ea575
scramv1 b -j 8 
runTheMatrix.py -l 1325.9,25202.1,250202.2,250202.3 -t 4 -j 8 --ibeos --command='-n 10'
```
it run to completion.

#### if this PR is a backport please specify the original PR:
https://github.com/cms-sw/cmssw/pull/27823, https://github.com/cms-sw/cmssw/pull/27941, https://github.com/cms-sw/cmssw/pull/28028, https://github.com/cms-sw/cmssw/pull/28081

cc:
@EmyrClement @pieterdavid @robervalwalsh @tsusa @mtosi @JanFSchulte

PR #28355, opened 2019-11-06
Activate SiStrip APV Simulation for 2016 MC

mmusich:
#### PR description:

This PR is a follow-up of PRs https://github.com/cms-sw/cmssw/pull/27823, https://github.com/cms-sw/cmssw/pull/27941, https://github.com/cms-sw/cmssw/pull/28028 and https://github.com/cms-sw/cmssw/pull/28081. 
In this PR the SiStrip APV saturation simulation is activated [**EDIT: assuming the 2 MC samples solution is used**]
   * on  ~~53%~~ 100% of the events (ratio of the 2016 luminosity affected by the APV saturation vs the total is 53%).  
   * providing the needed conditions of the APV baselines in the Global Tag (adding to the `run2_mc` & `run2_mc_cosmics` autoCond keys the appropriate tag: `SiStripApvSimulationParameters_2016preVFP_v1`) This corresponds to the APV baselines stored in https://patch-diff.githubusercontent.com/raw/cms-data/SimTracker-SiStripDigitizer/pull/2.patch.
   * introducing an era modifier `strips_vfp30_2016`which is used in `Era_Run2_2016_HIPM` ~~and excluded in `Era_Run2_2016_pA` and `Era_Run2_2017`~~
   * the parameter `includeAPVSimulation` is turned off in the premix stage-1 in accordance to what done for the special relval workflows introduced in  https://github.com/cms-sw/cmssw/pull/28024

#### PR validation:

Extensive technical validation has been carried out by the DPG (see presentation e.g. [here](https://indico.cern.ch/event/841672/contributions/3541970/attachments/1898878/3133702/Emyr_28082019.pdf)) and through production of dedicated RelVals (see [PDMVRELVALS-59]( https://its.cern.ch/jira/projects/PDMVRELVALS/issues/PDMVRELVALS-59)).
I report here few MC / MC comparisons in which 9k TTBar MC events produced with vanilla `CMSSW_11_0_0_pre10` (**black**) are compared with events in which the APV saturation simulation is run on 100% of the cases (**blue**) and with events in which the APV saturation simulation is run together with the HIP mitigation (**orange**). 

image

image

image

####  Follow-up still needed: 

At the moment of the submitting this PR a detailed Data/MC comparison at Tracking level is not yet available (subject to unavailability of the samples requested in [PDMVRELVALS-59](https://its.cern.ch/jira/projects/PDMVRELVALS/issues/PDMVRELVALS-59)), though a very rough data/mc comparison has been compiled [here](https://tinyurl.com/yygthm9n) for the part of the data affected by the issue. Also it is not yet decided if it is necessary or not to run the tracking HIP mitigation over all or part of the events - though this decision shall be taken in agreement to Tracking / other POG preferences.  

This PR is not a backport, but the commits proposed here will need to be included in PR https://github.com/cms-sw/cmssw/pull/28160 (or in a subsequent follow-up if that one is merged first.)
cc:
@pieterdavid @robervalwalsh  @EmyrClement @mtosi @JanFSchulte @tsusa

### DIFF
--- a/Configuration/AlCa/python/autoCond.py
+++ b/Configuration/AlCa/python/autoCond.py
@@ -16,9 +16,9 @@ autoCond = {
     # GlobalTag for MC production with perfectly aligned and calibrated detector for Run2
     'run2_design'       :   '110X_mcRun2_design_v3',
     #GlobalTag for MC production with optimistic alignment and calibrations for Run2
-    'run2_mc'           :   '110X_mcRun2_asymptotic_v3',
+    'run2_mc'           :   '110X_mcRun2_asymptotic_Candidate_2019_11_05_13_17_35',
     # GlobalTag for MC production (cosmics) with starup-like alignment and calibrations for Run2, Strip tracker in peak mode
-    'run2_mc_cosmics'   :   '110X_mcRun2cosmics_startup_deco_v2',
+    'run2_mc_cosmics'   :   '110X_mcRun2cosmics_startup_deco_Candidate_2019_11_05_13_38_13',
     # GlobalTag for MC production (Heavy Ions collisions) with optimistic alignment and calibrations for Run2
     'run2_mc_hi'        :   '110X_mcRun2_HeavyIon_v2',
     # GlobalTag for MC production (p-Pb collisions) with realistic alignment and calibrations for Run2
--- a/Configuration/AlCa/python/autoCond.py
+++ b/Configuration/AlCa/python/autoCond.py
@@ -16,9 +16,9 @@ autoCond = {
     # GlobalTag for MC production with perfectly aligned and calibrated detector for Run2
     'run2_design'       :   '110X_mcRun2_design_v3',
     #GlobalTag for MC production with optimistic alignment and calibrations for Run2
-    'run2_mc'           :   '110X_mcRun2_asymptotic_Candidate_2019_11_05_13_17_35',
+    'run2_mc'           :   '110X_mcRun2_asymptotic_v4',
     # GlobalTag for MC production (cosmics) with starup-like alignment and calibrations for Run2, Strip tracker in peak mode
-    'run2_mc_cosmics'   :   '110X_mcRun2cosmics_startup_deco_Candidate_2019_11_05_13_38_13',
+    'run2_mc_cosmics'   :   '110X_mcRun2cosmics_startup_deco_v3',
     # GlobalTag for MC production (Heavy Ions collisions) with optimistic alignment and calibrations for Run2
     'run2_mc_hi'        :   '110X_mcRun2_HeavyIon_v2',
     # GlobalTag for MC production (p-Pb collisions) with realistic alignment and calibrations for Run2
--- a/Configuration/Eras/python/Era_Run2_2016_HIPM_cff.py
+++ b/Configuration/Eras/python/Era_Run2_2016_HIPM_cff.py
@@ -2,7 +2,6 @@ import FWCore.ParameterSet.Config as cms
 
 from Configuration.Eras.Era_Run2_2016_cff import Run2_2016
 from Configuration.Eras.Modifier_tracker_apv_vfp30_2016_cff import tracker_apv_vfp30_2016
-from Configuration.Eras.Modifier_strips_vfp30_2016_cff import strips_vfp30_2016
 
-Run2_2016_HIPM = cms.ModifierChain(Run2_2016, strips_vfp30_2016, tracker_apv_vfp30_2016)
+Run2_2016_HIPM = cms.ModifierChain(Run2_2016, tracker_apv_vfp30_2016)
 
--- a/Configuration/Eras/python/Era_Run2_2016_HIPM_cff.py
+++ b/Configuration/Eras/python/Era_Run2_2016_HIPM_cff.py
@@ -2,6 +2,7 @@ import FWCore.ParameterSet.Config as cms
 
 from Configuration.Eras.Era_Run2_2016_cff import Run2_2016
 from Configuration.Eras.Modifier_tracker_apv_vfp30_2016_cff import tracker_apv_vfp30_2016
+from Configuration.Eras.Modifier_strips_vfp30_2016_cff import strips_vfp30_2016
 
-Run2_2016_HIPM = cms.ModifierChain(Run2_2016, tracker_apv_vfp30_2016)
+Run2_2016_HIPM = cms.ModifierChain(Run2_2016, strips_vfp30_2016, tracker_apv_vfp30_2016)
 
--- a/Configuration/Eras/python/Era_Run2_2016_cff.py
+++ b/Configuration/Eras/python/Era_Run2_2016_cff.py
@@ -13,5 +13,6 @@ from Configuration.Eras.Modifier_run2_jme_2016_cff import run2_jme_2016
 from Configuration.Eras.Modifier_strips_vfp30_2016_cff import strips_vfp30_2016
 
 Run2_2016 = cms.ModifierChain(run2_common, run2_25ns_specific,
-                              stage2L1Trigger, ctpps_2016, run2_HLTconditions_2016, run2_muon_2016, run2_egamma_2016, run2_L1prefiring, pixel_2016, run2_jme_2016, strips_vfp30_2016)
+                              stage2L1Trigger, ctpps_2016, run2_HLTconditions_2016, run2_muon_2016,
+                              run2_egamma_2016, run2_L1prefiring, pixel_2016, run2_jme_2016, strips_vfp30_2016)
 
--- a/Configuration/Eras/python/Era_Run2_2016_cff.py
+++ b/Configuration/Eras/python/Era_Run2_2016_cff.py
@@ -10,7 +10,8 @@ from Configuration.Eras.Modifier_run2_egamma_2016_cff import run2_egamma_2016
 from Configuration.Eras.Modifier_run2_L1prefiring_cff import run2_L1prefiring
 from Configuration.Eras.Modifier_pixel_2016_cff import pixel_2016
 from Configuration.Eras.Modifier_run2_jme_2016_cff import run2_jme_2016
+from Configuration.Eras.Modifier_strips_vfp30_2016_cff import strips_vfp30_2016
 
 Run2_2016 = cms.ModifierChain(run2_common, run2_25ns_specific,
- stage2L1Trigger, ctpps_2016, run2_HLTconditions_2016, run2_muon_2016, run2_egamma_2016, run2_L1prefiring, pixel_2016, run2_jme_2016)
+                              stage2L1Trigger, ctpps_2016, run2_HLTconditions_2016, run2_muon_2016, run2_egamma_2016, run2_L1prefiring, pixel_2016, run2_jme_2016, strips_vfp30_2016)
 
--- a/Configuration/Eras/python/Era_Run2_2016_cff.py
+++ b/Configuration/Eras/python/Era_Run2_2016_cff.py
@@ -10,7 +10,8 @@ from Configuration.Eras.Modifier_run2_egamma_2016_cff import run2_egamma_2016
 from Configuration.Eras.Modifier_run2_L1prefiring_cff import run2_L1prefiring
 from Configuration.Eras.Modifier_pixel_2016_cff import pixel_2016
 from Configuration.Eras.Modifier_run2_jme_2016_cff import run2_jme_2016
+from Configuration.Eras.Modifier_strips_vfp30_2016_cff import strips_vfp30_2016
 
 Run2_2016 = cms.ModifierChain(run2_common, run2_25ns_specific,
- stage2L1Trigger, ctpps_2016, run2_HLTconditions_2016, run2_muon_2016, run2_egamma_2016, run2_L1prefiring, pixel_2016, run2_jme_2016)
+ stage2L1Trigger, ctpps_2016, run2_HLTconditions_2016, run2_muon_2016, run2_egamma_2016, run2_L1prefiring, pixel_2016, run2_jme_2016, strips_vfp30_2016)
 
--- a/Configuration/Eras/python/Era_Run2_2016_cff.py
+++ b/Configuration/Eras/python/Era_Run2_2016_cff.py
@@ -10,9 +10,7 @@ from Configuration.Eras.Modifier_run2_egamma_2016_cff import run2_egamma_2016
 from Configuration.Eras.Modifier_run2_L1prefiring_cff import run2_L1prefiring
 from Configuration.Eras.Modifier_pixel_2016_cff import pixel_2016
 from Configuration.Eras.Modifier_run2_jme_2016_cff import run2_jme_2016
-from Configuration.Eras.Modifier_strips_vfp30_2016_cff import strips_vfp30_2016
 
 Run2_2016 = cms.ModifierChain(run2_common, run2_25ns_specific,
-                              stage2L1Trigger, ctpps_2016, run2_HLTconditions_2016, run2_muon_2016,
-                              run2_egamma_2016, run2_L1prefiring, pixel_2016, run2_jme_2016, strips_vfp30_2016)
+ stage2L1Trigger, ctpps_2016, run2_HLTconditions_2016, run2_muon_2016, run2_egamma_2016, run2_L1prefiring, pixel_2016, run2_jme_2016)
 
--- a/Configuration/Eras/python/Era_Run2_2016_pA_cff.py
+++ b/Configuration/Eras/python/Era_Run2_2016_pA_cff.py
@@ -2,6 +2,7 @@ import FWCore.ParameterSet.Config as cms
 
 from Configuration.Eras.Era_Run2_2016_cff import Run2_2016
 from Configuration.Eras.Modifier_pA_2016_cff import pA_2016
+from Configuration.Eras.Modifier_strips_vfp30_2016_cff import strips_vfp30_2016
 
-Run2_2016_pA = cms.ModifierChain(Run2_2016, pA_2016)
+Run2_2016_pA = cms.ModifierChain(Run2_2016.copyAndExclude([strips_vfp30_2016]), pA_2016)
 
--- a/Configuration/Eras/python/Era_Run2_2016_pA_cff.py
+++ b/Configuration/Eras/python/Era_Run2_2016_pA_cff.py
@@ -2,7 +2,6 @@ import FWCore.ParameterSet.Config as cms
 
 from Configuration.Eras.Era_Run2_2016_cff import Run2_2016
 from Configuration.Eras.Modifier_pA_2016_cff import pA_2016
-from Configuration.Eras.Modifier_strips_vfp30_2016_cff import strips_vfp30_2016
 
-Run2_2016_pA = cms.ModifierChain(Run2_2016.copyAndExclude([strips_vfp30_2016]), pA_2016)
+Run2_2016_pA = cms.ModifierChain(Run2_2016, pA_2016)
 
--- a/Configuration/PyReleaseValidation/python/relval_steps.py
+++ b/Configuration/PyReleaseValidation/python/relval_steps.py
@@ -1578,7 +1578,7 @@ step2Upg2015Defaults = {'-s'     :'DIGI:pdigi_valid,L1,DIGI2RAW,HLT:@relval2016'
                   }
 step2Upg2015Defaults50ns = merge([{'-s':'DIGI:pdigi_valid,L1,DIGI2RAW,HLT:@relval50ns','--conditions':'auto:run2_mc_50ns','--era':'Run2_50ns'},step2Upg2015Defaults])
 
-step2Upg2015DefaultsAPVSimu = merge([{'--customise': 'SimTracker/SiStripDigitizer/customizeSiStripAPVDynamicGainSimulation.activateSiStripAPVDynamicGain'},step2Upg2015Defaults])
+step2Upg2015DefaultsAPVSimu = merge([{'--era': 'Run2_2016_HIPM'},step2Upg2015Defaults])
 
 #for 2017
 step2Upg2017Defaults = {'-s'     :'DIGI:pdigi_valid,L1,DIGI2RAW,HLT:@relval2017',
@@ -1602,7 +1602,7 @@ step2Upg2018Defaults = {'-s'     :'DIGI:pdigi_valid,L1,DIGI2RAW,HLT:@relval2018'
 
 
 steps['DIGIUP15']=merge([step2Upg2015Defaults])
-steps['DIGIUP15APVSimu']=merge([{'--customise': 'SimTracker/SiStripDigitizer/customizeSiStripAPVDynamicGainSimulation.activateSiStripAPVDynamicGain'},step2Upg2015Defaults])
+steps['DIGIUP15APVSimu']=merge([{'--era': 'Run2_2016_HIPM'},step2Upg2015Defaults])
 
 steps['DIGIUP15PROD1']=merge([{'-s':'DIGI,L1,DIGI2RAW,HLT:@relval2016','--eventcontent':'RAWSIM','--datatier':'GEN-SIM-RAW'},step2Upg2015Defaults])
 steps['DIGIUP15_PU25']=merge([PU25,step2Upg2015Defaults])
@@ -1741,9 +1741,9 @@ digiPremixLocalPileupUp2018Defaults25ns = merge([digiPremixLocalPileup,
                                                  digiPremixUp2018Defaults25ns])
 
 steps['DIGIPRMXUP15_PU25']=merge([digiPremixUp2015Defaults25ns])
-steps['DIGIPRMXUP15APVSimu_PU25']=merge([{'--customise': 'SimTracker/SiStripDigitizer/customizeSiStripAPVDynamicGainSimulation.activateSiStripAPVDynamicGain'},digiPremixUp2015Defaults25ns])
+steps['DIGIPRMXUP15APVSimu_PU25']=merge([{'--era': 'Run2_2016_HIPM'},digiPremixUp2015Defaults25ns])
 steps['DIGIPRMXLOCALUP15_PU25']=merge([digiPremixLocalPileupUp2015Defaults25ns])
-steps['DIGIPRMXLOCALUP15APVSimu_PU25']=merge([{'--customise': 'SimTracker/SiStripDigitizer/customizeSiStripAPVDynamicGainSimulation.activateSiStripAPVDynamicGain'},digiPremixLocalPileupUp2015Defaults25ns])
+steps['DIGIPRMXLOCALUP15APVSimu_PU25']=merge([{'--era': 'Run2_2016_HIPM'},digiPremixLocalPileupUp2015Defaults25ns])
 
 steps['DIGIPRMXUP15_PU50']=merge([digiPremixUp2015Defaults50ns])
 steps['DIGIPRMXUP17_PU25']=merge([digiPremixUp2017Defaults25ns])
--- a/PhysicsTools/PatAlgos/python/slimming/MicroEventContent_cff.py
+++ b/PhysicsTools/PatAlgos/python/slimming/MicroEventContent_cff.py
@@ -144,6 +144,10 @@ MicroEventContentMC.outputCommands += [
                                         'keep L1GtTriggerMenuLite_l1GtTriggerMenuLite__*'
                                       ]
 
+from Configuration.Eras.Modifier_strips_vfp30_2016_cff import strips_vfp30_2016
+strips_vfp30_2016.toModify(MicroEventContentMC, outputCommands = MicroEventContentMC.outputCommands + [
+    'keep *_*_SimulatedAPVDynamicGain_*'
+])
 
 MiniAODOverrideBranchesSplitLevel = cms.untracked.VPSet( [
 cms.untracked.PSet(branch = cms.untracked.string("patPackedCandidates_packedPFCandidates__*"),splitLevel=cms.untracked.int32(99)),
--- a/PhysicsTools/PatAlgos/python/slimming/MicroEventContent_cff.py
+++ b/PhysicsTools/PatAlgos/python/slimming/MicroEventContent_cff.py
@@ -146,7 +146,7 @@ MicroEventContentMC.outputCommands += [
 
 from Configuration.Eras.Modifier_strips_vfp30_2016_cff import strips_vfp30_2016
 strips_vfp30_2016.toModify(MicroEventContentMC, outputCommands = MicroEventContentMC.outputCommands + [
-    'keep *_*_SimulatedAPVDynamicGain_*'
+    'keep *_mix_SimulatedAPVDynamicGain_*'
 ])
 
 MiniAODOverrideBranchesSplitLevel = cms.untracked.VPSet( [
--- a/PhysicsTools/PatAlgos/python/slimming/MicroEventContent_cff.py
+++ b/PhysicsTools/PatAlgos/python/slimming/MicroEventContent_cff.py
@@ -146,7 +146,7 @@ MicroEventContentMC.outputCommands += [
 
 from Configuration.Eras.Modifier_strips_vfp30_2016_cff import strips_vfp30_2016
 strips_vfp30_2016.toModify(MicroEventContentMC, outputCommands = MicroEventContentMC.outputCommands + [
-    'keep *_mix_SimulatedAPVDynamicGain_*'
+    'keep *_simAPVsaturation_SimulatedAPVDynamicGain_*'
 ])
 
 MiniAODOverrideBranchesSplitLevel = cms.untracked.VPSet( [
--- a/SimGeneral/MixingModule/python/stripDigitizer_cfi.py
+++ b/SimGeneral/MixingModule/python/stripDigitizer_cfi.py
@@ -23,7 +23,7 @@ premix_stage2.toModify(stripDigitizer,
     Noise = False, # will be added in DataMixer
     PreMixingMode = True, #Special mode to save all hit strips
     FedAlgorithm = 5, # special ZS mode: accept adc>0
-    includeAPVSimulation = False  # APV simulation is off for premix stage1
+    includeAPVSimulation = False  # APV simulation is off for the MixingModule configuration in premix stage2
 )
 
 from Configuration.Eras.Modifier_phase2_tracker_cff import phase2_tracker
--- a/SimGeneral/MixingModule/python/stripDigitizer_cfi.py
+++ b/SimGeneral/MixingModule/python/stripDigitizer_cfi.py
@@ -23,6 +23,7 @@ premix_stage2.toModify(stripDigitizer,
     Noise = False, # will be added in DataMixer
     PreMixingMode = True, #Special mode to save all hit strips
     FedAlgorithm = 5, # special ZS mode: accept adc>0
+    includeAPVSimulation = False  # APV simulation is off for premix stage1
 )
 
 from Configuration.Eras.Modifier_phase2_tracker_cff import phase2_tracker
--- a/SimTracker/Configuration/python/SimTracker_EventContent_cff.py
+++ b/SimTracker/Configuration/python/SimTracker_EventContent_cff.py
@@ -66,5 +66,5 @@ phase2_tracker.toModify(SimTrackerPREMIX, outputCommands = [
 from Configuration.Eras.Modifier_strips_vfp30_2016_cff import strips_vfp30_2016
 for _entry in [SimTrackerFEVTDEBUG,SimTrackerDEBUG,SimTrackerRAW,SimTrackerRECO,SimTrackerAOD,SimTrackerPREMIX]:
     strips_vfp30_2016.toModify(_entry, outputCommands = _entry.outputCommands + [
-        'keep *_*_SimulatedAPVDynamicGain_*'
+        'keep *_mix_SimulatedAPVDynamicGain_*'
     ])
--- a/SimTracker/Configuration/python/SimTracker_EventContent_cff.py
+++ b/SimTracker/Configuration/python/SimTracker_EventContent_cff.py
@@ -62,3 +62,9 @@ phase2_tracker.toModify(SimTrackerPREMIX, outputCommands = [
         'keep *_*_Phase2OTDigiSimLink_*',
         'keep *_simSiPixelDigis_*_*', # covers digis and digiSimLinks
 ])
+
+from Configuration.Eras.Modifier_strips_vfp30_2016_cff import strips_vfp30_2016
+for _entry in [SimTrackerFEVTDEBUG,SimTrackerDEBUG,SimTrackerRAW,SimTrackerRECO,SimTrackerAOD,SimTrackerPREMIX]:
+    strips_vfp30_2016.toModify(_entry, outputCommands = _entry.outputCommands + [
+        'keep *_*_SimulatedAPVDynamicGain_*'
+    ])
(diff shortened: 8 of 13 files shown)


christopheralanwest:
please test

mmusich:
in the last commit - as discussed in the [last PPD general meeting](https://indico.cern.ch/event/856064/#11-trackertracking-for-ul16-sl) the knowledge if the apv dynamic gain simulation has been run or not is persistent in the event.

perrotta:
please test

PhysicsTools/PatAlgos/python/slimming/MicroEventContent_cff.py
```diff
@@ -144,6 +144,10 @@
                                         'keep L1GtTriggerMenuLite_l1GtTriggerMenuLite__*'
                                       ]
 
+from Configuration.Eras.Modifier_strips_vfp30_2016_cff import strips_vfp30_2016
+strips_vfp30_2016.toModify(MicroEventContentMC, outputCommands = MicroEventContentMC.outputCommands + [
+    'keep *_*_SimulatedAPVDynamicGain_*'
```
slava77: please use specific module instance names

Configuration/Eras/python/Era_Run2_2016_pA_cff.py
```diff
@@ -2,6 +2,7 @@
 
 from Configuration.Eras.Era_Run2_2016_cff import Run2_2016
 from Configuration.Eras.Modifier_pA_2016_cff import pA_2016
+from Configuration.Eras.Modifier_strips_vfp30_2016_cff import strips_vfp30_2016
 
-Run2_2016_pA = cms.ModifierChain(Run2_2016, pA_2016)
+Run2_2016_pA = cms.ModifierChain(Run2_2016.copyAndExclude([strips_vfp30_2016]), pA_2016)
```
slava77: I'm not sure if it matters here.
pA_2016 is used only in reco; was the DIGI step made with pA as well?
Perhaps it's safer to remove, anyways.

slava77 (commented):
this looks backwards: the problem is enabled in MC by default specifically where the reco is not set to react to it as in Run2_2016_HIPM using tracker_apv_vfp30_2016, which is the case in data.
At least this is coherently picked up in Run2_2016_HIPM.

My understanding of the previous discussion was that without proper handling of this in reco on event by event we were left with just an option of a small sample with APV inefficiency simulation+mitigation ;  not with an option to by default enable the problem in sim without a mitigation in reco.

Should the strips_vfp30_2016 be also removed in Era_Run2_2016_trackingLowPU_cff? IIRC, this era applied also to the older data, but then I also recall that the APV issue is much less visible at low PU where this era is applied. Does the simulation account for pileup/occupancies automatically?

slava77:
> My understanding of the previous discussion was that without proper handling of this in reco on event by event we were left with just an option of a small sample with APV inefficiency simulation+mitigation ; not with an option to by default enable the problem in sim without a mitigation in reco.

Actually, considering how many analyses will eventually need this MC (not many, which in turn means not so many samples), perhaps it's more appropriate to have two sets of MC. The way we treat the eras in rerecos, Run2_2016 should not have the saturation problem, which should be covered with Run2_2016_HIPM.

Configuration/Eras/python/Era_Run2_2016_cff.py
```diff
@@ -10,7 +10,8 @@
 from Configuration.Eras.Modifier_run2_L1prefiring_cff import run2_L1prefiring
 from Configuration.Eras.Modifier_pixel_2016_cff import pixel_2016
 from Configuration.Eras.Modifier_run2_jme_2016_cff import run2_jme_2016
+from Configuration.Eras.Modifier_strips_vfp30_2016_cff import strips_vfp30_2016
 
 Run2_2016 = cms.ModifierChain(run2_common, run2_25ns_specific,
- stage2L1Trigger, ctpps_2016, run2_HLTconditions_2016, run2_muon_2016, run2_egamma_2016, run2_L1prefiring, pixel_2016, run2_jme_2016)
+                              stage2L1Trigger, ctpps_2016, run2_HLTconditions_2016, run2_muon_2016, run2_egamma_2016, run2_L1prefiring, pixel_2016, run2_jme_2016, strips_vfp30_2016)
```
slava77: please add some linebreaks to keep lines shorter than ~100 chars

PhysicsTools/PatAlgos/python/slimming/MicroEventContent_cff.py
slava77: BTW, is this collection always produced now, or just with this strips_vfp30_2016 era?
It may be easier to just add to the default ```keep``` list to keep this config simpler.

Configuration/Eras/python/Era_Run2_2016_pA_cff.py
mmusich: as far as I can see in the only relval using the era `Run2_2016_pA` in release the era is used in all steps: (including digitization):
```
281.0  EPOS_PPb_8160GeV_MinimumBias+DIGIUP15_PPb+RECOUP15_PPb+HARVESTUP15_PPb [1]: cmsDriver.py ReggeGribovPartonMC_EposLHC_4080_4080GeV_pPb_cfi  --conditions auto:run2_mc_pa -n 10 --era Run2_2016_pA --eventcontent RAWSIM --relval 9000,100 -s GEN,SIM --datatier GEN-SIM --beamspot RealisticPPbBoost8TeV2016Collision 
                                           [2]: cmsDriver.py step2  --conditions auto:run2_mc_pa -s DIGI:pdigi_valid,L1,DIGI2RAW,HLT:PIon --datatier GEN-SIM-DIGI-RAW-HLTDEBUG -n 10 --era Run2_2016_pA --eventcontent FEVTDEBUGHLT
                                           [3]: cmsDriver.py step3  --runUnscheduled  --conditions auto:run2_mc_pa -s RAW2DIGI,L1Reco,RECO,ALCA:TkAlMinBias+TkAlMuonIsolatedPA+TkAlUpsilonMuMuPA+TkAlZMuMuPA,EI,VALIDATION,DQM --datatier AODSIM,DQMIO -n 10 --era Run2_2016_pA --eventcontent AODSIM,DQM
                                           [4]: cmsDriver.py step4  --filetype DQM --mc  --conditions auto:run2_mc_pa --era Run2_2016_pA -s HARVESTING:validationHarvestingNoHLT+dqmHarvestingFakeHLT -n 100
```

PhysicsTools/PatAlgos/python/slimming/MicroEventContent_cff.py
mmusich: I added the module instance name.
>  is this collection always produced now, or just with this strips_vfp30_2016 era?

it is produced always, but it is not relevant in any other era.

>  It may be easier to just add to the default keep list to keep this config simpler.

Perhaps so, but as this eventually needs to be backported to 10.6.X I am not sure one can change the event content of the production UL release **also** for the years already processed. Maybe @fabiocos can clarify the stance on this.

mmusich:
@slava77 few answers on some of the points:

> this looks backwards: the problem is enabled in MC by default specifically where the reco is not set to react to it as in Run2_2016_HIPM using tracker_apv_vfp30_2016, which is the case in data.
At least this is coherently picked up in Run2_2016_HIPM.

it is implemented in a somewhat backwards way (for the moment) as it is not fully clear how to handle the situation. 
I limited myself to implement the proposal we presented [here](https://indico.cern.ch/event/856064/contributions/3632500/attachments/1940818/3218153/191107_TRKPOG.pdf), slide n. 11

> Actually, considering how many analyses will eventually need this MC (not many, which in turn means not so many samples), perhaps it's more appropriate to have two sets of MC. The way we treat the eras in rerecos, Run2_2016 should not have the saturation problem, which should be covered with Run2_2016_HIPM
 
Two sets of MC (dealt with two different eras, say add the `strips_vfp30_2016` modifier to `Run2_2016_HIPM` era and use that for both digitization **and** reconstruction steps would guarantee that if CMS produces a MC with these settings the APV simulation and the HIP mitigation are in synch by construction) is also my preferred solution, though this option has been met with negative feedback from x-pog and PPD (too much work to keep the books right for two separate MC campaigns, with implications on era-dependent scale factors, etc. etc.). @srimanob

 > Should the strips_vfp30_2016 be also removed in Era_Run2_2016_trackingLowPU_cff? IIRC, this era applied also to the older data, but then I also recall that the APV issue is much less visible at low PU where this era is applied. Does the simulation account for pileup/occupancies automatically?

It does, e.g. it is active also on cosmics by default but it doesn't degrade the performance. If there is preference I can remove it entirely from that era.

fabiocos:
please test

srimanob:
@mmusich @slava77 
It seems that having 2 MCs is the most possible option at the moment if we (CMS, Physics) accept to have 4 years of UL (instead of 3). This comes with the cheapest in the development of infrastructure, but expensive in terms of operation, book-keeping. 

It's not clear to me if this is the case, it means all POGs will need to do extra SF/JEC for the additional era.

Do I understand correctly from this PR that with era, we can handle both cases properly? Then we may need to discuss on the autocond, which scenario should be "run2_mc". Or we should introduce run2_mc_APV (or others) to reflect 2 eras.

mmusich:
@srimanob 

> Do I understand correctly from this PR that with era, we can handle both cases properly?

no, at the moment we are just activating the simulation on the whole 2016 and randomly selecting ~ one event out of two to undergo the APV simulation. The HIP mitigation is not addressed. 
On the other hand I can modify the PR in such a way (e.g. implementing what described in https://github.com/cms-sw/cmssw/pull/28355#issuecomment-552209789) that both cases are handled properly.

> Then we may need to discuss on the autocond, which scenario should be "run2_mc". Or we should introduce run2_mc_APV (or others) to reflect 2 eras.

I don't understand the need of having 2 Global Tags to reflect the two eras. The same GT can be used for both and the feature be steered by configuration.

SimTracker/Configuration/python/SimTracker_EventContent_cff.py
```diff
@@ -62,3 +62,9 @@
         'keep *_*_Phase2OTDigiSimLink_*',
         'keep *_simSiPixelDigis_*_*', # covers digis and digiSimLinks
 ])
+
+from Configuration.Eras.Modifier_strips_vfp30_2016_cff import strips_vfp30_2016
+for _entry in [SimTrackerFEVTDEBUG,SimTrackerDEBUG,SimTrackerRAW,SimTrackerRECO,SimTrackerAOD,SimTrackerPREMIX]:
+    strips_vfp30_2016.toModify(_entry, outputCommands = _entry.outputCommands + [
+        'keep *_mix_SimulatedAPVDynamicGain_*'
```
makortel: The `mix` label points to the `MixingModule` that does the digitization for the hard scatter event (~equivalent to the treatment of pileup in "stage 1"). Do you want to persist the product from that, or from the hard scatter + pileup overlay? For the latter the label is `mixData`.

(Naively I'd expect you'd want `mixData` here, and add `includeAPVSimulation=False` to
https://github.com/cms-sw/cmssw/blob/09562bc09c37b7207bb3f956b2834d9903572ea0/SimGeneral/MixingModule/python/stripDigitizer_cfi.py#L16-L26
In retrospect a comment along "these customizations should be replicated into stripDigitizer_cfi " in `SiStripSimParameters_cfi.py` above the `premix_stage1.toModify()` would have provided at least a reminder, even though a less error prone mechanism would be much better #26001.)

mmusich: @makortel, thanks for the comment, but please bear with me, I don't understand the suggestion in this jungle of configuration files.
* in the premixing case the APV simulation should be run *only* in the stage2. Why shall I set it false in `stripDigitizer_cfi`? I actually **want** it to be run in the stage2
* about `mix` or `mixdata`, as the APV simulation is run only in the stage2 I think we should persist the product from the hard scatter + pileup overlay. On the other hand it's a pity that the the name of the module is different in the premix case as now I need to speciate it for the premix case.

makortel: > please bear with me, I don't understand the suggestion in this jungle of configuration files.

No problem, it's a bit complicated.

> * in the premixing case the APV simulation should be run _only_ in the stage2. Why shall I set it false in `stripDigitizer_cfi`? I actually **want** it to be run in the stage2

Premixing has three "mixing" actors
* in "stage 1" job (producing the pileup library)
  * 1: `MixingModule` to accumulate the data for pileup (label `mix`)
* in "stage 2" job (overlaying the pileup over signal
  * 2: `MixingModule` to accumulate the data for signal (label `mix`)
  * 3: `PreMixingModule` to overlay the pileup on top of signal (label `mixData`)

The two `MixingModule` instances need to be configured as for "stage 1". For `MixingModule` 1 the customization is in 
https://github.com/cms-sw/cmssw/blob/09562bc09c37b7207bb3f956b2834d9903572ea0/SimGeneral/MixingModule/python/SiStripSimParameters_cfi.py#L130-L134
and for `MixingModule` 2 the customization is in 
https://github.com/cms-sw/cmssw/blob/09562bc09c37b7207bb3f956b2834d9903572ea0/SimGeneral/MixingModule/python/stripDigitizer_cfi.py#L22-L26 
(actually one option to simplify would be to move also the former customization into `stripDigitizer_cfi`)

The `PremixingModule` (3) is then configured as for "stage 2", and there the `SiStripSimBlock` is used as-is
https://github.com/cms-sw/cmssw/blob/09562bc09c37b7207bb3f956b2834d9903572ea0/SimGeneral/PreMixingModule/python/mixOne_premix_on_sim_cfi.py#L62-L64

> On the other hand it's a pity that the the name of the module is different in the premix case as now I need to speciate it for the premix case.

I fully agree...

slava77: >  On the other hand it's a pity that the the name of the module is different in the premix case as now I need to speciate it for the premix case.

An EDAlias can be defined to resolve this problem

mmusich: @makortel, thanks for the detailed explanation, it's much clearer now. I've implemented part of the suggestion in https://github.com/cms-sw/cmssw/pull/28355/commits/1e3c88cee5fda75042ba742fda32bd4c1af16fa3.

@slava77 I am afraid I am not really familiar with the `EDAlias`-ing. Shall an appropriate line be added here:
https://github.com/cms-sw/cmssw/blob/e626860d26692de5880c52c7c80aec7b859a0c60/SimGeneral/MixingModule/python/aliases_PreMix_cfi.py#L36
?

mmusich:
The last commit reverts the changes done in `Era_Run2_2016` and adds the modifier `strips_vfp30_2016` to the era `Era_Run2_2016_HIPM` assuming that, this one is used to process one chunk of MC samples with APV simulation ON and HIP mitigation ON, while the other part is processed with  `Era_Run2_2016` with both features inactive.

SimTracker/Configuration/python/SimTracker_EventContent_cff.py
makortel: > Shall an appropriate line be added here:
> https://github.com/cms-sw/cmssw/blob/e626860d26692de5880c52c7c80aec7b859a0c60/SimGeneral/MixingModule/python/aliases_PreMix_cfi.py#L36
> 
> 
> ?

I'd advise against using the (otherwise natural choice of) `simSiStripDigis` because I'm afraid that would make the already-confusing product naming even more confusing: as of today the `simSiStripDigis` is an alias for products of `mix`, i.e. from the "hard-scatter `MixingModule`". (and yes, the aliases really should be restructured such that downstream would see only `simSiStripDigis` for both classical mixing and premixing).

You could add a new alias though such that in `SimGeneral/MixingModule/python/aliases_cfi.py` the alias points to product from `mix` and in `SimGeneral/MixingModule/python/aliases_PreMix_cfi.py` it points to `mixData`. Then if/when #26001 gets resolved, the labels could be unified (resolving #26001 would anyway break nearly all product labels in DIGI).

christopheralanwest:
please test

Configuration/Eras/python/Era_Run2_2016_HIPM_cff.py
```diff
@@ -2,6 +2,7 @@
 
 from Configuration.Eras.Era_Run2_2016_cff import Run2_2016
 from Configuration.Eras.Modifier_tracker_apv_vfp30_2016_cff import tracker_apv_vfp30_2016
+from Configuration.Eras.Modifier_strips_vfp30_2016_cff import strips_vfp30_2016
 
-Run2_2016_HIPM = cms.ModifierChain(Run2_2016, tracker_apv_vfp30_2016)
+Run2_2016_HIPM = cms.ModifierChain(Run2_2016, strips_vfp30_2016, tracker_apv_vfp30_2016)
```
slava77: what is the motivation to have a separate modifier for the same purpose?
I think that we can just stay with the existing ```tracker_apv_vfp30_2016``` or if so desired, rename it to ```strips_vfp30_2016```

SimTracker/Configuration/python/SimTracker_EventContent_cff.py
mmusich: thanks @makortel I've tried my understanding of your proposed solution in this commit: https://github.com/cms-sw/cmssw/commit/435508be147e9ff70cf365d88be72dc1866e6f9e 
though while it seems to work fine for the standard mixing case, I don't see it working for the pre-mixing. Am I missing something trivial?

mmusich: actually, it seems that in the way it is configured now, the product is still coming from `mix` rather than `mixdata` (puzzled)

mmusich:
In the last commits:
   * the APV simulation steering is moved to the `tracker_apv_vfp30_2016` modifier
   * the `strips_vfp30_2016` is kept only in order to persist exclusively in the `Era_Run2_2016` the information if the APV simulation has been run or not (we can remove this modifier completely , depending on the outcome of the discussion at: https://github.com/cms-sw/cmssw/pull/28355#discussion_r344445384
   * the RelVals I had introduced in https://github.com/cms-sw/cmssw/pull/28024 are modified in order to use `Run2_2016_HIPM` instead of `--customise': 'SimTracker/SiStripDigitizer/customizeSiStripAPVDynamicGainSimulation.activateSiStripAPVDynamicGain'`
   * an `EDAlias` is introduced in both `aliases_cfi.py` and `aliases_PreMix_cfi.py` in order to change the product label name to `simAPVsaturation`. `keep` statements in the event content are modified accordingly. 

I checked that the boolean flag is persisted in all steps in 3 different workflows:

* `25202.2`

```bash
[25202.2_TTbar_13+TTbar_13+DIGIUP15APVSimu_PU25+RECOUP15_PU25_HIPM+HARVESTUP15_PU25]$ edmEventSize -v -a step2.root | grep APV
bool_simAPVsaturation_SimulatedAPVDynamicGain_HLT. 346.2 346.2
[25202.2_TTbar_13+TTbar_13+DIGIUP15APVSimu_PU25+RECOUP15_PU25_HIPM+HARVESTUP15_PU25]$ edmEventSize -v -a step3.root | grep APV
bool_simAPVsaturation_SimulatedAPVDynamicGain_HLT. 247 247
[25202.2_TTbar_13+TTbar_13+DIGIUP15APVSimu_PU25+RECOUP15_PU25_HIPM+HARVESTUP15_PU25]$ edmEventSize -v -a step3_inMINIAODSIM.root  | grep APV
bool_simAPVsaturation_SimulatedAPVDynamicGain_HLT. 247 247
```
* `250202.4`
```bash
[250202.4_TTbar_13+TTbar_13+DIGIPRMXUP15APVSimu_PU25+RECOPRMXUP15_HIPM_PU25+HARVESTUP15_PU25]$ edmEventSize -v -a step2.root | grep simAPV
bool_simAPVsaturation_SimulatedAPVDynamicGain_HLT. 346.2 346.2
[250202.4_TTbar_13+TTbar_13+DIGIPRMXUP15APVSimu_PU25+RECOPRMXUP15_HIPM_PU25+HARVESTUP15_PU25]$ edmEventSize -v -a step3.root | grep simAPV
bool_simAPVsaturation_SimulatedAPVDynamicGain_HLT. 271.8 271.8
[250202.4_TTbar_13+TTbar_13+DIGIPRMXUP15APVSimu_PU25+RECOPRMXUP15_HIPM_PU25+HARVESTUP15_PU25]$ edmEventSize -v -a step3_inMINIAODSIM.root  | grep simAPV
bool_simAPVsaturation_SimulatedAPVDynamicGain_HLT. 247 247
```

* `250202.5`
```
[250202.5_TTbar_13+TTbar_13+PREMIXUP15_PU25+DIGIPRMXLOCALUP15APVSimu_PU25+RECOPRMXUP15_HIPM_PU25+HARVESTUP15_PU25]$ edmEventSize -v -a step2.root | grep simAPV               
bool_simAPVsaturation_SimulatedAPVDynamicGain_DIGI. 348.4 348.4
[250202.5_TTbar_13+TTbar_13+PREMIXUP15_PU25+DIGIPRMXLOCALUP15APVSimu_PU25+RECOPRMXUP15_HIPM_PU25+HARVESTUP15_PU25]$ edmEventSize -v -a step3.root | grep simAPV
bool_simAPVsaturation_SimulatedAPVDynamicGain_HLT. 346.2 346.2
[250202.5_TTbar_13+TTbar_13+PREMIXUP15_PU25+DIGIPRMXLOCALUP15APVSimu_PU25+RECOPRMXUP15_HIPM_PU25+HARVESTUP15_PU25]$ edmEventSize -v -a step4.root  | grep simAPV
bool_simAPVsaturation_SimulatedAPVDynamicGain_HLT. 247 247
[250202.5_TTbar_13+TTbar_13+PREMIXUP15_PU25+DIGIPRMXLOCALUP15APVSimu_PU25+RECOPRMXUP15_HIPM_PU25+HARVESTUP15_PU25]$ edmEventSize -v -a step4_inMINIAODSIM.root  | grep simAPV
bool_simAPVsaturation_SimulatedAPVDynamicGain_HLT. 247 247
```

Open point: it's not clear to me why the `EDAlias` for the premixing case, if the old module label `mixData` is used doesn't work properly.

Configuration/Eras/python/Era_Run2_2016_cff.py
```diff
@@ -10,7 +10,8 @@
 from Configuration.Eras.Modifier_run2_L1prefiring_cff import run2_L1prefiring
 from Configuration.Eras.Modifier_pixel_2016_cff import pixel_2016
 from Configuration.Eras.Modifier_run2_jme_2016_cff import run2_jme_2016
+from Configuration.Eras.Modifier_strips_vfp30_2016_cff import strips_vfp30_2016
 
 Run2_2016 = cms.ModifierChain(run2_common, run2_25ns_specific,
- stage2L1Trigger, ctpps_2016, run2_HLTconditions_2016, run2_muon_2016, run2_egamma_2016, run2_L1prefiring, pixel_2016, run2_jme_2016)
+ stage2L1Trigger, ctpps_2016, run2_HLTconditions_2016, run2_muon_2016, run2_egamma_2016, run2_L1prefiring, pixel_2016, run2_jme_2016, strips_vfp30_2016)
```
slava77: I'm confused why we still have modifications in this file.
Wasn't the strips_vfp30_2016 supposed to be removed completely?

mmusich: Hi @slava77 quoting my previous comment:
* the `strips_vfp30_2016` is kept only in order to persist exclusively in the Era_Run2_2016 the information if the APV simulation has been run or not (we can remove this modifier completely , depending on the outcome of the discussion at: #28355 (comment) )

slava77: ah, thanks for clarifying.
But then it seems like this mod is not removed from Era_Run2_2017.
If the product is just a single float or int, it seems more straightforward to just always save it.

makortel:
@mmusich 

> Open point: it's not clear to me why the `EDAlias` for the premixing case, if the old module label `mixData` is used doesn't work properly.

I'm in the process of taking a look myself, but could the cause be that in premixing the product has an instance label of `siStripDigisDMSimulatedAPVDynamicGain` instead of plain `SimulatedAPVDynamicGain`?
https://github.com/cms-sw/cmssw/blob/e626860d26692de5880c52c7c80aec7b859a0c60/SimTracker/SiStripDigitizer/plugins/PreMixingSiStripWorker.cc#L176
https://github.com/cms-sw/cmssw/blob/e626860d26692de5880c52c7c80aec7b859a0c60/SimGeneral/PreMixingModule/python/mixOne_premix_on_sim_cfi.py#L72

mmusich:
@makortel thanks, you are right!
with these changes:
```diff
diff --git a/PhysicsTools/PatAlgos/python/slimming/MicroEventContent_cff.py b/PhysicsTools/PatAlgos/python/slimming/MicroEventContent_cff.py
index 95186b80942..c42b4eac91b 100644
--- a/PhysicsTools/PatAlgos/python/slimming/MicroEventContent_cff.py
+++ b/PhysicsTools/PatAlgos/python/slimming/MicroEventContent_cff.py
@@ -146,7 +146,7 @@ MicroEventContentMC.outputCommands += [
 
 from Configuration.Eras.Modifier_strips_vfp30_2016_cff import strips_vfp30_2016
 strips_vfp30_2016.toModify(MicroEventContentMC, outputCommands = MicroEventContentMC.outputCommands + [
-    'keep *_simAPVsaturation_SimulatedAPVDynamicGain_*'
+    'keep *_simAPVsaturation_*_*'
 ])
 
 MiniAODOverrideBranchesSplitLevel = cms.untracked.VPSet( [
diff --git a/SimGeneral/MixingModule/python/aliases_PreMix_cfi.py b/SimGeneral/MixingModule/python/aliases_PreMix_cfi.py
index 465e0e67cf3..7e8f5915eba 100644
--- a/SimGeneral/MixingModule/python/aliases_PreMix_cfi.py
+++ b/SimGeneral/MixingModule/python/aliases_PreMix_cfi.py
@@ -50,7 +50,7 @@ genPUProtons = cms.EDAlias(
 )
 
 simAPVsaturation = cms.EDAlias(
-    mix = cms.VPSet(
+    mixData = cms.VPSet(
         cms.PSet(type = cms.string('bool'))
     )
 )
diff --git a/SimTracker/Configuration/python/SimTracker_EventContent_cff.py b/SimTracker/Configuration/python/SimTracker_EventContent_cff.py
index 6f0a1f2c1b3..0cfda6a699e 100644
--- a/SimTracker/Configuration/python/SimTracker_EventContent_cff.py
+++ b/SimTracker/Configuration/python/SimTracker_EventContent_cff.py
@@ -66,5 +66,5 @@ phase2_tracker.toModify(SimTrackerPREMIX, outputCommands = [
 from Configuration.Eras.Modifier_strips_vfp30_2016_cff import strips_vfp30_2016
 for _entry in [SimTrackerFEVTDEBUG,SimTrackerDEBUG,SimTrackerRAW,SimTrackerRECO,SimTrackerAOD,SimTrackerPREMIX]:
     strips_vfp30_2016.toModify(_entry, outputCommands = _entry.outputCommands + [
-        'keep *_simAPVsaturation_SimulatedAPVDynamicGain_*'
+        'keep *_simAPVsaturation_*_*'
     ])
```
I get: 
```
bool_simAPVsaturation_siStripDigisDMSimulatedAPVDynamicGain_HLT. 377 377
```

Now this solves the issue, but it's not really elegant to have two different product names in the std mixing and premixing case. How do you make the two even?

Configuration/Eras/python/Era_Run2_2016_cff.py
mmusich: thanks for catching. Addressed in https://github.com/cms-sw/cmssw/pull/28355/commits/b91964f7baa964fae3a1b96a6e541b50db5b3c28

makortel:
@mmusich 
> Now this solves the issue, but it's not really elegant to have two different product names in the std mixing and premixing case. How do you make the two even?

Change the alias definition in `aliases_PreMix_cfi.py` to
```python
simAPVsaturation = cms.EDAlias(
    mixData = cms.VPSet(
        cms.PSet(
            type = cms.string('bool'),
            fromProductInstance = cms.string('siStripDigisDMSimulatedAPVDynamicGain'),
            toProductInstance = cms.string('SimulatedAPVDynamicGain'),
        ),
    )
)
```

mmusich:
> Change the alias definition in aliases_PreMix_cfi.py 

@makortel, addressed in https://github.com/cms-sw/cmssw/pull/28355/commits/ef7e755092e51e0c215bf9451a5f36c6dfc2ca70, thanks for the support (and the patience).

Configuration/AlCa/python/autoCond.py
```diff
@@ -16,9 +16,9 @@
     # GlobalTag for MC production with perfectly aligned and calibrated detector for Run2
     'run2_design'       :   '110X_mcRun2_design_v3',
     #GlobalTag for MC production with optimistic alignment and calibrations for Run2
-    'run2_mc'           :   '110X_mcRun2_asymptotic_v3',
+    'run2_mc'           :   '110X_mcRun2_asymptotic_Candidate_2019_11_05_13_17_35',
```
chayanit: @christopheralanwest , shall we have proper GT tag name instead?

tlampen: Here are new Global Tags which can be used instead of the candidates:

110X_mcRun2_asymptotic_v4 (replaces110X_mcRun2_asymptotic_Candidate_2019_11_05_13_17_35)

and

110X_mcRun2cosmics_startup_deco_v3 (replaces 110X_mcRun2cosmics_startup_deco_Candidate_2019_11_05_13_38_13)

diffs are here:
https://cms-conddb.cern.ch/cmsDbBrowser/diff/Prod/gts/110X_mcRun2_asymptotic_v4/110X_mcRun2_asymptotic_Candidate_2019_11_05_13_17_35
https://cms-conddb.cern.ch/cmsDbBrowser/diff/Prod/gts/110X_mcRun2cosmics_startup_deco_v3/110X_mcRun2cosmics_startup_deco_Candidate_2019_11_05_13_38_13

pieterdavid: Thanks, I just pushed a commit to update them in autoCond.py

christopheralanwest:
please test

pieterdavid:
The last commit should fix the problems with fast sim in the tests (thanks to @tsusa for finding the problem, and @robervalwalsh for running additional tests)

perrotta:
please test

Configuration/Eras/python/Era_Run2_2016_HIPM_cff.py
pieterdavid: Currently the `strips_vfp30_2016` modifier (included in `Run2_2016`) is the one that changes the event content, and `tracker_apv_vfp30_2016` (included in `Run2_2016_HIPM`) enables the APV simulation in the digitizer/mixing modules, and HIP mitigation in the reconstruction.
With the eventcontent customisation only in `tracker_apv_vfp30_2016` this boolean would only be there for samples/events with the APV simulation and HIP mitigation on (so it may need more care when merging the two, if needed).
What would be the preferred solution (also for the backport to 10_6)?

slava77: I think that it should be fine to always add the new boolean to flag if the simulation is on.

SimGeneral/MixingModule/python/stripDigitizer_cfi.py
```diff
@@ -23,6 +23,7 @@
     Noise = False, # will be added in DataMixer
     PreMixingMode = True, #Special mode to save all hit strips
     FedAlgorithm = 5, # special ZS mode: accept adc>0
+    includeAPVSimulation = False  # APV simulation is off for premix stage1
```
slava77: is the comment correct? It says "premix stage1" but this is inside ```premix_stage2.toModify```
should the last line perhaps be  ```(premix_stage2 & ~tracker_apv_vfp30_2016).toModify(stripDigitizer, includeAPVSimulation = False)``` ?

Configuration/PyReleaseValidation/python/relval_steps.py
```diff
@@ -1602,7 +1602,7 @@ def lhegensim2018ml(fragment,howMuch):
 
 
 steps['DIGIUP15']=merge([step2Upg2015Defaults])
-steps['DIGIUP15APVSimu']=merge([{'--customise': 'SimTracker/SiStripDigitizer/customizeSiStripAPVDynamicGainSimulation.activateSiStripAPVDynamicGain'},step2Upg2015Defaults])
+steps['DIGIUP15APVSimu']=merge([{'--era': 'Run2_2016_HIPM'},step2Upg2015Defaults])
```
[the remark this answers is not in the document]
mmusich: As far as I can see, there exists no 2015 MC relval in the matrix still in use.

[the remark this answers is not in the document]
mmusich: And yes, for how confusing it can be 2015 and 15 are used for 2015. Up to PdmV to fix it.

SimGeneral/MixingModule/python/stripDigitizer_cfi.py
mmusich: It should not read stage1 here. It's for the `MixingModule` instance used in the stage2 for signal digitization. See comment above from @makortel

Configuration/PyReleaseValidation/python/relval_steps.py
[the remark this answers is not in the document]
chayanit: @slava77 , @mmusich , thanks for bringing this up. The convention to use 2015 or 15 for 2016 has been there before my time. My understanding is we have used 2015/15 for both 2015 and 2016 relvals to indicate that these two years have the same detector geometry and conditions.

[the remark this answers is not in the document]
slava77: OK, this is not ideal and can become confusing if someone actually decides to make something more precise for 2015.
 L1 stage2 was not yet complete, and the PPS detectors were not present in 2015. This is probably close enough to 2016 to keep the workflows as is.

SimGeneral/MixingModule/python/stripDigitizer_cfi.py
slava77: > It should not read stage1 here. It's for the `MixingModule` instance used in the stage2 for signal digitization. See comment above from @makortel

just to be sure (because in the reco meeting earlier today it appeared that this PR is ready to go as is), 
do I understand correctly that here the comment is wrong, but the value ```includeAPVSimulation = False``` for all kinds of premix_stage2 is correct.
Please clarify.

mmusich: Hi @slava77, yes that's correct. The only case in which `includeAPVSimulation = True` is needed is for the `PremixingModule` configuration which does the signal + PU overlay. See this comment: https://github.com/cms-sw/cmssw/pull/28355#discussion_r344835469. I updated the description in the latest commit.

perrotta:
please test

slava77:
+1

for https://github.com/cms-sw/cmssw/pull/28355	5413741 
- jenkins tests pass and comparisons with the baseline show no differences, except for the expected addition of a bool simAPVsaturation:SimulatedAPVDynamicGain product in the 2016 workflow outputs

srimanob:
@tocheng @christopheralanwest @tlampen @chayanit @civanch Could you please consider on this PR? Thanks very much.

christopheralanwest:
+1

chayanit:
+1

fabiocos:
+operations

the update of Eras look coherent with the purpose of the PR and the discussion during the review

fabiocos:
@civanch @makortel so what is the conclusion from the simulation point of view? Do you have further comments/questions? Or can this PR be signed?

Configuration/PyReleaseValidation/python/relval_steps.py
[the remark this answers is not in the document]
srimanob: Do I understand from PR correctly that this is to define the new DIGI (signal+pu) step only, but no new relvals workflow pick it up?

Configuration/PyReleaseValidation/python/relval_steps.py
```diff
@@ -1741,9 +1741,9 @@ def lhegensim2018ml(fragment,howMuch):
                                                  digiPremixUp2018Defaults25ns])
 
 steps['DIGIPRMXUP15_PU25']=merge([digiPremixUp2015Defaults25ns])
-steps['DIGIPRMXUP15APVSimu_PU25']=merge([{'--customise': 'SimTracker/SiStripDigitizer/customizeSiStripAPVDynamicGainSimulation.activateSiStripAPVDynamicGain'},digiPremixUp2015Defaults25ns])
+steps['DIGIPRMXUP15APVSimu_PU25']=merge([{'--era': 'Run2_2016_HIPM'},digiPremixUp2015Defaults25ns])
```
srimanob: Same question here on new relvals workflow.

fabiocos:
@srimanob there is a list of Relval workflows available where the fragments are used as far as I can see:

```
13:47 cmsdev25 526> find Configuration/PyReleaseValidation/python/ -name "*.py" -exec grep -n -H APV {} \; | grep workflows
Configuration/PyReleaseValidation/python/relval_premix.py:23:workflows[250202.2]=['',['TTbar_13','DIGIPRMXUP15APVSimu_PU25','RECOPRMXUP15_PU25','HARVESTUP15_PU25']]
Configuration/PyReleaseValidation/python/relval_premix.py:24:workflows[250202.3]=['',['TTbar_13','PREMIXUP15_PU25','DIGIPRMXLOCALUP15APVSimu_PU25','RECOPRMXUP15_PU25','HARVESTUP15_PU25']]
Configuration/PyReleaseValidation/python/relval_premix.py:25:workflows[250202.4]=['',['TTbar_13','DIGIPRMXUP15APVSimu_PU25','RECOPRMXUP15_HIPM_PU25','HARVESTUP15_PU25']]
Configuration/PyReleaseValidation/python/relval_premix.py:26:workflows[250202.5]=['',['TTbar_13','PREMIXUP15_PU25','DIGIPRMXLOCALUP15APVSimu_PU25','RECOPRMXUP15_HIPM_PU25','HARVESTUP15_PU25']]
Configuration/PyReleaseValidation/python/relval_standard.py:580:workflows[1325.9] = ['', ['TTbar_13','DIGIUP15APVSimu','RECOUP15','HARVESTUP15','ALCATTUP15']]
Configuration/PyReleaseValidation/python/relval_standard.py:581:workflows[1325.91] = ['', ['TTbar_13','DIGIUP15APVSimu','RECOUP15_HIPM','HARVESTUP15','ALCATTUP15']]
Configuration/PyReleaseValidation/python/relval_pileup.py:37:workflows[25202.1]=['',['TTbar_13','DIGIUP15APVSimu_PU25','RECOUP15_PU25','HARVESTUP15_PU25']]
Configuration/PyReleaseValidation/python/relval_pileup.py:38:workflows[25202.2]=['',['TTbar_13','DIGIUP15APVSimu_PU25','RECOUP15_PU25_HIPM','HARVESTUP15_PU25']]
```

Configuration/PyReleaseValidation/python/relval_steps.py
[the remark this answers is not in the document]
mmusich: No, there are new relval workflows that excercise both the new DIGI and the RECO tracking mitigation too, those were defined in  https://github.com/cms-sw/cmssw/pull/28082. Here I just remove the customization function for the DIGI step in favour of more coincise era.

Configuration/PyReleaseValidation/python/relval_steps.py
mmusich: @srimanob same comment as above: No, there are new relval workflows that excercise both the new DIGI and the RECO tracking mitigation too, those were defined in #28082. Here I just remove the customization function for the DIGI step in favour of more coincise era.

makortel:
@fabiocos 
> so what is the conclusion from the simulation point of view? Do you have further comments/questions? Or can this PR be signed?

Visually the premixing treatment looks correct.

civanch:
+1

fabiocos:
@kpedro88 your signature is required just by PyRelval, which looks ok to me, pleas e have a look at your earliest convenience

kpedro88:
+upgrade

fabiocos:
+1

fabiocos:
merge

Dr15Jones:
This appears to have broken many of the IB RelVals.

fabiocos:
@Dr15Jones it has, the issue is partly addressed by #28454 but requires more input by @mmusich @makortel for a full solution

santocch:
+1

PR #28359, opened 2019-11-07
(DAQ) Revert to untracked parameters and new configuration

smorovic:
#### PR description:
Reverting to untracked parameters in EvFDaqDirector as discussed in #28151
Including also these changes:
- new parameter is added to control if fileBrokeHost parameter can be changed within the process by looking at hltd configuration (which is current behavior).
- default value is removed from fileBrokerHost, so it must always be present for the configuration to run successfully.
- other parameters in the service have also been regrouped, and one was renamed.

#### PR validation:

Tested in DAQ "VM" test cluster with several relevant combinations of parameters. RunBUFU.sh in EventFilter/Utilities/test also runs fine.

### DIFF
--- a/EventFilter/Utilities/interface/EvFDaqDirector.h
+++ b/EventFilter/Utilities/interface/EvFDaqDirector.h
@@ -189,20 +189,22 @@ namespace evf {
 
     std::string base_dir_;
     std::string bu_base_dir_;
-    bool directorBu_;
     unsigned int run_;
     bool useFileBroker_;
+    bool fileBrokerHostFromCfg_;
     std::string fileBrokerHost_;
     std::string fileBrokerPort_;
     bool fileBrokerKeepAlive_;
     bool fileBrokerUseLocalLock_;
-    unsigned int startFromLS_ = 1;
+    unsigned int fuLockPollInterval_;
     bool outputAdler32Recheck_;
     bool requireTSPSet_;
     std::string selectedTransferMode_;
-    std::string hltSourceDirectory_;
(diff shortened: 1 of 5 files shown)


Martin-Grunewald:
@smorovic 
On this issue:
```
default value is removed from fileBrokerHost, so it must always be present for the configuration to run successfully.
```
please do NOT do this: parameters without a default will not be written to the fillDescriptions generated cfi file (as the code does not know which value to put).
In turn, ConfDB parsing the cfi file will never know about this parameter.
So you will never be able to set it to any value in ConfDB!
Please revert and give a value, even is it is a value of "InValid" or so, and you check the string being "InValid" and issue an error message then!

smorovic:
Well, maybe "addUntracked" (or even "add") function without default should then be removed, unless there is use case I don't see.

It has now been changed to InValid as default.

Martin-Grunewald:
Thanks.
People not using ConfDB GUI are able to use the feature, but ConfDB uses cfi files to discover the parameters, so for that use case all parameters must appear in the cfi file.

smorovic:
Yes, that makes sense if they want description (and/or at least something in the generated cfi file), but no defaults.
Cheers,
Srecko

perrotta:
please test

slava77:
+1

for https://github.com/cms-sw/cmssw/pull/28359	1749535 
- reco code is not affected

fabiocos:
@mommsen comments?

mommsen:
+1

fabiocos:
+1

PR #28368, opened 2019-11-08
Hcal: A fix for DQM TDC inconsistency for 2017 workflows

lwang046:
#### PR description:

A fix for DQM TDC inconsistency of non-reproducibility of LETDCTime for 2017 workflows, caused by threshold settings in https://github.com/cms-sw/cmssw/pull/28330.

### DIFF
--- a/SimCalorimetry/HcalSimAlgos/src/HcalElectronicsSim.cc
+++ b/SimCalorimetry/HcalSimAlgos/src/HcalElectronicsSim.cc
@@ -145,8 +145,10 @@ void HcalElectronicsSim::analogToDigital(
   analogToDigitalImpl(engine, lf, result, preMixFactor, preMixBits);
   if (!PreMixDigis) {
     const HcalSimParameters& pars = static_cast<const HcalSimParameters&>(theParameterMap->simParameters(lf.id()));
-    HcalTDC theTDC((pars.threshold_currentTDC()));
-    theTDC.timing(lf, result);
+    if (pars.threshold_currentTDC() > 0) {
+      HcalTDC theTDC((pars.threshold_currentTDC()));
+      theTDC.timing(lf, result);
+    }
   }
 }
 
--- a/SimCalorimetry/HcalSimAlgos/src/HcalElectronicsSim.cc
+++ b/SimCalorimetry/HcalSimAlgos/src/HcalElectronicsSim.cc
@@ -145,7 +145,7 @@ void HcalElectronicsSim::analogToDigital(
   analogToDigitalImpl(engine, lf, result, preMixFactor, preMixBits);
   if (!PreMixDigis) {
     const HcalSimParameters& pars = static_cast<const HcalSimParameters&>(theParameterMap->simParameters(lf.id()));
-    if (pars.threshold_currentTDC() > 0) {
+    if (pars.threshold_currentTDC() > 0.) {
       HcalTDC theTDC((pars.threshold_currentTDC()));
       theTDC.timing(lf, result);
     }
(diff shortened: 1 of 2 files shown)


abdoulline:
@perrotta 
Andrea, I suppose this suggested workaround would help to avoid non-reproducibility of HCAL DQM TDC plots in 2017 MC workflows...

perrotta:
please test

abdoulline:
Well, one can see that only 2017 MC workflows are affected (as intended) and that for these workflows only DigiTask TDC plots changed (as intended),  having returned back to "all-rubbish" content (as intended, as it comes back to Mixing content, where TDC bits are used ), once we've switched off TDC simulation for 2017 (HEP17-only QIE11 module in HE)...

abdoulline:
@perrotta 
May I ask you to include this PR in some other PRs (subsequent) testing to see whether it cures Hcal DigiTask DQM issue in question?

civanch:
@lwang046 , @abdoulline , I understand this change as "negative threshold is not any more allowed".  Is it true for 2017 or in general?

abdoulline:
> @lwang046 , @abdoulline , I understand this change as "negative threshold is not any more allowed". Is it true for 2017 or in general?

In the original PR there were no negative thresholds. 
Now they (negaive th.) are added by default and the code checks it and invokes TDC simulation _solely_ in case of positive threshold(s). The latter are set for HBHE Phase1 (QIE11), for HE since 2018, for HB since Run3 (>=2019).  So, for 2017, where HCAL DQM has an issue with simulated (non-reproducible plots) TDC, we deliberately fall back to the situation of a ~month ago (before TDC simulation has been implemented).

civanch:
+1

fabiocos:
@abdoulline concerning https://github.com/cms-sw/cmssw/pull/28368#issuecomment-552310121 as soon as there is a clean baseline for comparison I will test my #28373 with this PR, to see whether the fix is working

perrotta:
> @perrotta
> May I ask you to include this PR in some other PRs (subsequent) testing to see whether it cures Hcal DigiTask DQM issue in question?

@abdoulline 

I have compared the DQM results of wf 10024, which shows those differences in the most recent IBs, using as the baseline the CMSSW_11_0_X_2019-11-10-2300 IB with this PR (for the record, I used for the comparison the results with such a baseline plus #28370). I can confirm you that no differences are observed between the baseline and the additional PR plus the baseline.

As such, one can say that the TDC time irreproducibility is removed by this PR.

On the other hand, if I understand correctly what you are doing here, you are just removing the TDC simulation from the 2017 workflows: isn't it something like hiding the dust under the carpet? If the TDC simulation in 2017 workflows gives non-reproducible results, once you switch off that simulation you obviously don't hit the problem any more, but still the TDC simulation, if activated, would provide the same non-reproducible results: as such I'd say that you didn't fix the problem, just hide it.

But maybe I am missing something.

abdoulline:
@perrotta 

(1) we have all-good reproducible TCD simulation results for full Phase1 HCAL subdetectors (>= 2018HE and >=2019 HB, where TDC actually exists).

(2) HCAL data-specialized DQM (DQM/HcalTasks: DigiTask, where the issue popped up) in its current form is not supposed to deal even with 2018, as I've previously mentioned. And there might be some randomness issue in it specifically for previous years (as no back-compatibility assumed) as @DryRun has acknowledged in a private communication. 

(3) Specifically for  MC RelVals we're using another package DQMOffline/Hcal, (fully back-compatible, era-customized) which we'll update to add TDC plots (in the near future) for Run3.
 
(4) I must admit this PR is a workaround indeed, not a full-fledged fix, but for TDC simulation we don't care much about 2017 MC (and even less so - how it looks in _actual_ DQM/HcalTasks) but solely about >= Run3...

fabiocos:
+1

abdoulline:
@perrotta 
<...>
On the other hand, if I understand correctly what you are doing here, you are just removing the TDC simulation from the 2017 workflows: isn't it something like hiding the dust under the carpet? 
<...>
You were right, Andrea. It did strike back again https://github.com/cms-sw/cmssw/issues/29076
and only now it's fully tracked down (to HcalTDC simulation issue) and fixed...

PR #28370, opened 2019-11-09
Keeping sumPt from MET Significance calculation

danbarto:
#### PR description:

The sum of the unclustered energy is one of the inputs to MET significance, but is currently not stored in miniAOD, making recomputation of the significance and the MET covariance matrix difficult. With this PR, the calculated sum of the pTs of the unclustered particles is stored, and can be accessed with metSumPt(), similar to metSignificance().
This makes a later calculation, e.g. in nanoAOD as proposed in [this PR](https://github.com/cms-nanoAOD/cmssw/pull/443), redundant.

#### PR validation:

All tests have been run, and were passed.

### DIFF
--- a/DataFormats/PatCandidates/src/classes_def_objects.xml
+++ b/DataFormats/PatCandidates/src/classes_def_objects.xml
@@ -272,9 +272,10 @@
   <ioread sourceClass = "pat::Jet" version="[1-]" targetClass="pat::Jet" source="" target="daughtersTemp_">
     <![CDATA[daughtersTemp_.reset();]]>
   </ioread>
-  <class name="pat::MET"  ClassVersion="15">
-    <version ClassVersion="15" checksum="428901429"/>
+  <class name="pat::MET"  ClassVersion="16">
+    <version ClassVersion="16" checksum="2931227204"/>
     <field name="corMap_" transient="true"/>
+   <version ClassVersion="15" checksum="428901429"/>
    <version ClassVersion="14" checksum="1795935545"/>
    <version ClassVersion="13" checksum="2368359386"/>
    <version ClassVersion="12" checksum="1474251442"/>
--- a/DataFormats/PatCandidates/src/classes_def_objects.xml
+++ b/DataFormats/PatCandidates/src/classes_def_objects.xml
@@ -272,10 +272,9 @@
   <ioread sourceClass = "pat::Jet" version="[1-]" targetClass="pat::Jet" source="" target="daughtersTemp_">
     <![CDATA[daughtersTemp_.reset();]]>
   </ioread>
-  <class name="pat::MET"  ClassVersion="17">
-    <version ClassVersion="17" checksum="2416242778"/>
+  <class name="pat::MET"  ClassVersion="16">
+    <version ClassVersion="16" checksum="2416242778"/>
     <field name="corMap_" transient="true"/>
-   <version ClassVersion="16" checksum="2931227204"/>
    <version ClassVersion="15" checksum="428901429"/>
    <version ClassVersion="14" checksum="1795935545"/>
    <version ClassVersion="13" checksum="2368359386"/>
--- a/DataFormats/PatCandidates/src/classes_def_objects.xml
+++ b/DataFormats/PatCandidates/src/classes_def_objects.xml
@@ -272,9 +272,10 @@
   <ioread sourceClass = "pat::Jet" version="[1-]" targetClass="pat::Jet" source="" target="daughtersTemp_">
     <![CDATA[daughtersTemp_.reset();]]>
   </ioread>
-  <class name="pat::MET"  ClassVersion="16">
-    <version ClassVersion="16" checksum="2931227204"/>
+  <class name="pat::MET"  ClassVersion="17">
+    <version ClassVersion="17" checksum="2416242778"/>
     <field name="corMap_" transient="true"/>
+   <version ClassVersion="16" checksum="2931227204"/>
    <version ClassVersion="15" checksum="428901429"/>
    <version ClassVersion="14" checksum="1795935545"/>
    <version ClassVersion="13" checksum="2368359386"/>
--- a/RecoMET/METProducers/src/PFMETProducer.cc
+++ b/RecoMET/METProducers/src/PFMETProducer.cc
@@ -94,7 +94,7 @@ namespace cms {
     event.getByToken(rhoToken_, rho);
 
     //Compute the covariance matrix and fill it
-    double sumPt;
+    double sumPt = 0;
     reco::METCovMatrix cov = metSigAlgo_->getCovariance(
         *inputJets, leptons, candInput, *rho, resPtObj, resPhiObj, resSFObj, event.isRealData(), sumPt);
 
--- a/RecoMET/METProducers/src/PFMETProducer.cc
+++ b/RecoMET/METProducers/src/PFMETProducer.cc
@@ -94,8 +94,9 @@ namespace cms {
     event.getByToken(rhoToken_, rho);
 
     //Compute the covariance matrix and fill it
+    double sumPt;
     reco::METCovMatrix cov = metSigAlgo_->getCovariance(
-        *inputJets, leptons, candInput, *rho, resPtObj, resPhiObj, resSFObj, event.isRealData());
+        *inputJets, leptons, candInput, *rho, resPtObj, resPhiObj, resSFObj, event.isRealData(), sumPt);
 
     return cov;
   }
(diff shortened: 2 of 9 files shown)


slava77:
@cmsbuild please test

RecoMET/METProducers/src/PFMETProducer.cc
```diff
@@ -94,8 +94,9 @@ namespace cms {
     event.getByToken(rhoToken_, rho);
 
     //Compute the covariance matrix and fill it
+    double sumPt;
```
ahinzmann: Doesn't this need initialization = 0?

danbarto: You're right, thanks for the catch! It's initialized with 0 in the other producers, but somehow missing there. I'll fix it.

perrotta:
please test

perrotta:
Adding the JetMET POG conveners @ahinzmann @lathomas , since a specific MET reco contact isn't appointed yet

ahinzmann:
Hi,
what is exactly the difference between met->sumEt and met->sumPt with this PR? (this is something non-experts will wonder).
If find the definition of sumEt here:
https://github.com/cms-sw/cmssw/blob/2ba5d421e10379d81760a899532b2c991b89c82c/RecoMET/METAlgorithms/src/METAlgo.cc#L44
So one difference is certainly that Et instead of Pt of particles is used (makes a difference for charged particles that are assigned a pi+- or lepton mass).
But is there any other difference in the computation of sumPt for signficance w.r.t. the general sumEt definition?
(depending on how significant the difference in definition is one may think of a more detailed name than sumPt)

danbarto:
sumEt is the scalar sum of the Et of all PF candidates above the globalThreshold (which is 0 if I'm not wrong), while for sumPt overlap with leptons and jets is removed.

ahinzmann:
so sumPt is rather sumPtUnclusered? Or is unclustered defined yet in another way?

danbarto:
Yes, it's the measure of the unclustered energy in the MET significance. I agree that sumPtUnclustered is probably a less ambiguous name.

slava77:
> Yes, it's the measure of the unclustered energy in the MET significance. I agree that sumPtUnclustered is probably a less ambiguous name.

https://github.com/cms-sw/cmssw/blob/81826f1ed7ccffddbd3d642838c057e31d309987/RecoMET/METAlgorithms/src/METSignificance.cc#L154-L157
this looks like more than unclustered

danbarto:
Yes, jets with pt below the threshold (15 GeV by default) are also added to sumPt, and their associated PF candidates are removed to avoid double counting.

ahinzmann:
So jets with pt<15 are treated as unclustered energy, so the name sumPtUnclustered could make sense.

ahinzmann:
So my suggestion would be to rename to sumPtUnclustered.

danbarto:
Fine with me, if there's no objection I'll rename it and update the PR.

ahinzmann:
Just to confirm. Also @lathomas supports sumPtUnclustered.

perrotta:
please test

DataFormats/PatCandidates/src/classes_def_objects.xml
```diff
@@ -272,9 +272,11 @@
   <ioread sourceClass = "pat::Jet" version="[1-]" targetClass="pat::Jet" source="" target="daughtersTemp_">
     <![CDATA[daughtersTemp_.reset();]]>
   </ioread>
-  <class name="pat::MET"  ClassVersion="15">
-    <version ClassVersion="15" checksum="428901429"/>
+  <class name="pat::MET"  ClassVersion="17">
+    <version ClassVersion="17" checksum="2416242778"/>
```
perrotta: Please use consecutive numbers for ClassVersions, here "16".

DataFormats/PatCandidates/src/classes_def_objects.xml
```diff
@@ -272,9 +272,11 @@
...
     <field name="corMap_" transient="true"/>
+   <version ClassVersion="16" checksum="2931227204"/>
```
perrotta: Intermediate versions can be neglected

DataFormats/PatCandidates/src/classes_def_objects.xml
danbarto: done

perrotta:
please test

perrotta:
+1
- "sumPtUnclustered" added to the MET DataFormat, according to the PR description and following review
- Jenkins tests pass and show no other differences than this additional double added to the dataformat

santocch:
+1

peruzzim:
A backport to 102X and 106X has to be discussed if you want this to be used in nanoAOD (and this implementation currently breaks the dataformat). Please follow up

danbarto:
It would be desirable to have sumPtUnclustered in nanoAOD v7, so a backport will be needed.

fabiocos:
@danbarto please provide it, so it can be merged once this feature is fully validated

fabiocos:
+1

danbarto:
@slava77 @fabiocos Do you have a suggestion/example on how to best keep sumPtUnclustered for later access in nanoAOD, without changing the data format (for the backport)? Thanks a lot for your help!

slava77:
> @slava77 @fabiocos Do you have a suggestion/example on how to best keep sumPtUnclustered for later access in nanoAOD, without changing the data format (for the backport)? Thanks a lot for your help!

Addition of a new variable should be OK for a backport. The algorithm to fill and read from it will likely need to be enabled with an era modifier so that this reading is consistent. For now in 10_6_X you can use ```run2_miniAOD_devel``` to do that.

peruzzim:
@slava77 I think that @danbarto was referring here to how to calculate it for using it in nanoAOD production without breaking MiniAOD consistency in ongoing campaigns, given that now the code is integrated with the producer that runs also during MiniAOD production.

slava77:
> @slava77 I think that @danbarto was referring here to how to calculate it for using it in nanoAOD production without breaking MiniAOD consistency in ongoing campaigns, given that now the code is integrated with the producer that runs also during MiniAOD production.

shouldn't this be just to recompute the pat MET by rerunning the corresponding producer in the nanoAOD workflow?
The more convenient way would be to make a producer which can read the available pat::MET and fill just this variable via e.g. ModifiedObjectProducer infrastructure.

peruzzim:
yes but backporting this will result in touching the MET dataformat...
sorry, maybe I'm not seeing the easy way out here... should one just backport the code and put the result somewhere transient?

slava77:
> yes but backporting this will result in touching the MET dataformat...

as I mentioned above, adding a new variable should be OK as is. Consistent reading and filling of it in the production workflows will need use of appropriate modifiers.

peruzzim:
This was not backported to 10_2_X, and is needed for NanoAODv7.
@danbarto would you have the chance to do it promptly?

#29180

danbarto:
> This was not backported to 10_2_X, and is needed for NanoAODv7.
> @danbarto would you have the chance to do it promptly?
> 
> #29180

I should have time later today. Just to be sure I understand correctly, we need some era modifier to turn on/off filling of sumPtUnclustered for the backport?

peruzzim:
@slava77 could you please provide some guidance on how to backport this to 10_2_X, technically? I'm also not sure. Thank you

peruzzim:
After discussion with @silviodonato, we would propose to do the backport (to both 10_2_X and 10_6_X) as follows:

- change the MET dataformat as in the master branch
- add a configuration variable to `PATMETProducer`, let's call it `storeSumPtUnclustered`, that switches the filling of this variable by calling or not the `setMETSumPtUnclustered` function, and default it to `False`
- `run2_miniAOD_devel` modifier to set this variable `True`, for any possible future usage in re-MiniAOD campaigns
- also set it to `True` in the NanoAOD workflow (I can take care of this in the NanoAOD PR)

Would you agree?

slava77:
> Would you agree?

please see https://github.com/cms-sw/cmssw/pull/28370#issuecomment-558644947
I think that a new variable can just be added without extra modifiers

peruzzim:
ok, in this case we can just do a trivial backport without worrying of modifiers. @danbarto can you please prepare it?

slava77:
> ok, in this case we can just do a trivial backport without worrying of modifiers. @danbarto can you please prepare it?

don't we also need a backport for 10_6_X?

peruzzim:
yes indeed, not only for re-mini but also for backporting the nanoaodv7 configuration there (I'm working on that)

danbarto:
Should I prepare the backport to 10_6_X?

ahinzmann:
> Should I prepare the backport to 10_6_X?

Yes, please!

PR #28371, opened 2019-11-09
Use isolated particle level photons

AndrewLevin:
#### PR description:

In order to measure isolated photon cross sections, we need to have isolated particle level photons.

#### PR validation:

I have run over some recent relval samples and checked that the number of isolated particles level photon is reasonable.

### DIFF
--- a/GeneratorInterface/RivetInterface/interface/RivetAnalysis.h
+++ b/GeneratorInterface/RivetInterface/interface/RivetAnalysis.h
@@ -39,7 +39,7 @@ namespace Rivet {
     double _jetConeSize, _jetMinPt, _jetMaxEta;
     double _fatJetConeSize, _fatJetMinPt, _fatJetMaxEta;
 
-    double _phoMinPt, _phoMaxEta ,_phoIsoConeSize, _phoMaxRelIso;
+    double _phoMinPt, _phoMaxEta, _phoIsoConeSize, _phoMaxRelIso;
 
     Particles _leptons, _photons, _neutrinos;
     Jets _jets, _fatjets;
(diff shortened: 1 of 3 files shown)


qliphy:
adding  @intrepid42 @xjanssen

qliphy:
please test

AndrewLevin:
I just updated the relevant NanoAOD config file. Those relval tests should be successful now.

qliphy:
please test

intrepid42:
Hi, looks fine to me, thank you for implementing! I verified that all the prompt photons are still stored in the nanoAod "GenPart" collection if some analysis needs them.

My only worry would be that some analysis might be using the current particleLevel:photons and unknowingly switches from prompt to isolated photons.
But we could send a mail to the Rivet HN to announce this change.

qliphy:
+1

peruzzim:
+xpog

I think at the moment we're not saving particleLevel:photons in NanoAOD, so it will not affect anything in the NanoAOD content. In this sense, the configuration change is purely technical.
Can you confirm this is totally unrelated to HTXSRivetProducer, and will not affect its results?

For the more general item of changing the default configuration of the producer, it's up to GEN to follow up - but I agree it should be clearly announced to physics groups.

fabiocos:
+1

@AndrewLevin please consider the question https://github.com/cms-sw/cmssw/pull/28371#issuecomment-552358244

fabiocos:
merge

AndrewLevin:
> +1
> 
> @AndrewLevin please consider the question [#28371 (comment)](https://github.com/cms-sw/cmssw/pull/28371#issuecomment-552358244)

I am pretty sure this is totally independent of the HTXSRivetProducer.

santocch:
+1

PR #28375, opened 2019-11-11
DQM: Configuration alias for QualityTester

schneiml:
#### PR description:

This PR is extracted from #28316, the first in a series to eventually replace #28316.

It replaces the direct invocations of `cms.EDAnalyzer("QualityTester", ...)`  with calls to an alias, to allow changing the module type of `QualityTester` to `EDProducer` in the future.

This PR touches lots of packages, but does not include *any* functional changes. Most of the changes are machine-generated.

#### PR validation:

No functional changes, alias seems to work.

PR #28247, opened 2019-10-23
DQM: Remove more DQMStore features

schneiml:
#### PR description:
This PR removes more rarely used functionality from the DQM infrastructure, namely
- `scaleElements`: not sure what this was ever used for, online maybe?
- `softReset`: this feature is rather useful, but the few usages don't really justify the complexity. Used mainly in ECAL DQM.
- `setAccumulate` flag: basically unused, not sure what it was supposed to signal.
- `remove*`: Used in a bunch of places mostly for non-standard workflows. Remove is very hard to support safely (even today, it might randomly invalidate pointers that other modules are holding), so we have to drop it. The most common pattern is *"check if ME exists, if yes remove, then book"*: this probably was not required in most cases, but it might be required to re-bin plots. We might change semantics of `book*` to *replace* an existing ME if *merge* is not possible, without invalidating the exisiting MEs (only replace the `TH1`). [update]
- `QStatisticalTests`: completely unused. [update]

#### PR validation:
Requires feedback from subsystem users, mainly ECAL (@tanmaymudholkar ?) regarding `softReset`. I suspect this PR will change behavior for ECAL online DQM.

The other user of `softReset` is `HLXMonitor`, which as far as I understand is history. Not sure if anybody feels responsible.

`scaleElements` was called in the `MEtoEDM` workflow, so it might affect AlCa/PCL. I suspect it does not (I can't see any configuration for it, and I don't see what it would be good for), but it might be worth another check.

Remove was used in a lot of places, I suspect in most cases this will only make additional MEs show up in the output. However, there might be cases where re-booking now behaves incorrectly. I typically removed the entire `cleanup` functionality incl. config options, unless that seemed to complicated, then I left a comment in the now-useless function stubs.

PR #28379, opened 2019-11-12
DQM: Restructure QTests (v2)

schneiml:
#### PR description:

This is the second, and main, PR replacing #28316. [It is based on #28375 (functionally required for this to work, separated to reduce the amount of changes to review), and  #28247 (massive conflicts between the two, and  #28247 will probably go in first).] Update: Rebased with both PRs merged to master.

This PR covers the reorganization of the DQM QualityTest functionality in CMSSW:
- `QualityTester` becomes a plugin in `DQMServices/Components`, handling all QTest things.
- The XML parsing is now done with `boost::property_tree` rather than `xerces`. Much less code, however more strict (some of the QTest configuration XML used in production was malformed!) and probably worse error reporting.
- The QTest implementation remain in `DQMServices/Core` `QTest.[h|cc]`. However, the *only* place that uses them now is `QualityTester`.
- `QReport`s (results) remain available via the `MonitorElement`, but don't contain pointers to `QCriterion`s (the actual tests) any more.
- `QualityTester` is now a `DQMEDHarvester`, rather than a legacy module.
- The ordering `QualityTester` vs. other harvesters is complicated. Some of the old behaviour was accidental and may be impossible to reproduce, since nothing is guaranteed by the EDM framework with respect to ordering in `endJob` (see #28354). In practice, some of SiStrip Certification output is broken now and there is no way to fix it, but it could also break spontaneously in production and there would be no way to fix it.

A few changes hitchhike this PR due to the history in #28316 and other requirements:
- The `DQMEDHarvester` base class is updated to the `new DQMStore` version, since some features (`dqmEndRun`, proper dependencies) are needed here.
- `DQMGenericClient` is moved and restructured, to prepare for the removal of the `ClientConfig` package.
- Some more random changes remove dependencies on `ClientConfig`. However, the package remains present and the Tracker modules relying on `DQMParserBase` are unaffected.

#### PR validation:
Some validation happend in #28316 and the PR test comparison should be clean. This needs re-checking now that the other PRs are part of the baseline.


schneiml:
please test

it looks like this compiles and does not blow up immediately in a local test, so the rebase might actually have succeeded. 

The last few commits include a change suggested in #28354, that ideally does not change anything but makes things more clean. Let's see how it goes.

schneiml:
please test

with the last changes, the module order in `SiStripMonitorClient` looks roughly correct. The results are now different, but look still wrong. Re-running tests to see if there are any useful hints in the comparison.

schneiml:
Reviewing the changes, there are additional MEs due to #28247, and
- a few changes in 20034.0 and similar (incl. 136.7611). Here, the baseline shows `<phiEfficiency_staMuon.EfficiencyInRange>qr=st:90:0:UNKNOWN_ALGORITHM:NO_MESSAGE_ASSIGNED` for some tests, even though the config looks normal and ok. The new implementation handles it correctly. I blame this on bugs in the old implementation.
- The Qtest `dttf_01_tracks_occupancy_test_summary.L1TOccupancyTest` is missing in 1000 and similar workflows. I *think* this is a QTest that applies to a summary plot in harvesting by mistake.
- The Pixel Phase0 `reportSummary` in 25202.0 changes. Not sure why this only affects this workflow.
- In 8.0 and similar, changes to `SiStrip/MechanicalView/sToNReportMap` and report summary/certification contents. This may be due to a bug affecting `hasError()` and similar calls, locals tests in progress.

schneiml:
please test

The last change should fix the differences in 8.0 and similar.

schneiml:
That last fix indeed resolves the issues with `SiStripMonitorClient` in 8.0 and similar, and also the Pixel Phase0 difference in 25202. I'd consider things done at this point, assuming the other differences (`dttf_01_tracks_occupancy_test_summary.L1TOccupancyTest`, `UNKNOWN_ALGORITHM` and new QTests in Tracking are bugs in the QTest setup and the old implementation, respectively.

The remaining differences in the comparison are due to  #28247 and we need to re-test (and probably rebase) after it is merged.

schneiml:
+1

Conditioned on that the two underlying PRs get merged first, #28247 and  #28375. This should also reduce the amount of signatures required for this one a bit.

schneiml:
please test

My local compilation has not finished yet, but the rebase was clean, so no reason to assume it won't work.

schneiml:
Mmmh... the bin-to-bin comparison is broken due to the "only copy stuff with x bits  to $PATH" change. Great. Waiting for #28494.

schneiml:
+1

There are some residual changes, but I have no reason to assume they are any worse than what was seen and explained before.

Martin-Grunewald:
+1

fabiocos:
please test 

@schneiml the bin-to-bin comparison should now be fixed, so we can repeat the test

fabiocos:
@qliphy @agrohsje @intrepid42 could you please have a look at the generator part? It involves a minimal change in Rivet interface

fabiocos:
@rekovic the change in L1T area are minimal, please check, I will move forward

fabiocos:
@schneiml I see that this PR adds an additional dependency on boost, anyway property_tree is already used in several areas of DQM. Anyway the xml parsing so far is used only in a test for the luminosity code

schneiml:
@fabiocos yes indeed, I use the `property_tree` XML parsing instead of Xerces (via `ClientConfig`, to be deleted in #28380 ). It seems clearly superior to me, and as I understood anything in boost can be used without much trouble.

fabiocos:
+1

fabiocos:
merge

schneiml:
Thanks everybody for the help with this big project!

Where there any suspicious problems in the IB tests over the weekend? Given there are actual behavior changes in this PR, I'd not be too surprised...

fabiocos:
@schneiml the only anomalies noticed in the week end are dealt with in #28581 to my knowledge

PR #28380, opened 2019-11-12
DQM: Remove ClientConfig package

schneiml:
#### PR description:

[This PR is the last one including changes from #28316. It is based on #28379, which includes the changes that make this PR possible] Edit: rebased after these PRs are merged.

This PR removes the `DQMServices/ClientConfig` package. The more important parts of its functionality (`DQMGenericClient`, `QualityTester`) where moved to `DQMServices/Components` in #28379. What remains is `DQMParserBase`, some sort of XML parsing helper based on Xerces, and a few more random files.

The few modules that where using `DQMParserBase` (`SiPixelMonitorClient`/`SiStripMonitorClient`/`SiStripCommissioningClients`) now have their parsers stubbed out. The only one of them used in the (short matrix) tests seems to be one part of `SiPixelMonitorClient`, which has been replaced with a `boost::property_tree` based parser. I also implemented a parser for `SiStripMonitorClient`, but could not actually test it yet (it may be used in online DQM, which I can test at some point). The others remain stubbed out, and while I could easily replace them, I am hesitant to do so without being able to test them at all.

#### PR validation:

None for this PR, but some in #28316. No changes in PR tests expected, but most of the affected code is not covered by the PR tests. The stubbed out implementations clearly won't work, but it is questionable if they are still used anywhere. The assertions should make sure that anything using the missing functionality crashes.


schneiml:
please test

Rebased to latest version of #28379.

schneiml:
please test

seems to pass a local test now, after fixing another Python dependence in `Validation/RPCRecHits`.

fabiocos:
@mmusich this PR removes some functionalities in the Tracker DQM, partly replacing them with ```boost``` as far as I can see, could you please comment about this proposal? The main question of @schneiml is about testing the alternative, boost-based, parsing

mmusich:
@fabiocos how urgent is this PR? I have my hands full now, this will have to be postponed to next year likely.
BTW: I see in the PR initial comment:

> The few modules that where using DQMParserBase (SiPixelMonitorClient/SiStripMonitorClient/SiStripCommissioningClients) now have their parsers stubbed out. 

This needs to be reviewed also by the Strip Online team: @rgerosa can you have a look and/or suggest who else should be made aware of these changes?

schneiml:
@fabiocos @mmusich We were/are in touch with this with Tracker DQM.

The issue here is that I can't test some of the code (incl. one of the new parsers and the stubbed out stuff), and we wondered if this is used at all, ever, today. I think the answer was "probably not" but not definitive yet. 

So, we are still waiting for the green-light from Tracker DQM with this.

Re urgency: Not that urgent, we just might want to finally get rid of this package before the next release, but nothing else depends on this PR (unlike most of my other PRs these days).

mmusich:
@schneiml I would really prefer if this gets a rather more detailed scrutiny than: 

>  I think the answer was "probably not" but not definitive yet.

especially given that 

>  nothing else depends on this PR.

In particular the packages `DQM/SiStripCommissioningClients` `DQM/SiStripCommissioningSources` are used at P5 for the commissioning of the detector and are most likely not exercised in the standard matrix.

rgerosa:
Thanks @mmusich for adding me to this thread. The SummaryPlotXmlParser is used by the "online" / commissioning SiStrip tracker workflow i.e. is part of the DQM based software that produced online conditions/calibrations for the strip-tracker.

Therefore, I confirm that the tool is used and that SummaryPlotXmlParser class needs to be adapted in order to parse the same set of XML based config files as before.

schneiml:
@mmusich I agree, also no pressure from our side. Though 11_1 does sound like a reasonable goal for me...

@rgerosa Ah, so we need to add that one. I can blindly do that (but not test it!) or leave it to you. Also keep in mind that SiStripConfigParser is implemented here but also untested.

That leaves SiPixelLayoutParser and the two writers in the gray. I think sb. mentioned that the writers are most likely unused, and the Pixel stuff... I guess if it was used for commissioning, the detector is history now, and the stuff used offline works.

rgerosa:
@schneiml thanks a lot for the reply. I can work on that but unfortunately not before the xmas break because I need to finalise other commitments. Therefore, depending on your schedule:

1) if you prefer to close this PR this week, I think it would be best if you can try to "blindly" implement it. Then, after the winter break, I can test the PR and add fixes if needed.
2) If it is not urgent, I can work on this in January.

schneiml:
@rgerosa January should be all fine.

silviodonato:
This is just a friendly reminder about this PR.

silviodonato:
This is just a friendly reminder about this PR.

silviodonato:
Weekly reminder of this PR

rgerosa:
Hi All,

sorry for the delay but I just had now time to fix the SummaryPlotXmlParser class. @schneiml in order to integrate this in your PR, I copied the two relevant files (header and implementation) in this public folder on my afs:

/afs/cern.ch/user/r/rgerosa/public/PR28380/
SummaryPlotXmlParser.h --> goes into DQM/SiStripCommissioningClients/interface/
SummaryPlotXmlParser.cc --> goes into DQM/SiStripCommissioningClients/src/

In this way you can just replace and update the PR without further overheads.

schneiml:
Thanks @rgerosa, I rebased this PR and added your changes.

schneiml:
please test

No changes expected, IIRC. The relevant code paths are not covered in the PR tests.

schneiml:
+1

assuming this is fine from Tracker side now (else, @rgerosa quickly jump in).

mmusich:
>  I also implemented a parser for SiStripMonitorClient, but could not actually test it yet (it may be used in online DQM, which I can test at some point). The others remain stubbed out, and while I could easily replace them, I am hesitant to do so without being able to test them at all.

@arossi83 @sroychow please comment if there is any remaining concern and / or if you can provide suitable test configurations.

silviodonato:
hold
waiting for feedback from @arossi83 @sroychow about the test of the parser of SiStripMonitorClient

sroychow:
@silviodonato Tracker DQM is okay with this PR to be merged. Few bugs were discovered in configuration files in DQM/SiStripMonitorClient package. We shall make a separate PR for those.

silviodonato:
unhold

silviodonato:
+1

PR #28389, opened 2019-11-13
Register consumes from the EventSetup in GsfElectronAlgo

guitargeek:
#### PR description:

As explained in https://github.com/cms-sw/cmssw/pull/28223 and in @Dr15Jones email, it's better to register the consumes from the EventSetup also for ED producers. I wanted to do this for some time in the GsfElectronAlgo anyway, because in this way I can get rid of the manual caching with the `unsigned long long cacheID` member variables, trusting the framework to handle this efficiently on its own and bringing us a good step closer to an algo with no mutable member variables. The remaining class members were also reorganized a bit to further reduce the mutable state and make it clearer what is still mutable: the HcalHelper, the RegressionHelper and the EcalClusterFunctions.

#### PR validation:

No backport intended.

### DIFF
--- a/RecoEgamma/EgammaElectronAlgos/src/GsfElectronAlgo.cc
+++ b/RecoEgamma/EgammaElectronAlgos/src/GsfElectronAlgo.cc
@@ -224,12 +224,12 @@ void GsfElectronAlgo::calculateShowerShape(const reco::SuperClusterRef& theClus,
   std::vector<int> recHitSeverityToBeExcluded;
   if (detector == EcalBarrel) {
     recHits = eventData.barrelRecHits.product();
-    recHitFlagsToBeExcluded = generalData_.recHitsCfg.recHitFlagsToBeExcludedBarrel;
-    recHitSeverityToBeExcluded = generalData_.recHitsCfg.recHitSeverityToBeExcludedBarrel;
+    recHitFlagsToBeExcluded = cfg_.recHits.recHitFlagsToBeExcludedBarrel;
+    recHitSeverityToBeExcluded = cfg_.recHits.recHitSeverityToBeExcludedBarrel;
   } else {
     recHits = eventData.endcapRecHits.product();
-    recHitFlagsToBeExcluded = generalData_.recHitsCfg.recHitFlagsToBeExcludedEndcaps;
-    recHitSeverityToBeExcluded = generalData_.recHitsCfg.recHitSeverityToBeExcludedEndcaps;
+    recHitFlagsToBeExcluded = cfg_.recHits.recHitFlagsToBeExcludedEndcaps;
+    recHitSeverityToBeExcluded = cfg_.recHits.recHitSeverityToBeExcludedEndcaps;
   }
 
   std::vector<float> covariances = ClusterTools::covariances(seedCluster, recHits, &topology, &geometry);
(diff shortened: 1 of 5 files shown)


perrotta:
please test

RecoEgamma/EgammaElectronAlgos/src/GsfElectronAlgo.cc
```diff
@@ -501,8 +540,18 @@ void GsfElectronAlgo::completeElectrons(reco::GsfElectronCollection& electrons,
...
+  auto const& trackerGeometry = eventSetup.getData(trackerGeometryToken_);
+  auto const& ecalSeveretyLevelAlgo = eventSetup.getData(ecalSeveretyLevelAlgoToken_);
+
   checkSetup(eventSetup);
-  auto eventData = beginEvent(event);
+  auto eventData = beginEvent(event, caloGeometry, ecalSeveretyLevelAlgo);
+  double magneticFieldInTesla = magneticField.inTesla(GlobalPoint(0., 0., 0.)).z();
+
+  MultiTrajectoryStateTransform mtsTransform(&trackerGeometry, &magneticField);
+  GsfConstraintAtVertex constraintAtVtx(eventSetup);
```
slava77: GsfConstraintAtVertex construction seems nontrivial; it was previously done once per IOV.
Please provide profiler report to justify that the cost change is negligible.

guitargeek: Nontrivial means here that it gets some things from the EventSetup. My assumption here was that the EventSetup is pretty smart itself, and getting something again that was already obtained in the beginning of the IOV should be basically instantaneous since it returns a cached value (that's also how I realized the manual caching in egamma code is superfluous). See here:

https://twiki.cern.ch/twiki/bin/view/CMSPublic/SWGuideHowToGetDataFromES

"Behind the scenes, the EventSetup system only runs the algorithm used to create the data in a Record the first time that data is requested after a Record has changed its IOV. All subsequent requests will get the cached value."

I can do some profiling anyway, it never hurts of course :)

slava77: non-trivial means that there is more than just getting the event setup here, at least at two-deep level I found the following (pseudo-code)
``` GsfConstraintAtVertex::GsfConstraintAtVertex () {
 gsfPropagator_ = new GsfPropagatorAdapter(AnalyticalPropagator(magField_, anyDirection)); //copies the temp prop argument
  tipExtrapolator_ = new TransverseImpactPointExtrapolator(*gsfPropagator_); //copies the prop argument
}
```
... after poking deeper now, it seems like the above reduces to some simple initializations and computations of perhaps \~100 operations (pointer assignments, a bunch of "new" for 5-10 words). So, this is probably fine.

guitargeek: Thanks a lot for writing this down, I came to the same conclusion back when I made the changes for this PR.

slava77:
+1

for https://github.com/cms-sw/cmssw/pull/28389	a8ca482 
- mostly technical/refactoring code changes are in line with the PR description and the follow up review

fabiocos:
+1

PR #28408, opened 2019-11-15
E/g HLT Pixel match Speedups: 11_0_X

Sam-Harper:
#### PR description:

This PR speeds up the E/gamma HLT pixel matching. It uses work done by Calib Fangmeier to allow the pixel matching to run on triplet cleaned doublets which at a small efficiency cost will greatly increase speed (the path needs to enable this so its of for now). 

Additionally it does some optimization where expensive variables are now computed outside a loop, it also uses parametric b field for speed where appropriate and finally adds a minimum et cut for superclusters to pixel match. 

This benefits Run2/Run3 configs but the real reason is for PhaseII 200PU where the e/gamma hlt pixel matching was taking ~30mins an event. It now takes ~2seconds and could be further optimised.

No / very small changes are expected for existing HLT workflows although the timing is reduced. 

I validated this on an electron gun sample for 2021 MC conditions (the samples used for the Run3 projections) using /dev/CMSSW_11_0_0/GRun/V10

image

Timing:
TimeReport   0.015958     0.032457     0.032457  hltEgammaElectronPixelSeeds
goes to
TimeReport   0.007012     0.014263     0.014263  hltEgammaElectronPixelSeeds

Doing a full timing study  on 2018 data I see a drop from 240ms to 236ms with the drops happening in electron paths as expected.

### DIFF
--- a/RecoEgamma/EgammaElectronAlgos/interface/TrajSeedMatcher.h
+++ b/RecoEgamma/EgammaElectronAlgos/interface/TrajSeedMatcher.h
@@ -259,6 +259,8 @@ private:
   std::string detLayerGeomLabel_;
 
   bool useRecoVertex_;
+  bool enableHitSkipping_;
+  bool requireExactMatchCount_;
   std::vector<std::unique_ptr<MatchingCuts> > matchingCuts_;
 
   //these two varibles determine how hits we require
--- a/RecoEgamma/EgammaElectronAlgos/interface/TrajSeedMatcher.h
+++ b/RecoEgamma/EgammaElectronAlgos/interface/TrajSeedMatcher.h
@@ -197,7 +197,7 @@ private:
                                       const GlobalPoint& candPos,
                                       const GlobalPoint& vprim,
                                       const float energy,
-                                      const int charge);
+                                      const TrajectoryStateOnSurface& initialTrajState);
 
   static float getZVtxFromExtrapolation(const GlobalPoint& primeVtxPos,
                                         const GlobalPoint& hitPos,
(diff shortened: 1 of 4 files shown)


perrotta:
please test

Sam-Harper:
ah I forgot to include Martin as this is really a HLT module
 @Martin-Grunewald 

At somepoint RECO will use this though, but not for 11_0_0

Martin-Grunewald:
+1

perrotta:
@Sam-Harper , just to verify that I understood it correctly:
- In single electron gun the time taken by pixel matching goes down from 32 ms to 14 ms with this PR
- In PU200 Phase2 samples the time taken by pixel matching goes down from ~1800 s to ~2 s with this PR ("and it could be further optimised")

Do you confirm that I got it correctly and that there is such an impressive speed-up with PU200?

Sam-Harper:
> Do you confirm that I got it correctly and that there is such an impressive speed-up with PU200?

Yes, phaseII e/gamma pixel matching was taking ~30mins to run per event at 200PU. Its not really an impressive speed up, we knew this module was doing things suboptimally just there were more important things to fix. As it combines pixel pairs, it scales extremely poorly with number of pixel hits.   Once it became a blocker for PhaseII, it got fixed. 

It was a combination of code optimizations and config changes for PhaseII, the Run2 stays with the old config and just gets the code optimizations. 

The 15ms->7ms was actually measured on data using a proper timing setup. 

This is something that would be nice to have in for 11_0_0 but then again it the 11_0_0 production will not use this as only a fake HLT menu will be run for the Phase II production.

perrotta:
+1
- This code is only run at HLT, by now, and not in offline reco
- From a visual inspection, its implementation corresponds to what declared in the PR description
- The validation provided shows the improvement in the cpu consumption of the pixel matcher, and no effect on the inspected HLT results
- Jenkins tests show no differences in reco outputs (of course), and at a first glance not even in the HLT outputs

fabiocos:
+1

PR #28410, opened 2019-11-15
Enforce proper produce calls for Begin and End Transitions

Dr15Jones:
When calling `produce` we now check that if non-event transition is chosen, then the `begin` and `end` part of the transition must match the type of producer making the call.

#### PR description:

Framework code compiles and all framework tests pass.


Dr15Jones:
please test

Dr15Jones:
+1

fabiocos:
+1

PR #28429, opened 2019-11-20
Phase out old pre-GED GsfElectron sequence

guitargeek:
#### PR description:

This PR suggests to remove the old gsfElectron sequence and it's producers, which are not in use anymore since the rewrite between Run I and II in the context of the GED (global event description) campaign. This would be a big step for the egamma code cleanup. Other producers would not have to support the old workflow anymore (which can result in a good reduction of the number of configuration parameters) helpers that were only used by the old producers can be removed, etc.

Still, this cleanup is not completely trivial because there is a whole relval-step (`RECODFROMRAWRECO`) which still depends on the old sequences. I am not completely familiar with these relval-steps so maybe I'm doing a mistake by proposing also the removal of this step, but it seems to me that the step is not supported anymore anyway: in the MatrixReader, it's hardcoded that this step is ignored [1,2], and the word `RECODFROMRAWRECO` appears nowhere else in CMSSW. The other remaining use of the old gsfElectronSequence was in the `cosmicElectronSequence`. However, this is not used anywhere: the cosmic reconstruction uses only the `egammarecoCosmics_woElectrons` sequence.

@Sam-Harper, @jainshilpi, @lsoffi, please let us know if you think if this is a good idea or if you  think the old GsfElectronProducer should still stick around. I think this is a very good opportunity to reduce quite a bit the complexity of the egamma reco area and hence render it easier to understand and to work with.

[1] https://github.com/cms-sw/cmssw/blob/master/Configuration/PyReleaseValidation/python/MatrixReader.py#L243
[2] https://github.com/cms-sw/cmssw/blob/master/Configuration/PyReleaseValidation/python/MatrixReader.py#L341
[3] https://github.com/cms-sw/cmssw/blob/master/Configuration/StandardSequences/python/ReconstructionCosmics_cff.py#L59

#### PR validation:

No backport intended.


perrotta:
please test

guitargeek:
That went better then expected! Unfortunately I removed the `InputTag` in 
```Python
fastSim.toModify(lowPtGsfElectrons,ctfTracksTag = cms.InputTag("generalTracksBeforeMixing"))
```
thinking it was not needed and it can take just a string. Of course that didn't work because the ctfTracksTag parameter was not explicitly used in python before. Pardon the force push, I imagine nobody started reviewing here anyhow and I preferred to not have an extra commit for this one-line correction.

perrotta:
please test

tlampen:
please test workflow 7.2,7.4

Configuration/PyReleaseValidation/python/relval_steps.py
```diff
@@ -2851,16 +2851,6 @@ def gen2018HiMix(fragment,howMuch):
 
 steps['RECOFROMRECOSt2']=steps['RECOFROMRECO']
 
-steps['RECODFROMRAWRECO']=merge([{'-s':'RAW2DIGI:RawToDigi_noTk,L1Reco,RECO:reconstruction_noTracking,EI',
```
chayanit: @guitargeek always nice to have some clean up :) shall we also clean this up everywhere in relval code?
https://github.com/cms-sw/cmssw/blob/02d4198c0b6615287fd88e9a8ff650aea994412e/Configuration/PyReleaseValidation/python/relval_production.py#L17
https://github.com/cms-sw/cmssw/blob/c1f0db57b8464dfb2388ed74ec25d80df9b6abf7/Configuration/PyReleaseValidation/python/MatrixReader.py#L341
https://github.com/cms-sw/cmssw/blob/c1f0db57b8464dfb2388ed74ec25d80df9b6abf7/Configuration/PyReleaseValidation/python/MatrixReader.py#L243

guitargeek: Hi, Good to know you are ok with this! To whom did you direct your question to? I agree of course that this should be cleaned so nobody in the future is confused about this, but I think  as PdmV convenor it's your decision to make?

tlampen:
+1

Configuration/PyReleaseValidation/python/relval_steps.py
kpedro88: I think @chayanit is proposing that you include the additional cleanup in this PR. I am also in favor of this.

guitargeek: Yes you are right, so let's do that

kpedro88:
please test

tlampen:
+1

chayanit:
+1

kpedro88:
+upgrade

RecoEgamma/EgammaElectronProducers/python/gsfElectronSequence_cff.py
```diff
@@ -7,31 +7,14 @@
 #==============================================================================
 
 from RecoEgamma.EgammaElectronProducers.ecalDrivenGsfElectronCores_cfi import ecalDrivenGsfElectronCores
-from RecoEgamma.EgammaElectronProducers.gsfElectronCores_cfi import gsfElectronCores
 from RecoEgamma.EgammaElectronProducers.ecalDrivenGsfElectronCoresFromMultiCl_cff import ecalDrivenGsfElectronCoresFromMultiCl
 from RecoEgamma.EgammaElectronProducers.gsfElectrons_cfi import *
```
perrotta: This import can probably also get removed now

RecoEgamma/EgammaElectronProducers/python/gsfElectronSequence_cff.py
```diff
@@ -7,31 +7,14 @@
...
-gsfElectronTask = cms.Task(ecalDrivenGsfElectronCores,ecalDrivenGsfElectrons,gsfElectronCores,gsfElectrons)
-gsfElectronSequence = cms.Sequence(gsfElectronTask)
 
 gsfEcalDrivenElectronTask = cms.Task(ecalDrivenGsfElectronCores,ecalDrivenGsfElectrons)
 gsfEcalDrivenElectronSequence = cms.Sequence(gsfEcalDrivenElectronTask)
```
perrotta: This sequence seems to be only used in a few (outdated?) test scripts.
To be possibly considered in a future further cleaning campaing, while I agree that it could add unnecessary complications to add it here.

perrotta:
@Sam-Harper, @jainshilpi, @lsoffi, @afiqaize, @SohamBhattachary please comment about the removal of modules and sequences in this PR: is EGamma ok with the proposed cleaning?

RecoEgamma/EgammaElectronProducers/python/gsfElectronSequence_cff.py
guitargeek: It needs to stay, because this file also contains the `ecalDrivenGsfElectrons`. I would not bother renaming it now thought, because at some point there will be probably an autogenerated ecalDrivenGsfElectrons_cfi anyway.

Sam-Harper:
for me, we might have some mild use cases for obscure analyses for gsfElectrons. However I think its very marginal and honestly I'm in favour of it deleting it and then adding it back / coming to a better solution.

jainshilpi:
We will be posting on hypernews as well just to make sure there is no special case about using run I electrons

guitargeek:
Had to do a quick rebase because https://github.com/cms-sw/cmssw/pull/28505 changed some line in `GsfElectronProducer.cc`, which gets deleted here. Nothing else changed.

perrotta:
please test

Sam-Harper:
So sorry for the late realisation here.  It is extremely handy to be able to produce GsfElectrons without going through particle flow for development purposes. Deleting the normal GsfElectron producer would break phase-II reco efforts right now. 

This is not to say I think it should exist, however the workflow to produce GsfElectrons without going through particle flow must be preserved. Maybe the GED version actually can be configured to do that, in fact to be honest, it should just be one producer which can do both. 

I'm also not opposed to deleting it and then adding a new version back in as a form of extreme code clean up but we do need to preserve this workflow.

jainshilpi:
We checked the sequence used in our phase II upgrade efforts and they run fine - so we are fine with it. We will just make sure one more time on the hypernews if we hear something by Wednesday. After that we are good to go - just want to make sure that we are not imperiling anything with this ...

perrotta:
@Sam-Harper , did I read correctly your message if I say that you suggest to keep (even if modified) a version of the GsfElectron producer in CMSSW? This would require at least to modify this PR. And isn't it in contradiction with what @jainshilpi is writing here? 
Please clarify

jainshilpi:
@perrotta , yes this comment from me was after discussions with Sam. But I let @Sam-Harper clarify ....

guitargeek:
Hi @Sam-Harper, thanks for your comment! I think the `GsfElectronBaseProducer` can indeed also be used to get GsfElectrons without particle flow. That's what the ecalDrivenElectrons and lowPtElectrons do, and the lowPtElectrons also use both ecalDriven and trackerDriven seeds just like your old gsfElectrons I think.

Actually, don't you use the ecalDrivenGsfElectrons for phase II studies? That's at least what i thought naively because some phaseII modifiers are playing around with them:
https://github.com/cms-sw/cmssw/blob/master/RecoEgamma/EgammaElectronProducers/python/gsfElectronSequence_cff.py#L36

Maybe that's why your and Shilpis statements seem to contradict here? Al least we have a good example how having all these electron sequences can be confusing :)

jainshilpi:
Hi @guitargeek , yes we use the ecalDriveGsfElectrons for phase II studies for now and that is why all the tests could be passed. So Sam's comment came before I mentioned to him about the checks I had done using your commits + phase II upgrade studies. All the checks had passed and from discussions with @Sam-Harper , he seems to be fine as well with the removal. So at least in our phase II efforts, we are not using the Run I electrons for now. And I haven't heard of it in hypernews as well. so in my opinion, we can remove this. If we need it in some rare case later, we can re-activate the flow (hoping that option would always be there).

kpedro88:
What is the status of this PR?

perrotta:
@Sam-Harper, @jainshilpi, @lsoffi @guitargeek 

Returning to this open PR after some time, I must admit that I'm still a bit lost about your comments.
My understanding after reading your exchanges would be: "ok for removing the sequences, but keep in the release the code for the GsfElectronProducer in view of some possible future use in Phase2".
If my understanding is correct, then this PR must be modified to recover that now removed plugin.

However, maybe I just misunderstood some intervention here, and all EGamma POG is fine to remove all GsfElectronProducer and GsfElectronCoreProducer plugins, together with the sequences wich refer to them.

Could you please make a clear statement about what above? After that, I think the review can proceed quickly to conclusion.

Thank you, and happy holidays!

jainshilpi:
Hi @perrotta ,

Currently the old sequence is not used in EGM. We can just remove it altogether (including the GsfElectronProducer and GsfElectronCoreProducer plugins etc). So the current PR is fine and no changes are needed. If needed at any point in any rare case, we can just bring the old code and copy to the new release.

Thanks and bonne année!

perrotta:
+1
- GsfElectrons and related producers are removed from CMSSW, as well as all (by now outdated) sequences and workflows which rely on them.
- The removal is [blessed by the EGM POG](https://github.com/cms-sw/cmssw/pull/28429#issuecomment-569986737), as those modules are not used any more in current configuration, nor are they expected to be used as such for Phase2
- Jenkins tests pass and show no changes, as expected

perrotta:
please test
(refresh the tests, to have this PR tested against the latest IB)

kpedro88:
+upgrade

chayanit:
+1

silviodonato:
+operations

silviodonato:
alca (@christopheralanwest  @franzoni  @tlampen  @pohsun  @tocheng ) please review/sign

tlampen:
+1

silviodonato:
+1

PR #28430, opened 2019-11-20
Add variables for PileUp Jet ID BDT in JME Custom-NanoAODs.

nurfikri89:
PR description:

This PR adds jet variables needed to train the BDT for PileUp jet ID in the JME Custom-NanoAODs which are not included in the main NanoAODs. The variables are not stored in MiniAODs so they need to be re-calculated again by scheduling [pileupJetIdCalculator](https://github.com/nurfikri89/cmssw/blob/0f44de31272b6a3807c27a7d5e536efc9608c1e2/PhysicsTools/NanoAOD/python/custom_jme_cff.py#L346). Then, the variables need to be stored in ValueMaps and this is done by [scheduling](https://github.com/nurfikri89/cmssw/blob/0f44de31272b6a3807c27a7d5e536efc9608c1e2/PhysicsTools/NanoAOD/python/custom_jme_cff.py#L357) PileupJetIDVarProducer [plugin](https://github.com/nurfikri89/cmssw/blob/0f44de31272b6a3807c27a7d5e536efc9608c1e2/PhysicsTools/NanoAOD/plugins/PileupJetIDVarProducer.cc) which will be added also. The [rest](https://github.com/nurfikri89/cmssw/blob/0f44de31272b6a3807c27a7d5e536efc9608c1e2/PhysicsTools/NanoAOD/python/custom_jme_cff.py#L366-L395) is just normal procedure to save variables in NanoAODs. At the moment, these variables are set to be calculated and stored only for AK4PFCHS jets.

This PR needs to be backported to 10_6_X, once merged, for the ultra legacy campaigns.

### DIFF
--- a/PhysicsTools/NanoAOD/plugins/PileupJetIDVarProducer.cc
+++ b/PhysicsTools/NanoAOD/plugins/PileupJetIDVarProducer.cc
@@ -11,7 +11,7 @@
 #include "DataFormats/JetReco/interface/PileupJetIdentifier.h"
 
 class PileupJetIDVarProducer : public edm::global::EDProducer<> {
-  
+
 public:
   explicit PileupJetIDVarProducer(const edm::ParameterSet& iConfig)
       : srcJet_(consumes<edm::View<pat::Jet>>(iConfig.getParameter<edm::InputTag>("srcJet"))),
@@ -26,12 +26,10 @@ public:
     produces<edm::ValueMap<float>>("frac04");
     produces<edm::ValueMap<float>>("ptD");
     produces<edm::ValueMap<float>>("beta");
-    produces<edm::ValueMap<float>>("betaStar");
     produces<edm::ValueMap<float>>("pull");
     produces<edm::ValueMap<float>>("jetR");
     produces<edm::ValueMap<float>>("jetRchg");
     produces<edm::ValueMap<int>>("nCharged");
-    produces<edm::ValueMap<int>>("nNeutral");
   }
   ~PileupJetIDVarProducer() override{};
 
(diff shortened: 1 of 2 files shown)


fabiocos:
@peruzzim @fgolf could you please scrutinize this PR? It is fully in your domain

fabiocos:
please test

peruzzim:
+xpog

The code looks ok to me, from a technical point of view. Modifications are confined to the `PrepJMECustomNanoAOD(process)` customise function, and will not affect standard NANOAOD production.
On the other hand, I think that no relval workflow is currently implemented to test the JME-custom NANOAOD production. I assume it was thoroughly tested by JME to ensure it satisfy their needs.
In any case, I would encourage JME to add this check, to be sure we don't inadvertently break this workflow in the future.

fabiocos:
@nurfikri89 could you please comment about the performed validation?

santocch:
+1

fabiocos:
@nurfikri89 could you please comment about the performed validation?

nurfikri89:
Hi @fabiocos, the only validation for this PR that we had done is to check that the variables we are adding are the same as the ntuples produced by PileUp Jet ID analyzers (like [this](https://github.com/singh-ramanpreet/PUjetID/blob/master2019/pileupJetId/plugins/JMEFlatTreeProducer.cc)) and that the values are not different between the CustomNanoAOD and the ntuples. Maybe the JMAR convenors can comment more on this. @camclean @alefisico
 
Apologies. I have not setup any validation workflow as suggested by @peruzzim. Is there an example which I can based on?

fabiocos:
@nurfikri89 for the PR itself I believe that your reference can be ok, but I agree with @peruzzim that adding a dedicated workflow to produce this non-standard NANOAOD (in the spirit of one of the existing 132X workflows) would be useful.

peruzzim:
If this workflow will be run centrally by PdmV, as I understand will be the case, they will need to define a pilot in the relval system to check it before injection. I think would be the right occasion to have them add the configuration to the relval matrix tests.

fabiocos:
+1

PR #28437, opened 2019-11-20
Low energy GFlash-like parameterization

dmitrySorokin:
**PR description:**
Implement GFlash-like parameterization for low energy electrons in PbWO4. 
This is meant to be used for EcalRegion. This model is disabled by default and controlled by LowEnergyGflashEcal flag. 

**PR validation:**
https://indico.cern.ch/event/861790/contributions/3630534/attachments/1941836/3220043/fast_sim.pdf


civanch:
@dmitrySorokin , it would be good to update this PR:
1) move two new classes to SimG4Core/GFlash
2) fitted parameters values can be defined as named constexp

SimG4Core/Application/python/g4SimHits_cfi.py
```diff
@@ -148,6 +148,9 @@
         EMPhysics   = cms.untracked.bool(True),
         HadPhysics  = cms.untracked.bool(True),
         FlagBERT    = cms.untracked.bool(False),
+        LowEnergyGflashEcal = cms.bool(False),
+        LowEnergyGflashEcalEmin = cms.double(0),
```
civanch: @dmitrySorokin , can Emin be non zero?

SimG4Core/Application/src/LowEnergyFastSimModel.cc
```diff
@@ -0,0 +1,62 @@
...
+  return energy > fEmin && energy < fEmax && fRegion == fastTrack.GetEnvelope();
+}
+
+void LowEnergyFastSimModel::DoIt(const G4FastTrack& fastTrack, G4FastStep& fastStep) {
+  fastStep.KillPrimaryTrack();
+  fastStep.SetPrimaryTrackPathLength(0.0);
+  fastStep.SetTotalEnergyDeposited(fastTrack.GetPrimaryTrack()->GetKineticEnergy());
+
+  const G4double energy = fastTrack.GetPrimaryTrack()->GetKineticEnergy();
+  const G4ThreeVector pos = fastTrack.GetPrimaryTrack()->GetPosition();
```
civanch: may be 
const G4ThreeVector&

SimG4Core/Application/src/LowEnergyFastSimModel.cc
```diff
@@ -0,0 +1,62 @@
...
+
+  G4double inPointEnergy = param.GetInPointEnergyFraction(energy) * energy;
+
+  const auto momDir = fastTrack.GetPrimaryTrack()->GetMomentumDirection();
```
civanch: here const G4ThreeVector& seems to be better

SimG4Core/Application/python/g4SimHits_cfi.py
dmitrySorokin: Yes. One can apply this parametrization in any range. 
Like in GFlash it works from 50 GeV to 10 GeV.

SimG4Core/Application/src/LowEnergyFastSimModel.cc
dmitrySorokin: Yes, sure

SimG4Core/Application/src/LowEnergyFastSimModel.cc
dmitrySorokin: Yes, thanks

SimG4CMS/Calo/src/CaloSD.cc
```diff
@@ -341,8 +341,7 @@ bool CaloSD::checkHit() {
 
 int CaloSD::getNumberOfHits() { return theHC->entries(); }
 
-CaloG4Hit* CaloSD::createNewHit(const G4Step* aStep) {
-  auto const theTrack = aStep->GetTrack();
+CaloG4Hit* CaloSD::createNewHit(const G4Step* aStep, const G4Track* theTrack) {
```
dmitrySorokin: @civanch I added theTrack parameter here because in case of GFlash hits aStep->GetTrack() will return nullptr and there is no way except const_cast to set it.

dmitrySorokin (commented):
> @dmitrySorokin , it would be good to update this PR:
> 
> 1. move two new classes to SimG4Core/GFlash

I think it would be better to keep them in Application/ because  in ParametrisedEMPhysics.cc GFlash models come from Application/ and not from GFlash/

civanch:
please test

civanch:
+1

civanch:
I agree, that this development should go to SimG4Core/Application.

fabiocos:
@civanch this development is switched off by default, so merging it is safe. The validation linked to the description is useful to understand the detailed response to the algorithm, anyway I think that a check on full physics events is needed.

fabiocos:
+1

PR #28442, opened 2019-11-21
PFlow fix/improvement in the HF region for phase2

hatakeyamak:
#### PR description:

This PR introduces the first attempt of track-HF cluster matching in PFAlgo for phase 2. It shouldn't impact Run2 and Run3 reconstruction, but improves phase2 reconstruction. Before this PR, tracks and HF clusters in |eta|=3-4 region have been heavily double-counting charged particles in this region, which will be largely remedied in this PR. This PR introduces kdtree-based linking between tracks and HF clusters, and links are utilized in mostly createCandidatesHF of PFAlgo.

Also, the propagator is fixed to make the track propagation to HF to work for particles with pt<3 GeV also (which was broken in the past, as it didn't matter).

Also, the HF face is updated to be in agreement with the actual geometry.

#### PR validation:

http://hep.baylor.edu/hatake/PFval/PFForwardTrack/stack_offset_SinglePiPt50Eta3p45_3p54_2026D41_PU200_v0_vs_v5.pdf
shows the improved offset quantities. [Horns around |eta|=3-4 much reduced with this update.]

I made sure, using PF validation code, that 2018 QCD plots show no differences.

This is not a backport.

@bendavid @jpata @iashvili @mariadalfonso


perrotta:
please test

slava77:
the crash looks related
```
PFAlgo::createCandidatesHF(const reco::PFBlock&, reco::PFBlock::LinkData&, c...): 
Assertion `inds.hfEmIs.size() + inds.hfHadIs.size() == elements.size()'
```

hatakeyamak:
> the crash looks related
> 
> ```
> PFAlgo::createCandidatesHF(const reco::PFBlock&, reco::PFBlock::LinkData&, c...): 
> Assertion `inds.hfEmIs.size() + inds.hfHadIs.size() == elements.size()'
> ```

Thx. This is good to know. Will address it soon.

slava77:
@hatakeyamak 
please clarify if this is required for 11_0_0, or just nice to have

hatakeyamak:
i really would love to have it in so that we can use various central samples for further validation and tuning. What's currently done is wrong..

kpedro88:
assign upgrade

hatakeyamak:
I think the segfault was due to PFBlocks with tracks in the HF region, but tracks are inactive when we reach createCandidatesHF (tracks linked to HFEM only, not to HFHAD, are already converted to charged hadrons in elementLoop). This case should be handled properly now..

slava77:
@cmsbuild please test

hatakeyamak:
Looks like comparisons show what we expect i.e. some reduction of pf candidates in eta=3-4 due to more proper calo-track matching. Example:
https://cmssdt.cern.ch/SDT/jenkins-artifacts/baseLineComparisons/CMSSW_11_0_X_2019-11-22-1100+f8d12e/34305/validateJR/all_mini_OldVSNew_TTbar14TeV2026D35wf20034p0/all_mini_OldVSNew_TTbar14TeV2026D35wf20034p0c_patPackedCandidates_packedPFCandidates__RECO_obj_eta.png

CommonTools/BaseParticlePropagator/src/BaseParticlePropagator.cc
```diff
@@ -506,11 +506,10 @@ bool BaseParticlePropagator::propagateToHcalEntrance(bool first) {
...
 bool BaseParticlePropagator::propagateToVFcalEntrance(bool first) {
   //
-  // Propagation to VFCAL entrance
-  // TODO: include proper geometry
-  // Geometry taken from DAQ TDR Chapter 13
+  // Propagation to VFCAL (HF) entrance
+  // Geometry taken from xml: Geometry/CMSCommonData/data/cms.xml
 
-  setPropagationConditions(400.0, 1110.0, first);
+  setPropagationConditions(400.0, 1114.95, first);
```
kpedro88: as a future development item, it would be even better to get these directly from the geometry instead of hardcoding the values

RecoParticleFlow/PFProducer/plugins/PFProducer.cc
```diff
@@ -448,5 +455,10 @@ void PFProducer::fillDescriptions(edm::ConfigurationDescriptions& descriptions)
   desc.add<std::vector<double>>("calibHF_b_HADonly", {1., 1., 1., 1., 1., 1., 1., 1., 1., 1.});
   desc.add<std::vector<double>>("calibHF_b_EMHAD", {1., 1., 1., 1., 1., 1., 1., 1., 1., 1.});
 
+  // resolution parameters for HF: EPJC 53(2008)139, doi:10.1140/epjc/s10052-007-0459-4
+  desc.add<double>("resolHF_a", 2.799)->setComment("HF resolution - stochastic term");
```
kpedro88: as another future development item, we should consider how these values change with aging

RecoParticleFlow/PFProducer/plugins/kdtrees/KDTreeLinkerTrackHF.cc
```diff
@@ -0,0 +1,245 @@
...
+  if (track->trackRefPF()->extrapolatedPoint(reco::PFTrajectoryPoint::VFcalEntrance).isValid()) {
+    targetSet_.insert(track);
+  }
+}
+
+void KDTreeLinkerTrackHF::insertFieldClusterElt(reco::PFBlockElement* hfCluster) {
+  const reco::PFClusterRef& clusterref = hfCluster->clusterRef();
+
+  // This test is more or less done in PFBlockAlgo.h. In others cases, it should be switch on.
+  //   if (!((clusterref->layer() == PFLayer::HF_ENDCAP) ||
```
kpedro88: delete commented-out code

RecoParticleFlow/PFProducer/plugins/kdtrees/KDTreeLinkerTrackHF.cc
```diff
@@ -0,0 +1,245 @@
...
+
+  const std::vector<reco::PFRecHitFraction>& fraction = clusterref->recHitFractions();
+
+  // We create a list of hfCluster
+  fieldClusterSet_.insert(hfCluster);
+  for (size_t rhit = 0; rhit < fraction.size(); ++rhit) {
+    const reco::PFRecHitRef& rh = fraction[rhit].recHitRef();
+    double fract = fraction[rhit].fraction();
+
+    if ((rh.isNull()) || (fract < 1E-4))
```
kpedro88: magic number 1e-4 should be named constant

RecoParticleFlow/PFProducer/plugins/kdtrees/KDTreeLinkerTrackHF.cc
```diff
@@ -0,0 +1,245 @@
...
+
+    if (trackphi > M_PI)
+      trackphi -= 2 * M_PI;
+    else if (trackphi < -M_PI)
+      trackphi += 2 * M_PI;
+
+    // Estimate the maximal envelope in phi/eta that will be used to find rechit candidates.
+    // Same envelope for cap et barrel rechits.
+    double inflation = 1.;
+    float rangeeta = (cristalPhiEtaMaxSize_ * (1.5 + 0.5) + 0.2 * fabs(dHeta)) * inflation;
```
kpedro88: there are a lot of magic numbers here. in particular, it seems that (1.5 + 0.5) could be simplified to 2.0...

RecoParticleFlow/PFProducer/src/PFAlgo.cc
```diff
@@ -1200,16 +1251,366 @@ int PFAlgo::decideType(const edm::OwnVector<reco::PFBlockElement>& elements,
...
+      double uncalibratedenergyHfEm = 0.;
+
+      // estimated HF resolution and track p error
+      double Caloresolution = hfEnergyResolution(totalChargedMomentum);
+      Caloresolution *= totalChargedMomentum;
+      double TotalError = sqrt(Caloresolution * Caloresolution + sumpError2);
+      double nsigmaHFEM = nSigmaHFEM(totalChargedMomentum);
+      double nsigmaHFHAD = nSigmaHFHAD(totalChargedMomentum);
+
+      //
```
kpedro88: delete unnecessary empty lines

RecoParticleFlow/PFProducer/src/PFAlgo.cc
```diff
@@ -1200,16 +1251,366 @@ int PFAlgo::decideType(const edm::OwnVector<reco::PFBlockElement>& elements,
...
+          if (active[iHfEm])
+            continue;
+          (*pfCandidates_)[tmpi].addElementInBlock(blockref, iHfEm);
+        }
+        double frac = 0.;
+        if (totalChargedMomentum)
+          frac = trackRef->p() / totalChargedMomentum;
+        (*pfCandidates_)[tmpi].setEcalEnergy(uncalibratedenergyHfEm * frac, energyHfEm * frac);
+        (*pfCandidates_)[tmpi].setHcalEnergy(uncalibratedenergyHfHad * frac, energyHfHad * frac);
+        (*pfCandidates_)[tmpi].setHoEnergy(0., 0.);
```
kpedro88: since these three calls (HO, PS energy) happen repeatedly, might be useful to encapsulate them in a helper function (to reduce repetition)

RecoParticleFlow/PFProducer/src/PFAlgo.cc
```diff
@@ -2968,6 +3375,28 @@ double PFAlgo::nSigmaHCAL(double clusterEnergyHCAL, double eta) const {
   return nS;
 }
 
+double PFAlgo::hfEnergyResolution(double clusterEnergyHF) const {
+  // Add a protection
+  if (clusterEnergyHF < 1.)
```
kpedro88: `clusterEnergyHF = std::max(clusterEnergyHF,1.);`

RecoParticleFlow/PFProducer/src/PFAlgo.cc
```diff
@@ -2968,6 +3375,28 @@ double PFAlgo::nSigmaHCAL(double clusterEnergyHCAL, double eta) const {
...
+
+  double resol =
+      sqrt(pow(resolHF_a_, 2) / clusterEnergyHF + pow(resolHF_b_, 2) + pow(resolHF_c_, 2) / pow(clusterEnergyHF, 2));
+  // a: stochastic term, b: constant term, c: noise term
+
+  return resol;
+}
+
+double PFAlgo::nSigmaHFEM(double clusterEnergyHF) const {
+  double nS = nSigmaHFEM_ * (1. + exp(-clusterEnergyHF / 100.));
```
kpedro88: magic number 100 should be named constant

kpedro88:
@hatakeyamak I'm also curious if this change has any impact on the CPU usage of particle flow.

hatakeyamak:
> @hatakeyamak I'm also curious if this change has any impact on the CPU usage of particle flow.

I haven't. It really shouldn't impact for run2/run3, but probably some for phase2. Is it possible to activate igprof test in jenkins's test.

Thanks for comments. I will try to address them over the next couple of days.

RecoParticleFlow/PFProducer/src/PFAlgo.cc
```diff
@@ -2717,8 +3118,14 @@ void PFAlgo::processBlock(const reco::PFBlockRef& blockref,
...
   // Reconstruct pfCandidate from HF (either EM-only, Had-only or both)
+  // For phase2, process also pfblocks containing HF clusters and linked tracks
   if (!(inds.hfEmIs.empty() && inds.hfHadIs.empty())) {
-    createCandidateHF(block, blockref, elements, inds);
+    createCandidatesHF(block, linkData, elements, active, blockref, inds);
+    if (inds.hcalIs.empty() && inds.ecalIs.empty())
+      return;
+    edm::LogWarning("PFAlgo::processBlock")
+        << "Block contains HF clusters, but also contains ECAL or HCAL clusters. Continue.";
+    edm::LogWarning("PFAlgo::processBlock") << block;
```
slava77: How frequent and unlikely is this condition?
LogWarnings are supposed to show some significant issues in the execution.
It looks like the condition for this warning is a property of the algorithm configuration, which means that this is not a warning but a debug.

Please issue only one LogWarning here, simply ```Continue.\n" << block``` should work here

RecoParticleFlow/PFProducer/src/PFAlgo.cc
```diff
@@ -2968,6 +3375,28 @@ double PFAlgo::nSigmaHCAL(double clusterEnergyHCAL, double eta) const {
   return nS;
 }
 
+double PFAlgo::hfEnergyResolution(double clusterEnergyHF) const {
+  // Add a protection
+  if (clusterEnergyHF < 1.)
+    clusterEnergyHF = 1.;
+
+  double resol =
+      sqrt(pow(resolHF_a_, 2) / clusterEnergyHF + pow(resolHF_b_, 2) + pow(resolHF_c_, 2) / pow(clusterEnergyHF, 2));
```
slava77: please save squares of resolHF_\*_ parameters to avoid unnecessary recomputations

perrotta:
Since in the review there are already requests to update this PR, you could profit and get rid of the dead assignments in CommonTools/BaseParticlePropagator/src/BaseParticlePropagator.cc pointed out by the static analyzer, and simply remove L90 (zProp) and L393 (dist2)

hatakeyamak:
Okay. Thank you.

Just to keep people updated, I am working on implementing suggestions.
Probably the hardest one to incorporate is introducing a template or one more layer of inheritance for TrackAndHFLinker.cc and KDTreeLinkerTrackHF.cc. I wonder if it's required.

slava77:
>  Probably the hardest one to incorporate is introducing a template or one more layer of inheritance for TrackAndHFLinker.cc and KDTreeLinkerTrackHF.cc. I wonder if it's required.

I prefer that this is resolved if not immediately then very soon after (considering the needs and the current deadlines).

RecoParticleFlow/PFProducer/src/PFAlgo.cc
hatakeyamak: Thx. For now, just removed these three lines inside createCandidatesHF, which should just give 0 energies anyway, assuming that's preferred.

hatakeyamak:
It took longer than I hoped, but I addressed most of comments I think.
Still thinking if KDTreeLinkerTrackHF and possibly TrackAndHFLinker can be simplified by template/inheritance.

slava77:
@cmsbuild please test

hatakeyamak:
> It took longer than I hoped, but I addressed most of comments/suggestions I think.
> Still thinking if KDTreeLinkerTrackHF and possibly TrackAndHFLinker can be simplified by template/inheritance.

One possibility I can think of is to try to use KDTreeLinkerTrackHCAL and TrackAndHCALLinker for HF too, get rid of KDTreeLinkerTrackHF/TrackAndHFLinker, and try to make PFTrajectoryPoint configurable (set to HCALEntrance/HCALExit or VFcalEntrance). Make some if statement for case we utilize Exit (for now yes for HCAL[HBHE] and no for HF).

If this sounds like a reasonable approach, I can give it a try to see if it really helps.

slava77:
> One possibility I can think of is to try to use KDTreeLinkerTrackHCAL and TrackAndHCALLinker for HF too, get rid of KDTreeLinkerTrackHF/TrackAndHFLinker, and try to make PFTrajectoryPoint configurable (set to HCALEntrance/HCALExit or VFcalEntrance). Make some if statement for case we utilize Exit (for now yes for HCAL[HBHE] and no for HF).
> 
> If this sounds like a reasonable approach, I can give it a try to see if it really helps.
> 
> Or do you have a better suggestion? @slava77

This sounds like a good solution to the code replication problem.
Please go ahead with it.

hatakeyamak:
ok. will give it a try.

hatakeyamak:
I implemented what we discussed above. We removing KDTreeLinkerTrackHF & TrackAndHFLinker, and now we use KDTreeLinkerTrackHcal & TrackAndHCALLinker for both HBHE and HF.

I introduced a new virtual function to KDTreeLinkerBase which is necessary only for KDTreeLinkerTrackHcal. I am not sure if this is the best way to code, but this is the best I could do.

kpedro88:
please test

kpedro88:
-1
@hatakeyamak please follow http://cms-sw.github.io/tutorial-resolve-conflicts.html in order to rebase PRs correctly

hatakeyamak:
> -1
> @hatakeyamak please follow http://cms-sw.github.io/tutorial-resolve-conflicts.html in order to rebase PRs correctly

Right. I think rebase is good now.

Martin-Grunewald:
please test

hatakeyamak:
I think this addresses the crash in the PPOnAA workflow.

slava77:
@cmsbuild please test

slava77:
> ---> test testTauEmbeddingProducers had ERRORS

this is unrelated to this PR.
The problem showed up after #28541 was integrated and I expect that it is fixed in #28567

slava77:
@cmsbuild please test with #28567

slava77:
+1

for https://github.com/cms-sw/cmssw/pull/28480	53c7190 
- code changes are in line with the PR description and the follow up review; the new code is not enabled in available standard configuration(s), to be used first at the analysis level

hatakeyamak:
Hi @slava77

Thank you. Just to be clear, this first PFAlgo in the HF region using tracks for phase2 is enabled in the phase2 configuration (of course, disabled for pre-phase2), and that's why we see changes only in phase2 samples, for example:
https://cmssdt.cern.ch/SDT/jenkins-artifacts/baseLineComparisons/CMSSW_11_1_X_2019-12-05-1100+f90831/34469/validateJR/all_OldVSNew_TTbar14TeV2026D35wf20034p0/all_OldVSNew_TTbar14TeV2026D35wf20034p0c_recoPFCandidates_particleFlow__RECO_obj_eta.png
Changes look as expected to me.

Since there were a few small changes during the review, I double-checked the offset plots to see changes w.r.t, the time of submission:
http://hep.baylor.edu/hatake/PFval/PFForwardTrack/stack_offset_SinglePiPt50Eta3p45_3p54_2026D41_PU200_v5_vs_v8.pdf
No major changes are seen as expected.

I am wondering if the backport of this PR can be considered. This PR was listed in:
https://github.com/cms-sw/cmssw/issues/28455
When various phase2 samples are produced with 11_0 release, I think it will be good to include this update.

slava77:
> Hi @slava77
> 
> Thank you.

sorry, the earlier +1 was posted to an incorrect thread. I still need to review the latest updates here.

DataFormats/ParticleFlowReco/interface/PFTrajectoryPoint.h
```diff
@@ -52,10 +52,35 @@ namespace reco {
...
+      m["ECALEntrance"] = LayerType::ECALEntrance;
+      m["ECALShowerMax"] = LayerType::ECALShowerMax;
+      m["HCALEntrance"] = LayerType::HCALEntrance;
+      m["HCALExit"] = LayerType::HCALExit;
+      m["HOLayer"] = LayerType::HOLayer;
+      m["VFcalEntrance"] = LayerType::VFcalEntrance;
+      return m;
+    }
+    static LayerType layerTypeFromString(std::string layerTypeString) {
+      std::map<std::string, LayerType> layerTypeMap = create_map();
```
kpedro88: I think it might be preferable for a few reasons to define the map as a constant within this function:
```cpp
const std::map<std::string, LayerType> layerTypeMap{
  {"ClosestApproach", LayerType::ClosestApproach},
  {"BeamPipeOrEndVertex", LayerType::BeamPipeOrEndVertex},
  {"PS1", LayerType::PS1},
  {"PS2", LayerType::PS2},
  {"ECALEntrance", LayerType::ECALEntrance},
  {"ECALShowerMax", LayerType::ECALShowerMax},
  {"HCALEntrance", LayerType::HCALEntrance},
  {"HCALExit", LayerType::HCALExit},
  {"HOLayer", LayerType::HOLayer},
  {"VFcalEntrance", LayerType::VFcalEntrance},
};
```

makortel: This looks like a very expensive way to map a name of an enum to the enum (even with the pre-constructed `std::map`). Linear search over a static array would require less memory and likely be faster (see e.g. 
https://github.com/cms-sw/cmssw/blob/431e2fdfedec052e73c16e9f06de98ade41ebc56/DataFormats/TrackReco/src/TrackBase.cc#L137-L142
for an example)

hatakeyamak: Okay. This will be called only at the beginning of each job 1-3 times, and not for each event, so I don't think it's critical to make it less expensive, but let me think..

makortel: If only used in job startup the `std::map` could be acceptable (assuming it gets destroyed after last use to not "hoard" memory). This leaves the main argument against to be "bad example for others".

hatakeyamak: Thanks. At the end, I just adopted your string to enum conversion method. I agree that it's nicer.

kpedro88:
please test

jfernan2:
Due to the change in PF candidates this PR is affecting many MEs across several folders including:  
Btag, JetMET, ParticleFlow, Physics, RecoTauV and Tracking
but only in Phase2 samples as expected.
Nevertheless, before signing, @hatakeyamak I would like to drive your attention to these plots that, despite statistics, look different to your validation (probably the observables are not comparable, it just got my attention since baseline indeed had entries in |eta|~3 but the PR seems to move entries to |eta|>4), just in case, probably it is understood.
https://tinyurl.com/t7wwvg3
https://tinyurl.com/r89a69h
Sorry if the comment is irrelevant.

hatakeyamak:
> Due to the change in PF candidates this PR is affecting many MEs across several folders including:
> Btag, JetMET, ParticleFlow, Physics, RecoTauV and Tracking
> but only in Phase2 samples as expected.
> Nevertheless, before signing, @hatakeyamak I would like to drive your attention to these plots that, despite statistics, look different to your validation (probably the observables are not comparable, it just got my attention since baseline indeed had entries in |eta|~3 but the PR seems to move entries to |eta|>4), just in case, probably it is understood.
> https://tinyurl.com/t7wwvg3
> https://tinyurl.com/r89a69h
> Sorry if the comment is irrelevant.

Actually, this is a nice catch. Please go to these links, and undo "normalization" by clicking CMS logo. We see what we expect to see i.e. reduction in HFEM and HFHAD because of linkings to track.

This "normalization" feature tricked me a couple of times in the past too. I think they are booked as histogram after processing in harvester, but for this pt density vs eta plots, we really don't want to normalize. Is there a good way to avoid this? This can be addressed in another validation/dqm PR, I think, if you have suggestions. Thank you.

Hopefully the review goes smoothly in coming days...

jfernan2:
+dqm
Thanks @hatakeyamak . Indeed this is an artifact of the normalization which is set ON by default in comparisons at the DQM GUI to cope with different stats, which is not the case in the bin by bin tool which compares the same number of entries in base and base+pr. 
I will urge DQM core to make as default the non normalized option. Thanks

RecoParticleFlow/PFProducer/src/PFBlockAlgo.cc
```diff
@@ -119,8 +119,22 @@ void PFBlockAlgo::setLinkers(const std::vector<edm::ParameterSet>& confs) {
...
+      const std::string trajectoryLayerExitString(conf.existsAs<std::string>("trajectoryLayerExit")
+                                                      ? conf.getParameter<std::string>("trajectoryLayerExit")
+                                                      : "");
+      // convert TrajectoryLayers info from string to enum
+      const reco::PFTrajectoryPoint::LayerType trajectoryLayerEntrance =
+          reco::PFTrajectoryPoint::layerTypeByName(trajectoryLayerEntranceString);
+      const reco::PFTrajectoryPoint::LayerType trajectoryLayerExit =
+          reco::PFTrajectoryPoint::layerTypeByName(trajectoryLayerExitString);
+      if (trajectoryLayerEntrance != reco::PFTrajectoryPoint::Unknown)
+        kdtrees_.back()->setTrajectoryPoints(trajectoryLayerEntrance, trajectoryLayerExit);
```
slava77: I would just pass conf to the corresponding tree linker and let the concrete linker constructor implementation figure out what to do with the inputs.
This way the ```setTrajectoryPoints``` extension to the interface is not needed (as it stands, it is needed only for one case).

RecoParticleFlow/PFProducer/plugins/linkers/TrackAndHCALLinker.cc
```diff
@@ -11,19 +12,38 @@ class TrackAndHCALLinker : public BlockElementLinkerBase {
...
+    _trajectoryLayerExit = reco::PFTrajectoryPoint::layerTypeByName(_trajectoryLayerExitString);
+    // make sure the requested setting is supported
+    assert((_trajectoryLayerEntrance == reco::PFTrajectoryPoint::HCALEntrance &&
+            _trajectoryLayerExit == reco::PFTrajectoryPoint::HCALExit) ||
+           (_trajectoryLayerEntrance == reco::PFTrajectoryPoint::HCALEntrance &&
+            _trajectoryLayerExit == reco::PFTrajectoryPoint::Unknown) ||
+           (_trajectoryLayerEntrance == reco::PFTrajectoryPoint::VFcalEntrance &&
+            _trajectoryLayerExit == reco::PFTrajectoryPoint::Unknown));
+    // flag if exit layer should be checked or not
+    _checkExit = (_trajectoryLayerExit == reco::PFTrajectoryPoint::Unknown) ? false : true;
```
slava77: ```checkExit_ = trajectoryLayerExit != reco::PFTrajectoryPoint::Unknown; ```

RecoParticleFlow/PFProducer/interface/KDTreeLinkerBase.h
```diff
@@ -23,6 +23,10 @@ class KDTreeLinkerBase {
 
   void setFieldType(const reco::PFBlockElement::Type &fld) { _fieldType = fld; }
 
+  // set trajectory points, necessary for track-HCAL links
+  virtual void setTrajectoryPoints(const reco::PFTrajectoryPoint::LayerType trajectoryPointEntrance,
+                                   const reco::PFTrajectoryPoint::LayerType trajectoryPointExit){};
```
slava77: looking at the use cases, I do not see a good justification to add this method at the base class level.
It seems more practical to me to use a more common pattern of passing configuration to the tool via a PSet.
I suggest to add a ```KDTreeLinkerBase(const edm::ParameterSet& conf)``` and add the appropriate logic to KDTreeLinkerTrackHcal(conf) constructor.

See more in comments for PFBlockAlgo.cc

slava77 (commented):
I think that the current iteration is fairly close to be ready to go. Still, some updates are needed.

DataFormats/ParticleFlowReco/interface/PFTrajectoryPoint.h
```diff
@@ -52,10 +52,15 @@ namespace reco {
...
       VFcalEntrance = 9,
-
-      NLayers = 10
+      // Number of valid layers
+      NLayers = 10,
+      // Unknown
+      Unknown = -1
     };
 
+    static const std::string layerTypeNames[];
```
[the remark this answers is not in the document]
hatakeyamak: Okay. Do you have a good example of doing a search with this fixed length array?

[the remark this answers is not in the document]
kpedro88: @hatakeyamak using `std::array` allows you to use the nicer standard library accessors:
```
std::find(layerTypeNames.begin(), layerTypeNames.end(), name)
```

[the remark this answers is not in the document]
hatakeyamak: Yeah, I thought so, but I get something a compile error and haven't found a solution.
```
>> Compiling  /cms/data/hatake/ana/PF/sl7/CMSSW_11_0_0_pre13_FwdTrkDev/src/DataFormats/ParticleFlowReco/src/PFTrajectoryPoint.cc 
/cms/data/hatake/ana/PF/sl7/CMSSW_11_0_0_pre13_FwdTrkDev/src/DataFormats/ParticleFlowReco/src/PFTrajectoryPoint.cc: In static member function 'static reco::PFTrajectoryPoint::LayerType reco::PFTrajectoryPoint::layerTypeByName2(const string&)':
/cms/data/hatake/ana/PF/sl7/CMSSW_11_0_0_pre13_FwdTrkDev/src/DataFormats/ParticleFlowReco/src/PFTrajectoryPoint.cc:62:24: error: invalid conversion from 'const std::__cxx11::basic_string<char>*' to 'int' [-fpermissive]
   int index = std::find(layerTypeNames2.begin(), layerTypeNames2.end(), name);
               ~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
gmake: *** [config/SCRAM/GMake/Makefile.rules:1742: tmp/slc7_amd64_gcc820/src/DataFormats/ParticleFlowReco/src/DataFormatsParticleFlowReco/PFTrajectoryPoint.cc.o] Error 1
gmake: *** [There are compilation/build errors. Please see the detail log above.] Error 2
```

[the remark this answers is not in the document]
hatakeyamak: Right. I was doing something silly which led to this question, but now I understand and I like this better.

fabiocos:
@hatakeyamak what is the ETA for the next iteration?

hatakeyamak:
> @hatakeyamak what is the ETA for the next iteration?

I think in 1-3 days I will make updates based on @slava77 's suggestions.

HLTrigger/Configuration/python/customizeHLTforCMSSW.py
```diff
@@ -164,10 +164,51 @@ def customiseFor2017DtUnpacking(process):
...
+               linkerName = cms.string( "ECALAndHCALLinker" )
+    ),
+    cms.PSet(  linkType = cms.string( "HFEM:HFHAD" ),
+               useKDTree = cms.bool( False ),
+               linkerName = cms.string( "HFEMAndHFHADLinker" )
+    )
+)
+def customiseFor28442(process):
+    if hasattr(process,'hltParticleFlowBlock'):
+        process.hltParticleFlowBlock.linkDefinitions = hltPFBlockLinkDefPrePhase2
```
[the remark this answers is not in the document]
hatakeyamak: I see what you mean, but syntax to do this wasn't obvious to me, and I think this is sort of temporary until new parameters go to confdbs, so I left this as is for now.

kpedro88:
please test

hatakeyamak:
Thanks @slava77 .
Apart from suggestions on HLTrigger/Configuration/python/customizeHLTforCMSSW.py 
I implemented all suggested changes. I really like the suggested changes.
How does this PR look now?

hatakeyamak:
> I will urge DQM core to make as default the non normalized option. Thanks

Thanks. It looks you already went ahead with this change!

slava77:
> Apart from suggestions on HLTrigger/Configuration/python/customizeHLTforCMSSW.py
> I implemented all suggested changes.

Thank you, I'll take a look ASAP.

I hope that the currently rather verbose/expanded variant in customizeHLTforCMSSW is OK for the HLT

slava77:
@cmsbuild please test

slava77:
@hatakeyamak 
please confirm if you are actually done with the commits.

hatakeyamak:
Now, it should be done. Sorry for additional commit. I was re-reviewing changes, and spotted a couple of things which we had better update.

slava77:
@cmsbuild please test

slava77:
looking at 1000 events in wf 23243.0 (QCD flat-PT, D49, no PU), things are roughly as expected (red is with this PR; black is CMSSW_11_1_X_2019-12-17-1100 out of the box)

- there are more charged hadrons and significantly fewer HF hadrons (and moderately fewer HF EMs) in the |eta| 3-4 range ; charged hadron appearance is restricted to pt>~3 GeV (as in the code)
all_sign1093VSorig_QCD14TeVFlatPt15s3000in2023D49wf23243p0c_recoPFCandidates_particleFlow__RECO_obj_eta39
all_sign1093VSorig_QCD14TeVFlatPt15s3000in2023D49wf23243p0c_recoPFCandidates_particleFlow__RECO_obj_eta74
all_sign1093VSorig_QCD14TeVFlatPt15s3000in2023D49wf23243p0c_recoPFCandidates_particleFlow__RECO_obj_eta81

pfMET is not affected significantly, although it looks a bit closer to gen
wf23243D49QCD_pfmet_genDiff

The jet response is down in the forward eta region
wf23243D49QCD_jetResp_F
wf23243D49QCD_ak4pf_40to200_eta

wf23243D49QCD_ak4pf_20to40_resp1d
the fraction of the reco response well above 1 is much lower now, which is apparently in the right direction (removing double-counting of objects).

slava77:
I added a few more plots to the post above.
Almost complete disappearance of the lost tracks in eta 3-4 seems a bit suspicious https://user-images.githubusercontent.com/4676718/71046511-614d1100-20ed-11ea-86fc-57a1328467c8.png
either the tracks are so pure or the track-HF linking manages to link fakes successfully.
Note that the high-purity fake rate is still not zero in the eta range:
sign1093_20443 0_hp_eff_eta

slava77:
+1

for https://github.com/cms-sw/cmssw/pull/28442	b07fda1 
- code changes are in line with the PR description and the follow up review. The energy flow double-counting issue fixed in this PR is effectively a bug-fix and qualifies for a backport to 11_0.
- jenkins tests pass and comparisons with the baseline show differences only in the phase-2 workflows in plots related to PF candidates, concentrated in the eta range of 3 to 4. This is in agreement with the changes in the code.
- plots made in an extended test of 1K QCD flat-pt events show roughly expected behavior, in line with the PR description and the reco meeting https://indico.cern.ch/event/866456/contributions/3651206/attachments/1954041/3245106/PF_Reco_Nov29-2019.pdf

HLTrigger/Configuration/python/customizeHLTforCMSSW.py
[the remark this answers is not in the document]
Martin-Grunewald: So the only actual change is to add two parameters in the PSet with link type TRACK:HCAL?
Everything else is supposed to be unchanged?

Martin-Grunewald:
+1
(pending answer to my question)

HLTrigger/Configuration/python/customizeHLTforCMSSW.py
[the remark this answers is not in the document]
hatakeyamak: Yes, that's correct.

hatakeyamak:
> +1
> (pending answer to my question)

I added the answer above, but yes, your understanding is correct.

Martin-Grunewald:
thanks

CommonTools/BaseParticlePropagator/src/BaseParticlePropagator.cc
fabiocos: although this is a number that is not supposed to change, I agree with @kpedro88 that this implementation is intrinsicattly weak, although its fix can be deferred to a second moment

hatakeyamak: I understand. If we make this update, I want to update a few other hardcoded numbers, but some hardcoded numbers may not be directly extractable from geometry. It's not a trivial change, so I prefer to keep this separate, while fixing this known small inconsistency for now.

kpedro88:
+upgrade

fabiocos:
@ssekmen @sbein please have a look at this PR, it affects tangentially fastsim and it has been discussed for fast integration into 11_1_X and backport to 11_0_X in view of the HLT TDR production. I would like to integrate it asap, so please check and comment (in case of hot remarks) or sign it

sbein:
+1

fabiocos:
@hatakeyamak in order to prepare the backport requested by @fwyzard could you please provide already the PR there, so as we can launch the tests?

slava77:
> @hatakeyamak in order to prepare the backport requested by @fwyzard could you please provide already the PR there, so as we can launch the tests?

it looks like the same topic branch will work
https://github.com/cms-sw/cmssw/compare/CMSSW_11_0_X...hatakeyamak:Phase2PFTracksForward_v6?expand=1

hatakeyamak:
> > @hatakeyamak in order to prepare the backport requested by @fwyzard could you please provide already the PR there, so as we can launch the tests?
> 
> it looks like the same topic branch will work
> https://github.com/cms-sw/cmssw/compare/CMSSW_11_0_X...hatakeyamak:Phase2PFTracksForward_v6?expand=1

I had impression that I don't have to do anything as @slava77 indicated, but please let me know if anything is necessary for backport.

slava77:
> I had impression that I don't have to do anything as @slava77 indicated, but please let me know if anything is necessary for backport.

you still need to submit a pull request (even though you do not need to create a new branch)

hatakeyamak:
> > I had impression that I don't have to do anything as @slava77 indicated, but please let me know if anything is necessary for backport.
> 
> you still need to submit a pull request (even though you do not need to create a new branch)

Ah, ok. Right, of course. I just did it. #28650

fabiocos:
+1

PR #28446, opened 2019-11-21
remove MTD 'fastsim' modules

kpedro88:
#### PR description:

This PR removes CPU-heavy 4D vertexing modules that use the MTD 'fastsim' time smearing (not to be confused with the actual CMS fast simulation, to which it has no relation). These are no longer required for any purpose, but still consume ~10% of Phase 2 reconstruction CPU time in 200PU.

This PR should be followed up by a reassessment of the usage of `TrackTimeValueMapProducer`, which provides the time smearing inputs (and, in principle, is no longer actually used anywhere).

#### PR validation:

Ran workflow 20434.0 successfully.


kpedro88:
please test

kpedro88:
assign upgrade

kpedro88:
+upgrade
Comparison changes just due to removed collections

perrotta:
+1
- Collections produced by the removed modules aren't used anywhere in the Phase2 workflows
- Jenkins tests pass and the only differences come from the removal of those products in output

fabiocos:
@pmeridian @parbol @bendavid @lgray FYI this update will be integrated tonight, please comment in case

fabiocos:
+1

PR #28453, opened 2019-11-22
GEMMasking :: FixBug

yeckang:
#### PR description:

* GEM strip numbering start from 0. But we used the strip number starting from 1. It has fixed for the GEM masking method.
* GEM masking should be applied at initial clusterization to conceal digis from maskedStrips. It has applied now.
 
#### PR validation:

* Code style has fixed with scram build code-format command.
* runTheMatrix.py is runnable.
* The result has not changed for GEM.

@jshlee @watson-ij

### DIFF
--- a/RecoLocalMuon/GEMRecHit/interface/GEMClusterizer.h
+++ b/RecoLocalMuon/GEMRecHit/interface/GEMClusterizer.h
@@ -6,12 +6,13 @@
 
 #include "RecoLocalMuon/GEMRecHit/interface/GEMClusterContainer.h"
 #include "RecoLocalMuon/GEMRecHit/interface/GEMCluster.h"
+#include "RecoLocalMuon/GEMRecHit/interface/GEMEtaPartitionMask.h"
 #include "DataFormats/GEMDigi/interface/GEMDigiCollection.h"
 
 class GEMClusterizer {
 public:
   GEMClusterizer(){};
   ~GEMClusterizer(){};
-  GEMClusterContainer doAction(const GEMDigiCollection::Range& digiRange);
+  GEMClusterContainer doAction(const GEMDigiCollection::Range& digiRange, const EtaPartitionMask& mask);
 };
 #endif
--- a/RecoLocalMuon/GEMRecHit/plugins/GEMRecHitProducer.cc
+++ b/RecoLocalMuon/GEMRecHit/plugins/GEMRecHitProducer.cc
@@ -108,13 +108,13 @@ void GEMRecHitProducer::beginRun(const edm::Run& r, const edm::EventSetup& setup
       for (const auto& tomask : theGEMMaskedStripsObj->getMaskVec()) {
         if (tomask.rawId == rawId) {
           const int bit = tomask.strip;
-          mask.set(bit - 1);
+          mask.set(bit);
         }
       }
       for (const auto& tomask : theGEMDeadStripsObj->getDeadVec()) {
         if (tomask.rawId == rawId) {
           const int bit = tomask.strip;
-          mask.set(bit - 1);
+          mask.set(bit);
         }
       }
       // add to masking map if masking present in etaPartition
--- a/RecoLocalMuon/GEMRecHit/src/GEMClusterizer.cc
+++ b/RecoLocalMuon/GEMRecHit/src/GEMClusterizer.cc
@@ -1,6 +1,6 @@
 #include "RecoLocalMuon/GEMRecHit/interface/GEMClusterizer.h"
 
-GEMClusterContainer GEMClusterizer::doAction(const GEMDigiCollection::Range& digiRange) {
+GEMClusterContainer GEMClusterizer::doAction(const GEMDigiCollection::Range& digiRange, const EtaPartitionMask& mask) {
   GEMClusterContainer initialCluster, finalCluster;
   // Return empty container for null input
   if (std::distance(digiRange.second, digiRange.first) == 0)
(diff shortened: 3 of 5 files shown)


perrotta:
please test

kpedro88:
+upgrade

perrotta:
@yeckang 

- Strip numbering is modified in GEM masking, taking into account that it starts from 0 and not 1. 
   - One could have get an error in case strip 0 was masked, something which is avoided with this fix. If I understand it correctly, no effect must be expected from this fix for masking any other strip than 0, and in fact no effect is visible in jenkins tests: do you confirm?
- GEM masking also applied at initial clusterization, which is also a fix according to the logic of the algo implemented in GEMRecHitBaseAlgo
  - Here I would have expected some difference in output, because masked/dead strips should have been previously included in the GEM clusters, but not now: is the lack of differences in tests due to the fact that dead/masked strips have so far no signal?
- Since I noticed it, I report here that some cleaning of the code could be achieved if
  - remove #include "RecoLocalMuon/GEMRecHit/interface/GEMClusterizer.h"  from GEMMaskReClusterizer.cc (as not needed)
  - move #include "RecoLocalMuon/GEMRecHit/interface/GEMCluster.h"  from GEMClusterizer.h to GEMClusterizer.cc

yeckang:
@perrotta 
* Strip numbering : It should affect on the other strips not only 0th one. But as far as I know, the masking is not applied at default option. Is Jenkins modifies the options and tests scripts? If it doesn't, I think we could not see any differences at here.

* I can remove #include "RecoLocalMuon/GEMRecHit/interface/GEMClusterizer.h" from GEMMaskReClusterizer.cc.
* The GEMClusterContainer.h doesn't include GEMCluster.h. So if I move it to GEMClusterizer.cc, the complier give an error. So I suggests to move #include "RecoLocalMuon/GEMRecHit/interface/GEMCluster.h" to GEMCluseterContainer.h.

perrotta:
> @perrotta
> 
> * Strip numbering : It should affect on the other strips not only 0th one. But as far as I know, the masking is not applied at default option. Is Jenkins modifies the options and tests scripts? If it doesn't, I think we could not see any differences at here.

Could you please remind us the config changes needed to apply the masking, just to verify that it can run?

> * I can remove #include "RecoLocalMuon/GEMRecHit/interface/GEMClusterizer.h" from GEMMaskReClusterizer.cc.
> * The GEMClusterContainer.h doesn't include GEMCluster.h. So if I move it to GEMClusterizer.cc, the complier give an error. So I suggests to move #include "RecoLocalMuon/GEMRecHit/interface/GEMCluster.h" to GEMCluseterContainer.h.

I see... GEMCluseterContainer.h was designed in analogy with the ​RPCClusterContainer.h.
I think that both them could/should be adjusted, but probably it is not for this PR to do it.

yeckang:
> > @perrotta
> > 
> > * Strip numbering : It should affect on the other strips not only 0th one. But as far as I know, the masking is not applied at default option. Is Jenkins modifies the options and tests scripts? If it doesn't, I think we could not see any differences at here.
> 
> Could you please remind us the config changes needed to apply the masking, just to verify that it can run?

To apply mask, you should add following line at the step3.
```
process.gemRecHits.applyMasking = cms.bool(True)
process.gemRecHits.maskFile = cms.FileInPath("RecoLocalMuon/GEMRecHit/data/maskedStrips.txt")
process.gemRecHits.deadFile = cms.FileInPath("RecoLocalMuon/GEMRecHit/data/deadStrips.txt")
```

> 
> > * I can remove #include "RecoLocalMuon/GEMRecHit/interface/GEMClusterizer.h" from GEMMaskReClusterizer.cc.
> > * The GEMClusterContainer.h doesn't include GEMCluster.h. So if I move it to GEMClusterizer.cc, the complier give an error. So I suggests to move #include "RecoLocalMuon/GEMRecHit/interface/GEMCluster.h" to GEMCluseterContainer.h.
> 
> I see... GEMCluseterContainer.h was designed in analogy with the ​RPCClusterContainer.h.
> I think that both them could/should be adjusted, but probably it is not for this PR to do it.

Then I will remove #include "RecoLocalMuon/GEMRecHit/interface/GEMClusterizer.h" from GEMMaskReClusterizer.cc and push again.

perrotta:
> To apply mask, you should add following line at the step3.
> 
> ```
> process.gemRecHits.applyMasking = cms.bool(True)
> process.gemRecHits.maskFile = cms.FileInPath("RecoLocalMuon/GEMRecHit/data/maskedStrips.txt")
> process.gemRecHits.deadFile = cms.FileInPath("RecoLocalMuon/GEMRecHit/data/deadStrips.txt")
> ```

Thank you. I ran succesfully several workflows with such a modified config, with no issues.
For me the PR is ready to go
 
> > > * I can remove #include "RecoLocalMuon/GEMRecHit/interface/GEMClusterizer.h" from GEMMaskReClusterizer.cc.

Just let me know if you actually plan to apply this trivial cleaning, so that we ca re-run the tests and I will sign for reco afterwards. Otherwise, it can be moved to a follow up PR

perrotta:
please test

perrotta:
+1
- Bug fixes does make sense
- They don't apply to the default config, where the applyMasking flag is disabled: tested that the code runs without errors when that flag is set to true
- Jenkins tests pass and show no differences, as expected for the default config

kpedro88:
+upgrade

fabiocos:
+1

PR #28459, opened 2019-11-22
Delay constructing transition wrappers

Dr15Jones:
#### PR description:

Do not create the edm::Run or edm::LuminosityBlock objects until they are requested.

#### PR validation:

The framework unit tests pass.

### DIFF
--- a/FWCore/Framework/interface/Event.h
+++ b/FWCore/Framework/interface/Event.h
@@ -33,6 +33,7 @@ For its usage, see "FWCore/Framework/interface/PrincipalGetAdapter.h"
 #include "FWCore/Common/interface/EventBase.h"
 #include "FWCore/Framework/interface/Frameworkfwd.h"
 #include "FWCore/Framework/interface/PrincipalGetAdapter.h"
+#include "FWCore/Framework/interface/LuminosityBlock.h"
 #include "FWCore/Utilities/interface/TypeID.h"
 #include "FWCore/Utilities/interface/EDGetToken.h"
 #include "FWCore/Utilities/interface/EDPutToken.h"
@@ -94,7 +95,12 @@ namespace edm {
     ///\return The id for the particular Stream processing the Event
     StreamID streamID() const { return streamID_; }
 
-    LuminosityBlock const& getLuminosityBlock() const { return *luminosityBlock_; }
+    LuminosityBlock const& getLuminosityBlock() const {
+      if (not luminosityBlock_) {
+        fillLuminosityBlock();
+      }
+      return *luminosityBlock_;
+    }
 
     Run const& getRun() const;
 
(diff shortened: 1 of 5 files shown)


Dr15Jones:
please test

Dr15Jones:
please test

Dr15Jones:
+1
The Reco difference is seen in other pull request.

fabiocos:
+1

slava77:
it looks like this PR added almost 10K issues in the static analysis report.
Should the mutables in the Event.h and LuminosityBlock.h be marked thread safe?
Combined they probably add up to most of the "mutable member if accessed via const pointer " kind of issues.

slava77:
> it looks like this PR added almost 10K issues in the static analysis report.
> Should the mutables in the Event.h and LuminosityBlock.h be marked thread safe?
> Combined they probably add up to most of the "mutable member if accessed via const pointer " kind of issues.

@Dr15Jones @smuzaffar  @makortel please comment if marking the mutables in the Event.h and LuminosityBlock.h thread safe is OK.

Dr15Jones:
@slava77 marking the change using CMS_THREAD_SAFE doesn't seem correct to me personally. That `mutable` is not safe to access across multiple threads, which is OK since `edm::Event` is itself only safe to use on one thread (each module actually get their own `edm::Event` instance).

So the proper thing would be able to mark `edm::Event` and `edm::LuminosityBlock` as being _thread friendly_ but not _const thread safe_. At the moment, we do not have a way to designate that.

slava77:
I see, is it practical to make these variables properly thread safe? (atomic or mutex)

Dr15Jones:
> I see, is it practical to make these variables properly thread safe? (atomic or mutex)

There are many `mutable` members inside `edm::Event`. Making them thread-safe would most likely cause performance issues and not give us any useful gain.

Dr15Jones:
@slava77 there are two static analysis checks for mutables. One (the test that you looked at) just looks for mutable in any classes and the second looks for thread-unsafe mutables in classes we know get used across threads (e.g. Event or EventSetup data products). To avoid the problem you pointed out we could

1. add a new C++ attribute which denotes that the mutable has been reviewed and found not to be _const thread safe_. This would only be applied to the general mutable checker, not to the data products checker.
1. we could remove the general mutable checker.

So the question to you is, do you find the general mutable finder helpful or is it just noise?

slava77:
> So the question to you is, do you find the general mutable finder helpful or is it just noise?

they are somewhat useful
- in some cases the variables do not need to be mutable with slight changes in the code
- in some cases there is a reasonably cheap/simple fix by using std::atomic 

The cases where a review was made and there will be no fix should not be counted/reported by default, assuming there is some annotation in the code to revisit these cases. This would be resolved by your case 1.

Dr15Jones:
@slava77 
Discussing the issue with Matti, we propose adding the attribute
```
[[cms::sa_allow]]
```
and the accompanying CPP define
```
CMS_SA_ALLOW
```
This can be used to tell the static analyzer (SA) to allow a construct it would previously have flagged as a problem. The policy would be that the use of the attribute also requires a comment stating why it should be allowed.

Matti and I feel this gives the most flexibility for the minimal change to the static analyzer as well as being easy to explain to developers.

PR #28463, opened 2019-11-25
HGCAL scintillator digitisation: adequate dynamic range (follow up 28349)

franzoni:
#### PR description:

This is a follow up to the [PR 28349](https://github.com/cms-sw/cmssw/pull/28349#issuecomment-554785930), which pointed out the necessity of extending the dynamic range for energy of the HGCAL scintillator. 
We do it by introducing the realistic digitisation also for the scintillators, even though not all parameters of the electronics and photo detection are currently known. By using the 10 bits of the adc (leaving the LSB un-changed) and the 12 bits for the amplitude measurement after the ToA switch, we are able to extend the saturation value to 1000 MIP (up from 275), which comfortably covers magnitude of physical deposits in the scintillator cells.

#### PR validation at digi level

HEback-Mod-sample

We verify that (for 100 ttbar events, no pu, same GEN-SIM for target-pr-> RED and reference->BLACK) with this PR the `mode` bit of the in-time bunch-crossing (3rd in a window of 5 samples) turns on for the hardest hits, 

HEback-DIGI-sample_2
and that the amplitude of the digi themselves is limited to below ~0.8*1024, the current coverage of the adc (the hits which exceed the threshold now use the toA way of recording the amplitude, which has a larger LSB and are therefore buried in the bulk of the distribution).
HEback-DIGI-sample_3

We verify also that in the bunch-crossing after the in-time sample (4th in a window of 5), more hits are present which stems from shape leakage of the realistic digitisation now used for the scintillators.
No changes in EE and HEfront.

#### PR validation at rechit level

HEback-RH-sample

The spectrum of HGCAL scintillator rechits in statistically unvaried for hits which don't hit the old saturation threshold. 

HEback-T-sample

The reconstructed time now happens for hits above the ToA threshold (previously no such thing was possible with the single-BX trivial digitisation). The ball-park values make sense and can enable studies, we don't aim at tuning the details since the actual parameters of the scintillator part are not known.


franzoni:
please test

SimCalorimetry/HGCalSimProducers/python/hgcalDigitizer_cfi.py
```diff
@@ -88,48 +90,8 @@
...
-            toaMode           = cms.uint32(1),
-            # TDC charge drain parameterisation (from J. Kaplon)
-            tdcChargeDrainParameterisation = cms.vdouble(
-                -919.13, 365.36, -14.10, 0.2,
-                 -21.85, 49.39,  22.21,  0.8,
-                 -0.28,   27.14,  43.95,
-                 3.89048 )
-            )
+        thresholdFollowsMIP = cms.bool(thresholdTracksMIP),
+        feCfg   = hgcROCSettings.clone()
```
rovere: Maybe it would be better to use (here and also elsewhere) a refToPset?
Something like

```
feCfg = cms.Pset(refToPset_ = cms.string("hgcROCSettings"))
```

should work. In this way, if we ever need to change parameters, e.g., via era upstream, they are properly propagated everywhere, irrespectively of the order in which files are loaded and imported.

slava77:
@franzoni 
please clarify if this should be in 11_0_X or if it's OK to have it later in 11_1_X.
Thank you.

franzoni:
greetings @slava77 ,
this PR ought to enter 11_ to the extent possible.

This

> This is a follow up to the PR 28349, which pointed out the necessity of extending the dynamic range 
> for energy of the HGCAL scintillator.

is the reason: saturation at 275 MIP is insufficient and known to generate un-physical saturations in the scintillators.

fabiocos:
@fwyzard @trtomei this PR is affecting digitization, so in principle it is needed in 11_0_X now if this feature is wanted for the HLT TDR. Of course it arrives quite late wrt to the schedule, please clarify whether it is desired (with the possible implied further delays) or not.

rovere:
Ciao @fabiocos, all, 
reading [here](https://twiki.cern.ch/twiki/bin/viewauth/CMS/CMSSW_11_0_0) I see that CMSSW_11 is supposed to be built in exactly 3 weeks' time. I know integration and validation are time-consuming, never-ending and often forgotten tasks, yet I think we could accommodate this PR for CMSSW_11.

SimCalorimetry/HGCalSimProducers/python/hgcalDigitizer_cfi.py
pfs: Hi, it makes sense to avoid overwriting the parameters inadvertedly. I did the exercise of changing this but there are several dependencies upstream (L1, RecHits) which are copying only specific parameters of this PSet (e.g. https://github.com/cms-sw/cmssw/blob/master/RecoLocalCalo/HGCalRecProducers/python/HGCalUncalibRecHit_cfi.py). As the refToPset_ is not resolved at python level we start raising errors related to unknown parameters. Of course one can also import directly the parameters from the factorized hgcROCSettings PSet, but that defeats the purpose of using refToPSet_. I would thus propose to leave this as it is for the moment, and check with you and @jbsauvan, the best way to structure passing the HGCROCSettings. This is rather important as in the plan for one of the near future PRs (not for this release) we want to pass also the gain information to have auto-calibrated recHits.

fwyzard:
> @fwyzard @trtomei this PR is affecting digitization, so in principle it is needed in 11_0_X now if this feature is wanted for the HLT TDR. 

I would say yes, at least based on the comments from @franzoni:
> saturation at 275 MIP is insufficient and known to generate un-physical saturations in the scintillators

SimCalorimetry/HGCalSimProducers/python/hgcalDigitizer_cfi.py
kpedro88: The correct way to synchronize parameters across multiple producers is to use an ESProducer (in lieu of an actual conditions database, which does not exist yet for HGCal). The `refToPSet_` option has always been something of a kludge for this case.

kpedro88:
+upgrade

RecoLocalCalo/HGCalRecProducers/python/HGCalUncalibRecHit_cfi.py
```diff
@@ -45,9 +45,16 @@
         ),
 
     HGCHEBConfig = cms.PSet(
-        isSiFE  = cms.bool(False),
+        #isSiFE  = cms.bool(False), # GF
```
slava77: is this commented out code needed?

SimCalorimetry/HGCalSimProducers/python/hgcalDigitizer_cfi.py
```diff
@@ -233,19 +155,18 @@
...
-            # n bits for the ADC (same as the silicon ROC)
-            adcNbits        = cms.uint32(12),
-            # ADC saturation : in this case we use the same variable but fC=MIP
-            adcSaturation_fC = cms.double(275.0), #value chosen to have 1MIP at 15ADC,
-            # together with the choice of adcNbits
-            # threshold for digi production : in this case we use the same variable but fC=MIP
-            adcThreshold_fC = cms.double(0.5),
-            )
+        feCfg = hgcROCSettings.clone(
+            adcNbits        = cms.uint32(10),      # standard ROC operations (was 2 bits more up to 11_0_0_pre12)
```
slava77: please drop type specifications for all parameters which already exist.

RecoLocalCalo/HGCalRecProducers/python/HGCalUncalibRecHit_cfi.py
franzoni: hello @slava77 , not needed, thanks for spotting.

SimCalorimetry/HGCalSimProducers/python/hgcalDigitizer_cfi.py
franzoni: addressed.

franzoni:
please test

slava77:
+1

for https://github.com/cms-sw/cmssw/pull/28463	47dc153 
- jenkins tests pass and comparisons with the baseline show differences only in the phase-2 workflows. Digis in HGCAL are affected, which throws off the random number sequence in other detectors. The final distributions are somewhat consistent. The DQM plots for HGCAL time/ToA show filled plots (the main feature change).

e.g. in 21234 (D44)
wf21234_hgcal_hescint_toa_9

fabiocos:
@kpedro88 given the technical nature of the latest updates, I will consider your previous signature as valid

kpedro88:
+upgrade

fabiocos:
@civanch any further comment?

jbsauvan:
Just to mention that since the HGCAL HSci digis are also used in the L1 TPG, I plan to transpose the exact same fix to the L1 sequence. I plan to make an other PR with these changes Today evening (CERN time).

fabiocos:
@jbsauvan do I correctly understand that this is needed to avoid inconsistencies?

jbsauvan:
The L1 TPG assumes the previous HCAL Sci digi format, so it will misinterpret the energy values in the TDC case.

civanch:
+1

fabiocos:
+1

PR #28422, opened 2019-11-19
Changes for new trained regression for UL 2016 - 110X

jainshilpi:
#### PR description:

These changes are essential for making the 2016 UL regression work. After discussion with Sam, I have made these changes (since 2016 uses the old PF cluster regression strategy and old ele/pho regressions codes). 
We had to create Modifier_run2_ECAL_2016_cff.py so that the training strategy of 2017 and 2018 could be used here as well.

PR #28466, opened 2019-11-25
Changes for new trained regression for UL 2016 - 110X 

jainshilpi:
***This is exactly the same as #28422. A new PR is created instead of continuing from there since the there were problems with the history and hence further merging***

These changes are essential for making the 2016 UL regression work. After discussion with Sam, I have made these changes (since 2016 uses the old PF cluster regression strategy and old ele/pho regressions codes).
We had to create Modifier_run2_ECAL_2016_cff.py so that the training strategy of 2017 and 2018 could be used here as well.

### DIFF
--- a/RecoParticleFlow/PFClusterProducer/python/particleFlowClusterOOTECAL_cff.py
+++ b/RecoParticleFlow/PFClusterProducer/python/particleFlowClusterOOTECAL_cff.py
@@ -6,7 +6,6 @@ particleFlowClusterOOTECAL.inputECAL = cms.InputTag("particleFlowClusterOOTECALU
 
 from Configuration.Eras.Modifier_run2_miniAOD_80XLegacy_cff import run2_miniAOD_80XLegacy
 run2_miniAOD_80XLegacy.toModify(
-    particleFlowClusterOOTECAL.energyCorrector, 
-    recHitsEBLabel = "reducedEcalRecHitsEB",
-    recHitsEELabel = "reducedEcalRecHitsEE"
+    particleFlowClusterOOTECAL,
+    energyCorrector = dict(applyMVACorrections = True, maxPtForMVAEvaluation = 90., srfAwareCorrection = False, recHitsEBLabel = "reducedEcalRecHitsEB", recHitsEELabel = "reducedEcalRecHitsEE")
 )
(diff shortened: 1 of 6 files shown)


slava77:
@cmsbuild please test

RecoParticleFlow/PFClusterProducer/python/particleFlowClusterOOTECAL_cff.py
```diff
@@ -6,7 +6,6 @@
 
 from Configuration.Eras.Modifier_run2_miniAOD_80XLegacy_cff import run2_miniAOD_80XLegacy
 run2_miniAOD_80XLegacy.toModify(
-    particleFlowClusterOOTECAL.energyCorrector, 
-    recHitsEBLabel = "reducedEcalRecHitsEB",
-    recHitsEELabel = "reducedEcalRecHitsEE"
+    particleFlowClusterOOTECAL,
+    energyCorrector = dict(applyMVACorrections = True, maxPtForMVAEvaluation = 90., srfAwareCorrection = False, recHitsEBLabel = "reducedEcalRecHitsEB", recHitsEELabel = "reducedEcalRecHitsEE")
```
slava77: please add linebreak for readability to keep the length within/below 100 chars.

Also, It looks like ```applyMVACorrections = True``` is redundant

jainshilpi: Hi Slava, yes that is right - removed applyMVACorrections = True.

slava77:
@cmsbuild please test

slava77:
@cmsbuild please test
it looks like the unit test failure is transient

slava77:
+1

for https://github.com/cms-sw/cmssw/pull/28466	857cfe1 
- jenkins tests pass and comparisons with the baseline show changes only in the 2016 workflows, starting from ECAL cluster corrected energies and then also propagating downstream with additional changes in the egamma objects
- a follow up update is expected to simplify the run2 configurations for the future releases as described in #28482 

Based on the discussion in the Nov 26 ORP, this PR is only introducing the configuration changes. There should be a follow up update to the GT to properly complete the update for the 2016 ecal/egamma regressions.

fabiocos:
+operations

the update of Eras and modifier is coherent with the purpose of the PR

fabiocos:
+1

PR #28467, opened 2019-11-25
Option to write out all ALCTs and CLCTs

dildick:
#### PR description:

Current CSC trigger simulation only allows for max 2 ALCTs and max 2 CLCTs per chamber. MC studies are ongoing that investigate possibilities to trigger on exotic signatures which produce >2 ALCTs or >2 CLCTs. Such signatures may include long-lived particles that decay to multiple muons.

This PR introduces an option in the CSC trigger to write out specific ALCT and CLCT collections with the label "All" for such exotica studies. These collections are only useful when `MAX_ALCTS_PER_PROCESSOR` and `MAX_CLCTS_PER_PROCESSOR` is greater than 2. By default the option is turned off, so there is no change in the trigger performance.

#### PR validation:

Tested it with a 2026 runTheMatrix command.

N/A

@tahuang1991

### DIFF
--- a/L1Trigger/CSCTriggerPrimitives/interface/CSCTriggerPrimitivesBuilder.h
+++ b/L1Trigger/CSCTriggerPrimitives/interface/CSCTriggerPrimitivesBuilder.h
@@ -71,6 +71,7 @@ public:
              const GEMPadDigiClusterCollection* gemPadClusters,
              CSCALCTDigiCollection& oc_alct,
              CSCCLCTDigiCollection& oc_clct,
+             CSCCLCTDigiCollection& oc_clct_all,
              CSCALCTPreTriggerDigiCollection& oc_alctpretrigger,
              CSCCLCTPreTriggerDigiCollection& oc_clctpretrigger,
              CSCCLCTPreTriggerCollection& oc_pretrig,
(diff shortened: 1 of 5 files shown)


dildick:
I fixed this in my last commit.

dildick:
@fabiocos Is the cmsbuild state machine stuck in this PR?

fabiocos:
code-checks

fabiocos:
please test

fabiocos:
please test

fabiocos:
@rekovic could you please have a look?

@dildick this PR in its default configuration does not modify anything. What is the expected increase in output size in case the flags are switched to true?

dildick:
That depends on the scenario. In low-pileup scenarios the "All" CLCT and ALCT collections will be just copies of the standard CLCT and ALCT collections. At PU200, there could be many more CLCTs and ALCTs per chamber... So the size could increase with a factor 2 to perhaps even 5. But in any case, the flags `writeOutAllCLCTs` and `writeOutAllALCTs` should not be set to True for any official MC production.

rekovic:
+1

fabiocos:
+1

PR #28468, opened 2019-11-25
HFNose: add 2D clusters

mariadalfonso:
This PR add 2D clusters for the HFnose


perrotta:
please test

RecoLocalCalo/HGCalRecProducers/interface/HGCalCLUEAlgo.h
```diff
@@ -19,16 +19,19 @@
 
 #include "RecoLocalCalo/HGCalRecAlgos/interface/RecHitTools.h"
 
+#include "tbb/task_arena.h"
```
kpedro88: Why are these includes added here when they weren't needed before?

RecoLocalCalo/HGCalRecProducers/interface/HFNoseTilesConstants.h
```diff
@@ -0,0 +1,38 @@
...
+
+#ifndef RecoLocalCalo_HGCalRecAlgos_interface_HFNoseTilesConstants_h
+#define RecoLocalCalo_HGCalRecAlgos_interface_HFNoseTilesConstants_h
+
+#include <cstdint>
+#include <array>
+
+namespace hfnosetilesconstants {
+
+  constexpr int32_t ceil(float num) {
```
kpedro88: this function should be put in some central area rather than copy-pasting it

RecoLocalCalo/HGCalRecProducers/python/hgcalLayerClusters_cff.py
```diff
@@ -17,3 +17,15 @@
 hgcalLayerClusters.plugin.fcPerEle = cms.double(fC_per_ele)
 hgcalLayerClusters.plugin.noises = cms.PSet(refToPSet_ = cms.string('HGCAL_noises'))
 hgcalLayerClusters.plugin.noiseMip = hgchebackDigitizer.digiCfg.noise
+
+
+hgcalLayerClustersHFNose = hgcalLayerClusters_.clone()
+
+hgcalLayerClustersHFNose.detector = cms.string('HFNose')
```
kpedro88: make all of these changes inside the `clone()` call

RecoLocalCalo/HGCalRecProducers/interface/HGCalCLUEAlgo.h
rovere: This is a relic of my previous template implementation. It can be safely moved to the .cc where it was.

rovere: Which already has it, in fact. Just drop it.

slava77:
@mariadalfonso 
please provide some estimates for disk size and CPU time needed to reconstruct the new clusters

mariadalfonso:
@slava77 ok I will check these.
rule of thumb: HFNose is a 10% the size of HGCAL

kpedro88:
please test

mariadalfonso:
@slava77 
on wf 21234.0_TTbar_14TeV here what i see

```
FastReport   CPU time avg.      when run  Real time avg.      when run     Alloc. avg.      when run   Dealloc. avg.      when run  Modules
FastReport        26.1 ms        26.1 ms        26.1 ms        26.1 ms       +6963 kB       +6963 kB       -1476 kB       -1476 kB    hgcalLayerClusters
FastReport        14.9 ms        14.9 ms        15.0 ms        15.0 ms       +4791 kB       +4791 kB        -950 kB        -950 kB    hgcalLayerClustersHFNose
```

```
recoCaloClusters_hgcalLayerClustersHFNose__RECO. 10685.2 2133.76
recoCaloClusters_hgcalLayerClusters__RECO. 58.14 47.48
```

slava77:
>  on wf 21234.0_TTbar_14TeV here what i see

I'm not sure that no-pileup results are informative; 21434.0 should be more appropriate.

hgcalLayerClusters look empty, which looks suspicious; having no other better reference, what is the RECO file size? (this may require changing the ```--eventcontent``` to RECOSIM in the cmsDriver coming out from the matrix).

slava77:
I got stuck in the review because the solution proposed here is not particularly trivial; it is in some parts using some creative c++ template specialization.

Is there a good evidence that the templated code is required?
I can only think of some limited performance benefits from knowing some ranges or values at compile time to optimize the compiled code. But then looking at the implementation, it looks like these benefits should be really minor.

It looks like there is a clear need to run the same code on data structures different only in the array size dimension. The amount of code at .cc level is about 1K lines with quite a bit of stuff inlined, which means that the replication coming from the two templates doubles the fairly large instruction base. A common approach to support different sizes is to use containers with size defined at runtime. If the ```HGCalLayerTiles::tiles_``` member is changed to be a vector instead of an array, the same compiled code can be used for both HGCal and HFNose.

If there is no clear evidence for better performance, I propose to change to a non-templated implementation.
- define a const TileConstants struct and two (static) const instances: one for HGCal and one for HFNose
- HGCalLayerTiles can get the parameters by holding a pointer or a copy to the relevant TileConstants instance defined at construction
- HGCalCLUEAlgo picks up HGCal or HFNose variant based on the configuration

rovere:
@slava77 
I'm not sure I share your review, sorry.

   * The move from using a dynamic vector to the fixed-size array was suggested by you during the review of #26295.
   * I did not measure the gain, but anything different will run slower. The static_asserts used pretty much everywhere require those numbers to be known at compile time. Anything else, including the _const_ you suggest, will run slower.
   * The template, for the moment, is such that the implementation is identical but gives us the freedom to chose, in future, different solutions w/o coupling (too much) the implementation (except for the underlying interface, of course).
   * `pluginRecoLocalCaloHGCalRecProducersPlugins.so` with this templated solution and compiled with both implementations is less than 700K. I hardly believe this is a problem for CMS/CMSSW.
   * I'm not willing to spend my time modifying code to support `HFNose`. I'm glad and willing to help as I did if this requires a reasonable amount of work and almost no impact on `HGCal` and `TICL` related code. If that's not the case, I suggest a complete copy&paste of the code base, eventually with dedicated and separate packages in CMSSW.

slava77:
>  I'm not sure I share your review, sorry.

At the point of proposing the non-templated way, I thought that I have too many comments which would require more work on the templated implementation compared to what may be a more straightforward non-templated implementation. I can continue with the review on the templated variant for you to perhaps reassess later what is easier.

> * The move from using a dynamic vector to the fixed-size array was suggested by you during the review of #26295.

this appears to refer to a different algorithm (TICL). 
Does this comment imply that the downstream TICL algorithms will also all be templated for HFNose? I hope not.

> * I did not measure the gain, but anything different will run slower. The static_asserts used pretty much everywhere require those numbers to be known at compile time. Anything else, including the const you suggest, will run slower.

Related to the static_asserts, thanks for noticing. They can be eliminated by adding rX, rY, rEta, and rPhi constants in the TileConstants. So, there is no extra cost here.

> * The template, for the moment, is such that the implementation is identical but gives us the freedom to chose, in future, different solutions w/o coupling (too much) the implementation (except for the underlying interface, of course).

OK, this is a valid reason for HGCalCLUEAlgo, except that the same can be achieved with inheritance.

> * pluginRecoLocalCaloHGCalRecProducersPlugins.so with this templated solution and compiled with both implementations is less than 700K. I hardly believe this is a problem for CMS/CMSSW.

this is not a problem for cvmfs, this is a problem for the executable instruction set. 700K is huge compared to the L1 instruction cache. The increased size will slow down the execution, and this may be more costly than the benefits from knowing const values at compile time.

> * I'm not willing to spend my time modifying code to support HFNose.

Roughly, I see rather similar or even smaller set of changes necessary compared with the current version. Perhaps your comment means "to spend any more of my time". If we discussed this in advance of the developments, this may have had a different conclusion.

mariadalfonso:
> > on wf 21234.0_TTbar_14TeV here what i see
> 
> I'm not sure that no-pileup results are informative; 21434.0 should be more appropriate.
> 
> hgcalLayerClusters look empty, which looks suspicious; having no other better reference, what is the RECO file size? (this may require changing the `--eventcontent` to RECOSIM in the cmsDriver coming out from the matrix).

```
FastReport   CPU time avg.      when run  Real time avg.      when run     Alloc. avg.      when run   Dealloc. avg.      when run  Modules
FastReport        36.0 ms        36.0 ms        37.3 ms        37.3 ms       +9081 kB       +9081 kB       -3843 kB       -3843 kB    hgcalLayerClusters
FastReport        10.1 ms        10.1 ms        10.1 ms        10.1 ms       +2436 kB       +2436 kB       -1253 kB       -1253 kB    hgcalLayerClustersHFNose
```

```
recoCaloClusters_hgcalLayerClusters__RECO. 651954 120156
recoCaloClusters_hgcalLayerClustersHFNose__RECO. 93394 18423.7
```

P.S. disregard the previous numbers, were on a singlePhoton no PU at eta 3.5

slava77:
> These are the results on 21234.0_TTbar_14TeV

just to make sure; are these still no-PU?
Please let me know if you can check the case with PU, e.g. 21434.
Thank you.

rovere:
> > I'm not sure I share your review, sorry.
> 
> At the point of proposing the non-templated way, I thought that I have too many comments which would require more work on the templated implementation compared to what may be a more straightforward non-templated implementation. I can continue with the review on the templated variant for you to perhaps reassess later what is easier.

The comments you make are suggestions, not requirements. If you intend them as requirements you can modify directly the code in the first place. Maybe we have to clarify the review procedure at a higher level.

> 
> > * The move from using a dynamic vector to the fixed-size array was suggested by you during the review of #26295.
> 
> this appears to refer to a different algorithm (TICL).
> Does this comment imply that the downstream TICL algorithms will also all be templated for HFNose? I hope not.

Not the algorithms, just the very same data structure.

> 
> > * I did not measure the gain, but anything different will run slower. The static_asserts used pretty much everywhere require those numbers to be known at compile time. Anything else, including the const you suggest, will run slower.
> 
> Related to the static_asserts, thanks for noticing. They can be eliminated by adding rX, rY, rEta, and rPhi constants in the TileConstants. So, there is no extra cost here.
> 
> > * The template, for the moment, is such that the implementation is identical but gives us the freedom to chose, in future, different solutions w/o coupling (too much) the implementation (except for the underlying interface, of course).
> 
> OK, this is a valid reason for HGCalCLUEAlgo, except that the same can be achieved with inheritance.

Yes, with virtual calls inside loops on 10^5 elements, that would be for sure faster...

> 
> > * pluginRecoLocalCaloHGCalRecProducersPlugins.so with this templated solution and compiled with both implementations is less than 700K. I hardly believe this is a problem for CMS/CMSSW.
> 
> this is not a problem for cvmfs, this is a problem for the executable instruction set. 700K is huge compared to the L1 instruction cache. The increased size will slow down the execution, and this may be more costly than the benefits from knowing const values at compile time.

I'm not sure this will have a real impact. And, just to make the point clear, you are discussing about L1 cache for a piece of code that runs in tens of ms. Do we really need to optimize this kind of already fast code at this level?

> 
> > * I'm not willing to spend my time modifying code to support HFNose.
> 
> Roughly, I see rather similar or even smaller set of changes necessary compared with the current version. Perhaps your comment means "to spend any more of my time". If we discussed this in advance of the developments, this may have had a different conclusion.

In any case, despite the code "complexity", I do not see any concrete measure that the proposed solution is not good enough for production.

slava77:
> > I can continue with the review on the templated variant for you to perhaps reassess later what is easier.
> 
> The comments you make are suggestions, not requirements.

OK, I will continue with the review path assuming the templated version is really strongly desired version.

slava77:
> > Does this comment imply that the downstream TICL algorithms will also all be templated for HFNose? I hope not.
> 
> Not the algorithms, just the very same data structure.

Sorry, the response was too terse for me.
Please clarify/confirm if a similar (or any) templating is expected for the needs of the HFNose also in the TICL algorithms.
Thank you.

rovere:
> > > Does this comment imply that the downstream TICL algorithms will also all be templated for HFNose? I hope not.
> > 
> > 
> > Not the algorithms, just the very same data structure.
> 
> Sorry, the response was too terse for me.
> Please clarify/confirm if a similar (or any) templating is expected for the needs of the HFNose also in the TICL algorithms.
> Thank you.

Ciao Slava,
honestly, I'm not in a position to comment on that, at least for the current status of the reconstruction.

Having said that, I think that we need at least another templating [here](https://github.com/cms-sw/cmssw/blob/master/DataFormats/HGCalReco/interface/TICLLayerTile.h), more or less along the lines of what I did in this PR. This assuming that also in the case of `HFNose` we will use the `TICLLayerTileProducer` (see [here](https://github.com/cms-sw/cmssw/blob/master/RecoHGCal/TICL/plugins/TICLLayerTileProducer.cc)).
For an example of what I mean, please look [here](https://github.com/rovere/cmssw/blob/TICL_inReco/DataFormats/HGCalReco/interface/TICLLayerTile.h#L49-L70). Or maybe even a level up, since eta-phi ranges will need to be changed for `HFNose`.

If we then make the further assumption that we want to re-use also the PatternRecognition in its current incarnation also for the `HFNose`, I fear something else needs to be touched [here](https://github.com/cms-sw/cmssw/blob/master/RecoHGCal/TICL/plugins/PatternRecognitionAlgoBase.h#L30-L37), at the very least, with changes propagated downstream.

At the current level of development for the `HGCal` reconstruction, I'm not sure how deeply I want the reconstruction and data-structure themselves coupled between `HGCal` and `HFNose`. Because we have not yet any real measure of the goodness of the full reconstruction in `HGCal` and, even more, no evidence that the very same approach/algorithms will transparently work elsewhere.

slava77:
> honestly, I'm not in a position to comment on that, at least for the current status of the reconstruction.
> 
> Having said that, I think that we need at least another templating ...

Thank you for the quick notes or estimates on how far this possibly spreads.
This apparently also reveals that with a similar strategy my idea/question in the last reco meeting to potentially use TICL for other detectors quickly falls apart due to such a strong compile time coupling to the geometry of the detectors.

Another related question is how stable are the geometry parameters which cast in stone (constexpr'ed) with the current implementation?
Considering -- perhaps quite unlikely for HGCAL -- experience with "plan-1" in HE, even the phi min/max is not necessarily fixed (well, +/-pi will always cover, although it may be less computationally efficient. ).
If there are benefits to sub-split by layers, then the x,y ranges change.
I'm guessing in LS3 or similar no-beam times, we may need to run with the endcaps open: is the hardcoded eta range safe against it? [I'm assuming that a few cm misalignment in x/y/z can be anyways tolerated]

Related to both TICL and CLUE tiling: is it still computationally efficient to define the array sizes that are just large enough to cover all reasonable needs ?
Before your message I was thinking/assuming that TICL is somewhat generic enough to handle different inputs.

rovere:
Ciao Slava,
we can make everything parametric, at the price of computing ranges here and there. Unless we fill code with const static. In this PR I think we can see that the CLUE algorithm can support both `HGCAL` and `HFNose` in a, more or less, transparent way without code-duplication. I think this could be extended to TICL in general.
The underlying data-structures have been optimized with `HGCal` in mind. The eta range could be extended to accommodate some tolerance, while the grid dimension has to be tailored to the specific detector and its 3D space location. The partitioning along the "layer-dimension" uses the layer number as an integer, not its Z coordinate. The further eta-phi space remains unchanged (and in fact we already have redundancy). x, y and z are accessed through the geometry, dynamically. `TICL` uses `Tiles` internally and, to the extent that those could be `made generic enough` (via templating is an example), we can support different scenarios.

slava77:
> And, just to make the point clear, you are discussing about L1 cache for a piece of code that runs in tens of ms. Do we really need to optimize this kind of already fast code at this level?

It looks like the total is more like many hundreds of ms: hgcalLayerClusters is 683.2 ms in your recent test
https://rovere.web.cern.ch/rovere/HGCAL/Timing/circles/web/piechart.html?dataset=ttbar_11_0_X_2019-11-18-1100_200PURelVal_TICL_inReco.
It's under 1% here; so it's still small and not changing the arguments quantitatively, I guess.

rovere:
> > And, just to make the point clear, you are discussing about L1 cache for a piece of code that runs in tens of ms. Do we really need to optimize this kind of already fast code at this level?
> 
> It looks like the total is more like many hundreds of ms: hgcalLayerClusters is 683.2 ms in your recent test
> https://rovere.web.cern.ch/rovere/HGCAL/Timing/circles/web/piechart.html?dataset=ttbar_11_0_X_2019-11-18-1100_200PURelVal_TICL_inReco.
> It's under 1% here; so it's still small and not changing the arguments quantitatively, I guess.

Indeed looking at the pie I'm even more convinced that you are chasing optimizations, especially at this level, in the wrong place.

slava77:
> Indeed looking at the pie I'm even more convinced that you are chasing optimizations, especially at this level, in the wrong place.

right; well, in the HLT context timing goals it's already a pretty large number comparable to the total budget per event; but in this context we also have imagination at play that everything will run infinitely faster on future hardware.

rovere:
In fact, it happens that we already have a GPU implementation of the CLUE algorithm that can run in ~30ms, more or less. That's not future software on future hardware, it's current algorithm on exiting GPUs.

FYI: [slides presented at CHEP](https://indico.cern.ch/event/773049/contributions/3473275/attachments/1936902/3210151/chep2019.pdf)

slava77:
> In fact, it happens that we already have a GPU implementation of the CLUE algorithm that can run in ~30ms, more or less. That's not future software on future hardware, it's current algorithm on exiting GPUs.

oh, very nice. Thank you for sharing.
On p13 I was not able to match to the 600 ms from the pie chart measured on a CPU. My guess is that there it corresponds to the orange 200 ms. Is the difference just from the faster hardware?
Is this  number normalized to an equivalent single CPU core cost used in the projections?
My naive match for one GPU costing as one cheaper CPU socket with ~20 HT-weighted threads in a socket (or 1/2 of a more expensive more core socket), the 30 ms on one GPU at full load is still fairly far from the target budget.
But this is digressing from this PR review.

mariadalfonso:
@slava77 
What is the status of this review ? Do you have some suggested change ? 
Ideally I would like to complete this before the winter break

slava77:
>  What is the status of this review ? Do you have some suggested change ?

I'm hoping to get back to you by the end of the week.
Sorry for the delay.

slava77 (commented):
this is from last week, but it still looks consistent/practical.
Please start looking at this part while I continue with the rest of the PR/

RecoLocalCalo/HGCalRecProducers/interface/HGCalTilesConstants.h
```diff
@@ -14,23 +14,24 @@ namespace hgcaltilesconstants {
...
-  constexpr int nRowsPhi = hgcaltilesconstants::ceil((maxPhi - minPhi) / tileSizeEtaPhi);
-  constexpr int nTiles = nColumns * nRows + nColumnsEta * nRowsPhi;
-
+  struct TileConstants {
+    static constexpr float tileSize = 5.f;
+    static constexpr float minX = -285.f;
+    static constexpr float maxX = 285.f;
+    static constexpr float minY = -285.f;
+    static constexpr float maxY = 285.f;
+    static constexpr int nColumns = hgcaltilesconstants::ceil((maxX - minX) / tileSize);
```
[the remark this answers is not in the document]
mariadalfonso: done

RecoLocalCalo/HGCalRecProducers/interface/HGCalTilesConstants.h
```diff
@@ -14,23 +14,24 @@ namespace hgcaltilesconstants {
                                                                   : static_cast<int32_t>(num) + ((num > 0) ? 1 : 0);
```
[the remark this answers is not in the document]
mariadalfonso: done

RecoLocalCalo/HGCalRecProducers/interface/HFNoseTilesConstants.h
```diff
@@ -0,0 +1,34 @@
...
+
+namespace hfnosetilesconstants {
+  struct TileConstants {
+    // inner radius ~ 3cm , other radius ~ 110
+    static constexpr float tileSize = 5.f;
+    static constexpr float minX = -110.f;
+    static constexpr float maxX = 110.f;
+    static constexpr float minY = -110.f;
+    static constexpr float maxY = 110.f;
+    static constexpr int nColumns = hgcaltilesconstants::ceil((maxX - minX) / tileSize);
```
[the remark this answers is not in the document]
mariadalfonso: done

RecoLocalCalo/HGCalRecProducers/interface/HGCalLayerTiles.h
```diff
@@ -5,14 +5,16 @@
...
 
 #include <vector>
 #include <array>
 #include <cmath>
 #include <algorithm>
 #include <cassert>
 
-class HGCalLayerTiles {
+template <typename T>
+class LayerTilesT {
```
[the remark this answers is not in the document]
mariadalfonso: done

RecoLocalCalo/HGCalRecProducers/interface/HFNoseTilesConstants.h
```diff
@@ -0,0 +1,38 @@
...
+#include <array>
+
+namespace hfnosetilesconstants {
+
+  constexpr int32_t ceil(float num) {
+    return (static_cast<float>(static_cast<int32_t>(num)) == num) ? static_cast<int32_t>(num)
+                                                                  : static_cast<int32_t>(num) + ((num > 0) ? 1 : 0);
+  }
+
+  struct TileConstants {
```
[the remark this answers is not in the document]
mariadalfonso: used struct instead of nampespace

RecoLocalCalo/HGCalRecProducers/interface/HFNoseTilesConstants.h
```diff
@@ -0,0 +1,38 @@
+// Authors: Felice Pantaleo - felice.pantaleo@cern.ch
+// Date: 03/2019
+
+#ifndef RecoLocalCalo_HGCalRecAlgos_interface_HFNoseTilesConstants_h
+#define RecoLocalCalo_HGCalRecAlgos_interface_HFNoseTilesConstants_h
```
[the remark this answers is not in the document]
mariadalfonso: ok

RecoLocalCalo/HGCalRecProducers/interface/HGCalTilesConstants.h
```diff
@@ -14,23 +14,24 @@ namespace hgcaltilesconstants {
...
-  constexpr float minEta = -3.f;
-  constexpr float maxEta = 3.f;
-  //To properly construct search box for cells in phi=[-3.15,-3.] and [3.,3.15], cells in phi=[3.,3.15] are copied to the first bin and cells in phi=[-3.15,-3.] are copied to the last bin
-  constexpr float minPhi = -3.3f;
-  constexpr float maxPhi = 3.3f;
-  constexpr int nColumnsEta = hgcaltilesconstants::ceil((maxEta - minEta) / tileSizeEtaPhi);
-  constexpr int nRowsPhi = hgcaltilesconstants::ceil((maxPhi - minPhi) / tileSizeEtaPhi);
-  constexpr int nTiles = nColumns * nRows + nColumnsEta * nRowsPhi;
-
+  struct TileConstants {
```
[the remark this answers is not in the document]
mariadalfonso: used struct instead of namespace

mariadalfonso:
@rovere , check the changes as well

slava77:
@cmsbuild please test

slava77:
So, the 
I wonder ```constexpr int a = std::ceil( ...``` did not work in clang; sigh.

Is the gcc not compliant or is it the clang ?
@Dr15Jones @makortel

Dr15Jones:
Non of the standard web reference I use say `std::ceil` is declared `constexpr` in the standard.

slava77:
OK, we need to put back a method.

I propose to put it in a common place e.g. DataFormats/Math/interface/constexpr_cmath.h
With lack of imagination, ```constexpr_ceil``` could be a good name.

slava77:
@cmsbuild please test

DataFormats/Math/interface/constexpr_cmath.h
```diff
@@ -0,0 +1,10 @@
+#ifndef DataFormats_Math_constexpr_cmath_h
+#define DataFormats_Math_constexpr_cmath_h
+
+namespace reco {
+  constexpr int32_t ceil(float num) {
```
slava77: the header consistency check is complaining about int32_t ; which means this file needs an include of ```<cstdint>```

mariadalfonso: done

slava77:
@cmsbuild please test

kpedro88:
@slava77 do you have further comments here?
@mariadalfonso can you confirm the latest version of the PR has the same technical performance as previously studied? Some initial validation of the physics performance, confirming the clusters come out as expected, would also be appreciated.

slava77:
> @slava77 do you have further comments here?

unfortunately, the review is still pending (distracted by other higher priority things).

I'm looking back at earlier comments and I'm still missing a timing and size estimate for PU200
https://github.com/cms-sw/cmssw/pull/28468#issuecomment-567094637

RecoLocalCalo/HGCalRecProducers/plugins/HGCalLayerClusterProducer.cc
```diff
@@ -63,7 +64,10 @@ HGCalLayerClusterProducer::HGCalLayerClusterProducer(const edm::ParameterSet& ps
       detector(ps.getParameter<std::string>("detector")),  // one of EE, FH, BH or "all"
       timeClname(ps.getParameter<std::string>("timeClname")),
       timeOffset(ps.getParameter<double>("timeOffset")) {
-  if (detector == "all") {
+  if (detector == "HFNose") {
+    hits_hfnose_token = consumes<HGCRecHitCollection>(ps.getParameter<edm::InputTag>("HFNoseInput"));
+    algoId = reco::CaloCluster::hfnose;
+  } else if (detector == "all") {
```
slava77: it looks like "all" nebulously means everything but HFNose. 
Some more descriptive name would be useful.
E.g. ```HGCalAll".
Minimally, please add a comment that "all" has a special meaning

RecoLocalCalo/HGCalRecProducers/plugins/HGCalLayerClusterProducer.cc
```diff
@@ -101,6 +110,7 @@ void HGCalLayerClusterProducer::fillDescriptions(edm::ConfigurationDescriptions&
   desc.add<edm::ParameterSetDescription>("plugin", pluginDesc);
   desc.add<std::string>("detector", "all");
```
slava77: please add a comment here to clarify that all does not include HFNose and possibly list all other variants
in ```->setComment("all (does not include HFNose); other options: EE, FH, HFNose; other value defaults to BH");```

RecoLocalCalo/HGCalRecProducers/python/hgcalLayerClusters_cff.py
```diff
@@ -17,3 +17,17 @@
...
 hgcalLayerClusters.plugin.noises = cms.PSet(refToPSet_ = cms.string('HGCAL_noises'))
 hgcalLayerClusters.plugin.noiseMip = hgchebackDigitizer.digiCfg.noise
+
+
+hgcalLayerClustersHFNose = hgcalLayerClusters_.clone(
+    detector = cms.string('HFNose'),
+    timeOffset = hfnoseDigitizer.tofDelay,
+    plugin = cms.PSet(
+        dEdXweights = cms.vdouble(dEdX.weightsNose),
+        fcPerMip = cms.vdouble(HGCalUncalibRecHit.HGCHFNoseConfig.fCPerMIP),
```
slava77: type specs for all existing parameters better be dropped for automated syntax checking.

```python
detector = 'HFNose',
plugin = dict(
  dEdXweights = dEdX.weightsNose,
  ...
```

slava77 (commented):
sorry for the delay with the review; in retrospect, this could've been finished last week

slava77:
some plots from 22434.0 on 200 events, apparently make sense

There are about 400 clusters per event
all_origVSsign1094_TTbar14TeV2026D47wf22434p0c_recoCaloClusters_hgcalLayerClustersHFNose__RECO_objAT_size

all_origVSsign1094_TTbar14TeV2026D47wf22434p0c_recoCaloClusters_hgcalLayerClustersHFNose__RECO_obj_eta


layerCluster time (non-default values)
all_origVSsign1094_TTbar14TeV2026D47wf22434p0c_floatedmValueMap_hgcalLayerClustersHFNose_timeLayerCluster_RECO_obj_values_39

energy
all_origVSsign1094_TTbar14TeV2026D47wf22434p0c_log10recoCaloClusters_hgcalLayerClustersHFNose__RECO_obj_energy

Martin-Grunewald:
-1
please rebase properly to target branch, rather than merging

perrotta:
please test

mariadalfonso:
@slava77 

Updated the branch with your latest comments.

I did not merge the HGCalCLUEAlgo.cc with the HGCalCLUEAlgoT.h , neither created HGCalCLUEAlgoBase as in , would be better if @rovere will reimplement this in a follow up PR
https://github.com/cms-sw/cmssw/pull/28468#discussion_r359528577
https://github.com/cms-sw/cmssw/pull/28468#discussion_r359533323
https://github.com/cms-sw/cmssw/pull/28468#discussion_r359539337


on wf  22434.0 I see

```
FastReport   CPU time avg.      when run  Real time avg.      when run     Alloc. avg.      when run   Dealloc. avg.      when run  Modules
FastReport        39.6 ms        39.6 ms        39.6 ms        39.6 ms      +10859 kB      +10859 kB       -4749 kB       -4749 kB    hgcalLayerClusters
FastReport         8.6 ms         8.6 ms         8.6 ms         8.6 ms       +2200 kB       +2200 kB       -1178 kB       -1178 kB    hgcalLayerClustersHFNose
```

```
recoCaloClusters_hgcalLayerClusters__RECO. 708535 132382
recoCaloClusters_hgcalLayerClustersHFNose__RECO. 68812 13769.1
```

slava77 (commented):
related to still unresolved https://github.com/cms-sw/cmssw/pull/28468#discussion_r359528577

I couldn't find examples of this as a recommended pattern online, but perhaps there is some well-based intent to do it this way instead of a "classic" way of having the template code implementation in the header files.

If there is a concern with moving quite a bit of source code lines, I can suggest to rename HGCalCLUEAlgo.cc to HGCalCLUEAlgoT.icc and include it at the end of HGCalCLUEAlgoT.h (before the template instantiation "using" lines).

kpedro88:
@slava77 regarding .icc files, indeed they are sometimes the best option, and they are used in some places in CMSSW. Here is an example from the framework, set up in such a way that only the .h file is needed externally:
https://github.com/cms-sw/cmssw/blob/master/FWCore/MessageLogger/interface/ErrorObj.h
https://github.com/cms-sw/cmssw/blob/master/FWCore/MessageLogger/interface/ErrorObj.icc

mariadalfonso:
@slava77 

After checking with @rovere I implemented solution discussed here
https://github.com/cms-sw/cmssw/pull/28468#pullrequestreview-338789223

let me know if is all in order

fwyzard:
> @slava77 regarding .icc files, indeed they are sometimes the best option, and they are used in some places in CMSSW. Here is an example from the framework, set up in such a way that only the .h file is needed externally:
> https://github.com/cms-sw/cmssw/blob/master/FWCore/MessageLogger/interface/ErrorObj.h
> https://github.com/cms-sw/cmssw/blob/master/FWCore/MessageLogger/interface/ErrorObj.icc

This approach still ends up inluding and compiling the .h/.icc files in each compilation unit.

If the possible types are known, I think a better alternative would be to use explicit template instantiation, with the declaration in the .h and definition a the .cc file.

slava77:
> This approach still ends up including and compiling the .h/.icc files in each compilation unit.

technically yes, but this code is supposed to be included/compiled only in one place.

rovere:
> > @slava77 regarding .icc files, indeed they are sometimes the best option, and they are used in some places in CMSSW. Here is an example from the framework, set up in such a way that only the .h file is needed externally: ...
> 
> This approach still ends up including and compiling the .h/.icc files in each compilation unit.
> 
> If the possible types are known, I think a better alternative would be to use explicit template instantiation, with the declaration in the .h and definition a the .cc file.

@fwyzard that was what has been proposed in the first place. It is also true that, at least at present, there's only 1 compilation unit and 2 know types.

fwyzard:
> @fwyzard that was what has been proposed in the first place.

ok, and the rationale for *not* doing it was ?

RecoLocalCalo/HGCalRecProducers/plugins/HGCalCLUEAlgoT.h
```diff
@@ -196,14 +197,19 @@ class HGCalCLUEAlgo : public HGCalClusteringAlgoBase {
...
                              float delta_c,
                              float delta_r);  // return max density
-  void calculateDistanceToHigher(const HGCalLayerTiles& lt, const unsigned int layerId, float delta_c, float delta_r);
+  void calculateDistanceToHigher(const TILE& lt, const unsigned int layerId, float delta_c, float delta_r);
   int findAndAssignClusters(const unsigned int layerId, float delta_c, float delta_r);
   math::XYZPoint calculatePosition(const std::vector<int>& v, const unsigned int layerId) const;
   void setDensity(const unsigned int layerId);
 };
 
+#include "RecoLocalCalo/HGCalRecProducers/plugins/HGCalCLUEAlgo.icc"
```
fwyzard:
```suggestion
// declaration of explicit template instantiation
extern template class HGCalCLUEAlgoT<HGCalLayerTiles>;
extern template class HGCalCLUEAlgoT<HFNoseLayerTiles>;
```

RecoLocalCalo/HGCalRecProducers/plugins/HGCalCLUEAlgo.icc
```diff
@@ -534,11 +541,15 @@ void HGCalCLUEAlgo::computeThreshold() {
...
   unsigned int numberOfCells = cellsOnLayer.detid.size();
   for (unsigned int i = 0; i < numberOfCells; ++i)
     density_[cellsOnLayer.detid[i]] = cellsOnLayer.rho[i];
 }
 
-Density HGCalCLUEAlgo::getDensity() { return density_; }
+template <typename T>
+Density HGCalCLUEAlgoT<T>::getDensity() {
+  return density_;
+}
```
fwyzard:
```suggestion
}

// definition of explicit template instantiation
template class HGCalCLUEAlgoT<HGCalLayerTiles>;
template class HGCalCLUEAlgoT<HFNoseLayerTiles>;
```

fwyzard (changes requested):
Here is how things would look like using explicit template instantiation.

It would also make sense to rename the files to their old names, then.

rovere:
> > @fwyzard that was what has been proposed in the first place.
> 
> ok, and the rationale for _not_ doing it was ?

The reviewer disliked it and required alternative approaches.

RecoLocalCalo/HGCalRecProducers/plugins/HGCalCLUEAlgo.icc
slava77: does this remove the need for explicit specialization of every method as it was before?
If so, I think that this is a good solution.

fwyzard: I think so - it does compile.

slava77:
nominally, the factory pattern would imply that only HGCalLayerClusterPluginFactory.cc is supposed to know (be able to include the headers of) the types. So, only one compilation unit and only HGCalLayerClusterPluginFactory.cc to have the algo template specialized.

Still, the last solution proposed in https://github.com/cms-sw/cmssw/pull/28468#pullrequestreview-340523998 looks fine for me as well

rovere:
> nominally, the factory pattern would imply that only HGCalLayerClusterPluginFactory.cc is supposed to know (be able to include the headers of) the types. So, only one compilation unit and only HGCalLayerClusterPluginFactory.cc to have the algo template specialized.
> 
> Still, the last solution proposed in [#28468 (review)](https://github.com/cms-sw/cmssw/pull/28468#pullrequestreview-340523998) looks fine for me as well

Fine for me too and thanks for pointing this out.

mariadalfonso:
```
diff --git a/RecoLocalCalo/HGCalRecProducers/plugins/HGCalCLUEAlgo.icc b/RecoLocalCalo/HGCalRecProducers/plugins/HGCalCLUEAlgo.icc
index 9489d0789f3..c604a02f63a 100644
--- a/RecoLocalCalo/HGCalRecProducers/plugins/HGCalCLUEAlgo.icc
+++ b/RecoLocalCalo/HGCalRecProducers/plugins/HGCalCLUEAlgo.icc
@@ -553,3 +553,7 @@ template <typename T>
 Density HGCalCLUEAlgoT<T>::getDensity() {
   return density_;
 }
+
+// definition of explicit template instantiation
+template class HGCalCLUEAlgoT<HGCalLayerTiles>;
+template class HGCalCLUEAlgoT<HFNoseLayerTiles>;
\ No newline at end of file
diff --git a/RecoLocalCalo/HGCalRecProducers/plugins/HGCalCLUEAlgoT.h b/RecoLocalCalo/HGCalRecProducers/plugins/HGCalCLUEAlgoT.h
index f3612fa1198..771a7f2c244 100644
--- a/RecoLocalCalo/HGCalRecProducers/plugins/HGCalCLUEAlgoT.h
+++ b/RecoLocalCalo/HGCalRecProducers/plugins/HGCalCLUEAlgoT.h
@@ -207,9 +207,10 @@ private:
   void setDensity(const unsigned int layerId);
 };
 
-#include "RecoLocalCalo/HGCalRecProducers/plugins/HGCalCLUEAlgo.icc"
+//#include "RecoLocalCalo/HGCalRecProducers/plugins/HGCalCLUEAlgo.icc"
 
-using HGCalCLUEAlgo = HGCalCLUEAlgoT<HGCalLayerTiles>;
-using HFNoseCLUEAlgo = HGCalCLUEAlgoT<HFNoseLayerTiles>;
+// declaration of explicit template instantiation
+extern template class HGCalCLUEAlgoT<HGCalLayerTiles>;
+extern template class HGCalCLUEAlgoT<HFNoseLayerTiles>;
 
 #endif
```

With the above modifications, the code does not compile

fwyzard:
I also renamed the `.icc` file back to `.cc` .

mariadalfonso:
> I also renamed the `.icc` file back to `.cc` .

still does not compile, see error here http://dalfonso.web.cern.ch/dalfonso/PR28468_Bocci.txt

fwyzard:
This does work for me:
```bash
$ cmsrel CMSSW_11_1_X_2020-01-09-2300
$ cd CMSSW_11_1_X_2020-01-09-2300/src
$ cmsenv
$ git cms-init
$ git cms-merge-topic 28468
$ patch -p1 < ~/public/HGCalCLUEAlgo.patch
$ mv RecoLocalCalo/HGCalRecProducers/plugins/HGCalCLUEAlgoT.h RecoLocalCalo/HGCalRecProducers/plugins/HGCalCLUEAlgo.h
$ mv RecoLocalCalo/HGCalRecProducers/plugins/HGCalCLUEAlgo.icc RecoLocalCalo/HGCalRecProducers/plugins/HGCalCLUEAlgo.cc
$ scram b -j10
>> Local Products Rules ..... started
>> Local Products Rules ..... done
>> Building CMSSW version CMSSW_11_1_X_2020-01-09-2300 ----
...
(roughly one hour later)
...
>> Done generating edm plugin poisoned information
>> All python modules compiled
>> Plugins of all types refreshed.
```

mariadalfonso:
but still the not for me, same error when compiling the HGCalLayerClusterPluginFactory.cc

```
diff --git a/RecoLocalCalo/HGCalRecProducers/plugins/HGCalCLUEAlgo.icc b/RecoLocalCalo/HGCalRecProducers/plugins/HGCalCLUEAlgo.icc
index 9489d0789f3..e6bd8cb3ee4 100644
--- a/RecoLocalCalo/HGCalRecProducers/plugins/HGCalCLUEAlgo.icc
+++ b/RecoLocalCalo/HGCalRecProducers/plugins/HGCalCLUEAlgo.icc
@@ -1,4 +1,4 @@
-#include "RecoLocalCalo/HGCalRecProducers/plugins/HGCalCLUEAlgoT.h"
+#include "RecoLocalCalo/HGCalRecProducers/plugins/HGCalCLUEAlgo.h"
 
 // Geometry
 #include "DataFormats/HcalDetId/interface/HcalSubdetector.h"
@@ -553,3 +553,7 @@ template <typename T>
 Density HGCalCLUEAlgoT<T>::getDensity() {
   return density_;
 }
+
+// explicit template instantiation
+template class HGCalCLUEAlgoT<HGCalLayerTiles>;
+template class HGCalCLUEAlgoT<HFNoseLayerTiles>;
diff --git a/RecoLocalCalo/HGCalRecProducers/plugins/HGCalCLUEAlgoT.h b/RecoLocalCalo/HGCalRecProducers/plugins/HGCalCLUEAlgoT.h
index f3612fa1198..efa17798bc9 100644
--- a/RecoLocalCalo/HGCalRecProducers/plugins/HGCalCLUEAlgoT.h
+++ b/RecoLocalCalo/HGCalRecProducers/plugins/HGCalCLUEAlgoT.h
@@ -1,5 +1,5 @@
-#ifndef RecoLocalCalo_HGCalRecProducers_HGCalCLUEAlgoT_h
-#define RecoLocalCalo_HGCalRecProducers_HGCalCLUEAlgoT_h
+#ifndef RecoLocalCalo_HGCalRecProducers_HGCalCLUEAlgo_h
+#define RecoLocalCalo_HGCalRecProducers_HGCalCLUEAlgo_h
 
 #include "RecoLocalCalo/HGCalRecProducers/interface/HGCalClusteringAlgoBase.h"
 
@@ -207,9 +207,9 @@ private:
   void setDensity(const unsigned int layerId);
 };
 
-#include "RecoLocalCalo/HGCalRecProducers/plugins/HGCalCLUEAlgo.icc"
+// explicit template instantiation
+extern template class HGCalCLUEAlgoT<HGCalLayerTiles>;
+extern template class HGCalCLUEAlgoT<HFNoseLayerTiles>;
 
-using HGCalCLUEAlgo = HGCalCLUEAlgoT<HGCalLayerTiles>;
-using HFNoseCLUEAlgo = HGCalCLUEAlgoT<HFNoseLayerTiles>;
 
 #endif
diff --git a/RecoLocalCalo/HGCalRecProducers/plugins/HGCalLayerClusterPluginFactory.cc b/RecoLocalCalo/HGCalRecProducers/plugins/HGCalLayerClusterPluginFactory.cc
index 03c17da416c..f41a2d0c5a7 100644
--- a/RecoLocalCalo/HGCalRecProducers/plugins/HGCalLayerClusterPluginFactory.cc
+++ b/RecoLocalCalo/HGCalRecProducers/plugins/HGCalLayerClusterPluginFactory.cc
@@ -1,7 +1,7 @@
 #include "RecoLocalCalo/HGCalRecProducers/interface/HGCalLayerClusterAlgoFactory.h"
 #include "RecoLocalCalo/HGCalRecProducers/interface/HGCalClusteringAlgoBase.h"
 #include "RecoLocalCalo/HGCalRecProducers/interface/HGCalImagingAlgo.h"
-#include "RecoLocalCalo/HGCalRecProducers/plugins/HGCalCLUEAlgoT.h"
+#include "RecoLocalCalo/HGCalRecProducers/plugins/HGCalCLUEAlgo.h"
 #include "FWCore/ParameterSet/interface/ValidatedPluginFactoryMacros.h"
 #include "FWCore/ParameterSet/interface/ValidatedPluginMacros.h"
 
mv RecoLocalCalo/HGCalRecProducers/plugins/HGCalCLUEAlgo.icc RecoLocalCalo/HGCalRecProducers/plugins/HGCalCLUEAlgo.cc
mv RecoLocalCalo/HGCalRecProducers/plugins/HGCalCLUEAlgoT.h RecoLocalCalo/HGCalRecProducers/plugins/HGCalCLUEAlgo.h
```

mariadalfonso:
@slava77 
If there are no issues with the current implementation ,  
I suggest to get going this long standing PRreview as in 4fe57d5789b76f616084370774656cd8f0d09352 and defer to the HGCAL group a future redesign of their classes

rovere:
You removed the using directives, didn't you?
Did you follow the instructions @fwyzard  sent?

fwyzard:
> but still not for me

I'm sorry it doesn't compile for you, but I posted a working recipe that can be copy-pasted on any lxplus, so I don't think the problem is with the recipe or with the proposed changes.

mariadalfonso:
> You removed the using directives, didn't you?
good, the using needed to stay

slava77:
@cmsbuild please test

mariadalfonso:
@slava77 
are there other outstanding things in this PR ?

slava77:
+1

for https://github.com/cms-sw/cmssw/pull/28468	5d29327 
- jenkins tests pass and comparisons with the baseline show no differences (new features are enabled only with phase2_hfnose, not included in the standard set used in jenkins)

Some plots with in a workflow with the HFNose enabled were posted earlier in https://github.com/cms-sw/cmssw/pull/28468#issuecomment-567329733

kpedro88:
+upgrade

silviodonato:
+1

PR #28474, opened 2019-11-26
make HB TDC conversion configurable

lwang046:
#### PR description:

- Adding 2 configurable variables to HcalDigiToRawuHTR class and passing them to convertHB() method in PackerHelp.h to allow easy management of HB tdc bits from job config. 

- Expect no impact on regular workflows.

- Default values of configurables tdc1/tdc2 are current "guesstimates" from the first-pass look at LLP physics and will be studied (once in the release) and adjusted later on, first with some private MC mini-productions, then central samples will be requested.

### DIFF
--- a/EventFilter/HcalRawToDigi/plugins/HcalDigiToRawuHTR.cc
+++ b/EventFilter/HcalRawToDigi/plugins/HcalDigiToRawuHTR.cc
@@ -48,8 +48,8 @@ public:
 
 private:
   int _verbosity;
-  int tdc_1;
-  int tdc_2;
+  int tdc1_;
+  int tdc2_;
   std::string electronicsMapLabel_;
 
   edm::EDGetTokenT<HcalDataFrameContainer<QIE10DataFrame> > tok_QIE10DigiCollection_;
@@ -63,8 +63,8 @@ private:
 
 HcalDigiToRawuHTR::HcalDigiToRawuHTR(const edm::ParameterSet& iConfig)
     : _verbosity(iConfig.getUntrackedParameter<int>("Verbosity", 0)),
-      tdc_1(iConfig.getParameter<int>("tdc1")),
-      tdc_2(iConfig.getParameter<int>("tdc2")),
+      tdc1_(iConfig.getParameter<int>("tdc1")),
+      tdc2_(iConfig.getParameter<int>("tdc2")),
       electronicsMapLabel_(iConfig.getParameter<std::string>("ElectronicsMap")),
       tok_QIE10DigiCollection_(
           consumes<HcalDataFrameContainer<QIE10DataFrame> >(iConfig.getParameter<edm::InputTag>("QIE10"))),
@@ -150,7 +150,7 @@ void HcalDigiToRawuHTR::produce(edm::StreamID id, edm::Event& iEvent, const edm:
 
       //   convert to hb qie data if hb
       if (HcalDetId(detid.rawId()).subdet() == HcalSubdetector::HcalBarrel)
-        qiedf = convertHB(qiedf, tdc_1, tdc_2);
+        qiedf = convertHB(qiedf, tdc1_, tdc2_);
 
       if (!uhtrs.exist(uhtrIndex)) {
         uhtrs.newUHTR(uhtrIndex, presamples);
--- a/EventFilter/HcalRawToDigi/plugins/HcalDigiToRawuHTR.cc
+++ b/EventFilter/HcalRawToDigi/plugins/HcalDigiToRawuHTR.cc
@@ -48,6 +48,8 @@ public:
 
 private:
   int _verbosity;
+  int tdc_1;
+  int tdc_2;
   std::string electronicsMapLabel_;
 
   edm::EDGetTokenT<HcalDataFrameContainer<QIE10DataFrame> > tok_QIE10DigiCollection_;
@@ -61,6 +63,8 @@ private:
 
 HcalDigiToRawuHTR::HcalDigiToRawuHTR(const edm::ParameterSet& iConfig)
     : _verbosity(iConfig.getUntrackedParameter<int>("Verbosity", 0)),
+      tdc_1(iConfig.getParameter<int>("tdc1")),
+      tdc_2(iConfig.getParameter<int>("tdc2")),
       electronicsMapLabel_(iConfig.getParameter<std::string>("ElectronicsMap")),
       tok_QIE10DigiCollection_(
           consumes<HcalDataFrameContainer<QIE10DataFrame> >(iConfig.getParameter<edm::InputTag>("QIE10"))),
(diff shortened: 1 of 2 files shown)


perrotta:
please test

mariadalfonso:
@perrotta @fabiocos 
This is a desirable PR to add in the 11_0.

Will not modify downstream workflow, but will add new features to the RAW that will enable to re-emulate different L1 triggers for LLP in view of Run3 studies

abdoulline:
Yes, we expect TDC timing plots to change.

perrotta:
This code do not only makes HB TDC conversion configurable, but  it also modifies the 
mapping between the HE QIE digies and the HB data format:
- "tdc1" from 24 to 4
- "tdc2" from 49 to 24
-  (possible) "tdc3 from being 63 only to the range 24-49

This change is what originates the change in the TDC plots that was already [mentioned ](https://github.com/cms-sw/cmssw/pull/28474#issuecomment-558786547) by @abdoulline 

Please add a line in the PR description to keep track of this change, and perhaps also to justify it.

Don't you need to make configurable also the threshold which is now hardcoded at "49"? If not, please at least define a named constant ("tdc3"?) for it.

abdoulline:
Andrea,

*49* is a fixed hardware parameter (max.resolution for physical TDC counts 
in QIE10/11 chip), it will not be changed, so it doesn't make  much sense 
to make it configurable, other than to "harmonize" the visual 
representation of the values.

Previous fully-hardcoded conversion was corresponding to 0-th version of 
the HCAL firmware, which is now re-designed to what's emulated in this PR.

Default values of configurables tdc1/tdc2 are current 
"guesstimates" from first-pass look at LLP physics and will be studied 
(once in the release) and adjusted later on, first with some private MC 
mini-productions, then central samples will be requested.

EventFilter/HcalRawToDigi/plugins/HcalDigiToRawuHTR.cc
```diff
@@ -48,6 +48,8 @@ class HcalDigiToRawuHTR : public edm::global::EDProducer<> {
 
 private:
   int _verbosity;
+  int tdc_1;
+  int tdc_2;
   std::string electronicsMapLabel_;
```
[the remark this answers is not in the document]
lwang046: Fixed

lwang046:
I'm not sure it could be called as a mapping. As Salavat mentioned above, The available 4 states from version 0 of the 2 bits of HB tdc were assigned as 0-24, 25-49, and 2 error states. Now these 2 bits are assigned as [0,tdc1], (tdc1, tdc2], (tdc2, 49], and 1 error state, where tdc1 and tdc2 can be easily changed according to future studies. 

> This code do not only makes HB TDC conversion configurable, but it also modifies the
> mapping between the HE QIE digies and the HB data format:
> 
> * "tdc1" from 24 to 4
> * "tdc2" from 49 to 24
> * (possible) "tdc3 from being 63 only to the range 24-49
> 
> This change is what originates the change in the TDC plots that was already [mentioned ](https://github.com/cms-sw/cmssw/pull/28474#issuecomment-558786547) by @abdoulline

abdoulline:
Right,  while tdc1 and tdc2 are firmware configurables, *49* is not (just chip specification).
Well, we can:
(1)  set private member  const tdcmax = 49; 
(2) add sanity check (and throw logWarning if not OK) when taking these values from the ParameterSet (like tdc1 <= tdc2 <= tdcmax) in producer constructor.  

@perrotta  would it be more, say, elegant?

perrotta:
Thank you @lwang046 and @abdoulline 

> Right, while tdc1 and tdc2 are firmware configurables, _49_ is not (just chip specification).
> Well, we can:
> (1) set private member const tdcmax = 49;
> (2) add sanity check (and throw logWarning if not OK) when taking these values from the ParameterSet (like tdc1 <= tdc2 <= tdcmax) in producer constructor.
> 
> @perrotta would it be more, say, elegant?

(1) is exactly what I had in mind (maybe using a ```constexpr```, rather than a ```const``` data member); (2) is certainly welcome

Please also add something like https://github.com/cms-sw/cmssw/pull/28474#issuecomment-559255518 to the PR description, just for reference

perrotta:
Not related, and NOT meant for this pull request.

When I digged into the module parameters I realized that  "verbosity" is only passed as parameter to the methods of UHTRpacker (defined in EventFilter/HcalRawToDigi/plugins/PackerHelp.h). These methods accept "verbosity" as argument, but then they never use it inside the code: that parameter could be fully removed here and there in a follow up cleaning PR of the code. Do you agree?

abdoulline:
OK, Andrea. I agree with all your points, including cleaning of verbosity parameter later on.

@lwang046 
Long, could you please, 
 - add, constexpr tdcmax to HcalDigiToRawuHTR.cc, 
 - make a sanity check of tcd1,2 ranges as mentioned above
 -  add tdcmax it as yet another parameter to  qiedf = convertHB(qiedf, tdc1_, tdc2_);

EventFilter/HcalRawToDigi/plugins/PackerHelp.h
```diff
@@ -637,7 +637,7 @@ QIE11DataFrame convertHB(QIE11DataFrame qiehe, int tdc1, int tdc2) {
       tdc = 0;
     else if (tdc > tdc1 && tdc <= tdc2)
       tdc = 1;
-    else if (tdc2 < 49 && tdc > tdc2 && tdc <= 49)
+    else if (tdc2 < tdcmax && tdc > tdc2 && tdc <= tdcmax)
```
perrotta: ```tdc2 < tdcmax``` is redundant here
(Given the sanity check you introduced before this method is called the decision structure could even be further simplified. However, I have nothing in contrary if the ```convertHB``` method can remain agnostic of the input parameters)

lwang046: Agreed and changed, since `convertHB` is unlikely to be used in other circumstances.

perrotta:
please test

perrotta:
+1
- HB TDC conversion are made configurable, and the mapping between the HE QIE bits and the four states of the HB TDC is also modified with respect to the previous one
- TDC plots gets modified as a consequence
- Jenkins test pass with no other effects in outputs

abdoulline:
@fabiocos 
Is there a chance we still have it in 11_0_0?

perrotta:
> @fabiocos
> Is there a chance we still have it in 11_0_0?

@abdoulline  maybe you can prepare a backport PR to 10_0 in the meanwhile, which can use the very same github branch at this point. And tomorrow at the ORP I think we can discuss with @fabiocos the strategies for the possible backports.

abdoulline:
@lwang046 
Long, could you "cherry-pick" (collect) the relevant modified files from you earlier submitted PRs #28330, #28368 and this one in your local 10_6_6 branch and submit/backport it as a PR to 10_6_X  ?

perrotta:
> @lwang046
> Long, could you "cherry-pick" (collect) the relevant modified files from you earlier submitted PRs #28330, #28368 and this one in your local 10_6_6 branch and submit/backport it as a PR to 10_6_X ?

I meant a backport to 11_0_X (I have corrected the typo in my previous post, sorry for it)

abdoulline:
@perrotta 
Sorry, I was confused by your typo "10_0" last night...
While your "can use very same github branch" can only be done (and should be) for 11_X. 
So I withdraw my  10_6_X request.

fabiocos:
@abdoulline we need to discuss its real impact and need, I'd like to avoid massive backports that are not strictly necessary

fabiocos:
+1

PR #28494, opened 2019-11-27
A fix for importing dependencies in DQM bin by bin tool

andrius-k:
#### PR description:

A fix for importing dependencies in DQM bin by bin tool

#### PR validation:

PR was validated locally.

PR #28407, opened 2019-11-15
DQM: new DQMGlobalEDAnalyzer

schneiml:
#### PR description:
This PR imports more features from the "new DQMStore" project:

- A new `DQMEDAnalyzer` base class, now based on `EDProducer`, like all the other DQM module base classes. This is to make sure that everything that _produces_ DQM histograms actually is known to the EDM framework as a _Producer_, even if the data still flows through the `DQMStore`; otherwise, we have no guarantees that modules will run in the correct order.

- Some cleanup removing unused declarations in `DQMStore.h`, including `fastmatch`. The previous cleanup PRs left these unused.

Update: some things went into #28612, removed here now.

#### PR validation:

The challenge with this PR is that `DQMGlobalEDAnalyzer`s are primarily used in HLT, and I am not sure how to test them there. For now, HLT configs should reliably crash due to the incorrect module type used in the auto-generated configs; I am not sure how to change this part of the code. Also `FastTimerService` and `ThroughputService` need to be checked. Let's see if any of that is picked up in the PR tests, else I'll have to defer that to the HLT experts.

PR #28501, opened 2019-11-28
DQM: Script to locally run the bin-to-bin comparison tool

schneiml:
#### PR description:
The bin-to-bin comparison tool has proven useful, but to use it, one first has to produce the input DQM files. 

This script downloads the data from the PR test result page and then (re-)runs the comparison locally. This can be useful to compare between PRs, or to apply local changes to the tool, or while the tool is broken (like right now).

#### PR validation:

Successfully produced a comparison for #28407.

Running it is rather tricky, since `cern-get-sso-cookie` does not work inside a cmssw release area currently. Also the bin-to-bin tool is broken at the moment and the fix in #28494 needs to be applied first.

### DIFF
--- a/DQMServices/FileIO/scripts/compareFilesFromPR.sh
+++ b/DQMServices/FileIO/scripts/compareFilesFromPR.sh
@@ -1,0 +1,34 @@
+#!/bin/bash
+RESULTURL="$1"
+BASELINEURL="$2"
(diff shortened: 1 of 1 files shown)


jfernan2:
please test

jfernan2:
+1

fabiocos:
+1

PR #28508, opened 2019-11-29
Backport of "Changes for new trained regression for UL 2016" in 10_6_X  

jainshilpi:
This is just a backport of #28466 

PR description:
These changes are essential for making the 2016 UL regression work. After discussion with Sam, I have made these changes (since 2016 uses the old PF cluster regression strategy and old ele/pho regressions codes).
We had to create Modifier_run2_ECAL_2016_cff.py so that the training strategy of 2017 and 2018 could be used here as well.

### DIFF
--- a/Configuration/Eras/python/Era_Run2_2016_cff.py
+++ b/Configuration/Eras/python/Era_Run2_2016_cff.py
@@ -11,7 +11,8 @@ from Configuration.Eras.Modifier_run2_tau_ul_2016_cff import run2_tau_ul_2016
 from Configuration.Eras.Modifier_run2_L1prefiring_cff import run2_L1prefiring
 from Configuration.Eras.Modifier_pixel_2016_cff import pixel_2016
 from Configuration.Eras.Modifier_run2_jme_2016_cff import run2_jme_2016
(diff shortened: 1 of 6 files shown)


perrotta:
@cmsbuild please test

slava77:
backport of #28466

slava77:
+1

for https://github.com/cms-sw/cmssw/pull/28508	0edccec 
- code changes are consistent with #28466	857cfe1 (merged on Nov 26)
- jenkins tests pass and comparisons with the baseline show differences only in the 2016 workflows starting from ecal superclusters and egamma objects
    - similar to the master version, it looks like the GT still needs an update to have consistent physics results; IIUC, this is going to be added/applied separately in another PR or at the relval/production job submission stage

slava77:
@fabiocos 
changes in this PR conflict with #28512; they can not be merged independently.
May I suggest to merge this PR so that #28512 can be consequently updated.
Otherwise we need an über-PR

fabiocos:
+operations

the PR is consistent with its master version

fabiocos:
@slava77 there is indeed a conflict in Eras, as #28512 needs to be anyway updated (and we need to try to integrate it quickly) I fully agree with your suggestion, no reason to make things more complex combining different things

fabiocos:
+1

PR #28528, opened 2019-12-03
APE tool: Files for 2016 trends, option to use/not use condor CAF queue

mteroerd:
#### PR description:
- The 2016 runs with their respective luminosities were added so they can be accessed by the plotting tool for APE trends. A typo in the trend plotting tool was also fixed.
- When a measurement is started that uses loaded conditions from sqlite files, checks are now performed to see if the sqlite files exist. Before, this was not caught so that jobs were submitted to condor and then crashed. At the moment there is still no check if the used tag and record are available in the source.
- An option is added to submit condor jobs for APE measurements to the CAF queue. It is off by default. With this option, one does not have to connect to lxplus-t0.cern.ch anymore.


christopheralanwest:
please test

pohsun:
+1

fabiocos:
+1

PR #28531, opened 2019-12-03
Fix two bugs in CommonTools/Utils unit tests found by ASAN

dan131riley:
#### PR description:

ASAN found two fairly trivial bugs in the CommonTools/Utils unit tests.  One bug used the wrong type for the local geometry, which violated an assumption made by the BaseTrackerRecHit here:

https://github.com/cms-sw/cmssw/blob/30a2ff8ff13aaa416259453b54ad0b62588edc91/DataFormats/TrackerRecHit2D/interface/BaseTrackerRecHit.h#L29

The second bug takes a reference to a vector element which can be invalidated if the vector resizes.  This fix preallocates the vector so that it does not resize.

#### PR validation:

Tests were re-run in an ASAN build and verified that the bugs were fixed.  No effect on anything outside the package unit tests.

### DIFF
--- a/CommonTools/Utils/test/ExpressionEvaluatorUnitTest.cpp
+++ b/CommonTools/Utils/test/ExpressionEvaluatorUnitTest.cpp
@@ -20,7 +20,7 @@ int main() {
   using MyExpr = reco::MaskCollection<eetest::CandForTest>;
 
   std::vector<Cand> oriC;
-  oriC.reserve(10); // pre-allocate so it doesn't reallocate
+  oriC.reserve(10);  // pre-allocate so it doesn't reallocate
   MyExpr::Collection c;
   for (int i = 5; i < 15; ++i) {
     oriC.emplace_back(Cand(i, 1, 1));
--- a/CommonTools/Utils/test/ExpressionEvaluatorUnitTest.cpp
+++ b/CommonTools/Utils/test/ExpressionEvaluatorUnitTest.cpp
@@ -20,6 +20,7 @@ int main() {
   using MyExpr = reco::MaskCollection<eetest::CandForTest>;
 
   std::vector<Cand> oriC;
+  oriC.reserve(10); // pre-allocate so it doesn't reallocate
   MyExpr::Collection c;
   for (int i = 5; i < 15; ++i) {
     oriC.emplace_back(Cand(i, 1, 1));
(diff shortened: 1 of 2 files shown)


smuzaffar:
please test

santocch:
+1

perrotta:
@dan131riley , could you please about the bug fixed in ```testCutParser.cc```?
(Ok, it is test code which does not enter in any workflow; still, I would like to understand it, for my own education).

SiStripMatchedRecHit2D derives from BaseTrackerRecHit, and in both cases a ```GeomDet``` argument is required in the constructor, and not a ```TrackerGeomDet```. I would (naively?) say that either there is a bug in our code (which assumes ```TrackerGeomDet``` functionalities without requiring that class as argument), or this is not really a "bug" but simply there is an implicit assumption (certified by the ```static_cast```) that a ```TrackerGeomDet``` will be fed instead.

What is the correct explanation?

dan131riley:
@perrotta Unfortunately, the GeomDet code is full of static casts that implicitly assume the types.  I took a look at what would be involved in changing BaseTrackerRecHit to take an explicit TrackerGeomDet, and found that the changes would ripple throughout the GeomDet code.  It would be a massive refactoring to make the code compile-time type safe.  It could be made run-time type safe by changing to dynamic casts.  I'm currently running a limited matrix with that one cast changed to be dynamic, but I don't expect it to find anything.  If it does, further investigation will certainly be warranted.

perrotta:
> @perrotta Unfortunately, the GeomDet code is full of static casts that implicitly assume the types. I took a look at what would be involved in changing BaseTrackerRecHit to take an explicit TrackerGeomDet, and found that the changes would ripple throughout the GeomDet code. It would be a massive refactoring to make the code compile-time type safe. It could be made run-time type safe by changing to dynamic casts. I'm currently running a limited matrix with that one cast changed to be dynamic, but I don't expect it to find anything. If it does, further investigation will certainly be warranted.

Thank you Dan, this roughly corresponds to what I suspected. Indeed, a massive refactoring of the code would require a major effort, which certainly goes beyond the scope of these fixes. If really needed/wanted it should be rather coordinated by the tracking POG, I guess.
For the moment I think that the fix you are proposing here is safe enough for its purposes.

perrotta:
+1
- Technical, and only affecting a test area
- Jenkins tests pass

fabiocos:
+1

PR #28540, opened 2019-12-04
rename tmp file to avoid race condition between PyDevParameterSet & PythonParameterSet

smuzaffar:
#### PR description:

This change makes sure that unit tests in PyDevParameterSet and PythonParameterSet generates different file name. Currently they both generate and delete `$CMSSW_BASE/src/tmp.py` which causes one of these tests to randomly fail


smuzaffar:
please test

Dr15Jones:
+1

fabiocos:
+1

@smuzaffar thanks for catching the problem

Issue #28566, opened 2019-12-05
Memory problem in HGCalBackendLayer2Producer


Dr15Jones:
The address sanitizer found the following when running workflow 21634.0 on step 2
```
=================================================================
==31360==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x61400106d5d0 at pc 0x2ad3810a3d48 bp 0x7ffcee46a550 sp 0x7ffcee46a548
READ of size 8 at 0x61400106d5d0 thread T0
    #0 0x2ad3810a3d47 in HGCalShowerShape::percentileLayer(l1t::HGCalMulticluster const&amp, HGCalTriggerGeometryBase const&amp, float) const (/cvmfs/cms-ib.cern.ch/nweek-02604/slc7_amd64_gcc820/cms/cmssw/CMSSW_11_0_ASAN_X_2019-11-29-2300/lib/slc7_amd64_gcc820/libL1TriggerL1THGCal.so+0x109d47)
    #1 0x2ad3810d6362 in HGCalShowerShape::fillShapes(l1t::HGCalMulticluster&amp, HGCalTriggerGeometryBase const&amp) const (/cvmfs/cms-ib.cern.ch/nweek-02604/slc7_amd64_gcc820/cms/cmssw/CMSSW_11_0_ASAN_X_2019-11-29-2300/lib/slc7_amd64_gcc820/libL1TriggerL1THGCal.so+0x13c362)
    #2 0x2ad38101ebb5 in HGCalHistoClusteringImpl::finalizeClusters(std::vector&lt;l1t::HGCalMulticluster, std::allocator&lt;l1t::HGCalMulticluster&gt; &gt;&amp, BXVector&lt;l1t::HGCalMulticluster&gt;&amp, HGCalTriggerGeometryBase const&amp) const (/cvmfs/cms-ib.cern.ch/nweek-02604/slc7_amd64_gcc820/cms/cmssw/CMSSW_11_0_ASAN_X_2019-11-29-2300/lib/slc7_amd64_gcc820/libL1TriggerL1THGCal.so+0x84bb5)
[...]

PR #28570, opened 2019-12-06
[HGCAL trigger] Fix percentile out of bound

jbsauvan:
#### PR description:

Fix out of bound in layer percentile interpolation. See issue #28566.

#### PR validation:
Ran on HGCAL V9 sample. 
Tested extreme case where no trigger cell are selected in the computation, which gives 0 energy in all the layers.

### DIFF
--- a/L1Trigger/L1THGCal/src/backend/HGCalShowerShape.cc
+++ b/L1Trigger/L1THGCal/src/backend/HGCalShowerShape.cc
@@ -124,7 +124,7 @@ float HGCalShowerShape::percentileLayer(const l1t::HGCalMulticluster& c3d,
   }
   std::partial_sum(layers.begin(), layers.end(), layers.begin());
   double pt_threshold = layers.back() * quantile;
-  int percentile = 0;
+  unsigned percentile = 0;
   for (double pt : layers) {
     if (pt > pt_threshold) {
       break;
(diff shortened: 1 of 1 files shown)


kpedro88:
please test

L1Trigger/L1THGCal/src/backend/HGCalShowerShape.cc
```diff
@@ -133,7 +133,7 @@ float HGCalShowerShape::percentileLayer(const l1t::HGCalMulticluster& c3d,
   }
   // Linear interpolation of percentile value
   double pt0 = (percentile > 0 ? layers[percentile - 1] : 0.);
-  double pt1 = layers[percentile];
+  double pt1 = (percentile < layers.size() ? layers[percentile] : layers.back());
```
Dr15Jones: Can `layers.size() == 0`? If so then calling `layers.back()` in that case reads off the end (and may cause a segmentation fault since it is likely to be reading memory location 0).

jbsauvan: The size of `layers` is given by the number of HGCAL layers :
https://github.com/cms-sw/cmssw/blob/04f9376ae2a8d215fb8f906b71e8ab5b2517caa8/L1Trigger/L1THGCal/src/backend/HGCalShowerShape.cc#L111-L112
where the trigger tools returns this:
https://github.com/cms-sw/cmssw/blob/04f9376ae2a8d215fb8f906b71e8ab5b2517caa8/L1Trigger/L1THGCal/src/HGCalTriggerTools.cc#L72-L73
Which is computed from the HGCAL geometry:
https://github.com/cms-sw/cmssw/blob/04f9376ae2a8d215fb8f906b71e8ab5b2517caa8/L1Trigger/L1THGCal/src/HGCalTriggerTools.cc#L38-L47

So unless something really bad happens it shouldn't be 0

fabiocos:
@kpedro88 @rekovic please have a look, this looks a technical fix which should be ok (unless there is a serious inconsistency in the geometry definition).

kpedro88:
+upgrade

fabiocos:
+1

@rekovic please have a look

fabiocos:
merge

PR #28572, opened 2019-12-06
Skip building onnx code on PowerPC. The external is not supported

mrodozov:
#### PR description:

Skip building code that requires onnxruntime external. The external doesn't build on PowerPC. 
Externals states the arch is not yet supported. Actual error is here:
https://github.com/cms-externals/onnxruntime/blob/cms/v1.0.0/onnxruntime/core/mlas/lib/mlasi.h#L714
and there are other errors but that's where it starts failing.

#### PR validation:

Builds & run tests.

### DIFF
--- a/PhysicsTools/ONNXRuntime/BuildFile.xml
+++ b/PhysicsTools/ONNXRuntime/BuildFile.xml
@@ -1,3 +1,6 @@
+<ifarchitecture name="_ppc64le_">
+<flags SKIP_FILES="%"/>
+</ifarchitecture>
 <use name="onnxruntime"/>
 <use name="FWCore/Framework"/>
 <use name="FWCore/Utilities"/>
--- a/PhysicsTools/ONNXRuntime/test/BuildFile.xml
+++ b/PhysicsTools/ONNXRuntime/test/BuildFile.xml
@@ -1,3 +1,6 @@
+<ifarchitecture name="_ppc64le_">
+<flags SKIP_FILES="%"/>
+</ifarchitecture>
 <bin name="testONNXRuntime" file="testRunner.cpp, testONNXRuntime.cc">
     <use name="boost_filesystem" />
     <use name="cppunit" />
--- a/RecoBTag/ONNXRuntime/plugins/BuildFile.xml
+++ b/RecoBTag/ONNXRuntime/plugins/BuildFile.xml
@@ -1,3 +1,6 @@
+<ifarchitecture name="_ppc64le_">
+<flags SKIP_FILES="%"/>
+</ifarchitecture>
 <use name="FWCore/Framework"/>
 <use name="FWCore/MessageLogger"/>
 <library file="*.cc" name="RecoBTagONNXRuntimePlugins">


mrodozov:
please test

slava77:
@mrodozov 
does "please test for slc7_ppc64le_gcc820" work?

@smuzaffar @fabiocos 
How serious are we about the ppc builds?
Disabling the compilation is not enough, about a half of the relval matrix workflows will fail due to missing plugin.

mrodozov:
it does. but I want to see how it will work for intel first

slava77:
@cmsbuild please test for slc7_ppc64le_gcc820

slava77:
So, the PPC build tests failed.
I somehow thought that it will be at the link time, but apparently the problem starts at the python configuration building time because the auto-generated python files were not created.

Is this an acceptable situation for the PPC builds?
I see two options:
- leave the PPC builds broken and benefit from the performance improvements in the ONNXRuntime-based producers
- revert the switch to ONNXRuntime in modules used for production

I somewhat prefer the former.

makortel:
> So, the PPC build tests failed.
> I somehow thought that it will be at the cmsRun run time, but apparently the problem starts at the python configuration building time because the auto-generated python files were not created.
> 
> Is this an acceptable situation for the PPC builds?
> I see two options:
> 
> * leave the PPC builds broken and benefit from the performance improvements in the ONNXRuntime-based producers
> * revert the switch to ONNXRuntime in modules used for production
> 
> I somewhat prefer the former.

Independently I realized the same will happen with CUDA plugins as well (on `ppc64le`). I created an issue #28576, and with @Dr15Jones we tentatively agreed to discuss on the subject in the next core meeting.

VinInn:
Pretty ridiculous not supporting altivec...

VinInn:
Let me add that this clearly shows how the use of intrinsics is basically flawed.
Compilers (at least gcc and clang) supports "hardware vectors" (ok not VC and this is a MS project). It would have been absolutely trivial to write those 100 lines or so (much less once all the ifdef are removed) to support ANY platform the compiler support.

Just lessons to learn before choosing/writing/supporting  a "generic" software tool

fabiocos:
@slava77 about https://github.com/cms-sw/cmssw/pull/28572#issuecomment-562637276 I have suggested to @smuzaffar that we discuss explicitly this and next Core Software meeting, also in the context of #28576

smuzaffar:
yes agree, we should discuss it tomorrow during core sw meeting.

slava77:
@hqucms 
please note this issue.
Did you see much already about ONNXRuntime on PPC?

mrodozov:
There's the project similar to what we might need, wrapping SSE2 to Altivec calls 
https://github.com/IvantheDugtrio/veclib

VinInn:
interesting. fail to understand why not fully inline, make no sense not letting the compiler see the altivec intrinsics

hqucms:
> @hqucms
> please note this issue.
> Did you see much already about ONNXRuntime on PPC?

@slava77 
I could not find any info about ONNXRuntime on PPC. Seems no plan to support it.
Actually do we need to run these ONNXRuntime-dependent plugins on PPC? If not, could we just make a dummy `PhysicsTools/ONNXRuntime/interface/ONNXRuntime.h` to just let the build pass?

slava77:
> Actually do we need to run these ONNXRuntime-dependent plugins on PPC? 

from the discussion in the core sw meeting today, I understood that we should not hide a problem (e.g. by making some code that should run to silently pass).

I'm curious if in the meantime (since v 1.6) the tensorflow inference was optimized to be similarly fast at the ONNX.

hqucms:
> I'm curious if in the meantime (since v 1.6) the tensorflow inference was optimized to be similarly fast at the ONNX.

@slava77 Do we have a newer TF build in CMSSW to check?

slava77:
> @slava77 Do we have a newer TF build in CMSSW to check?

There was a build at some point for  1.13 https://github.com/cms-sw/cmsdist/pull/4946 but it looks like it was only experimental and it is no longer available in the current test build IBs.

davidlange6:
No because it’s not yet t been made to work....




On 10 Dec 2019, at 20:26, Slava Krutelyov <notifications@github.com> wrote:

﻿

@slava77<https://github.com/slava77> Do we have a newer TF build in CMSSW to check?

There was a build at some point for 1.13 cms-sw/cmsdist#4946<https://github.com/cms-sw/cmsdist/pull/4946> but it looks like it was only experimental and it is no longer available in the current test build IBs.

slava77:
@mrodozov @smuzaffar 
can this PR be closed considering somewhat successful update of ONNX to make it work on PPC?
Please clarify.
Thank you.

mrodozov:
yes. I'll update you on it tomorrow

PR #28580, opened 2019-12-08
updating addition of new 2017 beamspot smearing for 2017 pp @ 5 TeV

gsfs:
#### PR description:

This PR adds a new beamspot option for MC. This beamspot is a modification of that previously used for MC for 2017 pp data @ 5 TeV. It was known for some time that the original beam smearing did not match that in data but it was felt that the difference was small enough not to matter. However, now that we have the much larger 2018 PbPb dataset, new analyses may be possible that are sensitive to the difference in beam smearing in pp. Therefore, it was decided to add a more accurate beam smearing option in CMSSW. This is being put in Master in preparation for back porting to CMSSW_9_4_X (which is the only CMSSW version used for analyzing that 2017 pp data).

#### PR validation:

This exact code has already been used successfully to generate numerous private MC samples with the new beamspot option. We now want the ability to generate official samples and so need this code in an official CMSSW release.


civanch:
please test

fabiocos:
+operations

the update of central configuration looks correct and coherent with the purpose of the PR

civanch:
+1

fabiocos:
+1

PR #11086, opened 2015-09-02
Replace mark-and-delete at next merge algorithm in DQMStore (75x)

dmitrijus:
Additionally, the DQMStore will now acquire locks for writing files.

This PR also add a new file saver for HLT->PB (DQMFileSaverPB).
Which should replace DQMFileSaver in the near future.

Please do not accept this PR until I say okay.

PR #28588, opened 2019-12-09
DQM: new DQMFileSaver

schneiml:
#### PR description:
In offline, the legacy TDirectory files for the DQMGUI are written by the `DQMFileSaver` module. While reading the code and its configurations, it became apparent that this module is not actually used in online or HLT (any more?), even though most of the code is dedicated to these cases. (Online and HLT use `DQMServices/FileIO` now.)

So, I stripped down this module to prove the hypothesis. 

In addition, this PR contains the new TDirectory writing code by @andrius-k, as a new independent class. It is still linked into the DQMStore and can be called as `edm::Service<DQMStore>()->save(...)`. I also updated the Online output modules to use this helper (`DQMFileSaverOnline`) or contain the writing code (`DQMFileSaverPB`). Consequently, `DQMStore` itself does not contain _file writing_ code any more. There is still a huge amount of _file reading_ code, which is largely useless and to be removed next.

#### PR validation:

This PR needs testing in online DQM (not done yet) and ideally also in HLT (or a solid confirmation that the module is never loaded in HLT). The changes to the ProtoBuf writer need to be tested in the HLT/fastHadd/online DQM chain. None of this is done so far. I'll try to provide a test setup in another PR.

This PR needs testing in Tier0 Express processing, where the `runIsComplete` logic is used.

The main point of the PR is to test
- That only a single mode of this module is used in offline, to be confirmed by the IB tests; (as far as it seems confirmed)
- That the new TDirectory writing code works; (confirmed)
- Which unit tests depend on behavior removed in this PR. (none, it seems)

[I suspect some tests to fail (assuming these tests are run in the PR tests, not sure about that), and some flaws in the TDirectory writing code (it does not support QTests yet, for example).] Edit: all fixed.


schneiml:
please test

I did remember to run code-format, this time!

schneiml:
I am surprised that no tests failed, even though #28407 indicated that the old `whiteRabbit` DQM tests run in the PR tests.

The comparison looks much more messy than expected though, we'll need to fix a bunch of things it seems.

schneiml:
Updated to fix the missing parts in the output.

Now, in my local tests it looks like the actual QTest results are different here. Let's see what the bot thinks. It might be a matter of propagating QTest results in all the required places after #28379.

schneiml:
please test

now there should be much less removed and also changed items.

schneiml:
The comparison results look great -- however, I will add some more changes to this PR to also get rid of the old saving code in the DQMStore.

schneiml:
please test

Everything should still be fine, the main changes added now are online/HLT only. However there might be crashes/changes if something calls `DQMStore::save` directly using (now) unsupported features.

schneiml:
Latest comparison (still) looks clean, apart from a QTest on a MessageLogger(?) plot in wf 8.0. Nothing I'd be worried about.

schneiml:
please test

Just for formal reasons.

I fixed a bug in `DQMFileSaverPB` (found with the tests in #28612). So, while I still can't read the PB output, I get bit-by-bit identical output which is a good sign.

Martin-Grunewald:
@schneiml 

DQMFileSaver is used in the online HLT production menus like this:

DQMHistograms = cms.EndPath(   hltGtStage2Digis +   hltScalersRawToDigi +   hltTriggerRatesMonitor +   hltTriggerBxMonitor +   hltTriggerObjectTnPMonitor +   hltLumiMonitor +   hltPSColumnMonitor +   hltDQMFileSaver )

RelVal tests use the offline HLT menus where the above endpath is not included, thus PR testing will not catch any problems!

schneiml:
@Martin-Grunewald where do the files saved by `hltDQMFileSaver` go? Do they go anywhere?

I have the suspicion that `hltDQMFileSaver` is actually a `DQMFileSaverPB`. But I can't prove it, full text search for `hltDQMFileSaver` comes up essentially empty. [1] seems relevant, but I don't understand what it is good for.

[1] https://cmssdt.cern.ch/dxr/CMSSW/source/HLTrigger/Configuration/python/Tools/confdb.py#628-657

Martin-Grunewald:
@schneiml 
The config of the module is this:
```
hltDQMFileSaver = cms.EDAnalyzer( "DQMFileSaver",
    runIsComplete = cms.untracked.bool( False ),
    referenceHandling = cms.untracked.string( "all" ),
    producer = cms.untracked.string( "DQM" ),
    forceRunNumber = cms.untracked.int32( -1 ),
    saveByRun = cms.untracked.int32( 1 ),
    saveAtJobEnd = cms.untracked.bool( False ),
    saveByLumiSection = cms.untracked.int32( 1 ),
    version = cms.untracked.int32( 1 ),
    referenceRequireStatus = cms.untracked.int32( 100 ),
    convention = cms.untracked.string( "FilterUnit" ),
    dirName = cms.untracked.string( "." ),
    fileFormat = cms.untracked.string( "PB" )
)
```

schneiml:
> fileFormat = cms.untracked.string( "PB" )

Oh. This is surprising. It might explain why the files created by `DQMFileSaverPB` are not actually readable (since `DQMFileSaver` in `PB` mode is used in production). 

Could you also tell me the status of the `enableMultiThread` flag in the `DQMStore` (ideally the entire `DQMStore` config PSet) and if the job runs multi-threaded (I assume yes)? I'll try to reproduce a test setup.

Martin-Grunewald:
The DQMStore is configured like this:
```
DQMStore = cms.Service( "DQMStore",
    referenceFileName = cms.untracked.string( "" ),
    verbose = cms.untracked.int32( 0 ),
    collateHistograms = cms.untracked.bool( False ),
    enableMultiThread = cms.untracked.bool( True ),
    forceResetOnBeginLumi = cms.untracked.bool( False ),
    saveByLumi = cms.untracked.bool( False ),
    LSbasedMode = cms.untracked.bool( True ),
    verboseQT = cms.untracked.int32( 0 )
)
```

schneiml:
@Martin-Grunewald Thanks! `LSbasedMode = True` with `enableMultiThread` is not a combination I expected to see...

Looking at #11086, I wonder if we should make "in the near future" today, 4 years later, or rather just ignore this. The discussion back then fully applies today, and I support the reasoning to switch to `DQMFileSaverPB`. But we have no evidence that it actually works. So apart from changing the HLT config, it also requires some validation.

Martin-Grunewald:
Hmm, ConfDB (which holds the HLT menus) knows only about ```DQMFileSaverOnline``` and ```DQMFileSaver``` but not about ```DQMFileSaverPB```, which is a problem.

A cfi file (used by ConfDB) exists but is faulty:
```
mport FWCore.ParameterSet.Config as cms
0002 
0003 # DQM file saver module
0004 dqmSaver = cms.EDAnalyzer("DQMFileSaverPB",
0005     # Name of the producer.
0006     producer = cms.untracked.string('DQM'),
0007     # Directory in which to save the files.
0008     path = cms.untracked.string('./'),
0009 
0010     # Tag, used in the filename as the third term.
0011     tag = cms.untracked.string('UNKNOWN'),
0012 
0013     # Control reference saving (default / skip / qtests / all)
0014     referenceHandling = cms.untracked.string('all'),
0015     # Control which references are saved for qtests (default: STATUS_OK)
0016     referenceRequireStatus = cms.untracked.int32(100),
0017 
0018     # If set, EvFDaqDirector is emulated and not used
0019     fakeFilterUnitMode = cms.untracked.bool(false),
0020     # Label of the stream
0021     streamLabel = cms.untracked.string("streamDQMHistograms"),
0022 )

```
thus ConfDb does not learn aboutthe plugin for use in a menu config.


So could you please fix in the above:
```
  fakeFilterUnitMode = cms.untracked.bool(false),
```
to be:
```
  fakeFilterUnitMode = cms.untracked.bool(False),
```
Thanks!

Further, it seems we never tried and it is hard to test as this is not run online.
It probably requires hilton playback of even an MWGR?

@gennai @Sam-Harper 
I assume FOG needs to get involved!

schneiml:
@Martin-Grunewald I quickly fixed the typo, if that helps you.

Yes, I am afraid that we'll need to test that in a MWGR, esp. since I'll also need to revise the code that reads these files later (`DQMStreamerReader`) and i failed to get this to run at all in a non-online setup.

However, for the `DQMFileSaverPB` change, we can hope for bit-by-bit identical output (at least as long as this PR is *not* merged -- the changes here modify the ME flags so the output will be not identical).

Sam-Harper:
@Martin-Grunewald , thanks, I have alerted FOG and we can see if anybody is around to do a quick hilton test

schneiml:
@Sam-Harper if you get to it, can you provide a sample of the `streamDQMHistograms` output stream? This is what we need to be able to read on online DQM side, and while I can't reproduce this currently (the files seem to be missing a "streamer" header -- is that because I used the `fakeFilterUnitMode` or does some DAQ mechanism add that later?), looking at the files themselves would be useful...

fabiocos:
please test

missirol:
> @Sam-Harper if you get to it, can you provide a sample of the `streamDQMHistograms` output stream? This is what we need to be able to read on online DQM side, and while I can't reproduce this currently (the files seem to be missing a "streamer" header -- is that because I used the `fakeFilterUnitMode` or does some DAQ mechanism add that later?), looking at the files themselves would be useful...

as suggested by @Sam-Harper , we are trying to test this PR on a Hilton machine (in my understanding, we would then be able to provide a sample of the `streamDQMHistograms` output stream), but at the moment we are having issues installing a recent CMSSW release there. I'm still trying, but given the limited availability of people this week, this might get delayed to the new year.

schneiml:
@missirol thanks for the update! I'll try to pull some information out of the DQM online playback system in parallel.

schneiml:
For the record: waiting for changes&validation at HLT.

schneiml:
@missirol ran some tests but there were unrelated problems. Still waiting for HLT validation.

schneiml:
The test by @missirol last week looks ok, so this is good to go. (Sorry, forgot to update this PR in time)

schneiml:
+1

silviodonato:
+1

Issue #28571, opened 2019-12-06
MTDGeomDet type problem


dan131riley:
This is a followup to PR #28531.  In that PR, address sanitizer found that some unit tests were passing the wrong GeomDet type to the BaseTrackerRecHit constructor.  That constructor implicitly assumes that the GeomDet it receives is a TrackerGeomDet via this static_cast:

https://github.com/cms-sw/cmssw/blob/30a2ff8ff13aaa416259453b54ad0b62588edc91/DataFormats/TrackerRecHit2D/interface/BaseTrackerRecHit.h#L29

As an integrity test I ran a limited matrix with all the pointer static_casts in DataFormats/TrackerRecHit2D changed to dynamic_casts, which resulted in crashes in MTDTrackingRecHitProducer::run here:

https://github.com/cms-sw/cmssw/blob/30a2ff8ff13aaa416259453b54ad0b62588edc91/RecoLocalFastTime/FTLRecProducers/plugins/MTDTrackingRecHitProducer.cc#L120

I suspect this is because MTDGeomDet is not a TrackerGeomDet:

https://github.com/cms-sw/cmssw/blob/30a2ff8ff13aaa416259453b54ad0b62588edc91/Geometry/CommonDetUnit/interface/MTDGeomDet.h#L6

In fact, MTDGeomDet appears to be a cut&paste of TrackerGeomDet, so it has the same layout as the TrackerGeomDet, which is likely why the code doesn't crash with the static_cast.  Nevertheless, this is a bug.  I would guess that the correct fix is to make MTDGeomDet a TrackerGeomDet, either by inheritance or a simple 'using' alias (since there don't seem to be any differences in behavior), but a subject expert should take a look.

As a secondary issue, there are a lot of other hidden assumptions in the GeomDet code where static casts are used.  The C++ Core Guidelines (which are referenced by the CMS coding style guide) recommends dynamic_cast for runtime type safety when navigating the class hierarchy is unavoidable.  Since the static_casts have proven to be fallible, should there be a campaign to change the GeomDet static casts to dynamic casts?

dan131riley:
@perrotta this is a followup to your question on #28531

perrotta:
assign geometry, reconstruction

dan131riley:
[...]

PR #28590, opened 2019-12-09
Correct MTDGeomDet type problems

dan131riley:
#### PR description:

The `BaseTrackerRecHit` constructor implicitly assumes that the `GeomDet` passed to it is a `TrackerGeomDet` by doing a `static_cast` to that type:

https://github.com/cms-sw/cmssw/blob/30a2ff8ff13aaa416259453b54ad0b62588edc91/DataFormats/TrackerRecHit2D/interface/BaseTrackerRecHit.h#L29

Changing the `static_cast` to `dynamic_cast` revealed that this assumption is violated by the `MTDTrackingRecHitProducer`, which passes an `MTDGeomDet` here:

https://github.com/cms-sw/cmssw/blob/30a2ff8ff13aaa416259453b54ad0b62588edc91/RecoLocalFastTime/FTLRecProducers/plugins/MTDTrackingRecHitProducer.cc#L120

`MTDGeomDet` does not inherit from `TrackerGeomDet`, but is instead a copy&paste copy of `TrackerGeomDet` that inherits directly from `GeomDet`:

https://github.com/cms-sw/cmssw/blob/30a2ff8ff13aaa416259453b54ad0b62588edc91/Geometry/CommonDetUnit/interface/MTDGeomDet.h#L6

This doesn't crash and isn't caught by the address sanitizer because the class layout is the same, but it is still a bug that could eventually lead to failures if the different classes evolve separately.  Since `MTDGeomDet` currently has functionality identical to `TrackerGeomDet`, this PR makes `MTDGeomDet` a typedef for `TrackerGeomDet`.  If in the future additional functionality is needed in `MTDGeomDet` it can be made a class that inherits from `TrackerGeomDet`.  This resolves issue #28571.

In addition, this PR also removes some casts that are unnecessary since `GeomDet` and `GeomDetUnit` are now the same type, and makes a small change to `AlignmentPositionError` in `GeomDet` and  `TrackerGeomDet` to improve encapsulation and reduce code replication.

#### PR validation:

A limited matrix was run with no new failures observed.  This is strictly a technical fix that should have no effect on physics results.


kpedro88:
please test

kpedro88:
Do the additional `dynamic_cast`s have a noticeable impact on CPU performance?

DataFormats/TrackerRecHit2D/interface/BaseTrackerRecHit.h
```diff
@@ -26,7 +26,8 @@ class BaseTrackerRecHit : public TrackingRecHit {
 
   BaseTrackerRecHit(const LocalPoint& p, const LocalError& e, GeomDet const& idet, trackerHitRTTI::RTTI rt)
```
cvuosalo: Isn't it the case that the `idet` parameter is always a derived `TrackerGeomDet`? The dynamic cast will not work if that is not true, since `det()` returns the pointer to `idet`. If `idet` is always a `TrackerGeomDet`, why not declare it as such, and then line 30 simply becomes:
```
LocalError lape = idet.localAlignmentError();
```

DataFormats/TrackerRecHit2D/src/ProjectedSiStripRecHit2D.cc
```diff
@@ -5,6 +5,6 @@
 
 void ProjectedSiStripRecHit2D::setDet(const GeomDet& idet) {
```
cvuosalo: Couldn't we just declare `idet` as `const GluedGeomDet&`? It is required to be `GluedGeomDet` by the dynamic cast.

VinInn:
If MTD pretends to be part of tracker better start to inherit from Tracker objects:
One cannot be a TrackerRecHit if NOT on a TrackerDet!
And please do not even think to use dynamic cast is TOO SLOW!
(and thanks if anybody volunteers to go around all code and make it fully consistent while preserving maximal computational efficiency)

VinInn:
-1

DataFormats/TrackerRecHit2D/interface/SiTrackerMultiRecHit.h
```diff
@@ -34,7 +34,7 @@ class SiTrackerMultiRecHit : public BaseTrackerRecHit {
   // at the momement nobody care of MultiHit!!!
   // used by trackMerger (to be improved)
   OmniClusterRef const& firstClusterRef() const override {
-    return static_cast<BaseTrackerRecHit const*>(&theHits.front())->firstClusterRef();
+    return dynamic_cast<BaseTrackerRecHit const&>(theHits.front()).firstClusterRef();
```
VinInn: this will abort! either test the result or assume is correct by costruction

DataFormats/TrackerRecHit2D/src/ProjectedSiStripRecHit2D.cc
```diff
@@ -5,6 +5,6 @@
...
   TrackingRecHit::setDet(idet);
-  const GluedGeomDet& gdet = static_cast<const GluedGeomDet&>(idet);
+  const GluedGeomDet& gdet = dynamic_cast<const GluedGeomDet&>(idet);
```
VinInn: it cannot be nothing else. It is correct by construction.
This is production code, not a debugging tool

dan131riley:
> -1 (at least for the dynamic_cast part:
> It has be correct at costruction time, full stop. no late test at run time. This is production code
> dynamic cast w/o an alternative is pure debug.

After running some performance tests **and** observing that no further logic errors were found in the integration tests, I agree, the cost is measurable and not justified.  I'll remove them from the PR.  I wouldn't have ever considered adding them except for the observation that the first one **did** find logic errors in what is supposed to be production code.

DataFormats/TrackerRecHit2D/interface/BaseTrackerRecHit.h
dan131riley: The callers for this constructor are all working from lists of generic `GeomDet*` pointers that are assumed to be `TrackerGeomDet*`.  If the type of the argument is changed to 'TrackerGeomDet*', then all the callers have to do a cast, so you end up with even more casts.  I started down that road and eventually concluded that making `GeomDet` usage compile-time type safe would require an enormous refactoring effort.

DataFormats/TrackerRecHit2D/src/ProjectedSiStripRecHit2D.cc
dan131riley: It's a virtual function, changing the argument type would lose polymorphism, and also likely force the callers to do a cast.

DataFormats/TrackerRecHit2D/interface/SiTrackerMultiRecHit.h
dan131riley: I'm reverting this to a static_cast so the subject is moot, but philosophically I disagree.  If a check finds a logic error in code that is supposed to be correct by construction, throwing an exception is a reasonable response.

DataFormats/TrackerRecHit2D/interface/BaseTrackerRecHit.h
VinInn: Welcome to the party!

DataFormats/TrackerRecHit2D/src/ProjectedSiStripRecHit2D.cc
VinInn: right!

DataFormats/TrackerRecHit2D/interface/SiTrackerMultiRecHit.h
VinInn: Well, CMS could finally decide to have two builds: validation )(with debug on) and production (no debug).
As long as  CMS does not switch to such a sound (to me) practice, we must enforce maximal efficiency.
I am also pretty annoyed that upgrade is polluting production code  that is supposed to run in production for Run3. Most probably NONE of this code will ever made it (I hope!) to Phase2,
still upgrade is finding her way in the most old and fragile components of cms software w/o any prior verification of the consequences upstream and downstream.
I sincerely do not know WHO has approved all this MTD code in Tracker and I have missed any design discussion of all this.

kpedro88: MTD had to write a TDR. Subsequently, they have not devoted much time to code maintenance or improvement tasks (currently ~40% of the upgrade issues on GitHub are related to MTD).

If there are issues that aren't being caught by our current tests, more tests should be added.

cvuosalo:
@cmsbuild please test

cvuosalo:
+1

kpedro88:
+upgrade

perrotta:
+1
- The bug described in the PR description is fixed by assigning 
```C++
using MTDGeomDet = TrackerGeomDet;
```
in Geometry/CommonTopologies/interface/MTDGeomDet.h
- Other reco related updates in this PR are mostly cleanings

perrotta:
@dan131riley please remove the mention to the move from static_cast's to dynamic_cast's which is still written in the PR description, for a correct reference

dan131riley:
@perrotta i updated the PR description and resolved the remaining conversation

silviodonato:
+1

Issue #28598, opened 2019-12-10
Problem with HGCalTowerMaps adds


Dr15Jones:
In the slc7_amd64_gcc900 IB, the RelVal 21234.0 step 2 failed with the exception

```
----- Begin Fatal Exception 10-Dec-2019 06:07:37 CET-----------------------
An exception of category 'StdException' occurred while
   [0] Processing  Event run: 1 lumi: 1 event: 3 stream: 0
   [1] Running path 'FEVTDEBUGHLToutput_step'
   [2] Prefetching for module PoolOutputModule/'FEVTDEBUGHLToutput'
   [3] Calling method for module HGCalTowerProducer/'hgcalTowerProducer'
Exception Message:
StdException HGCalTowerMap: Trying to add HGCalTowerMaps with different bins: 2592 and 0
----- End Fatal Exception -------------------------------------------------
```
See

https://cmssdt.cern.ch/SDT/cgi-bin/buildlogs/raw/slc7_amd64_gcc900/CMSSW_11_1_X_2019-12-09-2300/pyRelValMatrixLogs/run/21234.0_TTbar_14TeV+TTbar_14TeV_TuneCUETP8M1_2026D44_GenSimHLBeamSpotFull14+DigiFullTrigger_2026D44+RecoFullGlobal_2026D44+HARVESTFullGlobal_2026D44/step2_TTbar_14TeV+TTbar_14TeV_TuneCUETP8M1_2026D44_GenSimHLBeamSpotFull14+DigiFullTrigger_2026D44+RecoFullGlobal_2026D44+HARVESTFullGlobal_2026D44.log

This only appears to happen for the gcc 9 build. It is possible that it was just a 'random' failure (i.e. could happen in any of the IBs but just happened to be gcc 9 this time).

Dr15Jones:
I was able to reproduce the problem in my local area.

Dr15Jones:
I ran the code in gdb and did a stack trace for the exception throw
```
#1  0x00007f8f2b5c1090 in l1t::HGCalTowerMap::operator+=(l1t::HGCalTowerMap const&) [clone .cold] ()
   from /cvmfs/cms-ib.cern.ch/week0/slc7_amd64_gcc900/cms/cmssw/CMSSW_11_1_X_2019-12-09-2300/lib/slc7_amd64_gcc900/libDataFormatsL1THGCal.so
[...]

PR #28609, opened 2019-12-11
[HGCAL trigger] Remove HFNose from tower maps

jbsauvan:
#### PR description:
Fix crash in HFNose workflows coming from HFNose layers being mixed with HGCAL layers in the tower maps (see #28598).
Simply remove HFNose trigger cells from the tower maps (no HFNose towers have been used or are needed so far).

#### PR validation:
Tested with workflow 21234.0


Dr15Jones:
please test

kpedro88:
+upgrade

fabiocos:
please test for slc7_amd64_gcc900

fabiocos:
@rekovic please have a look

kpedro88:
@fabiocos is the different arch responsible for the comparison differences?

fabiocos:
@kpedro88 I do not see what else might cause this difference, which is anyway quite large (although we likely do not expect bit-by-bit identity). Please notice that the bulk fo the differences is in MC workflows, but not only (so not simply justified by some small numerical difference in the random number generator).

The crash looks solved, let's reverify the comparison before the integration (a cross compiler check goes beyond the scope of this PR)

fabiocos:
please test

rekovic:
+1

fabiocos:
+1

PR #28617, opened 2019-12-12
Make PixelChanneldentifier::thePacking constexpr

Dr15Jones:
#### PR description:

Converted the const static PixelChanneldentifier::thePacking to be constexpr
#### PR validation:

The code compiles.


Dr15Jones:
I happen to be working on a different problem which I thought might have to do with the static. It turned out not to be the case, but I thought this change was good none the less.

Dr15Jones:
please test

DataFormats/SiPixelDetId/interface/PixelChannelIdentifier.h
```diff
@@ -4,29 +4,17 @@
...
-
-  static std::pair<int, int> channelToPixel(int ch) {
-    int row = (ch >> thePacking.column_width) & thePacking.row_mask;
-    int col = ch & thePacking.column_mask;
-    return std::pair<int, int>(row, col);
-  }
-
-  static int pixelToChannel(int row, int col) { return (row << thePacking.column_width) | col; }
-
+namespace pixelchanelidentifierimpl {
```
Dr15Jones: The `Packing` class had to be removed from within `PixelChannelIdentifier` for a technicality about C++ and `constexpr`.  You can not declare a `constexpr` member data where the type is defined within the class. The reason is the class is not _defined_ until its closing `};` and that includes classes defined within it. The last bit is you can't use `constexpr` on an undefined type. Declaring `Packing` outside of `PixelChannelIdentifier` makes use `Packing` is fully defined when the compiler reaches the `constexpr`.

fabiocos:
@civanch this looks a technical update that does not show any regression as expected , I will integrate it, please have anyway a look at it

fabiocos:
+1

fabiocos:
merge

PR #28619, opened 2019-12-12
Use CmsRun python module for Validation/Geometry tests

Dr15Jones:
#### PR description:

The tests had been crashing in future looking IBs because ROOT
was doing cleanup once the ROOT python module is removed. This
caused a problem when using the module within a _cfg.py file as
cmsRun shutsdown python before doing the event processing.

Switching to using the CmsRun python module instead of running the
cmsRun executable avoids the problem as the CmsRun python module
does not shutdown (or spinup) python internally.

#### PR validation:

The tests now pass when run under CMSSW_11_1_ROOT618_X_2019-12-11-2300.

### DIFF
--- a/Validation/Geometry/test/genTrackerPlots.sh
+++ b/Validation/Geometry/test/genTrackerPlots.sh
@@ -9,7 +9,7 @@ mkdir $TEST_DIR && cd $TEST_DIR
 cmsRun ${VGEO_DIR}/test/single_neutrino_cfg.py nEvents=1000 >$TEST_DIR/single_neutrino_cfg.log 2>&1
 
 for geom in {'Extended2015','Extended2017Plan1'}; do
-    cmsRun ${VGEO_DIR}/test/runP_Tracker_cfg.py geom=$geom label=Tracker >$TEST_DIR/runP_Tracker_cfg.log 2>&1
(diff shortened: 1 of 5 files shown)


Dr15Jones:
please test

Dr15Jones:
@davidlange6 @smuzaffar this should fix the Validation/Geometry unit test failures we are seeing in the various IBs. From my thinking, this properly deals with the need of keeping ROOT 'alive'.

cvuosalo:
+1

jfernan2:
+dqm

fabiocos:
+1

fabiocos:
@Dr15Jones this is not a fully general alternative to cmsRun, looking at the code, or am I wrong?

Dr15Jones:
> @Dr15Jones this is not a fully general alternative to cmsRun, looking at the code, or am I wrong?

The CmsRun python module should be capable of running any CMSSW process that the cmsRun executable can. The CmsRun module is not intended to replace cmsRun, just to allow some activities which need to run a CMSSW process but then do other stuff as well (such as the test in this pull request).

fabiocos:
@Dr15Jones thanks for the clarification

PR #28621, opened 2019-12-13
PPS: configurable timing-tracking cut

jan-kaspar:
#### PR description:

This PR allows to configure (in python) the cuts for association between timing-RP and tracking-RP local tracks. Previously, these cuts were effectively hard-coded. This is a follow up of an observation presented e.g. in [Proton POG meeting 26 Nov 2019](https://indico.cern.ch/event/865518/contributions/3648775/attachments/1951600/3240029/kaspar_dpg_timing_eff.pdf) (slide 5), showing that the current cuts a little too strict. In order to maintain the backward compatibility, this PR keeps the default cut values as before.

#### PR validation:

The plot below compares proton reconstruction results for different periods (rows) and different CMSSW versions/settiings (colours). Comparing blue (before this PR) with red dashed (this PR, default cuts) shows that there is no difference in results. Comparing red dashed (default cuts) with green (enlarged cuts) shows that more reco-proton objects include non-zero number of timing-RP tracks.

make_cmp

### DIFF
--- a/RecoCTPPS/ProtonReconstruction/plugins/CTPPSProtonProducer.cc
+++ b/RecoCTPPS/ProtonReconstruction/plugins/CTPPSProtonProducer.cc
@@ -67,6 +67,9 @@ private:
     bool th_y_cut_apply;
     double th_y_cut_value;
 
+    double ti_tr_min;
+    double ti_tr_max;
+
     void load(const edm::ParameterSet &ps) {
       x_cut_apply = ps.getParameter<bool>("x_cut_apply");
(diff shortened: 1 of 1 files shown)


slava77:
@cmsbuild please test

perrotta:
+1
- This PR just defines two additional parameters which allow to configure the timing-tracking cut, now hardcoded
- Default values of those parameters reproduce those previously hardcoded cuts
- Jenkins tests pass

fabiocos:
+1

PR #28562, opened 2019-12-05
Added Service callbacks for Run and Lumi writing

Dr15Jones:
#### PR description:

The new signals are used to announce when starting and ending of the OutputModule write phases occur for Runs and LuminosityBlocks.

#### PR validation:

All framework unit tests pass.

Issue #28521, opened 2019-12-02
edm::Service callback for lumi/run cleanup


schneiml:
Dear framework experts,

while porting the "new DQMStore" work back to work with a `edm::Service` based `DQMStore`, I face a problem related to which callbacks are available to the service.

In the `DQMStore`, we now essentially emulate the product life cycle in edm: histograms are created as a module enters a lumisection/run, kept while that lumi/run has not ended, saved to one or more output files in output modules or EDAnalyzers, and then removed and recycled once the run/lumi is completely processed. One important detail here is that we _whenever possible_ reuse the `MonitorElement`/`TH1` objects: this is to make sure that when there are no concurrent runs/lumis, there is exactly one object per histogram and the legacy modules can keep pointers to that, and continue working correctly.

Now, the problem is that the `DQMStore` has to delete (or rather "mark for recycling") the histograms for a lumisection _after all output modules_ have written the their output. As far as I can see, there is no callback for that.

My current interim implementation cleans up at the beginning of the next lumisection, but that can't work with concurrent lumis. Also it does not work with the `DQMRootSource`, since it wants to read all histograms before triggering the start of the run/lumi. In the past, the cleanup was triggered by the output modules themselves, which is deeply flawed since there might be more than one of them (even in practice, usually `MEtoEDM` and `DQMRootOutputModule` were loaded at the same time).

Find the current WIP here:
```
mkdir -p /tmp/$USER
cd /tmp/$USER
cmsrel CMSSW_11_0_X_2019-11-26-2300
cd CMSSW_11_0_X_2019-11-26-2300/src
cmsenv
git cms-init
git cms-merge-topic schneiml:dqm-prepare-for-stream-v2
git checkout f603f9fd9dda8203fe5387ef7ba75b75c6a594bd
cd DQMServices/
scram b -j12
cd FwkIO/test
LOCAL_TEST_DIR=. LOCAL_TMP_DIR=. ./run_tests.sh
```
`read_file1_file2_cfg.py` fails because the second lumi is read *before* the first one is cleaned up, therefore new objects need to be allocated; after the first lumi is cleaned up, the old pointers are invalid and ` DummyReadDQMStore` crashes.. This could easily be avoided by fixing ` DummyReadDQMStore`, but I believe we should support this (full legacy behavior as long as there are no concurrent runs/lumis). The flawed `DQMStore` logic is `DQMServices/Core/src/DQMStore.cc:506-520`.

[...]

PR #28622, opened 2019-12-13
DQM: new DQMStore.

schneiml:
#### PR description:
This PR depends on many previous PR *and contains them*. Integration will take a while. [Edit: All dependencies are in!. Should be good to go now.]

This PR completely replaces the Core DQM infrastructure. Some things don't change, but are re-implemented. This state was achieved in previous pull requests:
- DQM Histograms are stored and managed by `edm::Service<DQMStore>`.
     - Using a service makes things like online DQM and the DQM@HLT protobuf chain easier, since we can ask the `DQMStore` for a snapshot of the current histograms.
     - Using a service is required to handle job-level products, which we need for multi-run harvesting (and for historical reasons also normal harvesting). 
    - Using a service allows legacy modules to work unchanged.

- In non-legacy modules, all dependencies across the the `DQMStore` are passed to edm via `DQMToken` products, that do not contain any data. (As far as possible -- there are no job products, so the job-level dependencies cannot be expressed.)
    - The tokens are divided into 4 _generations_: `DQMGenerationReco`, `DQMGenerationHarvesting`, `DQMGenerationQTest`, denoted in the instance label.
    - `DQMGenerationReco` is produced by `DQMEDAnalyzer`s, which consume only "normal" (non-DQM) products.
    - `DQMGenerationHarvesting` is produced by `DQMEDHarvester`s, which consume (by default) all `DQMGenerationReco` tokens. This allows all old code to work without explicitly declaring dependencies. `DQMEDHarvester`s provide a mechanism to consume more specific tokens, including `DQMGenerationHarvesting` tokens from other harvesters.
    - `DQMGenerationQTest` is produced only by the `QualityTester` module (of which we typically have many instances). The `QualityTester` has fully configurable dependencies to allow the more complicated setups sometimes required in harvesting, but typically consumes `DQMGenerationHarvesting`.
    - There is a hidden dependency between end run and end job in harvesting: Many harvesters effectively depend on `DQMGenerationQTest` products (the QTest results) and do not specify that, but things still work because they do their work in `endJob`.

- There are six supported types of DQM modules:
    - `DQMEDAnalyzer`, based on `edm::one::EDProducer`. Used for the majority of histogram filling in RECO jobs. Soon to be based on `edm::stream`.
    - `DQMOneEDAnalyzer` based on `edm::one::EDProducer`. Used when begin/end job transitions are required. Can accept more `edm::one` specific options. Cannot save per-lumi histograms.
    - `DQMOneLumiEDAnalyzer` based on `edm::one::EDProducer`. Used when begin/end lumi transitions are needed. Blocks concurrent lumisections.
    - `DQMGlobalEDAnalyzer` based on `edm::global::EDProducer`. Used for DQM@HLT and a few random other things. Cannot save per-lumi histograms (this is a conflict with the fact that HLT _typically_ saves _only_ per lumi histograms, see #28341).
    - `DQMEDHarvester` based on `edm::one::EDProducer`. Used in harvesting jobs to manipulate histograms in lumi, run, and job transitions. 
    - `edm::EDAnalyzer` legacy modules. Can do filling and harvesting. Not safe to use in the presence of concurrent lumisections. Safe for multi-threaded running from the DQM framework side.
- There are four supported file formats for DQM histograms:
    - DQMIO, `TTree` based ROOT files. Reading and writing implemented in EDM input and output modules in `DQMServices/FwkIO`: `DQMRootSource` and `DQMRootOutputModule`.
    - Legacy `TDirectory` based `DQM_*.root` ROOT files. Only write support, implemented in `DQMServices/Core` (see  #28588). Read support dropped in this PR, unneeded since references where removed. Only this format supports saving JOB histograms.
    - ProtoBuf streamer, a.k.a. `fastHadd` files. Implemented by `DQMFileSaverPB`/`DQMFileSaver` (see #11086) and the edm input module `DQMProtobufReader`, in "streamer" format for HLT/DAQ.
    - `MEtoEDM` edm based files. Read and written by the Pool I/O modules, after copying `DQMStore` content to edm products using `MEtoEDMConverter`.

In this PR, the following features are implemented/modified:
- Mode-less `DQMStore`. While there where many different modes before (`enableMultiThread`, `lsBasedMode`, `forceResetOnLumi`, `collate`, etc.) that could be configured in various ways and changed the behavior of the `DQMStore` (sometimes fundamentally, making certain features only work in certain modes), the new DQMStore has only a single mode. 
    - There are two options to control per-lumi saving (`saveByLumi` in the `DQMStore`) and harvesting (`reScope` in `DQMRootSource`). Both can be expressed in terms of _Scope_, see later.
    - Another option, `assertLegacySafe`, only adds assertions to make sure no operations that would be unsafe in the presence of legacy modules sneak in. It does not affect the behaviour.
    - The `verbose` option should not affect the behavior, though it has in the past (if only due to race conditions).
- Data-race free. The new implementation tries very hard to prevent data races even when legacy APIs are used, using fine-grained locking. 
    - This does not mean that there are no race conditions. Whenever histograms are read while they are being filled in a multi-threaded context, the results can be dependent on thread interleaving. But, there should be no undefined behavior on C++ level.
    - For safe manipulation of ROOT objects during booking, a callback interface is provided.
    - Currently, there is still only one global `IBooker` and booking methods cannot run concurrently, but it is easy to change this now (will probably be allowed with the `edm::stream` modules).
- _Global_ and _local_ histograms for concurrent lumisections.
    - In the DQM API, we face the conflict that `MonitorElement` objects are held in the modules (so their life cycle has to match that of the module) but also represent histograms whose life cycle depends the data processed (run and lumi transitions). This caused conflicts since the introduction of multi-threading.
    - The new `DQMStore` resolves this conflict by representing each monitor element using (at least) two objects: A _local_ `MonitorElement`, that follows the module life cycle but does not own data, and a _global_ `MonitorElement` that owns histogram data but does not belong to any module. There may be multiple _local_ MEs for one _global_ ME if multiple modules fill the same histogram (`edm::stream` or even independent modules). There may be multiple _global_ MEs for the same histogram if there are concurrent lumisections.
    - The live cycle of _local_ MEs is driven by callbacks from each of the module base classes (`enterLumi`, `leaveLumi`). For legacy `edm::EDAnalyzer`s, global begin/end run/lumi hooks are used, which only work as long as there are no concurrent lumisections. The _local_ MEs are kept in a set of containers indexed by the `moduleID`, with special value `0` for _all_ legacy modules and special values for `DQMGlobalEDAnalyzer`s, where the local MEs need to match the life cycle of the `runCache` (module id + run number). 
    - The live cycle of _global_ MEs  is driven by the enter/leave lumi calls (indirectly) and the `cleanupLumi` hook  using the edm feature added in #28562,  #28521 . They are kept in a set of containers indexed by run and lumi number. For `RUN` MEs, the lumi number is 0; for `JOB` MEs, run and lumi are zero. The special pair `(0,0)` is also used for _prototypes_: Global MEs that are not currently associated to any run or lumi, but can (and _have to_, for the legacy guarantees) be recycled once a run or lumi starts. 
    - If there are no concurrent lumisections, both _local_ and _global_ MEs live for the entire job and are always connected in the same way, which means all legacy interactions continue to work. `assertLegacySafe` (enabled by default) checks for this condition and crashes the job if it is violated.
- Harvesting controlled by _Scope_.
    - In the old `DQMStore`, the handling of per-job vs. per-run historgrams as well as the handling of per-lumi histograms in online, offline, and harvesting is very confusing.
    - To clarify this situation, we introduce the concept of _Scope_:
        - The _scope_ of a ME can be one of `JOB`, `RUN`, or `LUMI`.
        - The _scope_ defines how long a _global_ ME should be used before it is saved and replaced with a new histogram.
        - The _scope_ must be set during booking.
    - By default, the _scope_ for MEs is `RUN`.
        - Code can explicitly use `IBooker::setScope()` to change the scope to e.g. `LUMI`. This replaces the old `setLumiFlag`.
        - When setting `saveByLumi` option in the `DQMStore`, the default scope changes to `LUMI` for all modules that can support per-lumi saving. It could still be manually overridden in code.
        - In harvesting, the default scope is `JOB`. This works for single-run as well as multi-run harvesting, and emulates the old behavior. Moving to scope `RUN` for non-multi-run harvesting would be cleaner, but requires bigger changes to existing code.
    - When harvesting, we expect histograms to get merged. This merging can be controlled using a single option in `DQMRootSource`: `reScope`. This option sets the _finest allowed scope_ when reading histograms from the input files.
        - When just merging files (e.g. in the DQMIO merge jobs), `reScope` is set to `LUMI`. The scope of MEs is not changed, histograms are only merged if a run is split over multiple files.
        - When harvesting, `reScope` is set to `RUN`. Now, MEs saved with scope `LUMI` will be switched to scope `RUN` and merged. The harvesting modules can observe increasing statistics in the histogram as a run is processed (like in online DQM).
        - For multi-run harvesting, `reScope` is set to `JOB`. Now, even `RUN` histograms are merged. This is the default, since it also works for today's single-run harvesting jobs.
- In-order harvesting.
    - Harvesting jobs are always processed sequentially, like the data was taken: runs and lumisections are processed in increasing order. This is implemented in `DQMRootSource`.
- Thread-safe `MonitorElement`s.
    - There is (effectively) only one type of ME, and it uses locking on a local lock to allow safe access from multiple threads.
    - There are three different namespaces (`reco`, `harvesting`, `legacy`) providing this ME type, however the `reco` version does not expose some non-thread-safe APIs.
    - The `MonitorElement` can _own_ the `MonitorElementData` that holds the actual histogram (that is the case for _global_ MEs), or _share_ it with one ME that owns the data and others that don't (this is the case for _local_ MEs).
    - The `MonitorElementData` does not provide an API. It is always wrapped in a `MonitorElement`.
    - The `MonitorElement` has _no_ state, apart from the pointer to the data and the `is_owned` flag.
        - Actually, it does have some state in the `DQMNet::CoreObject`, which is present in the `MonitorElement` to remain compatible with `DQMNet` for online DQM. The values there (dir, name, flags, qtests) are to be considered cached copies of the "real" values in `MonitorElementData`. The dir/name copy in the `CoreObject` is also used as a set index for the `std::set`s in the `DQMStore`, since it remains available even when the `MonitorElementData` is detached from local MEs (e.g. between lumisections).
    - The `MonitorElement` still allows access to the underlying ROOT object (`getTH1()`), but this is unsafe and should be avoided whenever possible. However, there is no universal replacement yet, and DQM framework code uses direct access to the ROOT object in many places.
- Debugging tools: To understand how a ME is handled, set the `process.DQMStore.trackME = cms.untracked.string("<ME Name>")` option in the config file.
      - The `DQMStore` will then log all life cycle events affecting MEs matching this name. This does not include things done to the ME (like filling -- the `DQMStore` is not involved there), but it does include creation, reset, recycling, and saving of MEs.
      - The matching is a sub-string match on the full path, so it is also possible to watch folders or groups of MEs.
      - For the more difficult cases, it can make sense to put a debug breakpoint (`std::raise(SIGNINT)`) inside `DQMStore::debugTrackME` to inspect the stack when a certain ME is created/modified.
      - The previous functionality of logging the caller for all booking calls also still exists and can be enabled by setting `process.DQMStore.verbose = 4`.

#### PR validation:

Passes the relevant unit tests and also some `runTheMatrix`workflows. 

Known broken:
- Anything with `DQMStore::load()` (should be unused, but there still is code referring to it, related to references. This can be removed, since references are gone for a while now.)
- Standard Configuration files mostly still work, but need clean up. The DQMStore does not accept many options any more. Esp. check around `DQMRootSource` related to single/multi run harvesting.

Other flaws:
- Some parts  of the `MonitorElement` implementation can be moved around/removed.


schneiml:
please test

It would be surprising if this actually passed.

DQM/BeamMonitor/plugins/AlcaBeamMonitor.cc
```diff
@@ -174,6 +173,7 @@ void AlcaBeamMonitor::bookHistograms(DQMStore::IBooker& ibooker, edm::Run const&
     }
   }
   ibooker.setCurrentFolder(monitorName_ + "Service");
+  auto scope = ibooker.setScope(MonitorElementData::Scope::LUMI);
```
Dr15Jones: The better C++ way to do such behavior is to use RAII (Resource Acquisition Is Initialization - resource release is destruction).
```
{
     DQMScopeContext scope{ ibooker, MonitorElementData::Scope::LUMI };
     theValuesContainer_ = ...
}
```
at the end of the block scope, the `scope` object is destroyed and it resets the ibooker scope back to what it started at.

schneiml: This was not exactly how I expected this API to be used, but it turned out to be a common pattern for migrating the existing code. I might add a helper like this.

schneiml:
please test

maybe we make it further now.

DQMServices/Core/interface/MonitorElement.h
```diff
@@ -76,13 +96,37 @@ namespace dqm::impl {
...
+    // private access to the MEData, that is why it lives here.
+    struct MEComparison {
+      using is_transparent = int;  // magic marker to allow C++14 heterogeneous set lookup.
+
+      auto make_tuple(MonitorElement *me) const { return std::make_tuple(me->getPathname(), me->getName()); }
+      auto make_tuple(MonitorElementData::Path const &path) const {
+        return std::make_tuple(path.getDirname(), path.getObjectname());
+      }
+      bool operator()(MonitorElement *left, MonitorElement *right) const {
+        return make_tuple(left) < make_tuple(right);
```
makortel: If I'm not mistaken the `path.getDirname()` and `path.getObjectname()` return `string const&`. As far as I can tell, in this case `std::make_tuple()` seems to return `std::tuple<std::string, std::string>` and thus copies the string objects. I guess from the comment above that these functions are used as comparison functions in set insertion/sorting, and if the number of objects to be inserted/sorted is large, the string copying has a risk of becoming expensive.

By quick test with compiler explorer using `std::reference_wrapper` along
```cpp
  return std::make_tuple(std::reference_wrapper(path.getDirname()), std::reference_wrapper(path.getObjectname()));`
```
seems to avoid string copies (a hand-written two-string comparison might be even more performant).

DQMServices/Core/interface/MonitorElement.h
```diff
@@ -191,21 +244,17 @@ namespace dqm::impl {
...
-      if (!data_.dirname->empty())
-        path += '/';
-      path += data_.objname;
-      return path;
-    }
+    const std::string getFullname() const { return access().key.path_.getFullname(); }
+
+    const edm::LuminosityBlockID getRunLumi() { return access().key.id_; }
+
+    const MonitorElementData::Scope getScope() { return access().key.scope_; }
```
makortel: When returning a copy the `const` has very little meaning.

DQMServices/FwkIO/plugins/DQMRootSource.cc
```diff
@@ -42,204 +35,178 @@
...
-    if (name.find("EventInfo/processedEvents") != std::string::npos) {
-      iElement->Fill(iValue + iElement->getIntValue());
-    } else if (name.find("EventInfo/iEvent") != std::string::npos ||
-               name.find("EventInfo/iLumiSection") != std::string::npos) {
-      if (iValue > iElement->getIntValue()) {
-        iElement->Fill(iValue);
+      for (int i = 1; i <= a1->GetNbins(); ++i) {
+        TString label1 = a1->GetBinLabel(i);
+        TString label2 = a2->GetBinLabel(i);
+        if (label1 != label2) {
```
makortel: String copies seem unnecessary.

DQMServices/Core/interface/MonitorElement.h
schneiml: Thanks for the hint! -- my profiling also suggests that there are still string copies around there, but I did not change anything as long as performance was acceptable. Using `reference_wrapper` might be an easy way to fix that.

DQMServices/FwkIO/plugins/DQMRootSource.cc
schneiml: This code is taken unchanged from the old implementation, and I would not take any bets around root objects. 

`GetBinLabel` returns `const char*`, so making proper string object sounds like a slow but reasonable way to avoid `strcmp` and friends...

fwyzard: Creating string objects just to make a comparison is never "reasonable", and using `TString` outside of ROOT code is almost never reasonable in itself.
 
Use `strcmp` directly, or use `std::string_view" for a higher level interface.

schneiml: Well, Fede added that code back in 2014 (#3173)... but yes, I'll try `string_view` once I get to it.

mmusich:
@schneiml (maybe irrelevant given the generalized failures), but if it helps on this item:

> MEtoEDM not really tested, but should roughly work as before.

the strip gain calibration workflow is for sure broken in the current state of affairs, see failure of unit test 
>I found errors in the following unit tests:
>
>---> test testSSTGainPCL_fromRECO had ERRORS

but also: 

> produceTagFilter -> Return false: Statistics is too low : 0

in this [log file](https://cmssdt.cern.ch/SDT/jenkins-artifacts/pull-request-integration/PR-8cba95/3968/runTheMatrix-results/1001.0_RunMinBias2011A+RunMinBias2011A+TIER0EXP+ALCAEXP+ALCAHARVDSIPIXELCALRUN1+ALCAHARVD1+ALCAHARVD2+ALCAHARVD3+ALCAHARVD4+ALCAHARVD5/step9_RunMinBias2011A+RunMinBias2011A+TIER0EXP+ALCAEXP+ALCAHARVDSIPIXELCALRUN1+ALCAHARVD1+ALCAHARVD2+ALCAHARVD3+ALCAHARVD4+ALCAHARVD5.log).

This generally means the harvester did not manage to read the input histograms.

schneiml:
@mmusich thanks for the hint, I did not get to investigating this test failure yet. I just fixed a rather drastic bug that crashed some of the other workflows...

Re the "not tested": Actually there are a few tests now, but I'd like to add even more (maybe tomorrow). Great to see that there also is a unit test related to PCL.

schneiml:
please test

One commit fixes a bug in `cd` which caused the crash in harvesting in 8.0. I assume it also fixes some of the others (or at least uncovers new bugs).

I also changed the default value of `reScope` to be correct for almost all configurations. I'd have expected that some unit tests fail, but it seems not? Those tests need to be improved then...

schneiml:
please test

Fixed the config syntax error, now things should be better

@mmusich the PCL unit test seems to pass now (locally, we'll see the test results soon), and the resulting `promptCalibConditions.db` is identical to the one in a baseline test. I assume that is a good sign?

mmusich:
> the PCL unit test seems to pass now (locally, we'll see the test results soon), and the resulting promptCalibConditions.db is identical to the one in a baseline test. I assume that is a good sign?

@schneiml indeed. Let's the output of 1001.0 as well.

schneiml:
please test

I fixed the crash in 1330 by disabling the `normalise2DHistogramsToBinArea` in L1T DQM. I don't think it ever did anything useful, and that is why it made the job crashed: The `endJob` runs *after* the run is saved to DQMIO (it was always like that), and modifying the MEs after they are saved does not make much sense. 

The correct way would be to put that code into `endRun` (though that is not ok for `DQMEDAnalyzer`), but I want to achieve identical output, so disabling it is better for now.

There are also some updates to the unit`DQMServices/Demo` unit tests, actually I am not sure if they pass atm. They are being updated also in #28612.

schneiml:
I am slightly surprised by this second unit test failure.

And I am a bit worried for the comparison, I'd not expect it to take more than 12h...

schneiml:
please test

The unit tests should pass now, thanks to ProtoBuf input being implemented again. The `runTheMatrix` output should be a bit better, thanks to the default value of the `reScope` option now being properly set (it has to be `JOB` for all classic harvesting jobs, else the havresting code in `endJob` won't see any histograms any more).

There is a known difference in handling of the efficiency flag, which explains a few thousand missing items in the comparison.

schneiml:
Ok. The failing unit test may be due to the `reScope` now being applied by default, but the FwkIO test assuming that no merging happens (it used to use special `DQMStore` modes for that).

The comparison shows a number of interesting features:
- `SiPixelMonitorClient` still seems to get `cd` wrong in some places: https://tinyurl.com/s4cczgs
- Some formerly empty histograms are now filled. I'd consider this bugs fixed: https://tinyurl.com/s2aoe7y
- Some changes that seem to be normalisation related: https://tinyurl.com/tpwq8q4
- Some rather worrying changes in Muon validation: https://tinyurl.com/txqeg4t
- Some weird effects in Tracking, with 2 missing entries and NaNs: https://tinyurl.com/wytaz9j
- Many harvesting plots are missing. Maybe the they are booked with the wrong scope.

schneiml:
please test

The new commits fix a lot of bugs in harvesting. (Some) summaries and `SiPixelMonitorClient` are still broken, but some other things should work now.

schneiml:
Ok, seems something in the tests is inconsistent again. Probably commited too much or too little of my local changes.

Is there any good reason that neither DQMHistoSizes nor bin-to-bin comparison ran?

slava77:
@schneiml 
it looks like the only reco category code change is in RecoLocalCalo/EcalRecProducers/test/EcalLocalRecoTask.cc. This looks unused.
The edits made there could just go in as is in a separate PR.
After that there will be no need for a reco signature in this large PR.

schneiml:
@slava77 I am not too worried about signatures here for now, first #28552 needs to go in and then this one should have a lot less. Breaking out the changes in fc4ac212477f9803663d48fb7b56cf3abb2c1637 and 10dff9a151ef478fa1571b9a24453e1f1e755e9b would be an option as well, but first I want to get everything working here.

Sorry for all the noise that everybody gets with these stacked PRs.

schneiml:
Needs rebase after #28552, #28407, #28588, #28612 are merged. Should require only a few signatures then.

Also needs more work to fix online DQM, unit tests, and quality tests in harvesting.

schneiml:
Some more work needed here, to improve the unit tests, resolve more comparison differences and address @Dr15Jones comments (regarding using a scope guard for `Scope`). Also some more style improvements may happen once things work.

I will pause work on this until some more dependencies are integrated (#28612 at least).

slava77:
-1

mainly to take it off the active list of PRs for reco signature.
Also, there are merge conflicts and more work planned was declared

schneiml:
rebased to Sunday's IB and fixed the tests.

But it turns out #28588 got merged since then (thanks @silviodonato !) so I'll immediatly rebase again.

schneiml:
Rebased to today'y 11:00 IB (which is not out yet). Fails to compile locally, but I assume that is because I started from yesterday's 23:00 IB.

schneiml:
please test

Stuff still seems to work, but I suspect there is a lot in the comparison to be cleaned up. Also, I still want to address some of the old review comments.

schneiml:
Test failure does not reproduced merged on top of CMSSW_11_1_X_2020-01-14-2300. Priavately produced comparsion here, for a while: https://mschneid.web.cern.ch/mschneid/dqm-histo-comparison-summary.html

perrotta:
please test
(let try launching the central tests with the CMSSW_11_1_X_2020-01-14-2300 IB, and see whether the failures disappear, as in your private ones)

schneiml:
Great, the new, proper comparison is much cleaner than what I produced before. Apparently there was more things broken in the previous test run.

Now, I see some familiar things (SiStrip badmodules, `SiPixelActionExecutor`, some Tracking), let's see how much of that I can fix.

schneiml:
Rebased again for convenience and fixed a few bugs. 

Now, a lot of the remaining differences are histograms that have more entries with this PR than without. One reason (for the Tracking/PUmonitoring plots) is that sometimes multiple modules book the same MEs, and while in the past these where independent histograms (and a random one was written to the output, the rest silently dropped), now all the data ends up in the same histogram. So the PUmonitoring plots (for example) are most likely wrong now, but they where not really "correct" before either.

schneiml:
please test

Some Tracker things should be better now.

schneiml:
So, the biggest chuck of differences is now in wf 1000, where some CSC Alca histograms are booked by legacy modules and consequently not saved on a per-run basis. I'll need to fix that on the subsystem side. 

Other differences are per-lumi histograms, which now get different values due to numerical/rounding changes, and a few summary histograms.

schneiml:
I am mildly surprised. This should be a clean rebase on top of `CMSSW_11_1_X_2020-01-21-2300`, which includes my other PR?

schneiml:
Rebased to master, which is `CMSSW_11_1_X_2020-01-22-1100` at the time. Maybe that works better?

schneiml:
please test

the bot seems happy now, and Muon Alignment  in wf 1000 should work again.

schneiml:
The comparison looks increasingly clean. Inspecting the remaining differences, I see
- `Hcal / RecHitTask / TimingCut / HBHEPartition`: were empty, now filled.
- `Pixel / Clusters / OffTrack`: now double-filled, since two modules book them. Clearly a bug, but in the Pixel Phase0 code and since forever.
- ` Tracking / TrackParameters / PUmonitoring`: multi-filled, booked multiple times by multiple instances of the same analyzer.
- ` HLT / EGM / egHLTTrigSum`: summary broken, e.g. in wf 4.53
- `JetMET` some previously empty plots are filled now. Also numerical differences in `METRate`, due to per-lumi saving on this plot.
- `EcalHitsV / EcalSimHitsValidation` disappeared in in 5.1
- `Hcal / RunInfo / NumberOfEvents` shows more events in 7.3. Needs investigation.
- `DT / EventInfo / DCSContents` missing.
- `SiStrip / EventInfo / DCSContents` missing.
- `Muons / RecoMuonV / MuonTrack` shows a weird graphical difference in 9.0
- `AlCaReco / EventInfo` missing in 1001.
- `HLT / EXO / MET` has more entries in 11634.
- `HLT / JME / PFMETNoMu120` has more entries in 11634.
- 40 new objects in 11634. Not sure which though.

schneiml:
please test

This may be the final round.

schneiml:
+1

The remaining differences are explained a few comments above. All other concerns should be addressed, as far as I remember.

@makortel @fwyzard you left comments before, but Github has lost them and I don't remember what they were about. Feel free to have another look.

DQMServices/Core/interface/MonitorElement.h
schneiml: Ah, found the old comment! I use reference_wrapper now.

Calibration/TkAlCaRecoProducers/python/AlcaSiStripGainsAAGHarvester_cff.py
```diff
@@ -9,9 +9,4 @@
 
 DQMStore = cms.Service("DQMStore")
 
-from DQMServices.Core.DQMEDAnalyzer import DQMEDAnalyzer
```
mmusich: @schneiml why is this being removed?
It has been extremely useful in the past to be able to know the run number when checking the results of the PCL in DQM GUI. Also as far as I can see there are other instances of `DQMEventInfo` in the Harvesting step of other PCL workflows: https://github.com/cms-sw/cmssw/search?l=Python&q=DQMEventInfo
Please clarify or provide a suitable replacement.

Calibration/TkAlCaRecoProducers/test/DQMRootFileReader.cc
```diff
@@ -83,7 +81,6 @@ void DQMRootFileReader::analyze(const edm::Event &iEvent, const edm::EventSetup
   // NOTE: this is here just because we need it after the beginRun of
   // MEtoEDMCoverter which calls a Reset on all MEs.
   dbe->open(filename, false, "", "", DQMStore::OpenRunDirs::StripRunDirs);
-  dbe->showDirStructure();
```
mmusich: It looks like this module isn't doing much now that the call to `showDirStructure` has been removed.
Is there a suitable replacement?

DQMServices/Demo/plugins/DemoNormalDQMEDAnalyzer.cc
```diff
@@ -70,7 +70,7 @@ void DemoNormalDQMEDAnalyzer::bookHistograms(DQMStore::IBooker& ibook,
                                              edm::EventSetup const& iSetup) {
   ibook.setCurrentFolder(folder_);
 
-  example_ = ibook.book1D("EXAMPLE", "Example 1D", 20, 0., 10.);
+  example_ = ibook.book1D("EXAMPLE", "Example 1D", 20, 0., 10., [](TH1*) { std::cout << "booked!\n"; });
```
makortel: How about using MessageLogger?

Calibration/TkAlCaRecoProducers/test/DQMRootFileReader.cc
schneiml: The main point of this module was probably the `dbe->open()`, which loads a file into the `DQMStore`.

However, while the stub is still there, there is currently no code for reading legacy files any more, since there shouldn't be any applications any more:
- A common use in the past were references, but references where removed a while ago.
-  Another potential use might be file format conversions. But converting *from* `TDirectory` ROOT format should never really be required, if anything, we convert *into* that format.

Overall, I don't want to add code for reading this messy format unless there is a really good use case for it.

So, probably this module is just obsolete, but deciding such things is out of scope for this PR (which is giant already).

Calibration/TkAlCaRecoProducers/python/AlcaSiStripGainsAAGHarvester_cff.py
schneiml: You cannot run `DQMEDAnalyzer`s in harvesting. An equivalent `DQMEDHarvester` could be built, many of the things in `DQMEventInfo` don't make sense in harvesting anyways.

Re the other usages: It is technically save to load an run the module, but it will not produce any output. This has worked int the past by accident, but would silently produce incorrect results e.g. in multi-run harvesting. I removed it here after it's output disappeared in the PR comparison, to clarify that the module is no longer useful here.

mmusich: > An equivalent DQMEDHarvester could be built, many of the things in DQMEventInfo don't make sense in harvesting anyways.

can you please provide such implementation?

> Re the other usages: It is technically save to load an run the module, but it will not produce any output. This has worked int the past by accident, but would silently produce incorrect results e.g. in multi-run harvesting. I removed it here after it's output disappeared in the PR comparison, to clarify that the module is no longer useful here.

if it is so, then I think must be removed everywhere else...

DQMServices/Core/interface/DQMStore.h
```diff
@@ -189,440 +420,278 @@ namespace dqm::dqmstoreimpl {
...
+
+      // TODO: review and possibly rename the all methods below:
+      // get MEs that are direct children of full path `path`
+      virtual std::vector<dqm::harvesting::MonitorElement*> getContents(std::string const& path) const;
+
+      // get all elements below full path `path`
+      // we have to discuss semantics here -- are run/lumi ever used?
+      virtual std::vector<dqm::harvesting::MonitorElement*> getAllContents(std::string const& path) const;
+      DQM_DEPRECATED
+      virtual std::vector<dqm::harvesting::MonitorElement*> getAllContents(std::string const& path,
```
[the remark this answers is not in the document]
schneiml: No, in general harvesting needs to modify objects at any time. This is one of the reasons why we don't save MEs in products yet: Immutable MEs require nontrivial changes to lots of harvesting logic.

DQMServices/Core/interface/DQMStore.h
```diff
@@ -189,440 +420,278 @@ namespace dqm::dqmstoreimpl {
...
+      DQM_DEPRECATED
+      bool open(std::string const& filename,
+                bool overwrite = false,
+                std::string const& path = "",
+                std::string const& prepend = "",
+                OpenRunDirs stripdirs = KeepRunDirs,
+                bool fileMustExist = true);
+
+      DQMStore(DQMStore const&) = delete;
+      DQMStore& operator=(DQMStore const&) = delete;
```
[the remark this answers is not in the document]
schneiml: It's a singleton (still `edm::Service`) for now, so that does not really apply.

It might actually be safe to move, but I'll rather just ban it since there is no need for moving.

DQMServices/Core/interface/MonitorElement.h
```diff
@@ -76,25 +96,42 @@ namespace dqm::impl {
...
+    };
+
   protected:
     DQMNet::CoreObject data_;  //< Core object information.
-    // TODO: we only use the ::Value part so far.
-    // Still using the full thing to remain compatible with the new ME implementation.
 
-    std::atomic<MonitorElementData const *> frozen_;    // only set if this ME is in a product already
-    std::atomic<MutableMonitorElementData *> mutable_;  // only set if there is a mutable copy of this ME
+    std::atomic<MutableMonitorElementData *> mutable_;  // only set if this is a mutable copy of this ME
```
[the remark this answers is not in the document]
schneiml: Mostly because it is a left-over from the transparent-copy-on-write logic that lived here for a while (visible in red, was never actually used). As we settled on keeping the MEs in the `DQMStore` rather than products, I ripped it out to simplify things, but did not change the type.

DQMServices/Demo/plugins/DemoNormalDQMEDAnalyzer.cc
schneiml: Oh indeed, that was a very quick test if the new callbacks work...

DQMServices/Core/interface/DQMStore.h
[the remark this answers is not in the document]
makortel: Would the opposite, i.e. not declaring these functions `const`, work either?

Calibration/TkAlCaRecoProducers/python/AlcaSiStripGainsAAGHarvester_cff.py
schneiml: Ok, that should be separate PR, potentially to be merged before. I can give it a try tomorrow.

mmusich: thanks!

DQMServices/Core/interface/DQMStore.h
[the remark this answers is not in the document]
schneiml: Potentially, though I am not sure I support your concern: `getAllContents` is const on the `DQMStore` (which is important: the datastructure in `DQMStore` does not allow concurrent access etc.), while what you do the `ME`s is completely up to you (e.g. they support concurrent filling). There are some evil interconnections, e.g. around the `Key` of the `ME`, which are mostly because this needs to preserve most of the old API.

I'll give this a closer look tomorrow.

slava77:
+1

for https://github.com/cms-sw/cmssw/pull/28622	3da0db4 
- reco is only very minimally affected with somewhat technical modifications in  RecoLocalCalo/EcalRecProducers/test/EcalLocalRecoTask.cc

DQMServices/Core/interface/DQMStore.h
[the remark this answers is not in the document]
makortel: I'm merely thinking about our code rule 4.23 "a `const` member function must not give non-const access to any data directly or indirectly held by the object". (ok, this case might be one of those where one has to compromise, at least in the short term)

> Potentially, though I am not sure I support your concern: `getAllContents` is const on the `DQMStore` (which is important: the datastructure in `DQMStore` does not allow concurrent access etc.)

But I don't quite follow why "datastructure in `DQMStore`" not allowing concurrent access implies that `getAllContents()` is good to be `const`?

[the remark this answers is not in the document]
schneiml: > But I don't quite follow why "datastructure in DQMStore" not allowing concurrent access implies that getAllContents() is good to be const?

It doesn't really, but there is many places where const-ness got very weird in this design, due to the constraint that in event processing, the `DQMStore` must not be mutated while the MEs will be, while in harvesting, both can be modified. I tried hard to get everything to work without const-casts (and almost succeeded), I didn't care too much wat makes sense.

I'll have a closer look if this const-ness can be changed, that depends on how much code is affected by such an API change.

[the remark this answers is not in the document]
makortel: Ok. Given that the situation wrt. "leaking" non-`const` pointers from `const` functions probably was there already, I'm fine with leaving further investigation for later time (also given the size of this PR and further work depending on it). So let's say this is something to think about.

silviodonato:
Missing signatures
alca:   @christopheralanwest  @franzoni  @tlampen  @pohsun  @tocheng 
core:   @Dr15Jones  @smuzaffar  @makortel 
hlt:   @Martin-Grunewald  @fwyzard 
l1:   @benkrikler  @rekovic

silviodonato:
+operations

mmusich:
I think this PR should be deferred until https://github.com/cms-sw/cmssw/pull/28821 is merged and this one is rebased on top of it...

schneiml:
@mmusich yes, I agree. I might also add some minor changes here (following up on Matti's comments), but nothing fundamental should change, so I guess it is still good for people to have a look.

Martin-Grunewald:
+1

schneiml:
please test

rebased to an IB with #28821 in, and cherry-picked six commits out of #28813 which are more about the new `DQMStore` in general. This should address a few of @makortel's comments (though now I remember I forgot the `std::cout` in the demo code...).

makortel:
> (though now I remember I forgot the `std::cout` in the demo code...).

@schneiml Are you planning to update that? (just thinking when to sign, that one would be nice to get fixed, the rest (if any) can easily be addressed later; well maybe even that)

schneiml:
@makortel done, we'll need to rerun the tests eventually but now there are no more requested changes that I am aware of.

makortel:
@cmsbuild, please test

schneiml:
+1

No open issues as far as I remember.

makortel:
+core 

for `DataFormats/Histograms/interface/MonitorElementCollection.h`. 

> * DQMHistoTests: Total failures: 467

Still out of curiosity, does this (and the "DQMHistoSizes" follow expectations?

schneiml:
@makortel Some comparison failures are expected: The majority are cases where MEs of the same name are booked by multiple modules or multiple instances of the same module. Here, the old behavior was to silently drop all but one, while now the entries are merged (because actually there never is more than one histogram).

In `HLT/EGM`, the summary plot is broken, and I don't fully know why; The module is somehow unable to read the HLT config (see https://github.com/cms-sw/cmssw/pull/28622#issuecomment-577644454 for the full stories).

I am somehow worried by the `SiStrip/EventInfo` in wf 1000, because that change does not show up in the bin-to-bin comparison...

slava77:
+1

for https://github.com/cms-sw/cmssw/pull/28622	2f51722 
- changes in reco are the same as in the previous iteration in https://github.com/cms-sw/cmssw/pull/28622	3da0db4, only a somewhat technical modification in RecoLocalCalo/EcalRecProducers/test/EcalLocalRecoTask.cc

silviodonato:
Kind reminder
alca:  @franzoni @pohsun @tlampen @tocheng @christopheralanwest
hlt:  @Martin-Grunewald @fwyzard
l1:  @benkrikler @rekovic

Martin-Grunewald:
+1

silviodonato:
+operations

silviodonato:
alca: @franzoni @pohsun @tlampen @tocheng @christopheralanwest
l1: @benkrikler @rekovic
Could you review and eventually sign, please?

silviodonato:
urgent

pohsun:
+1

silviodonato:
merge
@benkrikler @rekovic please have a look when you have time.
I wanted to integrate this PR today in order to have the IB tests before the ORP meeting of tomorrow.

silviodonato:
@schneiml I think this PR causes a failure in workflow 1361.181 of CMSSW_11_1_X_2020-02-10-2300 (see https://cmssdt.cern.ch/SDT/cgi-bin/logreader/slc7_amd64_gcc820/CMSSW_11_1_X_2020-02-10-2300/pyRelValMatrixLogs/run/1361.181_GluGluHToZZTo4L_M125_Pow_py8_Evt_13UP18ml+GluGluHToZZTo4L_M125_Pow_py8_Evt_13UP18ml+DIGIUP18ml+RECOUP18ml+HARVESTUP18+NANOUP18ml/step3_GluGluHToZZTo4L_M125_Pow_py8_Evt_13UP18ml+GluGluHToZZTo4L_M125_Pow_py8_Evt_13UP18ml+DIGIUP18ml+RECOUP18ml+HARVESTUP18+NANOUP18ml.log#/ ).
Could you have a look please?

schneiml:
@silviodonato is that the only failing WF? The crash is "expected" for anything that attempts to use concurrent lumisections without explicitly saying so in the configuration, so probably the respective configs just need to use the correct option. I'll try to track down what is going on in  1361.181.

silviodonato:
Many tests are ok https://cmssdt.cern.ch/SDT/html/cmssdt-ib/#/ib/CMSSW_11_1_X
but some of them are still running

schneiml:
@silviodonato 
```
$ pwd
/dev/shm/schneiml/newdqmstore/CMSSW_11_1_X_2020-02-10-2300/src
$ runTheMatrix.py -l 1361.181
...
1 1 1 1 1 tests passed, 0 0 0 0 0 failed
```

Seems like I am not running what the buildbot runs. Which additional options do I need to pass?

I can see why that workflow would fail (it does enable concurrent lumis, and it does run DQM/VALIDATION, and even if concurrent lumis are blocked by one-modules the assertion usually fires, but I don't get it now. It could be that it depends on some non-reproducible ordering inside EDM...

I can try to make a blind fix, but it would be better to understand what is going on. Especially since this assertion is there to protect the legacy modules from the effects of concurrent lumisections, and VALIDATION typically runs quite a few legacy `edm::EDAnalyzer`s.

silviodonato:
In ROOT6_X and ROOT620_X I see that also 136.8861 is failing
If this error is due to the running order of the multithreding process, I would try to increase the number of --nConcurrentLumis to increase the probability of failure. I will try and let you know.
Let me cc @smuzaffar @makortel @fabiocos

schneiml:
For 1361.181, I see in my outputs that the magic `#Setup FWK for multithreaded` only appears for step5, which does not run DQM (so fine). The other steps have concurrent lumis enabled in cmsDriver, but not multiple threads. This looks different in the IB result, so I do think I am missing some important options here.

For 136.8861, I am running that for a while now. It seems to have locked up at Event 36, like in the IB test. Debugging now.

silviodonato:
The exact command can be found by clicking on the step number from
https://cmssdt.cern.ch/SDT/html/cmssdt-ib/#/relVal/CMSSW_11_1/2020-02-10-2300?selectedArchs=slc7_amd64_gcc820&selectedFlavors=ROOT620_X&selectedStatus=failed

Eg. step3 of 136.8861
```
cmsDriver.py step3  --conditions auto:run2_data -s RAW2DIGI,L1Reco,RECO,SKIM:ZElectron+SinglePhotonJetPlusHOFilter+EXOMONOPOLE,EI,PAT,ALCA:SiStripCalZeroBias+SiStripCalMinBias+SiStripCalSmallBiasScan+TkAlMinBias+EcalESAlign,DQM:@standardDQMFakeHLT+@miniAODDQM+@L1TEgamma --runUnscheduled  --process reRECO --nStreams 4 --data  --era Run2_2018 --eventcontent RECO,MINIAOD,DQM --hltProcess reHLT --scenario pp --nConcurrentLumis 2 --datatier RECO,MINIAOD,DQMIO --customise Configuration/DataProcessing/RecoTLR.customisePostEra_Run2_2018  --customise Validation/Performance/TimeMemorySummary.customiseWithTimeMemorySummary --prefix '/data/cmsbld/jenkins/workspace/ib-run-relvals/cms-bot/monitor_workflow.py timeout --signal SIGTERM 9000 '  -n 100  --filein  file:step2.root  --fileout file:step3.root  --suffix "-j JobReport3.xml "  --nThreads 4 > step3_RunEGamma2018Dml1+RunEGamma2018Dml1+HLTDR2_2018ml+RECODR2_2018reHLT_skimEGamma_Prompt_L1TEgDQMml+HARVEST2018_L1TEgDQM_Prompt.log  2>&1
```

schneiml:
Ah, I can indeed just run it all manually from the `cmsDriver` commands (which do show `nThreads = 4`). Though I'd really like to be able to reproduce this using `runTheMatrix`. It's fine in this case, but we had bugs inside `runTheMatrix` itself, which are impossible to debug that way. 

Re 136.8861, that turns out to be a deterministic deadlock in the error path of `setBinLabel(...)`. I'll have a PR in a few minutes. I don't really get why it happens at event 36 (the relevant code runs in `bookHistograms`, called from `beginRun`), but I guess edm has good reasons for what it does (maybe there are filters involved?).

smuzaffar:
try using the following ( this is what we run for IB tests)
```
runTheMatrix.py  -l 1361.181  -i all --job-reports -t 4
```

silviodonato:
@schneiml  The workflow 250202.183 ( 250202.183_TTbar_13UP18_RD_test+TTbar_13UP18_RD_test+DIGIPRMXUP18_PU25_RD_test+RECOPRMXUP18_PU25_RD_test+HARVESTUP18_PU25 ) is crashing as well.
I'm still try to understand if this is due to this PR or not.

schneiml:
@silviodonato I think this is the same deadlock, the suspicious module is mentioned there as well

schneiml:
Thanks @smuzaffar , the problem with 1361.181 reproduces now.

@silviodonato #28916 should fix the two timeouts.

schneiml:
For the record: #28920 prevents the crash in 1361.181.

silviodonato:
please test workflow 1361.181

mtosi:
ciao
sorry for coming to this PR only now, but running the usual script of the tracking efficiency monitoring which makes the stack plot of the efficiency per tracking iteration, we realized that from pre2 to pre3 the stack plot does not make much sense (while comparing the single distributions they seem comparable)

in particular, the resulting stack in 11_1_0_pre2
1110pre2_phase2_efficiency_eta_cum
while in 11_1_0_pre3
1110pre3_phase2_efficiency_eta_cum

if I loop on the single distributions I get the integral and the maximum of each histo and of the updated stack
in pre2
h effic 59.3535502967 90
stack 59.3535502967 0.929356098175
h effic 13.3359933533 90
stack 72.6895434856 0.93662327528
h effic 0.400576615764 90
stack 73.0901202559 0.940172374249
h effic 0.0202829334739 90
stack 73.1104031205 0.940172374249
h effic 0.228987028415 90
stack 73.3393899798 0.944059491158
h effic 0.632008572386 90
stack 73.9713985324 0.946399509907
h effic 0.0 90
stack 73.9713985324 0.946399509907
h effic 0.00213345608063 90
stack 73.973532021 0.946399509907

both the integral and the maximum increase
in pre3, instead

h effic 59.2632414947 90
stack 59.2632414947 0.922642588615
h effic 13.3893704087 90
stack 23.2090102354 0.443937152624
h effic 0.430286794202 90
stack 2.78062788786 0.0663475841284
h effic 0.0170889779401 90
stack 0.122144599902 0.004930596333
h effic 0.234343903809 90
stack 0.110274714562 0.00367779331282
h effic 0.642823928472 90
stack 0.105616362159 0.0040756505914
h effic 0.0 90
stack 0.105616360762 0.0040756505914
h effic 0.00245980154432 90
stack 0.033769877245 0.00149156176485

the stack gets smaller !?!?

so you see that in the latter case something gets wrong in the handling of the histo


do you have any suggestions ?
thanks

@fwyzard

mtosi:
I've checked the ROOT version in 11_1_0_pre2 and 11_1_0_pre3, by running
`scram tool info root` (thanks @fwyzard )
CMSSW_11_1_0_pre2 : 6.18.04-bcolbf
CMSSW_11_1_0_pre3 : 6.18.04-bcolbf2

therefore, I think this issue is not related to the ROOT version
@schneiml @silviodonato @slava77 @Dr15Jones @jfernan2 @mmusich : any clue / suggestions ?

jfernan2:
@mtosi which is "the usual script" you are running?

mtosi:
ciao
thanks for the head up !
the script is https://github.com/cms-sw/cmssw/blob/CMSSW_11_1_0_pre2/Validation/RecoTrack/test/publicPlots/plot.py
and in particular the function for making the stacked plot is 
https://github.com/cms-sw/cmssw/blob/CMSSW_11_1_0_pre2/Validation/RecoTrack/test/publicPlots/plot.py#L734
which calls
https://github.com/cms-sw/cmssw/blob/CMSSW_11_1_0_pre2/Validation/RecoTrack/test/publicPlots/plot.py#L187

the printout I quoted in my comment comes from that part, playing w/ the `h` and `self._stack`

schneiml:
Hi Mia,

after running the crystal ball for a while (I haven't really tried executing the script yet), maybe this is related: https://github.com/cms-sw/cmssw/pull/28622/files#diff-8e94a26c1f038018d957101ae048c7c5L228-R275

In theory that should make bare ROOT respect the efficiency flag we had in DQM since a long time.

mtosi:
thanks @schneiml
that was the other PR, indeed 
what should I do ? I'm sorry, but as now it is not clear to me :(
thanks

schneiml:
Try resetting that bit in your Python code. There is a good chance that these plots have the efficiency flag in CMSSW set, and that is saved into the ROOT object now, and that that causes the different behaviour.

Now, if this is actually what we want is not that clear, but for a workaround, un-setting that bit in your plotting code should work (and prove that this is actually the cause, which I am not sure about).

DQMServices/Core/interface/MonitorElement.h
```diff
@@ -69,176 +96,197 @@ namespace dqm::impl {
...
+    bool getLumiFlag() const { return access().key.scope_ == MonitorElementData::Scope::LUMI; }
 
     /// this ME is meant to be an efficiency plot that must not be
     /// normalized when drawn in the DQM GUI.
-    void setEfficiencyFlag() { data_.flags |= DQMNet::DQM_PROP_EFFICIENCY_PLOT; }
-    bool getEfficiencyFlag() { return data_.flags & DQMNet::DQM_PROP_EFFICIENCY_PLOT; }
+    void setEfficiencyFlag() {
+      auto access = this->accessMut();
+      if (access.value.object_)
+        access.value.object_->SetBit(TH1::kIsAverage);
```
schneiml, line 270: @mtosi Looks like Github screwed up that link before. This is the relevant line.

mtosi: thanks, but I'm sorry, I'm still confused
the single histogram used by the stacked plot is an efficiency, right, 
but it comes from
https://github.com/cms-sw/cmssw/blob/CMSSW_11_1_0_pre3/Validation/RecoTrack/python/PostProcessorTracker_cfi.py#L21
which is making use of the `DQMGenericClient`

what am I missing ?
thanks

mtosi: forget, I missed your message above
and indeed, I'm trying to rollback this bit, but I'm not able to :(
how should this be done ?
thanks

schneiml: `th1.SetBit(ROOT.kIsAverage, 0)` or `th1.ResetBit(ROOT.kIsAverage)` would be my guess in Python. If you have trouble finding the constant, you can just use the value directly (`1 << 18`).

mtosi:
ciao
I can confirm that is the source of the problem,
but not knowing how to rollback that setting,
and experimenting that the Clone() clones even such a setting
I had to pass through a tmp histogram which stores all the relevant information (binning, content, style, etc)
it seems to work, but it is not super clean :(
1110pre3_phase2_efficiency_eta_cum
1110pre2_phase2_efficiency_eta_cum

schneiml:
Ok, maybe consider asking in the ROOT forum if this is the expected behaviour.

We could still roll back this change and go back to our own definition of the efficiency flag if there are systematic problems, but that should be pretty quickly (given that we have a sort-of-final 11_1 now and this code is actually used for large-scale productions now).

mtosi:
ciao
about asking ROOT, while looking for info about that setting (I was not aware of it :( ) I think it is working as expected / designed
then, the fact that I cannot unset the TH1 is probably a different thing

why do we move our MEs to make use of that setting is even another point, and you (DQM core) have decided that --I admit my ignorance on the reason :( --
I can image you have done because it supports some needs, even if they are not clear to me --I have to admit--

about having this "feature" in 11_1_0, but a rollback in 11_2 and even 11_1_X I do not see a major problem, because as far as I understood it affects the DQM/Validation only, and I think 11_1_0 will be used for the offline reconstruction of the samples for the HLT TDR

I managed to find a --very ugly-- solution to our issue, therefore I'm not feeling to push towards a rollback,
but I would appreciate if someone could explain to me what we gain w/ this setting, because it is not super clear to me, as now

said that thanks a lot for the support ! it was really helpful in finding the source of the issue

schneiml:
Regarding the "why" of that change: In the past, we had the TH1 and the "efficiency flag" stored separately. That meant anything handling DQM MEs had to manually check the efficiency flag and adapt it's behaviour accordingly (e.g. `hAdd` in harvesting, `fastHadd`, reference/overlay in DQMGUI) to handle our efficiency-stored-in-TH1 MEs. (TProfile always did the right thing by default). 

According to the docs, setting this bit should make ROOT do the right thing by default, simplifying things downstream (also the `flags` field of the MEs is completely unused now, which helps e.g. DQMIO storage, which never (correctly) supported storing flags...). However, I took this decision based on the docs, and have not extensively tested which behaviors of ROOT actually change. 

Your corner case is odd, because you have an efficiency, but due to how tracking works, summing up efficiencies makes sense in your application. I think it is fair that you have to adjust your code here since in the common case (e.g. summing up Lumisections, or runs, or datasets) the new behaviour is correct: 90% efficiency in 7 runs is 90% efficiency, not 630% efficiency. Only in the special case of tracking, 10% efficiency in each of 7 steps is actually 70% efficiency (and not 0.00001%, which would be another common interpretation...).

fwyzard:
There is nothing special about tracking in these plots - the same behaviour applies whenever you have some kind of "iterative" reconstruction, or the like.

For example, we could make the same plots for the HLT muon efficiency (which use 3 iterations based on outside-in, inside-out, inside-out based on L1), etc.

mtosi:
uhmmm, averaging the efficiency through different LSs is different than summing the efficiency from independent categories

schneiml:
> uhmmm, averaging the efficiency through different LSs is different than summing the efficiency from independent categories

Yes, that is the fundamental issue. But ROOT only offers one "hadd" (to my knowlegde?) and that one is predominantly used to assemle complete statistics from partial histograms (to my knowledge, again). IMO we should just make sure that this bit can be easily reset to change the behavior to what you need, and according to the docs that should absolutely work, so this feels like a ROOT issue to me.

mtosi:
can the DQM core handle this, please ?

mtosi:
@pcanal could you please help on this task ?
thanks

fioriNTU:
@mtosi , for what I can understand, this is a buggy behavior in ROOT, and we don't really want to fix a root bug through CMSSW. I think there are multiple ways to solve your issue, the more straightforward I can see is to compute the efficiiencies of the various steps summing up the numerators of the steps before, and then doing a simple Draw("same") with any need of a THStack. This would require minor changes to the script you linked.

davidlange6:
Could someone summarize a reproducer for the bug? Thx

On Jun 29, 2020, at 2:58 PM, fioriNTU <notifications@github.com<mailto:notifications@github.com>> wrote:



@mtosi<https://github.com/mtosi> , for what I can understand, this is a buggy behavior in ROOT, and we don't really want to fix a root bug through CMSSW. I think there are multiple ways to solve your issue, the more straightforward I can see is to compute the efficiiencies of the various steps summing up the numerators of the steps before, and then doing a simple Draw("same") with any need of a THStack. This would require minor changes to the script you linked.

schneiml:
@davidlange6 There is not really a bug, more of a feature.

DQM (since ~forever) has a mechanism called "Efficiency Flag", that is used to mark histograms that are not histograms so DQMGUI can handle them correctly.

This PR sets `kIsAverage` in ROOT if the efficiency flag is set so that ROOT also handles them correctly.

This now breaks a downstream script that relied on `TH1::Add` on a efficiency TH1 being handled like if it was a histogram. 

According to @mtosi it was not that easy to just `TH1::ResetBit(kIsAverage)`, which indicates a bug somewhere on ROOT side.

mtosi:
@schneiml **SORRY**
as I wrote in my previous messages, I was not able to reset the bit, but actually I was not running `TH1::ResetBit(kIsAverage)` !!!!!
now, I remade the check by using it, and I can confirm that it works --as expected--

sorry the noise, I'm going to make the PR for adding to our script the handling of it

davidlange6:
Ok, thanks - sounds like there is no bug here after all.

On Jun 29, 2020, at 3:42 PM, Marcel Schneider <notifications@github.com<mailto:notifications@github.com>> wrote:



@davidlange6<https://github.com/davidlange6> There is not really a bug, more of a feature.

DQM (since ~forever) has a mechanism called "Efficiency Flag", that is used to mark histograms that are not histograms so DQMGUI can handle them correctly.

This PR sets kIsAverage in ROOT if the efficiency flag is set so that ROOT also handles them correctly.

This now breaks a downstream script that relied on TH1::Add on a efficiency TH1 being handled like if it was a histogram.

According to @mtosi<https://github.com/mtosi> it was not that easy to just TH1::ResetBit(kIsAverage), which indicates a bug somewhere on ROOT side.

PR #28623, opened 2019-12-13
Prevent segfault if dd4hep::Solid is not valid

cvuosalo:
#### PR description:

A test script for reading geometry XML files terminated with a segmentation fault when it created an incomplete `dd4hep::Solid` object and then tried to output the `Solid::name()` in a debug statement. This `dd4hep::Solid::name()` method tries to de-reference a null pointer, causing a segfault, when the object is invalid.
To fix this problem, a call to `Solid::isValid()` is added to ensure that the object is valid before trying to get its name.
The DD4hep team will be changing the behavior of `Solid::name()` to raising an exception instead of segfaulting in the case of an invalid object, but the approach in this PR of checking for validity is our preferred solution to the problem

#### PR validation:

The script with the segfault was tested with this fix to ensure that the bug is gone. This fix should not change any standard workflows.

No backport needed.

### DIFF
--- a/DetectorDescription/DDCMS/src/DDNamespace.cc
+++ b/DetectorDescription/DDCMS/src/DDNamespace.cc
@@ -153,11 +153,14 @@ dd4hep::Volume DDNamespace::addVolumeNS(dd4hep::Volume vol) const {
   dd4hep::Material m = vol.material();
   vol->SetName(n.c_str());
   m_context->volumes[n] = vol;
+  const char* solidName = "Invalid solid";
+  if (s.isValid())         // Protect against seg fault
+    solidName = s.name();  // If Solid is not valid, s.name() will seg fault.
   dd4hep::printout(m_context->debug_volumes ? dd4hep::ALWAYS : dd4hep::DEBUG,
                    "DD4CMS",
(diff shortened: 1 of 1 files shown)


cvuosalo:
@cmsbuild please test

cvuosalo:
+1

DetectorDescription/DDCMS/src/DDNamespace.cc
```diff
@@ -153,11 +153,14 @@ dd4hep::Volume DDNamespace::addVolumeNS(dd4hep::Volume vol) const {
   dd4hep::Material m = vol.material();
   vol->SetName(n.c_str());
   m_context->volumes[n] = vol;
+  const char* solidName = "Invalid solid";
```
fabiocos: @cvuosalo our guidelines suggest to prefer ```string```  to ```char*```:
```
Use string or string_view, not char * (string).
```

fabiocos: ok, ```s.name()``` is a ```char*``` as well...

fabiocos:
+1

cvuosalo:
@slomeo FYI, the segmentation fault in the script `testGeometry2026D49.py` is fixed by this PR, which has just been merged.

PR #28626, opened 2019-12-15
[Geant] Updates needed for Geant 10.6

smuzaffar:
New Geant4 10.6 complains about ( https://github.com/cms-sw/cmsdist/pull/5424#issuecomment-565573822 )  
```
error: comparison of integer expressions of different signedness: 'int' and 'size_t' {aka 'long unsigned int'} [-Werror=sign-compare]
```

This PR uses the correct type `size_t` to fix this warnings

### DIFF
--- a/SimG4Core/PrintGeomInfo/src/PrintGeomInfoAction.cc
+++ b/SimG4Core/PrintGeomInfo/src/PrintGeomInfoAction.cc
@@ -173,7 +173,7 @@ void PrintGeomInfoAction::dumpG4LVLeaf(G4LogicalVolume *lv,
   //--- If a volume is placed n types as daughter of this LV, it should only be counted once
   std::map<G4LogicalVolume *, unsigned int> lvCount;
   std::map<G4LogicalVolume *, unsigned int>::const_iterator cite;
-  for (int ii = 0; ii < lv->GetNoDaughters(); ii++) {
+  for (size_t ii = 0; ii < lv->GetNoDaughters(); ii++) {
     cite = lvCount.find(lv->GetDaughter(ii)->GetLogicalVolume());
     if (cite != lvCount.end())
       lvCount[cite->first] = (cite->second) + 1;
@@ -192,8 +192,7 @@ int PrintGeomInfoAction::countNoTouchables() {
 }
 
 void PrintGeomInfoAction::add1touchable(G4LogicalVolume *lv, int &nTouch) {
-  int siz = lv->GetNoDaughters();
-  for (int ii = 0; ii < siz; ii++)
+  for (size_t ii = 0; ii < lv->GetNoDaughters(); ii++)
     add1touchable(lv->GetDaughter(ii)->GetLogicalVolume(), ++nTouch);
 }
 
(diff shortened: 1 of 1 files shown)


civanch:
+1

SimG4Core/PrintGeomInfo/src/PrintGeomInfoAction.cc
```diff
@@ -173,7 +173,7 @@ void PrintGeomInfoAction::dumpG4LVLeaf(G4LogicalVolume *lv,
   //--- If a volume is placed n types as daughter of this LV, it should only be counted once
   std::map<G4LogicalVolume *, unsigned int> lvCount;
   std::map<G4LogicalVolume *, unsigned int>::const_iterator cite;
-  for (int ii = 0; ii < lv->GetNoDaughters(); ii++) {
+  for (size_t ii = 0; ii < lv->GetNoDaughters(); ii++) {
```
VinInn: as discussed elsewhere I prefer "int" to unsigned if operations (such as --) are involved.
So in my opinion this one should be modified using the pattern in the other below and those left unchanged or all modified as
```
for (int ii = 0, siz = lv->GetNoDaughters(); ii < siz;  ++ii)
```

civanch: @VinInn , this code itself is rarely used  for tests only, so here we only remove compilation warnings. However, can you, please, clarify, what you mean:

  - it is more correct semantics?
  - potential problem with --i?
  - or potentially more performant code with int instead of size_t?

This may useful for other classes where we use such loops frequently.

civanch:
@smuzaffar , what I see as a default in recent IB: 
CLHEP_BASE=/cvmfs/cms-ib.cern.ch/nweek-02607/slc7_amd64_gcc820/external/clhep/2.4.0.0-nmpfii

However, this PR has compilation problems:
Exception Message:
unable to load /data/cmsbld/jenkins/workspace/ib-run-pr-tests/CMSSW_11_1_GEANT4_X_2019-12-12-2300/tmp/slc7_amd64_gcc820/src/Validation/HGCalValidation/test/testValidationHGCalValidation/libtestValidationHGCalValidation.so because libCLHEP-2.4.1.2.so: cannot open shared object file: No such file or directory

but in Geant4 download page it is required to use 2.4.1.3 

I do not think it is a strong requirement but it seems that clhep use is not coherent.

Another moment: comparisons between signed and unsigned is frequent - I see in several SIM sub-packages. Not only in this place. I will try to make PR with fixes.

smuzaffar:
@civanch , we have included clhep 2.4.1.3 for new geant4. The errors we are getting here are due to this `int` vs `size_t` comparisons https://cmssdt.cern.ch/SDT/jenkins-artifacts/pull-request-integration/PR-6720ce/3984/build-logs/ . We need cmssw changes to go with new geant4. @mrodozov  is working on it

civanch:
OK, I will wait for @mrodozov PR(s) and not propose mine.

SimG4Core/PrintGeomInfo/src/PrintGeomInfoAction.cc
VinInn: there was a thread with David at some point.
In general unsigned should not be used for arithmetics and not even for size: alo the C++ committee now agree that was a mistake to use unsigned for sizes.
Definetively -- can be a problem as it DOES NOT become negative
unsigned shall be used only for bit manipulation or when "wrapping" is intended (modular arithmetics)
Everybody agrees in particular that the index of a loop SHALL be a signed integer

smuzaffar:
please test

smuzaffar:
closing it in favor of https://github.com/cms-sw/cmssw/pull/28637

PR #28638, opened 2019-12-17
APE estimation tool: Enable submission of skims to condor, rearrange skimming tool, add phase 0 geometry support for plotting tools.

mteroerd:
#### PR description:

In the past, multiple skims had to be done quickly for multiple APE measurements on different data sets (for example during ReReco campaigns with many IOVs). When starting all skims simultaneously, too much CPU time was used on lxplus, resulting in some skims being killed. Therefore, submission to condor is now possible. There is also an option to submit to the CAF queue. Skimming-related scripts were moved from /test/batch to /test/SkimProducer and old, unused bash scripts were removed.

Additionally, some additional code was added so that GEN-SIM-RECO samples can be skimmed out of the box in case that should be needed again.

Also, support for the phase 0 geometry was added to the plotting tools for compatibility.


christopheralanwest:
please test

Alignment/APEEstimation/test/SkimProducer/skimProducer_cfg.py
```diff
@@ -178,13 +182,13 @@
...
-    #process.GlobalTag = GlobalTag(process.GlobalTag, '101X_dataRun2_Prompt_v11', '')
-elif isMc:
-    #process.GlobalTag = GlobalTag(process.GlobalTag, 'auto:phase1_2018_realistic', '')
-    process.GlobalTag = GlobalTag(process.GlobalTag, 'auto:phase1_2018_design', '')
-
+if globalTag == None:
+    if isData:
+        process.GlobalTag = GlobalTag(process.GlobalTag, 'auto:run2_data', '')
+    elif isMc:
+        process.GlobalTag = GlobalTag(process.GlobalTag, 'auto:run2_mc', '')
```
christopheralanwest: Is `auto:run2_mc` ("post-VFP" 2016 MC) a sensible default? It seems that for MC it doesn't make sense to specify a default global tag, and should always be specified explicitly.

mteroerd: Indeed, the global tag to use changes per skim. This option is there to allow changing the global tag centrally and was kept from previous versions of the tool. auto:run2_mc is used here only because it was the last used option for a post-VFP APE measurement. This part of the code is probably not necessary anymore.

mteroerd: @christopheralanwest I removed that part from the code now.

tocheng:
@cmsbuild please test

silviodonato:
@mteroerd please update the title of the PR adding the other changes you are proposing

tocheng:
+1

silviodonato:
+1

PR #28648, opened 2019-12-18
Support of plot-specific input paramters for Payload Inspector

ggovi:
#### PR description:

The developer level interface of the Payload Inspector has been extended to support plot-specific input parameters

#### PR validation:

The new functionality has been tested with the test options of the getPayloadData script
More validation is expected after the deployment on the development service

### DIFF
--- a/CondCore/Utilities/interface/PayloadInspector.h
+++ b/CondCore/Utilities/interface/PayloadInspector.h
@@ -11,6 +11,8 @@
 #include <boost/uuid/uuid.hpp>
 #include <boost/uuid/uuid_generators.hpp>
 #include <boost/uuid/uuid_io.hpp>
+#include <boost/python/list.hpp>
+#include <boost/python/dict.hpp>
 
 namespace cond {
 
@@ -159,6 +161,12 @@ namespace cond {
       // required in the browser
       bool isTwoTags() const;
 
+      // required in the browser
+      boost::python::list inputParams() const;
+
+      // required in the browser
+      void setInputParamValues( const boost::python::dict& values );
+
       // returns the json file with the plot data
       std::string data() const;
 
@@ -189,6 +197,8 @@ namespace cond {
 
       void setTwoTags(bool flag);
 
+      void addInputParam( const std::string& paramName );
+
       // access to the fetch function of the configured reader, to be used in the processData implementations
       template <typename PayloadType>
       std::shared_ptr<PayloadType> fetchPayload(const cond::Hash& payloadHash) {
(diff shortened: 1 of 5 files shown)


ggovi:
please test

ggovi:
+1

fabiocos:
@davidlange6 you have invested wor in the past to get rid of some boost dependencies within CMSSW, here we have a further addition of boost dependency (ok, the DB area is entirely dependent on boost...)
@smuzaffar @Dr15Jones @makortel Do we have a clear definition of a policy wrt boost?

davidlange6:
Hi Fabio-

The guidance I’d suggest is if the boost component has been implemented in the C++ standard we are using in CMSSW, then that standard implementation should be used rather than the boost one.

This PR is adding more boost python, which indeed we should replace in conditions code with pybind11  and then drop from our dependency tree (so separate from the advice above) - I think this replacement is not completely trivial to do (having tried it once). But this is a migration that needs to start elsewhere before this PR could be changed (I believe)



On Dec 22, 2019, at 4:46 PM, Fabio Cossutti <notifications@github.com<mailto:notifications@github.com>> wrote:


@davidlange6<https://github.com/davidlange6> you have invested wor in the past to get rid of some boost dependencies within CMSSW, here we have a further addition of boost dependency (ok, the DB area is entirely dependent on boost...)
@smuzaffar<https://github.com/smuzaffar> @Dr15Jones<https://github.com/Dr15Jones> @makortel<https://github.com/makortel> Do we have a clear definition of a policy wrt boost?

mmusich:
Hello all,
in the light of the https://github.com/cms-sw/cmssw/pull/28648#issuecomment-568254426 may I ask what is the foreseen trajectory of this PR?
It seems to me that more work is needed to move with the migration described above, and in the meanwhile this addition would solve the long standing issue [CMSDBBROWS-82](https://its.cern.ch/jira/projects/CMSDBBROWS/issues/CMSDBBROWS-82). Additionally there is already a new line of development based on the changes proposed here which is waiting this PR to be merged.

silviodonato:
@ggovi

silviodonato:
hold

mmusich:
@silviodonato for my education what holds this PR?

silviodonato:
@mmusich @ggovi I was expecting a confirmation that this PR should go in without waiting for the migration to pybind11 ( https://github.com/cms-sw/cmssw/pull/28648#issuecomment-568254426 )

ggovi:
@silviodonato I'm not sure who is supposed to provide this confirmation. Fabio/David?

silviodonato:
As far as I understood @davidlange6 agreed that the migration is complicated and should be made in a different PR. So I think we can proceed with merging this PR in the next IB.

silviodonato:
unhold

silviodonato:
+1

PR #28656, opened 2019-12-19
Remove track jets from AOD

rappoccio:
#### PR description:

AK4 track jets and their constituents are removed from AOD. 

#### PR validation:

I ran runTheMatrix workflows and this is dropping the expected content from AOD.


rappoccio:
Backport is not needed, removing this PR.

PR #28539, opened 2019-12-04
Run3-hcx244 Switch of debug prints

bsunanda:
#### PR description:

Switch off debug prints

#### PR validation:

Standard tests performed

#### if this PR is a backport please specify the original PR:

It is strictly not back porting. But the debug prints will be off in the master thread as well

PR #28506, opened 2019-11-29
Run3-hcx243 Second step to make HCAL parameters to be read in DDD/DD4Hep

bsunanda:
#### PR description:

Second step to read Hcal parameters using DDD as well as DD4Hep

#### PR validation:

Use the standard runTheMatrix.py

#### if this PR is a backport please specify the original PR:

Nothing special

PR #28657, opened 2019-12-19
Remove track jets from AOD

rappoccio:
#### PR description:

This PR removes AK4 track jets and the track p4 candidates used for them from AOD. 

#### PR validation:

Ran matrix workflows, the output of AOD drops these as expected. 

No backport needed.

### DIFF
--- a/RecoJets/Configuration/python/RecoJets_EventContent_cff.py
+++ b/RecoJets/Configuration/python/RecoJets_EventContent_cff.py
@@ -101,8 +101,6 @@ RecoJetsAOD = cms.PSet(
                                            'keep *_cmsTopTagPFJetsCHS_*_*',
                                            'keep *_ak4PFJets_*_*',
                                            'keep *_JetPlusTrackZSPCorJetAntiKt4_*_*',    
-                                           'keep *_ak4TrackJets_*_*',
-                                           'keep recoRecoChargedRefCandidates_trackRefsForJets_*_*',                                             
                                            'keep *_caloTowers_*_*', 
                                            'keep *_CastorTowerReco_*_*',                                           
                                            'keep *_ak4JetTracksAssociatorAtVertex_*_*',


Geometry/HcalCommonData/src/CaloDDDSimulationConstants.cc
```diff
@@ -1,7 +1,7 @@
 #include "Geometry/HcalCommonData/interface/CaloDDDSimulationConstants.h"
 #include "FWCore/MessageLogger/interface/MessageLogger.h"
 
-#define EDM_ML_DEBUG
+//#define EDM_ML_DEBUG
```
slava77: I'm curious how this change made it to this PR.
@rappoccio please clarify

rappoccio: Goood question. I rebased to 11_1_X and apparently this got picked up.

slava77: @bsunanda @cvuosalo @fabiocos 
please clarify if this more recent commit/change is actually present in 11_0_X, but is not in 11_1_X.

https://github.com/cms-sw/cmssw/pull/28539#issuecomment-562222682 refers to #28506 as the master version for the change, but that PR did not have the change at least when it was merged (there was a similar HcalParametersFromDD.cc)

slava77:
@rappoccio 
I'd rather you re-"rebase" without this unrelated commit.

slava77:
> @rappoccio
> I'd rather you re-"rebase" without this unrelated commit.

smth like ```git reset --hard HEAD~2; git cherry-pick b683e8a ``` should do it

rappoccio:
OK should be fixed. Sorry about that.

slava77:
@cmsbuild please test

Geometry/HcalCommonData/src/CaloDDDSimulationConstants.cc
cvuosalo: It looks like one file was left out of #28506. Maybe it can be fixed in @bsunanda 's next PR.

slava77:
+1

for https://github.com/cms-sw/cmssw/pull/28657	0808eb6 
- this is in line with the earlier discussion with JME; the main target in the quest for reducing the AOD size is the p4 candidate collection, which is about 4.3% of the AODSIM in PU50 ttbar events (the track jets alone are an extra 0.2%). The dropped data can be reproduced from AOD, it is not used in the miniAOD workflows and will not trigger any downstream CPU use increase in the production workflows.
- jenkins tests pass

Geometry/HcalCommonData/src/CaloDDDSimulationConstants.cc
fabiocos: @cvuosalo @slava77 @rappoccio @bsunanda this is a useful fix that apparently was missing in master (and I overlooked this when integrating its backport) . Ok to avoid mixing it here, I'll fix it separately

fabiocos:
+1

PR #28659, opened 2019-12-19
PUPPI in RECO

rappoccio:
#### PR description:

- Adding PUPPI weight to PFCandidate. 
- Adding AK4Puppi and AK8PFPuppi to RECO by default. 
- Removing AK8CHS. 
- Removing CMSTopTagger.

This will add two floats (standard and no-lep puppi) to PFCandidate. It becomes the "final" particleFlow module after linking. There are therefore now two "particleFlowTmp" objects (Tmp and Tmp2). 

This is the first in an expected long sequence of follow-on PRs that will make use of PUPPI in RECO, simplifying many downstream workflows. 

#### PR validation:

```
runTheMatrix.py -l limited
```

The expected PUPPI weights are present in the PFCandidates and are used (for now) only for ak4PFPuppi and ak8PFPuppi. 

No backports are planned.


slava77:
@cmsbuild please test

jfernan2:
@rappoccio  As expected this PR introduces changes in DQM for JetMET and Physics (and minor ones in Trigger). Removals from CMSTopTagger and AK8CHS are evident.

A general change is that for Physics B2G and Exotica lose 65 and 14 plots resp., which were mainly empty, but gain 50 and 14 empty plots; hence there seems to be just a renaming in the case of B2G but plots remain empty still.

JetMET has modifications but this results in many histograms becoming empty like many observables on JetMET/Jet/CleanedslimmedJetsAK8  or  JetMET/JetValidation/slimmedJetsPuppi. Nevertheless, the only change in code for the JetAnalyzer is not filling the mPrunedMass* histograms, so I am not able to explain this.

HLT gets minor modifications due to change in collection to monitor. All fine I believe.

So, the number of modifications creating empty histograms makes me wonder if the change to Puppi as default options has been completely propagated to the DQM modules. Do we understand this? Thanks

slava77:
@cmsbuild please test

to refresh 2+weeks old tests.

slava77:
@rappoccio 
please comment on the status of addressing the failing unit test
testTauEmbeddingProducers.
The error there is apparently related to this PR changes.

rappoccio:
Hi, @slava77 

Sorry for the delay, I didn't get to this before the holiday break. I will get to it this week.

Cheers,
Sal

CommonTools/PileupAlgos/plugins/PuppiProducer.cc
```diff
@@ -288,9 +294,17 @@ void PuppiProducer::produce(edm::Event& iEvent, const edm::EventSetup& iSetup) {
...
-      pfCand->setP4(puppiP4s.back());
+      pfCandWeighted->setP4(puppiP4s.back());
+      pfCandWeighted->setSourceCandidatePtr(aCand.sourceCandidatePtr(0));
+      pfCand->setP4(aCand.p4());
       pfCand->setSourceCandidatePtr(aCand.sourceCandidatePtr(0));
+      if (fPuppiForLeptons) {
+        pfCandWeighted->setPuppiWeight(1.0, lWeights[iCand]);
+      } else {
+        pfCandWeighted->setPuppiWeight(lWeights[iCand]);
+      }
```
ahinzmann: I don't quite understand the logic behind this.
My understanding from the code is that the "PFCandidates_puppi__*" collection would contain the original PF candidates and the PUPPI weights, while the "PFCandidates_puppi_weighted_*" collection would contain the scaled candidates and is not used anywhere after the full transition of all downstream code.

From this I would expect here
pfCand->setPuppiWeight
rather than
pfCandWeighted->setPuppiWeight

Did I misunderstand?

rappoccio: Shoot, you're right. I will fix it.

slava77:
```particleFlow_{muons,electrons,photons}``` collections (valueMaps) are missing now. 
It looks like they are actually not used in the tested workflows. I did not find any examples for muons and photons (```git grep  "particleFlow[^$]photons"```). The electrons have a mention in the release via patElectrons, but in the end it looks like miniAOD is overriding that inputTag to use another (reducedEGamma) collection and in the plain AOD->PAT it looks like the default is to use the original full PFCandidates particleFlow.

@bendavid @hatakeyamak do you know if there are actual use cases?
Can these be fully recovered from the AOD collections (based on particle ID and refs in the particleFlow PFCandidates)?
If there are no active use cases and the collections can be recovered, we could just drop them altogether (perhaps after a recovery module is implemented).
@rappoccio  For now, I propose to keep the intermediate Tmp2 collections (and update CommonTools/ParticleFlow/test/pfIsolation_cfg.py and PhysicsTools/PatAlgos/python/producersLayer1/electronProducer_cfi.py where there are hints of potential use).

Apart from this specific case for the somewhat less used collection, @bendavid @hatakeyamak please check the updates in the PF reco introduced in this PR and let me know if these look OK.
Thank you.

**Comment by slava77** [comment]
@rappoccio 
it looks like the packedPFCandidates puppiWeightNoLep is now always set to 1.0, which is most visible in the plots of neutrals (e.g. a reminiAOD https://tinyurl.com/yh26qcsz vs https://tinyurl.com/yj296p79 after)
I suppose these should be fixed; it's not obvious to me if this is related to https://github.com/cms-sw/cmssw/pull/28659#discussion_r363733477

**Comment by ahinzmann** [comment]
Currently only the PUPPI-with-lepton weight (for jets) is calculated and stored at AOD. To avoid confusion (and possibly solve the issue Slava reports about packedPFCandidate puppiWeightNoLep), it may be better to compute PUPPI-no-lepton (for MET) in AOD as well.

File: CommonTools/PileupAlgos/plugins/PuppiProducer.cc (hunk 1 above)
Comment (ahinzmann) [reply CommonTools/PileupAlgos/plugins/PuppiProducer.cc]: And also this should keep the "other" weight, like a few lines before in the code, e.g.
pCand->setPuppiWeight(pCand->puppiWeight(), lWeights[iCand]);

File: RecoJets/JetProducers/plugins/VirtualJetProducer.cc [hunk 2]
```diff
@@ -456,9 +458,18 @@ void VirtualJetProducer::inputTowers() {
 	std::cout << "PF cand:" << pfc << '\n';
       }
       */
-      fjInputs_.emplace_back(input.px(), input.py(), input.pz(), input.energy());
+      if (!applyPuppiWeight_) {
+        fjInputs_.emplace_back(input.px(), input.py(), input.pz(), input.energy());
+        fjInputs_.back().set_user_index(i - inBegin);
+      } else {
+        reco::PFCandidate const* pPF = dynamic_cast<reco::PFCandidate const*>(i->get());
```
slava77: it makes sense to me to also support the packedPF here

RecoJets/JetProducers/python/ak4PFJets_cfi.py
```diff
@@ -17,7 +17,8 @@
     )
 
 ak4PFJetsPuppi = ak4PFJets.clone(
-    src = cms.InputTag("puppi")
+    src = cms.InputTag("particleFlow"),
+    applyPuppiWeight=cms.bool(True)
```
slava77: please drop type specifications for all parameters which already exist.

RecoJets/JetProducers/python/ak8PFJets_cfi.py
```diff
@@ -16,7 +16,8 @@
 # AK8 jets with various pileup subtraction schemes
 ##############################################################################
 ak8PFJetsPuppi = ak8PFJets.clone(
-    src = cms.InputTag("puppi")
+    src = cms.InputTag("particleFlow"),
+    applyPuppiWeight = cms.bool(True)
```
slava77: please drop type specifications for all parameters which already exist.

RecoParticleFlow/Configuration/python/RecoParticleFlow_cff.py
```diff
@@ -39,9 +42,15 @@
...
-particleFlowLinksTask = cms.Task( particleFlow, particleFlowPtrs, chargedHadronPFTrackIsolation, particleBasedIsolationTask)
+# SRR 17-Dec-2019: Add puppi weights by default to PF
+particleFlow = puppi.clone(candName='particleFlowTmp2')
+
+particleFlowLinksTask = cms.Task( particleFlowTmp2, particleFlow, particleFlowPtrs, chargedHadronPFTrackIsolation, particleBasedIsolationTask)
 particleFlowLinks = cms.Sequence(particleFlowLinksTask)
 
+
+
+
```
slava77: do we need these 3 empty lines?

RecoParticleFlow/PFProducer/python/pfLinker_cff.py
```diff
@@ -2,10 +2,10 @@
 
 import RecoParticleFlow.PFProducer.pfLinker_cfi
 import RecoParticleFlow.PFProducer.particleFlowTmpPtrs_cfi
-particleFlow = RecoParticleFlow.PFProducer.pfLinker_cfi.pfLinker.clone()
-particleFlow.PFCandidate = [cms.InputTag("particleFlowTmp")]
+particleFlowTmp2 = RecoParticleFlow.PFProducer.pfLinker_cfi.pfLinker.clone()
+particleFlowTmp2.PFCandidate = [cms.InputTag("particleFlowTmp")]
 particleFlowPtrs = RecoParticleFlow.PFProducer.particleFlowTmpPtrs_cfi.particleFlowTmpPtrs.clone()
-particleFlowPtrs.src = "particleFlow"
+particleFlowPtrs.src = "particleFlow"  # SRR 17-Dec-2019: 'particleFlow' will be post-puppi now.
```
slava77: I think that this should be moved downstream after ```particleFlow``` is known

RecoParticleFlow/PFProducer/python/pfLinker_cff.py
```diff
@@ -2,10 +2,10 @@
 
 import RecoParticleFlow.PFProducer.pfLinker_cfi
 import RecoParticleFlow.PFProducer.particleFlowTmpPtrs_cfi
-particleFlow = RecoParticleFlow.PFProducer.pfLinker_cfi.pfLinker.clone()
-particleFlow.PFCandidate = [cms.InputTag("particleFlowTmp")]
+particleFlowTmp2 = RecoParticleFlow.PFProducer.pfLinker_cfi.pfLinker.clone()
+particleFlowTmp2.PFCandidate = [cms.InputTag("particleFlowTmp")]
```
slava77: "Tmp" is already confusing enough.
Since this is is new, it could be a good opportunity to name this more explicitly e.g. ```particleFlowLinked```.
Another option is to regress in the name wrt the standard particleFlow and say ```particleFlowPrePuppi```; it looks more like something incomplete.

slava77 (commented):
In addition to the inline comments:
it looks like Puppi_cff can now also have ```useExistingWeights = True```, to acknowledge the new default. This will be set to False in the ```particleFlow``` clone.

CommonTools/PileupAlgos/plugins/PuppiProducer.cc
rappoccio: That would keep the existing puppi weight, which is handled elsewhere, right?

rappoccio (commented):
> Currently only the PUPPI-with-lepton weight (for jets) is calculated and stored at AOD. To avoid confusion (and possibly solve the issue Slava reports about packedPFCandidate puppiWeightNoLep), it may be better to compute PUPPI-no-lepton (for MET) in AOD as well.

Well, we're then beyond even what I expected for this particular solution because we then need **two** separate puppi modules run in sequence. Can we finish the existing scope first?

CommonTools/PileupAlgos/plugins/PuppiProducer.cc
```diff
@@ -288,9 +294,17 @@ void PuppiProducer::produce(edm::Event& iEvent, const edm::EventSetup& iSetup) {
...
     } else {
-      pfCand->setP4(puppiP4s.back());
+      pfCandWeighted->setP4(puppiP4s.back());
+      pfCandWeighted->setSourceCandidatePtr(aCand.sourceCandidatePtr(0));
+      pfCand->setP4(aCand.p4());
       pfCand->setSourceCandidatePtr(aCand.sourceCandidatePtr(0));
+      if (fPuppiForLeptons) {
+        pfCandWeighted->setPuppiWeight(1.0, lWeights[iCand]);
+      } else {
+        pfCandWeighted->setPuppiWeight(lWeights[iCand]);
```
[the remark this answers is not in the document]
rappoccio: But the packed candidates have separate logic there. Those are if the weights are kept, not computed (which is what we're doing in RECO). The logic is different. 

I think we should keep the weights for both. If people want to "uncorrect" for whatever reason, the weights are there (unless it is zero of course).

CommonTools/PileupAlgos/plugins/PuppiProducer.cc
```diff
@@ -288,9 +294,17 @@ void PuppiProducer::produce(edm::Event& iEvent, const edm::EventSetup& iSetup) {
       pCand->setSourceCandidatePtr(aCand.sourceCandidatePtr(0));
       fPackedPuppiCandidates->push_back(*pCand);
     } else {
-      pfCand->setP4(puppiP4s.back());
+      pfCandWeighted->setP4(puppiP4s.back());
+      pfCandWeighted->setSourceCandidatePtr(aCand.sourceCandidatePtr(0));
```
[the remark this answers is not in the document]
rappoccio: Not exactly. I don't think we envision a case where we are inputting the PFCandidates and keeping the weights (which is what is done before).

CommonTools/PileupAlgos/plugins/PuppiProducer.cc
```diff
@@ -308,6 +322,7 @@ void PuppiProducer::produce(edm::Event& iEvent, const edm::EventSetup& iSetup) {
       values[ic] = pkref;
     }
   } else {
+    iEvent.put(std::move(fPuppiCandidatesWeighted), "weighted");
     edm::OrphanHandle<reco::PFCandidateCollection> oh = iEvent.put(std::move(fPuppiCandidates));
     for (unsigned int ic = 0, nc = oh->size(); ic < nc; ++ic) {
       reco::CandidatePtr pkref(oh, ic);
```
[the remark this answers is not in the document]
rappoccio: No, this is basically for backwards-compatibility for now. The information is duplicated in the new PFCandidate members and the map. We don't need the map at all for the "weighted" (to be changed to "p4scaled") collection.

RecoParticleFlow/PFProducer/python/pfLinker_cff.py
rappoccio: There are a ton of places where people assume "particleFlowTmp":

https://cmssdt.cern.ch/dxr/CMSSW/search?q=particleFlowTmp&case=true&redirect=true

Can we descope that to some other time? This is already complicated enough =(

CommonTools/PileupAlgos/plugins/PuppiProducer.cc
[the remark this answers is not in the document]
slava77: I actually thought that we will (thinking of getting a similarity of running puppi in the miniAOD workflows and leaving unchanged the miniAOD/PAT workflow variants that may need to remake objects downstream from puppi in the same way they did before).

CommonTools/PileupAlgos/plugins/PuppiProducer.cc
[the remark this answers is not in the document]
slava77: If it were for physics use backward compatibility, then this would've used a map of fPuppiCandidatesWeighted (that's what the old code was producing at least for the physics purpose).

I guess if the bw compat is for the collection naming pattern (not caring if it works, if anyone uses it), then OK, mapping the collection without an instance name makes sense.

RecoParticleFlow/PFProducer/python/pfLinker_cff.py
slava77: My comment was about a name of what's currently particleFlowTmp2 in this PR

rappoccio: Ahh okay I can change that.

CommonTools/PileupAlgos/plugins/PuppiProducer.cc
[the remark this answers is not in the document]
rappoccio: It's not for "physics use" but for migration in the next couple of months. It should be dropped when the migration is complete, I agree.

CommonTools/PileupAlgos/plugins/PuppiProducer.cc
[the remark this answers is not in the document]
rappoccio: If you have the PFCandidates and are "keeping the weights" then there's no reason to run the module at all. It would just be left out of the workflow since it would just copy the PFCandidates to a new collection.

CommonTools/PileupAlgos/plugins/PuppiProducer.cc
```diff
@@ -48,8 +48,10 @@ PuppiProducer::PuppiProducer(const edm::ParameterSet& iConfig) {
 
   if (fUseExistingWeights || fClonePackedCands)
     produces<pat::PackedCandidateCollection>();
-  else
+  else {
     produces<reco::PFCandidateCollection>();
+    produces<reco::PFCandidateCollection>("weighted");
```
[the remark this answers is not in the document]
rappoccio: I agree on the p4scaled name. Tokens will be addressed too.

RecoParticleFlow/PFProducer/python/pfLinker_cff.py
rappoccio: I don't think I understand this one. It's being imported by the particle flow sequence, so it doesn't need to be defined ahead of time, right?

RecoParticleFlow/Configuration/python/RecoParticleFlow_cff.py
rappoccio: Removed

RecoParticleFlow/PFProducer/python/pfLinker_cff.py
slava77: I mean that particleFlowPtrs should be defined in another file; it logically does not belong here anymore

rappoccio:
OK I addressed all of the comments except this one, which I don't necessarily understand:

https://github.com/cms-sw/cmssw/pull/28659/#discussion_r364867587

slava77:
@cmsbuild please test

rappoccio:
The one failing unit test is pretty complicated. It is a multi-step customized workflow with a lot of extra options. I may need someone from the Tau group to look it over. The remaining MINIAOD-level checks work fine. This one may need special treatment.

slava77:
there are logerrors of JetPtMismatch kind 
e.g.  wf 136.85
https://cmssdt.cern.ch/SDT/jenkins-artifacts/pull-request-integration/PR-acfbc8/4215/runTheMatrix-results/136.85_RunEGamma2018A+RunEGamma2018A+HLTDR2_2018+RECODR2_2018reHLT_skimEGamma_Offline_L1TEgDQM+HARVEST2018_L1TEgDQM/step3_RunEGamma2018A+RunEGamma2018A+HLTDR2_2018+RECODR2_2018reHLT_skimEGamma_Offline_L1TEgDQM+HARVEST2018_L1TEgDQM.log

```
Begin processing the 23rd record. Run 315489, Event 47618378, LumiSection 78 on stream 0 at 13-Jan-2020 21:14:32.903 CET
%MSG-e JetPtMismatch:   CandSecondaryVertexProducer:pfInclusiveSecondaryVertexFinderTagInfosAK8PFPuppiSoftDropSubjets  13-Jan-2020 21:14:42 CET Run: 315489 Event: 47618378
The reclustered and original fat jet 0 have different Pt's (263.743 vs 261.035 GeV, respectively).
Please check that the jet algorithm and jet size match those used for the original fat jet collection and also make sure the original fat jets are uncorrected. In addition, make sure you are not using CaloJets which are presently not supported.

In extremely rare instances the mismatch could be caused by a difference in the machine precision in which case make sure the original jet collection is produced and reclustering is performed in the same job.

```

@rappoccio 
is this a bug?

RecoParticleFlow/PFProducer/python/pfLinker_cff.py
```diff
@@ -2,10 +2,10 @@
 
 import RecoParticleFlow.PFProducer.pfLinker_cfi
 import RecoParticleFlow.PFProducer.particleFlowTmpPtrs_cfi
-particleFlow = RecoParticleFlow.PFProducer.pfLinker_cfi.pfLinker.clone()
-particleFlow.PFCandidate = [cms.InputTag("particleFlowTmp")]
+particleFlowPrePuppi = RecoParticleFlow.PFProducer.pfLinker_cfi.pfLinker.clone()
+particleFlowPrePuppi.PFCandidate = [cms.InputTag("particleFlowTmp")]
 particleFlowPtrs = RecoParticleFlow.PFProducer.particleFlowTmpPtrs_cfi.particleFlowTmpPtrs.clone()
-particleFlowPtrs.src = "particleFlow"
```
slava77: this should now be moved to RecoParticleFlow_cff.py, where the particleFlow module is defined

CommonTools/PileupAlgos/plugins/PuppiProducer.cc
```diff
@@ -283,58 +290,81 @@ void PuppiProducer::produce(edm::Event& iEvent, const edm::EventSetup& iSetup) {
...
+      pfCand->setP4(aCand.p4());
       pfCand->setSourceCandidatePtr(aCand.sourceCandidatePtr(0));
-      fPuppiCandidates->push_back(*pfCand);
+      if (fPuppiForLeptons) {
+        pfCandWeighted->setPuppiWeight(1.0, lWeights[iCand]);
+        pfCand->setPuppiWeight(1.0, lWeights[iCand]);
+      } else {
+        pfCandWeighted->setPuppiWeight(lWeights[iCand]);
+        pfCand->setPuppiWeight(lWeights[iCand]);
+      }
```
slava77: Shouldn't there be a way to set both weights here (by reading a collection with the other weight already set)?

is there a way to compute both weights in one go?

slava77 (commented):
There are no use cases in this PR for the p4scaled collection.
Please outline which modules will be changed to use it.
Having at least one use case would be useful.

I see that ```recoPFCandidateedmPtredmValueMap_particleFlow_{photons,muons,electrons}``` are still missing. They should be recovered unless there is an OK from the PF group that these are not needed.

The puppiNoLep weight in the standard miniAOD packed candidates is still broken (always 1.0).
This should be fixed.

rappoccio:
> there are logerrors of JetPtMismatch kind
> e.g. wf 136.85
> https://cmssdt.cern.ch/SDT/jenkins-artifacts/pull-request-integration/PR-acfbc8/4215/runTheMatrix-results/136.85_RunEGamma2018A+RunEGamma2018A+HLTDR2_2018+RECODR2_2018reHLT_skimEGamma_Offline_L1TEgDQM+HARVEST2018_L1TEgDQM/step3_RunEGamma2018A+RunEGamma2018A+HLTDR2_2018+RECODR2_2018reHLT_skimEGamma_Offline_L1TEgDQM+HARVEST2018_L1TEgDQM.log
> 
> ```
> Begin processing the 23rd record. Run 315489, Event 47618378, LumiSection 78 on stream 0 at 13-Jan-2020 21:14:32.903 CET
> %MSG-e JetPtMismatch:   CandSecondaryVertexProducer:pfInclusiveSecondaryVertexFinderTagInfosAK8PFPuppiSoftDropSubjets  13-Jan-2020 21:14:42 CET Run: 315489 Event: 47618378
> The reclustered and original fat jet 0 have different Pt's (263.743 vs 261.035 GeV, respectively).
> Please check that the jet algorithm and jet size match those used for the original fat jet collection and also make sure the original fat jets are uncorrected. In addition, make sure you are not using CaloJets which are presently not supported.
> 
> In extremely rare instances the mismatch could be caused by a difference in the machine precision in which case make sure the original jet collection is produced and reclustering is performed in the same job.
> ```
> 
> @rappoccio
> is this a bug?

Nope, it's a feature of the "First Circle of Hell" solution. This is one of the things BTV will need to fix in the new scheme after this is migrated.

CommonTools/PileupAlgos/plugins/PuppiProducer.cc
rappoccio: Not how it's currently set up. The weights are computed with the same code upstream. If we want them to be run together I have to rewrite the module.

slava77:
> Nope, it's a feature of the "First Circle of Hell" solution. This is one of the things BTV will need to fix in the new scheme after this is migrated.

the PR is likely to be accepted only if it does not break any production downstream use cases.
I'm not sure if the comment implies there is something to fix in this context; if yes, the PR is incomplete.

rappoccio:
> > Nope, it's a feature of the "First Circle of Hell" solution. This is one of the things BTV will need to fix in the new scheme after this is migrated.
> 
> the PR is likely to be accepted only if it does not break any production downstream use cases.
> I'm not sure if the comment implies there is something to fix in this context; if yes, the PR is incomplete.

I've been mentioning for a while now that this will be incomplete, and that's what I mean by "This is the first in an expected long sequence of follow-on PRs that will make use of PUPPI in RECO, simplifying many downstream workflows." from the very first comment of the PR ;). There is zero chance of me being able to change every downstream module that makes use of the "p4" candidate model assumption. I don't have the expertise to do that, nor the time, unfortunately.

CommonTools/PileupAlgos/plugins/PuppiProducer.cc
```diff
@@ -283,58 +290,81 @@ void PuppiProducer::produce(edm::Event& iEvent, const edm::EventSetup& iSetup) {
...
+    // source candidate appropriately, and set the p4 of the packed candidate.
     if (fUseExistingWeights || fClonePackedCands) {
       pCand->setP4(puppiP4s.back());
       pCand->setSourceCandidatePtr(aCand.sourceCandidatePtr(0));
-      fPackedPuppiCandidates->push_back(*pCand);
+      fPackedPuppiCandidates.push_back(*pCand);
     } else {
-      pfCand->setP4(puppiP4s.back());
+      // Here, we are not using packed candidates. That is, this is MINIAOD
+      // and we are computing puppi for the PFCandidates themselves.
```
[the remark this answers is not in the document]
rappoccio: Sorry, I meant RECO. Will fix the comment.

slava77:
> I've been mentioning for a while now that this will be incomplete, and that's what I mean by "This is the first in an expected long sequence of follow-on PRs that will make use of PUPPI in RECO, simplifying many downstream workflows." from the very first comment of the PR ;).

I know that more work will be needed to adopt to using the weights from the PFCands.
However, all of this activity is under an assumption that the sequence of PRs does not break anything created/used in production/analysis.

rappoccio:
> I know that more work will be needed to adopt to using the weights from the PFCands.
> However, all of this activity is under an assumption that the sequence of PRs does not break anything created/used in production/analysis.

We're going to be at an impasse, then. Such a major change will not be able to be integrated. So what do we do?

slava77:
> We're going to be at an impasse, then. Such a major change will not be able to be integrated. So what do we do?

perhaps a different circle is the way to go?

I think that temporary CPU and file size increase is acceptable for the steps towards completion. This, however should be clear that it's temporary with a clear path to what needs to be done.
The requirement for not breaking the standard production/analysis objects would have to be followed though.

slava77:
> The one failing unit test is pretty complicated. It is a multi-step customized workflow with a lot of extra options. I may need someone from the Tau group to look it over. The remaining MINIAOD-level checks work fine. This one may need special treatment.

The problem in this unit tests is related subtly to this PR and can appear in other workflows.
I traced this down to 
https://github.com/cms-sw/cmssw/blob/88b29b8a0401d97f62b19c822eebb75d5511cc28/PhysicsTools/PatAlgos/python/tools/coreTools.py#L151-L153
coming from the call to removeMCMatching called by runOnData in the call to the standard miniAOD_customizeAllData.
Here the code assumes that the configured slimmedGenJets$prefix  have to have a companion patJets$prefix. This PR finally breaks this symmetry by removing patJetsAK8.

Curiously enough, now the prefix relationship was broken a while back in https://github.com/cms-sw/cmssw/blob/88b29b8a0401d97f62b19c822eebb75d5511cc28/PhysicsTools/PatAlgos/python/slimming/applySubstructure_cff.py#L193-L195, but we still were creating the patJetsAK8 (in the CHS chain), which was removed in https://github.com/cms-sw/cmssw/pull/28659/files#diff-416ad16d53535660648a4a4bb98636d6L39-L44

I think that an easy solution is to either explicitly add slimmedGenJetsAK8 in the list to be removed https://github.com/cms-sw/cmssw/blob/88b29b8a0401d97f62b19c822eebb75d5511cc28/PhysicsTools/PatAlgos/python/tools/coreTools.py#L105
or to add the AK8 prefix explicitly in https://github.com/cms-sw/cmssw/blob/88b29b8a0401d97f62b19c822eebb75d5511cc28/PhysicsTools/PatAlgos/python/tools/coreTools.py#L150

slava77:
@rappoccio 
please clarify on your plan to update this PR.
It looks like at least the unit test is fixable and some other updates appear to be doable as well from the review.

rappoccio:
We're in discussion with EGM/MUO and BTV about the remaining issues. It's going to be a bit longer.

emilbols:
@rappoccio @slava77 
this https://github.com/emilbols/cmssw/tree/BTV_LO_fix should fix the issues regarding btv.

However since we need to have different behavior for the puppi and CHS jet collection, a flag is needed to determine if we are looking at puppi or chs. Passing a boolean flag via configuration files which can end up being mismatched by the user, is really not ideal in my opinion. This fix i am linking here relies on checking if the jet collection label contains the string "Puppi", but i am not sure if this is always a safe assumption. Maybe we could add a boolean flag into the jet collection itself, which is set in the virtual jet producer, indicating whether it is a puppi jet or chs jet.

rappoccio:
Thanks Emil, that's a good suggestion. I have merged your fix for now, but we can implement that shortly as well.

slava77:
> -code-checks

@rappoccio 
please apply the patch, in case you missed the message.
Thanks.

silviodonato:
This is a kind reminder about this PR @rappoccio

rappoccio:
Sorry for the delay, it's the first week of classes as I mentioned. 

Please hold off on checks, cmsdev20 is full and my checks have not completed. I have to clean out the area and restart because something broke due to the disk being full.

slava77:
> Please hold off on checks, cmsdev20 is full and my checks have not completed. I have to clean out the area and restart because something broke due to the disk being full.

@rappoccio 
please clarify if the PR checks are now done on your side.
Thank you.

slava77:
@cmsbuild please test

rappoccio:
Sorry, missed this earlier (last week was completely hectic). My tests passed. I don't know if there are other cases that need to be addressed. There is still an open question about what to do with the EGM and MUO value maps. @bendavid and @hatakeyamak have weighed in, perhaps we can discuss on Wednesday at the JME meeting?

silviodonato:
@rappoccio Do you news about this PR after the JME meeting?

rappoccio:
Yes apparently we're killing this idea. The next circle of hell will come at a later time.

PR #28662, opened 2019-12-19
Fix muon track producer and muon hl tvalidation

abbiendi:
This PR is meant to fix the muon HLT validation, which has not been running as wished for the new HLT reconstruction (few validation modules not working) . To do so an extension of the MuonTrackProducer is needed, to deal correctly with the various kinds of muon tracks, which can have muon/tracker/mu+tracker hits. Validation of L2Muons is also recovered as it used to be there as an essential step in the chain.
Taking this occasion the offline validation is also completed, adding the highest-level reco::Muon objects to the standard validation: PF, TuneP, and All muons which were still missing it. 
The customization for the Run3 and Phase2 scenarios has been modified to bring Run3 to the complete validation sequence, while still keeping Phase2 with a reduced sequence.

A bug has been fixed in the logic of the MuonTrackProducer which is now used for the new modules (not affecting already existing workflows, that's why perhaps it went unobserved from 740 to now).

A flat-P (total momentum) particle gun generator has been added, which has already been widely used within the Muon POG, particularly useful for studying the performance of high-momentum muons. 

#### PR validation:

Tested on CMSSW_11_0_0_pre3 RelVals for ZMM_14TeV RelVals for 2021 scenario with and without PU:
- rerunning step2+step3+step4 for no-pileup  (hence testing all the new features)
- rerunning step3+step4 for 2021 PU scenario
- also on few events from single muon gun with pt=100 and the new flat-P gun
All the new results make sense.
The PR can be tested simply with the standard matrix test.
None of the existing validation modules will change.


HLTrigger/Configuration/python/HLTrigger_EventContent_cff.py
```diff
@@ -209,6 +209,7 @@
         'keep *_hltIter4Merged_*_*',
         'keep *_hltIterL3GlbMuon_*_*',
         'keep *_hltIterL3MuonMerged_*_*',
+        'keep *_hltIterL3MuonAndMuonFromL1Merged_*_*',
```
Martin-Grunewald: This is an auto-generated file, thus changes made directly will be overwritten.
You need to file a JIRA ticket requeting addition of this product!

HLTrigger/Configuration/python/HLTrigger_EventContent_cff.py
```diff
@@ -548,6 +549,7 @@
...
```
Martin-Grunewald: This is an auto-generated file, thus changes made directly will be overwritten.
You need to file a JIRA ticket requeting addition of this product!

Martin-Grunewald:
-1
for changes in HLTrigger/Configuration

HLTrigger/Configuration/python/HLTrigger_EventContent_cff.py
abbiendi: I never created a JIRA ticket, and would not know what is to be indicated in several fields... would you please help me ? 
The point is simple: this is the highest-level collection of muon tracker tracks (linked to the reco::Muon object in HLT), which enter in many trigger paths. The validation cannot run for it if it is not saved in the Event. Of course the trigger (or its offline version) will run anyway since it is all happening in the same step. 
Being one of the most important ingredients in the Muon HLT it *should* be validated. 
@Martin-Grunewald @folguera @Fedespring @cericeci @calderona

Martin-Grunewald: Please remove the changes to this file. 
It is part of https://github.com/cms-sw/cmssw/pull/28681 done through ConfDB

abbiendi: Thanks ! doing it

civanch:
please test

abbiendi:
The errors found in matrix tests have two different origins, although similar:
- errors in the Run3 and Phase2 workflows depend on a missing track collection, which is fixed in #28681. So this is removed only after that PR is merged.
- errors in Run2 workflows depend on a missing HLT product, since they all have been processed with a Fake HLT configuration. This has to be fixed, and I am updating the PR now.

abbiendi:
please could we move on with tests

jfernan2:
please test

jfernan2:
@abbiendi this PR is adding ~2k-4k new histograms and removing ~1k empty ones, as stated in the PR description. The numbers seem to be in line with the expected changes: +696 for Run1/special FS workflows, +1606 for Run2, +4341 for Run3 and +1387 for Phase2
Could you please have a last look to see if it makes sense to you too?
https://cmssdt.cern.ch/SDT/jenkins-artifacts/baseLineComparisons/CMSSW_11_1_X_2020-01-06-1100+56dd37/34695/dqm-histo-comparison-summary.html
There are some empty news ones in HLT / Muon / MuonTrack / hltL2Muons  folder and others, not sure if due to stats.

@calderona FYI

abbiendi:
From my side I confirm that the changes are the expected ones. I think some of the histos are empty due to low stat (the process is in general TTbar where there could be no muons passing the trigger on 10 events only !)
Basically the change in customization for the different eras reflects the idea to have full validation for Run3, reasonable validation for Phase2 (yet no HLT, but the most important offline distributions). Since the HLT menu for Run2 and before is now "Fake" there are no HLT muons in those cases and hence less folders in the output.

@calderona please have a look if it seems fine for you too

abbiendi:
small addition: the expected changes are in DQM folders "HLT/Muon" and "Muons/RecoMuonV". I have not changed anything outside, so I have no idea of changes to distributions in other folders (if any)

calderona:
> From my side I confirm that the changes are the expected ones. I think some of the histos are empty due to low stat (the process is in general TTbar where there could be no muons passing the trigger on 10 events only !)
> Basically the change in customization for the different eras reflects the idea to have full validation for Run3, reasonable validation for Phase2 (yet no HLT, but the most important offline distributions). Since the HLT menu for Run2 and before is now "Fake" there are no HLT muons in those cases and hence less folders in the output.
> 
> @calderona please have a look if it seems fine for you too

It is ok with me. Please go ahead.

jfernan2:
+1

civanch:
+1

abbiendi:
@qliphy  @agrohsje   please, could you review this and possibly sign it ?

abbiendi:
@qliphy @agrohsje   what are we waiting for here ?  can we move on ?
@calderona

qliphy:
+1

silviodonato:
+1

PR #28665, opened 2019-12-20
Mf regression v2

namapane:
#### PR description:

- Update MF regression code to use binary files from standard data path, cf. PR in cms-data;
- regression.py updated with command-line arguments to cover all possible cases. (This is intended to be added as a unit test in a next PR);
- Added consumes to DDCompactViewMFESProducer; cf. conversation in #28346

This PR replaces #28642 as per discussion therein.

#### PR validation:

Ran regression tests against earlier regression files (produced in past CMSSW releases) for all possible combinations of era, current, and MF producer type.


namapane:
@slava77, regarding your comments to #28642:
- modified volumeBasedMagneticField_dd4hep_160812_cfi.py to avoid duplication as you requested, but without adding one new files, I hope that is OK.
- Could not rename the DDDetectorESProducer since renaming breaks it for reasons I don't understand (see my comment in #28642). I hope review of this PR can go on while this is sorted out.

Dr15Jones:
please test

slava77:
> @slava77, regarding your comments to #28642:
> 
>     * modified volumeBasedMagneticField_dd4hep_160812_cfi.py to avoid duplication as you requested, but without adding one new files, I hope that is OK.
> 
>     * Could not rename the DDDetectorESProducer since renaming breaks it for reasons I don't understand (see my comment in #28642). I hope review of this PR can go on while this is sorted out.

please clarify how the modified volumeBasedMagneticField_dd4hep_160812_cfi was tested.

namapane:
> 
> 
> > @slava77, regarding your comments to #28642:
> > ```
> > * modified volumeBasedMagneticField_dd4hep_160812_cfi.py to avoid duplication as you requested, but without adding one new files, I hope that is OK.
> > 
> > * Could not rename the DDDetectorESProducer since renaming breaks it for reasons I don't understand (see my comment in #28642). I hope review of this PR can go on while this is sorted out.
> > ```
> 
> please clarify how the modified volumeBasedMagneticField_dd4hep_160812_cfi was tested.

As specified in the PR description:
"Ran regression tests against earlier regression files (produced in past CMSSW releases) for all possible combinations of era, current, and MF producer type." 
[which was re-done after the modification]

cvuosalo:
+1

slava77:
+1

for https://github.com/cms-sw/cmssw/pull/28665	3b771bf 
- code changes are in line with the PR description and the follow up review (initially submitted for #28642 ). These chances look fine for the needs of the standalone setup tests with magnetic field.
    - The discussion in https://github.com/cms-sw/cmssw/pull/28642#discussion_r359587376 may resurface if the volumeBasedMagneticField_dd4hep_160812_cfi will need to be used in a full detector setup (not just MF tests), which is likely not needed if we can switch to the production workflow setup using the MF from DB as it's done for the older DDD case.

fabiocos:
+1

PR #28666, opened 2019-12-20
Properly handle file open for Streamer based sources

Dr15Jones:
#### PR description:
If the ProductRegistry can be modified upon opening a new file, the framework must be notified of the new file before the registry is updated. The streamer based sources were not following that rule.

#### PR validation:

The code compiles and the old and new unit tests all pass.

### DIFF
--- a/CalibCalorimetry/EcalLaserSorting/src/WatcherStreamFileReader.cc
+++ b/CalibCalorimetry/EcalLaserSorting/src/WatcherStreamFileReader.cc
@@ -215,7 +215,7 @@ const EventMsgView* WatcherStreamFileReader::getNextEvent() {
   edm::StreamerInputFile* inputFile;
 
   //go to next input file, till no new event is found
-  while ((inputFile = getInputFile()) != nullptr && inputFile->next() == 0) {
+  while ((inputFile = getInputFile()) != nullptr && inputFile->next() != edm::StreamerInputFile::Next::kEvent) {
     closeFile();
   }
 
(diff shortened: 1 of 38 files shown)


Dr15Jones:
please test

silviodonato:
hold

silviodonato:
please test

jfernan2:
+1

Dr15Jones:
+1

perrotta:
+1
- FOr reco: FedRawDataInputSource in EventFilter updated to comply with the new handling of file opening in RawInputSource

mommsen:
The changes look fine to me. But @smorovic should have a look at it, too.

smorovic:
The changes look fine.

New unit tests cover changes in the Streamer source code. I ran a few tests with the FedRawDataInputSource. With scripts included with the code it runs fine. There is one small difference: if source issued "newLumi" twice for same lumisection (which it avoids, so it would be a bug) it would result in IsStop being with the old RawInoutSource version (premature and silent end of Run). In the new version this will trigger assert (due to no cached event) which is a better behavior.

Dr15Jones:
ping

mommsen:
+1
Thanks for the ping. It fell off my radar. Remi

silviodonato:
+1
@christopheralanwest  @franzoni  @tlampen  @pohsun  @tocheng 
please have a look

silviodonato:
merge

PR #28701, opened 2019-12-31
Apply strict memory ownership to muon tracking

Dr15Jones:
#### PR description:

The static analyzer had identified a path through the muon code which would cause a memory leak. After fixing that one case, it was apparent that the code base would benefit from a memory handling overhaul.

#### PR validation:

The code compiles.

### DIFF
--- a/RecoMuon/CosmicMuonProducer/src/CosmicMuonTrajectoryBuilder.cc
+++ b/RecoMuon/CosmicMuonProducer/src/CosmicMuonTrajectoryBuilder.cc
@@ -133,7 +133,7 @@ void CosmicMuonTrajectoryBuilder::setEvent(const edm::Event& event) {
 }
 
 MuonTrajectoryBuilder::TrajectoryContainer CosmicMuonTrajectoryBuilder::trajectories(const TrajectorySeed& seed) {
-  vector<Trajectory*> trajL = vector<Trajectory*>();
+  TrajectoryContainer emptyContainer;
 
   MuonPatternRecoDumper debug;
 
@@ -163,7 +163,7 @@ MuonTrajectoryBuilder::TrajectoryContainer CosmicMuonTrajectoryBuilder::trajecto
   LogTrace(category_) << "found " << navLayers.size() << " compatible DetLayers for the Seed";
 
   if (navLayers.empty())
-    return trajL;
+    return emptyContainer;
 
   vector<DetWithState> detsWithStates;
   LogTrace(category_) << "Compatible layers: ";
@@ -188,13 +188,13 @@ MuonTrajectoryBuilder::TrajectoryContainer CosmicMuonTrajectoryBuilder::trajecto
   }
   detsWithStates.clear();
   if (!lastTsos.isValid())
-    return trajL;
+    return emptyContainer;
 
   TrajectoryStateOnSurface secondLast = lastTsos;
 
   lastTsos.rescaleError(10.0);
 
-  Trajectory* theTraj = new Trajectory(seed, alongMomentum);
+  Trajectory theTraj(seed, alongMomentum);
 
   navLayers.clear();
 
@@ -237,14 +237,14 @@ MuonTrajectoryBuilder::TrajectoryContainer CosmicMuonTrajectoryBuilder::trajecto
       continue;
 
     for (vector<TrajectoryMeasurement>::const_iterator theMeas = measL.begin(); theMeas != measL.end(); ++theMeas) {
-      pair<bool, TrajectoryStateOnSurface> result = updator()->update((&*theMeas), *theTraj, propagator());
+      pair<bool, TrajectoryStateOnSurface> result = updator()->update((&*theMeas), theTraj, propagator());
 
       if (result.first) {
         LogTrace(category_) << "update ok ";
         incrementChamberCounters(
             (*rnxtlayer), DTChamberUsedBack, CSCChamberUsedBack, RPCChamberUsedBack, TotalChamberUsedBack);
         secondLast = lastTsos;
-        if ((!theTraj->empty()) && result.second.isValid()) {
+        if ((!theTraj.empty()) && result.second.isValid()) {
           lastTsos = result.second;
           LogTrace(category_) << "get new lastTsos here " << lastTsos.globalMomentum() << " at "
                               << lastTsos.globalPosition();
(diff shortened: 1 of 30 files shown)


Dr15Jones:
please test

Dr15Jones:
There are further changes which could be done to this code. I only made these changes as I figured they would have the least likelihood to cause any negative performance issues.

Some further changes could be
- The `TrackCand` class (which is a `std::pair`) should be looked at to have its memory management localized.
- The `CandidateContainer` could probably be changed to `std::vector<MuonCandidate>` with little to no impact on performance
- The `MuonCandidate` could be changed to use `std::optional<Trajectory>` instead of `std::unique_ptr<Trajectory>` which would decrease the number of `new` and `delete` calls and consolidate memory better.

RecoMuon/CosmicMuonProducer/src/CosmicMuonTrajectoryBuilder.cc
```diff
@@ -133,7 +133,7 @@ void CosmicMuonTrajectoryBuilder::setEvent(const edm::Event& event) {
 }
 
 MuonTrajectoryBuilder::TrajectoryContainer CosmicMuonTrajectoryBuilder::trajectories(const TrajectorySeed& seed) {
-  vector<Trajectory*> trajL = vector<Trajectory*>();
```
Dr15Jones: `trajL` was always empty until the very end of the subroutine.

RecoMuon/CosmicMuonProducer/src/CosmicMuonTrajectoryBuilder.cc
```diff
@@ -133,7 +133,7 @@ void CosmicMuonTrajectoryBuilder::setEvent(const edm::Event& event) {
...
+  TrajectoryContainer emptyContainer;
```
Dr15Jones: The compiler will not allow this to be `const TrajectoryContainer` because that would require a copy constructor for `std::unique_ptr`.

RecoMuon/GlobalTrackingTools/src/GlobalTrajectoryBuilderBase.cc
```diff
@@ -242,86 +241,63 @@ MuonCandidate::CandidateContainer GlobalTrajectoryBuilderBase::build(const Track
       refitted1 = theGlbRefitter->refit(*it->trackerTrack(), tTT, allRecHits, theMuonHitsOption, theTopo);
       LogTrace(theCategory) << "     This track-sta refitted to " << refitted1.size() << " trajectories";
 
-      Trajectory* glbTrajectory1 = nullptr;
+      std::unique_ptr<Trajectory> glbTrajectory1;
       if (!refitted1.empty())
-        glbTrajectory1 = new Trajectory(*(refitted1.begin()));
+        glbTrajectory1 = std::make_unique<Trajectory>(*(refitted1.begin()));
```
Dr15Jones: Maybe this could be `glbTrajectory1 = std::move(refitted1.begin());` ?

RecoMuon/GlobalTrackingTools/src/GlobalTrajectoryBuilderBase.cc
```diff
@@ -242,86 +241,63 @@ MuonCandidate::CandidateContainer GlobalTrajectoryBuilderBase::build(const Track
...
         tmpSeed = it->trackerTrack()->seedRef();
 
       TC refitted0;
-      Trajectory* tkTrajectory = nullptr;
+      std::unique_ptr<Trajectory> tkTrajectory;
       if (!(it->trackerTrajectory() && it->trackerTrajectory()->isValid())) {
         refitted0 = theTrackTransformer->transform(it->trackerTrack());
         if (!refitted0.empty()) {
-          tkTrajectory = new Trajectory(*(refitted0.begin()));
+          tkTrajectory = std::make_unique<Trajectory>(*(refitted0.begin()));
```
Dr15Jones: possibly use `std::move` here as well to avoid making another `Trajectory`?

RecoMuon/GlobalTrackingTools/src/GlobalTrajectoryBuilderBase.cc
```diff
@@ -242,86 +241,63 @@ MuonCandidate::CandidateContainer GlobalTrajectoryBuilderBase::build(const Track
...
+      refittedResult.emplace_back(std::make_unique<MuonCandidate>(
+          std::move(tkTrajectory), it->muonTrack(), it->trackerTrack(), std::move(cpy)));
     }
   }
 
   // choose the best global fit for this Standalone Muon based on the track probability
   CandidateContainer selectedResult;
-  MuonCandidate* tmpCand = nullptr;
-  if (!refittedResult.empty())
-    tmpCand = *(refittedResult.begin());
```
Dr15Jones: The start condition was moved into the loop

RecoMuon/TrackingTools/interface/MuonTrackFinder.h
```diff
@@ -61,16 +62,16 @@ class MuonTrackFinder {
   void setEvent(const edm::Event&);
 
   /// convert the trajectories into tracks and load them in to the event
-  edm::OrphanHandle<reco::TrackCollection> load(const TrajectoryContainer&, edm::Event&, const TrackerTopology& ttopo);
+  edm::OrphanHandle<reco::TrackCollection> load(TrajectoryContainer&, edm::Event&, const TrackerTopology& ttopo);
```
Dr15Jones: Passing the container as `const` was not appropriate as the function deletes items held by the container. Although C++ allows that, it is not strictly const correct.

RecoMuon/TrackingTools/interface/MuonTrackLoader.h
```diff
@@ -55,18 +55,18 @@ class MuonTrackLoader {
 
   /// Convert the trajectories into tracks and load the tracks in the event
   edm::OrphanHandle<reco::TrackCollection> loadTracks(
-      const TrajectoryContainer&, edm::Event&, const TrackerTopology& ttopo, const std::string& = "", bool = true);
```
Dr15Jones: The containers elements are deleted by the function so strictly speaking this is not const.

perrotta:
Adding the muon reco contacts, to take note ot these changes and the further memory management updates proposed by Chris in https://github.com/cms-sw/cmssw/pull/28701#issuecomment-569980202: @Fedespring + Carlos Erice Cid (which is his github name, by the way?)

perrotta:
+1
- A set of technical changes that go along what outlined in the PR description
- Further possible updates are listed in https://github.com/cms-sw/cmssw/pull/28701#issuecomment-569980202, but as far as I understand they are not for this PR
- Jenkins tests pass and show no differences in reco or other objects

Martin-Grunewald:
+1

RecoMuon/CosmicMuonProducer/src/CosmicMuonTrajectoryBuilder.cc
VinInn: The new implementation "intentionally" prevents return value optimization (the previous one allowed it).

Dr15Jones: Playing around with compiler explorer seems to show minimal differences between reusing a container and creating a new one:

https://godbolt.org/z/k36w2x

The difference appears to be a second call to a default constructor. In all cases a move constructor is still called. I originally used a std::vector<int> as the container but with all the inlining, it was difficult to tell exactly what the compiler was doing (although it appeared to also just calling an additional default constructor call).

silviodonato:
+1

PR #28714, opened 2020-01-07
Changes to the PPS pixels Online/Offline DQM

AndreaBellora:
Changed PPS Offline DQM plots for the pixel tracker to include efficiency plots and exclude some unnecessary ones. Added flags to remove the plots of specific detector planes when needed.

### DIFF
--- a/DQM/CTPPS/plugins/CTPPSPixelDQMSource.cc
+++ b/DQM/CTPPS/plugins/CTPPSPixelDQMSource.cc
@@ -356,11 +356,50 @@ void CTPPSPixelDQMSource::bookHistograms(DQMStore::IBooker &ibooker,
 
         ibooker.setCurrentFolder(rpd);
 
+        const float x0Maximum = 70.;
+        const float y0Maximum = 15.;
+        string st = "track intercept point";
         string st2 = ": " + stnTitle;
+        h2trackXY0[indexP] = ibooker.book2D(
+            st, st + st2 + ";x0;y0", int(x0Maximum) * 2, 0., x0Maximum,
+            int(y0Maximum) * 4, -y0Maximum, y0Maximum);
+        h2trackXY0[indexP]->getTH2F()->SetOption("colz");
 
-        string st = "hit multiplicity in planes";
-        string st3 = ";PlaneIndex(=pixelPot*PlaneMAX + plane)";
+        st = "number of tracks per event";
+        htrackMult[indexP] = ibooker.bookProfile(
+            st, rpTitle + ";number of tracks", NLocalTracksMAX + 1, -0.5,
+            NLocalTracksMAX + 0.5, -0.5, NLocalTracksMAX + 0.5, "");
+        htrackMult[indexP]->getTProfile()->SetOption("hist");
+
+        hRPotActivPlanes[indexP] = ibooker.bookProfile(
+            "number of fired planes per event",
+            rpTitle + ";nPlanes;Probability", NplaneMAX + 1, -0.5,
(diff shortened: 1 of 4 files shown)


AndreaBellora:
@fabferro

jfernan2:
@AndreaBellora we don't have you identified as PPS DQM developer, can you please add yourself to the corresponding e-group in
https://twiki.cern.ch/twiki/bin/viewauth/CMS/DQMContacts#CT_PPS
?
Thanks in advance

AndreaBellora:
No problem, I've added myself to both the online and offline developers

jfernan2:
please test

jfernan2:
@AndreaBellora the 28 plots that this PR is adding seem to be empty on all workflows, not sure if it is stats of dataset, but both data and MC are empty. Have you done some private checks with more stats or different process?
https://cmssdt.cern.ch/SDT/jenkins-artifacts/baseLineComparisons/CMSSW_11_1_X_2020-01-07-1100+dbb278/34701/dqm-histo-comparison-summary.html

AndreaBellora:
I have done some private testing and the plots appeared to be full and working, I'll check why they appear empty on these workflows. Many of the plots have just been reorganized, but not changed, so I would not expect them to be empty, if they weren't already.

AndreaBellora:
@jfernan2 I've run some tests, also with the help of @fabferro. Locally everything is fine and the problem with the workflows is that only a few of them use real data runs and in none of them PPS pixels were in. 
The only one that produces some results is 136.85. With enough statistics we got a few background tracks that fill some of the plots and they appear to be working.
I have however noticed that some plots that appear in the files are not in the GUI. How can I add them in?

jfernan2:
Thanks @AndreaBellora 
The GUIs in the Bin by Bin tool are only showing those histograms which change w.r.t. baseline, that's why you don't see the full set of plots. You can find the full DQM root file in the Matrix test output folder: https://cmssdt.cern.ch/SDT/jenkins-artifacts/pull-request-integration/PR-dbb278/4159/runTheMatrix-results/136.85_RunEGamma2018A+RunEGamma2018A+HLTDR2_2018+RECODR2_2018reHLT_skimEGamma_Offline_L1TEgDQM+HARVEST2018_L1TEgDQM/

Indeed I do see entries in the 136.85 wf, but do you understand why the efficiency plots in all the plane folders are empty? 
https://cmsweb.cern.ch/dqm/dev/start?runnr=315489;dataset=/wf136_85_pr/CMSSW_11_1_X-PRcmssw_28714-4159/DQMIO;sampletype=offline_relval;filter=all;referencepos=overlay;referenceshow=customise;referencenorm=True;referenceobj1=refobj;referenceobj2=none;referenceobj3=none;referenceobj4=none;search=;striptype=object;stripruns=;stripaxis=run;stripomit=none;workspace=Everything;size=M;root=CTPPS/TrackingPixel/sector%2045/station%20220/fr_hr;focus=CTPPS/TrackingPixel/sector%2045/station%20220/fr_hr/ROCs%20hits%20multiplicity%20per%20event%20vs%20LS;zoom=yes;

AndreaBellora:
The workflow takes data from run 315489, which appears to be a 2018 run where PPS was not inserted. This means that the detectors were on, but out of acceptance and thus leads to a very small number of background tracks observed. Since we do not have tracks, the efficiency cannot be computed and the plots remain empty. With higher statistics (5000 events) we observed that a few bins get filled, thus we believe that the plot works and this issue is just given by an unfortunate dataset choice.

jfernan2:
+1

silviodonato:
+1

PR #28723, opened 2020-01-09
Fixing HB thresholds E/gamma saved rec-hits for run3 

Sam-Harper:
#### PR description:

This bug fix for Run3 MC / data enables the proper saving of HCAL rec-hits for the barrel. The current threshold is 0.8 GeV which is not appropriate for the phaseI HB.   This adjusts the rec-hits e/gamma saves to have a minimum energy of 0.1 for phaseI HB workflows.  Thanks to @swagata87 for raising the issue!

This fix would be needed for the upcoming Run3 production, the samples will be sigificantly less useful for e/gamma without it as it wouldnt allow us to develop Hcal depth based variables in the barrel. 

#### PR validation:

This will increase the AOD and miniAOD sizes for workflows with the phaseI HB detector. 

This was done with 200 TTbar events with standard pileup. 

sample: /RelValTTbar_14TeV/CMSSW_11_0_0_pre13-PU_110X_mcRun3_2023_realistic_v6-v1/GEN-SIM-DIGI-RAW
reco config cmsDriver: 
```
cmsDriver.py step3 --conditions auto:phase1_2021_realistic -n 10 --era Run3 --eventcontent RECOSIM,AODSIM,MINIAODSIM --runUnscheduled -s RAW2DIGI,L1Reco,RECO,RECOSIM,EI,PAT --datatier GEN-SIM-RECO,AODSIM,MINIAODSIM --geometry DB:Extended --io RecoFull_2021.io --python RecoFull_2021.py --conditions=110X_mcRun3_2023_realistic_v6 --no_exec --filein /store/relval/CMSSW_11_0_0_pre13/RelValTTbar_14TeV/GEN-SIM-DIGI-RAW/PU_110X_mcRun3_2023_realistic_v6-v1/20000/C27EA1AE-ADDB-3149-9867-6ABE5F653A97.root --fileout file:ttbar.root --nThreads 8
```

size increases (via ls) for 200 events:
AOD: 138061572 -> 138231541 (0.1% increase)
MiniAOD: 20615277 -> 20496350. (0.6% increase)

In miniAOD:
HBHERecHitsSorted_reducedEgamma_reducedHBHEHits_RECO  compressed size 1236 -> 1830 (50% increase)

Some of this size increase could be reduced (probably by 1/3) by properly applying the depth specific thresholds of 0.1,0.2,0.3,0.3 but this needs c++ changes to do so. 

red is energy of rec-hits saved before fix, blue is after, behaves as expected.

image

PR #28724, opened 2020-01-10
Fixing HB thresholds E/gamma saved rec-hits for run3 : 11_0_0

Sam-Harper:
#### PR description:

Bug fixes to the HB threshold to be correct for Run3 for the hits e/gmama saves for Hcal studies

backport of https://github.com/cms-sw/cmssw/pull/28723

### DIFF
--- a/RecoEgamma/EgammaIsolationAlgos/python/interestingEgammaIsoDetIdsSequence_cff.py
+++ b/RecoEgamma/EgammaIsolationAlgos/python/interestingEgammaIsoDetIdsSequence_cff.py
@@ -148,3 +148,7 @@ egamma_lowPt_exclusive.toModify(interestingGedEgammaIsoHCALDetId,
 		           minEleEt= 1.0, #default 20
 			   minPhoEt= 1.0 #default 20
 ) 
+
+from Configuration.Eras.Modifier_run3_HB_cff import run3_HB
+run3_HB.toModify(interestingEgammaIsoHCALSel,
+                 minEnergyHB = 0.1)


slava77:
@cmsbuild please test

slava77:
abort test

slava77:
this changes the default behavior for a production release for run3 and phase-2 workflows.
Nominally this is not allowed by the no-change policy.
The usual solution is to provide the new/changing feature with a (non-default) configuration change.
It seems to me that a customise_command can easily be passed at job submission; it can also be formalized a bit more robustly with a customise python method.

However, the reducedHcalRecHits have a limited use scope and one may argue for an exception.
But this requires some discussion and signoff in some (PPD or/and ORP) meeting.

slava77:
>  Nominally this is not allowed by the no-change policy.

ah, maybe I'm wrong and we still did not validate 11_0_0 for production. In that case, it may be fine for this bugfix.

@silviodonato please clarify where we are now

slava77:
@cmsbuild please test

Sam-Harper:
okay let me know if this in policy or not. If not I'll try and find another solution. Will be at the ORP meeting on tuesday to discuss  further.

slava77:
> > Nominally this is not allowed by the no-change policy.
> 
> ah, maybe I'm wrong and we still did not validate 11_0_0 for production. In that case, it may be fine for this bugfix.
> 
> @silviodonato please clarify where we are now

@srimanob , in case you know the current status

srimanob:
@slava77 @Sam-Harper 
Thanks for pointing me to this PR. Final validation on 11_0_0 for Run-3/Phase-2 is just starting. We assumed that no much change wrt. to pre13. I think we should be able to cover this bug fix.

slava77:
+1

for https://github.com/cms-sw/cmssw/pull/28724	6dfe29f 
- the backport is correct by virtue of using the same topic branch as in #28723
- changes in jenkins are only in the reduced hcal rechit collection, as in the master version
- the changes in reco output in this PR were accepted in the Jan 14 ORP.

silviodonato:
+1

PR #28726, opened 2020-01-10
PPS: vector classes used in reco, geometry etc.

jan-kaspar:
#### PR description:

This is a technical PR to reduce the number of different vector classes used in PPS reconstruction, geometry, etc. In particular, the dependence on CLHEP is dropped in favour of ROOT. Several unnecessary vector-format conversion are avoided. In the geometry-consumer code, whenever possible, the auto statement or CTPPSGeometry::Vector typedef is used in order simplify possible future type transitions.

#### PR validation:

Running
```
runTheMatrix.py -l limited --ibeos
```
gave
```
34 33 32 26 16 4 1 1 1 tests passed, 0 0 0 0 0 0 0 0 0 failed
```

This PR does not change the proton reconstruction results - see the comparison below. Blue = before this PR, red dashed = with this PR.
make_cmp

### DIFF
--- a/Geometry/VeryForwardRPTopology/interface/RPTopology.h
+++ b/Geometry/VeryForwardRPTopology/interface/RPTopology.h
@@ -10,17 +10,7 @@
 #ifndef Geometry_VeryForwardRPTopology_RPTopology
 #define Geometry_VeryForwardRPTopology_RPTopology
 
-#include <HepMC/SimpleVector.h>
-
-#include "TMath.h"
-#include "DataFormats/GeometryVector/interface/LocalVector.h"
-#include "DataFormats/Common/interface/Ref.h"
-#include "DataFormats/Math/interface/LorentzVector.h"
-#include "DataFormats/Math/interface/Point3D.h"
-
-namespace HepMC {
-  class ThreeVector;
-}
+#include "Math/Vector3D.h"
 
 /**
  *\brief Geometrical and topological information on RP silicon detector.
@@ -28,10 +18,13 @@ namespace HepMC {
  **/
 class RPTopology {
 public:
+  using Vector = ROOT::Math::DisplacementVector3D<ROOT::Math::Cartesian3D<double>>;
+  ;
+
   RPTopology();
-  inline const HepMC::ThreeVector& GetStripReadoutAxisDir() const { return strip_readout_direction_; }
-  inline const HepMC::ThreeVector& GetStripDirection() const { return strip_direction_; }
-  inline const HepMC::ThreeVector& GetNormalDirection() const { return normal_direction_; }
+  inline const Vector& GetStripReadoutAxisDir() const { return strip_readout_direction_; }
+  inline const Vector& GetStripDirection() const { return strip_direction_; }
+  inline const Vector& GetNormalDirection() const { return normal_direction_; }
 
   /// method converts strip number to a hit position [mm] in det readout coordinate
   /// in the origin in the middle of the si detector
@@ -65,9 +58,9 @@ public:
   static const double last_strip_to_border_dist_;
   static const double last_strip_to_center_dist_;
 
-  HepMC::ThreeVector strip_readout_direction_;
-  HepMC::ThreeVector strip_direction_;
-  HepMC::ThreeVector normal_direction_;
+  Vector strip_readout_direction_;
+  Vector strip_direction_;
+  Vector normal_direction_;
 };
 
 #endif  //Geometry_VeryForwardRPTopology_RPTopology
--- a/Geometry/VeryForwardRPTopology/interface/RPTopology.h
+++ b/Geometry/VeryForwardRPTopology/interface/RPTopology.h
@@ -19,7 +19,6 @@
 class RPTopology {
 public:
   using Vector = ROOT::Math::DisplacementVector3D<ROOT::Math::Cartesian3D<double>>;
-  ;
 
   RPTopology();
   inline const Vector& GetStripReadoutAxisDir() const { return strip_readout_direction_; }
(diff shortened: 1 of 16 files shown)


jfernan2:
please test

jfernan2:
+1
No changes to dqm

cvuosalo:
+1

RecoCTPPS/PixelLocal/src/RPixRoadFinder.cc
```diff
@@ -92,17 +92,17 @@ void RPixRoadFinder::findPattern() {
...
     while (it_gh2 != temp_all_hits.end()) {
       bool same_pot = false;
       CTPPSPixelDetId tmpGh2Id = CTPPSPixelDetId(it_gh2->detId);
       if (currDet.rpId() == tmpGh2Id.rpId())
         same_pot = true;
-      CLHEP::Hep3Vector subtraction = currPoint - it_gh2->globalPoint;
+      const auto subtraction = currPoint - it_gh2->globalPoint;
 
-      if (subtraction.perp() < roadRadius_ && same_pot) {  /// 1mm
+      if (subtraction.Rho() < roadRadius_ && same_pot) {  /// 1mm
```
perrotta: ```perp()``` and ```Rho()``` are not the same: are you fixing a bug here?

perrotta:
@jan-kaspar , although unrelated to this PR, the static analyzer points at what looks like a bug in ```DQM/CTPPS/plugins/CTPPSPixelDQMSource.cc```: at L702 it seems to me that
```C++
        if (clusize > ClusterSizeMax)
          clusize = ClusterSizeMax;
```
should be instead
```C++
        if (clusize > ClusterSizeMax)
          ClusterSizeMax = clusize;
```

Please check, and fix (if you agree with it)

jan-kaspar:
> @jan-kaspar , although unrelated to this PR, the static analyzer points at what looks like a bug in `DQM/CTPPS/plugins/CTPPSPixelDQMSource.cc`: at L702 it seems to me that

@fabferro @AndreaBellora can you please check the suggestion above?

RecoCTPPS/PixelLocal/src/RPixRoadFinder.cc
jan-kaspar: Despite the confusing names, perp() in CLHEP and Rho() in ROOT do the same, see e.g. https://root.cern.ch/root/html608/classROOT_1_1Math_1_1DisplacementVector3D.html#a92c6c1ddbb0e06be2aa87e2cc0eb0ddb
I've verified it by explicit print outs.

Geometry/VeryForwardRPTopology/interface/RPTopology.h
```diff
@@ -10,28 +10,21 @@
...
+#include "Math/Vector3D.h"
 
 /**
  *\brief Geometrical and topological information on RP silicon detector.
  * Uses coordinate a frame with origin in the center of the wafer.
  **/
 class RPTopology {
 public:
+  using Vector = ROOT::Math::DisplacementVector3D<ROOT::Math::Cartesian3D<double>>;
+  ;
```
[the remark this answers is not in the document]
jan-kaspar: Done in 8c7259c

perrotta:
please test

AndreaBellora:
> 
> 
> > @jan-kaspar , although unrelated to this PR, the static analyzer points at what looks like a bug in `DQM/CTPPS/plugins/CTPPSPixelDQMSource.cc`: at L702 it seems to me that
> 
> @fabferro @AndreaBellora can you please check the suggestion above?
As far as I can see, this line has not been introduced by the recent PR (#28714), but nevertheless it looks suspicious.
L699 and L701 seem to perform the same check. The fix would although not work, because ClusterSizeMax is a constexpr and thus can't be modified. 
I would assume that the check against ClusterSizeMax is done to set clusize to the maximum allowed if it gets higher than that. The histogram fill should then performed afterwards, otherwise it makes no sense to assign a new value to the variable that goes immediately out of scope.

The check against cluSizeMax is instead done for "monitoring" (reading the comment), but both cluSizeMax and multHitsMax appear to serve no purpose in the code. I would suggest to remove both variables and the if statement at L665. Furthermore move the if statement at L697 below the one at L 701.

jan-kaspar:
> The check against cluSizeMax is instead done for "monitoring" (reading the comment), but both cluSizeMax and multHitsMax appear to serve no purpose in the code. I would suggest to remove both variables and the if statement at L665. Furthermore move the if statement at L697 below the one at L 701.

Thanks @AndreaBellora ! Also to @perrotta , do you prefer to make these modifications in this PR or a dedicated one? Doing it here somewhat kills the "no change" character of this PR.

AndreaBellora:
> 
> 
> > The check against cluSizeMax is instead done for "monitoring" (reading the comment), but both cluSizeMax and multHitsMax appear to serve no purpose in the code. I would suggest to remove both variables and the if statement at L665. Furthermore move the if statement at L697 below the one at L 701.
> 
> Thanks @AndreaBellora ! Also to @perrotta , do you prefer to make these modifications in this PR or a dedicated one? Doing it here somewhat kills the "no change" character of this PR.

In case of a new PR I can follow with that. The change should be pretty quick.

perrotta:
> > > The check against cluSizeMax is instead done for "monitoring" (reading the comment), but both cluSizeMax and multHitsMax appear to serve no purpose in the code. I would suggest to remove both variables and the if statement at L665. Furthermore move the if statement at L697 below the one at L 701.
> > 
> > 
> > Thanks @AndreaBellora ! Also to @perrotta , do you prefer to make these modifications in this PR or a dedicated one? Doing it here somewhat kills the "no change" character of this PR.
> 
> In case of a new PR I can follow with that. The change should be pretty quick.

Yes, please make another "DQM only" PR.
I only wanted to report about my observation here.

jan-kaspar:
> > ...
> > In case of a new PR I can follow with that. The change should be pretty quick.
> 
> Yes, please make another "DQM only" PR.
> I only wanted to report about my observation here.

Thanks to both Andreas!

jfernan2:
+1

perrotta:
+1
- Technical pull request which cleans up and make uniform use of several vector classes in PPS reconstruction
- No changes in reco outputs are expected, and none is observed, also in the private dedicated validation reported in the PR description

perrotta:
@cvuosalo you already signed for geometry, and the last update only removed one useless empty line

cvuosalo:
+1

silviodonato:
+1

PR #28727, opened 2020-01-10
Fix reading memory after freed in HLTPerformanceInfo

Dr15Jones:
#### PR description:

The memory in the string was deleted at the end of the function so it is not safe to return as a const char*.

The problem was found by the clang static analyzer.

#### PR validation:

The code compiles.


Dr15Jones:
please test

Martin-Grunewald:
+1

silviodonato:
+1

PR #28574, opened 2019-12-06
 Introducing SiPixelCalSingleMuon ALCARECO

tvami:
Backport of #28535

cc:
@mmusich @tsusa @tocheng @christopheralanwest

PR #28731, opened 2020-01-11
Bugfix for PR28574

tvami:
#### PR description:

Fixing the issue reported in https://github.com/cms-sw/cmssw/pull/28574#issuecomment-573296950

#### PR validation:

Tested with `runTheMatrix.py -l 7.22,7.21,7.4,7.3 -t 4 -j 8`

This is a fix for a backport, PR #28574.

@silviodonato @mmusich @tsusa


silviodonato:
please test

chayanit:
+1

silviodonato:
please test workflow 136.726

silviodonato:
merge

boudoul:
Hello @tvami , I start to see a  dataset called /StreamExpressCosmics/Commissioning2019-PromptCalibProdSiPixel-Express-v1/ALCAPROMPT  in the current MWGR , this is what you introduced here , correct ? I guess this is already a good sign  - Of course  there is at this point no pixel in the run, but sounds like this config is ok  . You may want to x-check.

Btw @tvami and @silviodonato , it is appreciated (and useful!) to have better PR titles than this one. Titles are heavily used to understand quickly  what is in a given PR (at least the context : subdetector , etc..) , so just giving a number of a PR in the title is not explicit enough to understand the purpose of this bugfix . Just a suggestion for next time - Thank you.

mmusich:
>  I start to see a dataset called /StreamExpressCosmics/Commissioning2019-PromptCalibProdSiPixel-Express-v1/ALCAPROMPT in the current MWGR , this is what you introduced here , correct ? 

Hi @boudoul, no that's not correct. The sample you point out is the one created by the pixel bad components PCL, and being a PCL workflow runs at Express (in this case irrespective of the fact that pixel are in the run or not).
The new ALCARECO producers introduced here were not asked to be added to the Express configuration (that being pointless, after all there are no pixels in global run) hence we don't expect any new sample to appear in this MWGR...

boudoul:
oh ok, thanks I was simply confused , sorry for the noise

PR #28733, opened 2020-01-13
Removing unused variables from "CTPPSPixelDQMSource.cc" - fix suggested in PR #28726

AndreaBellora:
#### PR description:

This is a technical PR aimed at removing two unused variables as pointed out in PR #28726

#### PR validation:

Running

`runTheMatrix.py -l limited --ibeos`

gave

`34 33 32 26 16 4 1 1 1 tests passed, 0 0 0 0 0 0 0 0 0 failed`

### DIFF
--- a/DQM/CTPPS/plugins/CTPPSPixelDQMSource.cc
+++ b/DQM/CTPPS/plugins/CTPPSPixelDQMSource.cc
@@ -36,7 +36,8 @@ public:
 
 protected:
   void dqmBeginRun(edm::Run const &, edm::EventSetup const &) override;
-  void bookHistograms(DQMStore::IBooker &, edm::Run const &, edm::EventSetup const &) override;
+  void bookHistograms(DQMStore::IBooker &, edm::Run const &,
+                      edm::EventSetup const &) override;
   void analyze(edm::Event const &e, edm::EventSetup const &eSetup) override;
 
 private:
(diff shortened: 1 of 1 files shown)


AndreaBellora:
@fabferro @jan-kaspar

jfernan2:
please test

jfernan2:
+dqm
No impact

DQM/CTPPS/plugins/CTPPSPixelDQMSource.cc
```diff
@@ -694,12 +690,10 @@ void CTPPSPixelDQMSource::analyze(edm::Event const &event, edm::EventSetup const
       for (const auto &p : ds) {
         int clusize = p.size();
 
-        if (onlinePlots)
-          h2CluSize[arm][station]->Fill(prIndex(rpot, plane), clusize);
-        if (cluSizeMax < clusize)
-          cluSizeMax = clusize;
         if (clusize > ClusterSizeMax)
           clusize = ClusterSizeMax;
+        if (onlinePlots)
```
perrotta: One could gain something if this check is performed before the ```for``` loop at L690, as the whole loop has now only the scope of filling ```h2CluSize``` if ```onlinePlots```.

I would say more: the whole scope between L668 and L696, steered by 
```C++
  if (pixClus.isValid())
```
only makes two things:
- increment the bidimensional vector ```ClusMultPlane```;
- fill ```h2CluSize``` if ```onlinePlots```.

```ClusMultPlane``` is by now only filled in the code, but never used; as far as I can see that class member is not even accessed from elsewhere. One could further gain in code efficiency by:
- removing the so far useless array ```ClusMultPlane```;
- putting the check on  ```onlinePlots``` at the very beginning of the scope, i.e. replacing L667 with
```C++
  if (onlinePlots && pixClus.isValid())
```

AndreaBellora: You are completely right. By applying just a few changes here and there I didn't realize that there was another unused variable. Would you suggest fixing it and submitting another PR?

perrotta: > You are completely right. By applying just a few changes here and there I didn't realize that there was another unused variable. Would you suggest fixing it and submitting another PR?

If you plan to apply that further fix I would do it all in once here.

AndreaBellora: I'm fixing it now, so that everything goes in a single PR. Thanks

fabiocos:
@AndreaBellora @silviodonato I suggest to expand the title into something more informative (as it will become an integral part of the release notes)

AndreaBellora:
> 
> 
> @AndreaBellora @silviodonato I suggest to expand the title into something more informative (as it will become an integral part of the release notes)

Would "Removing unused variables from "CTPPSPixelDQMSource.cc" be a suitable title for the purpose?

silviodonato:
please test

jfernan2:
+1

silviodonato:
+1

PR #28735, opened 2020-01-14
Add parameter for treatment of charged high eta in PUPPI

ahinzmann:
PR description:

This is a followup of #28558, introducing a parameter to steer the eta-range for the treatment of charged particles not associated to vertices. This is to optimize PUPPI with phase-1 tracking. The performance was discussed here:
https://indico.cern.ch/event/876142/#5-update-on-puppi-tuning

This PR implements the necessary switches without changing the current behavior of the algorithm, i.e. no changes to RECO+MiniAOD are expected from this PR.

PR validation:

runtests

Backports

We would like to have these options backported to 10_6_X, to introduce a new PUPPI configuration for UL MiniAOD-v2 and for JMEcustomNanoAODs as soon as the PR is merged.


slava77:
@cmsbuild please test

CommonTools/PileupAlgos/plugins/PuppiProducer.h
```diff
@@ -51,6 +51,7 @@ class PuppiProducer : public edm::stream::EDProducer<> {
   bool fUseDZ;
   float fDZCut;
   double fPtMaxCharged;
+  double fEtaMaxCharged;
   double fPtMaxPhotons;
   double fEtaMaxPhotons;
   double fPtMaxNeutrals;
```
perrotta: ```fPtMaxNeutrals``` and ```fPtMaxNeutralsStartSlope``` are not used here, only in PuppiContainer: you can remove from PuppiProducer then

CommonTools/PileupAlgos/plugins/PuppiProducer.cc
```diff
@@ -351,6 +356,7 @@ void PuppiProducer::fillDescriptions(edm::ConfigurationDescriptions& description
   desc.add<bool>("UseDeltaZCut", true);
   desc.add<double>("DeltaZCut", 0.3);
   desc.add<double>("PtMaxCharged", 0.);
+  desc.add<double>("EtaMaxCharged", 0.);
```
perrotta: I see some potential (even if perhaps unlikely) danger due to rounding approximations in setting these parameters at "0", when there are checks like
```C++
   if ((fPtMaxCharged > 0) ...
```

Wouldn't be identical, for the point of view of the results, but safer, setting the value to "-1" (e.g) instead of "0" for the following parameters?
- PtMaxCharged
- EtaMaxCharged
- PtMaxPhotons

CommonTools/PileupAlgos/plugins/PuppiProducer.cc
```diff
@@ -161,6 +162,8 @@ void PuppiProducer::produce(edm::Event& iEvent, const edm::EventSetup& iSetup) {
             pReco.id = 0;
             if ((fPtMaxCharged > 0) and (pReco.pt > fPtMaxCharged))
               pReco.id = 1;
+            else if ((fEtaMaxCharged > 0) and (std::abs(pReco.eta) > fEtaMaxCharged))
```
perrotta: When one wants to say "do nothing", if one sets "EtaMaxCharged" to some large value (instead of 0), this line could be replaced by the more intuitive
```C++
            else if  (std::abs(pReco.eta) > fEtaMaxCharged)
``` 
(which is by the way what is done later on in this same code for "EtaMaxPhotons")

ahinzmann: done

CommonTools/PileupAlgos/plugins/PuppiProducer.h
ahinzmann: done

CommonTools/PileupAlgos/plugins/PuppiProducer.cc
```diff
@@ -160,9 +158,9 @@ void PuppiProducer::produce(edm::Event& iEvent, const edm::EventSetup& iSetup) {
             pReco.id = 1;
           } else if (tmpFromPV == 1 || tmpFromPV == 2) {
             pReco.id = 0;
-            if ((fPtMaxCharged > 0) and (pReco.pt > fPtMaxCharged))
+            if (pReco.pt > fPtMaxCharged)
```
perrotta: I think this should be reverted. As such it will not reproduce the previous behavior in case of (pathologically) large pt's. And in any case, as such it creates an asymmetry wrt how ```PtMaxPhotons``` is dealt with.
My suggestion was simply to assign a value "-1" instead of "0" to the  ```PtMaxCharged``` and  ```PtMaxPhotons``` parameters, so that the condition ```(fPtMaxCharged > 0)``` is always satisfied, even in the unlikely case of floating point upward fluctuations of "0".

CommonTools/PileupAlgos/plugins/PuppiProducer.cc
```diff
@@ -191,9 +189,9 @@ void PuppiProducer::produce(edm::Event& iEvent, const edm::EventSetup& iSetup) {
           } else if (lPack->fromPV() == (pat::PackedCandidate::PVTight) ||
                      lPack->fromPV() == (pat::PackedCandidate::PVLoose)) {
             pReco.id = 0;
-            if ((fPtMaxCharged > 0) and (pReco.pt > fPtMaxCharged))
+            if (pReco.pt > fPtMaxCharged)
```
perrotta: Same here

CommonTools/PileupAlgos/plugins/PuppiProducer.cc
```diff
@@ -355,12 +353,10 @@ void PuppiProducer::fillDescriptions(edm::ConfigurationDescriptions& description
   desc.add<bool>("UseFromPVLooseTight", false);
   desc.add<bool>("UseDeltaZCut", true);
   desc.add<double>("DeltaZCut", 0.3);
-  desc.add<double>("PtMaxCharged", 0.);
-  desc.add<double>("EtaMaxCharged", 0.);
+  desc.add<double>("PtMaxCharged", 7000.);
```
perrotta: As I said, 7000 cannot guarantee that all particles, even the pathological ones, are kept

CommonTools/PileupAlgos/plugins/PuppiProducer.cc
```diff
@@ -355,12 +353,10 @@ void PuppiProducer::fillDescriptions(edm::ConfigurationDescriptions& description
...
+  desc.add<double>("EtaMaxCharged", 10.);
```
perrotta: That's in principle fine.
Perhaps, to make clear to everyone reading the config that you want to keep "all eta's" you should rather use an obviously large number, like (e.g.) "99999'"

CommonTools/PileupAlgos/plugins/PuppiProducer.cc
```diff
@@ -374,6 +370,7 @@ void PuppiProducer::fillDescriptions(edm::ConfigurationDescriptions& description
   desc.add<double>("MinPuppiWeight", .01);
 
   PuppiAlgo::fillDescriptionsPuppiAlgo(desc);
+  PuppiContainer::fillDescriptionsPuppiContainer(desc);
```
perrotta: My review request was simply to remove from this class the local variables ```fPtMaxNeutrals``` and ```fPtMaxNeutralsStartSlope```, because they are not used here.

I'm not against to moving the definition of the parameters specific to PuppiContainer to the ```fillDescriptions``` method of that class, as you are doing here. However, if you do so you must do the same with all parameters specific to PuppiContainer, not only these two (there are a few more).

To keep the intervention at its minimum you could however simply revert to the previous behavior and have all parameters defined again inside PuppiProducer

CommonTools/PileupAlgos/python/Puppi_cff.py
```diff
@@ -31,6 +31,7 @@
                        UseDeltaZCut   = cms.bool(True),
                        DeltaZCut      = cms.double(0.3),
 		       PtMaxCharged   = cms.double(0.),
+		       EtaMaxCharged   = cms.double(0.),
```
perrotta: Please don't forger to update also here, as this config will overwrite the setting defined in the ```fillDescriptions``` methods

ahinzmann: all done

CommonTools/PileupAlgos/python/Puppi_cff.py
```diff
@@ -30,7 +30,8 @@
                        UseFromPVLooseTight = cms.bool(False),
                        UseDeltaZCut   = cms.bool(True),
                        DeltaZCut      = cms.double(0.3),
-		       PtMaxCharged   = cms.double(0.),
+		       PtMaxCharged   = cms.double(-1.),
+		       EtaMaxCharged   = cms.double(99999.),
 		       PtMaxPhotons = cms.double(0),
```
perrotta: For analogy, please also initialize to "-1" ```PtMaxPhotons```, both here and in the fillDescrption method

perrotta:
please test

perrotta:
+1
- A new parameter added in PuppiProducer to select charged particle up to a give pseudorapidity
- The default value set for it reproduces the behavior of the producer without it

santocch:
+1

silviodonato:
+1

PR #28742, opened 2020-01-15
PPS: DQM update

jan-kaspar:
#### PR description:

  * better structure of PPS DQM sequences
  * introduced flag to enable/disable creation of proton-reconstruction plots
  * adjusted range of proton time histogram
  * strip-RP plots removed from the online and offline sequences since these RPs will not be used in Run 3 (the plots still remain in the "calibration" sequence for special runs)

#### PR validation:

Running
```
runTheMatrix.py --ibeos -l 136.731,136.7611,136.788,136.8311,136.85
```
gave
```
5 5 5 3 tests passed, 0 0 0 0 failed
```

This PR doesn't change the results of the reconstruction - see the plot below - blue = before the PR, red = with this PR.

make_cmp

### DIFF
--- a/DQM/CTPPS/plugins/CTPPSCommonDQMSource.cc
+++ b/DQM/CTPPS/plugins/CTPPSCommonDQMSource.cc
@@ -51,6 +51,8 @@ private:
   const edm::EDGetTokenT<std::vector<CTPPSLocalTrackLite>> tokenLocalTrackLite;
   const edm::EDGetTokenT<std::vector<reco::ForwardProton>> tokenRecoProtons;
 
+  bool makeProtonRecoPlots_;
+
   int currentLS;
   int endLS;
 
@@ -90,7 +92,7 @@ private:
 
     ArmPlots() {}
 
-    ArmPlots(DQMStore::IBooker &ibooker, int _id);
+    ArmPlots(DQMStore::IBooker &ibooker, int _id, bool makeProtonRecoPlots);
   };
 
   std::map<unsigned int, ArmPlots> armPlots;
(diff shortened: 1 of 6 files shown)


jfernan2:
please test

jfernan2:
@jan-kaspar since these changes affect Online DQM which is not tested in githug, have you tested the PR in a Online DQM configuration?
runTheMatrix does NOT check Online DQM in any workflow
Thanks

jan-kaspar:
> @jan-kaspar since these changes affect Online DQM which is not tested in githug, have you tested the PR in a Online DQM configuration?
> runTheMatrix does NOT check Online DQM in any workflow

Sort of. I tried directly by putting `True` here:
https://github.com/CTPPS/cmssw/blob/pps_dqm_update/DQM/Integration/python/clients/ctpps_dqm_sourceclient-live_cfg.py#L6
but it didn't work out of the box: there was a "list index out of range" when quering DAS, then PPSTimingCalibrationRcd was not found (so I had to use a different GT) ... so I had to modify the config a little to get it work. For sure, the reorganisation of the PPS DQM sequences introduced in this PR doesn't seem to pose any problem.

jan-kaspar:
> @jan-kaspar despite you made makeProtonRecoPlots = True in DQMOffline, you are removing 1k plots, can you please check?

This PR removes the plots from tracking-strip RPs as these will not be used in Run 3 (in normal runs). I checked your link and this is exactly what happens for WFs 136.731 and 136.788.

I realise that I didn't mention this in the PR description, will fix it.

jfernan2:
+1

silviodonato:
+1

PR #28751, opened 2020-01-16
Bugfix: Crash in DaqTestHistograms

smorovic:
#### PR description:

Problem was introduced in #28092 commit:
bb6f53fef153630ef609802597b34ab3b495bd19
Fixes (removes) a call into uninitialized pointer.

#### PR validation:

Verified to work in DAQ (HLT) test environment.
PR will be issued with backport to 110X.

### DIFF
--- a/DQM/HLTEvF/plugins/DaqTestHistograms.cc
+++ b/DQM/HLTEvF/plugins/DaqTestHistograms.cc
@@ -20,7 +20,7 @@ namespace {
 
     RunBasedHistograms()
         :  // overall event count and event types
-          events_processed(),
+          events_processed(nullptr),
           element_array() {}
   };
 }  // namespace
@@ -63,7 +63,6 @@ DaqTestHistograms::DaqTestHistograms(edm::ParameterSet const &config)
 void DaqTestHistograms::dqmBeginRun(edm::Run const &run,
                                     edm::EventSetup const &setup,
                                     RunBasedHistograms &histograms) const {
-  histograms.events_processed->Reset();
   histograms.element_array.resize(m_num_histograms);
 }
 


jfernan2:
please test

silviodonato:
As far as I remember, you can fix the online DQM code at P5 without creating a new CMSSW release. Is it correct? I'm trying to understand how much this PR is urgent. The MWGR will use CMSSW_10_6_8_patch1 and is scheduled on Tuesday (January 21)

silviodonato:
I see there is no bug in CMSSW_10_6_8_patch1. False alarm :-) 
https://github.com/cms-sw/cmssw/blob/CMSSW_10_6_8_patch1/DQM/HLTEvF/plugins/DaqTestHistograms.cc#L23

jfernan2:
@silviodonato Anyway, yes, you were correct: we can take a PR and integrate it on P5 DQM Machines on top of a existing CMSSW release.
However, to avoid developers to take a wrong code, it is good to have it integrated sooner or later so that it does NOT become a "hanging PR" in the future. 
MWGR is not afected as you realized, we used development CMSSW releases for the DQM core changes in order to let 10_X stable. Thanks
FYI @schneiml @andrius-k

smorovic:
this is module is not used in production DQM or HLT. It is a DAQ test module for generating histograms in HLT farm in test runs.

It is not necessary to merge it to 10_X. However I see it is already in the CMSSW_10_6_8_patch1 so it's fine (we will probably use it only from 11_X though).

No patch release or prerelease is needed either for 11_0 /11_1_X there is no urgency.

jfernan2:
+1

Martin-Grunewald:
+1

silviodonato:
+1

PR #28753, opened 2020-01-16
Fixed sign in pat::Muon::dB() to match one obtained from the best track definition 

cericeci:
**PR description:**

The the transverse impact parameter for pat::Muon was being computed using the IPTools package. It takes the sign from the scalar product of the tangent vector to the best track at the point of closest approach to the PCA to PV vector. While this makes sense for complex objects like jets, for muon tracks it should be ~0 by definition.  Thus, while the absolute dxy value makes sense the sign did not. This fixes it by switching to use the same IP sign as the dxy obtained from the best track.

This has been noted and later corrected from the nanoAOD side in
[https://github.com/cms-nanoAOD/cmssw/pull/475](https://github.com/cms-nanoAOD/cmssw/pull/475)

An open JIRA ticket was opened discussing the problem in the muon group
[https://its.cern.ch/jira/browse/CMSMUONS-311](https://its.cern.ch/jira/browse/CMSMUONS-311)

**Expected changes**

The sign obtained from pat::Muon.dB(pat::Muon::PV2D) and pat::Muon.dB(pat::Muon::BS2D) should now match what could be obtained from the track collection as pat::Muon.bestTrack()->dxy(PV _or_ BS). Expect no changes in the absolute value itself.

**PR validation**

Tested locally in a couple of Z->mumu RelVal AOD samples obtaining the desired results.
No additional changes from code checks.
Minimal integration checks from runTheMatrix.py -l limited -i all  passed.

### DIFF
--- a/DataFormats/TrackReco/src/TrackBase.cc
+++ b/DataFormats/TrackReco/src/TrackBase.cc
@@ -144,3 +144,20 @@ TrackBase::TrackAlgorithm TrackBase::algoByName(const std::string &name) {
   // cast
   return TrackAlgorithm(index);
 }
+
+double TrackBase::dxyError(Point const &vtx, math::Error<3>::type const &vertexCov) const {
+  // Gradient of TrackBase::dxy(const Point &myBeamSpot) with respect to track parameters
+  TrackBase::ParameterVector trackGrad =
+      TrackBase::ParameterVector(0,
+                                 0,
+                                 (vtx.x() * px() + vtx.y() * py()) / pt(),  // x_vert * cos(phi) + y_vert * sin(phi)
+                                 1,
+                                 0);
+  // Gradient with respect to point parameters
+  math::Vector<3>::type pointGrad = math::Vector<3>::type(py() / pt(),   // sin(phi)
+                                                          -px() / pt(),  // -cos(phi)
+                                                          0);
+  // Propagate covariance assuming cross-terms of the covariance between track and vertex parameters are 0
+  return std::sqrt(ROOT::Math::Dot(trackGrad, covariance() * trackGrad) +
+                   ROOT::Math::Dot(pointGrad, vertexCov * pointGrad));
+}
--- a/DataFormats/TrackReco/src/TrackBase.cc
+++ b/DataFormats/TrackReco/src/TrackBase.cc
@@ -146,18 +146,14 @@ TrackBase::TrackAlgorithm TrackBase::algoByName(const std::string &name) {
 }
 
 double TrackBase::dxyError(Point const &vtx, math::Error<3>::type const &vertexCov) const {
-  // Gradient of TrackBase::dxy(const Point &myBeamSpot) with respect to track parameters
-  TrackBase::ParameterVector trackGrad =
-      TrackBase::ParameterVector(0,
-                                 0,
-                                 (vtx.x() * px() + vtx.y() * py()) / pt(),  // x_vert * cos(phi) + y_vert * sin(phi)
-                                 1,
-                                 0);
+  // Gradient of TrackBase::dxy(const Point &myBeamSpot) with respect to track parameters. Using unrolled expressions to avoid calling for higher dimension matrices
+  // ( 0, 0, x_vert * cos(phi) + y_vert * sin(phi), 1, 0 )
   // Gradient with respect to point parameters
-  math::Vector<3>::type pointGrad = math::Vector<3>::type(py() / pt(),   // sin(phi)
-                                                          -px() / pt(),  // -cos(phi)
-                                                          0);
+  // ( sin(phi), -cos(phi))
   // Propagate covariance assuming cross-terms of the covariance between track and vertex parameters are 0
-  return std::sqrt(ROOT::Math::Dot(trackGrad, covariance() * trackGrad) +
-                   ROOT::Math::Dot(pointGrad, vertexCov * pointGrad));
+  return std::sqrt((vtx.x() * px() + vtx.y() * py()) * (vtx.x() * px() + vtx.y() * py()) / (pt() * pt()) *
+                       covariance(i_phi, i_phi) +
+                   2 * (vtx.x() * px() + vtx.y() * py()) / pt() * covariance(i_phi, i_dxy) + covariance(i_dxy, i_dxy) +
+                   py() * py() / (pt() * pt()) * vertexCov(0, 0) - 2 * py() * px() / (pt() * pt()) * vertexCov(0, 1) +
+                   px() * px() / (pt() * pt()) * vertexCov(1, 1));
 }
--- a/PhysicsTools/PatAlgos/plugins/PATMuonProducer.cc
+++ b/PhysicsTools/PatAlgos/plugins/PATMuonProducer.cc
@@ -1102,7 +1102,7 @@ void PATMuonProducer::embedHighLevel(pat::Muon& aMuon,
   double d0_corr = result.second.value();
   double d0_err = primaryVertexIsValid ? result.second.error() : -1.0;
   // Now correct the sign using information from the track
-  d0_corr = (track->dxy() > 0 ? 1 : -1)*fabs(d0_corr);
+  d0_corr = (track->dxy(primaryVertex.position()) > 0 ? 1 : -1)*fabs(d0_corr);
   aMuon.setDB(d0_corr, d0_err, pat::Muon::PV2D);
 
   // PV3D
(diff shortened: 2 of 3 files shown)


PhysicsTools/PatAlgos/plugins/PATMuonProducer.cc
```diff
@@ -1101,6 +1101,8 @@ void PATMuonProducer::embedHighLevel(pat::Muon& aMuon,
       IPTools::signedTransverseImpactParameter(tt, GlobalVector(track->px(), track->py(), track->pz()), primaryVertex);
   double d0_corr = result.second.value();
   double d0_err = primaryVertexIsValid ? result.second.error() : -1.0;
+  // Now correct the sign using information from the track
+  d0_corr = (track->dxy(primaryVertex.position()) > 0 ? 1 : -1) * fabs(d0_corr);
```
slava77: this can be 

```suggestion
  d0_corr = std::copysign(d0_corr, track->dxy(primaryVertex.position()));
```

slava77: actually, I'd modify on L1102 instead. There is no significant need to assign one value and then to apply this correction.

arizzi:
I guess we should just use the value returned by bestTrack->dxy(pv / bs ) instead of calling IPTools ...
The linearization made in IPTools from the PCA is meaningless anyhow without lifetime

slava77:
> I guess we should just use the value returned by bestTrack->dxy(pv / bs ) instead of calling IPTools ...

even better. IPTools are not the cheapest tool to use computationally. So, if we can read just dxy, better use that.

slava77:
> > I guess we should just use the value returned by bestTrack->dxy(pv / bs ) instead of calling IPTools ...
> 
> even better. IPTools are not the cheapest tool to use computationally. So, if we can read just dxy, better use that.

@cericeci please comment on the status of this PR regarding the latest comments.
Thank you.

cericeci:
Hi @slava77, I updated in the JIRA ticket but forget to do it here.
While the definition of the sign itself was checked, I wanted to check that there is no real case why IPTools was used. I'll need a day or two to check it all. Should we close the PR and later open another?

slava77:
> I'll need a day or two to check it all. Should we close the PR and later open another?

Thank you for the clarification.
I don't have a strong preference about open/reopen. Feel free to just leave this open and update here when the new code is ready.

cericeci:
It took some time to convince myself that there is no caveat in using the best track dxy instead of IPTools. I ran over a mixture (i.e. isolated, non isolated, displaced, etc.) of ~200M events and see around ~100 muons in which both are not compatible without uncertainties. Those have displacement >= 40cm so it seems that the linear approximations start to fail anyway.

A problem arised when computing the uncertainty in the dxy measurement, as using reco::Track::dxyError() would ignore the -small- uncertainty in the vertex coordinates that is already in the IPTools approach. A quadratic sum of track uncertainty + vertex uncertainty would be an overestimation as one needs to correct by the relative direction of the track (in the curved coordinates) with respect to the position of the vertex. In other words, propagate uncertainties over this formula

https://github.com/cms-sw/cmssw/blob/8a74ab5f53ead2d4f02abe808031941d78449bad/DataFormats/TrackReco/interface/TrackBase.h#L742

I see that this functionality is not available in the reco::TrackBase object. Is it somewhere in other package I might be missing? If not I can add it to reco::TrackBase as it is of general purpose, not only restricted to muons

slava77:
I think that having this in the TrackBase might work.
We can't easily pass a reco::Vertex to the TrackBase member methods due to a circular dependence.
I guess the interface will be ```dxyError(Point const& vtx, math::Error<3>::type const& vertexCov)```

@VinInn @mtosi @arizzi please check/comment.
Thank you.

DataFormats/TrackReco/src/TrackBase.cc
```diff
@@ -144,3 +144,20 @@ TrackBase::TrackAlgorithm TrackBase::algoByName(const std::string &name) {
   // cast
   return TrackAlgorithm(index);
 }
+
+double TrackBase::dxyError(Point const &vtx, math::Error<3>::type const &vertexCov) const {
```
cericeci: This is the proposed way to propagate uncertainties properly for the dxy measurement with respect to a given PV. 

Going from the original frame (dxy) to the new frame (dxy') translated from PV coordinates (xv, yv)  it uses the transformation:
dxy' = dxy + x_v*sin(phi) - y_v*cos(phi)

And propagates accordingly using the track and vertex covariance matrices plus gradients

(0,0,xv*cos(phi) + yv*sin(phi), 1, 0) (gradient with respect to the track parameters in the curvilinear basis)

(sin(phi), -cos(phi), 0) (gradient with respect to the vertex parameters)

slava77:
@cmsbuild please test

PhysicsTools/PatAlgos/plugins/PATMuonProducer.cc
```diff
@@ -1097,27 +1097,24 @@ void PATMuonProducer::embedHighLevel(pat::Muon& aMuon,
...
 
   // PV2D
+  aMuon.setDB(track->dxy(primaryVertex.position()),
+              track->dxyError(primaryVertex.position(), primaryVertex.covariance()),
+              pat::Muon::PV2D);
+
+  // PV3D
   std::pair<bool, Measurement1D> result =
-      IPTools::signedTransverseImpactParameter(tt, GlobalVector(track->px(), track->py(), track->pz()), primaryVertex);
+      IPTools::signedImpactParameter3D(tt, GlobalVector(track->px(), track->py(), track->pz()), primaryVertex);
```
slava77: doesn't the 3D also have the same problem as the 2D case?
Also, does this agree (in a linear approximation) with the reco::Track::dsz ?

DataFormats/TrackReco/src/TrackBase.cc
```diff
@@ -144,3 +144,20 @@ TrackBase::TrackAlgorithm TrackBase::algoByName(const std::string &name) {
...
+                                 (vtx.x() * px() + vtx.y() * py()) / pt(),  // x_vert * cos(phi) + y_vert * sin(phi)
+                                 1,
+                                 0);
+  // Gradient with respect to point parameters
+  math::Vector<3>::type pointGrad = math::Vector<3>::type(py() / pt(),   // sin(phi)
+                                                          -px() / pt(),  // -cos(phi)
+                                                          0);
+  // Propagate covariance assuming cross-terms of the covariance between track and vertex parameters are 0
+  return std::sqrt(ROOT::Math::Dot(trackGrad, covariance() * trackGrad) +
+                   ROOT::Math::Dot(pointGrad, vertexCov * pointGrad));
```
slava77: both can be expanded; at least the first term should be to avoid the 5x5 matrix operations, while only 2x2 is actually touched.

cericeci: Done, added a comment to at least make it easier to understand the formula.

PhysicsTools/PatAlgos/plugins/PATMuonProducer.cc
cericeci: Yes, the difference in the sign between that of the best track and the one obtained by IPTools remains. Nonetheless, the sign itself is not used elsewhere as that of the 2D version is. I understand that using track information would be preferred from the usage of computational resources point of view?

From a quick set of tests I see, surprisingly, that in most cases (~90%) both computations of the 3D IP agree, but there is a significant set of deviations between the tracks' dsz and iptools' value for ~10% of the events. This seems to be correlated with the linear approximation but not completely caused by it (i.e. it happens for several cases in which vertex ~ (0,0,0) and with relatively high pT/low bending). 

The sip3D = abs(d3d)/error(d3d) variable is used down the line in terms of ID criteria. I.e. for the prompt MVA discriminant (https://github.com/cms-sw/cmssw/blob/e4c6faa3322fcebf71f864322ab1a20cb00d7604/PhysicsTools/PatAlgos/src/MuonMvaEstimator.cc#L70 ). This deserves further studies in terms of muon selection optimality before changing it which goes slightly out of scope for this PR. Should we put it as a separate issue?

slava77: > Should we put it as a separate issue?

OK.
Thank you for checking.

slava77:
@cmsbuild please test

slava77:
+1

for https://github.com/cms-sw/cmssw/pull/28753	8dd5d67 
- jenkins tests pass and comparisons with the baseline show differences in pat muon DB (e.g. in nanoAOD muon dxy and dxyError)  or downstream in deepTau ID (which uses these 2D muon dxy and error variables)

silviodonato:
+1
@peruzzim @arizzi FYI

santocch:
+1

PR #28758, opened 2020-01-17
Update SimTrackSimVertexDumper to use tokens

fabiocos:
#### PR description:

Fix for the simple test analyzer class SimTrackSimVertexDumper, similar to that previously applied to the hit dumper. The collections are accessed now using tokens, and the useless separate header is now integrated into the code.

#### PR validation:

Tested on the step1 of workflow 22007, it produces the expected output (previously it was broken).


fabiocos:
please test

testing nothing apart for build, this code is used nowhere in regular production

civanch:
+1

silviodonato:
+1

PR #26094, opened 2019-03-07
Simps to customphysics 7 1

lowette:
#### PR description:

This PR includes the code necessary to simulate SIMPs, implementing them as heavy neutrons with a configurable mass and as only interaction a clone of Geant's neutron's inelastic process.
This is very specific code for signal generation of the EXO-17-010 analysis. More information can be found in this presentation: https://indico.cern.ch/event/792362/contributions/3299414/attachments/1791002/2917852/190205_Lowette_SIMPSimulation.pdf

#### PR validation:

This code has been validated, among other things, by showing this SIMP simulation at 1GeV to give identical results to a simulation with a real Geant neutron replacing the SIMP. More information also in the talk mentioned above.

Note that this code has so far only been used to simulate a signal in 7_1_X, thus the initial PR  in this branch. Next, it will be forward-ported to 10_6_X.

PR #28763, opened 2020-01-20
Forward port of Geant4 SIMP custom physics model to master branch 

fabiocos:
#### PR description:

Port of #26094 (targeting 7_1_X) to master, based on the original proposal by @lowette .
The code has been adapted to the new 11_1_X branch trying to mimic as closely as possible the original proposal. The  ```G4GlauberGribovCrossSection``` class, no more present in Geant4, has been replaced by the available ```G4ComponentGGHadronNucleusXsc```. Include files have been moved to ```interface```, the ```G4``` names have been turned into ```CMS``` as done for many other models, and include guards have been updated according to the standard we adopt.

#### PR validation:

Code compiles and passes ```scram b code-checks```, but no dedicated test workflow is provided to probe the real code behaviour in the original proposal. @civanch should confirm the correctness of the adopted approach, and possible tests to be performed.

### DIFF
--- a/SimG4Core/CustomPhysics/interface/CMSSIMPInelasticProcess.h
+++ b/SimG4Core/CustomPhysics/interface/CMSSIMPInelasticProcess.h
@@ -1,0 +1,167 @@
+#ifndef SimG4Core_CustomPhysics_CMSSIMPInelasticProcess_H
+#define SimG4Core_CustomPhysics_CMSSIMPInelasticProcess_H
+
+#include "G4HadronicProcess.hh"
+
+class G4ParticleDefinition;
+
+class CMSSIMPInelasticProcess : public G4HadronicProcess {
+public:
+  CMSSIMPInelasticProcess(const G4String& processName = "SIMPInelastic");
+
+  ~CMSSIMPInelasticProcess() override;
+
+  G4bool IsApplicable(const G4ParticleDefinition& aParticleType) override;
+
+  // register generator of secondaries
+  void RegisterMe(G4HadronicInteraction* a);
+
+  // get cross section per element
+  inline G4double GetElementCrossSection(const G4DynamicParticle* part,
+                                         const G4Element* elm,
+                                         const G4Material* mat = nullptr) {
+    G4double x = theCrossSectionDataStore->GetCrossSection(part, elm, mat);
+    if (x < 0.0) {
+      x = 0.0;
+    }
+    return x;
+  }
+
+  // obsolete method to get cross section per element
+  inline G4double GetMicroscopicCrossSection(const G4DynamicParticle* part,
+                                             const G4Element* elm,
+                                             const G4Material* mat = nullptr) {
+    return GetElementCrossSection(part, elm, mat);
+  }
+
+  // generic PostStepDoIt recommended for all derived classes
+  G4VParticleChange* PostStepDoIt(const G4Track& aTrack, const G4Step& aStep) override;
+
+  // initialisation of physics tables and CMSSIMPInelasticProcessStore
+  void PreparePhysicsTable(const G4ParticleDefinition&) override;
+
+  // build physics tables and print out the configuration of the process
+  void BuildPhysicsTable(const G4ParticleDefinition&) override;
+
+  // dump physics tables
+  inline void DumpPhysicsTable(const G4ParticleDefinition& p) { theCrossSectionDataStore->DumpPhysicsTable(p); }
+
+  // add cross section data set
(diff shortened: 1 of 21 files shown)


fabiocos:
please test

SimG4Core/CustomPhysics/src/CMSQGSPSIMPBuilder.cc
```diff
@@ -0,0 +1,57 @@
...
+  theMin = 12 * GeV;
+
+  theModel = new G4TheoFSGenerator("QGSP");
+
+  theStringModel = new G4QGSModel<G4QGSParticipants>;
+  theStringDecay = new G4ExcitedStringDecay(theQGSM = new G4QGSMFragmentation);
+  theStringModel->SetFragmentationModel(theStringDecay);
+
+  theCascade = new G4GeneratorPrecompoundInterface;
+  thePreEquilib = new G4PreCompoundModel(new G4ExcitationHandler);
```
civanch: @fabiocos , lines 22-23 should be removed.

SimG4Core/CustomPhysics/src/CMSQGSPSIMPBuilder.cc
```diff
@@ -0,0 +1,57 @@
...
+    theModel->SetQuasiElasticChannel(theQuasiElastic);
+  } else {
+    theQuasiElastic = nullptr;
+  }
+}
+
+CMSQGSPSIMPBuilder::~CMSQGSPSIMPBuilder() {
+  delete theStringDecay;
+  delete theStringModel;
+  delete thePreEquilib;
```
civanch: @fabiocos , lines 38-39 should be removed (delete thePreEquilib; delete theCascade)

SimG4Core/CustomPhysics/interface/CMSSIMPInelasticProcess.h
```diff
@@ -0,0 +1,167 @@
+#ifndef SimG4Core_CustomPhysics_CMSSIMPInelasticProcess_H
+#define SimG4Core_CustomPhysics_CMSSIMPInelasticProcess_H
+
+#include "G4HadronicProcess.hh"
+
+class G4ParticleDefinition;
+
+class CMSSIMPInelasticProcess : public G4HadronicProcess {
```
civanch: @fabiocos , most of virtual methods must be removed - they are implemented inside G4HadronicProcess. In different Geant4 releases they may be different.

civanch: @fabiocos , The code was working in 10.0 where we run in single threading mode, in 10.6 many things were changed except interfaces, so the code is well compiling but there may be problems, which I cannot catch promptly.

civanch:
@fabiocos , would you agree if we merge this PR and after I will try to shape these classes to recent Geant4? I foresee in this 7_1 implementation some thread unsafe things not seen in an easy way. Suspect, that it would be ineffective if will put too many comments.

civanch:
+1
@fabiocos , I would think, that if we do forward port we should not change code a lot. So, let us integrate this but add modifications on top.

fabiocos:
@civanch thanks for your comment, as said mine was a basic attempt to do the porting with the minimal needed code change. I might easily implement the first comments of yours, but if more is needed, may be it is better to follow your suggestion, let @silviodonato integrate this (it will not break anyway anything as far as I can see) and let you clean the code in a second step. Just to be sure, do we have some test workflow using CustomPhysics, just to have a small cross check?

silviodonato:
+1

smuzaffar:
@fabiocos , @civanch looks like this broke GEANT4 IBs 
https://cmssdt.cern.ch/SDT/cgi-bin/buildlogs/slc7_amd64_gcc820/CMSSW_11_1_GEANT4_X_2020-01-27-1100/SimG4Core/CustomPhysics

civanch:
@smuzaffar , thanks! 

I expect problems for GEANT4 IB but they happens on dramatically high level. Will try to make next PR asap. The reason of the problem: author of this code rewrite all Geant4 base class virtual methods for physics and x-sections with old variants of 6 years ago. Because he is using inheritance, he should rewrite instead only few methods, which really required his modifications.

fabiocos:
@civanch @smuzaffar as far as I can see, the issue to be solved in the code cleaning is the redefinition of the particle performed in https://github.com/cms-sw/cmssw/blob/master/SimG4Core/CustomPhysics/src/CMSSIMPInelasticProcess.cc#L145
If that could be moved inside the particle definition it would be straightforward to clean the CMSSIMPInelasticProcess class.

PR #28777, opened 2020-01-22
HGCAL: fix interpolation boundaries for ToA

amartelli:
This PR 
- fixes a bug spotted by @adas1994 in the computation of the time-of-arrival in the HGCAL digi
- adds some comments to help the understanding of the code
- fixes in the way in which the PCaloHits are cumulated, to deal with PCaloHIts having exactly the same time stamps (very few cases only observed in high PU)

Quick reminder (how ToA firing works):
- In each Silicon cell, the ToA is fired when the cumulative charge is at threshold (12fC).
- Since a full pulse shape is not available, the single time stamps deposited by G4 in each cell (each with a time and a charge) are used:
      - the time-sorted list of time stamps is considered, and their cumulative charge computed.
      - the 2 time stamps with cumulative charge around the threshold (the one before, and the one after) are picked as reference
       - the ToA is extracted with the linear interpolation between the 2 chosen time-stamps, taking the value at which the interpolation crosses the 12fC threshold.

Explanation of  the bug fix:
- The bug is in the interpolation of the cumulated charge vs time, to compute the ToA at threshold, 
and consists in the further cumulation of the already cumulated charges.
- The first cumulation of time-stamps is correct
- The identification of the cells with ToA available is correct
- The bug only involves the identification of the time stamps used to interpolate and extract the ToA

We expect an impact if the delay among time stamps is "big", and this is what happens with PU.
The validation run, comparing the performance between the TDR version (with bug) and this PR version show a negligible impact also at 200PU.

Bug, fix and validation documented in the slides:
https://cernbox.cern.ch/index.php/s/gMcLoFLjHXBpNdi

For reference @rovere

### DIFF
--- a/SimCalorimetry/HGCalSimProducers/src/HGCDigitizer.cc
+++ b/SimCalorimetry/HGCalSimProducers/src/HGCDigitizer.cc
@@ -456,7 +456,7 @@ void HGCDigitizer::accumulate(edm::Handle<edm::PCaloHitContainer> const& hits,
                                                  : std::pair<float, float>((findPos - 1)->first + charge, tof));
 	}
 
-	//cumulate the charge of new entry for all elements that follows in the sorted list
+	//cumulate the charge of new entry for all elements that follow in the sorted list
 	//and resize list accounting for cases when the inserted element itself crosses the threshold
         for (std::vector<std::pair<float, float>>::iterator step = insertedPos; step != hitRefs_bx0[id].end();
              ++step) {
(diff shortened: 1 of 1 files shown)


civanch:
please test

kpedro88:
@amartelli I can't seem to access the slides on CERNbox. Can you reupload them or share them in a different way?

civanch:
+1

amartelli:
Hi @kpedro88 
for me CERNbox works, so in case I attach the slides here
[HGCAL_genTOA.pdf](https://github.com/cms-sw/cmssw/files/4116564/HGCAL_genTOA.pdf)

kpedro88:
+upgrade

silviodonato:
please test 23434.1001

silviodonato:
@trtomei @fwyzard be aware of this bug in 11_0_X GEN-SIM-DIGI, even if the bug has a "negligible impact also at 200PU"

silviodonato:

silviodonato:
+1

fwyzard:
On Wed, 29 Jan 2020, 09:54 Silvio Donato, <notifications@github.com> wrote:

> @trtomei <https://github.com/trtomei> @fwyzard
> <https://github.com/fwyzard> be aware of this bug in 11_0_X GEN-SIM-DIGI,
> even if the bug has a "negligible impact also at 200PU"
>

Thanks for the heads up.

I checked with HGCal DPG, and indeed this should not be an issue for the
HLT Phase 2 studies.

>

PR #28790, opened 2020-01-24
Add 2 workflows for default PR test, target for PPS plots

srimanob:
#### PR description:
PPS would like to request to add 2 workflows for PR test by default as the workflows used by default in CMSSW PR tests do not cover well the needs of PPS. Workflows include 2017C, 2018C Egamma (was ZeroBias but change after discussed with RECO conveners)

FYI
@jan-kaspar 

#### PR validation:
These workflows are part of relval_standard.py, and run regularly in relvals submitted by PdmV.

### DIFF
--- a/Configuration/PyReleaseValidation/scripts/runTheMatrix.py
+++ b/Configuration/PyReleaseValidation/scripts/runTheMatrix.py
@@ -62,7 +62,9 @@ if __name__ == '__main__':
                      136.7611, #2016E JetHT reMINIAOD from 80X legacy
                      136.8311, #2017F JetHT reMINIAOD from 94X reprocessing
                      136.788, #2017B Photon data
+                     136.801, #2017C ZeroBias
                      136.85, #2018A Egamma data
+                     136.880, #2018C ZeroBias
                      140.53, #2011 HI data
                      140.56, #2018 HI data
                      158.0, #2018 HI MC with pp-like reco
--- a/Configuration/PyReleaseValidation/scripts/runTheMatrix.py
+++ b/Configuration/PyReleaseValidation/scripts/runTheMatrix.py
@@ -61,10 +61,8 @@ if __name__ == '__main__':
                      136.731, #2016B Photon data
                      136.7611, #2016E JetHT reMINIAOD from 80X legacy
                      136.8311, #2017F JetHT reMINIAOD from 94X reprocessing
-                     136.788, #2017B Photon data
-                     136.801, #2017C ZeroBias
-                     136.85, #2018A Egamma data
-                     136.880, #2018C ZeroBias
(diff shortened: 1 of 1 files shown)


slava77:
> Workflows include 2017C, 2018C ZeroBias

isn't it possible to pick up a more physics-populated dataset
so that we can replace an existing workflow, instead of extending the "short" matrix?

srimanob:
I would let @jan-kaspar comments on this, as the proposal from PPS on WFs which are sensible PPS cover only ZeroBias.

jan-kaspar:
Let me copy here my answer to the parallel email thread.

We are almost indifferent about the trigger choice - we look at pile-up protons anyway. What is important to us is the run/LS selection. Therefore I believe that we can find a good candidate to suit the needs both of PPS and other sub-systems. For example, I looked at
  * 136.793, RunDoubleEG2017C -> about 20 reco protons in each PPS arm
  * 136.874, RunEGamma2018C -> about 40 reco protons in each PPS arm

The statistics is not huge, but probably OK. I looked in the generated cfg.py files and see maxEvents.input = 100. I presume that this is applied to all runTheMatrix jobs and therefore we cannot get (many) more protons by choosing a different dataset.

slava77:
The number of events could be adjusted, but it has to be done with an assumption that the job still finishes in reasonable time (order 10 mins for the short matrix tests). So, 100 stays as a reasonable choice so far.

Did we have enough useful data in the low-pileup (or the high-beta*) runs?
This could be an alternative to get some more events with PPS, perhaps mostly for private tests rather than for the standard short matrix set up.

srimanob:
List of workflows is updated as agreed.

jan-kaspar:
> The number of events could be adjusted, but it has to be done with an assumption that the job still finishes in reasonable time (order 10 mins for the short matrix tests). So, 100 stays as a reasonable choice so far.

Thanks, I understand. I believe that those altogether ~100 reco protons would be sufficient
  * to spot a difference, when strictly no difference is expected
  * to judge on a trend, when it is expected.

> Did we have enough useful data in the low-pileup (or the high-beta*) runs?
> This could be an alternative to get some more events with PPS, perhaps mostly for private tests rather than for the standard short matrix set up.

For sure we will keep running our private tests - processing PPS data only we can afford many more events in a short time.

chayanit:
+1

jan-kaspar:
For completeness, I put here the proton xi histograms (sector 45, +z side) from the newly introduced WFs 136.793 (top) and 136.874 (bottom)

xi_hist_136 793

xi_hist_136 874

kpedro88:
+upgrade

silviodonato:
+1

PR #28793, opened 2020-01-26
Update TTTrack for chi2 split

mdhildreth:
#### PR description:

Added capability of splitting chisq into separate xy and z calculations for different track fits. Added three new (as yet undefined) MVA-based track quality variables.  Cleaned up some of the code, including some variable assignments that were previously missing.

#### PR validation:

Track finding code runs with these new modifications.  TTTrack just stores information.

### DIFF
--- a/DataFormats/L1TrackTrigger/interface/TTTrack.h
+++ b/DataFormats/L1TrackTrigger/interface/TTTrack.h
@@ -167,6 +167,8 @@ TTTrack<T>::TTTrack() {
   theMomentum = GlobalVector(0.0, 0.0, 0.0);
   theRInv = 0.0;
   thePOCA = GlobalPoint(0.0, 0.0, 0.0);
+  theD0 = 0.;
+  theZ0 = 0.;
   theTanL = 0;
   thePhi = 0;
   theTrkMVA1 = 0;
@@ -187,6 +189,8 @@ TTTrack<T>::TTTrack(std::vector<edm::Ref<edmNew::DetSetVector<TTStub<T> >, TTStu
   theMomentum = GlobalVector(0.0, 0.0, 0.0);
   theRInv = 0.0;
   thePOCA = GlobalPoint(0.0, 0.0, 0.0);
+  theD0 = 0.;
+  theZ0 = 0.;
   theTanL = 0;
   thePhi = 0;
   theTrkMVA1 = 0;
@@ -215,10 +219,12 @@ TTTrack<T>::TTTrack(double aRinv,
                     unsigned int nPar,
                     double aBfield) {
   theStubRefs.clear();
-  double thePT = MagConstant * 1.0 / aRinv * aBfield / 100.0;  // Rinv is in cm-1
+  double thePT = fabs(MagConstant * 1.0 / aRinv * aBfield / 100.0);  // Rinv is in cm-1
   theMomentum = GlobalVector(GlobalVector::Cylindrical(thePT, aphi0, thePT * aTanlambda));
   theRInv = aRinv;
   thePOCA = GlobalPoint(ad0 * cos(aphi0), ad0 * sin(aphi0), az0);
+  theD0 = ad0;
+  theZ0 = az0;
   thePhi = aphi0;
   theTanL = aTanlambda;
   thePhiSector = 0;      // must be set externally
@@ -251,10 +257,12 @@ TTTrack<T>::TTTrack(double aRinv,
                     unsigned int nPar,
                     double aBfield) {
   theStubRefs.clear();
-  double thePT = MagConstant * 1.0 / aRinv * aBfield / 100.0;  // Rinv is in cm-1
+  double thePT = fabs(MagConstant * 1.0 / aRinv * aBfield / 100.0);  // Rinv is in cm-1
   theMomentum = GlobalVector(GlobalVector::Cylindrical(thePT, aphi0, thePT * aTanlambda));
   theRInv = aRinv;
   thePOCA = GlobalPoint(ad0 * cos(aphi0), ad0 * sin(aphi0), az0);
+  theD0 = ad0;
+  theZ0 = az0;
   thePhi = aphi0;
   theTanL = aTanlambda;
   thePhiSector = 0;      // must be set externally
--- a/DataFormats/L1TrackTrigger/interface/TTTrack.h
+++ b/DataFormats/L1TrackTrigger/interface/TTTrack.h
@@ -115,7 +115,6 @@ public:
 
   /// POCA
   GlobalPoint POCA() const;
-  GlobalPoint getPOCA(unsigned int npar = Npars4) const;
 
   /// MVA Track quality variables
   double trkMVA1() const;
--- a/DataFormats/L1TrackTrigger/interface/TTTrack.h
+++ b/DataFormats/L1TrackTrigger/interface/TTTrack.h
@@ -35,12 +35,16 @@ private:
   double theD0;
   double theZ0;
   unsigned int thePhiSector;
+  unsigned int theEtaSector;
   double theStubPtConsistency;
   double theChi2;
+  double theChi2XY;
+  double theChi2Z;
   unsigned int NumFitPars;
   unsigned int theHitPattern;
-  float an_MVA_Value;
-  float another_MVA_Value;
+  double theTrkMVA1;
+  double theTrkMVA2;
+  double theTrkMVA3;
   int theTrackSeed;
   double theBField;  // needed for unpacking
   static constexpr unsigned int Npars4 = 4;
@@ -58,6 +62,24 @@ public:
           double az0,
           double ad0,
           double aChi2,
+	  double trkMVA1,
+	  double trkMVA2,
+	  double trkMVA3,
+          unsigned int aHitpattern,
+          unsigned int nPar,
+          double Bfield);
+
+
+  TTTrack(double aRinv,
+          double aphi,
+          double aeta,
+          double az0,
+          double ad0,
+          double aChi2xyfit,
+          double aChi2zfit,
+	  double trkMVA1,
+	  double trkMVA2,
+	  double trkMVA3,
           unsigned int aHitpattern,
           unsigned int nPar,
           double Bfield);
@@ -99,10 +121,21 @@ public:
   GlobalPoint POCA() const;
   GlobalPoint getPOCA(unsigned int npar = Npars4) const;
 
+
+  /// MVA Track quality variables
+  double trkMVA1() const;
+  double trkMVA2() const;
+  double trkMVA3() const;
+
   /// Phi Sector
-  unsigned int PhiSector() const { return thePhiSector; }
-  unsigned int getSector() const { return thePhiSector; }
+  unsigned int phiSector() const { return thePhiSector; }
+  unsigned int getPhiSector() const { return thePhiSector; }
   void setPhiSector(unsigned int aSector) { thePhiSector = aSector; }
+
+  /// Eta Sector
+  unsigned int etaSector() const { return theEtaSector; }
+  unsigned int getEtaSector() const { return theEtaSector; }
+  void setEtaSector(unsigned int aSector) { theEtaSector = aSector; }
 
   /// Track seeding (for debugging)
   unsigned int TrackSeed() const { return theTrackSeed; }
@@ -142,7 +175,14 @@ TTTrack<T>::TTTrack() {
   theMomentum = GlobalVector(0.0, 0.0, 0.0);
   theRInv = 0.0;
   thePOCA = GlobalPoint(0.0, 0.0, 0.0);
+  theTanL = 0;
+  thePhi = 0;
+  theEta = 0;
+  theTrkMVA1 = 0;
+  theTrkMVA2 = 0;
+  theTrkMVA3 = 0;
   thePhiSector = 0;
+  theEtaSector = 0;
   theTrackSeed = 0;
   theChi2 = 0.0;
   theStubPtConsistency = 0.0;
@@ -156,7 +196,14 @@ TTTrack<T>::TTTrack(std::vector<edm::Ref<edmNew::DetSetVector<TTStub<T> >, TTStu
   theMomentum = GlobalVector(0.0, 0.0, 0.0);
   theRInv = 0.0;
   thePOCA = GlobalPoint(0.0, 0.0, 0.0);
+  theTanL = 0;
+  thePhi = 0;
+  theEta = 0;
+  theTrkMVA1 = 0;
+  theTrkMVA2 = 0;
+  theTrkMVA3 = 0;
   thePhiSector = 0;
+  theEtaSector = 0;
   theTrackSeed = 0;
   theChi2 = 0.0;
   theStubPtConsistency = 0.0;
@@ -166,22 +213,70 @@ TTTrack<T>::TTTrack(std::vector<edm::Ref<edmNew::DetSetVector<TTStub<T> >, TTStu
 /// Meant to be default constructor
 template <typename T>
 TTTrack<T>::TTTrack(double aRinv,
-                    double aphi,
-                    double aeta,
+                    double aphi0,
+                    double aTanlambda,
                     double az0,
                     double ad0,
                     double aChi2,
+		    double trkMVA1,
+		    double trkMVA2,
+		    double trkMVA3,
                     unsigned int aHitPattern,
                     unsigned int nPar,
                     double aBfield) {
   theStubRefs.clear();
   double thePT = MagConstant * aRinv * aBfield;
-  theMomentum = GlobalVector(GlobalVector::Cylindrical(thePT, aphi, thePT * sinh(aeta)));
+  theMomentum = GlobalVector(GlobalVector::Cylindrical(thePT, aphi0, thePT * aTanlambda));
   theRInv = aRinv;
-  thePOCA = GlobalPoint(ad0 * cos(aphi), ad0 * sin(aphi), az0);
+  thePOCA = GlobalPoint(ad0 * cos(aphi0), ad0 * sin(aphi0), az0);
+  thePhi = aphi0;
+  theTanL = aTanlambda;
+  theEta = theMomentum.eta();
   thePhiSector = 0;  // must be set externally
+  theEtaSector = 0;  // must be set externally
   theTrackSeed = 0;  // must be set externally
   theChi2 = aChi2;
+  theTrkMVA1 = trkMVA1;
+  theTrkMVA2 = trkMVA2;
+  theTrkMVA3 = trkMVA3;
+  theStubPtConsistency = 0.0;  // must be set externally
+  NumFitPars = nPar;
+  theHitPattern = aHitPattern;
+  theBField = aBfield;
+  // should probably fill the momentum vectur
+}
+
+/// Second default constructor with split chi2
+template <typename T>
+TTTrack<T>::TTTrack(double aRinv,
+                    double aphi0,
+                    double aTanlambda,
+                    double az0,
+                    double ad0,
+                    double aChi2XY,
+                    double aChi2Z,
+		    double trkMVA1,
+		    double trkMVA2,
+		    double trkMVA3,
+                    unsigned int aHitPattern,
+                    unsigned int nPar,
+                    double aBfield) {
+  theStubRefs.clear();
+  double thePT = MagConstant * aRinv * aBfield;
+  theMomentum = GlobalVector(GlobalVector::Cylindrical(thePT, aphi0, thePT * aTanlambda));
+  theRInv = aRinv;
+  thePOCA = GlobalPoint(ad0 * cos(aphi0), ad0 * sin(aphi0), az0);
+  thePhi = aphi0;
+  theTanL = aTanlambda;
+  theEta = theMomentum.eta();
+  thePhiSector = 0;  // must be set externally
+  theEtaSector = 0;  // must be set externally
+  theTrackSeed = 0;  // must be set externally
+  theChi2XY = aChi2XY;
+  theChi2Z = aChi2Z;
+  theTrkMVA1 = trkMVA1;
+  theTrkMVA2 = trkMVA2;
+  theTrkMVA3 = trkMVA3;
   theStubPtConsistency = 0.0;  // must be set externally
   NumFitPars = nPar;
   theHitPattern = aHitPattern;
@@ -314,6 +409,33 @@ double TTTrack<T>::getChi2Red(unsigned int npar) const  //backwards compatibilit
   return chi2Red();
 }
 
+/// MVA quality variables
+template <typename T>
+double TTTrack<T>::trkMVA1() const {
+  if (NumFitPars == Npars5 || NumFitPars == Npars4) {
+    return theTrkMVA1;
+  } else
+    return 0.0;
+}
+
+template <typename T>
+double TTTrack<T>::trkMVA2() const {
+  if (NumFitPars == Npars5 || NumFitPars == Npars4) {
+    return theTrkMVA2;
+  } else
+    return 0.0;
+}
+
+template <typename T>
+double TTTrack<T>::trkMVA3() const {
+  if (NumFitPars == Npars5 || NumFitPars == Npars4) {
+    return theTrkMVA3;
+  } else
+    return 0.0;
+}
+
+
+
 /// StubPtConsistency
 template <typename T>
 void TTTrack<T>::setStubPtConsistency(double aStubPtConsistency) {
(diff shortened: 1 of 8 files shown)


tomalin:
Some questions/comments:
1)	Can't we delete data member theEta, since we have theTanL? And why do we need theMomentum if we have thePhi, theTanL & theRInv? Converting between them on the fly (using calc at https://github.com/cms-sw/cmssw/blob/540bc00e40edb2d4f54c486dcbe8656fe3d858cf/DataFormats/L1TrackTrigger/interface/TTTrack.h#L229 ) presumably wouldn’t use much CPU. And they will waste disk space.
2)	All function & variables names should start with a lower case letter. e.g. trackSeed(),  numFitPars.
3)	Many functions are included for backwards compatibility. Now we have switched to CMSSW11, can’t we forget backwards compatiblity and just delete these? This would substantially shorten & simplify the code, and the redundant parameter nPar in the old functions is confusing. (Though we’d have to check if any code in CMSSW11 aside from L1Trk uses TTTrack). This affects getSector(), getPhiSector(), getEtaSector() & getPOCA(nPar), getRInv(nPar), getChi2(nPar), getChi2Red(nPar), getStubPtConsistentcy(nPar), setFitParNo(nPar). 
4)	Why do we need the old constructor that used chi2 instead of chi2(rphi) and chi2(rz)? – Is this just to support the pure tracklet algo?
5)	Why do we need the constructor TTTrack (aStubs)?  I can’t imagine when one would know the stubs on a track, but not the helix parameters.
6)	Add function stubPtConsistency(){return theStubPtConsistency;}
7)	Add function numFitPars(){return NumFitPars;}
8)	What does function trackSeed() return? If it is the seed type (i.e. layers used), then I suggest renaming it to trackSeedType().
9)	Function testTrackWordBits() doesn’t do anything?
10)	Can we delete Npars4=4 and Npars5=5? 
11)	Why do we need all the checks such as https://github.com/cms-sw/cmssw/blob/540bc00e40edb2d4f54c486dcbe8656fe3d858cf/DataFormats/L1TrackTrigger/interface/TTTrack.h#L303 that the number of fit params must be 4 or 5? Can’t we just delete them and shorten the code?
12)	The speed of light is inaccurate. Please change “MagConstant=0.3” to 0.299792458.
Cheers,
Ian

kpedro88:
@mdhildreth please apply the code-format, otherwise the PR cannot be tested

DataFormats/L1TrackTrigger/interface/TTTrack.h
```diff
@@ -99,11 +121,22 @@ class TTTrack : public TTTrack_TrackWord {
...
+  unsigned int getPhiSector() const { return thePhiSector; }
   void setPhiSector(unsigned int aSector) { thePhiSector = aSector; }
 
+  /// Eta Sector
+  unsigned int etaSector() const { return theEtaSector; }
+  unsigned int getEtaSector() const { return theEtaSector; }
+  void setEtaSector(unsigned int aSector) { theEtaSector = aSector; }
+
   /// Track seeding (for debugging)
   unsigned int TrackSeed() const { return theTrackSeed; }
```
skinnari: this should be trackSeed() for consistency

DataFormats/L1TrackTrigger/interface/TTTrack.h
```diff
@@ -35,12 +35,16 @@ class TTTrack : public TTTrack_TrackWord {
...
+  unsigned int theEtaSector;
   double theStubPtConsistency;
   double theChi2;
+  double theChi2XY;
+  double theChi2Z;
   unsigned int NumFitPars;
   unsigned int theHitPattern;
-  float an_MVA_Value;
-  float another_MVA_Value;
+  double theTrkMVA1;
```
skinnari: do we want these to be doubles or unsigned ints?

skinnari:
@mdhildreth @tomalin important bug still, this is wrong: 

https://github.com/mdhildreth/cmssw/blob/uptdateTTTrackForChi2Split/DataFormats/L1TrackTrigger/interface/TTTrack.h#L228

that is a calculation that uses radius of curvature in the formula, but the variable is inverse radius of curvature! must also clarify in the code that the assumed unit is [m-1]. at least the tracklet uses by default internally [cm-1] and then does the conversion when calculating pt, so would be good to add that as a comment in the class.

mdhildreth:
please test

mdhildreth:
please test

jfernan2:
+1

tomalin:
A few remaining comments:
1) Constructor https://github.com/cms-sw/cmssw/blob/fcdd9ea5a45680d3d642700974a01bea46a060b0/DataFormats/L1TrackTrigger/interface/TTTrack.h#L165 should initialize theChi2XY & theChi2Z to zero.
2) Can't we delete the constructor https://github.com/cms-sw/cmssw/blob/fcdd9ea5a45680d3d642700974a01bea46a060b0/DataFormats/L1TrackTrigger/interface/TTTrack.h#L185 ? When would we know stubs but not helix params?
3) Constructor https://github.com/cms-sw/cmssw/blob/fcdd9ea5a45680d3d642700974a01bea46a060b0/DataFormats/L1TrackTrigger/interface/TTTrack.h#L240 should initialize theChi2=theChi2XY+theChi2Z;
4) Constructor https://github.com/cms-sw/cmssw/blob/fcdd9ea5a45680d3d642700974a01bea46a060b0/DataFormats/L1TrackTrigger/interface/TTTrack.h#L240 should initialize theChi2XY & theChi2Z to some value to indicate not set.
5) Add function chi2XY(). Suggest renaming chi2z() to chi2Z(). And add function dof() to return (2 * theStubRefs.size() - numFitPars); -- it can be called also from chi2Red().
6) Suggest setting B-field value via setBfield() function instead of constructor. In data, best estimate of B-field may change after dataset is produced. This would require calculating Pt on the fly instead of in constructor. (Alternative if so keep B-field in constructor, but provide setBfield() to set B and update thePt and theMomentum).
7) TrackWord should take chi2XY and chi2Z (corresponding to firmware), not chi2Z and chi2.

skinnari:
@mdhildreth this class is still buggy. please do not merge yet. 

(1) this: 

https://github.com/cms-sw/cmssw/blob/fcdd9ea5a45680d3d642700974a01bea46a060b0/DataFormats/L1TrackTrigger/interface/TTTrack.h#L219

should be sinh(aTanlambda) 

(2) neither theD0 or theZ0 are set in the constructors.

mdhildreth:
@kpedro88 @rekovic This PR is ready to go now.  If you have additional comments, please go ahead and make them.  Thanks.

skinnari:
> A few remaining comments:
> 
>     1. Constructor https://github.com/cms-sw/cmssw/blob/fcdd9ea5a45680d3d642700974a01bea46a060b0/DataFormats/L1TrackTrigger/interface/TTTrack.h#L165
>         should initialize theChi2XY & theChi2Z to zero.
> 
>     2. Can't we delete the constructor https://github.com/cms-sw/cmssw/blob/fcdd9ea5a45680d3d642700974a01bea46a060b0/DataFormats/L1TrackTrigger/interface/TTTrack.h#L185
>         ? When would we know stubs but not helix params?
> 
>     3. Constructor https://github.com/cms-sw/cmssw/blob/fcdd9ea5a45680d3d642700974a01bea46a060b0/DataFormats/L1TrackTrigger/interface/TTTrack.h#L240
>         should initialize theChi2=theChi2XY+theChi2Z;
> 
>     4. Constructor https://github.com/cms-sw/cmssw/blob/fcdd9ea5a45680d3d642700974a01bea46a060b0/DataFormats/L1TrackTrigger/interface/TTTrack.h#L240
>         should initialize theChi2XY & theChi2Z to some value to indicate not set.
> 
>     5. Add function chi2XY(). Suggest renaming chi2z() to chi2Z(). And add function dof() to return (2 * theStubRefs.size() - numFitPars); -- it can be called also from chi2Red().
> 
>     6. Suggest setting B-field value via setBfield() function instead of constructor. In data, best estimate of B-field may change after dataset is produced. This would require calculating Pt on the fly instead of in constructor. (Alternative if so keep B-field in constructor, but provide setBfield() to set B and update thePt and theMomentum).
> 
>     7. TrackWord should take chi2XY and chi2Z (corresponding to firmware), not chi2Z and chi2.

@mdhildreth were these addressed? in particular there are some initializations that were missing?

skinnari:
just to note also here, it seems get_ichi2() results in seg fault, and that the TTTrack word does not produce the expected results. @mdhildreth

DataFormats/L1TrackTrigger/interface/TTTrack.h
```diff
@@ -31,21 +31,24 @@ class TTTrack : public TTTrack_TrackWord {
...
   double theD0;
   double theZ0;
   unsigned int thePhiSector;
+  unsigned int theEtaSector;
   double theStubPtConsistency;
   double theChi2;
-  unsigned int NumFitPars;
+  double theChi2XY;
+  double theChi2Z;
+  unsigned int numFitPars;
```
kpedro88: to fit the style used in this class, should be named `theNumFitPars`
(to fit CMS style, all variables should be changed to have a trailing underscore...)

DataFormats/L1TrackTrigger/interface/TTTrack.h
```diff
@@ -31,21 +31,24 @@ class TTTrack : public TTTrack_TrackWord {
...
-  int theTrackSeed;
+  double theTrkMVA1;
+  double theTrkMVA2;
+  double theTrkMVA3;
+  int theTrackSeedType;
   double theBField;  // needed for unpacking
   static constexpr unsigned int Npars4 = 4;
   static constexpr unsigned int Npars5 = 5;
-  static constexpr float MagConstant = 0.3;
+  static constexpr float MagConstant = 0.299792458;
```
kpedro88: better to define this in terms of `CLHEP::c_light`?

DataFormats/L1TrackTrigger/interface/TTTrack.h
```diff
@@ -156,34 +189,92 @@ TTTrack<T>::TTTrack(std::vector<edm::Ref<edmNew::DetSetVector<TTStub<T> >, TTStu
...
+                    double trkMVA1,
+                    double trkMVA2,
+                    double trkMVA3,
                     unsigned int aHitPattern,
                     unsigned int nPar,
                     double aBfield) {
   theStubRefs.clear();
-  double thePT = MagConstant * aRinv * aBfield;
-  theMomentum = GlobalVector(GlobalVector::Cylindrical(thePT, aphi, thePT * sinh(aeta)));
+  double thePT = fabs(MagConstant * 1.0 / aRinv * aBfield / 100.0);  // Rinv is in cm-1
```
kpedro88: use `std::abs()` rather than `fabs()`

DataFormats/L1TrackTrigger/interface/TTTrack.h
```diff
@@ -156,34 +189,92 @@ TTTrack<T>::TTTrack(std::vector<edm::Ref<edmNew::DetSetVector<TTStub<T> >, TTStu
...
   theChi2 = aChi2;
+  theTrkMVA1 = trkMVA1;
+  theTrkMVA2 = trkMVA2;
+  theTrkMVA3 = trkMVA3;
   theStubPtConsistency = 0.0;  // must be set externally
-  NumFitPars = nPar;
+  numFitPars = nPar;
+  theHitPattern = aHitPattern;
+  theBField = aBfield;
+  // should probably fill the momentum vectur
```
kpedro88: is this needed for this PR?

DataFormats/L1TrackTrigger/interface/TTTrack.h
```diff
@@ -156,34 +189,92 @@ TTTrack<T>::TTTrack(std::vector<edm::Ref<edmNew::DetSetVector<TTStub<T> >, TTStu
...
+                    double aChi2XY,
+                    double aChi2Z,
+                    double trkMVA1,
+                    double trkMVA2,
+                    double trkMVA3,
+                    unsigned int aHitPattern,
+                    unsigned int nPar,
+                    double aBfield) {
+  theStubRefs.clear();
+  double thePT = fabs(MagConstant * 1.0 / aRinv * aBfield / 100.0);  // Rinv is in cm-1
```
kpedro88: use `std::abs()` rather than `fabs()`

DataFormats/L1TrackTrigger/interface/TTTrack.h
```diff
@@ -156,34 +189,92 @@ TTTrack<T>::TTTrack(std::vector<edm::Ref<edmNew::DetSetVector<TTStub<T> >, TTStu
...
+  theTrkMVA3 = trkMVA3;
   theStubPtConsistency = 0.0;  // must be set externally
-  NumFitPars = nPar;
+  numFitPars = nPar;
+  theHitPattern = aHitPattern;
+  theBField = aBfield;
+  // should probably fill the momentum vectur
+}
+
+/// Second default constructor with split chi2
```
kpedro88: rather than duplicating all this code, call the previous constructor and then just change the chi2 values?

DataFormats/L1TrackTrigger/interface/TTTrack.h
```diff
@@ -205,137 +296,96 @@ void TTTrack<T>::setFitParNo(unsigned int nPar) {
...
 }
 
 /// Chi2 reduced
 template <typename T>
 double TTTrack<T>::chi2Red() const {
-  if (NumFitPars == Npars5 || NumFitPars == Npars4) {
-    return theChi2 / (2 * theStubRefs.size() - NumFitPars);
-  } else
-    return 0.0;
+  return theChi2 / (2 * theStubRefs.size() - numFitPars);
```
kpedro88: is it worth caching this value (maybe as a transient to maintain the size of the class on disk)?

DataFormats/L1TrackTrigger/src/TTTrack_TrackWord.cc
```diff
@@ -297,8 +314,19 @@ unsigned int TTTrack_TrackWord::get_BendChi2Bits() {
   return bits;
 }
 
+float TTTrack_TrackWord::unpack_ichi2Z() {
+  unsigned int bits = (TrackWord3 & 0x00003C00) >> 10;
```
kpedro88: all of the masks and offsets in this file should be named constants

skinnari:
@mdhildreth In addition to what has been pointed out above (e.g. related to chi2 from @tomalin) and general comments from @kpedro88), also a getter function for hitpattern is missing!

DataFormats/L1TrackTrigger/interface/TTTrack.h
mdhildreth: done.

DataFormats/L1TrackTrigger/interface/TTTrack.h
mdhildreth: done

DataFormats/L1TrackTrigger/interface/TTTrack.h
mdhildreth: done

DataFormats/L1TrackTrigger/interface/TTTrack.h
mdhildreth: done

DataFormats/L1TrackTrigger/interface/TTTrack.h
mdhildreth: was just a random comment. It's done.

DataFormats/L1TrackTrigger/interface/TTTrack.h
mdhildreth: done

DataFormats/L1TrackTrigger/interface/TTTrack.h
mdhildreth: @kpedro88 Wouldn't caching make the class bigger?  I don't understand the sign of this comment?

mdhildreth:
back out pushing entire L1 Trigger suite.  Oops.  More later.

mdhildreth:
@tomalin @skinnari  This last commit addresses all of the comments and suggestions except it doesn't
1) fix the get_ichi2() bug/crash
2) rename all of the bit patterns in the TrackWord as variables.  Too much work for now.

DataFormats/L1TrackTrigger/interface/TTTrack.h
```diff
@@ -63,7 +63,7 @@ class TTTrack : public TTTrack_TrackWord {
           double aChi2,
           double trkMVA1,
           double trkMVA2,
-          double trkMVA3,
+          double trkMVA3_,
```
skinnari: @mdhildreth why did trkMVA3 change to trkMVA3_ in the two constructors? This seems odd to me?

DataFormats/L1TrackTrigger/interface/TTTrack.h
```diff
@@ -26,34 +27,33 @@ class TTTrack : public TTTrack_TrackWord {
...
+  double theRInv_;
+  double thePhi_;
+  double theTanL_;
+  double theD0_;
+  double theZ0_;
+  unsigned int thePhiSector_;
+  unsigned int theEtaSector_;
+  double theStubPtConsistency_;
+  double theChi2_;
+  double theChi2_XY_;
```
skinnari: Were these meant to be theChi2XY_ and theChi2Z_ ?

DataFormats/L1TrackTrigger/interface/TTTrack.h
```diff
@@ -296,96 +280,119 @@ void TTTrack<T>::setFitParNo(unsigned int nPar) {
...
 }
 
 /// StubPtConsistency
 template <typename T>
 double TTTrack<T>::stubPtConsistency() const {
-  return theStubPtConsistency;
+  return theStubPtConsistency_;
 }
 
+/// StubPtConsistency
```
skinnari: Would be good to correct comment to avoid confusion...

DataFormats/L1TrackTrigger/interface/TTTrack.h
```diff
@@ -296,96 +280,119 @@ void TTTrack<T>::setFitParNo(unsigned int nPar) {
...
+template <typename T>
+unsigned int TTTrack<T>::hitPattern() const {
+  return theHitPattern_;
+}
+
+/// StubPtConsistency
```
skinnari: Same here, would be good to correct comment to avoid confusion.

DataFormats/L1TrackTrigger/interface/TTTrack.h
mdhildreth: oh, that's a typo.  Shouldn't be...  It doesn't really matter, but...

DataFormats/L1TrackTrigger/interface/TTTrack.h
kpedro88: Member variables can be marked as transient so they do not persist, but also do not need to be recalculated every time the accessor is called.

DataFormats/L1TrackTrigger/interface/TTTrack.h
tomalin: Agreed. I have a few further comments:

1) The digitisation ranges assumed at https://github.com/mdhildreth/cmssw/blob/uptdateTTTrackForChi2Split/DataFormats/L1TrackTrigger/interface/TTTrack_TrackWord.h#L219 appear wildly inconsistent with those specified in https://twiki.cern.ch/twiki/bin/view/CMS/HybridDataFormat#Fitted_Tracks_written_by_KalmanF . Please check.

2) Please add two extra functions to TTTrack.h:

chi2XYred(){return theChi2XY_/(theStubRefs.size() - (theNumFitPars_-2));}
chi2Zred(){return theChi2Z_/(theStubRefs.size() - 2);}

3) If you have adopted the convention that class data members are distinguished by naming them theXYZ_, then variables such as https://github.com/mdhildreth/cmssw/blob/uptdateTTTrackForChi2Split/DataFormats/L1TrackTrigger/interface/TTTrack_TrackWord.h#L26 should not include "the"

4) Please eliminate the hard-wired constants, such as 16 at https://github.com/mdhildreth/cmssw/blob/uptdateTTTrackForChi2Split/DataFormats/L1TrackTrigger/src/TTTrack_TrackWord.cc#L167 . This could be written instead (NZ0Bits+1)+NChi2Bits .

Cheers,
Ian

DataFormats/L1TrackTrigger/interface/TTTrack.h
mdhildreth: done

DataFormats/L1TrackTrigger/interface/TTTrack.h
mdhildreth: **Responding to comments:**

The last commit gets most of these, including Louise's suggestions.

Responding to Ian's comments: @tomalin 

1. Fixed, except for phi0. I assumed that phi would be digitized with respect to the center of a nonant, which gets you maximum resolution (if we could figure out how to calculate it).  I don't know/understand why phi0 should range between +/- 1.026 ?  Why is that the magic number?

2. Done
3. Done
4. Sorry, I don't have time to do this now.  Since the Track Word format is likely to change, I also don't want to spend a whole bunch of time prettying up this code at this point.  I suggest we press on with this version unless it's critical.  
5. Done

Cheers,
Mike

tomalin: 1) The digitisation range in phi0 of +-1.026 rads about the centre of the sector. This is because the required stub phi digi range is 2*pi/9 + 2*hourglass_overlap (2*0.1638) = 1.026; i.e. from -1.026/2 to +1.026/2. To simplify maths, the phi0 digi range can differ from this only by powers of 2. And unfortunately, this range is very slightly too small to cover the allowed phi0 values of tracks in a nonant, so we have to double it to (-1.026,+1.026).
2) I just noticed that TTTrackWord is digitising eta instead of tanL. This is also inconsistent with the agreement documented on the twiki.

mdhildreth: @tomalin @skinnari
Ok, I have fixed the last two issues with the phi digitization and switched eta -> tanl in the Track Word.   I guess someone downstream will do the conversion from tanl back to eta so it can be used in the trigger.

silviodonato:
please test

jfernan2:
+1
Changes unrelated to this PR

kpedro88:
https://github.com/cms-sw/cmssw/pull/28793/files#r378293112 has not been addressed, leaving several hardcoded values in TTTrack_TrackWord.

mdhildreth:
@kpedro88 Hi Kevin - I updated the fixed masks and shifts in TTTrack_TrackWord.cc  Everything else should be fixed now.

mdhildreth:
please test

mdhildreth:
please test

jfernan2:
+1

kpedro88:
+upgrade

silviodonato:
@benkrikler @rekovic do you have any comments?

DataFormats/L1TrackTrigger/interface/TTTrack.h
tomalin: My comment (2) above was incorrectly implemented in the C++. The number of degrees of freedom of the chi2XY should be "(theStubRefs.size() - (theNumFitPars_-2))", but at https://github.com/cms-sw/cmssw/blob/9deefb0d87de9ad4a85b9c12cf28723725636493/DataFormats/L1TrackTrigger/interface/TTTrack.h#L347 it is instead given as "(theStubRefs.size() - theNumFitPars_ - 2)". 

N.B. As sanity check, the number of degrees of freedom in XY plus the number in Z should equal the number in total, which is not currently the case in the C++.

tomalin:
Please check my report of a bug in TTTrack.h in https://github.com/cms-sw/cmssw/pull/28793#discussion_r392960098 . (I mention it again here, as github "hides" old 
conversations, and I suspect it could be missed).

DataFormats/L1TrackTrigger/interface/TTTrack.h
```diff
@@ -205,137 +279,130 @@ void TTTrack<T>::setFitParNo(unsigned int nPar) {
...
-    return theChi2 / (2 * theStubRefs.size() - NumFitPars);
-  } else
-    return 0.0;
+  return theChi2_ / (2 * theStubRefs.size() - theNumFitPars_);
+}
+
+/// Chi2XY reduced
+template <typename T>
+double TTTrack<T>::chi2XYRed() const {
+  return theChi2_XY_ / (theStubRefs.size() - theNumFitPars_ - 2);
```
silviodonato:
```suggestion
  return theChi2_XY_ / (theStubRefs.size() - (theNumFitPars_ - 2));
```
@mdhildreth, @tomalin suggest to make this fix

DataFormats/L1TrackTrigger/interface/TTTrack.h
mdhildreth: fixed

rekovic:
+1

jfernan2:
+1

kpedro88:
+upgrade

silviodonato:
+1

PR #28808, opened 2020-01-28
Introduce fnmatch in a DQM script

fioriNTU:
#### PR description:

This PR add the possibility to use wildcards in the script which extract DQMIO data in the form of SQMLITE tables

#### PR validation:

This PR has been validated in a private setup successfully. The standard PR test will not run this code, which is a standalone script, so I expect no changes at all in whatever comparison due to this PR

### DIFF
--- a/DQMServices/Components/scripts/dqmiodatasetharvest.py
+++ b/DQMServices/Components/scripts/dqmiodatasetharvest.py
@@ -7,7 +7,7 @@ import sqlite3
 import argparse
 import subprocess
 import multiprocessing
-
+import fnmatch
 
 ROOTPREFIX = "root://cms-xrd-global.cern.ch/"
 #ROOTPREFIX = "root://eoscms//eos/cms" # for more local files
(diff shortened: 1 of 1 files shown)


fioriNTU:
please test

fioriNTU:
+1

DQMServices/Components/scripts/dqmiodatasetharvest.py
```diff
@@ -23,83 +23,27 @@
...
-
-  "L1T/L1TObjects/L1TEGamma/timing/egamma_eta_phi_bx_0",
-  "L1T/L1TObjects/L1TJet/timing/jet_eta_phi_bx_0",
-  "L1T/L1TObjects/L1TMuon/timing/muons_eta_phi_bx_0",
-  "L1T/L1TObjects/L1TTau/timing/tau_eta_phi_bx_0",
-  "L1T/L1TObjects/L1TEGamma/timing/denominator_egamma",
-  "L1T/L1TObjects/L1TJet/timing/denominator_jet",
-  "L1T/L1TObjects/L1TMuon/timing/denominator_muons",
-  "L1T/L1TObjects/L1TTau/timing/denominator_tau",
-}
```
silviodonato: These names do not match  "PixelPhase1/Phase1_MechanicalView/PXBarrel/adc_PXLayer*".
Are you removing them on purpurse? If yes, could update the PR description specifying what you are changing?

DQMServices/Components/scripts/dqmiodatasetharvest.py
```diff
@@ -208,13 +151,27 @@ def harvestfile(fname):
...
+            if mename.find("Isolated") != -1:
+              continue
+            
+            if mename.find("HLT") != -1:
+              continue
+            
+            if not ((mename.find("SiStrip") >= 0) or (mename.find("OfflinePV") >= 0) or (mename.find("PixelPhase1") >= 0) or (mename.find("Tracking") >= 0 )):    
+              continue
+
+            if check_interesting(mename):
```
silviodonato: Do you have any special reasons to apply a veto on "AlCaReco", "Isolated", "HLT" and requiring at least one match among "SiStrip", "OfflinePV", "PixelPhase1", "Tracking"? I expected you specify them directly with "interesting_mes"

fioriNTU: well, no, I couldn't do it in interesting_mes once added the fnmatch. The point is that AlCaReco, IsolatedBunches and HLT workspaces all contains a SiStrip folder, which is a copy with different selections of the original SIStrip folder in the GUI. Not adding those lines I would collect also the histograms with the same name in the other folders. This could be removed for the general case, I left these on purpose to remind people, there can be cases such this one.

silviodonato:
Thanks @fioriNTU. Could you also confirm that you are removing many "interesting_mes" on purpose (see the other comment above)?

fioriNTU:
@silviodonato , yes I am removing a lot of interesting_mes on purpose, the reason is that the choice of the particular histogram to be extracted will mostly depend by the specific use case.

silviodonato:
+1

PR #28279, opened 2019-10-25
Add in beamspot constraints for timing hits and an improved track/MTD-hit selection BDT.

lgray:
This PR introduces an ID MVA for tracks associated to MTD hits so that poor-quality associations can be flagged. It also introduces additional criteria for selecting hits in the MTD track extended to require compatibility with the beamspot.

Validation can largely be found in the MTD TDR, can reproduce appropriate plots for this PR.

This is one remaining TODO item, create an external package to hold the BDT weights for the new track-quality MVA.

@bendavid @pmeridian

PR #28815, opened 2020-01-28
MTD TDR Reconstruction + MTD Track Association BDT 

pmeridian:
#### PR description:
Reopening of PR #28279: adds the tuning of the MTD track extender done for MTD TDR (using also a beam spot/vertex time constrain when finding new hits) and add a track quality MTD associator score based on a BDT classifier. 

Most of the comments raised for PR #28279 are addressed (not addressing all the style comments made for TrackPIDinfo, as this will need to completely evolve once MTD reconstruction is fully integrated into the tracker)

The TrackPUIDMVA Producer and TOFPIDProducer are now stored into a new package in RecoMTD/TimingIDTools as suggested by @slava77 .

Also the BDT evaluation now used TMVAEvaluator with the GBRForest evaluation. 

Future backport into 11_0_X can be imagined once this is fully integrated.

#### PR validation:
Validated this technically rerecoing 1000 events from RelValZMM_14TeV. To run a real test is required to make a repository to store the BDT xml file (now residing under RecoMTD/TimingIDTools/data). Please create a cms-data repository for this package.

@parbol @hatakeyamak @bendavid @simonepigazzini


pmeridian:
Don't know how to fix the "Found files with invalid states" error (showing files I've removed with git rm)

slava77:
> Don't know how to fix the "Found files with invalid states" error (showing files I've removed with git rm)

These warnings are tolerable.
However, if you could review the commit history of this PR and drop the commits that introduced the later removed files, it would be nice.

slava77:
@cmsbuild please test

pmeridian:
For tests to succeed need to create a cms-data repository for RecoMTD/TimingIDTools (once the package name is signed off). Meantime, I've fixed a typo in a module name. 
runTheMatrix test  work locally:
23234.0_TTbar_14TeV+TTbar_14TeV_TuneCUETP8M1_2026D49_GenSimHLBeamSpotFull14+DigiFullTrigger_2026D49+RecoFullGlobal_2026D49+HARVESTFullGlobal_2026D49 Step0-PASSED Step1-PASSED Step2-PASSED Step3-PASSED  - time date Wed Jan 29 10:25:07 2020-date Wed Jan 29 10:14:32 2020; exit: 0 0 0 0
1 1 1 1 tests passed, 0 0 0 0 failed

fabiocos:
@pmeridian the check on added and deleted files is relevant for large data files that might erroneously enter into the history, and pollute the repository even if deleted from the HEAD of the branch. In this case I guess the files are small. If the history of the commits that introduces them is not particularly significant a squash could simply remove the warning (and the addition of useless stuff to the master history).

slava77:
@cmsbuild please test

slava77:
abort test

slava77:
> To run a real test is required to make a repository to store the BDT xml file (now residing under RecoMTD/TimingIDTools/data). Please create a cms-data repository for this package.

@smuzaffar @mrodozov 
please create a repository

slava77:
@cmsbuild please test

I realized that the test failure was not due to a missing training input (yet)

smuzaffar:
@slava77 , you mean new data repo for RecoMTD/TimingIDTools?

slava77:
> @slava77 , you mean new data repo for RecoMTD/TimingIDTools?

yes, 
cms-data/RecoMTD-TimingIDTools

smuzaffar:
new data repository is now available https://github.com/cms-data/RecoMTD-TimingIDTools
Should we add `RecoMTD/TimingIDTools` under `reconstruction` category?

slava77:
> new data repository is now available https://github.com/cms-data/RecoMTD-TimingIDTools

Thank you.
@pmeridian you can now make a PR to this repository to add the training files

> Should we add `RecoMTD/TimingIDTools` under `reconstruction` category?

yes. I'd also say to add to the "upgrade", but there is a proposal under review to drop that category.

slava77:
the error is somewhat expected 
```
RuntimeError: An exception of category 'FileInPathError' occurred.
Exception Message:
edm::FileInPath unable to find file RecoMTD/TimingIDTools/data/clf4D_MTDquality_bo.xml anywhere in the search path.
```

We can restart the tests after the training file is available in https://github.com/cms-data/RecoMTD-TimingIDTools (at least as a PR)

pmeridian:
PR is now available: https://github.com/cms-data/RecoMTD-TimingIDTools/pull/1

slava77:
@cmsbuild please test with cms-data/RecoMTD-TimingIDTools#1

RecoMTD/TimingIDTools/plugins/TrackPUIDMVAProducer.cc
```diff
@@ -0,0 +1,148 @@
...
+  for (unsigned int itrack = 0; itrack < tracks.size(); ++itrack) {
+    const reco::Track& track = tracks[itrack];
+    const reco::TrackRef trackref(tracksH, itrack);
+
+    if (trackAssoc[trackref] == -1) {
+      mvaOutRaw.push_back(-1.);
+      continue;
+    } else {
+      //---training performed only above 0.5 GeV
+      if (track.pt() < 0.5)
```
kpedro88: magic number 0.5 should be named constant

RecoMTD/TrackExtender/plugins/TrackExtenderWithMTD.cc
```diff
@@ -267,21 +374,78 @@ void TrackExtenderWithMTDT<TrackCollection>::produce(edm::Event& ev, const edm::
...
+    edm::Handle<vector<SimVertex>> genVtxH;
+    ev.getByToken(genVtxToken_, genVtxH);
+    if (genVtxH.isValid())
+      genPV = &(genVtxH.product()->at(0));
+  }
+
+  double vtxTime = 0.;
+  if (useVertex_) {
+    if (useSimVertex_ && genPV) {
+      vtxTime = genPV->position().t() * CLHEP::second;  //convert to ns
```
kpedro88: @cvuosalo can you recommend a better unit conversion here?

RecoMTD/TrackExtender/plugins/TrackExtenderWithMTD.cc
```diff
@@ -358,38 +545,260 @@ void TrackExtenderWithMTDT<TrackCollection>::produce(edm::Event& ev, const edm::
...
+      edm::LogError("CannotPropagateToBeamLine") << "the state on the closest measurement isnot valid. skipping track.";
+      return false;
+    }
+
+    const FreeTrajectoryState& stateForProjectionToBeamLine = *stateForProjectionToBeamLineOnSurface.freeState();
+
+    TSCBLBuilderWithPropagator tscblBuilder(*thePropagator);
+    tscbl = tscblBuilder(stateForProjectionToBeamLine, bs);
+
+    if (!tscbl.isValid()) {
```
kpedro88: simplify to: `return tscbl.isValid();`

kpedro88:
@pmeridian please also let us know if the comparison differences are expected and match the private validation

pmeridian:
> @pmeridian please also let us know if the comparison differences are expected and match the private validation

Yes, the small differences in the trackExtender properties match a somewhat stricter requirements which now includes also a time constraint (from the beamspot). It will be nice also to have some PU200 comparison if possible.

slava77:
> It will be nice also to have some PU200 comparison if possible.

@pmeridian 
this comment is a bit scary.
It sounds like the developments were not tested in the standard expected environment.
Please point to some slides where the physics side of updates in this PR was described.
It would be nice to have it in the PR description.
Thank you.

slava77:
@cmsbuild please test with cms-data/RecoMTD-TimingIDTools#1

pmeridian:
> > It will be nice also to have some PU200 comparison if possible.
> 
> @pmeridian
> this comment is a bit scary.
> It sounds like the developments were not tested in the standard expected environment.
> Please point to some slides where the physics side of updates in this PR was described.
> It would be nice to have it in the PR description.
> Thank you.

Well you asked me if the differences are as expected, I said yes but  of course it will be nice to see also some events at PU200. Comparison on the MTD quality BDT I've done on ntuples used for TDR versus this PR on(rerecoing some TTbar PU200 RelVal events) seems to indicate that everything is as expected. 

Screenshot 2020-01-30 at 17 01 12

slava77:
> Comparison on the MTD quality BDT I've done on ntuples used for TDR versus this PR on(rerecoing some TTbar PU200 RelVal events) seems to indicate that everything is as expected.

Nice.
Thanks for reassurance.

Are these variables in the standard DQM already, or is this something still based on private tools?

pmeridian:
Private tools still. @casarsa is working on a RECO validation DQM package

RecoMTD/TrackExtender/plugins/TrackExtenderWithMTD.cc
cvuosalo: Yes, please remove

`#include "CLHEP/Units/GlobalSystemOfUnits.h"`

and add

`#include DataFormats/Math/interface/GeantUnits.h`

I will provide the suggested change for line 396.

cvuosalo:
```suggestion
      vtxTime = genPV->position().t() * geant_units::seconds;  //convert to ns
```

RecoMTD/TrackExtender/plugins/TrackExtenderWithMTD.cc
```diff
@@ -358,38 +545,256 @@ void TrackExtenderWithMTDT<TrackCollection>::produce(edm::Event& ev, const edm::
...
+                    double t_vtx,
+                    double t_vtx_err,
+                    bool addPIDError = true) {
+      constexpr double m_pi = 0.13957018;
+      constexpr double m_pi_inv2 = 1.0 / m_pi / m_pi;
+      constexpr double m_k = 0.493677;
+      constexpr double m_k_inv2 = 1.0 / m_k / m_k;
+      constexpr double m_p = 0.9382720813;
+      constexpr double m_p_inv2 = 1.0 / m_p / m_p;
+      constexpr double c_cm_ns = CLHEP::c_light * CLHEP::ns / CLHEP::cm;  //[cm/ns]
```
cvuosalo:
```suggestion
      constexpr double c_cm_ns = geant_units::operators::convertMmToCm(CLHEP::c_light);  // [mm/ns] -> [cm/ns]
```

RecoMTD/TrackExtender/plugins/TrackExtenderWithMTD.cc
```diff
@@ -358,38 +545,256 @@ void TrackExtenderWithMTDT<TrackCollection>::produce(edm::Event& ev, const edm::
...
+        validpropagation = false;
+      }
+      pathlength = pathlength1 + pathlength2;
+    }
+
+    return validpropagation;
+  }
+
+  struct TrackTofPidInfo {
+    TrackTofPidInfo(double magp2,
```
slava77: this struct exposes all data members which are supposed to have some relationship.
This is a case for making a class.
Perhaps a more effective way out here is to keep a plain ```struct TrackTofPidInfo ``` without the constructor and replace the constructor with a standalone function that returns a ```const TrackTofPidInfo```.

RecoMTD/TrackExtender/plugins/TrackExtenderWithMTD.cc
```diff
@@ -358,38 +545,256 @@ void TrackExtenderWithMTDT<TrackCollection>::produce(edm::Event& ev, const edm::
...
+              double t_vtx_err = useVtxConstraint ? vtx_res : beamSpot_res;  //should use beam spot in the future
+
+              constexpr double t_res_manual = 0.035;
+              TrackTofPidInfo tof(p.mag2(),
+                                  tot_pl,
+                                  itr->time(),
+                                  t_res_manual,
+                                  t_vtx,
+                                  t_vtx_err,
+                                  false);  //put hit error by hand for the moment
```
slava77: it looks like the end-line comment made clang-format break up this call into many lines.

does the comment refer to the ```t_res_manual``` or to the full call?
For the former case, better make it on the line of the argument.
For the latter, perhaps move above the line ```TrackTofPidInfo tof(```

RecoMTD/TrackExtender/plugins/TrackExtenderWithMTD.cc
```diff
@@ -398,47 +803,107 @@ namespace {
...
+                                                              const reco::BeamSpot& bs,
+                                                              const double& vtxTime,
+                                                              const bool matchVertex,
+                                                              TransientTrackingRecHit::ConstRecHitContainer& output,
+                                                              MTDHitMatchingInfo& bestHit) const {
+  std::set<MTDHitMatchingInfo> hitsInLayer;
+  bool hitMatched = false;
+
+  if (useVertex_ && matchVertex)
+    find_hits_in_dets(hits, traj, ilay, tsos, vtxTime, bs, prop, *theEstimator, *hitbuilder, true, hitsInLayer);
```
slava77: most parameters are the same in the 3 calls here.  std::bind may make this more readable
```c++
using namespace std::placeholders;
short_hits_in_dets = std::bind(find_hits_in_dets, hits, traj, ilay, tsos, _1, bs, prop, *theEstimator, *hitbuilder, _2, hitsInLayer);
...
if (useVertex_ && matchVertex)
   short_hits_in_dets(vtxTime, true);
else 
   short_hits_in_dets(0, false);
...
short_hits_in_dets(0, false);
```

RecoMTD/TrackExtender/plugins/TrackExtenderWithMTD.cc
```diff
@@ -494,83 +940,41 @@ reco::Track TrackExtenderWithMTDT<TrackCollection>::buildTrack(const reco::Track
...
-          pathlength = pathlength1 + pathlength2;
-          const MTDTrackingRecHit* mtdhit = static_cast<const MTDTrackingRecHit*>(it->recHit()->hit());
-          thit = mtdhit->time();
-          thiterror = mtdhit->timeError();
-          validmtd = true;
-          break;
-        }
+
+    //need to better handle the cases with >1 hit in MTD
+    for (auto it = trajWithMtd.measurements().begin(); it != trajWithMtd.measurements().end(); ++it) {
```
slava77: ```it``` is not used in the loop -> use a range loop ```for (auto const& hit : trajWithMtd.measurements())```

RecoMTD/TrackExtender/plugins/TrackExtenderWithMTD.cc
```diff
@@ -494,83 +940,41 @@ reco::Track TrackExtenderWithMTDT<TrackCollection>::buildTrack(const reco::Track
...
-      double betaerror = beta_p - beta_pi;
-
+      TrackTofPidInfo tofInfo(
+          p.mag2(),
+          pathlength,
+          thit,
+          thiterror,
+          0.,
+          0.,
+          true);                   //here add the PID uncertainty for later use in the 1st step of 4D vtx reconstruction
```
slava77: this looks like a case of clang-format breaking up the code lines for no reason.
Please move the comment above ```TrackTofPidInfo tofInfo```

RecoVertex/Configuration/python/RecoVertex_cff.py
```diff
@@ -82,3 +84,4 @@
 phase2_timing_layer.toReplaceWith(offlinePrimaryVertices4DWithBS, offlinePrimaryVertices4DwithPIDWithBS.clone())
 phase2_timing_layer.toModify(offlinePrimaryVertices4D, vertices = "unsortedOfflinePrimaryVertices4D", particles = "trackRefsForJetsBeforeSorting4D")
 phase2_timing_layer.toModify(offlinePrimaryVertices4DWithBS, vertices = "unsortedOfflinePrimaryVertices4D:WithBS", particles = "trackRefsForJetsBeforeSorting4D")
+phase2_timing_layer.toModify(tofPID, vtxsSrc='unsortedOfflinePrimaryVertices4D')
```
slava77: normally only modules defined in a given file can have toModify applied.
Please put the ```phase2_timing_layer.toModify(tofPID``` in RecoVertex_phase2_timing_cff.py

RecoVertex/Configuration/python/RecoVertex_phase2_timing_cff.py
```diff
@@ -31,20 +31,26 @@
...
 offlinePrimaryVertices4DwithPIDWithBS=offlinePrimaryVertices4DwithPID.clone(vertices="unsortedOfflinePrimaryVertices4DwithPID:WithBS")
 
 from SimTracker.TrackerHitAssociation.tpClusterProducer_cfi import tpClusterProducer
 from SimTracker.TrackAssociatorProducers.quickTrackAssociatorByHits_cfi import quickTrackAssociatorByHits
 from SimTracker.TrackAssociation.trackTimeValueMapProducer_cfi import trackTimeValueMapProducer
-from CommonTools.RecoAlgos.tofPID_cfi import tofPID
+from RecoMTD.TimingIDTools.tofPIDProducer_cfi import tofPIDProducer
+
+tofPID4DnoPID=tofPIDProducer.clone()
+tofPID4DnoPID.vtxsSrc=cms.InputTag('unsortedOfflinePrimaryVertices4DnoPID')
```
slava77: please drop type specifications for all parameters which already exist.
```tofPID4DnoPID=tofPIDProducer.clone(vtxsSrc = "unsortedOfflinePrimaryVertices4DnoPID")```

File: RecoMTD/TimingIDTools/plugins/BuildFile.xml [hunk 12]
```diff
@@ -0,0 +1,13 @@
+<export>
+</export>
+<library   name="RecoMTDTimingIDTools_plugins" file="*.cc">
+   <use   name="FWCore/Framework"/>
```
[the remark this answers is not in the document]
smuzaffar: @slava77 , in `plugins` directory every `<library> </library>` is by default an edm plugin, so need to explicitly add this `<flags EDM_PLUGIN="1"/>` flag. 

Yes the `<export> </export>` section has no use here. `<export>` is only needed in package level BuildFiles

RecoMTD/TimingIDTools/plugins/TOFPIDProducer.cc
```diff
@@ -174,7 +177,7 @@ void TOFPIDProducer::produce(edm::Event& ev, const edm::EventSetup& es) {
     float t0 = t0In[trackref];
     float t0safe = t0;
     float sigmat0safe = sigmat0In[trackref];
-    float sigmatmtd = sigmatmtdIn[trackref];
+    float sigmatmtd = (sigmatmtdIn[trackref] > 0. && fixedT0Error_ > 0.) ? fixedT0Error_ : sigmatmtdIn[trackref];
```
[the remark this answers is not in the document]
pmeridian: Yes, we want to keep the sigmatmd = 0 when to time is associated

RecoMTD/TimingIDTools/src/TrackPUIDMVA.cc
```diff
@@ -0,0 +1,63 @@
...
+  vars_.push_back("etlMatchChi2");
+  vars_.push_back("etlMatchTimeChi2");
+  vars_.push_back("mtdt");
+  vars_.push_back("path_len");
+
+  std::string options("!Color:Silent");
+  std::string method("BDT");
+
+  mva_ = std::make_unique<TMVAEvaluator>();
+  mva_->initialize(options, method, weights_file, vars_, spec_vars_, true, false);  //use GBR, GradBoost
```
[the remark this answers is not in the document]
pmeridian: Yes it works with GBR so we use this as default, but it may be good to have the functionality to use TMVA if eg different training is used in the future

slava77:
@cmsbuild please test with cms-data/RecoMTD-TimingIDTools#1

slava77:
@cmsbuild please test with cms-data/RecoMTD-TimingIDTools#1

kpedro88:
@pmeridian is there some validation of the physics performance that you can share?

pmeridian:
2026D49 scenario:
DeltaT (tofpid_t0-gent0) at PU200 (20 ttbar events) for tracks with |dz|<1cm for MVA>0.8:
deltaT
behaviour is as expected: peak with 35ps resolution (primary vertex tracks) over a larger bump of pile-up tracks, very small tails. These performance are very similar to the "fastsim".

MVA helps to reduce tracks in the tails as expected (reducing tails of a factor >4):
Screenshot 2020-02-05 at 01 26 53
Next step after discussing with PFlow will be to apply a cut in the PATPackedCandidate, to remove fake association at high pile-up/

Single muon (Pt10) MTD association efficiency vs eta is as expected (~85% in BTL, ~100% in ETL):
Screenshot 2020-02-05 at 01 21 59

pmeridian:
It will be nice if this PR can be already integrated in CMSSW_11_1_0_pre3. Any further request/action from my side?

silviodonato:
Kind reminder to 
reconstruction: @perrotta @slava77
analysis: @santocch

silviodonato:
urgent

RecoMTD/TrackExtender/plugins/TrackExtenderWithMTD.cc
```diff
@@ -382,18 +504,19 @@ void TrackExtenderWithMTDT<TrackCollection>::produce(edm::Event& ev, const edm::
       pv = &(vtxH.product()->at(0));
```
slava77: the vertex validity check should not be needed. Absence of a vertex in this context is more likely due to a misconfiguration of the module.
Perhaps a check for the collection size is still OK.

slava77 (commented):
@pmeridian 
thank you for the thorough follow up to the last iteration of review.
Sorry for the few days delay now with this check from my side.

I think that we are pretty close now from the signoff.

silviodonato:
please test

slava77:
please abort

the cms-data PR parameter is missing

slava77:
@cmsbuild please test with cms-data/RecoMTD-TimingIDTools#1

mrodozov:
test parameters:
- pull_requests = cms-data/RecoMTD-TimingIDTools#1

mrodozov:
now both will work

slava77:
> now both will work

thanks. I hope we do not need another re-test before merging ;)

RecoMTD/TrackExtender/plugins/TrackExtenderWithMTD.cc
```diff
@@ -787,15 +789,13 @@ namespace {
...
-              TrackTofPidInfo tof;
-              computeTrackTofPidInfo(tof,
-                                     p.mag2(),
-                                     tot_pl,
-                                     itr->time(),
-                                     t_res_manual,  //put hit error by hand for the moment
-                                     t_vtx,
-                                     t_vtx_err,  //put vtx error by hand for the moment
-                                     false);
+              TrackTofPidInfo tof = computeTrackTofPidInfo(p.mag2(),
```
slava77: ```const TrackTofPidInfo tof =  ``` or event  ```auto tof = ```

slava77 (commented):
these are mostly for posterity. No need to update in this PR.

slava77:
it looks like this PR is adding about 5% to the total RECO time in PU200 (in a production setup, writing AODSIM), which is coming from about 5-fold increase in CPU use by the trackExtenderWithMTD (from 5.7 s to ~28 s/ev). [this is still based on a partial set of events in the test that's still running]

@pmeridian 
from the context or earlier responses I understood that you did not rerun this code on high PU and did not study the technical performance.
Please correct me if I'm wrong.

@kpedro88 please take a note for something to follow up.

slava77:
+1

for https://github.com/cms-sw/cmssw/pull/28815	d4a01be 
- jenkins tests pass and comparisons with the baseline show differences in variables related to MTD variables
- local tests show that this PR adds almost 6% to the total RECO time in the ttbar sample with PU200 in D49 setup in 11_1_0_pre2 (all MTD-related reco is back to ~30% of the total RECO time).
    - there is no significant change in the total AODSIM size: newly added floats ValueMaps are balanced with about 0.5% decrease in size of Tracks and TrackExtras from the trackExtenderWithMTD [BTW, trackExtra and hits from trackExtenderWithMTD are about 20% of AODSIM; something to be dropped in some not too distant future].
    - there is no visible change in the peak RSS use

silviodonato:
Kind reminder to @kpedro88 @santocch

kpedro88:
+upgrade

slava77:
some inclusive plots from the root files for the ttbar PU200 D49 setup, 200 events

At the track level, it looks like most of the population changes are in the ETL
all_sign1096VSorig_TTbar14TeV2026D49PU200localwf23434p21c_floatedmValueMap_trackExtenderWithMTD_pathLength_RECO_obj_values_

pf CH times are closer to 0
all_sign1096VSorig_TTbar14TeV2026D49PU200localwf23434p21c_max-250,min150,recoPFCandidates_particleFlow__RECO_obj_time_18

pf  electrons show the most significant effect
all_sign1096VSorig_TTbar14TeV2026D49PU200localwf23434p21c_max-250,min150,recoPFCandidates_particleFlow__RECO_obj_time_25



the time error of the 4D PVs is better (with some migration to the default no-time-set value [displayed at 0])
all_sign1096VSorig_TTbar14TeV2026D49PU200PATlocalwf23434p21c_log10recoVertexs_offlineSlimmedPrimaryVertices4D__PAT_obj_tError

The time is closer to zero as well with significantly reduced tail on the positive side
all_sign1096VSorig_TTbar14TeV2026D49PU200PATlocalwf23434p21c_recoVertexs_offlineSlimmedPrimaryVertices4D__PAT_obj_t

x (and y) errors are a bit better, but there is also a small component that gets worse
all_sign1096VSorig_TTbar14TeV2026D49PU200PATlocalwf23434p21c_log10recoVertexs_offlineSlimmedPrimaryVertices4D__PAT_obj_xError

At a glance, this all looks like going roughly in the direction of improving the performance.

silviodonato:
merge
@santocch

smuzaffar:
@silviodonato , this also needs https://github.com/cms-sw/cmsdist/pull/5538  which is not merged yet. I am going to merge it now. You might want to restart 11.1.0.pre3 release to get latest cmsdist

smuzaffar:
@silviodonato , Ok, I have restarted the 11.1.0.pre3 release jobs , so every thing should be fine now

santocch:
+1

PR #28819, opened 2020-01-29
do not run test-valgrind-memleak with full path

smuzaffar:
#### PR description:

This should fix the `test-valgrind-memleak` unit test which fails when run for IBs. As for IBs, we use tests from release area instead of re-building them locally so that is why `$CMSSW_BASE/test/$SCRAM_ARCH/test-valgrind-memleak` is not available during IB tests. We need to run `test-valgrind-memleak` directly which will be in `PATH` 

#### PR validation:

Local test builds and runs

### DIFF
--- a/Utilities/ReleaseScripts/test/test-valgrind.sh
+++ b/Utilities/ReleaseScripts/test/test-valgrind.sh
@@ -2,8 +2,9 @@
 
 SCRIPT_NAME=$(basename $0)
 TEST_NAME="test-valgrind-memleak"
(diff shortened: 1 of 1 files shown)


smuzaffar:
please test

smuzaffar:
@silviodonato , this should fix the unittest issue we get in latest 11.1.X IBs.

smuzaffar:
+core

silviodonato:
+1
thanks!

PR #28821, opened 2020-01-29
DQM: New EventInfo module for Harvesting

schneiml:
#### PR description:

In some harvesting workflows, `DQMEventInfo` is run to record the current run number etc. in the DQM output. This is not really legal, since `DQMEventInfo` is a `DQMEDAnalyzer` and not supposed to run in harvesting. 

While this worked by coincidence in the past, in #28622 such modules' output is no longer written to file. To fix this, I added a new module that is based on `DQMEDHarvester` which only records the values which make sense in harvesting.

On request of @mmusich , who wanted to have this functionality preserved for AlCa harvesting jobs.

I changed a few configurations where it is rather obvious that `DQMEventInfo` was used in harvesting, however, that list might be not complete.

#### PR validation:

Runs wf 1001, where the issue was spotted. Some configuration files changed here might not be tested in the PR tests.


mmusich:
Thanks @schneiml !

schneiml:
please test

Some changes in  a few `EventInfo` folders expected.

schneiml:
please test

Now without the function that clang complained about.

schneiml:
+1 

The new values appear exactly where they should be.

mmusich:
@schneiml there is something which is not working as I originally meant...
In the baseline the GUI header looks like:
image
while after this PR it looks like:
image

so the information about the run number and the time is lost.
Is there anything we can do about it?

schneiml:
@mmusich ah, good catch. DQMGUI reads that specific `iRun`, so I should keep that in a format that it understands. Probably set it to 999999 or sth. for multi-run harvesting. Give me a few minutes...

schneiml:
please test

This should make the run number also understandable for the DQMGUI again.

schneiml:
+1

The last change fixes the behavior that @mmusich pointed out. Note that the new comparison is a bit more confusing now: Now neither baseline nor pr show the correct header, since the `iRun` ME responsible for that is now unchanged and therefore not included in the comparison output.

mmusich:
@schneiml thanks!

I presume there is no way of including the begin run timestamp too, right?

schneiml:
@mmusich There probably is a way, the question is exactly what to use and where to put it. It looks like the old code looked at the first event (which should not work in harvesting), but it also does all sorts of other things

schneiml:
please test

Now with run start time, and I also `iLumisection`, just in case the GUI is picky about having that.

schneiml:
+1

Unless @mmusich finds another problem, that is. The "big" difference is now that `iLumisection` actually has a valid value. That value may not make much sense in case we process lumisections out-of-order (the normal case in offline), but then that value usually does not mean much anyways so I hope it is not too confusing. Actually, I wonder how that currently works in offline DQM... 

(And it turns out the GUI is perfectly happy with only one of `iRun` or `iLumisection` being set, good to know.)

mmusich:
thanks so much @schneiml!
I streamed the output `ALCAPROMPT` DQM file of wf 1001.0 from the last round of tests (file is available [here](https://cmssdt.cern.ch/SDT/jenkins-artifacts/pull-request-integration/PR-87ec37/4445/runTheMatrix-results/1001.0_RunMinBias2011A+RunMinBias2011A+TIER0EXP+ALCAEXP+ALCAHARVDSIPIXELCALRUN1+ALCAHARVD1+ALCAHARVD2+ALCAHARVD3+ALCAHARVD4+ALCAHARVD5/DQM_V0001_R000165121__Express__PCLTest__ALCAPROMPT.root))
into a private development GUI and that's what I see:

image

PR good to go on my side!

christopheralanwest:
+1

silviodonato:
+1

PR #28827, opened 2020-01-30
Adding parameters needed for PUjetID 94X and 102X

alefisico:
#### PR description:

  - Including the files and parameters needed for the latest PUjetID training for 94X and 102X samples. 
  - Once merged, we need to backport this PR to the corresponding CMSSW release for the latest nanoAOD. 
  - There is also a PR for the new weights for PUJetID https://github.com/cms-data/RecoJets-JetProducers/pull/10#issuecomment-579821257

#### PR validation:

Successfully run the runTheMatrix tests and here is the output: https://cernbox.cern.ch/index.php/s/XNTEPDN5VU01O84

### DIFF
--- a/RecoJets/JetProducers/python/PileupJetIDCutParams_cfi.py
+++ b/RecoJets/JetProducers/python/PileupJetIDCutParams_cfi.py
@@ -1,12 +1,62 @@
 import FWCore.ParameterSet.Config as cms
 
 ###########################################################
+## Working points for the 102X training
+###########################################################
+full_102x_chs_wp  = cms.PSet(
+    #4 Eta Categories  0-2.5 2.5-2.75 2.75-3.0 3.0-5.0
+
+    #Tight Id
+    Pt010_Tight    = cms.vdouble( 0.69, -0.35, -0.26, -0.21),
+    Pt1020_Tight   = cms.vdouble( 0.69, -0.35, -0.26, -0.21),
+    Pt2030_Tight   = cms.vdouble( 0.69, -0.35, -0.26, -0.21),
+    Pt3050_Tight   = cms.vdouble( 0.86, -0.10, -0.05, -0.01),
+
+    #Medium Id
+    Pt010_Medium   = cms.vdouble( 0.18, -0.55, -0.42, -0.36),
+    Pt1020_Medium  = cms.vdouble( 0.18, -0.55, -0.42, -0.36),
+    Pt2030_Medium  = cms.vdouble( 0.18, -0.55, -0.42, -0.36),
+    Pt3050_Medium  = cms.vdouble( 0.61, -0.35, -0.23, -0.17),
+
+    #Loose Id
+    Pt010_Loose    = cms.vdouble(-0.97, -0.68, -0.53, -0.47),
+    Pt1020_Loose   = cms.vdouble(-0.97, -0.68, -0.53, -0.47),
+    Pt2030_Loose   = cms.vdouble(-0.97, -0.68, -0.53, -0.47),
+    Pt3050_Loose   = cms.vdouble(-0.89, -0.52, -0.38, -0.30)
+)
+
+###########################################################
+## Working points for the 94X training
+###########################################################
+full_94x_chs_wp  = cms.PSet(
+    #4 Eta Categories  0-2.5 2.5-2.75 2.75-3.0 3.0-5.0
+
+    #Tight Id
+    Pt010_Tight    = cms.vdouble( 0.69, -0.35, -0.26, -0.21),
+    Pt1020_Tight   = cms.vdouble( 0.69, -0.35, -0.26, -0.21),
+    Pt2030_Tight   = cms.vdouble( 0.69, -0.35, -0.26, -0.21),
+    Pt3050_Tight   = cms.vdouble( 0.86, -0.10, -0.05, -0.01),
+
+    #Medium Id
+    Pt010_Medium   = cms.vdouble( 0.18, -0.55, -0.42, -0.36),
+    Pt1020_Medium  = cms.vdouble( 0.18, -0.55, -0.42, -0.36),
+    Pt2030_Medium  = cms.vdouble( 0.18, -0.55, -0.42, -0.36),
+    Pt3050_Medium  = cms.vdouble( 0.61, -0.35, -0.23, -0.17),
+
+    #Loose Id
+    Pt010_Loose    = cms.vdouble(-0.97, -0.68, -0.53, -0.47),
+    Pt1020_Loose   = cms.vdouble(-0.97, -0.68, -0.53, -0.47),
+    Pt2030_Loose   = cms.vdouble(-0.97, -0.68, -0.53, -0.47),
+    Pt3050_Loose   = cms.vdouble(-0.89, -0.52, -0.38, -0.30)
+)
+
+###########################################################
 ## Working points for the 81X training (completed in 80X with variable fixes)
 ###########################################################
 full_81x_chs_wp  = cms.PSet(
     #4 Eta Categories  0-2.5 2.5-2.75 2.75-3.0 3.0-5.0
 
-    #Tight Id            
+    #Tight Id
     Pt010_Tight    = cms.vdouble( 0.69, -0.35, -0.26, -0.21),
     Pt1020_Tight   = cms.vdouble( 0.69, -0.35, -0.26, -0.21),
     Pt2030_Tight   = cms.vdouble( 0.69, -0.35, -0.26, -0.21),
@@ -31,7 +81,7 @@ full_81x_chs_wp  = cms.PSet(
 full_80x_chs_wp  = cms.PSet(
     #4 Eta Categories  0-2.5 2.5-2.75 2.75-3.0 3.0-5.0
 
-    #Tight Id            
+    #Tight Id
     Pt010_Tight    = cms.vdouble( 0.26, -0.34, -0.24, -0.26),
     Pt1020_Tight   = cms.vdouble( 0.26, -0.34, -0.24, -0.26),
     Pt2030_Tight   = cms.vdouble( 0.26, -0.34, -0.24, -0.26),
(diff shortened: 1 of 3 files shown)


slava77:
@cmsbuild please test

slava77:
> There is also a PR for the new weights for PUJetID [cms-data/RecoJets-JetProducers#10 (comment)](https://github.com/cms-data/RecoJets-JetProducers/pull/10#issuecomment-579821257)

@alefisico 
what is the plan to use these?
Are the new IDs going to be available only for analysis purposes or are they going to be imminently added to the production setup (mini or nano)?
For the latter case, perhaps it's worth to make the change now.
Please comment.
Thank you.

alefisico:
Hi @slava77 
yes, the idea is for these PRs to be included in imminently in the next production setup.

slava77:
> Hi @slava77
> yes, the idea is for these PRs to be included in imminently in the next production setup.

OK, but what is the plan to enable the new IDs in the master?
If these new IDs are the new recommended ones, I think that they should already be enabled with this PR.

alefisico:
sorry, maybe I got confused, but we *are* including the latest IDs: https://github.com/cms-sw/cmssw/blob/086826ca6731bf82abde15a9341a7e7eed033aa0/RecoJets/JetProducers/python/PileupJetIDCutParams_cfi.py#L6-L51 
or maybe I didn't understand the comment

slava77:
> sorry, maybe I got confused, but we _are_ including the latest IDs:
> 
> https://github.com/cms-sw/cmssw/blob/086826ca6731bf82abde15a9341a7e7eed033aa0/RecoJets/JetProducers/python/PileupJetIDCutParams_cfi.py#L6-L51
> 
> or maybe I didn't understand the comment

This part just defines a new PSet, which has no impact during run time of the standard workflows like miniAOD or nanoAOD.
I expect that the new PSet is used to make the pileupJetId, in pileupJetId.algos configuration parameter values, to replace the current default ```full_81x_chs```

alefisico:
isn't it this part? https://github.com/cms-sw/cmssw/blob/086826ca6731bf82abde15a9341a7e7eed033aa0/RecoJets/JetProducers/python/PileupJetID_cfi.py#L14-L15
or you mean, the explicit part for mini/nanoAOD to change for each dataset and year?

slava77:
> isn't it this part?
> 
> https://github.com/cms-sw/cmssw/blob/086826ca6731bf82abde15a9341a7e7eed033aa0/RecoJets/JetProducers/python/PileupJetID_cfi.py#L14-L15
> 
> or you mean, the explicit part for mini/nanoAOD to change for each dataset and year?

This PR is only adding new PSets which are not used anywhere in running modules.

From the earlier messages I understood that the new IDs will be used in production.
We need some matrix workflows where this is actually enabled.
It also seems like the latest IDs should be applied to the latest Run2 simulation.

alefisico:
thanks again for the help, I think I got it, you mean we need to have this line: https://github.com/cms-sw/cmssw/blob/086826ca6731bf82abde15a9341a7e7eed033aa0/RecoJets/JetProducers/python/PileupJetID_cfi.py#L17
for the samples for different years. Is this correct?
Do you perhaps have an example how to implement something like this? I haven't seen a producer which this functionality.

slava77:
> thanks again for the help, I think I got it, you mean we need to have this line:
> 
> https://github.com/cms-sw/cmssw/blob/086826ca6731bf82abde15a9341a7e7eed033aa0/RecoJets/JetProducers/python/PileupJetID_cfi.py#L17
> 
> for the samples for different years. Is this correct?
> Do you perhaps have an example how to implement something like this? I haven't seen a producer which this functionality.

I'm not sure about the reference to the different years.

I will need some clarification on the plans for processing.
@ahinzmann @lathomas please advise
Thank you.

ahinzmann:
Hi Slava,
the situation is the following. We have separate trainings for each year of running now.
This means we will need modifiers to adapt for each year.

The shortterm goal is to include this is a re-nanoAOD. For this I assume the modifier and era is likely to be defined in cms-nanoaod? @peruzzim

The longterm goal would to include it into UL-re-miniAOD. Do we have an era for each year of UL-re-miniAOD?

slava77:
@ahinzmann 
thank you for the clarifications.

This covers my question regarding the narrowly defined re-miniAOD in 106X release.
It looks straightforward to me now that this update applies directly to re-nanoAOD running on the old Run-2 miniAODs made in 94X and in 102X.
Do you expect that the same training will work for the 106X UL samples?

In the master version processing workflows we should be changing the defaults,
unless the 81X-based training is still somehow better.
We have eras defined by the year (Run2_2018, Run2_2017 etc) with corresponding jme run2_jme_2017; there is no run2_jme_2018, but it can be added.

There is no setup yet for a targeted reminiAOD from UL.
I hope we get it in a matter of a week or so. This would define what to do precisely for the backport.

ahinzmann:
Discussing a bit more with @alefisico the most senseful thing to do here is to simply put the 102X-training into the master (regardless of year). This should be better than the 80X training. And a good starting point for re-miniAOD from UL (through we don't know how far it is from optimal yet).

slava77:
@cmsbuild please test

alefisico:
The runTheMatrix tests are failing because this PR is still not merged: https://github.com/cms-data/RecoJets-JetProducers/pull/10 
Please let me know if there is something I need to do from my side

slava77:
test parameters
- pull_request = cms-data/RecoJets-JetProducers#10

slava77:
@cmsbuild please test

perrotta:
@cmsbuild please test

ahinzmann:
@alefisico In these tests, the discriminator seems to always peak at -1, regardless of how much pileup is in the sample, e.g.
https://cmssdt.cern.ch/SDT/jenkins-artifacts/baseLineComparisons/CMSSW_11_1_X_2020-02-04-2300+35ca6d/34963/validateJR/all_mini_OldVSNew_RunEGamma2018Cwf136p874/all_mini_OldVSNew_RunEGamma2018Cwf136p874c_patJets_slimmedJets__reRECO_obj___userFloats__3_.png

Was this also the case in the tests done in 10_2_X? It looks fishy to me.

camclean:
@ahinzmann, did you apply a pt cut to the jets in this plot?

camclean:
Did the code change so that jets with pt > 50 GeV return -1?

perrotta:
Is there any update on the explanation of the outputs?

RecoJets/JetProducers/python/PileupJetIDCutParams_cfi.py
```diff
@@ -1,12 +1,62 @@
...
+)
+
+###########################################################
+## Working points for the 94X training
+###########################################################
+full_94x_chs_wp  = cms.PSet(
+    #4 Eta Categories  0-2.5 2.5-2.75 2.75-3.0 3.0-5.0
+
+    #Tight Id
+    Pt010_Tight    = cms.vdouble( 0.69, -0.35, -0.26, -0.21),
```
slava77: the values appear to be the same.
Do we really need to copy-paste this fairly large number of values?
Simply ```full_94x_chs_wp = full_102x_chs_wp.clone()``` would work.
Same comment applies to the other points.

RecoJets/JetProducers/python/PileupJetIDParams_cfi.py
```diff
@@ -1,6 +1,212 @@
...
+                    "jetRchg"   ,
+                )
+                ),
+            cms.PSet(
+                jEtaMin = cms.double(2.5),
+                jEtaMax = cms.double(2.75),
+                tmvaWeights   = cms.FileInPath("RecoJets/JetProducers/data/pileupJetId_102X_Eta2p5To2p75_chs_BDT.weights.xml.gz"),
+                tmvaVariables = cms.vstring(
+                    "nvtx"      ,
+                    "dR2Mean"   ,
```
slava77: it looks like the same variable name vector repeats for the first 3 ranges. These can be initialized from a separately defined (temporary) vector

RecoJets/JetProducers/python/PileupJetIDParams_cfi.py
```diff
@@ -1,6 +1,212 @@
...
+            "jetPt" ,
+            "jetEta",
+        ),
+        JetIdParams = full_102x_chs_wp,
+        label = cms.string("full")
+)
+
+###################################################################################################################
+full_94x_chs = cms.PSet(
+        impactParTkThreshold = cms.double(1.),
```
slava77: is anything changing here other than the training file names and the working point PSet?
If not, a more compact and perhaps more maintainable way would be to use

```python
full_94x_chs = full_102x_chs.clone(JetIdParams = full_94x_chs_wp)
full_94x_chs.trainings[0].tmvaWeights = "RecoJets/JetProducers/data/pileupJetId_94X_Eta0p0To2p5_chs_BDT.weights.xml.gz"
...
```

Having the older IDs listed first may also be more maintainable for cases of future additions.

RecoJets/JetProducers/python/PileupJetIDCutParams_cfi.py
alefisico: These numbers are preliminary the same, but it will potentially change in the near future. (and could *not* be the same for all the years like now) Perhaps is better to include the structure now?

RecoJets/JetProducers/python/PileupJetIDParams_cfi.py
alefisico: True, I will change it

RecoJets/JetProducers/python/PileupJetIDParams_cfi.py
alefisico: same as previous comment. We are currently using the same values for WP in these years, but they will potentially change in the near future. What is your suggestion?

slava77: ```full_94x_chs = full_102x_chs.clone(JetIdParams = full_94x_chs_wp)``` is applying different working points for 94X vs 102X. The suggestion in the earlier comment simply allows not to repeat what's known to be the same.

RecoJets/JetProducers/python/PileupJetIDCutParams_cfi.py
slava77: Looking back, indeed the values of the working points are rather different for different training cases.
What are the targets in training?

If the development strategy is to derive the working points from some fixed numerical target (like x% efficiency or rejection), then there would be very likely no accidental agreement. This, however doesn't seem to be the case this time.

If the WPs are weakly defined [this time?] with a replacement based on judgement calls,
it would be OK to copy-paste if the older values can update independently.

However it sounds like the development strategy this time is to first introduce placeholders to be replaced after some later processing. 
[I'm judging from the history of the past updates and guessing from the symptoms of potentially broken WPs visible in the validation results].
If you feel like a shorter ```new_wp = old_wp.clone() #TEMPORARY ``` is less acceptable,
please at least add an explicit comment inline that these are to be replaced soon.

alefisico:
Hi all, 
we needed to change a couple of things after the bug that @ahinzmann found. Since now several things changed in CMSSW, I think it is better to close this PR and start clean. 
New PR is #28984

PR #28829, opened 2020-01-30
DQM: Remove DQMDcsInfo

schneiml:
#### PR description:

The `Info/EventInfo/reportSummaryMap` DCS plot was broken since a long time. I have never seen the 'Physics Declared' actually trigger, and since some changes early 2019 also the actual DCS bit information is completely broken (I have not that much confidence that it was actually correct before, e.g. when some lumisections were not processed).

DQMProvInfo provides roughly the same functionality (and more), is much
simpler and actually works, and is used in online since a long time.

This PR replaces DQMDcsInfo with DQMProvInfo also in offline.

#### PR validation:

Runs 136.85, more to be seen. DQMGUI render plugins need to be checked if they work on the new plot also in offline.

Update: runs and seems to work correctly, Render Plugin handles both plots correctly.

We might want to backport this in the future to fix the DCS plot for the coming UL rereco.


schneiml:
please test

some new plots expected, as `DQMProvInfo` records more things compared to `DQMDcsInfo`. Also, ideally I caught all the references to `DQMDcsInfo` and nothing crashes.

schneiml:
It seems we'll need to have a look at optional vs. required dependencies on the old vs. new DCS products.

schneiml:
So, actually what crashed here are *all* MC based workflows. While the old `DQMDcsInfo` silently ignores the absence of a DCS record, `DQMProvInfo` does not...

schneiml:
please test

Should run through now, if the output is correct is another question.

schneiml:
Changes are mostly as expected.

- The new ProvInfo plots will probably need to be blacklisted in the comparison (hostname/directory etc.), or we disable them. 
- We lost the harvesting GT (used to be in `CMSSWInfo`), since the plot is now all done in RECO.
- `reportSummary` might have the wrong value; not sure how it is supposed to be defined.
- `processedLS` is now gone, but `reportSummaryMap` has the relevant information in the `valid` row.

boudoul:
Hi @schneiml For my curiosity, do you have a screenshot (or something)to show me  how the reportsummaryMap  looks like in the GUI  ? thanks.

schneiml:
please test

Rebased to remove the merge conflict. The new behaviour was considered fine so far, so we'll move forward here soon.

@boudoul The previous test result produced a nice comparison: https://tinyurl.com/tsq8649
Note that there is only a few events from a single lumi processed, which the new (left) plot correctly shows while the old (right) plot shows... something. The new plot also adds the beam status, compared to the old one.

schneiml:
+1

Seems fine on our side. Be prepared for some false positives in the bin-to-bin comparison once this is merged: We'll need to put some of the `ProvInfo` values into the blacklist (if they are not there already).

rekovic:
+1

silviodonato:
+1

PR #28844, opened 2020-01-31
Add customizer for PUPPI tune v13

ahinzmann:
#### PR description:

This PR adds a customizer to run PUPPI tune v13 and recompute slimmedJetsPuppi and slimmedMETsPuppi.
The general goal is to switch to PUPPI tune v13 in the master now and for the UL-re-MiniAOD in Summer. Until then we would like to use this customizer to produce JetMET-specific-NanoAODs (custom_jme_cff) with PUPPI tune v13 for validation and calibration.

The PR also adds puppi weighted multiplicity observables to JetMET-specific-NanoAODs.

The most recent presentation of the performance of PUPPI v13 is here:
https://indico.cern.ch/event/876142/#5-update-on-puppi-tuning

#### PR validation:

It was checked that slimmedJetsPuppi and slimmedMETsPuppi are modified as expected and the corresponding JetMET-specific-NanoAOD entries change accordingly using workflow 10824.0 with this customizer:

cmsDriver.py step6  --conditions auto:phase1_2018_realistic -n 10 --era Run2_2018 --eventcontent NANOEDMAODSIM --filein file:step3_inMINIAODSIM.root -s NANO --datatier NANOAODSIM --geometry DB:Extended --customise_commands "from CommonTools.PileupAlgos.customizePuppiTune_cff import UpdatePuppiTuneV13; UpdatePuppiTuneV13(process); from PhysicsTools.NanoAOD.custom_jme_cff import PrepJMECustomNanoAOD; PrepJMECustomNanoAOD(process)" --fileout file:step6.root  > step6_TTbar_13+TTbar_13TeV_TuneCUETP8M1_2018_GenSimFull+DigiFull_2018+RecoFull_2018+HARVESTFull_2018+ALCAFull_2018+NanoFull_2018.log

This should be backported to 10_6 for JetMET-specific-NanoAOD production.


slava77:
@ahinzmann 
a process modifier would be a preferred solution to the customize method proposed here.

slava77:
also, for the master I would expect that the new tune becomes a default

slava77:
@cmsbuild please test

ahinzmann:
Hi Slava,
I believe we will need the following:
1. The customization in the PR, in order to re-compute PUPPI on top of existing UL-MiniAODs to produce UL-JME-custom-NanoAOD
2. Enable the tune via modifier for the UL-re-MiniAOD. Do we have an era for this?
3. Enable the tune for Run3 (assume this is master)
4. Adapt the tune via modified for Phase-2
For point 2. I would need to know the era, the rest I can implement.

perrotta:
@cmsbuild please test

peruzzim:
please see #27889

silviodonato:
@ahinzmann what is the status of this PR? Are you waiting for the review or are you going to have an update of the code?

ahinzmann:
this is ready to be reviewed

PhysicsTools/NanoAOD/python/custom_jme_cff.py
```diff
@@ -309,8 +309,14 @@ def addTable(self, proc, recoJetInfo):
...
           rawFactor = jetTable.variables.rawFactor,
         )
+    elif "puppi" in recoJetInfo.jet:
+      tableContents = JETVARS.clone(
+        puppiMultiplicity = Var("userFloat('patPuppiJetSpecificProducer:puppiMultiplicity')",float,doc="Sum of PUPPI weights of particles in the jet"),
+        neutralPuppiMultiplicity = Var("userFloat('patPuppiJetSpecificProducer:neutralPuppiMultiplicity')",float,doc="Sum of PUPPI weights of neutral particles in the jet")
+      )
     else:
       tableContents = JETVARS.clone()
+
```
perrotta: This extra empty line is not needed

perrotta:
> Hi Slava,
> I believe we will need the following:
> 
> 1. The customization in the PR, in order to re-compute PUPPI on top of existing UL-MiniAODs to produce UL-JME-custom-NanoAOD
> 2. Enable the tune via modifier for the UL-re-MiniAOD. Do we have an era for this?
> 3. Enable the tune for Run3 (assume this is master)
> 4. Adapt the tune via modified for Phase-2
>    For point 2. I would need to know the era, the rest I can implement.

About this plan: 
- This PR represents point 1, isn't it?
- Then, for point 2, do I understand correctly that you will remove this customization and substitute it with a modifier? And why not preparing a modifier since now also for nanoAOD, which could be ```run2_nanoAOD_106Xv2``` or (probably better) another name that will get renamed as ```run2_nanoAOD_106Xv2``` as soon as that campaign will get started?

@peruzzim @fgolf do you have any comment/suggestion?

ahinzmann:
> > Hi Slava,
> > I believe we will need the following:
> > 
> > 1. The customization in the PR, in order to re-compute PUPPI on top of existing UL-MiniAODs to produce UL-JME-custom-NanoAOD
> > 2. Enable the tune via modifier for the UL-re-MiniAOD. Do we have an era for this?
> > 3. Enable the tune for Run3 (assume this is master)
> > 4. Adapt the tune via modified for Phase-2
> >    For point 2. I would need to know the era, the rest I can implement.
> 
> About this plan:
> 
>     * This PR represents point 1, isn't it?
> 
>     * Then, for point 2, do I understand correctly that you will remove this customization and substitute it with a modifier? And why not preparing a modifier since now also for nanoAOD, which could be `run2_nanoAOD_106Xv2` or (probably better) another name that will get renamed as `run2_nanoAOD_106Xv2` as soon as that campaign will get started?
> 
> 
> @peruzzim @fgolf do you have any comment/suggestion?

This PR implements 1., 3. and 4. The 106-backport I implemented 1. and 2.

perrotta:
please test

perrotta:
@ahinzmann , why there are changes also in AK8 jets, and not only in Puppi ones?

image

About results with Puppi jets: with v13 there are new and large horns at large pseudorapidity: is it expected/accepted? (I don't see them cited in the presentation linked from the PR description).

image

(All results from the PR tests with RunDoubleEG2017C wf 136.793

ahinzmann:
> @ahinzmann , why there are changes also in AK8 jets, and not only in Puppi ones?

because the AK8 jets in MiniAOD are puppi AK8 jets.

> image
> 
> About results with Puppi jets: with v13 there are new and large horns at large pseudorapidity: is it expected/accepted? (I don't see them cited in the presentation linked from the PR description).
> 
> image
> 
> (All results from the PR tests with RunDoubleEG2017C wf 136.793

the new tune is expected to behave like CHS in eta>2.5, is increase of horns from underlying detector calibration issues in 2017 is possibly. Is there a comparison of DoubleEG2017C also with UL or Run3 conditions where we don't want such nasty features to appear?

slava77:
> Is there a comparison of DoubleEG2017C also with UL or Run3 conditions where we don't want such nasty features to appear?

I naively hoped that the master version setup has the UL (or even better) conditions in place.
@christopheralanwest   @tocheng  please clarify.

ahinzmann:
another caveat of the puppi jet eta distribution is that the jet response increases signficantly in 2.5<eta<3.0, thus the JEC applied in MiniAOD are over-correcting. We are deriving updated JECs for this PUPPI tune v13 in the context of UL2017. Once we have these corrections (order of 2 weeks) we can look again at this particular eta distribution.

ahinzmann:
The fact that the horns in 2.5<eta<3.0 also appear in 2018 data seems to point to JECs rather than low level calibration/noise:
https://cmssdt.cern.ch/SDT/jenkins-artifacts/baseLineComparisons/CMSSW_11_1_X_2020-02-12-2300+5cc635/35064/validateJR/all_mini_OldVSNew_RunEGamma2018Cwf136p874/all_mini_OldVSNew_RunEGamma2018Cwf136p874c_patJets_slimmedJetsPuppi__reRECO_obj_eta.png

ahinzmann:
is it possible to re-run the comparisons with non-standard global tags (once we have them)?

perrotta:
> is it possible to re-run the comparisons with non-standard global tags (once we have them)?

Automatic tests are based on the default GT assignment of any workflow.
Please provide some comparison with what you believe is the "more correct" GT to be used: as you can see, issues as such are already visible when running on a limited number of events (here 100), therefore once you have the "correct" setup it should be rather fast to provide them. These plots should be then approved by the POG before the new Puppi tune can be integrated in CMSSW.

ahinzmann:
agree. In the meantime, could we integrate the 10_6 backport that does not change the default behavior and integrates the customization that we need to produce the JME NanoAODs to derive the full set of MC and data JECs for the new tune?

perrotta:
> agree. In the meantime, could we integrate the 10_6 backport that does not change the default behavior and integrates the customization that we need to produce the JME NanoAODs to derive the full set of MC and data JECs for the new tune?

The policy is that whatever backport must be integrated in the master first.
I would suggest to prepare two identical PRs, one for the master and one for 10_6, which do not change any default.
Then, as soon as your studies and calibrations are finalized, you can submit an additional PR which changes the default in the master

slava77:
>  I would suggest to prepare two identical PRs, one for the master and one for 10_6, which do not change any default.

I'm not sure that this is efficient.
This way we will not be able to test fully the new code/configuration with the standard tools (especially the parts that go beyond the short matrix test in the PR testing) until after it gets integrated for use in special production(s) in the older release.

perrotta:
> > I would suggest to prepare two identical PRs, one for the master and one for 10_6, which do not change any default.
> 
> I'm not sure that this is efficient.
> This way we will not be able to test fully the new code/configuration with the standard tools (especially the parts that go beyond the short matrix test in the PR testing) until after it gets integrated for use in special production(s) in the older release.

The change in the Puppi tune still has to be corredated with the correct conditions, according to https://github.com/cms-sw/cmssw/pull/28844#issuecomment-585849163
And after that, it has to be approved by the POG

Having it coded in the customizer allows them to derive the correct JEC for it. At that point, it can be imported in the release.

Did you have a different procedure in mind, @slava77 ?

slava77:
> The change in the Puppi tune still has to be corredated with the correct conditions, according to [#28844 (comment)](https://github.com/cms-sw/cmssw/pull/28844#issuecomment-585849163)
> And after that, it has to be approved by the POG
> 
> Having it coded in the customizer allows them to derive the correct JEC for it. At that point, it can be imported in the release.

with this plan in mind, I agree that it's  more practical to accumulate the changes until a one-step update.

ahinzmann:
This PR is now only adding the customizer, not changing the master config.
A PR to change the master config will follow in a couple of weeks, once we have reran our validations with JECs derived for the new tune.

perrotta:
please test

peruzzim:
> > Hi Slava,
> > I believe we will need the following:
> > 
> > 1. The customization in the PR, in order to re-compute PUPPI on top of existing UL-MiniAODs to produce UL-JME-custom-NanoAOD
> > 2. Enable the tune via modifier for the UL-re-MiniAOD. Do we have an era for this?
> > 3. Enable the tune for Run3 (assume this is master)
> > 4. Adapt the tune via modified for Phase-2
> >    For point 2. I would need to know the era, the rest I can implement.
> 
> About this plan:
> 
> * This PR represents point 1, isn't it?
> * Then, for point 2, do I understand correctly that you will remove this customization and substitute it with a modifier? And why not preparing a modifier since now also for nanoAOD, which could be `run2_nanoAOD_106Xv2` or (probably better) another name that will get renamed as `run2_nanoAOD_106Xv2` as soon as that campaign will get started?
> 
> @peruzzim @fgolf do you have any comment/suggestion?

On this, if I understand correctly I would suggest the following:

- 1. this can be done with a specific modifier or - better - just an option that is local to the JME-NanoAOD customizer function, because it will be needed only there in view of 3.
- 2. see #27889 i.e. protect the new tune by the `run2_miniAOD_devel` _in the backport to 10_6_X only_
- 3. for this I think we can just switch it on by default in master, right? as suggested by @slava77 
- 4. same as point 3. maybe, but which release will we use for that - 11_0_X or 11_1_X? Of which production for Phase-2 are we talking about here?

ahinzmann:
We need 1. (what is in this PR) asap to derive UL JEC calibrations with the new PUPPI tune. It cannot be done with a modifier since it it is MiniAOD->NanoAOD workflow (and PUPPI is typically ran in AOD->MiniAOD). I can move this piece of code to the JME Nano area if preferred?
So 2.+3.+4. will be in a separate PR

peruzzim:
I don't see why the function should be moved to the NanoAOD area, you can just call it as you plan to do already when you want to run this workflow. The NanoAOD package is not a temporary staging area for customizations.

PhysicsTools/NanoAOD/python/custom_jme_cff.py
```diff
@@ -392,7 +397,7 @@ def AddPileUpJetIDVars(proc):
   proc.jetTable.variables.pull     = Var("userFloat('pull')",    float, doc="magnitude of pull vector", precision= 6) 
   proc.jetTable.variables.jetR     = Var("userFloat('jetR')",    float, doc="fraction of jet pT carried by the leading constituent", precision= 6) 
   proc.jetTable.variables.jetRchg  = Var("userFloat('jetRchg')", float, doc="fraction of jet pT carried by the leading charged constituent", precision= 6) 
-  proc.jetTable.variables.nCharged = Var("userInt('nCharged')",  float, doc="number of charged constituents", precision= 6) 
+  proc.jetTable.variables.nCharged = Var("userInt('nCharged')",  int, doc="number of charged constituents", precision= 6)
```
peruzzim: please remove precision here, with integer type it makes no sense

PhysicsTools/NanoAOD/python/custom_jme_cff.py
```diff
@@ -309,6 +309,11 @@ def addTable(self, proc, recoJetInfo):
           area      = jetTable.variables.area,
           rawFactor = jetTable.variables.rawFactor,
         )
+    elif "puppi" in recoJetInfo.jet:
+      tableContents = JETVARS.clone(
+        puppiMultiplicity = Var("userFloat('patPuppiJetSpecificProducer:puppiMultiplicity')",float,doc="Sum of PUPPI weights of particles in the jet"),
```
peruzzim: please evaluate whether you really want full precision for these float numbers

perrotta:
please test
(just to check whether the non openable input file is a transient)

perrotta:
+1
- A customizer is added  to run PUPPI tune v13 and recompute slimmedJetsPuppi and slimmedMETsPuppi
- Once corresponding JEC corrections will be prepared, v13 will be enabled for Run3 and UL re-miniAOD (with a modifier, instead than a customizer)
- As of now, the PR does not modify any workflow: jenkins tests pass and show no differences

peruzzim:
+xpog

Are `puppiMultiplicity` and `neutralPuppiMultiplicity` normalized to 1? Otherwise careful, 6-bit precision can be little for numbers with a wide range. I assume you checked it as requested.

santocch:
+1

silviodonato:
@ahinzmann
could you reply to Marco's comment?

> Are `puppiMultiplicity` and `neutralPuppiMultiplicity` normalized to 1? Otherwise careful, 6-bit precision can be little for numbers with a wide range. I assume you checked it as requested.

silviodonato:
+1
@ahinzmann I think you can ignore my comment above, since you are moving in the opposite way float --> int

PR #28847, opened 2020-02-01
cmsDriver Test workflow based on DD4hep: first steps

fabiocos:
#### PR description:

This PR introduces some configuration updates in view of preparing a cmsDriver test workflow based on DD4hep, following a request by @ianna . The code is not yet mature for a regular insertion in PyReleaseValidation, so I add a simple script to ```Geometry/CMSCommonData/test/dd4hep``` with the step1 cmsDriver command for a SingleMuon test workflow based on the 2021 DD4hep geometry made available.

In order to interface this seamlessly with ```ConfigBuilder.py``` without ad-hoc hacks, I update the name of ```DD4hep_GeometrySim_cff``` to the more standard ```GeometryDD4hepExtended2021_cff```, add a Reco counterpart of it, and plug this into ```Configuration/StandardSequences/python/GeometryConf_cff``` so as to let ```ConfigBuilder.py``` to pick it from the command line in a standard way.

The Reco configuration needs to be adapted further, as the test workflow is crashing in the HCAL part: @bsunanda do you see what is wrong?

At the same time I move the process modifier ```dd4hep_cff``` from ```Configuration/Eras``` where I had originally accepted it to ```Configuration/ProcessModifiers``` as it looks a more suitable location to me, in the spirit of #22718 . The places where this is added so far by @bsunanda are updated accordingly, and its use is also extended to the ```g4SimHits``` declaration.

#### PR validation:

Code compiles, a test configuration is created but fails at runtime. It may be used to progressively test the completion of a geometry configuration fully operational for DD4hep.


fabiocos:
please test

ianna:
@fabiocos - thanks!

cvuosalo:
+1

civanch:
+1

silviodonato:
+operations

Configuration/Geometry/python/GeometryDD4hepExtended2021Reco_cff.py
```diff
@@ -0,0 +1,37 @@
...
+#Muon
+from Geometry.MuonNumbering.muonNumberingInitialization_cfi import *
+from RecoMuon.DetLayers.muonDetLayerGeometry_cfi import *
+from Geometry.GEMGeometryBuilder.gemGeometry_cfi import *
+
+#  Alignment
+from Geometry.TrackerGeometryBuilder.idealForDigiTrackerGeometry_cff import *
+from Geometry.CSCGeometryBuilder.idealForDigiCscGeometry_cff import *
+from Geometry.DTGeometryBuilder.idealForDigiDtGeometry_cff import *
+trackerGeometry.applyAlignment = cms.bool(False)
```
mmusich: out of curiosity why is this switched off (also in the regular 2021 reco geometry)?

fabiocos: @mmusich as far as this PR is concerned, the RECO part is just a copy of the regular geometry, and should be considered as a place holder, waiting for further adjustments. Whether the  alignment being false in the regular geometry is meaningless should be clarified by you and AlCa,
as far as I can see this is there since many years probably for historical reason, when the 2019 scenario (its original until recently) was just a study scenario for the far future. Most probably it should be updated nowadays (also to avoid odd results, and being forgotten), do you confirm?

mmusich: unless there is some hidden reason I am not aware of, it looks a plain bug to me... @ianna

ianna: @mmusich - The XML scenarios describe an Ideal geometry, e.g. without alignment. They are used to produce the ideal geometry payloads. Unless the *RECOGeometry in DB payloads should be already aligned, it's not a bug.
A corresponding DB scenario is used in production.

mmusich: then, will you please explain why all the run2 reco geometries (i.e.: https://github.com/cms-sw/cmssw/blob/master/Configuration/Geometry/python/GeometryExtended2018Reco_cff.py) is not treated in the same way?

ianna: as far as I understand they use "fakeForIdeal" alignment which is == no alignment

kpedro88:
+upgrade

Configuration/Geometry/python/GeometryDD4hepExtended2021Reco_cff.py
silviodonato: ok, I think the mystery is solved... shall we proceed with the integration?

fabiocos: for this specific PR I believe that flag is anyway ineffective for the time being. Whether it makes sense or not to have it in the DDD version should be addressed separately. In regular production we are using the DB version of the configuration, @ianna are these RECO versions used just for tests/developments (outside of Phase2 I mean)? Anyway if a fake alignement is used, the value of the flag looks irrelevant. 

I run
```
cmsDriver.py step2 --conditions auto:phase1_2018_realistic -s DIGI:pdigi_valid,L1,DIGI2RAW,HLT:@relval2018 --datatier GEN-SIM-DIGI-RAW -n 10 --geometry Extended2018 --era Run2_2018 --eventcontent FEVTDEBUGHLT --filein file:step1.root --fileout file:step2.root --no_exec
```

expand the configuration with ```edmConfigDump``` and search for ```applyAlignment```, used by all the geometry ES producers, finding 
```
11:29 cmsdev25 1181> grep applyAlignment full2.py 
    applyAlignment = cms.bool(True),
    applyAlignment = cms.bool(True),
    applyAlignment = cms.bool(False)
    applyAlignment = cms.bool(False)
    applyAlignment = cms.bool(False)
    applyAlignment = cms.bool(False),
    applyAlignment = cms.bool(False),
    applyAlignment = cms.bool(False),
    applyAlignment = cms.bool(False),
    applyAlignment = cms.bool(True),
```

Indeed we find inside

```
process.fakeForIdealAlignment = cms.ESProducer("FakeAlignmentProducer",
    appendToDataLabel = cms.string('fakeForIdeal')
)
```

mmusich: >  In regular production we are using the DB version of the configuration

yes in production we are using the DB geometry (aligned) so it's not a problem. My concern is exactly about the use-case of running the geometry from XML (as in your tests) in which it gives results different from production. As long as the person running the test knows it it's fine, but otherwise...

silviodonato:
+1
I think the discussion about `trackerGeometry.applyAlignment = cms.bool(False)` is more general and goes beyond the target of this PR

PR #28872, opened 2020-02-05
HGCAL Validation technical changes

rovere:
#### PR description:

This PR fixes a number of bugs in the validation package (mostly related to the score computation for `MultiClusters`)  for the `HGCal` detector.
It also adds the possibility to filter `CaloParticles` based on the number of `SimClusters` associated with the `CaloParticle` itself. 
Efficiencies and related metrics are now computed with respect to `CaloParticles` with at most 1 `SimCluster`, to make their interpretation easier.

This PR will have a limited impact on existing `Phase2` workflows, but its inclusion is preferable before including `TICL` in the official `Phase2` reconstruction. This will allow making some preliminary validation, even if also this has to be considered as a work in progress and not as a final validation suite for the full `HGCAL` reconstruction.

#### PR validation:

Usual limited `runTheMatrix`


jfernan2:
please test

jfernan2:
@rovere There are many changes in histograms at folders  SelectedCaloParticles and hgcalLayerClusters
I understand they are expected with this PR, please let me know if not. Thanks

rovere:
@jfernan2 the changes are related to the selection I now apply on the `CaloParticles`, so they are expected. I took a look at them and they goes in the right direction.

jfernan2:
+1
Thanks @rovere

silviodonato:
+1

PR #28874, opened 2020-02-05
Make FastTimerService easier to use

kpedro88:
#### PR description:

The [FastTimerService](https://twiki.cern.ch/twiki/bin/viewauth/CMS/FastTimerService) is used as the basis for [circles](https://github.com/fwyzard/circles) (among other things).

The twiki page (linked above) describes how to use it, but requires manually editing configs generated by cmsDriver.

To make this easier to use, I put the code from the twiki into some functions that can be given to cmsDriver using `--customise`:
* `customise_timer_service`
* `customise_timer_service_singlejob`
* `customise_timer_service_multirun`
* `customise_timer_service_print`

I also added a simple harvesting config (which has to run standalone, so a customise function is not appropriate), which includes a `multirun` command-line argument.

@fwyzard if you agree with this approach, we can update the twiki to point out the availability of these convenience functions.

#### PR validation:

Tested the various customise functions and observed that they ran successfully and provided the expected input.

### DIFF
--- a/HLTrigger/Timer/python/FastTimer.py
+++ b/HLTrigger/Timer/python/FastTimer.py
@@ -1,0 +1,104 @@
+import FWCore.ParameterSet.Config as cms
+
+def addService(process, multirun=False):
+    # remove any instance of the FastTimerService
+    if 'FastTimerService' in process.__dict__:
+        del process.FastTimerService
+
+    # instrument the menu with the FastTimerService
+    process.load("HLTrigger.Timer.FastTimerService_cfi")
+
+    # print a text summary at the end of the job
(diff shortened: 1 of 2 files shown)


kpedro88:
please test

fwyzard:
@kpedro88
thanks, looks good.

I cannot actually test it this week, and I'm thinking it could be useful to have the default behaviour in a cff file as well - but I can add that next week.

fwyzard:
+1

kpedro88:
ping @cmsbuild

silviodonato:
+1

PR #28875, opened 2020-02-06
MTD reco geometry: preliminary cleaning

fabiocos:
#### PR description:

This PR proposes a preliminary cleaning of the ```Geometry/MTDNumberingBuilder``` and ```Geometry/MTDGeometryBuilder``` in view of the migration to DD4hep and insertion of the new TRD ETL. The DD4hep ```DDFilteredView``` is updated by adding a method ```copyNumbers``` that has the same meaning as in DDD.

The ```GeometricTimingDetAnalyzer``` has also been revisited in order to provide a detailed dump of the product of the MTDNumberingBuilder and serve for comparisons in the rewriting of this component.

#### PR validation:

The old tests run on scenario D50 without issues, and now the printout in ```MTDTopologyAnalyzer``` is fixed (previously partially not working).


fabiocos:
please test

Geometry/MTDNumberingBuilder/test/MTDTopologyAnalyzer.cc
```diff
@@ -90,20 +92,29 @@ void MTDTopologyAnalyzer::analyze(const edm::Event& iEvent, const edm::EventSetu
 
     size_t num = fv.geoHistory().size();
 
-    if (fv.geoHistory()[num - 1].logicalPart().name() == "BarrelTimingLayer") {
+    if (fv.geoHistory()[num - 1].logicalPart().name().name() == "BarrelTimingLayer") {
```
kpedro88: Is there a reason this interface requires calling `name()` twice? It seems suboptimal.

kpedro88:
+upgrade

Geometry/MTDNumberingBuilder/test/MTDTopologyAnalyzer.cc
fabiocos: @kpedro88 this is the old DDD syntax to strip the namespace. Its absence was causing the bulk of the test to be ineffective. In DD4hep this is handled in a different way (@ianna can better comment in case)

fabiocos:
please test

fabiocos:
the failure in the DT unit test is already present in the IB since the build following the previous test of this PR, and it appears unrelated to this PR (@slomeo FYI)

fabiocos:
FYI, unrelated issue in unit test understood, see https://github.com/cms-sw/cmssw/pull/28796/files#r377507493

fabiocos:
please test with #28921

fabiocos:
@cvuosalo @kpedro88 @silviodonato again the failure is unrelated to this PR, the fix by @slomeo is not doing what it should. Any comment on this PR? Or can it move forward?

kpedro88:
+upgrade

Geometry/MTDNumberingBuilder/test/MTDTopologyAnalyzer.cc
ianna: @kpedro88 and @fabiocos - yes, a ```DDLogicalPart``` has a ```name()``` which is a ```DDName```. The latter is a two strings object: a ```namespace()``` and a ```name()```. Actually, every DD object has a ```DDName```.

ianna (approved):

ianna:
+1

silviodonato:

silviodonato:
+1

PR #28882, opened 2020-02-07
Clean BuildFiles in RecoVertex and RecoTracker

guitargeek:
#### PR description:

#### PR validation:

CMSSW compiles.

No backport intended.


christopheralanwest:
please test

guitargeek:
Pardon, I missed the test with `git cms-checkdeps -a -A`. It should work now.

christopheralanwest:
please test

pohsun:
+1

RecoPixelVertexing/PixelTriplets/plugins/BuildFile.xml
```diff
@@ -1,4 +1,5 @@
 <use   name="CommonTools/RecoAlgos"/>
+<use   name="CondFormats/SiPixelObjects"/>
```
slava77: this appears unnecessary.
```git grep CondFormats/SiPixelObjects | grep  RecoPixelVertexing/PixelTriplets``` finds nothing

File: RecoTracker/DeDx/plugins/BuildFile.xml [hunk 2]
```diff
@@ -1,4 +1,6 @@
-<use   name="RecoTracker/DeDx"/>
 <library   file="*.cc" name="RecoTrackerDeDxPlugins">
+  <use   name="HLTrigger/HLTcore"/>
+  <use   name="PhysicsTools/UtilAlgos"/>
```
slava77: ```git grep PhysicsTools/UtilAlgos | grep  RecoTracker/DeDx/plugins ``` shows nothing.

guitargeek:
Thanks for your comment @slava77, I should have checked that. I fixed this in my last commit by adding the correct dependency for DeDx/plugins, which was `DataFormats/RecoCandidate`. Sorry for the force push, but I preferred rebasing it to make sure it still compiles (I didn't change the first two commits). Furthermore I applied the same procedure to RecoTBCalo so the latest commit isn't just a fixup.

guitargeek:
For all the dependencies I added in this PR now, the git grepping shows they are actually used.

slava77:
@cmsbuild please test

slava77:
+1

for https://github.com/cms-sw/cmssw/pull/28882	f6dba84 
- these technical buildfile changes are in line with the PR description and the follow up review

pohsun:
+1

CalibTracker/SiStripHitEfficiency/BuildFile.xml
```diff
@@ -24,5 +24,6 @@
 <use   name="TrackingTools/PatternTools"/>
 <use   name="TrackingTools/TrackFitters"/>
 <use   name="TrackingTools/MeasurementDet"/>
+<use   name="TrackingTools/KalmanUpdators"/>
```
silviodonato: For my education, why do you need to add TrackingTools/KalmanUpdators?

Alignment/OfflineValidation/plugins/BuildFile.xml
```diff
@@ -5,6 +5,7 @@
 <use   name="CondCore/DBOutputService"/>
 <use   name="CondFormats/Alignment"/>
 <use   name="CondFormats/RunInfo"/>
+<use   name="CondFormats/SiPixelObjects"/>
```
silviodonato: same comment here

CalibTracker/SiStripHitEfficiency/BuildFile.xml
guitargeek: Because it's included here and in the line after:
https://github.com/cms-sw/cmssw/blob/master/CalibTracker/SiStripHitEfficiency/interface/HitEff.h#L14
Before, it somehow got this dependency indirectly by one of the used RecoTracker packages, but since the dependencies got cleaned there this didn't happen anymore. That's why we have to declare the dependency at this right place now.

Alignment/OfflineValidation/plugins/BuildFile.xml
guitargeek: It is used here: https://github.com/cms-sw/cmssw/blob/master/Alignment/OfflineValidation/plugins/TrackerGeometryCompare.cc#L18

Again, this was not an issue before because it somehow got this dependency indirectly via some RecoTracker package, but now it's done right.

silviodonato:
+1

PR #28897, opened 2020-02-09
Converted GenHTFilter to a global module

Dr15Jones:
#### PR description:

The filter modules in GeneratorInterface/GenFilters need to be converted from legacy modules to thread-friendly ones in order for the GEN step to be CPU efficient. I converted this one as it was being used in an low CPU job example I was given.

#### PR validation:

The code compiles and with a small external test I was given the filter appears to work.


Dr15Jones:
please test

qliphy:
+1

silviodonato:
+1

PR #28899, opened 2020-02-09
Concurrent ExternalLHEProducer

Dr15Jones:
#### PR description:

Add the ability to use multiple TBB tasks to execute the script concurrently during the begin run phase.

#### PR validation:

The code compiles and the new unit tests succeed. In addition, I ran the code using an example job I had been given. If the new parameter is not used, the old code is used.

### DIFF
--- a/GeneratorInterface/LHEInterface/plugins/BuildFile.xml
+++ b/GeneratorInterface/LHEInterface/plugins/BuildFile.xml
@@ -7,6 +7,7 @@
 <use name="boost"/>
 <library name="GeneratorInterfaceLHEProducer" file="LHEFilter.cc LHE2HepMCConverter.cc ExternalLHEProducer.cc ExternalLHEAsciiDumper.cc">
         <use name="FWCore/Framework"/>
+        <use name="FWCore/Concurrency"/>
         <use name="SimDataFormats/GeneratorProducts"/>
         <flags EDM_PLUGIN="1"/>
 </library>
--- a/GeneratorInterface/LHEInterface/plugins/ExternalLHEProducer.cc
+++ b/GeneratorInterface/LHEInterface/plugins/ExternalLHEProducer.cc
@@ -22,12 +22,14 @@ Implementation:
 #include <vector>
 #include <string>
 #include <fstream>
+#include "boost/filesystem.hpp"
 #include <unistd.h>
 #include <dirent.h>
 #include <fcntl.h>
 #include <sys/wait.h>
 #include <sys/time.h>
 #include <sys/resource.h>
+#include "tbb/task_arena.h"
 
 #include "boost/bind.hpp"
 
@@ -41,6 +43,8 @@ Implementation:
 #include "FWCore/Framework/interface/Run.h"
 #include "FWCore/Framework/interface/Event.h"
 #include "FWCore/Framework/interface/MakerMacros.h"
+
+#include "FWCore/Concurrency/interface/FunctorTask.h"
 
 #include "FWCore/ParameterSet/interface/FileInPath.h"
 #include "FWCore/ParameterSet/interface/ParameterSet.h"
@@ -66,7 +70,6 @@ Implementation:
 class ExternalLHEProducer : public edm::one::EDProducer<edm::BeginRunProducer, edm::EndRunProducer> {
 public:
   explicit ExternalLHEProducer(const edm::ParameterSet& iConfig);
-  ~ExternalLHEProducer() override;
 
   static void fillDescriptions(edm::ConfigurationDescriptions& descriptions);
 
@@ -76,30 +79,31 @@ private:
   void endRunProduce(edm::Run&, edm::EventSetup const&) override;
   void preallocThreads(unsigned int) override;
 
-  int closeDescriptors(int preserve);
-  void executeScript();
-  std::unique_ptr<std::string> readOutput();
+  std::vector<std::string> makeArgs(uint32_t nEvents, unsigned int nThreads, std::uint32_t seed) const;
+  int closeDescriptors(int preserve) const;
+  void executeScript(std::vector<std::string> const& args, int id) const;
 
   void nextEvent();
 
   // ----------member data ---------------------------
   std::string scriptName_;
   std::string outputFile_;
-  std::vector<std::string> args_;
+  const std::vector<std::string> args_;
   uint32_t npars_;
   uint32_t nEvents_;
   bool storeXML_;
   unsigned int nThreads_{1};
   std::string outputContents_;
+  bool generateConcurrently_{false};
 
   // Used only if nPartonMapping is in the configuration
   std::map<unsigned, std::pair<unsigned, unsigned>> nPartonMapping_{};
 
   std::unique_ptr<lhef::LHEReader> reader_;
-  std::shared_ptr<lhef::LHERunInfo> runInfoLast;
-  std::shared_ptr<lhef::LHERunInfo> runInfo;
-  std::shared_ptr<lhef::LHEEvent> partonLevel;
-  boost::ptr_deque<LHERunInfoProduct> runInfoProducts;
+  std::shared_ptr<lhef::LHERunInfo> runInfoLast_;
+  std::shared_ptr<lhef::LHERunInfo> runInfo_;
+  std::shared_ptr<lhef::LHEEvent> partonLevel_;
+  boost::ptr_deque<LHERunInfoProduct> runInfoProducts_;
   bool wasMerged;
 
   class FileCloseSentry : private boost::noncopyable {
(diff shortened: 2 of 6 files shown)


Dr15Jones:
please test

qliphy:
@Dr15Jones   Thanks. Looks interesting!  Can you give an example how to test this?  Will the LHE produced in each thread be merged at last?

intrepid42:
Hi, I tried to produce a couple of Powheg ttbar events, it seems to work :)

I am just wondering about the use of consecutive random seeds. These would overlap when consecutive job numbers are used as seed. How is it done in production? Consecutive or "random" random seeds?

@qliphy Yes, the gridpack is unpacked and used multiple times in directories `thread0`, `thread1`, etc.

I am using this fragment:
```
process.externalLHEProducer = cms.EDProducer("ExternalLHEProducer",
    args = cms.vstring('/cvmfs/cms.cern.ch/phys_generator/gridpacks/slc6_amd64_gcc481/13TeV/powheg/V2/TT_hdamp_Tune4_NNPDF30/TT_hdamp_Tune4_NNPDF30_13TeV_powheg_hvq.tgz'),
    nEvents = cms.untracked.uint32(process.maxEvents.input.value()),
    numberOfParameters = cms.uint32(1),
    outputFile = cms.string('cmsgrid_final.lhe'),
    scriptName = cms.FileInPath('GeneratorInterface/LHEInterface/data/run_generic_tarball_cvmfs.sh'),
    generateConcurrently = cms.untracked.bool(True)
)
```

Dr15Jones:
@intrepid42 already answered these but just for completeness.

> Can you give an example how to test this?

You can take any existing configuration and add `generateConcurrently = cms.untracked.bool(True)` to the `ExternalLHEProducer` configuration.

> Will the LHE produced in each thread be merged at last?

Yes

> Current gridpacks (MadGraph, Powheg...) doesn't support multi-thread yet, so does this mean it needs to copy and tar/untar gridpacks multi times?

Yes

Now as for the random numbers

>I am just wondering about the use of consecutive random seeds. These would overlap when consecutive job numbers are used as seed. How is it done in production? Consecutive or "random" random seeds?

The framework already makes use of consecutive random numbers to handle random number assignments for threading. The product system assigns a single random number seed to each module in a configuration. Then for each stream in a job, the random number seed is modifed by adding the stream index (a number from 0 to #streams -1).

qliphy:
Thanks @Dr15Jones @intrepid42 
I have done a test based on [SMP-RunIISummer19UL17wmLHEGEN-00001](https://cms-pdmv.cern.ch/mcm/public/restapi/requests/get_test/SMP-RunIISummer19UL17wmLHEGEN-00001)

which is a complicated case, i.e. WJetsToLNu_TuneCP5_13TeV-madgraph261MLM-pythia8  (W+up to 4 jets).

(1).   

cmsrel CMSSW_11_1_X_2020-02-12-2300
cd CMSSW_11_1_X_2020-02-12-2300/src
git cms-addpkg GeneratorInterface/LHEInterface
git cms-merge-topic Dr15Jones:concurrentExternalLHEProducer
scram b
curl -s --insecure https://cms-pdmv.cern.ch/mcm/public/restapi/requests/get_fragment/SMP-RunIISummer19UL17wmLHEGEN-00001 --retry 2 --create-dirs -o Configuration/GenProduction/python/SMP-RunIISummer19UL17wmLHEGEN-00001-fragment.py 
scram b

cmsDriver.py Configuration/GenProduction/python/SMP-RunIISummer19UL17wmLHEGEN-00001-fragment.py --fileout file:SMP-RunIISummer19UL17wmLHEGEN-00001.root --mc --eventcontent RAWSIM,LHE --datatier GEN,LHE --conditions 106X_mc2017_realistic_v6 --beamspot Realistic25ns13TeVEarly2017Collision --step LHE,GEN --geometry DB:Extended --era Run2_2017 --python_filename SMP-RunIISummer19UL17wmLHEGEN-00001_1_cfg.py --no_exec --customise Configuration/DataProcessing/Utils.addMonitoring --customise_commands 
process.source.numberEventsInLuminosityBlock="cms.untracked.uint32(2222)" -n 200


(2)  same as (1), 
but add "generateConcurrently = cms.untracked.bool(True)"  
and in addition 
process.options.numberOfThreads=cms.untracked.uint32(4)
process.options.numberOfStreams=cms.untracked.uint32(0)
process.options.numberOfConcurrentLuminosityBlocks=cms.untracked.uint32(1)

I indeed see 4 thread running, and each with a gridpack untarring and running.
However, from the output below, it seems the concurrent even takes longer time to get finished. And it seems initialization of gridpack takes quite long.

Probably this is  due to this example is with very complicated physics process.  I will have a look with simple case.


**The output of (1):**
TimeReport> Time report complete in 1141.85 seconds
 Time Summary: 
 - Min event:   0.039495
 - Max event:   6.11634
 - Avg event:   1.53845
 - Total loop:  1128.07
 - Total init:  12.4372
 - Total job:   1141.85
 - EventSetup Lock:   5.00679e-06
 - EventSetup Get:   0.00770998
 Event Throughput: 0.177295 ev/s
 CPU Summary: 
 - Total loop:  518.395
 - Total init:  4.91161
 - Total extra: 0
 - Total job:   523.881
 Processing Summary: 
 - Number of Events:  200
 - Number of Global Begin Lumi Calls:  1
 - Number of Global Begin Run Calls: 1

**The output of (2)**
TimeReport> Time report complete in 1788.17 seconds
 Time Summary: 
 - Min event:   0.179609
 - Max event:   17.4009
 - Avg event:   6.54978
 - Total loop:  1778.47
 - Total init:  8.827
 - Total job:   1788.17
 - EventSetup Lock:   6.19888e-06
 - EventSetup Get:   0.022414
 Event Throughput: 0.112456 ev/s
 CPU Summary: 
 - Total loop:  1383.01
 - Total init:  4.88403
 - Total extra: 0
 - Total job:   1388.48
 Processing Summary: 
 - Number of Events:  200
 - Number of Global Begin Lumi Calls:  1
 - Number of Global Begin Run Calls: 1

Dr15Jones:
Thanks for doing the study! If we look just at the single threaded case we see that the CPU efficiency is poor

	Wallclock Total loop: 1128.07
	CPU Total loop: 518.395

	 518.395/1128.07 = 0.46

I was able to replicate what you ran and by watching the job with 'top' I was able to see that the generation step is incredibly CPU inefficient. It does lots of work with 'tar' and even when 'madevent' is running, that application is also very CPU inefficient. So given that the single threaded job is so IO bound, it stands to reason that running 4 IO bound jobs (which is what the changes to externalLHEProducer replicates) would be slower since the 4 applications are fighting each other for the disk resources. There is one thing to keep in mind, the workflow management system will not just run 1 single threaded GEN job on a node as a batch slot typically has 8 cores assigned to it. Instead, the workflow system is likely to assign 8 single threaded GEN jobs to a node. Therefore a better test would be to compare the 4 threaded case to the time it takes to run 4 single threaded jobs on the node.

As a test, I ran the configuration you created under 3 different conditions

1) just the way it was creates using just 1 thread (only 1 job, not 4 of them)
2) setting generateConcurrently to true and running with 4 threads
3) including the Pythia8ConcurrentHadronizerFilter from my other pull request and running 4 threads

The results were as follows.

Single threaded 
Wallclock Total loop:  3134.77
CPU Total loop:  2826.53
	efficiency 2826.53/3134.77 = 0.90

4 threads with only ExternalLHEProducer
Wallclock Total loop:  3295.3
CPU Total loop:  5828.19
	efficiency  5828.19/3295.3/4 = 0.44

4 threads with ExternalLHEProducer and Pythia8ConcurrentHadronizerFiter
Wallclock Total loop:  1699.53
CPU Total loop:  4559.94
	efficiency 4559.94/1699.53/4 = 0.67

So using all the changes did process the code 3134.77/1699.53 = 1.8 times faster than the single threaded case.

Dr15Jones:
As a test, I ran 4 single-threaded jobs concurrently. The timing was
Wallclock Total loop:  3610.61
CPU Total loop:  3006.08

Wallclock Total loop:  3737.87
CPU Total loop:  3186.13

Wallclock Total loop:  3663.9
CPU Total loop:  3130.87

Wallclock Total loop:  3780.75
CPU Total loop:  3182.42

So much slower than the case of just 1 job. Now this did process 4x as many events as the previous test but does show that multiple IO bound jobs to interfere with one another.

qliphy:
Thanks @Dr15Jones    It makes sense.


One question,

You got two single thread numbers:

(1)
Wallclock Total loop: 1128.07
CPU Total loop: 518.395
 518.395/1128.07 = 0.46

(2)
Single threaded
Wallclock Total loop: 3134.77
CPU Total loop: 2826.53
efficiency 2826.53/3134.77 = 0.90

The efficiency differ a lot.  Why?

Dr15Jones:
> The efficiency differ a lot. Why?

So we each ran them on different machines. It looks like one machine has a much faster CPU than the other. If the IO systems on the two machines were about the same then the slower CPU machine would have spent a proportionally less time doing IO than the fast machine. That would mean the faster machine would have a proportionally worse CPU utilization.

Dr15Jones:
What further checks/code do you want in order to finish this pull request?

qliphy:
This PR looks good to me.  Probably we can check with not only GENonly but also GEN+SIM, but that can be done later.

qliphy:
+1

silviodonato:
+1

silviodonato:
+1

bendavid:
@silviodonato I guess you meant to +1 to https://github.com/cms-sw/cmssw/pull/30481 instead?

PR #28903, opened 2020-02-10
[HGCAL] TICL technical changes

rovere:
#### PR description:

This PR is another step toward the inclusion of `TICL` into the `Phase2` reconstruction sequence. This PR, in particular, fixes a few bugs and add templated data structures that will be used in future internally by `TICL` and, possibly, other detectors (`HFNose`, in particular). In more details:

- Correct the cone direction while running the `Pattern Recognition`
- Fix the `phi` search window
- Fix the charge assignment to `TICL Candidates`
- Fix energy regression input variables
- Template `TICL Tile` data structures. 

#### PR validation:

The usual limited Matrix.

### DIFF
--- a/RecoHGCal/TICL/plugins/HGCGraph.cc
+++ b/RecoHGCal/TICL/plugins/HGCGraph.cc
@@ -54,18 +54,18 @@ void HGCGraph::makeAndConnectDoublets(const TICLLayerTiles &histo,
         auto const &outerLayerHisto = histo[currentOuterLayerId];
         auto const &innerLayerHisto = histo[currentInnerLayerId];
 
-        for (int oeta = startEtaBin; oeta < endEtaBin; ++oeta) {
-          auto offset = oeta * nPhiBins;
-          for (int ophi_it = startPhiBin; ophi_it < endPhiBin; ++ophi_it) {
-            int ophi = ((ophi_it % nPhiBins + nPhiBins) % nPhiBins);
-            for (auto outerClusterId : outerLayerHisto[offset + ophi]) {
+        for (int ieta = startEtaBin; ieta < endEtaBin; ++ieta) {
+          auto offset = ieta * nPhiBins;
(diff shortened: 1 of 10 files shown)


kpedro88:
please test

kpedro88:
+upgrade

RecoHGCal/TICL/plugins/HGCGraph.cc
```diff
@@ -54,29 +54,29 @@ void HGCGraph::makeAndConnectDoublets(const TICLLayerTiles &histo,
         auto const &outerLayerHisto = histo[currentOuterLayerId];
         auto const &innerLayerHisto = histo[currentInnerLayerId];
 
-        for (int oeta = startEtaBin; oeta < endEtaBin; ++oeta) {
-          auto offset = oeta * nPhiBins;
-          for (int ophi_it = startPhiBin; ophi_it < endPhiBin; ++ophi_it) {
-            int ophi = ((ophi_it % nPhiBins + nPhiBins) % nPhiBins);
-            for (auto outerClusterId : outerLayerHisto[offset + ophi]) {
+        for (int ieta = startEtaBin; ieta < endEtaBin; ++ieta) {
```
perrotta: Is this inversion of the order of looping through innerLayer and outerLayer due to some performance improvement expected?

rovere: Ciao @perrotta 
this change will orient the search cone from inner to outer, rather than the opposite (which was wrong).

perrotta:
+1
- Code changes are in line with the PR description
- Jenkins tests pass with no difference (no TICL reco in them)
- Verified running in a wf which includes TICL:  20493.52
   - Changes in objects follow what implemented (eg. charge of the TICL candidate)
   - A few more candidates (241 vs 239) selected in a test run with that wf, mostly, with apparently no changes in the other parameters (energy, eta, etc.) of the candidates in common)

silviodonato:
+1

Issue #28772, opened 2020-01-21
Memory problem in HGCalCLUEAlgoT


kpedro88:
The address sanitizer build (CMSSW_11_1_ASAN_X_2020-01-17-2300) reports an issue for workflows 21234.0 and 22434.0:

```
==12521==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x60c000a2da40 at pc 0x2ae0cd38df66 bp 0x2ae0c0e5a640 sp 0x2ae0c0e5a638
READ of size 8 at 0x60c000a2da40 thread T4
    #0 0x2ae0cd38df65 in HGCalCLUEAlgoT<HGCalLayerTilesT<HFNoseTilesConstants> >::computeThreshold() (/cvmfs/cms-ib.cern.ch/nweek-02611/slc7_amd64_gcc820/cms/cmssw/CMSSW_11_1_ASAN_X_2020-01-17-2300/lib/slc7_amd64_gcc820/pluginRecoLocalCaloHGCalRecProducersPlugins.so+0x107f65)
    #1 0x2ae0cd3945bf in HGCalCLUEAlgoT<HGCalLayerTilesT<HFNoseTilesConstants> >::populate(edm::SortedCollection<HGCRecHit, edm::StrictWeakOrdering<HGCRecHit> > const&) (/cvmfs/cms-ib.cern.ch/nweek-02611/slc7_amd64_gcc820/cms/cmssw/CMSSW_11_1_ASAN_X_2020-01-17-2300/lib/slc7_amd64_gcc820/pluginRecoLocalCaloHGCalRecProducersPlugins.so+0x10e5bf)
    #2 0x2ae0cd3a8f2c in HGCalLayerClusterProducer::produce(edm::Event&, edm::EventSetup const&) (/cvmfs/cms-ib.cern.ch/nweek-02611/slc7_amd64_gcc820/cms/cmssw/CMSSW_11_1_ASAN_X_2020-01-17-2300/lib/slc7_amd64_gcc820/pluginRecoLocalCaloHGCalRecProducersPlugins.so+0x122f2c)
[...]

PR #28917, opened 2020-02-11
HFNose: fix rechitTools for HGCalCLUEAlgo

mariadalfonso:
This should solve the memory problem introduced with HFNose 2D clusters
https://github.com/cms-sw/cmssw/issues/28772

+ add the HF to the list of simClusters (needed since the HFNose will not contains the full hadron shower)

### DIFF
--- a/DataFormats/ForwardDetId/interface/HFNoseDetId.h
+++ b/DataFormats/ForwardDetId/interface/HFNoseDetId.h
@@ -26,7 +26,7 @@ public:
   static const int HFNoseCoarseN = 8;
   static const int HFNoseFineTrigger = 3;
   static const int HFNoseCoarseTrigger = 2;
-  static const int kHFNoseLayerEEmax = 6;
+  static const int HFNoseLayerEEmax = 6;
 
   /** Create a null cellid*/
   HFNoseDetId();
@@ -101,8 +101,8 @@ public:
   std::pair<int, int> triggerCellUV() const { return std::pair<int, int>(triggerCellU(), triggerCellV()); }
 
   /// consistency check : no bits left => no overhead
-  bool isEE() const { return (layer() <= kHFNoseLayerEEmax); }
-  bool isHE() const { return (layer() > kHFNoseLayerEEmax); }
+  bool isEE() const { return (layer() <= HFNoseLayerEEmax); }
+  bool isHE() const { return (layer() > HFNoseLayerEEmax); }
   bool isForward() const { return true; }
 
   static const HFNoseDetId Undefined;
--- a/DataFormats/ForwardDetId/interface/HFNoseDetId.h
+++ b/DataFormats/ForwardDetId/interface/HFNoseDetId.h
@@ -26,6 +26,7 @@ public:
   static const int HFNoseCoarseN = 8;
   static const int HFNoseFineTrigger = 3;
   static const int HFNoseCoarseTrigger = 2;
+  static const int kHFNoseLayerEEmax = 6;
 
   /** Create a null cellid*/
   HFNoseDetId();
@@ -107,7 +108,6 @@ public:
   static const HFNoseDetId Undefined;
 
 private:
-  static const int kHFNoseLayerEEmax = 6;
   static const int kHFNoseCellUOffset = 0;
   static const int kHFNoseCellUMask = 0x1F;
   static const int kHFNoseCellVOffset = 5;
--- a/RecoLocalCalo/HGCalRecProducers/plugins/HGCalCLUEAlgo.cc
+++ b/RecoLocalCalo/HGCalRecProducers/plugins/HGCalCLUEAlgo.cc
@@ -518,7 +518,7 @@ void HGCalCLUEAlgoT<T>::computeThreshold() {
 
   std::vector<double> dummy;
   const unsigned maxNumberOfThickIndices = 3;
-  dummy.resize(maxNumberOfThickIndices + isNose_, 0);  // +1 to accomodate for the Scintillators
+  dummy.resize(maxNumberOfThickIndices + !isNose_, 0);  // +1 to accomodate for the Scintillators
   thresholds_.resize(maxlayer_, dummy);
   v_sigmaNoise_.resize(maxlayer_, dummy);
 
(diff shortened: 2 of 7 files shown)


perrotta:
please test

DataFormats/ForwardDetId/interface/HFNoseDetId.h
```diff
@@ -26,6 +26,7 @@ class HFNoseDetId : public DetId {
   static const int HFNoseCoarseN = 8;
   static const int HFNoseFineTrigger = 3;
   static const int HFNoseCoarseTrigger = 2;
+  static const int kHFNoseLayerEEmax = 6;
```
kpedro88: the other public consts don't start with `k`, change this one to fit the style?

RecoLocalCalo/HGCalRecProducers/plugins/HGCalCLUEAlgo.cc
```diff
@@ -518,7 +518,10 @@ void HGCalCLUEAlgoT<T>::computeThreshold() {
 
   std::vector<double> dummy;
   const unsigned maxNumberOfThickIndices = 3;
-  dummy.resize(maxNumberOfThickIndices + 1, 0);  // +1 to accomodate for the Scintillators
+  if (isNose_)
```
kpedro88: simplify to: `dummy.resize(maxNumberOfThickIndices + isNose_, 0);`

SimCalorimetry/HGCalSimProducers/python/hgcalDigitizer_cfi.py
```diff
@@ -273,3 +273,8 @@ def HGCal_disableNoise(process):
 )
 phase2_hgcalV9.toModify(HGCAL_noise_fC, values = [x*fC_per_ele for x in nonAgedNoises_v9])
 phase2_hgcalV9.toModify(HGCAL_noises, values = [x for x in nonAgedNoises_v9])
+
+from Configuration.Eras.Modifier_phase2_hfnose_cff import phase2_hfnose
```
kpedro88: The `phase2_hfnose` modifier is not just used with the v9 HGCal, so this is not correct. It should also be unnecessary, since when it is used with the v9 HGCal, the `phase2_hgcalV9` modifications will already be picked up.

If it is necessary for some reason I'm missing, then rather than copying the modifications, just change the lines above this one to: `(phase2_hgcalV9 | phase2_hfnose).toModify(...)`

mariadalfonso: ok I agree the pset is already replaced for both HGC and HFnose, so it's unnecessary at this stage. I removed the commit

RecoLocalCalo/HGCalRecProducers/plugins/HGCalCLUEAlgo.cc
mariadalfonso: done

DataFormats/ForwardDetId/interface/HFNoseDetId.h
mariadalfonso: done

perrotta:
please test

perrotta:
please test

kpedro88:
In the short matrix workflow for the HF nose (21234.0, D44 geometry), there is an increase in clusters with energy = 0, eta = 0, phi = 0, and time = -100. Is this expected?

cluster energy
cluster eta
cluster phi
cluster time

slava77:
> In the short matrix workflow for the HF nose (21234.0, D44 geometry), there is an increase in clusters with energy = 0, eta = 0, phi = 0, and time = -100. Is this expected?

even if it's expected, I'd say it's  not really wanted.
Why are we wasting memory and disk space on this?

mariadalfonso:
> > In the short matrix workflow for the HF nose (21234.0, D44 geometry), there is an increase in clusters with energy = 0, eta = 0, phi = 0, and time = -100. Is this expected?
> 

I did not expect any change in the HGCAL,  only in the HFnose, I will investigate.

> even if it's expected, I'd say it's not really wanted.
> Why are we wasting memory and disk space on this?

do you know if the 0 in the particleFlowClusterHGCAL appear in both the 21234.0 (D44 w/ HFNose) and 20434.0 (D41 w/o HFnose) ?

kpedro88:
Based on https://cmssdt.cern.ch/SDT/jenkins-artifacts/baseLineComparisons/CMSSW_11_1_X_2020-02-12-2300+bb1ad4/35063/validateJR.html, the comparison changes only show up in WF 21234.0.

mariadalfonso:
After a bit of investigation of the zeros,
in the HFnose scenarios the simClusters (w/ hFNose + HF) are not yet mapped any particleFlowRecHitHFNose
https://github.com/cms-sw/cmssw/blob/2ba5d421e10379d81760a899532b2c991b89c82c/RecoParticleFlow/PFClusterProducer/python/particleFlowClusterHGC_cfi.py#L54

Here I show how the zeros are indeed reduce after last commit (from the red to the blu line).
http://dalfonso.web.cern.ch/dalfonso/particleFlowHGCal.png
However I compared with D49 w/o HFNose scenario and some residual zero is still there.

For now I comments the 2D sim clusters part: and re-activate at later stage when the full reconstruction will be in place for the HFnose as well.
For now at analysis level we can just uncomment those sim caloparticle.

kpedro88:
please test

kpedro88:
A lot of the 0 clusters are gone now:
cluster eta

kpedro88:
+upgrade

slava77:
> For now I comments the 2D sim clusters part: and re-activate at later stage when the full reconstruction will be in place for the HFnose as well.

It's a bit of an odd approach to break the "sim" (DIGI, formally) side to address the issue in making of reco objects. However, it is apparently acceptable.

slava77:
+1

for https://github.com/cms-sw/cmssw/pull/28917	50b7588 
- the update in reco is visually in the part relevant to fix the issue in #28772 [IIUC, this does not lead to any differences by itself]. Changes come from modifications in the digi step, in SimGeneral/MixingModule/python/caloTruthProducer_cfi.py
- jenkins tests pass and comparisons with the baseline show changes only in the workflow with HFNose enabled and only in the [reduction of] "zero energy" hits.

civanch:
+1

silviodonato:
please test workflow 23434.1001, 23234.1001

silviodonato:
+1

PR #27928, opened 2019-09-04
Add an LHE-based mc relval with multiple lumi sections and concurrent lumi settings

fabiocos:
#### PR description:

This PR adds an LHE-based simulation relval workflow where 2 lumi sections are produced, and the concurrent luminosity block settings are activated. This is a first attempt to address the request by @Dr15Jones discussed in #26348 , and test the lumi section boundary crossing that is usual in production and that sometimes shows problems (see for instance #25708 ).

In order to test the boundary crossing from LHE_based input onwards, the workflow 1361.18 is modified by forcing the EmptySource to have 5 events per lumi section, and therefore forcing the standard 10 events sample to contain 2 lumi sections.

#### PR validation:

If ```--nThreads 4``` the new workflow 1361.181 has the new correct settings, for instance in step2 (DIGI):

```
%MSG-i ThreadStreamSetup:  (NoModuleName) 04-Sep-2019 16:12:47 CEST pre-events
setting # threads 4
setting # streams 4
%MSG
04-Sep-2019 16:12:53 CEST  Initiating request to open file file:step1.root
04-Sep-2019 16:12:54 CEST  Successfully opened file file:step1.root
%MSG-s ModulesSynchingOnLumis:  AfterModConstruction 04-Sep-2019 16:12:59 CEST  pre-events
The following modules require synchronizing on LuminosityBlock boundaries:
  HLTrigReport hltTrigReport
%MSG
Begin processing the 1st record. Run 1, Event 2, LumiSection 1 on stream 3 at 04-Sep-2019 16:13:05.814 CEST
Begin processing the 2nd record. Run 1, Event 3, LumiSection 1 on stream 0 at 04-Sep-2019 16:13:05.814 CEST
Begin processing the 3rd record. Run 1, Event 4, LumiSection 1 on stream 2 at 04-Sep-2019 16:13:05.814 CEST
Begin processing the 4th record. Run 1, Event 5, LumiSection 1 on stream 1 at 04-Sep-2019 16:13:05.814 CEST
DTTSPhi::addTracoT: preview not present in TRACO trigger or its code=0  trigger not added!
Begin processing the 5th record. Run 1, Event 1, LumiSection 1 on stream 1 at 04-Sep-2019 16:14:14.785 CEST
Begin processing the 6th record. Run 1, Event 9, LumiSection 2 on stream 3 at 04-Sep-2019 16:14:16.953 CEST
[...]

PR #28920, opened 2020-02-11
DQM: Disable assertLegacySafe when concurrent lumis are enabled.

schneiml:
#### PR description:

This prevents the crashes in 1361.181 reported in #28622. Prevents, not fixes.

I'd like to summon @Dr15Jones and @makortel to this discussion: What happens here is that a job with VALIDATION enabled (which I am pretty sure does contain some `edm::EDAnalyzer`s -- I have not checked it though) is requested to run with concurrent lumisections. EDM should prevent this, since it is not safe to have concurrent lumisections with legacy modules [1]. However, the new `DQMStore` still detects that a new lumisection starts before the previous one is saved, and that consequently it needs to copy MEs (this triggers the `assertLegacySafe` assertion by default, unless it is explicitly turned off). The problem with that is now that `edm::EDAnalyzer` based DQM code could hold pointers that get `free`'d by the `DQMStore` later (as the first lumisection ends). For this reason, it is only safe to disable `assertLegacySafe` when there are no `edm::EDAnalyzer` based `DQMStore` users in the process.

So, with this PR, 1361.181 runs but is unsafe. What we should do instead is either make sure that EDM actually does not use concurrent lumisections at all [2] when there are legacy modules (then we can keep the assertion enabled and it will not fire), or remove all legacy modules from the jobs using concurrent lumis (that is really what we need to do, but much harder than it sounds).

We can of course also just keep the unsafe behaviour. It seems to work for now (maybe because the legacy modules involved don't do anything dangerous [3]), but I can't make any guarantees about that.

[1] actually, the majority of DQM currently runs in `edm::one::EDProducer`s watching lumis which should cause the same effect.
[2] I *think* EDM might overlap writing the current lumi and processing the next even when there are modules blocking concurrent lumis -- that would explain the behaviour.
[3] (Edit:) Since this entire story is about lumis, but legacy modules by default deal with per-job MEs, we should actually be safe as long as the legacy modules don't explicitly use `Scope::LUMI`. But then the (meaningful) legacy plugins *do* actually set a scope different from `JOB` manually (else they would not produce any output in the reco step), and *technically* the same problem exists with `RUN` MEs. Except, we don't do anything close to concurrent runs currently. But, we have reasons to believe that there is no risk of use-after-free in this workflow today.

#### PR validation:
```
runTheMatrix.py  -l 1361.181  -i all --job-reports -t 4
```
passes. Note the `-t 4`, bare `1361.181` did not trigger the problem.

See concerns above.

### DIFF
--- a/Configuration/Applications/python/ConfigBuilder.py
+++ b/Configuration/Applications/python/ConfigBuilder.py
@@ -2232,6 +2232,8 @@ class ConfigBuilder(object):
             self.pythonCfgCode +="process.options.numberOfThreads=cms.untracked.uint32("+self._options.nThreads+")\n"
             self.pythonCfgCode +="process.options.numberOfStreams=cms.untracked.uint32("+self._options.nStreams+")\n"
             self.pythonCfgCode +="process.options.numberOfConcurrentLuminosityBlocks=cms.untracked.uint32("+self._options.nConcurrentLumis+")\n"
+            if self._options.nConcurrentLumis > 1:
+              self.pythonCfgCode +="if process.DQMStore: process.DQMStore.assertLegacySafe=cms.untracked.bool(False)\n"
             self.process.options.numberOfThreads=cms.untracked.uint32(int(self._options.nThreads))
             self.process.options.numberOfStreams=cms.untracked.uint32(int(self._options.nStreams))
             self.process.options.numberOfConcurrentLuminosityBlocks=cms.untracked.uint32(int(self._options.nConcurrentLumis))


schneiml:
please test 1361.181

silviodonato:
please test workflow 1361.181

Dr15Jones:
> What happens here is that a job with VALIDATION enabled (which I am pretty sure does contain some edm::EDAnalyzers -- I have not checked it though) is requested to run with concurrent lumisections. EDM should prevent this, since it is not safe to have concurrent lumisections with legacy modules

@schneiml I'm afraid you have the wrong mental model for how this work. Concurrent LumiBlocks are not prevented, instead the framework still starts the new LuminosityBlock before the last ends while still guaranteeing that such legacy and edm::one modules will see the previous end LuminosityBlock call before the new begin LuminosityBlock call. So it isn't that the legacy prevent concurrent lumis, it is more they prevent processing the events in the next lumi before all events in the previous lumi finish. Sorry we gave you the wrong impression for what is happening.

schneiml:
@Dr15Jones yes, I pretty much inferred that this is what's going on, and it is basically fine. Except for the problem explained above: The new `DQMStore` tries to not allocate new ME objects as long as possible, using only a single instance of each ME, which allows the legacy code holding pointers to (global) MEs to work all fine (even if they are filled at the weirdest times). This obviously does not work as soon a new lumisection starts before the last one is saved, and once we have more than a single copy of an ME, the "all legacy interactions keep working" guarantee no longer holds. 

So as written above, we now have three options:
- Change these workflows to absolutely not use concurrent lumis.
- or Clean up and remove the legacy modules from these workflows.
- or Merge this PR and accept the risk.

I'd prefer the first or second, but the third is obviously much easier.

silviodonato:
urgent

silviodonato:
> What we should do instead is either make sure that EDM actually does not use concurrent lumisections at all [2] when there are legacy modules (then we can keep the assertion enabled and it will not fire)

@schneiml Can you force to use `--nConcurrentLumis=1` if the sequence contains legacy modules? I would prefer to temporary *disable* a feature instead of keeping it with the risk of having errors or buggy results

schneiml:
@silviodonato this may be possible, but I am not really sure how to do that. The heuristic could be that *any* `EDAnalyzer` on the DQM/VALIDATION sequences is probably a legacy module (at configuration stage, it is not possible to tell apart legacy `edm::EDAnalyzer` and non-legacy `edm::*::EDAnalyzer`, as far as I know, but we do know that all non-legacy DQM modules are based on `EDProducer` now).

But that still requires traversing the expanded sequence (which is much harder than I hoped, and lead me to build this contraption a few days ago: https://github.com/cms-sw/cmssw/compare/master...schneiml:dqm-dqmdumpsequence ), and then propagating the information to the right place in `ConfigBuilder`... I will take me a while to get that to work.

schneiml:
The cheap option would be to manually disable concurrent lumis in that one workflow that has DQM/VALIDATION in... Like it actually is when `runTheMatrix` is used without the `-t 4`. The concurrent lumis are only enabled in step5, where they actually work. Is that maybe what we actually want and having them in step3 a bug?

silviodonato:
Let's cc  @chayanit @pgunnell @zhenhu

schneiml:
Also, this behavior is not actually tested in the PR tests -- they are run without the `-t 4`!
https://cmssdt.cern.ch/SDT/jenkins-artifacts/pull-request-integration/PR-ffb461/4598/runTheMatrix-results/1361.181_GluGluHToZZTo4L_M125_Pow_py8_Evt_13UP18ml+GluGluHToZZTo4L_M125_Pow_py8_Evt_13UP18ml+DIGIUP18ml+RECOUP18ml+HARVESTUP18+NANOUP18ml/step3_RAW2DIGI_L1Reco_RECO_RECOSIM_EI_PAT_VALIDATION_DQM.py

A random other IB test that passes uses threads/streams, but no concurrent lumis: https://cmssdt.cern.ch/SDT/cgi-bin/logreader/slc7_amd64_gcc820/CMSSW_11_1_X_2020-02-11-1100/pyRelValMatrixLogs/run/136.722_RunDoubleEG2016B+RunDoubleEG2016B+HLTDR2_2016+RECODR2_2016reHLT_skimDoubleEG_HIPM+HARVESTDR2/step3_RunDoubleEG2016B+RunDoubleEG2016B+HLTDR2_2016+RECODR2_2016reHLT_skimDoubleEG_HIPM+HARVESTDR2.log#/

pgunnell:
The change looks fine to me, but does anybody know why only 1361.181 uses the concurrent lumis?

@chayanit

schneiml:
Concurrent lumis seem to be turned on deliberatly in this WF: https://github.com/cms-sw/cmssw/blob/master/Configuration/PyReleaseValidation/python/relval_steps.py#L2167
And this step is only used in 1361.181. 

`nThreads` overridden in `NANOUP18` (note the lack of `ml`/concurrent lumis), which then gets extended to `NANOUP18ml` with concurrent lumis, used only in this wf. This means in the end that `step5`/NANOAOD always runs with multiple threads and concurrent lumis, while RECO only uses concurrent lumis and mutliple threads when that is passed as an additional option to`runTheMatrix`. Note that `ConfigBuilder` does not turn on concurrent lumis if there are not also multiple threads.

silviodonato:
@schneiml do we need to disable concurrentLumis in all steps
NANOUP18ml
DIGIUP18ml
HLTDR2_2018ml
RECOUP18ml
RECODR2_2018reHLT_skimEGamma_Prompt_L1TEgDQMml
? Is it sufficient to disable it in NANOUP18ml ?

schneiml:
@silviodonato no, NANOUP18ml is fine, the one causing trouble is `RECOUP18ml` (I think, have not actually tried it)

silviodonato:
Ok, thanks. So the problem should be only in 1361.181.
These are the workflows using at least one `ml` step
```
136.8861 RunEGamma2018Dml1+HLTDR2_2018ml+RECODR2_2018reHLT_skimEGamma_Prompt_L1TEgDQMml+HARVEST2018_L1TEgDQM_Prompt 
136.8862 RunEGamma2018Dml2+HLTDR2_2018ml+RECODR2_2018reHLT_skimEGamma_Prompt_L1TEgDQMml+HARVEST2018_L1TEgDQM_Prompt 
1361.181 GluGluHToZZTo4L_M125_Pow_py8_Evt_13UP18ml+DIGIUP18ml+RECOUP18ml+HARVESTUP18+NANOUP18ml 
```

silviodonato:
@schneiml @pgunnell @chayanit 
I see three prossibilitie. We can temporary 
1. disable 1361.181
2. move 1361.181 from NANOUP18ml to NANOUP18
3. disable concurrent lumis in NANOUP18ml.

Any preference?

silviodonato:
wf 136.8861 and 136.8862 use concurrentLumis and `DQM:@standardDQMFakeHLT+@miniAODDQM+@L1TEgamma`. I'm not sure if it can create problems

chayanit:
> The change looks fine to me, but does anybody know why only 1361.181 uses the concurrent lumis?
> 
> @chayanit

Hi @silviodonato, 
This feature was introduced in PR #27937 and #27928 created by @fabiocos to enable concurrent lumi in both data and MC relvals but so far we have never tested them on the campaign level and probably only PR tests (?). I propose we disable all three concurrent lumi workflows until we understand how to fix the issues. 

@pgunnell

schneiml:
@silviodonato 136.8861 is one of the others that crashed, for the other problem. We'll see if it causes trouble after the fixes were merged (since the PR tests probably did not test multi-threaded running there either).

136.8862 seems to have run just fine, but probably only with luck: It could probably trigger the assertion if edm decides to do thing in a different order.

Re concurrent lumis in the tests in general: I think disabling them for RECO would make sense, at least as long as DQM blocks concurrent running anyways. After #28813 we can talk about that again. But for the other steps we can keep them enabled, to exercise that functionality in EDM.

And please, can we make the PR tests consistent with the IB tests? Use `-t 4` with both or with neither?

Dr15Jones:
>  I propose we disable all three concurrent lumi workflows until we understand how to fix the issues, unless @fabiocos has further comments.

I strongly advise against disabling the testing of concurrent lumis. We need to start using this feature in product to increase our CPU utilization. In order to use in production we have to be sure we do not break it in the IBs.

Dr15Jones:
Matti and I looked at the backtrace for the failing IB. It is not caused by an `edm::EDAnalyzer`. It is caused by a `DQMOneLumiEDAnalyzer`, in particular `AlcaBeamMonitor`.

fabiocos:
@schneiml @chayanit @silviodonato @Dr15Jones just to refresh everybody's memory, I have added last year these workflows to give the FWK group and every developer a first simple tool to test concurrent lumis behaviour at PR test/IB and possibly Relval level. So the test workflow has done its work, it tells that the code is not yet functional for this feature. IMHO the solution is not disabling the test in some way, hiding the problem under the carpet, but making it work by adjusting what is still not functional.

The failure of this workflow does not prevent building the pre-release and running existing standard non-concurrent workflows, but it is useful to monitor progresses in adjusting the code. So I personally agree with @Dr15Jones .

makortel:
Do I understand correctly that the problems are confined in the three "ml" workflows 136.8861, 136.8862, and 1361.181? And that these workflows are the only ones for which the concurrent lumis are enabled?

Do the failures of these workflows cause any problems for the pre-release RelVal production by PdmV? 
As far as I can tell, these workflows look like variations of other workflows (136.886 and 1361.18, respectively), so I'd guess the answer to be "no" (but I could be wrong).

If my assumption of only these workflows, that are run only in the IBs, are affected is correct, the strong preference of the core group is to keep these workflows failing (i.e. concurrent lumis enabled) until the real problem is fixed. Otherwise we would be just unnecessarily hiding problems. (we should probably move the discussion of the underlying problem and how to fix it elsewhere, e.g. an issue)

schneiml:
@Dr15Jones There is no actual problem here -- the `assertLegacySafe` is just a safety check inside the DQMStore. The new DQMStore guarantees to never free objects that were handed to subsystem code. This is violated for legacy modules when there are concurrent lumisections. In this case the assertion fires, to make the user think twice if they are really sure there are no legacy modules present that could hold pointers that might be randomly deleted in the middle of the run.

We can't just only check this for legacy modules, since legacy modules have the nasty property that they don't actually tell us that they are using the DQMStore. They can just come at any moment, go to the edm::Service and request pointers to MEs that any other module (inc. non-legacy module) might have created and then hold them forever. This is why the assertion fires as soon as any ME is copied -- now an object of limited lifetime exists and might leak to the subsystem code.

Of course, that does not mean that anything bad happens. Most legacy code does completely harmless things. Most legacy code does not read other modules MEs. But we can't know without reviewing the code, which has not happened yet. That is why this assertion is there by default. 

Obviously, if you want to make any attempt of getting concurrent lumis you need to turn off the assertion (which is what this PR does, and this is fine, and will be required at some point), but now I can't guarantee that the new DQM framework is memory-safe any more. If that is a problem is another question -- the old DQM code was probably not memory-safe either, and we probably have bigger problems in CMSSW. But what this PR does is disable a warning, and as long as there is random legacy code on the sequences of these jobs (which is the case), this is a valid warning.

I'd prefer if we could run these workflows with a reduced DQM and VALIDATION sequence that does not contain legacy modules. Then, we'd actually even have a chance to get true concurrent lumis after #28813, and everything is safe. But I can't provide such a sequence today.

silviodonato:
Thanks everybody for your replies. Since 1361.181 is not used in the RelVal production and then we will not break the RelVal production, I agree to can go ahead and make CMSSW_11_1_0_pre3 keeping 1361.181.
The failure of 1361.181 in the IB test will remind us that concurrent lumis feature is broken and need to be fixed.

schneiml:
While I agree that we should keep these tests to remind us of the problem, I'd like to point out that
- We will need this PR eventually. It is not safe to enable concurrent lumis today, but once it is, we need to merge this PR to remove the crash.
- Getting rid of all legacy code on the default DQM and VALIDATION sequences is a big task. Moving to a small, limited but safe set of modules similar to e.g. what runs in prompt reco is much more feasible.

fabiocos:
@makortel yes, these "ml" workflows were just a counterpart of the main (so far default) existing version where the concurrentLumi feature is switched on. In order to make this choice effective, I had chosen the input so as to trigger a boundary crossing within the few event range allowed by a PR test.

https://cmssdt.cern.ch/lxr/source/Configuration/PyReleaseValidation/python/relval_steps.py#1523

https://cmssdt.cern.ch/lxr/source/Configuration/PyReleaseValidation/python/relval_steps.py#1840

schneiml:
BTW thanks @fabiocos for providing these tests -- last time we had bigger changes in DQM, I was very worried that we don't really test lumi transitions etc. Having these greatly improves our ability to quickly test things.

But -- we should really run them multi-threaded also in the PR tests. Had we run wf 1361.181 in the PR tests of #28622, it would have succeeded! And the test result above in this PR is in fact meaningless, it does not test the relevant behavior.

davidlange6:
>The failure of 1361.181 in the IB test will remind us that concurrent lumis feature is broken and ?>need to be fixed.

perhaps before building releases one should agree a plan for a fix? (its not that known-to-be broken IB tests are appreciated)

makortel:
@Dr15Jones and I poked around the DQMStore code to try to understand what is going on.

It appears to us that the underlying problem is not with the "legacy DQM modules" themselves, but with any module using DQMStore that holds a pointer to a `TH*` within its member variables rather than asking it from the MonitorElement each time. For right now we assume that only the legacy modules could have this problem, since the modules inheriting from any `DQM`-analyzer base class have been rewritten already. We searched for all legacy modules using DQMStore directly and found 66 classes. Of those, we found that 23 classes had called the `getTH*()` and assigned the pointer to some variable. Quick visual inspection indicated that only two, APVValidationPlots and DummyFillDQMStore, actually hold pointers to the `TH*` objects via their member variables (and both look straightforward to fix).

After those get fixed, we don't see any reason for that particular parameter/assertion.

One thing we wondered was that is it intentional to call `DQMStore::enterLumi()` twice in `DQMOneEDAnalyzer::beginRun()`
https://github.com/cms-sw/cmssw/blob/30b42ebb664ecb39f5ddd45fc3c77acad3c3901a/DQMServices/Core/interface/DQMOneEDAnalyzer.h#L34-L45
All other `DQM*EDAnalyzer`s have call `enterLumi()` only once (as the last action).

We also re-discovered that any `edm::global` module making use of the DQMStore and per-lumi histograms would not work correctly with concurrent lumis (https://github.com/cms-sw/cmssw/issues/28341).

makortel:
> Quick visual inspection indicated that only two, APVValidationPlots and DummyFillDQMStore, actually hold pointers to the `TH*` objects via their member variables (and both look straightforward to fix).

These are addressed in https://github.com/cms-sw/cmssw/pull/28942 and https://github.com/cms-sw/cmssw/pull/28941, respectively.

schneiml:
@makortel Your analysis is mostly correct, but I am afraid you are missing one point:

- You observed that if modules hold pointers to `TH1`, they will be affected by this problem, if they are legacy or not. That is true, I mostly forgot about it because it is not very common in practice (which you also observed). Holding pointers to `TH1` is still guaranteed to be safe as long as `assertLegacySafe` does not fire *and* the job is single-threaded.

- But, the problem appears also when modules hold `MonitorElement` pointers, which is pretty much how the API is defined to work right now. In the non-legacy modules, what we actually hand out is *local* MEs, where the `DQMStore` keeps track of who holds them and adjusts which histogram is connected to them as the modules enter and leave lumis. But, to drive this mechanism, the `DQMstore` relies on callbacks from the modules itself, and these callbacks are only sent by the `DQM*EDAnalyzer` base classes. For legacy modules, there is no base class, and there are no callbacks, and they might end up filling the wrong histogram. And even worse, once that histogram is saved, it might be deleted.

- We guarantee that the same ME object will be used for every lumi/run as long as there are no concurrent lumis (so we can get away with a single one), and all of the switching logic is essentially pointless (it still updates some metadata, but the histogram data will always go to the right place). But when there are multiple conucurrent lumis, there will be multiple objects, and they will be allocated and deleted as needed, and if there is no callback to bend the pointers inside the `local` MEs, they will go stale.

- (Now, just to make things a bit more confusing, the legacy modules actually get returned pointers to `global` MEs in booking. But that does not really change anything. The reasoning was simply that this way, `book("thing") == get("thing")` and we don't have to worry about `local` MEs in harvesting.)

Getting back to the actual issue here: I do think that in this workflow nothing evil happens and the assertion can be safely disabled. To actually trigger a problem, we'd need a legacy module booking per-lumi MEs *and* holding pointers to them. There are not many modules booking per-lumi MEs (I touched them all manually), and I don't remember seeing anything that seemed terribly unsafe. But, the code could still be changed in the future to do dangerous things, and these changes might be very unobvious (I am worried about deep hierarchies of helper classes, the `TkMap` code comes to mind.) Also, this would become way more problematic with concurrent runs. So, I still think we have a timebomb there, and I put in this assertion specifically to remind of that.

makortel:
> There are not many modules booking per-lumi MEs (I touched them all manually), and I don't remember seeing anything that seemed terribly unsafe.

Thanks, that is somewhat reassuring.

> But, the code could still be changed in the future to do dangerous things, and these changes might be very unobvious (I am worried about deep hierarchies of helper classes, the `TkMap` code comes to mind.)

I think we would need to establish a policy that no such changes are allowed anymore (enforced at the code review), and make a plan to migrate (or remove) the legacy code (at least from any possible workflow run centrally).

makortel:
After https://github.com/cms-sw/cmssw/pull/28966, only workflow 1000.0 from the full matrix has legacy DQM modules being run in before-harvesting steps. First of those is `MuonAlignment`, for which I opened an issue https://github.com/cms-sw/cmssw/issues/28975. I don't know if there are more.

makortel:
> First of those is `MuonAlignment`, for which I opened an issue #28975. I don't know if there are more.

A further test reveals that the `MuonAlignment` is the only legacy DQM module in the step2 of 1000.0.

silviodonato:
With #29027 we removed `MuonAlignment` from wf 1000.0 and from CMSSW

makortel:
Now that full matrix non-harvesting steps should be clean from legacy modules, in theory this configuration flag should no longer be needed (in full matrix). I'm left to wonder why the assert fires also for `DQMOneLumiEDAnalyzer`:

https://github.com/cms-sw/cmssw/pull/28920#issuecomment-585260181
> Matti and I looked at the backtrace for the failing IB. It is not caused by an `edm::EDAnalyzer`. It is caused by a `DQMOneLumiEDAnalyzer`, in particular `AlcaBeamMonitor`.

(is it perhaps not detecting legacy modules properly?)

makortel:
@Dr15Jones reminded me that my question was essentially answered in https://github.com/cms-sw/cmssw/pull/28920#issuecomment-585268392.

Can `assertLegacySafe` be removed then?

schneiml:
@makortel the key point is that this assert does not detect the presence of legacy modules, but simply watches the behavior of the `DQMStore` (which is in the end driven by EDM). That makes it less useful, but also much simpler, and much easier to understand:
- If `assertLegacySafe` is set: There is only a single histogram for each ME name. No cloning, duplicating, free'ing happens ever. Holding any sort of pointers is safe.
- Else: there may be multiple histograms for each ME. In some cases, it may not be safe to touch pointers handed out earlier. It is safe in the transitions exposed by non-legacy modules.

Ideally the assert would check if legacy modules are present in the job, but that is quite hard, since the very specific property of legacy modules is that they might spontaneously, at any point in time, get the `edm::Service` and call into the `DQMStore`. So the `DQMStore` cannot know if there actually are legacy modules without asking for this information somewhere else. Therefore, the assert is conservative and has to be removed manually.

makortel:
Ok, so the change in this PR should cover all jobs whose configuration is generated with `cmsDriver`. Does that propagate to T0 configurations automatically or not? (I'm a bit afraid of the latter)

davidlange6:
The t0 uses config builder. So should be good. Can be checked via the unit tests in Configuration/dataProcessing

On Feb 25, 2020 1:09 PM, Matti Kortelainen <notifications@github.com> wrote:

Ok, so the change in this PR should cover all jobs whose configuration is generated with cmsDriver. Does that propagate to T0 configurations automatically or not? (I'm a bit afraid of the latter)

silviodonato:
@schneiml have you tried to use the tool `edmCheckMultithreading` to check the modules in the DQM sequence?

schneiml:
@silviodonato I am at it. It cave three hot candidates: https://cmssdt.cern.ch/dxr/CMSSW/source/Validation/RecoTau/plugins/DQMHistNormalizer.cc , https://cmssdt.cern.ch/dxr/CMSSW/source/DQM/TrigXMonitorClient/src/L1ScalersClient.cc , and https://cmssdt.cern.ch/dxr/CMSSW/source/Validation/RecoTau/plugins/DQMFileLoader.cc . 

Though, I think these are false positives, since `edmCheckMultithreading` will check all modules _configured_ in the configuration, even if they are not on any sequence. So no I am marrying https://github.com/cms-sw/cmssw/compare/master...schneiml:dqm-dqmdumpsequence with the logic from `edmCheckMultithreading` for a more precise check.

schneiml:
The new combined tool [1] indicates that there are no legacy DQM modules on any DQM or VALIDATION sequences defined in `autoDQM`/`autoValidation`, at least for scenario pp and a recent era. That is a good sign. There are a few non-DQM legacy modules though, but they should not hurt.

There might be legacy modules hiding in other corners of the configuration parameter space, but chances are not very high.

[1] https://github.com/cms-sw/cmssw/compare/master...schneiml:dqm-dqmdumpsequence

makortel:
@schneiml 

> There are a few non-DQM legacy modules though, but they should not hurt.

Could you please list them in #25090?

silviodonato:
@schneiml , we can try to run your tests https://github.com/cms-sw/cmssw/compare/master...schneiml:dqm-dqmdumpsequence in a special IB and check if this is sustainable or not.
@smuzaffar how do you suggest to proceed? Perhaps the easiest (quick&dirty) way is to simply integrate these test (https://github.com/cms-sw/cmssw/compare/master...schneiml:dqm-dqmdumpsequence). If they take too much time we can revert the change. A nicer and more complex way is to have a special IB including these new tests.

schneiml:
@silviodonato as you might have seen I just updated the test with @smuzaffar 's comments. Now it should not interfer with other tests, but it might take ~2h [1] to complete. Splitting it into more pieces would be better, but then we'd need to arbitrarily split the workflow numbers, and I'd prefer to avoid that complexity. 

[1] Extrapolating from that it took 15min at 16 threads. Actually I run it single-threaded on `cmsswconfigexplore` to serve the html UI, and it takes almost 24h to run, but that is on the smallest VM openstack has to offer...

silviodonato:
+operations

silviodonato:
+1
See https://github.com/cms-sw/cmssw/pull/29150

Configuration/Applications/python/ConfigBuilder.py
```diff
@@ -2232,6 +2232,8 @@ def prepare(self, doChecking = False):
             self.pythonCfgCode +="process.options.numberOfThreads=cms.untracked.uint32("+self._options.nThreads+")\n"
             self.pythonCfgCode +="process.options.numberOfStreams=cms.untracked.uint32("+self._options.nStreams+")\n"
             self.pythonCfgCode +="process.options.numberOfConcurrentLuminosityBlocks=cms.untracked.uint32("+self._options.nConcurrentLumis+")\n"
+            if self._options.nConcurrentLumis > 1:
+              self.pythonCfgCode +="if process.DQMStore: process.DQMStore.assertLegacySafe=cms.untracked.bool(False)\n"
```
makortel: This should have probably been something like
```python
self.pythonCfgCode +="if hasattr(process, 'DQMStore'): process.DQMStore.assertLegacySafe=cms.untracked.bool(False)\n"
```
The latest IB (CMSSW_11_1/2020-03-16-1100) shows failures like
```
LHE input from article  18334
Note: this tool is DEPRECATED, use xrdfs instead.
customising the process with customiseWithTimeMemorySummary from Validation/Performance/TimeMemorySummary
Starting /data/cmsbld/jenkins/workspace/ib-run-relvals/cms-bot/monitor_workflow.py timeout --signal SIGTERM 9000  cmsRun -j JobReport1.xml  step1_NONE.py
----- Begin Fatal Exception 16-Mar-2020 12:51:11 CET-----------------------
   [0] Processing the python configuration file named step1_NONE.py
Exception Message:
AttributeError: 'Process' object has no attribute 'DQMStore'

At:
  step1_NONE.py(92): &lt;module&gt;

```
https://cmssdt.cern.ch/SDT/cgi-bin/logreader/slc7_amd64_gcc820/CMSSW_11_1_X_2020-03-16-1100/pyRelValMatrixLogs/run/1370.0_GluGluHToGG_M125_Pow_MINLO_NNLOPS_py8_13+GluGluHToGG_M125_Pow_MINLO_NNLOPS_py8_13+Hadronizer_TuneCUETP8M1_13TeV_powhegEmissionVeto_2p_HToGG_M125_13+DIGIUP15+RECOUP15+HARVESTUP15+NANOUP15Had/step1_GluGluHToGG_M125_Pow_MINLO_NNLOPS_py8_13+GluGluHToGG_M125_Pow_MINLO_NNLOPS_py8_13+Hadronizer_TuneCUETP8M1_13TeV_powhegEmissionVeto_2p_HToGG_M125_13+DIGIUP15+RECOUP15+HARVESTUP15+NANOUP15Had.log#/

Issue #15248, opened 2016-07-21
C++ Modules support (based on Clang)


davidlt:
The current very basic draft exist here: https://github.com/davidlt/cmssw/tree/cpp-modules
It can only be used with DEVEL IBs as it requires Clang pre-3.9. We will be updating Clang as needed in DEVEL IBs.

How to use it?

```
scram p CMSSW_8_1_DEVEL_X_2016-07-19-1100
cmsenv
git cms-merge-topic davidlt:cpp-modules
rm -rf $CMSSW_BASE/cppmodulescache
mkdir -p $CMSSW_BASE/cppmodulescache

# Modify $CMSSW_BASE/config/BuildFile.xml
vim $CMSSW_BASE/config/BuildFile.xml

# Add the following line:
#
#   <flags INCLUDE="$(LOCALTOP)/src/FWCore/GNUCppStandardLibrary/interface"/>
#
# after:
#
#   <classpath path="+Project/LCG+SubSystem/+Package/tests+binary"/>
#

USER_CXXFLAGS="-fmodules -fmodules-cache-path=$CMSSW_BASE/cppmodulescache" scram b -v -j 1 COMPILER=llvm 2>&1 | tee b.log

# Alternative command:
USER_CXXFLAGS="-fmodules -Xclang -fmodules-local-submodule-visibility -fmodules-cache-path=$CMSSW_BASE/cppmodulescache" scram b -v -j 1 COMPILER=llvm 2>&1 | tee b.log
```

davidlt:
@vgvassilev: iostream.h -> iostream

davidlt:
@vgvassilev if we use `-Xclang -fmodules-local-submodule-visibility` we get:
```
>> Compiling  /mnt/build/davidlt/cppmodules/CMSSW_8_1_DEVEL_X_2016-07-17-2300/src/FWCore/MessageService/bin/Standalone.cpp
[...]

PR #28925, opened 2020-02-11
[cxxmodules] Workaround missing CLHEP modulemap.

vgvassilev:
Currently all contents are pasted N times in the including modules due to the lack of a CLHEP module for the external to cmssw package.

This walks on a very fragile path in the compiler (and roocling) as it needs to figure out while parsing the header content that a module already contained the definitions being parsed.

This patch avoids an issue where rootcling thinks that CLHEP/.../ThreeVector.h is included multiple times. We essentially tell the compiler to open the definition from a module rather than parsing.

Part of #15248

cc: @davidlange6, @smuzaffar, @oshadura

### DIFF
--- a/DataFormats/BTauReco/src/classes.h
+++ b/DataFormats/BTauReco/src/classes.h
@@ -16,7 +16,7 @@
 #include "DataFormats/Common/interface/ValueMap.h"
 #include "DataFormats/Candidate/interface/Candidate.h"
 #if defined __has_feature
-#  if __has_feature(modules)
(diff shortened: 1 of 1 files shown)


smuzaffar:
Please test

smuzaffar:
please test

slava77:
+1

for https://github.com/cms-sw/cmssw/pull/28925	d8f03f7 
- technical; looks in line with the description

silviodonato:
+1

PR #28933, opened 2020-02-12
Fix string operation in DDCMS

mrodozov:
#### PR description:

The latest ROOT integration builds failed because of string_view + string operation. 
Seems ns.name() was returning string before but it returns string view now. To be confirmed

#### PR validation:

It compiles with the updated DD4Hep externals (and ROOT)

### DIFF
--- a/DetectorDescription/DDCMS/plugins/dd4hep/DDAngular.cc
+++ b/DetectorDescription/DDCMS/plugins/dd4hep/DDAngular.cc
@@ -35,7 +35,7 @@ static long algorithm(Detector& /* description */, cms::DDParsingContext& ctxt, 
   Volume mother = ns.volume(args.parentName());
   string childName = args.value<string>("ChildName");
   if (strchr(childName.c_str(), NAMESPACE_SEP) == nullptr)
-    childName = ns.name() + childName;
+    childName = ns.prepend(childName);
   Volume child = ns.volume(childName);
 
   double delta = 0e0;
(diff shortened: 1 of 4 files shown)


ianna:
-1

> The latest ROOT integration builds failed because of string_view + string operation.
> Seems ns.name() was returning string before but it returns string view now. To be confirmed
> 
@mrodozov - IMHO this has nothing to do with ROOT, the DDNamespace API has not changed:
https://cmssdt.cern.ch/lxr/source/DetectorDescription/DDCMS/interface/DDNamespace.h#0068

There is a member function to append the namespace correctly:
https://cmssdt.cern.ch/lxr/source/DetectorDescription/DDCMS/src/DDNamespace.cc#0066

ianna:
@mrodozov - please, use #28951 instead

DetectorDescription/DDCMS/plugins/dd4hep/DDAngular.cc
```diff
@@ -35,7 +35,7 @@ static long algorithm(Detector& /* description */, cms::DDParsingContext& ctxt,
   Volume mother = ns.volume(args.parentName());
   string childName = args.value<string>("ChildName");
   if (strchr(childName.c_str(), NAMESPACE_SEP) == nullptr)
-    childName = ns.name() + childName;
+    childName = ns.prepend(childName);
```
ianna: The line above should be removed since ```ns.prepend``` does this comparison to avoid adding a namespace to a name that already has it. Yet another reason to forbid a blind ```+``` operation :-)

ianna:
-1

mrodozov:
No worries I'm closing this in favor of your pr

PR #28934, opened 2020-02-12
DQM: Fix EDMtoMEConverter.

schneiml:
#### PR description:

This PR fixes a problem pointed out by @mmusich related to the AlCa multi-run harvesting workflows, caused by #28622.

In the past, the DQMStore, when used with "collate" mode, would
automatically merge the `TH1` object passed into a `book*()` call with an
existing object. However, the merge logic in the `DQMStore` is rather
conservative and didn't merge objects with `SetCanExtend` set. So, to
work around this, `EDMtoMEConverter` would merge these objects itself.

The new `DQMStore` does not have the "collate" mode any more and all the
merging has to happen in the input modules. This fixes
`EDMtoMEConverter` to *always* use its local merge logic.

Arguably the ME merge logic should be factored out of DQMRootSource and
EDMtoMEConverter and shared between the two, but for now this small fix
should work to get the AlCa workflows working again.

#### PR validation:

Privately provided test configuration seems to provide correct results, needs further testing by AlCa.

I have not much confidence that the `EDMtoMEConverter` behaves correctly for all combinations of settings, but in the end this module is there to support the AlCa workflows dating back to Run1 and the only thing that matters is that these work correctly.

### DIFF
--- a/DQMServices/Components/plugins/EDMtoMEConverter.cc
+++ b/DQMServices/Components/plugins/EDMtoMEConverter.cc
@@ -149,18 +149,17 @@ namespace {
 
       if (me) {
         auto histo = HistoTraits<T>::get(me);
-        if (histo && me->getTH1()->CanExtendAllAxes()) {
-          TList list;
-          list.Add(metoedmobject);
-          if (histo->Merge(&list) == -1)
-            std::cout << "ERROR EDMtoMEConverter::getData(): merge failed for '" << metoedmobject->GetName() << "'"
-                      << std::endl;
-          return me;
-        }
+        assert(histo);
+        TList list;
+        list.Add(metoedmobject);
+        if (histo->Merge(&list) == -1)
+          edm::LogError("EDMtoMEConverter") << "ERROR EDMtoMEConverter::getData(): merge failed for '"
+                                            << metoedmobject->GetName() << "'" << std::endl;
+        return me;
+      } else {
+        iBooker.setCurrentFolder(dir);
+        return HistoTraits<T>::book(iBooker, metoedmobject->GetName(), metoedmobject);
       }
-
-      iBooker.setCurrentFolder(dir);
-      return HistoTraits<T>::book(iBooker, metoedmobject->GetName(), metoedmobject);
     }
   };
 


schneiml:
please test

Wondering about the unit tests and wf 1001, otherwise this code should not affect anything.

mmusich:
thanks @schneiml !
@silviodonato I am wondering if there is a way to run the tests with higher statistics in the PR tests.

silviodonato:
> @silviodonato I am wondering if there is a way to run the tests with higher statistics in the PR tests.

I don't think it is possible. @smuzaffar ? @fabiocos ?

smuzaffar:
currently not possible

mmusich:
For some reason relmon for wf 1001.0 is not available... https://cmssdt.cern.ch/SDT/jenkins-artifacts/baseLineComparisons/CMSSW_11_1_X_2020-02-12-1100+ac6b9c/35044/

looking to other PRs seems a persistent issue in PR tests. Who's responsible to fix that? PdmV?

schneiml:
@tocheng  @mmusich please let me know if this properly fixes the EDMtoME related problems. I'd sign it then.

mmusich:
@schneiml, I'll hopefully get to this later this week.

mmusich:
@schneiml I tested your commit with the simple setup I PR'ed at https://github.com/cms-sw/cmssw/pull/28994 and it works fine (at least it removes the problem I had originally observed). For the rest of the AlCa workflows I strongly suggest the AlCa/DB conveners @tocheng @christopheralanwest @tlampen to comment (and possibly to test further) but I don't see red flags with the proposed fix.

schneiml:
+1

Not sure yet if this is the last word for getting all AlCa workflows working again, but certainly better than what we have now.

Thanks @mmusich for improving test coverage of these corners of CMSSW!

silviodonato:
assign alca

> For the rest of the AlCa workflows I strongly suggest the AlCa/DB conveners @tocheng @christopheralanwest @tlampen to comment (and possibly to test further) but I don't see red flags with the proposed fix.

silviodonato:
@tocheng @christopheralanwest @tlampen do you have any comments on this PR?

schneiml:
We should get this merged at some point -- without this PR, the AlCa jobs _are_ broken. `11_1_0_pre3` _is_ broken in that respect and we should certainly have a fix for pre4.

mmusich:
I fully support Marcel here!

pohsun:
+1

PR #28943, opened 2020-02-12
[cxxmodules] Workaround missing CLHEP modulemap for JetReco.

vgvassilev:
In the same manner as in cms-sw/cmssw#28925 we force the definition of Hep3Vector to be read from the module that contains it rather than parsing it.

Part of #15248

cc: @davidlange6, @smuzaffar, @oshadura.

### DIFF
--- a/DataFormats/JetReco/src/classes_3.h
+++ b/DataFormats/JetReco/src/classes_3.h
@@ -5,6 +5,12 @@
 
 #include "DataFormats/JetReco/interface/BasicJetCollection.h"
 #include "DataFormats/JetReco/interface/CaloJetCollection.h"
+#if defined __has_feature
+#if __has_feature(modules)
+// Workaround the missing CLHEP.modulemap: CLHEP/Vector/ThreeVector.h:41:7: error: redefinition of 'Hep3Vector'
+#include "DataFormats/JetReco/interface/PFClusterJetCollection.h"
+#endif
+#endif
 #include "DataFormats/JetReco/interface/JPTJetCollection.h"
 #include "DataFormats/JetReco/interface/GenJetCollection.h"
 #include "DataFormats/JetReco/interface/PFJetCollection.h"


smuzaffar:
please test

davidlange6:
please test

(Changes have no effect on mainstream)

On Feb 12, 2020, at 3:17 PM, cmsbuild <notifications@github.com<mailto:notifications@github.com>> wrote:


+code-checks

Logs: https://cmssdt.cern.ch/SDT/code-checks/cms-sw-PR-28943/13752

  *   This PR adds an extra 16KB to repository

DataFormats/JetReco/src/classes_3.h
```diff
@@ -5,6 +5,12 @@
 
 #include "DataFormats/JetReco/interface/BasicJetCollection.h"
 #include "DataFormats/JetReco/interface/CaloJetCollection.h"
+#if defined __has_feature
+#if __has_feature(modules)
+// Workaround the missing CLHEP.modulemap: CLHEP/Vector/ThreeVector.h:41:7: error: redefinition of 'Hep3Vector'
+#include "DataFormats/JetReco/interface/PFClusterJetCollection.h"
```
slava77: similar to #28925 , I was expecting that DataFormats/JetReco/interface/PFClusterJetCollection.h (already mentioned below) is paired up with ```#include "CLHEP/Vector/ThreeVector.h"```

Otherwise, what is the benefit of this conditional double-inclusion.
Can this include be here just unconditionally or (still unconditionally) twice, if double-mention is needed?

vgvassilev: Good questions.

I think PR #28925 may have been good enough with the single include without the `ThreeVector.h`. The include order matters because we do not yet have a module for CLHEP. This means that CLHEP headers will be contained in other modules while some includes (which include CLHEP headers) will parse.

Moving the include up activates the copy of the CLHEP headers from `PFClusterJetCollection.h` rather than reparsing and thus avoiding a bug in rootcling (clang).

slava77: > PS: I would rather keep the duplicate include so that once we have the CLHEP module I will remove the workarounds altogether.

did you mean to keep it unconditionally (change this PR), or stay with the conditional duplication?
Please clarify.

vgvassilev: I meant to keep it conditional as it is now in the PR.

slava77:
+1

for https://github.com/cms-sw/cmssw/pull/28943	bcfd273 
- technical, even though somewhat kludgy 
- jenkins tests pass

silviodonato:
+1

PR #28945, opened 2020-02-13
Sync with stand-alone Pixel Calibration Code

OzAmram:
This is a PR to sync with changes made to the pixel calibration procedure. The idea is to call the actual CMSSW functions that are used in the pixel calibrations procedure rather than separately coded versions of them (like was done previously). To do this we needed to modify the objects and code structure a bit. The most significant changes are adding a sideload() function to SiPixelTemplate object to allow loading of partial templates, and moving the generic_position_formula() to a separate file.

These changes are all internal to the pixel code and we expect no changes in output.

We have tested by running the tracking validation and seeing that the PR makes no difference in the output at all.

Slides about this were shown in the pixel offline meeting. [Here](https://indico.cern.ch/event/879467/contributions/3746113/attachments/1985814/3308855/sideload_PR_feb11.pdf)

@cmantill @pmaksim1 @tvami @mmusich @tsusa

### DIFF
--- a/CondFormats/SiPixelTransient/interface/SiPixelTemplate.h
+++ b/CondFormats/SiPixelTransient/interface/SiPixelTemplate.h
@@ -700,7 +700,7 @@ private:
   float cota_current_;  //!< current cot alpha
   float cotb_current_;  //!< current cot beta
   float abs_cotb_;      //!< absolute value of cot beta
-  int Dtype_;           //!< flags BPix (=0) or FPix (=1)
+  int dtype_;           //!< flags BPix (=0) or FPix (=1)
   bool flip_y_;         //!< flip y sign-sensitive quantities
   bool flip_x_;         //!< flip x sign-sensitive quantities
   bool success_;        //!< true if cotalpha, cotbeta are inside of the acceptance (dynamically loaded)
--- a/CondFormats/SiPixelTransient/interface/SiPixelTemplate.h
+++ b/CondFormats/SiPixelTransient/interface/SiPixelTemplate.h
@@ -1,5 +1,5 @@
 //
-//  SiPixelTemplate.h (v10.20)
+//  SiPixelTemplate.h (v10.24)
 //
 //  Add goodness-of-fit info and spare entries to templates, version number in template header, more error checking
 //  Add correction for (Q_F-Q_L)/(Q_F+Q_L) bias
@@ -79,6 +79,7 @@
 //  V10.20 - Add directory path selection to the ascii pushfile method
 //  V10.21 - Address runtime issues in pushfile() for gcc 7.X due to using tempfile as char string + misc. cleanup [Petar]
 //  V10.22 - Move templateStore to the heap, fix variable name in pushfile() [Petar]
+//  V10.24 - Add sideload() + associated gymnastics [Petar and Oz]
 
 // Created by Morris Swartz on 10/27/06.
 //
@@ -258,17 +259,24 @@ public:
     index_id_ = -1;
     cota_current_ = 0.;
     cotb_current_ = 0.;
+    entry_sideloaded_ = nullptr; 
   }  //!< Constructor for cases in which template store already exists
 
 // Load the private store with info from the file with the index (int) filenum from directory dir:
 //   ${dir}template_summary_zp${filenum}.out
 #ifdef SI_PIXEL_TEMPLATE_STANDALONE
   static bool pushfile(int filenum, std::vector<SiPixelTemplateStore>& pixelTemp, std::string dir = "");
+
+   // For calibrations only: load precalculated values -- no interpolation.
+   void sideload(SiPixelTemplateEntry* entry, int iDtype, float locBx, float locBz, float lorwdy, float lorwdx, float q50, float fbin[3], float xsize, float ysize, float zsize);
+
 #else
   static bool pushfile(int filenum,
                        std::vector<SiPixelTemplateStore>& pixelTemp,
                        // *&^%$#@!  Different default dir -- remove once FastSim is updated.
                        std::string dir = "CalibTracker/SiPixelESProducers/data/");
+  
+  //load from DB (default in CMSSW)
   static bool pushfile(const SiPixelTemplateDBObject& dbobject,
                        std::vector<SiPixelTemplateStore>& pixelTemp);  // load the private store with info from db
 #endif
@@ -296,7 +304,7 @@ public:
   //Method to estimate the central pixel of the interpolated x-template
   int cxtemp();
 
-  // new methods to build templates from two interpolated clusters (for splitting)
+  //Methods to build templates from two interpolated clusters (for splitting)
   void ytemp3d_int(int nypix, int& nybins);
 
   void ytemp3d(int j, int k, std::vector<float>& ytemplate);
@@ -682,6 +690,9 @@ private:
   float cota_current_;  //!< current cot alpha
   float cotb_current_;  //!< current cot beta
   float abs_cotb_;      //!< absolute value of cot beta
+  int Dtype_;           //!< flags BPix (=0) or FPix (=1)
+  bool flip_y_;         //!< flip y sign-sensitive quantities
+  bool flip_x_;         //!< flip x sign-sensitive quantities
   bool success_;        //!< true if cotalpha, cotbeta are inside of the acceptance (dynamically loaded)
 
   // Keep results of last interpolation to return through member functions
@@ -706,6 +717,16 @@ private:
   float qmin_;              //!< minimum cluster charge for valid hit (keeps 99.9% of simulated hits)
   float clsleny_;           //!< y-cluster length of smaller interpolated template in pixels
   float clslenx_;           //!< x-cluster length of smaller interpolated template in pixels
+  float scalexavg_;         //!< average x-error scale factor
+  float scaleyavg_;         //!< average y-error scale factor
+  float delyavg_;           //!< average difference between clsleny_ and cluster length [with threshold effects]
+  float delysig_;           //!< rms of difference between clsleny_ and cluster length [with threshold effects]
+  float scalex_[4];         //!< x-error scale factor in charge bins
+  float scaley_[4];         //!< y-error scale factor in charge bins
+  float offsetx_[4];        //!< x-offset in charge bins
+  float offsety_[4];        //!< y-offset in charge bins
+
+
   float yratio_;            //!< fractional distance in y between cotbeta templates
   float yparl_[2][5];       //!< projected y-pixel uncertainty parameterization for smaller cotbeta
   float yparh_[2][5];       //!< projected y-pixel uncertainty parameterization for larger cotbeta
@@ -774,8 +795,21 @@ private:
   boost::multi_array<float, 2> temp2dy_;  //!< 2d-primitive for spltting 3-d template
   boost::multi_array<float, 2> temp2dx_;  //!< 2d-primitive for spltting 3-d template
 
-  // The actual template store is a std::vector container
+  // Pointers to presently interpolated point:
+  const SiPixelTemplateEntry* enty0_;   // enty[ilow]
+  const SiPixelTemplateEntry* enty1_;   // enty[iylow][ilow]
 
+  const SiPixelTemplateEntry* entx00_ ; // entx[iylow][ilow]
+  const SiPixelTemplateEntry* entx02_ ; 
+  const SiPixelTemplateEntry* entx20_ ;
+  const SiPixelTemplateEntry* entx22_ ;
+  const SiPixelTemplateEntry* entx21_ ;
+
+  // Pointer to the sideloaded Entry: use this one if set.
+  const SiPixelTemplateEntry * entry_sideloaded_ ;
+   
+
+  // The actual template store is a std::vector container
   const std::vector<SiPixelTemplateStore>& thePixelTemp_;
 };
 
--- a/CondFormats/SiPixelTransient/interface/SiPixelTemplate.h
+++ b/CondFormats/SiPixelTransient/interface/SiPixelTemplate.h
@@ -259,7 +259,7 @@ public:
     index_id_ = -1;
     cota_current_ = 0.;
     cotb_current_ = 0.;
-    entry_sideloaded_ = nullptr; 
+    entry_sideloaded_ = nullptr;
   }  //!< Constructor for cases in which template store already exists
 
 // Load the private store with info from the file with the index (int) filenum from directory dir:
@@ -267,15 +267,25 @@ public:
 #ifdef SI_PIXEL_TEMPLATE_STANDALONE
   static bool pushfile(int filenum, std::vector<SiPixelTemplateStore>& pixelTemp, std::string dir = "");
 
-   // For calibrations only: load precalculated values -- no interpolation.
-   void sideload(SiPixelTemplateEntry* entry, int iDtype, float locBx, float locBz, float lorwdy, float lorwdx, float q50, float fbin[3], float xsize, float ysize, float zsize);
+  // For calibrations only: load precalculated values -- no interpolation.
+  void sideload(SiPixelTemplateEntry* entry,
+                int iDtype,
+                float locBx,
+                float locBz,
+                float lorwdy,
+                float lorwdx,
+                float q50,
+                float fbin[3],
+                float xsize,
+                float ysize,
+                float zsize);
 
 #else
   static bool pushfile(int filenum,
                        std::vector<SiPixelTemplateStore>& pixelTemp,
                        // *&^%$#@!  Different default dir -- remove once FastSim is updated.
                        std::string dir = "CalibTracker/SiPixelESProducers/data/");
-  
+
   //load from DB (default in CMSSW)
   static bool pushfile(const SiPixelTemplateDBObject& dbobject,
                        std::vector<SiPixelTemplateStore>& pixelTemp);  // load the private store with info from db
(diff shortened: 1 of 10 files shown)


OzAmram:
Hmmm it seems the version of code-format that I have locally (in CMSSW_11_1_0_pre2) disagrees with the version being used here. When I run code-format in my local version it didn't catch these last two changes and when I applied them manually and then ran it again it actually changed them back.

I applied them manually again so the test can hopefully proceed

perrotta:
please test

ggovi:
+1

perrotta:
> These changes are all internal to the pixel code and we expect no changes in output.

There are indeed several changes in the reco outputs of the FastSim workflows, see https://cmssdt.cern.ch/SDT/jenkins-artifacts/baseLineComparisons/CMSSW_11_1_X_2020-02-12-2300+047c9d/35061/validateJR.html

@OzAmram please check

CondFormats/SiPixelTransient/interface/SiPixelTemplate.h
```diff
@@ -682,30 +700,42 @@ class SiPixelTemplate {
   float cota_current_;  //!< current cot alpha
   float cotb_current_;  //!< current cot beta
   float abs_cotb_;      //!< absolute value of cot beta
+  int Dtype_;           //!< flags BPix (=0) or FPix (=1)
```
perrotta: Lower case initial should be used for class members

CondFormats/SiPixelTransient/src/SiPixelUtils.cc
```diff
@@ -0,0 +1,90 @@
+#ifndef SI_PIXEL_TEMPLATE_STANDALONE
```
perrotta: Perhaps, if you include the implementation in the header file you can avoid this specialization with ```#ifndef``` etc.

CondFormats/SiPixelTransient/src/SiPixelTemplate.cc
```diff
@@ -1483,120 +1489,103 @@ bool SiPixelTemplate::interpolate(int id, float cotalpha, float cotbeta, float l
...
-                    yratio * thePixelTemp_[index_id_].enty[ihigh].chi2xmin[i];
-      yavgc2m_[i] = (1.f - yratio) * thePixelTemp_[index_id_].enty[ilow].yavgc2m[i] +
-                    yratio * thePixelTemp_[index_id_].enty[ihigh].yavgc2m[i];
-      if (flip_y) {
+      yavg_[i] = (1.f - yratio) * enty0_->yavg[i] + yratio * enty1_->yavg[i];
+      if (flip_y_) {
+        yavg_[i] = -yavg_[i];
+      }
+      yrms_[i] = (1.f - yratio) * enty0_->yrms[i] + yratio * enty1_->yrms[i];
+      //	      ygx0_[i]=(1.f - yratio)*enty0_->ygx0[i] + yratio*enty1_->ygx0[i];
```
perrotta: Is this commented out code needed: add one line explaining its usage, if so; otherwise just remove

CondFormats/SiPixelTransient/src/SiPixelTemplate.cc
```diff
@@ -1605,21 +1594,15 @@ bool SiPixelTemplate::interpolate(int id, float cotalpha, float cotbeta, float l
...
+    chi2yavgone_ = (1.f - yratio) * enty0_->chi2yavgone + yratio * enty1_->chi2yavgone;
+    chi2yminone_ = (1.f - yratio) * enty0_->chi2yminone + yratio * enty1_->chi2yminone;
+    chi2xavgone = (1.f - yratio) * enty0_->chi2xavgone + yratio * enty1_->chi2xavgone;
+    chi2xminone = (1.f - yratio) * enty0_->chi2xminone + yratio * enty1_->chi2xminone;
+
+    fracyone_ = (1.f - yratio) * enty0_->fracyone + yratio * enty1_->fracyone;
+    fracytwo_ = (1.f - yratio) * enty0_->fracytwo + yratio * enty1_->fracytwo;
     //       for(i=0; i<10; ++i) {
-    //		    pyspare[i]=(1.f - yratio)*thePixelTemp_[index_id_].enty[ilow].yspare[i] + yratio*thePixelTemp_[index_id_].enty[ihigh].yspare[i];
+    //		    pyspare[i]=(1.f - yratio)*enty0_->yspare[i] + yratio*enty1_->yspare[i];
```
perrotta: Is this commented out code needed: add one line explaining its usage, if so; otherwise just remove

CondFormats/SiPixelTransient/src/SiPixelTemplate.cc
```diff
@@ -1739,57 +1720,49 @@ bool SiPixelTemplate::interpolate(int id, float cotalpha, float cotbeta, float l
...
-      xrms_[i] = (1.f - yxratio) * ((1.f - xxratio) * thePixelTemp_[index_id_].entx[iylow][ilow].xrms[i] +
-                                    xxratio * thePixelTemp_[index_id_].entx[iylow][ihigh].xrms[i]) +
-                 yxratio * ((1.f - xxratio) * thePixelTemp_[index_id_].entx[iyhigh][ilow].xrms[i] +
-                            xxratio * thePixelTemp_[index_id_].entx[iyhigh][ihigh].xrms[i]);
+      xrms_[i] = (1.f - yxratio) * ((1.f - xxratio) * entx00_->xrms[i] + xxratio * entx02_->xrms[i]) +
+                 yxratio * ((1.f - xxratio) * entx20_->xrms[i] + xxratio * entx22_->xrms[i]);
 
-      //	      xgx0_[i]=(1.f - yxratio)*((1.f - xxratio)*thePixelTemp_[index_id_].entx[iylow][ilow].xgx0[i] + xxratio*thePixelTemp_[index_id_].entx[iylow][ihigh].xgx0[i])
-      //		          +yxratio*((1.f - xxratio)*thePixelTemp_[index_id_].entx[iyhigh][ilow].xgx0[i] + xxratio*thePixelTemp_[index_id_].entx[iyhigh][ihigh].xgx0[i]);
+      //	      xgx0_[i]=(1.f - yxratio)*((1.f - xxratio)*entx00_->xgx0[i] + xxratio*entx02_->xgx0[i])
```
perrotta: Is this commented out code needed: add one line explaining its usage, if so; otherwise just remove

CondFormats/SiPixelTransient/src/SiPixelTemplate.cc
```diff
@@ -1739,57 +1720,49 @@ bool SiPixelTemplate::interpolate(int id, float cotalpha, float cotbeta, float l
...
-      xrmsc2m_[i] = (1.f - yxratio) * ((1.f - xxratio) * thePixelTemp_[index_id_].entx[iylow][ilow].xrmsc2m[i] +
-                                       xxratio * thePixelTemp_[index_id_].entx[iylow][ihigh].xrmsc2m[i]) +
-                    yxratio * ((1.f - xxratio) * thePixelTemp_[index_id_].entx[iyhigh][ilow].xrmsc2m[i] +
-                               xxratio * thePixelTemp_[index_id_].entx[iyhigh][ihigh].xrmsc2m[i]);
-
-      //	      chi2xavgc2m_[i]=(1.f - yxratio)*((1.f - xxratio)*thePixelTemp_[index_id_].entx[iylow][ilow].chi2xavgc2m[i] + xxratio*thePixelTemp_[index_id_].entx[iylow][ihigh].chi2xavgc2m[i])
-      //		          +yxratio*((1.f - xxratio)*thePixelTemp_[index_id_].entx[iyhigh][ilow].chi2xavgc2m[i] + xxratio*thePixelTemp_[index_id_].entx[iyhigh][ihigh].chi2xavgc2m[i]);
+      xrmsc2m_[i] = (1.f - yxratio) * ((1.f - xxratio) * entx00_->xrmsc2m[i] + xxratio * entx02_->xrmsc2m[i]) +
+                    yxratio * ((1.f - xxratio) * entx20_->xrmsc2m[i] + xxratio * entx22_->xrmsc2m[i]);
+      //	      chi2xavgc2m_[i]=(1.f - yxratio)*((1.f - xxratio)*entx00_->chi2xavgc2m[i] + xxratio*entx02_->chi2xavgc2m[i])
```
perrotta: Is this commented out code needed: add one line explaining its usage, if so; otherwise just remove

CondFormats/SiPixelTransient/src/SiPixelTemplate.cc
```diff
@@ -1842,30 +1809,24 @@ bool SiPixelTemplate::interpolate(int id, float cotalpha, float cotbeta, float l
...
-                           xxratio * thePixelTemp_[index_id_].entx[iyhigh][ihigh].fracxtwo);
+    fracxone_ = (1.f - yxratio) * ((1.f - xxratio) * entx00_->fracxone + xxratio * entx02_->fracxone) +
+                yxratio * ((1.f - xxratio) * entx20_->fracxone + xxratio * entx22_->fracxone);
+    fracxtwo_ = (1.f - yxratio) * ((1.f - xxratio) * entx00_->fracxtwo + xxratio * entx02_->fracxtwo) +
+                yxratio * ((1.f - xxratio) * entx20_->fracxtwo + xxratio * entx22_->fracxtwo);
 
     //       for(i=0; i<10; ++i) {
-    //	      pxspare[i]=(1.f - yxratio)*((1.f - xxratio)*thePixelTemp_[index_id_].entx[iylow][ilow].xspare[i] + xxratio*thePixelTemp_[index_id_].entx[iylow][ihigh].xspare[i])
-    //		          +yxratio*((1.f - xxratio)*thePixelTemp_[index_id_].entx[iyhigh][ilow].xspare[i] + xxratio*thePixelTemp_[index_id_].entx[iyhigh][ihigh].xspare[i]);
+    //	      pxspare[i]=(1.f - yxratio)*((1.f - xxratio)*entx00_->xspare[i] + xxratio*entx02_->xspare[i])
```
perrotta: Is this commented out code needed: add one line explaining its usage, if so; otherwise just remove

CondFormats/SiPixelTransient/src/SiPixelTemplate.cc
```diff
@@ -1884,14 +1845,12 @@ bool SiPixelTemplate::interpolate(int id, float cotalpha, float cotbeta, float l
       for (j = 0; j < TXSIZE; ++j) {
         //  Take next largest x-slice for the x-template (it reduces bias in the forward direction after irradiation)
         //		   xtemp_[i][j+2]=(1.f - xxratio)*thePixelTemp_[index_id_].entx[imaxx][ilow].xtemp[i][j] + xxratio*thePixelTemp_[index_id_].entx[imaxx][ihigh].xtemp[i][j];
-        //		   xtemp_[i][j+2]=(1.f - xxratio)*thePixelTemp_[index_id_].entx[iyhigh][ilow].xtemp[i][j] + xxratio*thePixelTemp_[index_id_].entx[iyhigh][ihigh].xtemp[i][j];
-        if (flip_x) {
+        //		   xtemp_[i][j+2]=(1.f - xxratio)*entx20_->xtemp[i][j] + xxratio*entx22_->xtemp[i][j];
```
perrotta: Is this commented out code needed: add one line explaining its usage, if so; otherwise just remove

CondFormats/SiPixelTransient/src/SiPixelTemplate.cc
```diff
@@ -1950,6 +1909,207 @@ bool SiPixelTemplate::interpolate(int id, float cotalpha, float cotbeta, float l
...
+  zsize_ = zsize;
+  s50_ = q50;
+  qscale_ = 1.f;
+  for (int i = 0; i < 3; ++i) {
+    fbin_[i] = fbin[i];
+  }
+
+  // Set other class variables
+
+  // adcota_ = 0.f;
```
perrotta: These two can be removed, as no used anywhere

CondFormats/SiPixelTransient/src/SiPixelTemplate.cc
```diff
@@ -1950,6 +1909,207 @@ bool SiPixelTemplate::interpolate(int id, float cotalpha, float cotbeta, float l
...
+      }
+      if (locBx < 0.f) {
+        flip_y_ = true;
+      }
+      break;
+    default:
+      // #ifndef SI_PIXEL_TEMPLATE_STANDALONE
+      //         throw cms::Exception("DataCorrupt") << "SiPixelTemplate::illegal subdetector ID = " << Dtype_ << std::endl;
+      //#else
+      std::cout << "SiPixelTemplate:illegal subdetector ID = " << Dtype_ << std::endl;
```
perrotta: This is only used standalone, isn't it?

CondFormats/SiPixelTransient/src/SiPixelTemplate.cc
```diff
@@ -2987,22 +3145,22 @@ int SiPixelTemplate::qbin(int id,
```
perrotta: ```ilow``` and ```ihigh``` as assigned in previous lines 3142 and 3143 are never used, and that assignment can be removed (see static analyzer output)

Same for the assignment of the same two variables at line 3631 and 3815

RecoLocalTracker/SiPixelRecHits/src/SiPixelTemplateReco.cc
```diff
@@ -175,19 +174,26 @@ int SiPixelTemplateReco::PixelTempReco1D(int id,
...
   // First, interpolate the template needed to analyze this cluster
   // check to see of the track direction is in the physical range of the loaded template
 
-  if (!templ.interpolate(id, cotalpha, cotbeta, locBz, locBx)) {
-    if (theVerboseLevel > 2) {
-      LOGDEBUG("SiPixelTemplateReco") << "input cluster direction cot(alpha) = " << cotalpha
-                                      << ", cot(beta) = " << cotbeta << ", local B_z = " << locBz
-                                      << ", local B_x = " << locBx << ", template ID = " << id
-                                      << ", no reconstruction performed" << ENDL;
+  if (id > 0) {  //if id == 0 bypass interpolation (used in calibration)
```
perrotta: Can this be originating the differences in the FastSim? (just guessing....)

CondFormats/SiPixelTransient/interface/SiPixelTemplate.h
OzAmram: Ok I will change

CondFormats/SiPixelTransient/src/SiPixelUtils.cc
OzAmram: Sorry I am not sure I understand, you are saying it would be preferable to have the implementation of generic_position_formula in interface/SiPixelUtils.h removing the need for src/SiPixelUtils.cc ? I thought it was desired to have function specifications and only simple functions actually implemented in the header file. Or am I mis-understanding?

CondFormats/SiPixelTransient/src/SiPixelTemplate.cc
OzAmram: Removed

CondFormats/SiPixelTransient/src/SiPixelTemplate.cc
OzAmram: Added comment

CondFormats/SiPixelTransient/src/SiPixelTemplate.cc
OzAmram: removed

CondFormats/SiPixelTransient/src/SiPixelTemplate.cc
OzAmram: added comment

CondFormats/SiPixelTransient/src/SiPixelTemplate.cc
OzAmram: I left this one, I think the comment above explains its usage

CondFormats/SiPixelTransient/src/SiPixelTemplate.cc
OzAmram: removed

CondFormats/SiPixelTransient/src/SiPixelTemplate.cc
OzAmram: Yes this is only used in standalone.  the commented out lines were leftovers, I removed them

CondFormats/SiPixelTransient/src/SiPixelTemplate.cc
OzAmram: Ok I removed them

RecoLocalTracker/SiPixelRecHits/src/SiPixelTemplateReco.cc
OzAmram: I don't think fastsim uses the template reco. I think the differences stem from the calls made [here](https://github.com/cms-sw/cmssw/blob/master/FastSimulation/TrackingRecHitProducer/src/PixelTemplateSmearerBase.cc). I'm trying to understand now

pmaksim1: Possibly.

I want to comment on FastSim stuff: we are actually responsible for the FastSim code.  We looked at the diffs, and they are very small.

That being said, the improved template calibration infrastructure is actually going to help FastSim since we can merge the code that makes the FastSim pixel rechit resolution files.  This is vital especially for HL-LHC, since the current FastSim resolution files there have the wrong electronics response (which is included in this PR...)

pmaksim1: FastSim does *not* use the template reco, but it is querying the template object to figure things out that are otherwise not simulated (e.g. the typical dimensions of the cluster and such).

OzAmram:
Ok I just pushed some new commits that address all of the comments & static analyzer issues.
I also think I understand the changes in the fastsim workflow now. There are two functions used only in the fastsim workflow, qbin_dist and temperrors, that re-did the interpolation between templates even though interpolate is required to have been called before them for proper usage. The issue is that they don't do the coordinate system flipping that interpolate does (see the commented out lines [here](https://github.com/cms-sw/cmssw/blob/master/CondFormats/SiPixelTransient/src/SiPixelTemplate.cc#L3610-L3622) and [here](https://github.com/cms-sw/cmssw/blob/master/CondFormats/SiPixelTransient/src/SiPixelTemplate.cc#L3426-L3437)). After talking to @pmaksim1 we agree behavior is not correct. So in the new version we just use the pointers and yratio that interpolate has already computed (and thus has the flipping correct) and don't redo it in these functions. I also did some other code cleanup and removed some unnecessary local versions of class variables. 

So now we do expect/understand the changes for fastsim and they are improvements/fixes. 
For regular we reco we still expect no changes.

CondFormats/SiPixelTransient/src/SiPixelUtils.cc
perrotta: It was just a suggestion in case you wanted to avoid the ```#ifndef``` directive to specialize for the standalone case: it can work either way, anyhow

perrotta:
please test

perrotta:
+1
- Changes in the pixel calibration procedure were implemented as described
- They are mostly technical for what concerns the reconstruction related part of the code
- Jenkins tests pass and show no differences in the reco outputs
- There are changes in the FastSim output, which look like having been understood as consequential to a more correct handling of the interpolation between templates which is allowed by this PR: the changes observed in the FastSim outputs, while rather small in size, must therefore be considered as improvements with respect to the current "bugged" approach.

perrotta:
@ssekmen, @lveldere, @civanch, @mdhildreth, @sbein you may possibly want to have a look, see the explanation provided [here](https://github.com/cms-sw/cmssw/pull/28945#issuecomment-589815532)

silviodonato:
@ggovi Do you have any comments?

ggovi:
+1

silviodonato:
+1

PR #28946, opened 2020-02-13
Migrate legacy modules in GenFilters to global

wddgit:
#### PR description:

Migrate 4 legacy modules in GeneratorInterface/GenFilters
to be global filters to improve multi threaded performance.

BCToEFilter
FourLepFilter
LHEGenericFilter
LHEVpTFilter

Includes some cleanup. (This is part 1, more of these PRs are coming
and they will repeat the patterns in this one. If there is something you
do not like, please let me know.)

#### PR validation:

Relies on existing tests. Output should not change, except that some
output to log files containing a count of events and events passed
is deleted.


wddgit:
please test

wddgit:
please test

wddgit:
please test

rebased plus updated to be more careful about std::abs usage. I'm not sure that it was broken either by this PR or before, but make sure the needed headers are there and it's always looking in std:: for the overloads with no dependence on includes from includes.

silviodonato:
@alberto-sanchez  @agrohsje @efeyazgan @mkirsano @qliphy @SiewYan
do you have any comments?

agrohsje:
Looks good to me. Just one question. The old code kept track of the efficiency. Not sure this is used by someone, but I couldn't find the same info in the updated code. Is this still available?

wddgit:
The Framework provides something like this for free. Here is an example. Run
```
FWCore/​Framework/​test/​testBitsCount_cfg.py
```
Here are selected lines from the output log file:
```
...
TrigReport ---------- Module Summary ------------
TrigReport    Visited   Executed     Passed     Failed      Error Name
...
TrigReport         99         99          3         96          0 f1
TrigReport         99         99          1         98          0 f2
TrigReport         99         99          8         91          0 f3
TrigReport         99         99         30         69          0 f4
TrigReport         99         99         70         29          0 f5
TrigReport         99         99         12         87          0 f6
...
```
where f1, f2, f3, f4, f5, and f6 are the module labels of some
filters in the test. If you look in the configuration, this is the
part that turns it on.

```
    wantSummary = cms.untracked.bool(True)
)
```
The only thing that is missing is the efficiency itself which
can easily be calculated from the passed and executed columns.
The only other weaknesses of this compared to what was removed
from the modules is that this enables a significant amount of printout
in the log file and you have find your module label, plus you have
to remember to enable it in the configuration.

The downside of the counters that were in the modules is
that they are not const. In a global module, simple "int" counters
would not work. One could make them std::atomic's
and make that work, but that seems like a lot of overhead
for something the Framework provides for free. You are adding
some performance degradation when you add an atomic because
it has to be synchronized as the job runs. Also it clutters the code.
If you really think these are important, I will put them back as atomics
(assuming no one else objects).  But I think the better solution is
just to remove them.

I'll also note some modules in GenFilters have these counters and some
don't and the printout is inconsistent, so it is not implemented in
a uniform way in the package...

agrohsje:
Thanks @wddgit .
+1

wddgit:
@agrohsje Are you the generators official approver? For some reason, it says it still needs approval?  Maybe the +1 needs to be on the first line? Thanks.

qliphy:
+1

silviodonato:
+1

PR #28948, opened 2020-02-13
More flexible SOI handling

igv4321:
Added more flexible handling of SOI (sample of interest) so that the case of SOI=5 will be correctly handled. This is expected due to the impending change of SiPM dark current modelling in MC, where placement of SOI at a later time slice allows for skipping two initial time slices at which the dark current is not well modelled (as there is not enough history).

This change is not supposed to modify any test results.

The usual "runTheMatrix" sequence was run to validate.

### DIFF
--- a/RecoLocalCalo/HcalRecProducers/src/HBHEPhase1Reconstructor.cc
+++ b/RecoLocalCalo/HcalRecProducers/src/HBHEPhase1Reconstructor.cc
@@ -465,12 +465,45 @@ void HBHEPhase1Reconstructor::processData(const Collection& coll,
     const RawChargeFromSample<DFrame> rcfs(sipmQTSShift_, sipmQNTStoSum_, cond, cell, cs, soi, frame, maxTS);
     int soiCapid = 4;
 
-    // Use only 8 TSs when there are 10 TSs
-    const int shiftOneTS = use8ts_ && maxTS == static_cast<int>(HBHEChannelInfo::MAXSAMPLES) ? 1 : 0;
-    const int nCycles = maxTS - shiftOneTS;
+    // Typical expected cases:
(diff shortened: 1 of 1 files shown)


slava77:
@cmsbuild please test

abdoulline:
NB: during Run1 we have had several HCAL modules with "precession" (unstability) of the clock synchronization wrt LHC clock. So that their SOI was moving back and force in TS range=(3,5). 
We did discuss this in the past in several PRs...
And once this PR add a special case with SOI=5 for our coming (soon) modification of MC Data frame size (with SOI=5) for >=Run3, it's "sensitive" to this Run1 problem.
I suppose it's not a showstopper.

slava77:
+1

for https://github.com/cms-sw/cmssw/pull/28948	12f899b 
- code changes are in line with the PR description. The sample/range selection looks more appropriate now even for the drifting channels from Run 1.
- jenkins tests pass and comparisons with the baseline show differences only in Run 1 data tests, which are affected by the drifting channels mentioned in https://github.com/cms-sw/cmssw/pull/28948#issuecomment-585990138

silviodonato:
>  And once this PR adds a special case with SOI=5 for our coming (soon) modification of MC Data frame size (with SOI=5) for >=Run3, it's "sensitive" to this old Run1 problem.

Could you give more details on this? Merging this PR, will we get a better or worse data(run1)/MC agreement?

abdoulline:
@silviodonato 
Flexibility in general sounds better :) 
Now, in particular (for this Run1 issue)  - when in the data SOI happens to arrive to TS=5, with this PR it can be handled, which wasn't the case before. 
So, with this PR it becomes better, albeit it doesn't fully cure the Run1  issue, as the timing alignment was seamless/gradual (rather constant "precession" than quantized jumping between TS' 3,4,5)

silviodonato:
+1
thanks

PR #28484, opened 2019-11-27
HCALDQM: Run3 era, add compatibility for new uTCA HO (master)

DryRun:
#### PR description:
Some mandatory updates for HCAL DQM to run on the upgraded detector (HB and HO). 
- Move online DQM cfgs to use Run3 era.
- More importantly, add crate-FED mapping for HO. Without this, the new HO electronics channels correspond to unknown crates, and HCAL DQM crashes.

#### PR validation:

Ran `runTheMatrix.py -l limited -i all --ibeos` with no problems. Ran online DQM over run 328676.

PR #28952, opened 2020-02-13
HCALDQM: Fix ngHO crate-FED mapping (master)

DryRun:
#### PR description:
This PR fixes the ngHO crate-FED mapping implemented in https://github.com/cms-sw/cmssw/pull/28484. The mapping changed a bit since that PR, so HCAL is crashing the DQM playback again. There was some further work done, and crate 38 ended up being used, instead of 33. See that thread for more info on the expected changes.

This probably has cropped up a few times now, and might happen again in the future. We plan to move away from the hard-coded crate-FED map. A couple ideas:

1. Change from FED to something like RBX. HCAL ops doesn't like FED anyways, since it doesn't correspond to anything physical and is subject to change in the future (depending on the zero-suppression strategy at higher radiation damage, HCAL may move to 3 FEDs / crate). 
2. Make FED available through conditions. Seems like a lot of work just to have FED on a plot axis.

#### PR validation:
Fixes the crash in playback. Passed `runTheMatrix.py -l limited -i all --ibeos` without error.


srimanob:
Does crash in playback (without this PR) can be seen with any relval workflows? Or we just need to test in DQM environment only?

jfernan2:
@srimanob I am afraid the crash can only be seen in Online DQM, otherwise master would be already afected

jfernan2:
please test
Even if backport #28955 has already been tested in Online DQM playback

jfernan2:
+1
Changes expected due to new crate numbering

silviodonato:
+1

jfernan2:
+1
Tested in Online DQM playback

PR #28955, opened 2020-02-13
HCALDQM: Fix ngHO crate-FED mapping (11_0_X) 

DryRun:
Backport of https://github.com/cms-sw/cmssw/pull/28952

#### PR description:
This PR fixes the ngHO crate-FED mapping implemented in https://github.com/cms-sw/cmssw/pull/28484. The mapping changed a bit since that PR, so HCAL is crashing the DQM playback again. There was some further work done, and crate 38 ended up being used, instead of 33. See that thread for more info on the expected changes.

This probably has cropped up a few times now, and might happen again in the future. We plan to move away from the hard-coded crate-FED map. A couple ideas:

1. Change from FED to something like RBX. HCAL ops doesn't like FED anyways, since it doesn't correspond to anything physical and is subject to change in the future (depending on the zero-suppression strategy at higher radiation damage, HCAL may move to 3 FEDs / crate). 
2. Make FED available through conditions. Seems like a lot of work just to have FED on a plot axis.

#### PR validation:
Fixes the crash in playback. Passed `runTheMatrix.py -l limited -i all --ibeos` without error.


fioriNTU:
@DryRun  thanks for this PR! Will be soon deployed in playback.

jfernan2:
+1
Tested already in Online DQM playback

silviodonato:
+1

PR #28960, opened 2020-02-14
Slightly refactor PF endcap cluster calibration

guitargeek:
#### PR description:

In the PFEgammaAlgo, there is still this comment about code "stolen" from the PFECALSuperClusterAlgo:
https://github.com/cms-sw/cmssw/blob/master/RecoParticleFlow/PFProducer/src/PFEGammaAlgo.cc#L1774

I was thinking of ways to avoid this duplication, but it's not quite trivial because the code diverged a bit. However, a bit part of the duplicate code serves to compute cluster energy corrections, so I could move this into the PFEnergyCalibration class which is used by both algos. I think this is a good step to avoid this code duplication, but I don't want to go further for now to not risk accidentally changing the logic. What do you think?

#### PR validation:

No backport intended.

### DIFF
--- a/RecoEcal/EgammaClusterAlgos/src/PFECALSuperClusterAlgo.cc
+++ b/RecoEcal/EgammaClusterAlgos/src/PFECALSuperClusterAlgo.cc
@@ -320,7 +320,7 @@ void PFECALSuperClusterAlgo::buildSuperCluster(CalibClusterPtr& seed, CalibClust
       energyweighttot(0);
   for (auto& clus : clustered) {
     double ePS1 = 0.0;
-    double ePS2 = 0;
+    double ePS2 = 0.0;
     energyweight = clus->energy_nocalib();
     bare_ptrs.push_back(clus->the_ptr().get());
     // update EE calibrated super cluster energies
--- a/RecoEcal/EgammaClusterAlgos/src/PFECALSuperClusterAlgo.cc
+++ b/RecoEcal/EgammaClusterAlgos/src/PFECALSuperClusterAlgo.cc
@@ -325,52 +325,16 @@ void PFECALSuperClusterAlgo::buildSuperCluster(CalibClusterPtr& seed, CalibClust
     bare_ptrs.push_back(clus->the_ptr().get());
     // update EE calibrated super cluster energies
     if (isEE) {
-      double ps1_energy_sum = 0.;
-      double ps2_energy_sum = 0.;
-      ePS1 = 0.;
-      ePS2 = 0.;
-      int condP1 = 1;
-      int condP2 = 1;
       auto ee_key_val = std::make_pair(clus->the_ptr().key(), edm::Ptr<reco::PFCluster>());
       const auto clustops = std::equal_range(EEtoPS_->begin(), EEtoPS_->end(), ee_key_val, sortByKey);
+      std::vector<reco::PFCluster const*> psClusterPointers;
       for (auto i_ps = clustops.first; i_ps != clustops.second; ++i_ps) {
-        edm::Ptr<reco::PFCluster> psclus(i_ps->second);
-
-        auto const& recH_Frac = psclus->recHitFractions();
-
-        switch (psclus->layer()) {
-          case PFLayer::PS1:
-            ps1_energy_sum += psclus->energy();
-            for (auto const& recH : recH_Frac) {
-              ESDetId strip1 = recH.recHitRef()->detId();
-              if (strip1 != ESDetId(0)) {
-                // getStatusCode() == 1 => dead channel
-                //apply correction if all recHits in dead region
-                if (channelStatus_->getMap().find(strip1)->getStatusCode() == 0)
-                  condP1 = 0;  //active
-              }
-            }
-            break;
-          case PFLayer::PS2:
-            ps2_energy_sum += psclus->energy();
-            for (auto const& recH : recH_Frac) {
-              ESDetId strip2 = recH.recHitRef()->detId();
-              if (strip2 != ESDetId(0)) {
-                if (channelStatus_->getMap().find(strip2)->getStatusCode() == 0)
-                  condP2 = 0;
-              }
-            }
-            break;
-          default:
-            break;
-        }
+        psClusterPointers.push_back(i_ps->second.get());
       }
-      if (condP1 == 1)
-        ePS1 = -1.;
-      if (condP2 == 1)
-        ePS2 = -1.;
-      _pfEnergyCalibration->energyEm(
-          *(clus->the_ptr()), ps1_energy_sum, ps2_energy_sum, ePS1, ePS2, applyCrackCorrections_);
+      auto calibratedEnergies = _pfEnergyCalibration->calibrateEndcapClusterEnergies(
+          *(clus->the_ptr()), psClusterPointers, *channelStatus_, applyCrackCorrections_);
+      ePS1 = calibratedEnergies.ps1Energy;
+      ePS2 = calibratedEnergies.ps2Energy;
     }
 
     if (ePS1 == -1.)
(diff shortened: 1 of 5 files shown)


slava77:
@cmsbuild please test

silviodonato:
Do you have any comments? @perrotta @slava77

RecoParticleFlow/PFClusterTools/src/PFEnergyCalibration.cc
```diff
@@ -161,6 +161,59 @@ void PFEnergyCalibration::initializeCalibrationFunctions() {
...
+PFEnergyCalibration::CalibratedEndcapPFClusterEnergies PFEnergyCalibration::calibrateEndcapClusterEnergies(
+    reco::PFCluster const& eeCluster,
+    std::vector<reco::PFCluster const*> const& psClusterPointers,
+    ESChannelStatus const& channelStatus,
+    bool applyCrackCorrections) const {
+  double ps1_energy_sum = 0.;
+  double ps2_energy_sum = 0.;
+  double ePS1 = 0.;
+  double ePS2 = 0.;
+  int condP1 = 1;
```
perrotta: Maybe if you make these two some "bool" variable, everything gets more intuitive

RecoParticleFlow/PFClusterTools/src/PFEnergyCalibration.cc
```diff
@@ -161,6 +161,59 @@ void PFEnergyCalibration::initializeCalibrationFunctions() {
...
+    switch (psclus->layer()) {
+      case PFLayer::PS1:
+        ps1_energy_sum += psclus->energy();
+        for (auto const& recH : recH_Frac) {
+          ESDetId strip1 = recH.recHitRef()->detId();
+          if (strip1 != ESDetId(0)) {
+            //getStatusCode() == 0 => active channel
+            // apply correction if all recHits are dead
+            if (channelStatus.getMap().find(strip1)->getStatusCode() == 0)
+              condP1 = 0;
```
perrotta: you can break here

RecoParticleFlow/PFClusterTools/src/PFEnergyCalibration.cc
```diff
@@ -161,6 +161,59 @@ void PFEnergyCalibration::initializeCalibrationFunctions() {
...
+          }
+        }
+        break;
+      case PFLayer::PS2:
+        ps2_energy_sum += psclus->energy();
+        for (auto const& recH : recH_Frac) {
+          ESDetId strip2 = recH.recHitRef()->detId();
+          if (strip2 != ESDetId(0)) {
+            if (channelStatus.getMap().find(strip2)->getStatusCode() == 0)
+              condP2 = 0;
```
perrotta: you can break here

RecoEcal/EgammaClusterAlgos/src/PFECALSuperClusterAlgo.cc
```diff
@@ -316,62 +316,25 @@ void PFECALSuperClusterAlgo::buildSuperCluster(CalibClusterPtr& seed, CalibClust
...
   // calculate necessary parameters and build the SC
-  double posX(0), posY(0), posZ(0), corrSCEnergy(0), corrPS1Energy(0), corrPS2Energy(0), ePS1(0), ePS2(0),
-      energyweight(0), energyweighttot(0);
-  std::vector<double> ps1_energies, ps2_energies;
-  int condP1(1), condP2(1);
+  double posX(0), posY(0), posZ(0), corrSCEnergy(0), corrPS1Energy(0), corrPS2Energy(0), energyweight(0),
+      energyweighttot(0);
   for (auto& clus : clustered) {
-    ePS1 = ePS2 = 0;
+    double ePS1 = 0.0;
```
perrotta: (just for symmetry) you could write "0" for both them...

guitargeek:
Hi @perrotta, I addressed your comments and also got rid of an "temporaty ugly fix" in the PFEnergyCalibration class. Just reading the code of the PFCluster and CaloCluster classes, it seems equivalent to just use the `eta()` and `phi()` member functions without having to copy the cluster, but I guess the bot comparisons will tell if this is really correct.

perrotta:
please test

guitargeek:
Hi @perrotta, I made a mistake in my previous commit. Now I fixed that commit and also made sure the local tests pass. So the bot should succeed now!

perrotta:
please test

guitargeek:
Hi @perrotta, inspired by your previous comments that suggested to use `bool` and not `double` to encode whether the PS planes should be used for the corrections or not, I propagated this change now through all the functions of the class. That makes it more understandable, as using the same variables to declare the PS plane status and store the output was confusing.

perrotta:
please test

perrotta:
> * Reco comparison results: 1530 differences found in the comparisons

The latest commit wasn't that innocuos... We must check where those differences come from.
Perhaps, factorizing the updates whenever possible could be a better option: it would at least be easier to review

guitargeek:
Yes that's right, it would have been nice if it worked but I tried to do too much at once at this PR. I will amend the last commit and only keep the changes that are hopefully really innocent, and then maybe we can do the rest at some other time.

guitargeek:
Hi @perrotta, now it's ready to be tested again. I only kept the change with the other overloaded version of the `energyEm` function in the `ConvBremPFTrackFinder` function, so that the interface of `PFEnergyCalibration` needs fewer public functions. This should not change the result.

perrotta:
@cmsbuild please test

perrotta:
+1
- It allows some refactorization of the PF energy calibration 
- Jenkins tests pass and show no difference with the current reco outputs, as it should

silviodonato:
+1

slava77:
[I'm not sure if this is the PR with the problem, but the error messages appear to point to this]

@smuzaffar @mrodozov 
please remind me how to test on PPC interactively.
I suspect that the problem came from this PR, based on 140.52 working in CMSSW_11_1_X_2020-03-08-2300 but not in CMSSW_11_1_X_2020-03-09-2300.
In the latter we have
```
   [0] Processing  stream begin Run run: 152698 stream: 0
   [1] Calling method for module PFProducer/'particleFlowTmp'
   [2] Using EventSetup component PoolDBESSource/'GlobalTag' to make data PerformancePayload/'' in record PFCalibrationRcd
Exception Message:
A std::exception was thrown.
getrandom
```

https://cmssdt.cern.ch/SDT/cgi-bin/buildlogs/slc7_ppc64le_gcc820/CMSSW_11_1_X_2020-03-08-2300/pyRelValMatrixLogs/run/140.52_RunHI2010+RunHI2010+RECOHID10+RECOHIR10D11+HARVESTDHI/step2_RunHI2010+RunHI2010+RECOHID10+RECOHIR10D11+HARVESTDHI.log
vs
https://cmssdt.cern.ch/SDT/cgi-bin/buildlogs/slc7_ppc64le_gcc820/CMSSW_11_1_X_2020-03-09-2300/pyRelValMatrixLogs/run/140.52_RunHI2010+RunHI2010+RECOHID10+RECOHIR10D11+HARVESTDHI/step2_RunHI2010+RunHI2010+RECOHID10+RECOHIR10D11+HARVESTDHI.log

slava77:
ping

https://github.com/cms-sw/cmssw/pull/28960#issuecomment-597599381 
> @smuzaffar @mrodozov
please remind me how to test on PPC interactively.

smuzaffar:
@slava77 send me your ssh public key

slava77:
I managed to get to the error in PerformancePayload.
It is apparently unrelated to this PR.
The same IB where the problem appeared apparently also has https://github.com/cms-sw/cmsdist/pull/5573
the gdb stack trace points to a problem in boost and simply merging this PR on top of 08-2300 IB does not introduce any problems.

PR #28951, opened 2020-02-13
[DD4hep] DDNamespace::name Issue #28949

ianna:
#### PR description:

  * Use ```DDNamespace::prepend(name)``` instead of ```DDNamespace::name()``` in all ```std::string``` operations. 
  * Remove unused ```std::string```s

Addresses issue #28949 and replaces PR #28933

#### if this PR is a backport please specify the original PR and why you need to backport that PR:

Before submitting your pull requests, make sure you followed this checklist:
- verify that the PR is really intended for the chosen branch
- verify that changes follow [CMS Naming, Coding, And Style Rules](http://cms-sw.github.io/cms_coding_rules.html)
- verify that the PR passes the basic test procedure suggested in the [CMSSW PR instructions](https://cms-sw.github.io/PRWorkflow.html)

PR #28965, opened 2020-02-14
[Do not merge ] Copy of #28951 to test with ROOT

mrodozov:
#### PR description:

Purpose of this PR is to test if ROOT and DD4Hep build without poluting 28951 signatures

#### PR validation:

No validation, lets see how it goes


cvuosalo:
+1

PR #28976, opened 2020-02-17
[HGCal] Add a lower limit to the fraction of recHits energy in realistic SimClusters

felicepantaleo:
#### PR description:

When a realistic simcluster becomes invisible, all its exclusive rechits are redistributed to every visible simcluster that has at least one hit in common with it.
In pathological events, this could lead to an increase of hits per realistic simcluster, with fractions below 1e-10, making the loop much longer. 
Changes expected: 

- decrease in the running time for pathological events
- decrease in the number of rechits per PFCluster with negligible change in energy

#### PR validation:
I have executed `step3` of wf `23203.0` on one pathological event provided by the HLT upgrade convener @trtomei

### DIFF
--- a/RecoParticleFlow/PFClusterProducer/plugins/SimMappers/RealisticHitToClusterAssociator.h
+++ b/RecoParticleFlow/PFClusterProducer/plugins/SimMappers/RealisticHitToClusterAssociator.h
@@ -212,13 +212,16 @@ public:
               // hits that belonged completely to the absorbed cluster are redistributed
               // based on the fraction of energy shared in the shared hits
               float sharedFraction = pair.second / totalSharedEnergy;
-              float assignedEnergy = realisticHit.totalEnergy_ * sharedFraction;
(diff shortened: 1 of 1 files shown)


felicepantaleo:
@rovere fyi

felicepantaleo:
@cmsbuild please test

slava77:
@cmsbuild please test

slava77:
+1

for https://github.com/cms-sw/cmssw/pull/28976	e68e48a 
- jenkins tests pass and comparisons with the baseline show no differences (the conditions addressed by this PR require some larger occupancy to show any difference).
- tested on 200 events with PU 300 D49 setup, running RECO:hgcalLocalRecoSequence in CMSSW_11_1_X_2020-02-19-1100: the CPU time in particleFlowClusterHGCal went down from an average of 14 s/ev with a variance of 59 s (max 756 s/ev) to an average of 2.1 s/ev with a variance of just 0.5 s.

silviodonato:
+1

PR #28986, opened 2020-02-18
DQM: Re-add harvesting test tools.

schneiml:
#### PR description:

These got accidentally deleted in the big DQM test cleanup (#28612), originally added in #28002.

The files are not used for any automated tests (primarily because we don't have any data reliably available), but can be useful to debug issues on real data. Downloading the DQMIO files before running the job should no longer be required (the input source now supports opening remote files), but it still helps to remove some unknowns from the test setup.

#### PR validation:

No changes to existing workflows. `dqmiofilecopy.py` still works to download DQMIO files when supplied with a dataset currently on disk.

### DIFF
--- a/DQMServices/Components/test/dqmiofilecopy.sh
+++ b/DQMServices/Components/test/dqmiofilecopy.sh
@@ -1,0 +1,27 @@
+#!/bin/bash
+# like '/SingleElectron/Run2017D-09Aug2019_UL2017-v1/DQMIO'
+DATASET=$1
+# like 302663
+RUN=$2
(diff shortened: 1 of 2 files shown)


andrius-k:
please test

silviodonato:
@andrius-k @jfernan2 @kmaeshima @fioriNTU @schneiml
do you have any comments on this?

schneiml:
+1

Looks still good to me, and used to be in CMSSW before.

silviodonato:
+1

PR #28991, opened 2020-02-18
Migrate 10 EDFilters in GenFilters from legacy

wddgit:
#### PR description:

10 more legacy EDFilters in the GenFilters package
migrated from being legacy types to global or one
types. This is the second in a series of such PRs
(more are on the way).

There is a considerable amount of cleanup mixed in with
this. Headers fixed, commented code removed, cout lines
removed, event counters removed, .h content moved to
.cc files,  destructors just defaulted, esConsumes migration
done for one module...

The intent is nothing has been done that would change
the filter behavior and in fact the changes to the
algorithmic code in the filter methods are very small.
I didn't touch any complicated logical code.

#### PR validation:

Relying on existing tests. Behavior should not change.


wddgit:
please test

fyi @Dr15Jones @makortel

wddgit:
please test

Force pushed a commit that fixed the problem clang noticed. It was related to the std::abs function. Mostly the fix was that I restored "using namespace std" everywhere I had deleted it and also put a few other related things back the way they were before. A few less cleanup changes now.

silviodonato:
@alberto-sanchez @agrohsje @efeyazgan @mkirsano @qliphy @SiewYan
do you have any comments?

agrohsje:
+1

silviodonato:
+1

PR #28997, opened 2020-02-19
MTD RECO geometry: MTDNumberingBuilder code simplification

fabiocos:
#### PR description:

The existing MTD RECO geometry is based on a clone and minimal adjustment of the existing tracker geometry performed mostly by @lgray in view of the MTD TDR release. The need to 1) add the new ETL design and 2) to migrate this code to DD4hep strengthen the need for a code simplification and refresh, already emerged in the 2018 review and in more recent discussions with @ianna.

This PR is the first step in this direction: the aim is to preserve the existing geometry description, to be tested in detail through the refreshed test classes in ```Geometry/MTDNumberingBuilder``` and ```Geometry/MTDGeometryBuilder```, while moving from a layered build of the ```GeometricTimingDet``` structure from DDD to a much simpler overall loop, where only the subdetectors (BTL and ETL, two sides) and the final sensitive modules are kept (and appear necessary).
Several intermediate classes are removed, with no apparent impact on the final geometry structure produced in ```MTDGeometryBuilder```. 

The following step is the modification of the ```DDDCmsMTDConstruction``` class (and auxiliary code) to include the new geometry structure of ETL as implemented by @icosivi at xml level.

#### PR validation:

The dumps of the geometries, provided the same version of the test analyzers in ```mtd_cfg.py``` for the benchmark scenario D50, differ only  in ```GeometricTimingDetAnalyzer``` for the fix of the type of modules performed in GeometricTimingDet (but apparently not relevant in the final construction of the geometry):

```
173c100
< Module = BModule33Layer1NegativeZTimingwafer type = 4 rawId = 1652669440 Sub/side/RR = 1 0 1
---
> Module = BModule33Layer1NegativeZTimingwafer type = 100 rawId = 1652669440 Sub/side/RR = 1 0 1
```

```
79638c72655
< Module = EModule1Disc1Timingwafer type = 10 rawId = 1665204352 Sub/side/RR = 2 1 1
---
> Module = EModule1Disc1Timingwafer type = 100 rawId = 1665204352 Sub/side/RR = 2 1 1
```


fabiocos:
please test

cvuosalo:
+1

fabiocos:
@kpedro88 comments?

fabiocos:
please test

the update cannot modify any test workflow as the test class is used only in standalone test configurations, not running as unit test

fabiocos:
The DQM differences cannot be related to the latest commit after previous test

Geometry/MTDNumberingBuilder/plugins/CmsMTDConstruction.cc
```diff
@@ -1,42 +1,31 @@
...
+
+bool CmsMTDConstruction::mtdOrderPhi(const GeometricTimingDet* a, const GeometricTimingDet* b) {
+  MTDDetId id1(a->geographicalId());
+  MTDDetId id2(b->geographicalId());
+  return (id1.mtdRR() == id2.mtdRR()) && (angle0to2pi::make0To2pi(a->phi()) < angle0to2pi::make0To2pi(b->phi()));
 }
 
 void CmsMTDConstruction::buildBTLModule(DDFilteredView& fv, GeometricTimingDet* mother, const std::string& attribute) {
-  GeometricTimingDet* det = new GeometricTimingDet(&fv, theCmsMTDStringToEnum.type(fv.name()));
+  GeometricTimingDet* det = new GeometricTimingDet(&fv, theCmsMTDStringToEnum.type(fv.name().substr(0, 7)));
```
kpedro88: magic number 7 should be named constant and reused below

Geometry/MTDNumberingBuilder/plugins/CmsMTDConstruction.cc
```diff
@@ -80,3 +68,23 @@ void CmsMTDConstruction::buildETLModule(DDFilteredView& fv, GeometricTimingDet*
 
   mother->addComponent(det);
 }
+
+GeometricTimingDet* CmsMTDConstruction::buildSubdet(DDFilteredView& fv,
+                                                    GeometricTimingDet* mother,
+                                                    const std::string& attribute) {
+  auto thisDet = theCmsMTDStringToEnum.type(fv.name());
+  GeometricTimingDet* subdet = new GeometricTimingDet(&fv, thisDet);
```
kpedro88: A better pattern would be for the mother volume to create the subdetector as a smart pointer that it already owns, and then return it to this code for configuration. This would avoid a bare `new `, which always makes me wary of possible memory leaks. This could be addressed in a followup PR, since I think this is a common issue throughout the geometry code.

Geometry/MTDNumberingBuilder/plugins/DDDCmsMTDConstruction.cc
```diff
@@ -63,9 +69,86 @@ std::unique_ptr<GeometricTimingDet> DDDCmsMTDConstruction::construct(const DDCom
...
+  }
+
+#ifdef EDM_ML_DEBUG
+  comp.clear();
+  comp = mtd->deepComponents();
+  std::stringstream after(std::stringstream::in | std::stringstream::out);
+  for (const auto& it : comp) {
+    after << "ORDER2 " << it->geographicalId().rawId() << " " << it->type() << " " << it->translation().z() << "\n";
+  }
+  edm::LogVerbatim("MTDNumbering") << "GeometricTimingDet order before sorting \n" << after.str();
```
kpedro88: should this message say "after sorting"?

fabiocos: of course, a leftover of many cut-and-paste during debugging additions and removals

Geometry/MTDNumberingBuilder/plugins/CmsMTDConstruction.cc
fabiocos: not the only one, but annoying enough to be dealt with, I agree

Geometry/MTDNumberingBuilder/plugins/CmsMTDConstruction.cc
fabiocos: it is a common pattern in the parallel tracker code, from which this code is derived, and this PR reduces the number of instances. I would prefer to defer this update to a later moment, as there are already too many things boiling in the plate at present (new ETL, DD4hep migration)

fabiocos:
please test

fabiocos:
hold

I need to verify the management of the number of layers. The dump of the MTDGeometry is identical, but the number of layers looks not shown, I am looking into this now that I am adding the new ETL.
@pmeridian @parbol with the two discs layout in principle we have multiple layers (4, 2 discs x 2 sides I would say).

fabiocos:
Now there is an additional node as layer that preserves the calculation in MTDGeometry, even if I believe this is not used so far, and for the old geometries in practices there is just one layer. This could change in the new ETL geometry. The only difference is in the detId of layers, which anyway for what I can see is fake and irrelevant.

fabiocos:
unhold

fabiocos:
please test

fabiocos:
@cvuosalo @kpedro88 @silviodonato for the time being, unless you have further comments, I consider this PR as stable, and defer further developments to the next PR I am preparing, essentially adapt this new structure to include the new ETL design (which has a different structure and requires some non-trivial code adjustment).

kpedro88:
+upgrade

cvuosalo:
+1

silviodonato:
please test workflow 23234.1001, 23434.1001

silviodonato:
+1

PR #29002, opened 2020-02-19
remove tuple copies

davidlange6:
perhaps saves 8% of a hl-lhc premixing job

http://mkortela.web.cern.ch/mkortela/cgi-bin/navigator/premix_d35_1040mtd5/test_06_perf/43

(probably the same issue is in the MTD geometry since they've copied all the tracker code by value)


davidlange6:
please test

cvuosalo:
+1

ianna:
@davidlange6 - Can we use a linter to determine the cases? There is a flag in clang which would give a warning if the copy is unavoidable and that sort of change would be futile.

silviodonato:
+1

davidlange6:
sure why not? This case was particularly simple - looping over member data inside of a const function.

On Feb 20, 2020, at 7:36 AM, Ianna Osborne <notifications@github.com<mailto:notifications@github.com>> wrote:


@davidlange6<https://github.com/davidlange6> - Can we use a linter to determine the cases? There is a flag in clang which would give a warning if the copy is unavoidable and that sort of change would be futile.

fabiocos:
@davidlange6 concerning your comment about MTD yes, the issue is partly there (I will fix it), but it looks to me that the affected functions are not used , it was a massive port which needs a cleaning, I am slowly going through it.

PR #29006, opened 2020-02-20
Pin the key function in a single object file.

vgvassilev:
The itanium ABI spec section 5.2.3 Virtual Tables defines a term "key function" as "first non-pure virtual function that is not inline at the point of class definition"

If there is no function the vtable is emitted in every object file. This in turn triggers redundant deserializations from the module files showing a bug in the incremental setup for DataFormats/Provenance.

This patch chooses the key function to be out of line, this the vtable to be pinned in a single object file.

Part of #15248

cc: @davidlange6, @oshadura.

### DIFF
--- a/DataFormats/Common/src/RefHolder.cc
+++ b/DataFormats/Common/src/RefHolder.cc
@@ -1,0 +1,8 @@
+#include "DataFormats/Common/interface/RefHolderBase.h"
+
+namespace edm {
+  namespace reftobase {
+    // Pin the vtable here.
+    RefHolderBase::RefHolderBase() {}
+  }  // namespace reftobase
+}  // namespace edm
--- a/DataFormats/Common/interface/RefHolderBase.h
+++ b/DataFormats/Common/interface/RefHolderBase.h
@@ -16,7 +16,7 @@ namespace edm {
 
     class RefHolderBase {
     public:
-      RefHolderBase() {}
+      RefHolderBase();
       template <class T>
       T const* getPtr() const;
       virtual ~RefHolderBase();


Dr15Jones:
I do not see how this change accomplished what you want.

vgvassilev:
Fair point. I overlooked `getPtr`. I will outline the constructor instead. Thanks!

davidlange6:
please test




On 21 Feb 2020, at 07:28, cmsbuild <notifications@github.com> wrote:

﻿

Pull request #29006<https://github.com/cms-sw/cmssw/pull/29006> was updated. @makortel<https://github.com/makortel>, @smuzaffar<https://github.com/smuzaffar>, @cmsbuild<https://github.com/cmsbuild>, @Dr15Jones<https://github.com/Dr15Jones> can you please check and sign again.

davidlange6:
please test




On 22 Feb 2020, at 09:18, cmsbuild <notifications@github.com> wrote:

﻿

Pull request #29006<https://github.com/cms-sw/cmssw/pull/29006> was updated. @makortel<https://github.com/makortel>, @smuzaffar<https://github.com/smuzaffar>, @cmsbuild<https://github.com/cmsbuild>, @Dr15Jones<https://github.com/Dr15Jones> can you please check and sign again.

DataFormats/Common/src/RefHolder.cc
```diff
@@ -0,0 +1,27 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
```
Dr15Jones: Please remove the copyright notice

Dr15Jones:
please test

DataFormats/Common/src/RefHolder.cc
```diff
@@ -0,0 +1,8 @@
+#include <DataFormats/Common/interface/RefHolderBase.h>
```
makortel:
```suggestion
#include "DataFormats/Common/interface/RefHolderBase.h"
```
We tend to use the quotation marks for `#include`s inside CMSSW.

vgvassilev: Quotes force the compiler to make an extra lookup in the current folder where we already know it will find nothing.

makortel: Point taken, but our policy is to use quotes for headers inside CMSSW. Please change.

makortel:
@cmsbuild, please test

DataFormats/Common/src/RefHolder.cc
makortel: Thanks!

makortel:
+1

silviodonato:
+1

PR #29007, opened 2020-02-20
[HGCal] backport 11_0_X Add a lower limit to the fraction of recHits energy in realistic SimClusters 

felicepantaleo:
#### PR description:
Backport of #28976
When a realistic simcluster becomes invisible, all its exclusive rechits are redistributed to every visible simcluster that has at least one hit in common with it.
In pathological events, this could lead to an increase of hits per realistic simcluster, with fractions below 1e-10, making the loop much longer. 
Changes expected: 

- decrease in the running time for pathological events
- decrease in the number of rechits per PFCluster with negligible change in energy

#### PR validation:
I have executed `step3` of wf `23203.0` on one pathological event provided by the HLT upgrade convener @trtomei

### DIFF
--- a/RecoParticleFlow/PFClusterProducer/plugins/SimMappers/RealisticHitToClusterAssociator.h
+++ b/RecoParticleFlow/PFClusterProducer/plugins/SimMappers/RealisticHitToClusterAssociator.h
@@ -212,13 +212,16 @@ public:
               // hits that belonged completely to the absorbed cluster are redistributed
               // based on the fraction of energy shared in the shared hits
               float sharedFraction = pair.second / totalSharedEnergy;
-              float assignedEnergy = realisticHit.totalEnergy_ * sharedFraction;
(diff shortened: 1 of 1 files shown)


felicepantaleo:
backport #28976

felicepantaleo:
@cmsbuild please test

slava77:
+1

for https://github.com/cms-sw/cmssw/pull/29007	267d0bb 
- backport is correct (matches the diffs in the master version)
- Technically, this PR does not comply with the no-change policy for production code. Since this fix is addressing issues in somewhat small fraction of events [or clusters], it seems appropriate to apply this backport in this production branch.

silviodonato:
@trtomei @fwyzard Do you think it is still useful to have this fix in CMSSW_11_0_X for the HLT TDR ?

fwyzard:
@silviodonato the plan is to move to 11.1.x and not to use the realistic simclusters any more...

But for the moment, yes, this should help with some pathological events.

silviodonato:
+1

PR #29025, opened 2020-02-22
Remove drop and update the RecoJets event content formats

jeongeun:
#### PR description:
Update event content definitions to explicitly have RECO to be a superset of AOD and for FEVT to be a superset of RECO : Defining the RecoJetsAOD PSets first and then RecoJetsRECO content format imports the RecoJetsAOD one and adds additional keep collections to reproduce the exact RECO event content. The FEVT is also done similarly. (example: [RecoMuon_EventContent_cff.py](https://github.com/cms-sw/cmssw/blob/CMSSW_11_1_X/RecoMuon/Configuration/python/RecoMuon_EventContent_cff.py)).

In RecoJetsAOD case,  we use "keep" instead of "drop" (drop doubles_\*Jets_sigmas_\*,  drop doubles_\*Jets_rhos_\*) 
 
#### PR validation:
Event Content comparison check was also done and there is no change with this updates.
Tested in CMSSW_11_1_X_2020-02-20-2300, the basic test all passed in the [CMSSW PR instructions](https://cms-sw.github.io/PRWorkflow.html).

### DIFF
--- a/RecoJets/Configuration/python/RecoJets_EventContent_cff.py
+++ b/RecoJets/Configuration/python/RecoJets_EventContent_cff.py
@@ -47,6 +47,44 @@ RecoGenJetsAOD = cms.PSet(
                                            'keep *_genParticle_*_*'
                                            )
 )
+from Configuration.Eras.Modifier_pA_2016_cff import pA_2016
+from Configuration.Eras.Modifier_peripheralPbPb_cff import peripheralPbPb
+from Configuration.Eras.Modifier_pp_on_XeXe_2017_cff import pp_on_XeXe_2017
+from Configuration.Eras.Modifier_pp_on_AA_2018_cff import pp_on_AA_2018
+#products from regular pp which does not fit the normal AOD
+for e in [pA_2016, peripheralPbPb, pp_on_XeXe_2017, pp_on_AA_2018]:
+    e.toModify( RecoJetsAOD.outputCommands, 
+                func=lambda outputCommands: outputCommands.extend(['keep *_towerMaker_*_*'])
+                )
+for e in [pp_on_XeXe_2017, pp_on_AA_2018]:
+    e.toModify( RecoJetsAOD.outputCommands,
+                func=lambda outputCommands: outputCommands.extend(['keep recoCentrality*_hiCentrality_*_*',
+                                                                   'keep recoClusterCompatibility*_hiClusterCompatibility_*_*'
+                                                                   ])
+                )
+#HI-specific products: needed in AOD, propagate to more inclusive tiers as well
+pA_2016.toModify( RecoJetsAOD.outputCommands, 
+                  func=lambda outputCommands: outputCommands.extend(['keep recoCentrality*_pACentrality_*_*',
+                                                                     'keep *_hiFJGridEmptyAreaCalculator_*_*',
+                                                                     'keep *_hiFJRhoProducer_*_*'
+                                                                     ])
+                )
+#HI-specific products: needed in AOD, propagate to more inclusive tiers as well
+peripheralPbPb.toModify( RecoJetsAOD.outputCommands, 
+                         func=lambda outputCommands: outputCommands.extend(['keep recoCentrality*_pACentrality_*_*'])
+                         )
+
+pp_on_AA_2018.toModify( RecoJetsAOD.outputCommands, 
+                        func=lambda outputCommands: outputCommands.extend(['keep *_hiCentrality_*_*',
+                                                                           'keep *_hiFJRhoProducer_*_*',
+                                                                           'keep *_akPu3PFJets_*_*',
+                                                                           'keep *_akPu4PFJets_*_*',
+                                                                           'keep *_kt4PFJetsForRho_*_*',
+                                                                           'keep *_akCs4PFJets_*_*',
+                                                                           'keep *_akPu4CaloJets_*_*',
+                                                                           'drop *_caloTowers_*_*'
+                                                                           ])
+                        )
 #RECO content
 RecoJetsRECO = cms.PSet(
     outputCommands = cms.untracked.vstring('keep *_ak4CaloJets_*_*',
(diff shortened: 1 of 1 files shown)


perrotta:
please test

perrotta:
The output of the Phase2 workflows automatically run during the jenkins tests shows that in the **reco** output the following products are not present any more: ```recoGenJets_slimmedGenJets```

Those Phase2 workflows use at step3 the FEVT event content (instead of the RECO one, as the step3 in other workflows):
```
    --eventcontent FEVTDEBUGHLT,MINIAODSIM,DQM
```
I assume this is something wanted (i.e using the FEVT event content instead of the RECO one), but I'd let @kpedro88 comment if not

Anyhow, in order to have them back the additional content must be included here in the FEVT content, i.e.
```
RecoGenJetsFEVT = cms.PSet(	
    outputCommands = cms.untracked.vstring('keep recoGenJets_*_*_*')	
)
```

@rappoccio @knash FYI

RecoJets/Configuration/python/RecoJets_EventContent_cff.py
```diff
@@ -121,19 +36,62 @@
...
+                                           'keep *_kt4PFJets_*_*', 
+                                           'keep *_kt6PFJets_*_*',
+                                           'keep *_ak5PFJets_*_*',
+                                           'keep *_ak7PFJets_*_*',
+                                           'keep *_kt4TrackJets_*_*',
+                                           'keep *_ca*Mass_*_*',
+                                           'keep *_ak*Mass_*_*'
+        )
+)
+RecoGenJetsFEVT = cms.PSet(
```
perrotta:
```
RecoGenJetsFEVT = cms.PSet(	
    outputCommands = cms.untracked.vstring('keep recoGenJets_*_*_*')	
)
```

RecoJets/Configuration/python/RecoJets_EventContent_cff.py
```diff
@@ -145,38 +103,35 @@
...
+                                                                           'keep *_hiFJRhoProducer_*_*',
+                                                                           'keep *_akPu3PFJets_*_*',
+                                                                           'keep *_akPu4PFJets_*_*',
+                                                                           'keep *_kt4PFJetsForRho_*_*',
+                                                                           'keep *_akCs4PFJets_*_*',
+                                                                           'keep *_akPu4CaloJets_*_*',
+                                                                           'drop *_caloTowers_*_*'
+                                                                           ])
+                        )
+RecoJetsRECO.outputCommands.extend(RecoJetsAOD.outputCommands)
```
perrotta: To improve the clarity, I would add these ```extend``` commands right after the definition of the corresponding output commands above, so that it will become more evident that the RecoEventContent (e.g.) is an extention of the AOD event content, and similarly the FEVT event content is an extention of the Reco one

kpedro88:
@perrotta ideally we would like to define a reasonable RECO format for Phase2, but nobody has taken up this task, so for now we just use FEVT.

In general, I think FEVT should be a superset of RECO, so the collection should be added back to that tier.

slava77:
> In general, I think FEVT should be a superset of RECO, so the collection should be added back to that tier.

note that "slimmed" is from miniAOD.
IIRC, we are not saving other miniAOD products in FEVT (except for the accidental recoGenJets_slimmed* pattern match).

kpedro88:
Ah, I see. I agree that miniAOD products should not be saved in FEVT in general, so maybe we should just leave it out.

perrotta:
please test

perrotta:
> Ah, I see. I agree that miniAOD products should not be saved in FEVT in general, so maybe we should just leave it out.

I expect that with the latest update from @jeongeun there will be no differences any more with respect to the current output (let wait for the results of the tests)

Then, if we want to remove those miniAOD products from the FEVT output, then probably another line that drops " recoGenJets_slimmed*_*_*" could simply be added to the RecoGenJetsFEVT PSet.

I wonder whether we want to do it already in this PR, thus accepting some change of output, or  if a separate PR (which also checks more carefully for other possible special cases) is more indicated to touch that 12 years old line of code. What do you think @kpedro88 ?

perrotta:
Uhm, now there is some missing content in he HI workflow 158 ("Run2_2018_pp_on_AA", RECOSIM event content)...

Indeed,the modifier only acts on the AOD event content: as it was before, the RECO and FEVT event contents were added of the AOD event content after it passed through the modifier, while moving it back closer to the definition of the additional contents (as I asked in my previous comment) do not feed the RECO/FEVT contents with the modified AOD one.

One possibility would be to put at the end of the file once again the ```outputCommands.extend```'s, where they were in your first commit. I personally believe that as such it would be less intuitive understanding what is in each event content.

Another possibility is to move up the whole set of AOD modifiers right after the definitions of the ```RecoJetsAOD``` and ```RecoGenJetsAOD``` event contents. 

I'd vote for option 2, but I'm also open to suggestions

slava77:
> Then, if we want to remove those miniAOD products from the FEVT output, then probably another line that drops " recoGenJets_slimmed*_*_*" could simply be added to the RecoGenJetsFEVT PSet.

I would rather the FEVT not know about names of the miniAOD collections.
Is there a good "keep" pattern that would work? If not, perhaps having these jets from miniAOD is a lesser evil.

jeongeun:
Thank you @perrotta for the suggestion, I think option 2 is a good idea so I will apply it.
To check if it's ok to use "keep" pattern as @slava77 commented,
I've checked the recoGenJets_* collection from the FEVT event contents 
I considered mc workflows as follows.

**5.1**_TTbar+TTbarFS+HARVESTFS 
```
edmDumpEventContent step1.root (=FEVTDEBUGHLToutput)

vector<reco::GenJet>                  "ak4GenJets"                ""                "RECO"
vector<reco::GenJet>                  "ak4GenJetsNoNu"            ""                "RECO"
vector<reco::GenJet>                  "ak8GenJets"                ""                "RECO"
vector<reco::GenJet>                  "ak8GenJetsNoNu"            ""                "RECO"
```
**9.0**_Higgs200ChargedTaus+Higgs200ChargedTausINPUT+DIGI+RECO+HARVEST  
**10024.0**_TTbar_13+TTbar_13TeV_TuneCUETP8M1_2017_GenSimFullINPUT+DigiFull_2017+RecoFull_2017+HARVESTFull_2017+ALCAFull_2017+Null17
**20034.0**_TTbar_14TeV+TTbar_14TeV_TuneCUETP8M1_2026D35_GenSimHLBeamSpotFull14INPUT+DigiFullTrigger_2026D35+RecoFullGlobal_2026D35+HARVESTFullGlobal_2026D35/
```
edmDumpEventContent step2.root (=FEVTDEBUGHLToutput)

vector<reco::GenJet>                  "ak4GenJets"                ""                "HLT"
vector<reco::GenJet>                  "ak4GenJetsNoNu"            ""                "HLT"
vector<reco::GenJet>                  "ak8GenJets"                ""                "HLT"
vector<reco::GenJet>                  "ak8GenJetsNoNu"            ""                "HLT"

edmDumpEventContent step3_inMINIAODSIM.root (=MINIAODSIMoutput)

vector<reco::GenJet>                  "slimmedGenJets"            ""                "RECO"
vector<reco::GenJet>                  "slimmedGenJetsAK8"         ""                "RECO"
vector<reco::GenJet>                  "slimmedGenJetsAK8SoftDropSubJets"   ""                "RECO"

```

**135.4**_ZEE_13+ZEEFS_13+HARVESTUP15FS+MINIAODMCUP15FS 

```
edmDumpEventContent step1.root (=FEVTDEBUGHLToutput)

vector<reco::GenJet>                  "ak4GenJets"                ""                "RECO"
vector<reco::GenJet>                  "ak4GenJetsForPUid"         ""                "RECO"
vector<reco::GenJet>                  "ak4GenJetsNoNu"            ""                "RECO"
vector<reco::GenJet>                  "ak8GenJets"                ""                "RECO"
vector<reco::GenJet>                  "ak8GenJetsNoNu"            ""                "RECO"
vector<reco::GenJet>                  "chargedak4GenJets"         ""                "RECO"
vector<reco::GenJet>                  "kinematicSelectedTauValDenominatorQCD"   ""                "RECO"
vector<reco::GenJet>                  "kinematicSelectedTauValDenominatorZTT"   ""                "RECO"
vector<reco::GenJet>                  "objectTypeSelectedTauValDenominatorModuleZTT"   ""                "RECO"
vector<reco::GenJet>                  "tauGenJetsForVal"          ""                "RECO"
vector<reco::GenJet>                  "patJetsBDHadron"           "genJets"         "RECO"

```
They give a total of 11 different module labels ( ak4GenJets, ak4GenJetsForPUid, ak4GenJetsNoNu, ...etc). I think using drops " recoGenJets_slimmed***"  seems an easier way.
I wonder what your opinion is on this.

slava77:
> I think using drops " recoGenJets_slimmed***" seems an easier way.
> I wonder what your opinion is on this.

Just to clarify, if the is no simple RecoJets-related keep statement, I'd rather everything is still written and we do not add drop statements with collection names coming from outside.

In the examples above, it seems to me that  ```keep recoGenJets_ak*_*_*``` covers everything RecoJets-related. All other products are defined somewhere else (mainly in Validation) and should not really be collected by the RecoJets_EventContent_cff.py

perrotta:
please test

perrotta:
+1
- The reco Jets event contents have been reorganized: AOD content is now explicitly a subset of the RECO one, which is a substet of the FEVT
- While doing so, it was realized that slimmedGenJets (from miniAOD) had slipped in the FEVT event content, due to the too wide prescription which was previously set up: this has also been fixed here
- Jenkins tests pass, and the only difference observed is the removal of those slimmedGenJets from the FEVT content (used for now in Phase2 workflows)
- This is the first of a serie of such updates, eventually intended to reorganize the whole event contents for reco objects

silviodonato:
+1

PR #29026, opened 2020-02-22
Clean BuildFiles in some more reco subsystems

guitargeek:
#### PR description:

After this one, there is probably only one more PR and then all of the Reco* Build files don't use dependencies anymore which are not included in the source files.

#### PR validation:


slava77:
@cmsbuild please test

Martin-Grunewald:
+1

slava77:
+1

for https://github.com/cms-sw/cmssw/pull/29026	5d259a4 
- code changes are in line with the PR description and the previous similar PRs

silviodonato:
+1

PR #29032, opened 2020-02-25
Minimal era for Run 3 PbPb and corresponding relVal workflows

mandrenguyen:
#### PR description:

Adds Run3_PbPb to run PbPb reconstruction with 2021 conditions.  This era contains the minimal modifications to produce a running workflow.  It uses the same modifiers as in the Run2_2018_pp_on_AA era.  Also adding 2021 heavy-ion MC relval workflows by copying the corresponding 2018 ones. 

#### PR validation:

Tested locally with the new relval workflow 159.1, after adding PR #29078 by hand.


mandrenguyen:
@abaty

boudoul:
Great , thanks @mandrenguyen ! Presumably we should define a workflow using it, do you want me to take care ?

mandrenguyen:
@boudoul I can setup a relval workflow. Let's get some feedback first on this proposal for an era from @slava77 @perrotta

boudoul:
Perfect, thank you

Configuration/Eras/python/Era_Run3_PbPb_cff.py
```diff
@@ -0,0 +1,7 @@
+import FWCore.ParameterSet.Config as cms
+
+from Configuration.Eras.Era_Run3_cff import Run3
+from Configuration.Eras.Modifier_pp_on_AA_2018_cff import pp_on_AA_2018
+from Configuration.Eras.Modifier_pbpb_run3_cff import pbpb_run3
+
+Run3_PbPb = cms.ModifierChain(Run3, pp_on_AA_2018, pbpb_run3)
```
slava77: why not keep the same style as before and have "pp_on_AA" in the name[s]?

slava77 (commented):
> except it removes the HI-specific track MVA calibrations, which do not yet exist in a 2021 global tag

why do they not exist for the 2021 GT?
IIRC, the regular tracking selectors are the same in 2021 as in 2018.
Did I miss some essential tracking changes that would invalidate HIMVASelector[s]?

Configuration/Eras/python/Era_Run3_PbPb_cff.py
mandrenguyen: pp_on_AA was intended to mean that we are running a pp-like reco on AA collisions.  This is in contrast to the dedicated heavy-ion reco, which was also still being maintained in 2018.  Now that such a dedicated reco sequence has been deprecated, I thought it best to simplify the naming.   I'm open to suggestion though.

mandrenguyen:
> > except it removes the HI-specific track MVA calibrations, which do not yet exist in a 2021 global tag
> 
> why do they not exist for the 2021 GT?
> IIRC, the regular tracking selectors are the same in 2021 as in 2018.
> Did I miss some essential tracking changes that would invalidate HIMVASelector[s]?

In my tests I have used the standard pp GT (and corresponding beamspot) in my tests.   A crash in reco indicated to me that they do not appear to contain this record.   I assume there is not matching GT/beamspot set yet for heavy-ion beam conditions.   I don't believe there's anything fundamental.  One could simply add these records to the pp GTs as well.   Of course the selection  has to be retrained for Run 3 in any case.

slava77:
> One could simply add these records to the pp GTs as well.

I prefer this solution; it seems more appropriate for the current state of the physics performance setup.

Configuration/Eras/python/Era_Run3_PbPb_cff.py
slava77: IIUC, the HeavyIon scenario still exists. So, there is still a clear case to still call this pp_on_AA. To me the naming should reflect the status of the release in a coherent way rather than a specific intention for the future.
I prefer a coherent migration instead of changing in a single point.

Also, would AA or HI still work, or does "PbPb" need to be used here?

mandrenguyen:
@abaty Can you upload the HIMVA tags into the GT queue for 2021 realistic conditions, please?

Configuration/Eras/python/Era_Run3_PbPb_cff.py
mandrenguyen: I don't think there's much potential for confusion of --scenario HeavyIons with pbpb_run3, whereas most people will not understand the meaning of pp_on_AA, but I don't insist on the "pp_on_X" part.   AA denotes generic ion collisions.  Up to now we had PbPb and XeXe.  If we ran these today, we would like use the same pp_on_AA modifier on both, so I used AA to distinguish from pA.   However, in Run 3 we are likely to have Oxygen-Oxygen collisions, for which we would use the standard pp high lumi reconstruction settings.  So pp_on_AA doesn't make a whole lot of sense.   Should I call it pp_on_PbPb_run3?  Or we just live with pp_on_AA_2021, in the name of continuity?

abaty:
@mandrenguyen I've asked Xiao Wang to upload the relevant tags to the GT queue.

tocheng:
I prefer to have a separate GT for heavy Ion MC. AlCa can make a heavy Ion queue for 2021 based on pp and heavy Ion group can queue tags to add new conditions or replace conditions in pp GT.

mandrenguyen:
@tocheng Is there a reason not to add this record to the pp GT?  We may want to test the HI era with various pp GTs in the future.  In any case, this is fine for us if @slava77  agrees.  What would be the ETA to produce a HI GT that is the same as the one pointed to by auto:phase1_2021_realistic, except that it has the add'l HIMVA records?

tocheng:
@mandrenguyen What do you mean various pp GTs?
We could make a new queue  based on pp 2021 and you can add the missing tracker selector. The HI GT can be made soon after.

mandrenguyen:
@tocheng I just meant that adding this record to pp GTs would allow us to run PbPb_run3 era with pp GTs without crashing.   That sounds like something that could come in handy to me.  I think we can drop this discussion though.    We will need the HI queue anyway, since there will eventually be beamspot tag to upload there.  Let's just get this era operational.   Once a HI GT is available I can update the autocond, and add relval wf.

tocheng:
111X_mcRun3_2021_realistic_HI_Queue is ready and is equal to auto:phase1_2021
https://cms-conddb.cern.ch/cmsDbBrowser/diff/Prod/gts/111X_mcRun3_2021_realistic_HI_Queue/110X_mcRun3_2021_realistic_v8

Taburis:
Hi @tocheng, @mandrenguyen, and @slava77, the MVA tags added to the queue: 111X_mcRun3_2021_realistic_HI_Queue.

christopheralanwest:
PR #29078 contains a new autoCond key, `auto:phase1_2021_realistic_hi`, with the requested global tag. A new relval workflow should be created that uses this new key.

christopheralanwest:
assign pdmv

This PR should introduce a relval workflow that uses the `auto:phase1_2021_realistic_hi` GT from PR #29078 prior to merging these two PRs. We don't want to introduce new autoCond keys that are not tested by any relval workflow.

mandrenguyen:
@christopheralanwest I can introduce a new relval workflow, but I'm not clear on your proposed order of operations.  If I add the workflow to this PR it won't pass tests until #29078 is merged right?  So shouldn't we merge #29078 and then update this PR (or even just make a 3rd PR with the relval wf)?

chayanit:
> @christopheralanwest I can introduce a new relval workflow, but I'm not clear on your proposed order of operations. If I add the workflow to this PR it won't pass tests until #29078 is merged right? So shouldn't we merge #29078 and then update this PR (or even just make a 3rd PR with the relval wf)?

@mandrenguyen two PR can be tested together and no need to wait for one to be merged. I think you can add relval wf in this PR.

mandrenguyen:
@christopheralanwest I have copied the 2018 HI wfs to the 2021 era and rolled back the use of the pp track MVA calibrations, since the HI ones are now available from the GT.   The driver options are as described in the original PR description.

christopheralanwest:
please test workflow 159,159.1,159.2,159.3 with #29078

Configuration/PyReleaseValidation/python/relval_steps.py
```diff
@@ -2568,6 +2603,7 @@ def gen2018HiMix(fragment,howMuch):
                      '--filein':'file:PromptCalibProdEcalPedestals.root'}
 
 steps['RECOHISt4']=steps['RECOHI2015']
+steps['RECOHI2021MIX']=merge([steps['RECOHI2021PPRECO'],{'--pileup':'HiMix','--pileup_input':'das:/RelValHydjetQ_B12_5020GeV_2021/%s/GEN-SIM'%(baseDataSetRelease[9])}])
```
chayanit: @mandrenguyen you may need to introduce the new baseDataSetRelease for 2021 string

chayanit (changes requested):

Configuration/PyReleaseValidation/python/relval_steps.py
```diff
@@ -932,25 +939,33 @@ def genS(fragment,howMuch):
...
+steps['HydjetQ_B12_5020GeV_2021_ppReco']=merge([{'-n':1},hiDefaults2021_ppReco,gen2021hiprod('Hydjet_Quenched_B12_5020GeV_cfi',U2000by1)])
 
 steps['QCD_Pt_80_120_13_HI']=merge([hiDefaults2018_ppReco,gen2018hiprod('QCD_Pt_80_120_13TeV_TuneCUETP8M1_cfi',Kby(9,150))])
 steps['PhotonJets_Pt_10_13_HI']=merge([hiDefaults2018_ppReco,gen2018hiprod('PhotonJet_Pt_10_13TeV_TuneCUETP8M1_cfi',Kby(9,150))])
 steps['ZMM_13_HI']=merge([hiDefaults2018_ppReco,gen2018hiprod('ZMM_13TeV_TuneCUETP8M1_cfi',Kby(18,100))])
 steps['ZEEMM_13_HI']=merge([hiDefaults2018_ppReco,gen2018hiprod('ZEEMM_13TeV_TuneCUETP8M1_cfi',Kby(18,300))])
 
+steps['QCD_Pt_80_120_13_HI_2021']=merge([hiDefaults2021_ppReco,gen2021hiprod('QCD_Pt_80_120_13TeV_TuneCUETP8M1_cfi',Kby(9,150))])
+steps['PhotonJets_Pt_10_13_HI_2021']=merge([hiDefaults2021_ppReco,gen2021hiprod('PhotonJet_Pt_10_13TeV_TuneCUETP8M1_cfi',Kby(9,150))])
+steps['ZMM_13_HI_2021']=merge([hiDefaults2021_ppReco,gen2021hiprod('ZMM_13TeV_TuneCUETP8M1_cfi',Kby(18,100))])
```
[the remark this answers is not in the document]
mandrenguyen: @chayanit  Yes, it would make sense to switch to 14 TeV.  Now that it occurs to me, it would be very useful to have one gen fragment that's run with both the standard pp settings and the HI era.   Would it be ok to add a TTbar_14TeV_TuneCP5 workflow with heavy-ion settings  for this purpose?   Regarding the other point, I will clean up the ZMM, which is no longer used.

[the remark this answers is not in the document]
chayanit: @mandrenguyen we do have these GEN fragments for 14TeV. Please update
https://github.com/cms-sw/cmssw/blob/master/Configuration/Generator/python/QCD_Pt_80_120_14TeV_TuneCUETP8M1_cfi.py
https://github.com/cms-sw/cmssw/blob/master/Configuration/Generator/python/PhotonJet_Pt_10_14TeV_TuneCUETP8M1_cfi.py
but not ZEEMM. Would you like to switch from ZEEMM to TTbar?

boudoul:
Hello @mandrenguyen , just a small detail:  may you please update the title : In addition to 'Minimal Era for Run2 PbPb' you may want to add 'and corresponding RunTheMatrix Workflows' (or something like that) . Thanks !

Configuration/PyReleaseValidation/python/relval_steps.py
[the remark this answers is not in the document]
mandrenguyen: Thanks!  Z bosons tend to be pretty useful for us, I'm not sure I would like get rid of these wfs.   I see that there are separate ZMM and ZEE gen fragments for 14 TeV already.  What if  switched to these and reduced the number of events for each wf by a factor of 2?  
While the 14 TeV fragments exist, are any of these workflows actually run in the IB?    That was my idea for adding ttbar.  On the other hand, the scope of this PR is already spreading beyond my initial intention, so maybe we can drop that idea.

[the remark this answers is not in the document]
chayanit: @mandrenguyen  Instead of split them to two more workflows, I would recommend to create new GEN fragment for ZEEMM 14TeV under Configuration/Generator/python. You can just copy this GEN fragment and update energy to 14TeV https://github.com/cms-sw/cmssw/blob/master/Configuration/Generator/python/ZEEMM_13TeV_TuneCUETP8M1_cfi.py
I don't think any of these workflows are included in the IB, usually we use TTbar. You can try to add and I will have a look. 

I have other comments elsewhere. Please also consider the changes. Thanks.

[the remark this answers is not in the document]
mandrenguyen: @chayanit Your other comments seem straightforward, so I will address them as soon as I can find some time :-)

Configuration/Eras/python/Era_Run3_PbPb_cff.py
slava77: I still think that ```pp_on_``` should still be a part of the era name. I'm fine to stay with PbPb if other options may need a different setup

Configuration/Eras/python/Modifier_pbpb_run3_cff.py
```diff
@@ -0,0 +1,4 @@
+import FWCore.ParameterSet.Config as cms
+
+pbpb_run3 =  cms.Modifier()
```
slava77: this is not needed anymore.
If a placeholder is desired, a name following existing pattern would be pp_on_AA_run3 or pp_on_PbPb_run3 if finer specification is desired

Configuration/PyReleaseValidation/python/relval_steps.py
chayanit: @mandrenguyen your current setup will call baseDataSetRelease[9] string which is from Run2. I was asking you to implement new index of baseDataSetRelease but yes you're right we never produced any Run3 mixing yet and it will fail. You can continue with what you have for now.

kpedro88:
please test

christopheralanwest:
please abort

The tests should be run with PR #29078 and test the workflows introduced in this PR.

christopheralanwest:
please test workflow 159,159.1,159.2,159.3,159.4,310,311,312 with #29078

kpedro88:
@christopheralanwest in the future, please use the "test parameters" [bot command](http://cms-sw.github.io/cms-bot-cmssw-cmds.html) so that these special settings don't have to be included every time someone calls "please test"

mandrenguyen:
@christopheralanwest 310 and related workflows fail b/c of the lack of an appropriate pileup_dataset.  The GEN-SIM output of the new workflow 159 will be used for this.  See the discussion with @chayanit in this PR's comments.

christopheralanwest:
test parameters
workflow = 159,159.1,159.2,159.3,159.4
pull_request = #29078

christopheralanwest:
please test

slava77:
+1

for https://github.com/cms-sw/cmssw/pull/29032	5714287 
- changes in the reco category are trivial (an empty line added in RecoTracker/IterativeTracking/python/LowPtTripletStep_cff.py); the Era modifier name (not explicitly in reco category) is in line with the review comments.
- jenkins tests passed (where expected); there are no diffs to make in the results given that the rest of the PR is introducing a new workflow. The non-PU workflows ran OK already in https://cmssdt.cern.ch/SDT/jenkins-artifacts/pull-request-integration/PR-3e9604/5037/summary.html

christopheralanwest:
please test

The global tag in #29078 needed to be updated to include an update to the HCAL zero suppression for Run 3.

silviodonato:
+operations

silviodonato:
Do you have any comments?
upgrade:  @kpedro88
pdmv:  @chayanit @pgunnell @zhenhu

Configuration/PyReleaseValidation/python/relval_steps.py
```diff
@@ -932,25 +937,33 @@ def genS(fragment,howMuch):
...
+steps['HydjetQ_B12_5020GeV_2021_ppReco']=merge([{'-n':1},hiDefaults2021_ppReco,gen2021hiprod('Hydjet_Quenched_B12_5020GeV_cfi',U2000by1)])
 
 steps['QCD_Pt_80_120_13_HI']=merge([hiDefaults2018_ppReco,gen2018hiprod('QCD_Pt_80_120_13TeV_TuneCUETP8M1_cfi',Kby(9,150))])
 steps['PhotonJets_Pt_10_13_HI']=merge([hiDefaults2018_ppReco,gen2018hiprod('PhotonJet_Pt_10_13TeV_TuneCUETP8M1_cfi',Kby(9,150))])
-steps['ZMM_13_HI']=merge([hiDefaults2018_ppReco,gen2018hiprod('ZMM_13TeV_TuneCUETP8M1_cfi',Kby(18,100))])
 steps['ZEEMM_13_HI']=merge([hiDefaults2018_ppReco,gen2018hiprod('ZEEMM_13TeV_TuneCUETP8M1_cfi',Kby(18,300))])
 
+steps['QCD_Pt_80_120_14_HI_2021']=merge([hiDefaults2021_ppReco,gen2021hiprod('QCD_Pt_80_120_14TeV_TuneCUETP8M1_cfi',Kby(9,150))])
+steps['PhotonJets_Pt_10_14_HI_2021']=merge([hiDefaults2021_ppReco,gen2021hiprod('PhotonJet_Pt_10_14TeV_TuneCUETP8M1_cfi',Kby(9,150))])
+#steps['ZEEMM_13_HI_2021']=merge([hiDefaults2021_ppReco,gen2021hiprod('ZEEMM_13TeV_TuneCUETP8M1_cfi',Kby(18,300))])
```
chayanit: @mandrenguyen I thought we agreed to have ZEEMM not separated. Could you add corresponding GEN fragment and update this part?

chayanit (changes requested):

Configuration/PyReleaseValidation/python/relval_steps.py
mandrenguyen: @chayanit What is the logic of running them together?

chayanit: @mandrenguyen what is the logic to run them separate? We used to run them together in 2018 no?

mandrenguyen: @chayanit  The logic is to keep them in sync easily with the high lumi pp workflows.  I don't see why we should add an extra layer of complexity to anyone analyzing the output, as well as add additional gen fragments to maintain.  If you're worried about the extra processing, we can cut the number of events in half.

chayanit: @mandrenguyen ZEEMM has been used as HI workflow since 2018 and now you said it's adding extra layer of complexity by just selecting Z from muon or electron? If you want to keep them in sync with high lumi pp, you will need to update all workflows to TuneCP5.

mandrenguyen: Yes, was the one who made the PR to add ZEEMM IIRC.  Can we not be stuck with it forever, please?

chayanit: So please remove ZEEMM line you commented out completely if you wanna move to separate workflows. How about the Tuning? Do we stay with the old tuning for HI?

mandrenguyen: Ok, will do this afternoon, thanks!

mandrenguyen: @chayanit  I just pushed the change to CP5 for the ZEE and ZMM wfs (and removed the commented out wf).  I don't see CP5 fragments for the QCD and gamma-jet.  Did you want me to migrate those as well?

slava77:
@cmsbuild please test

Configuration/PyReleaseValidation/python/relval_steps.py
chayanit: @mandrenguyen it would be nice to have everything in the same setup. Thanks!

RecoTracker/IterativeTracking/python/LowPtTripletStep_cff.py
```diff
@@ -294,6 +294,7 @@
         mva = dict(GBRForestLabel = 'HIMVASelectorLowPtTripletStep_Phase1'),
         qualityCuts = [-0.8, -0.4, 0.5],
 )
+
```
kpedro88: remove errant newline so this PR doesn't change this file for no reason

mandrenguyen: The offending newline is removed in the latest commit

Configuration/PyReleaseValidation/python/relval_steps.py
mandrenguyen: I have added the CP5 fragments and updated the relevant workflows

slava77:
@cmsbuild please test

notably, a reco signature is not needed anymore

chayanit:
+1

silviodonato:
+operations

kpedro88:
+upgrade

silviodonato:
Kind reminder for generators:  @alberto-sanchez @agrohsje @efeyazgan @mkirsano @qliphy @SiewYan

qliphy:
+1

silviodonato:
+1

Issue #29036, opened 2020-02-25
Geometry/DTGeometryBuilder unit test fails for ARM


Dr15Jones:
The aarch64 IB has the unit tests GeometryDTGeometryBuilderTestDriver failing. The reason is doing a diff on the geometry output is showing an irrelevant difference. One such difference is

```
< Chamber  Wh:-2 St:1 Se:1  Position  (431.175,39.12,-533.35)  normVect  (-1,1.80411e-16,-6.21021e-20)  bounds W/H/L: 218/36.2/251.1
> Chamber  Wh:-2 St:1 Se:1  Position  (431.175,39.12,-533.35)  normVect  (-1,3.53686e-16,3.02612e-16)  bounds W/H/L: 218/36.2/251.1
```

where the coordinates of the normal vector which mathematically should be identical to 0. are instead very tiny numbers where x86 and aarch64 disagree on exactly which tiny number to use.

Computationally, any component of a normal vector which is on the order of 10^-7 will be completely irrelevant for bit wise comparisions for calculating the normal and for values produced using a dot product when using floats. For doubles values on the order of 10^-16 will be irrelevant for bitwise comparisons. If we do not care about bit wise, than substantially larger values will be irrelevant for actual physics results.

My suggestion is to either
1. have the geometry dump code look at the values of the normals and replace anything on the order of 10^-7 with 0. or
1. have the code that does computes the normal vectors replace values on order 10^-x (where x is some arbirary number > 6) with 0.

Dr15Jones:
assign geometry

Dr15Jones:
A recent example can be found here
https://cmssdt.cern.ch/SDT/cgi-bin/buildlogs/raw/cc8_aarch64_gcc8/CMSSW_11_1_X_2020-02-24-2300/unitTestLogs/Geometry/DTGeometryBuilder

cvuosalo:
I think I will create `DataFormats/Math/interface/Rounding.h`. It will contain:
```
namespace rounding {
template <class valType>
inline constexpr valType roundIfCloseTo0(valType value, double tolerance = 1.e-7) {
  if (std::abs(value) < tolerance)
    return (0.0);
  return (value);
}

cvuosalo:
@Dr15Jones How can I test this issue? I don't know about the ARM machines, and I probably don't have access to them.

Dr15Jones:
I don't know how to access the ARM machines either. I can only look at the results from the related IB.

[...]

PR #29052, opened 2020-02-26
Fix failing test on ARM by rounding small values to 0

cvuosalo:
Certain units tests are failing on the ARM architecture because tiny floating-point values near 0 are slightly different than on the standard CPU, as described in issue #29036.

To fix this problem, the printed values used for comparison will be rounded to 0 if they are less than 1e-7. The internal values will not be changed, only the printed values used to compare the test results and the baseline reference file, which is being updated.

This change means the DT Geometry unit test will be less sensitive to small differences in norm vector values near zero, which could possibly mean it might miss a real problem, but the current test is too sensitive in failing because of 1e-16 fluctuations. I think the version of the test in this PR is sufficiently sensitive to catch real errors while avoiding false positives.

This PR adds a new utility file `Rounding.h`. I looked to see if a function to round values near 0 already exists, but I couldn't find one, so I added this new file. If there is already such a rounding function, or if I put the new file in the wrong place, this PR can be revised.

#### PR validation:

The unit test should now be platform-independent. I hope this PR can be tested against the ARM IB because I have not be able to test the code change on an ARM machine.

No backport is needed.


cvuosalo:
@cmsbuild please test

ianna:
@cvuosalo please, remove `Geometry/DTGeometryBuilder/test/dtGeometry.log.org` from this PR and generate it for testing on the fly in a temporary test area.

ianna:
hold

ianna:
unhold

ianna:
please test for slc7_aarch64_gcc820

ianna:
+1

cvuosalo:
@ianna It looks like you changed your mind about generating the reference file on-the-fly. Could you please delete your comment about it, or add a comment saying so? I wouldn't want any reviewers to be confused.

ianna:
@cvuosalo - IMHO, the reference file should be generated on the fly, but ultimately, it's your decision how high this is on the priority list.

cvuosalo:
type bugfix

cvuosalo:
@ianna I will leave the PR as it is for now. I think on-the-fly generation of the reference file is lower priority than other issues I need to work on.

perrotta:
+1
- A new function is added ti DataFormats/Math to round up to zero small values (default below 10-7)
- That function is used in some DT Geometry unit tests, whose reference file I'd let the Geometry conveners to decide whether has to be generated on the fly, or not
- Jenkins tests pass

silviodonato:
+1

PR #29054, opened 2020-02-27
Remove redundant looping of elements for link tests in PFBlockAlgo

hatakeyamak:
#### PR description:

Tuning the looping logic so that we can remove the redundant looping of elements for link tests in PFBlockAlgo. This simple change should make the PFBlockAlgo [one of the most timing consuming part of PF] faster by a factor up to 1.8 with no change in outputs.

Using HLT's OnLine_HLT_GRun.py on:
/eos/cms/store/data/Run2018D/EphemeralHLTPhysics5/RAW/v1/000/324/897/00000/F5F13777-D664-C747-88AA-0459C1E35254.root
(1000 events)
**FastReport**
```
[Before]
CPU time avg.      when run  Real time avg.      when run     Alloc. avg.      when run   Dealloc. avg.      when run  Modules
        44.0 ms        44.0 ms        44.7 ms        44.7 ms       +3349 kB       +3349 kB       -2607 kB       -2607 kB    hltParticleFlowBlock
         0.9 ms        44.0 ms         0.9 ms        44.4 ms         +72 kB       +3434 kB         -55 kB       -2663 kB    hltParticleFlowBlockForTaus
         1.5 ms        43.0 ms         1.6 ms        45.2 ms        +112 kB       +3121 kB         -85 kB       -2381 kB    hltParticleFlowBlockReg

[With this PR]
        24.7 ms        24.7 ms        26.3 ms        26.3 ms       +3349 kB       +3349 kB       -2607 kB       -2607 kB    hltParticleFlowBlock
         0.5 ms        25.2 ms         0.6 ms        28.7 ms         +72 kB       +3438 kB         -56 kB       -2668 kB    hltParticleFlowBlockForTaus
         0.9 ms        24.5 ms         1.0 ms        27.8 ms        +112 kB       +3120 kB         -85 kB       -2382 kB    hltParticleFlowBlockReg
```

And, just for reference, if we also drop PS from PFBlockProducer for HLT as discussed in the TSG meeting, March 4th, (https://github.com/hatakeyamak/cmssw/pull/new/PFBlockProducerDropPS2)
```
        14.7 ms        14.7 ms        15.9 ms        15.9 ms       +1742 kB       +1742 kB       -1400 kB       -1400 kB    hltParticleFlowBlock
         0.3 ms        16.2 ms         0.4 ms        17.3 ms         +36 kB       +1752 kB         -29 kB       -1399 kB    hltParticleFlowBlockForTaus
         0.5 ms        12.6 ms         0.5 ms        12.9 ms         +52 kB       +1460 kB         -41 kB       -1146 kB    hltParticleFlowBlockReg
```

#### PR validation:

Check with 100 events from 2018 data that this change yields the identical list of PF candidates.

This is not backport.

@bendavid @jsalfeld @jpata

### DIFF
--- a/RecoParticleFlow/PFProducer/plugins/kdtrees/KDTreeLinkerPSEcal.cc
+++ b/RecoParticleFlow/PFProducer/plugins/kdtrees/KDTreeLinkerPSEcal.cc
@@ -139,7 +139,7 @@ void KDTreeLinkerPSEcal::buildTree(const RecHitSet &rechitsSet, KDTreeLinkerAlgo
 }
 
 void KDTreeLinkerPSEcal::searchLinks() {
-  // Must of the code has been taken from LinkByRecHit.cc
+  // Most of the code has been taken from LinkByRecHit.cc
 
   // We iterate over the PS clusters.
   for (BlockEltSet::iterator it = targetSet_.begin(); it != targetSet_.end(); it++) {
--- a/RecoParticleFlow/PFProducer/plugins/kdtrees/KDTreeLinkerTrackEcal.cc
+++ b/RecoParticleFlow/PFProducer/plugins/kdtrees/KDTreeLinkerTrackEcal.cc
@@ -135,7 +135,7 @@ void KDTreeLinkerTrackEcal::buildTree() {
 }
 
 void KDTreeLinkerTrackEcal::searchLinks() {
-  // Must of the code has been taken from LinkByRecHit.cc
+  // Most of the code has been taken from LinkByRecHit.cc
 
   // We iterate over the tracks.
   for (BlockEltSet::iterator it = targetSet_.begin(); it != targetSet_.end(); it++) {
--- a/RecoParticleFlow/PFProducer/plugins/kdtrees/KDTreeLinkerTrackHcal.cc
+++ b/RecoParticleFlow/PFProducer/plugins/kdtrees/KDTreeLinkerTrackHcal.cc
@@ -159,7 +159,7 @@ void KDTreeLinkerTrackHcal::buildTree() {
 }
 
 void KDTreeLinkerTrackHcal::searchLinks() {
-  // Must of the code has been taken from LinkByRecHit.cc
+  // Most of the code has been taken from LinkByRecHit.cc
 
   // We iterate over the tracks.
   for (BlockEltSet::iterator it = targetSet_.begin(); it != targetSet_.end(); it++) {
@@ -239,7 +239,7 @@ void KDTreeLinkerTrackHcal::searchLinks() {
 void KDTreeLinkerTrackHcal::updatePFBlockEltWithLinks() {
   //TODO YG : Check if cluster positionREP() is valid ?
 
-  // Here we save in each HCAL cluster the list of phi/eta values of linked clusters.
+  // Here we save in each HCAL cluster the list of phi/eta values of linked clusters (actually tracks).
   for (BlockElt2BlockEltMap::iterator it = cluster2TargetLinks_.begin(); it != cluster2TargetLinks_.end(); ++it) {
     reco::PFMultiLinksTC multitracks(true);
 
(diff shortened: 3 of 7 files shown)


slava77:
code-check

if the bot lost the last attempt

slava77:
code-checks

slava77:
@cmsbuild please test

hatakeyamak:
I expected absolutely no difference, but I see some failures.

Most of reported differences are related to:
recoCandidateedmPtrsrecoJetTagInforecoIPTagInfos_pfImpactParameterTagInfos__(re)RECO_obj_m_data_ghostTrackWeight
recoCandidateedmPtrsrecoJetTagInforecoIPTagInfos_pfImpactParameterTagInfos__RECO_obj_m_data_distanceToGhostTrack_value
and some muon-related DQM plots, but I am not sure if I understand the reported differences. Any suggestions will be appreciated to move this forward.

jpata:
Hi @hatakeyamak, just a quick question - did the number of PFBlocks stay the same before and after the change? I wonder if we would benefit from automatic tests of the code on one event with full verbose printout (LogDebug) - this can easily be diffed. I could set this up.

hatakeyamak:
> Hi @hatakeyamak, just a quick question - did the number of PFBlocks stay the same before and after the change? I wonder if we would benefit from automatic tests of the code on one event with full verbose printout (LogDebug) - this can easily be diffed. I could set this up.

Thanks. Yes, the number of PFBlocks stays the same in the test I looked at. The list of PF candidates also stays the same as far as I can tell, but the ordering of candidates in the collection change, which I think is due to the order of PFBlocks change, and this is probably expected because the order of union-find algo in PFBlockAlgo is not kept the same with this update. Check of this assumption could be very useful as you say.

All these changes shouldn't be a problem as long as physical observables are at the end sorted in some ways. I just cannot quite connect this expectation with reported failures, as I don't fully understand all the reported failed tests.

perrotta:
please test

hatakeyamak:
In my investigation of
```
recoCandidateedmPtrsrecoJetTagInforecoIPTagInfos_pfImpactParameterTagInfos__(re)RECO_obj_m_data_ghostTrackWeight
recoCandidateedmPtrsrecoJetTagInforecoIPTagInfos_pfImpactParameterTagInfos__RECO_obj_m_data_distanceToGhostTrack_value
```
I see some small changes in:
```
vector<reco::Track>                   "pfImpactParameterTagInfos"   "ghostTracks"     "(re)RECO"    
vector<reco::IPTagInfo<vector<edm::Ptr<reco::Candidate> >,reco::JetTagInfo> >    "pfImpactParameterTagInfos"   ""                "(re)RECO"    
```
For example, in output of 136.874_RunEGamma2018C, 100 events, I see 3 cases of tiny differences in ghostTracks:
```
108066c108066
< tracks: pt   0.6561851033277 eta  -0.98355321390117 phi   1.87568111755637 
---
> tracks: pt   0.6540341172427 eta  -0.98789902418708 phi   1.87540677496420 

118614c118614
< tracks: pt   0.8130750166088 eta  -0.66572216238707 phi  -2.40236842625495 
---
> tracks: pt   0.8130774022123 eta  -0.66571723565971 phi  -2.40238250896182 

208658c208658
< tracks: pt   0.6187283287962 eta   1.05984628567672 phi  -2.38975964612986 
---
> tracks: pt   0.6167970203661 eta   1.06382187233677 phi  -2.39066671374844 
```
My impression is that ghosttrack reco is very sensitive to numerical differences, which can be caused by changes of pf candidates order, but the differences is very small. I doubt it's something we need to worry about.

Regarding some changes in DQM plots, I think they are showing up for plots involving
vector<reco::PFCandidate>             "pfIsolatedMuonsEI"         ""                "(re)RECO"
and it's because these muons are not sorted in any particular way [and there is no better or worse in how this collection was defined before or after this PR]. I think the better way can be explored after carefully check also how other pf*EI collections are also defined/sorted, and it can be postponed to another PR.

perrotta (commented):
Sorry, I had written those comments since a while already, but I realize now that they were never submitted

hatakeyamak:
Just checking in to see the review status. There are planned follow-up updates, so it will be nice to have this converged hopefully soon.

perrotta:
please test

hatakeyamak:
Thank you @perrotta for your review and also for initiating the test.

perrotta:
+1
- Redundant looping  in PFBlockAlgo is removed, as explained in the PR description
- Reco outputs are expected to remain the same: tiny differences observed in the jenkins tests are compatible with small numerical differences due to a different ordering of the PFBlocks, see alsohttps://github.com/cms-sw/cmssw/pull/29054#issuecomment-596620086

silviodonato:
+1

PR #29060, opened 2020-02-28
Delete obsolete filter modules in GenFilters

wddgit:
#### PR description:

These obsolete modules were identified by the fact that they have been broken for years and no one has fixed them. If needed they will be in the git history and can be restored (and fixed, migrated to call consumes, migrated to not be legacy...).

Also deleted related helper classes and cfi files.

#### PR validation:

Nothing to test, only deletions.


wddgit:
please test

SiewYan:
+1

silviodonato:
+1

PR #29067, opened 2020-03-02
DQM: Avoid hoarding lumi MEs in reco jobs.

schneiml:
#### PR description:

An issue discussed in https://its.cern.ch/jira/projects/CMSCOMPPR/issues/CMSCOMPPR-10942 seems to be caused by per-lumi MEs being cloned per lumi, but never actually cleaned up in RECO jobs, since the cleanup operation `deleteUnusedLumiHistograms` is only called in the `DQMFileSaver` which only runs in HARVESTING jobs.

This apparently never caused trouble, since either the amount of per-lumi MEs or lumisections per job was quite small.

#### PR validation:
None so far.

No master PR yet. The relevant code does not exist in master/11_1.

Forward port to 11_0 might bepossible.

### DIFF
--- a/DQMServices/FwkIO/plugins/DQMRootOutputModule.cc
+++ b/DQMServices/FwkIO/plugins/DQMRootOutputModule.cc
@@ -442,6 +442,8 @@ DQMRootOutputModule::writeLuminosityBlock(edm::LuminosityBlockForOutput const& i
     m_indicesTree->Fill();
   }
 
+  dstore->deleteUnusedLumiHistograms(m_enableMultiThread ? m_run : 0, m_lumi);
+
   edm::Service<edm::JobReport> jr;
   jr->reportLumiSection(m_jrToken, m_run, m_lumi);
 }


schneiml:
please test

Let's see if anything blows up here.

silviodonato:
backport of #29068

silviodonato:
What about changing also here in master
https://github.com/cms-sw/cmssw/blob/master/DQMServices/FwkIO/plugins/DQMRootOutputModule.cc#L424 ?

cc: @srimanob

schneiml:
@silviodonato The location still exists, but `deleteUnusedLumiHistograms` doesn't exist any more.

silviodonato:
As discussed at the ORP meeting, we can close this PR. Feel free to open it back @schneiml  @srimanob

slava77:
another issue apparently related to this
https://its.cern.ch/jira/browse/CMSCOMPPR-11931

PR #29079, opened 2020-03-03
DQM: Fix type for DeadChannel.threshold QTest

schneiml:
#### PR description:

A small bug spotted running DT Online DQM.

This QTest seems to be not used in any offline workflows, else it would have probably blown up there before.

#### PR validation:

Works on online DQM playback system.

No changes in offline expected.


schneiml:
+1

silviodonato:
+1

PR #29085, opened 2020-03-03
Change globals in BasicHepMCValidation to constexpr

Dr15Jones:
#### PR description:

This was found by the static analyzer. As this module is now a stream module, we need to avoid use of non-const globals. In this case they were never changed but they were still filling up the static analyzer log.

#### PR validation:

The code compiles.


Dr15Jones:
please test

jfernan2:
+1
Changes unrelated to this PR

agrohsje:
+1

silviodonato:
+1

PR #29086, opened 2020-03-03
Add header guards for DYTThrObject.h and Serializable.h

oshadura:
For Serializable.h, we should change #pragma once, because it gets the filename and creates a include guard based on it. If it relocates it becomes problematic, because it is could be possible to have two files with the same name in two different locations.
(to avoid such cases we should rely less on #pragma once)

#### PR description:

This PR is a part of https://github.com/cms-sw/cmsdist/pull/5600

cc: @davidlange6 @vgvassilev

### DIFF
--- a/CondFormats/RecoMuonObjects/interface/DYTThrObject.h
+++ b/CondFormats/RecoMuonObjects/interface/DYTThrObject.h
@@ -1,3 +1,6 @@
+#ifndef CondFormats_RecoMuonObjects_DYTThrObject_h
+#define CondFormats_RecoMuonObjects_DYTThrObject_h
+
 #include "DataFormats/DetId/interface/DetId.h"
 #include <vector>
 #include "CondFormats/Serialization/interface/Serializable.h"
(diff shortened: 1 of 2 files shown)


vgvassilev:
Part of #15248

davidlange6:
please test

mrodozov:
@oshadura is it because of this missing the 5600 is failing ? in other words should we test 5600 along with this by default ?

oshadura:
@mrodozov yup! it should go together, sorry for late notice!

davidlange6:
the 'requires external' label isn't really appropriate here... the external PR requires this PR, but this PR does not require the external PR.

davidlange6:
please test

pohsun:
+1

davidlange6:
hi @ggovi, all - any further comments on this PR?

davidlange6:
seems not - could we move on with this PR?

silviodonato:
@ggovi I'm going to integrate this PR in the next IB. Please let me know if you have objections.

silviodonato:
merge
In case you have any comments, please let me know

PR #29095, opened 2020-03-04
[DQMOffline/Trigger] improved DQM plugin for monitoring of HLT path/filter efficiencies

missirol:
#### PR description:

This PR is meant as an improvement of the DQM plugin used to monitor the efficiencies of all HLT paths in a given HLT-Menu, as well as the efficiencies of the individual filters making up a given (HLT) path.

Currently, this monitoring is done by the plugin `GeneralHLTOffline`, which builds a set of histograms (`TH1F`s) counting how many times a given trigger fired, and how many times a given filter was executed.

The PR replaces `GeneralHLTOffline` with a new plugin named `HLTFiltersDQMonitor`, which serves the same purpose, but produces DQM outputs in a slightly different format.
In `HLTFiltersDQMonitor`, `TProfile`s are used to obtain efficiencies, of either a trigger path or an individual filter; this (i.e. using efficiencies via `TProfile`s instead of counts via `TH1F`s) allows to make comparisons straightforward between runs/samples with different number of events; using `TProfile`s also allows to have the efficiency estimates directly, i.e. without any dedicated harvesting of these DQM outputs (thanks to @schneiml for the suggestion to use `TProfile`s). More specifically, the new DQM outputs (under `HLT/Filters/`) include
* one plot with the efficiencies of every path in the HLT menu
* for every primary dataset (PD) associated to physics-related streams, a directory which contains (1) a plot with the efficiencies of the paths associated to that PD, and (2) a plot for every path, showing the efficiency of each individual filter making up the path.

Since the PR changes the format of these DQM outputs (from counts to efficiencies), the comparisons will not be very meaningful; we will instead have to check the new outputs and verify that they are in agreement with expectations (an apples-to-apples comparison to the outputs of `GeneralHLTOffline` would not be possible anyway, because of a couple of fixes
included in the new plugin, with respect to the old one).

The new plugin was validated with a number of local tests with MC, but it will be useful to check what happens when testing it on a larger scale during the PR review.

The changes in `test/` are mostly just renaming, for consistency, to accommodate for the new plugin (most of the configs in `test/` seem fairly old anyway).

FYI @Sam-Harper @gennai

#### PR validation:

Successfully ran a few standard workflows, e.g. `runTheMatrix.py -l 11634.0`.


jfernan2:
please test

missirol:
@jfernan2

I think the 2 lines removed in the last commit were previously having no effect (`hltResults` was an instance of `GeneralHLTOffline`, which did not have members named `RecHitsEBTag` and `RecHitsEBTag`).

jfernan2:
please test

Sam-Harper:
thank you @missirol!

missirol:
fwiw, I didn't catch something obviously incorrect in the new outputs; the efficiency values themselves are harder to judge because of the limited stats, but things look as expected.

jfernan2:
+1

silviodonato:
+1

PR #29097, opened 2020-03-04
DQM: Switch L1TStage2EMTF to DQMOneEDAnalyzer.

schneiml:
#### PR description:

... to avoid issue #29096.

@benkrikler @rekovic This should prevent the crash, but the code definitely needs some review. Doing a `TH1::Divide` on every event does not seem like a great idea, and in general such operations should happen in harvesting (or be avoided entierely, e.g. by using `TProfile`).

#### PR validation:
None so far.

Changing to `DQMOneEDAnalyzer` should completely return the behaviour to what it was before #28813 .


schneiml:
+1

Assuming a proper fix will not happen any time soon.

silviodonato:
+1

Issue #27343, opened 2019-06-26
Add a Production Cut on Proton equal to the one on Electron for MTD, Tracker and CTPPS


ianna:
See #27299

kpedro88:
assign geometry,simulation

kpedro88:
@ianna what about HGCal?

kpedro88:
HGCal is already addressed in the aforementioned PR

fabiocos:
I do not fully follow the logic choice of detectors to be updated first, anyway with this PR we can track is left to be done.

@vargasa @jan-kaspar please see the needs of Tracker and CTPPS

ianna:
@cvuosalo - The issue is still relevant. The 2021 scenario does not define a production cut on Protons, but a 2018 does. So, it could be undefined for a sub detector if a sub detector does not define it. In reality the sim code sets an undefined proton cut equal to electron. Perhaps, it is sufficient? 

@vargasa , @jan-kaspar - FYI

jan-kaspar:
Thanks @ianna ! I forward to @mundim who is responsible for the PPS simulation. Also FYI @fabferro

ianna:
@jan-kaspar - please, let me know if there are any news on CTPPS. 
@cvuosalo - I suppose, the CTPPS should be integrated in the DB payload?

mundim:
Sorry, somehow I overlooked this thread. I was not aware we have not defined production cuts for 2021. Since the simulation geometry (and all related files) are basically copied from 2018, the cuts should also be the same. However, I have tested the 2021 geometry (including the one in the 112 GT) and it worked fine. But we need to make some more detailed studies about the impact of such cuts.

vargasa:
Due to time constraints I haven't been able to look into this. FYI @mmusich

fabiocos:
@ianna as far as MTD is concerned, this issues should be solved now.

civanch:
+1

PR #29100, opened 2020-03-04
MTD ETL TDR Geometry update of structure

icosivi:
#### PR description:

A DD4hep version of the `DDMTDLinear` plugin is implemented. Such plugin is used to algorithmically place volumes on ETL `DiscSectors`.
The logical parts `LGAD_2`, `LGAD_r` and `LGAD_r_2` are removed from etl.xml: `LGAD_2` is identical to the `LGAD` logical part therefore it is not needed; whereas `LGAD_r` and `LGAD_r_2` are the LGAD volumes on the back face of the ETL disks and they can be replaced by an `LGAD` rotated by 180 degrees around the y-axis.
The name of the LGAD active volume is changed from `LGAD_ActiveThickness` to `LGAD_TimingActive` to be compatible with BTL definition.
The names of the `SensorModule` logical parts are changed to better define their position within an ETL `DiscSector`:
`SensorModule` --> `SensorModule_Front_Right`
`SensorModule_2` --> `SensorModule_Front_Left`
`SensorModule_r` --> `SensorModule_Back_Left`
`SensorModule_r_2` --> `SensorModule_Back_Right`

ETL-TDR_Front-Face

#### PR validation:

`testMTDinDDD.py` and `testMTDinDD4hep.py` allow comparing the DDD and DD4hep geometries. The dump geometries are identical except from the signs of the rotation matrix components (-0.000000 in DDD, 0.000000 in DD4hep).

The new geometry with rotated LGAD volumes on the back faces of the ETL disks is equal to the previous one : this was tested with `testMTDinDDD.py`, veryfiing that volumes' center are unchanged.

The volume-DataID association has changed because we changed the `SensorModule` names. The ideal numbering scheme has been adapted to that.

### DIFF
--- a/Geometry/MTDSimData/data/v2/mtdProdCuts.xml
+++ b/Geometry/MTDSimData/data/v2/mtdProdCuts.xml
@@ -3,23 +3,23 @@
 <DDDefinition xmlns="http://www.cern.ch/cms/DDL" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.cern.ch/cms/DDL ../../../DetectorDescription/Schema/DDLSchema.xsd">
 
   <SpecParSection label="regions.xml" eval="true">
-    <SpecPar name="fasttime">                                                  
-      <PartSelector path="//BarrelTimingLayer"/>                          
-      <PartSelector path="//EndcapTimingLayer"/>                          
-      <Parameter name="CMSCutsRegion" value="FastTimerRegion" eval="false"/>   
-      <Parameter name="ProdCutsForElectrons" value="0.1*mm"/>                  
-      <Parameter name="ProdCutsForPositrons" value="0.1*mm"/>                  
-      <Parameter name="ProdCutsForGamma" value="0.1*mm"/>                      
-    </SpecPar>                                                                                  
+    <SpecPar name="fasttime">
+      <PartSelector path="//BarrelTimingLayer"/>
+      <PartSelector path="//EndcapTimingLayer"/>
+      <Parameter name="CMSCutsRegion" value="FastTimerRegion" eval="false"/>
+      <Parameter name="ProdCutsForElectrons" value="0.1*mm"/>
+      <Parameter name="ProdCutsForPositrons" value="0.1*mm"/>
+      <Parameter name="ProdCutsForGamma" value="0.1*mm"/>
+    </SpecPar>
     <SpecPar name="fasttime-sens">
       <PartSelector path="//BModule1Layer1Timingactive"/>
       <PartSelector path="//BModule17Layer1Timingactive"/>
       <PartSelector path="//BModule33Layer1Timingactive"/>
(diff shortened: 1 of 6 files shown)


cvuosalo:
+1

fabiocos:
please test

changes should not be related to this PR

Geometry/MTDSimData/data/v2/mtdProdCuts.xml
```diff
@@ -3,23 +3,23 @@
...
-      <PartSelector path="//LGAD_ActiveThickness"/>
-      <Parameter name="CMSCutsRegion" value="FastTimerRegion" eval="false"/>   
-      <Parameter name="ProdCutsForElectrons" value="0.1*mm"/>                  
-      <Parameter name="ProdCutsForPositrons" value="0.1*mm"/>                  
-      <Parameter name="ProdCutsForGamma" value="0.1*mm"/>                      
+      <PartSelector path="//LGAD_TimingActive"/>
+      <Parameter name="CMSCutsRegion" value="FastTimerRegion" eval="false"/>
+      <Parameter name="ProdCutsForElectrons" value="0.1*mm"/>
+      <Parameter name="ProdCutsForPositrons" value="0.1*mm"/>
+      <Parameter name="ProdCutsForGamma" value="0.1*mm"/>
```
ianna: what about adding a production cut on Proton equal to the one on Electron: see issue https://github.com/cms-sw/cmssw/issues/27343?

ianna (changes requested):
@fabiocos - wouldn't it be better to create new versions on the xml files?

kpedro88:
@ianna these XML files are only used in the development geometry I12, which is not part of a central production yet. I would prefer not to spawn repeated versions of development geometries. We should allow developers to build up new geometries before enforcing strict versioning.

kpedro88:
+upgrade

Geometry/MTDSimData/data/v2/mtdProdCuts.xml
fabiocos: @ianna thanks for reminding me about this issue, I will update the production cuts in a next round, where I also update the name of some volumes to make the RECO geometry management easier

fabiocos:
@kpedro88 I agree in general with your comment, I believe anyway that scenario D53 is not frozen and that we may further update it as it is strictly speaking "work in progress". In this sense, I would also like to have from @civanch a confirmation that the above production cuts are still meaningful for a silicon sensor that now is 50 microns thick, as the LGAD active area is. I see that tracker for phaseII is moving to a 0.01 mm cut, allowing for secondary production even within very thin layers. I wonder whether we should do the same here for the new ETL (while for BTL, LYSO crystals, the present value looks reasonable).

silviodonato:
please test workflow 23234.1001, 23434.1001

fabiocos:
The observed difference in DQM summary cannot be due to this PR

silviodonato:
+1

PR #29111, opened 2020-03-05
CSC migration to DD4HEP

slomeo:
#### PR description:

CSC migration to DD4HEP:

This PR covers Sim / Numbering / Validation
No editing for Reco files
It doesn't cover Sensitive Detectors.
There are not overlaps for CSC (All Muon Overlaps were fixed in old PRs related to Scenario 2021 and 2026)

Files changed:

Geometry/CSCGeometryBuilder/plugins/CSCGeometryESModule.cc
Geometry/CSCGeometryBuilder/plugins/CSCGeometryESModule.h
Geometry/CSCGeometryBuilder/src/CSCGeometryBuilderFromDDD.cc
Geometry/CSCGeometryBuilder/src/CSCGeometryBuilderFromDDD.h
Geometry/CSCGeometryBuilder/src/CSCGeometryParsFromDD.cc
Geometry/CSCGeometryBuilder/src/CSCGeometryParsFromDD.h

New files adedd:

Geometry/CSCGeometryBuilder/plugins/CSCGeometryValidate.cc
Geometry/MuonNumbering/src/DD4hep_CSCNumberingScheme.cc
Geometry/MuonNumbering/interface/DD4hep_CSCNumberingScheme.h

Validation made by scripts present in Geometry/CSCGeometryBuilder/test/python (i.e see pdf files attached)
 
[CSC5.pdf](https://github.com/cms-sw/cmssw/files/4292852/CSC5.pdf)
[CSC6.pdf](https://github.com/cms-sw/cmssw/files/4292853/CSC6.pdf)
[CSC7.pdf](https://github.com/cms-sw/cmssw/files/4292855/CSC7.pdf)
[CSC8.pdf](https://github.com/cms-sw/cmssw/files/4292856/CSC8.pdf)
[CSC9.pdf](https://github.com/cms-sw/cmssw/files/4292858/CSC9.pdf)
[CSC10.pdf](https://github.com/cms-sw/cmssw/files/4292859/CSC10.pdf)
[CSC11.pdf](https://github.com/cms-sw/cmssw/files/4292860/CSC11.pdf)
[CSC12.pdf](https://github.com/cms-sw/cmssw/files/4292861/CSC12.pdf)
[CSC13.pdf](https://github.com/cms-sw/cmssw/files/4292862/CSC13.pdf)
[CSC1.pdf](https://github.com/cms-sw/cmssw/files/4292846/CSC1.pdf)
[CSC2.pdf](https://github.com/cms-sw/cmssw/files/4292847/CSC2.pdf)
[CSC3.pdf](https://github.com/cms-sw/cmssw/files/4292849/CSC3.pdf)
[CSC4.pdf](https://github.com/cms-sw/cmssw/files/4292851/CSC4.pdf)


mrodozov:
@slomeo is this root file a leftover ? 
```
The following packages do not have a category, yet:
cmsRecoGeom-2021.root
```

slomeo:
Probably do I need a "git squash"?

civanch:
@slomeo , better to close this and make a new one without root file. Another variant : 

git rm yourfile.root
git commit -m "remove root file"
git push ....

slomeo:
I tried again...

slomeo:
I think now there is no root file anymore

cvuosalo:
@slomeo Please edit the PR description to state what parts of the CSC DD4hep migration are completed in this PR. If some migration steps are still to be done, that would be helpful to mention in the description.

cvuosalo:
@cmsbuild please test

Geometry/CSCGeometryBuilder/plugins/CSCGeometryESModule.cc
```diff
@@ -11,10 +21,16 @@
...
+#include "Geometry/MuonNumbering/interface/DD4hep_MuonNumbering.h"
+#include "Geometry/MuonNumbering/interface/MuonDDDNumbering.h"
+#include "Geometry/MuonNumbering/interface/MuonBaseNumber.h"
+
 using namespace edm;
 
 CSCGeometryESModule::CSCGeometryESModule(const edm::ParameterSet& p)
-    : alignmentsLabel_(p.getParameter<std::string>("alignmentsLabel")),
+    : useDDD_(p.getParameter<bool>("useDDD")),
+      useDD4hep_{p.getUntrackedParameter<bool>("useDD4hep", false)},
```
ianna: it has to be tracked, I think

Geometry/CSCGeometryBuilder/plugins/CSCGeometryESModule.cc
```diff
@@ -1,4 +1,14 @@
...
+/*
+// \class CSCGeometryESModule
+//
+//  Description: CSC ESModule for DD4hep
+//              
+//
+// \author Sergio Lo Meo (sergio.lo.meo@cern.ch) following what Ianna Osburne made for DTs (DD4HEP migration)
+//         Created:  Thu, 05 March 2020 
+//   
+//         Old DD version author: Tim Cox
```
ianna: I'd say original author :-)

Geometry/CSCGeometryBuilder/plugins/CSCGeometryESModule.h
```diff
@@ -16,10 +21,10 @@
 #include <Geometry/Records/interface/MuonGeometryRecord.h>
 #include <Geometry/CSCGeometry/interface/CSCGeometry.h>
 #include "DetectorDescription/Core/interface/DDCompactView.h"
-
+#include <DetectorDescription/DDCMS/interface/DDCompactView.h>
```
ianna: please, use `""`

Geometry/CSCGeometryBuilder/plugins/CSCGeometryESModule.h
```diff
@@ -16,10 +21,10 @@
...
 #include "Geometry/Records/interface/IdealGeometryRecord.h"
 #include "Geometry/Records/interface/MuonNumberingRecord.h"
-
+#include <Geometry/MuonNumbering/interface/DD4hep_MuonNumbering.h>
```
ianna: please, use `""`

Geometry/CSCGeometryBuilder/plugins/CSCGeometryValidate.cc
```diff
@@ -0,0 +1,358 @@
...
+#include "FWCore/Framework/interface/EventSetup.h"
+#include "FWCore/Framework/interface/ESHandle.h"
+#include "FWCore/MessageLogger/interface/MessageLogger.h"
+#include "FWCore/ParameterSet/interface/ParameterSet.h"
+
+#include "Geometry/CSCGeometry/interface/CSCGeometry.h"
+#include "Geometry/CSCGeometry/interface/CSCChamber.h"
+#include "Geometry/CSCGeometry/interface/CSCLayer.h"
+#include "Geometry/CSCGeometry/interface/CSCLayerGeometry.h"
+#include <Geometry/CSCGeometry/interface/CSCStripTopology.h>
```
ianna: please, use `""`

Geometry/CSCGeometryBuilder/plugins/CSCGeometryValidate.cc
```diff
@@ -0,0 +1,358 @@
...
+#include <Geometry/CSCGeometry/interface/CSCWireTopology.h>
```
ianna: please, use `""`

Geometry/CSCGeometryBuilder/plugins/CSCGeometryValidate.cc
```diff
@@ -0,0 +1,358 @@
...
+#include <TH1.h>
+
+#include <limits>
+#include <string>
+#include <type_traits>
+#include <algorithm>
+
+using namespace std;
+
+template <class T>
```
ianna: this is not used here, please, delete from line #42 to line #49

Geometry/CSCGeometryBuilder/src/CSCGeometryBuilderFromDDD.cc
```diff
@@ -1,18 +1,34 @@
...
+*/
+//
 #include "CSCGeometryBuilderFromDDD.h"
 #include "CSCGeometryBuilder.h"
 #include "CSCGeometryParsFromDD.h"
 
 #include <CondFormats/GeometryObjects/interface/RecoIdealGeometry.h>
 #include <CondFormats/GeometryObjects/interface/CSCRecoDigiParameters.h>
 
+#include <DetectorDescription/DDCMS/interface/DDFilteredView.h>
```
ianna: please, use `""`

Geometry/CSCGeometryBuilder/src/CSCGeometryParsFromDD.cc
```diff
@@ -17,6 +28,17 @@
 
 #include <FWCore/MessageLogger/interface/MessageLogger.h>
 
+#include <DetectorDescription/DDCMS/interface/DDFilteredView.h>
```
ianna: same here for all includes

Geometry/CSCGeometryBuilder/src/CSCGeometryParsFromDD.cc
```diff
@@ -159,17 +183,56 @@ bool CSCGeometryParsFromDD::build(const DDCompactView* cview,
...
+
     fpar.clear();
-    //    dpar = fv.logicalPart().solid().parameters();
+
     if (fv.logicalPart().solid().shape() == DDSolidShape::ddsubtraction) {
       const DDSubtraction& first = fv.logicalPart().solid();
       const DDSubtraction& second = first.solidA();
       const DDSolid& third = second.solidA();
       dpar = third.parameters();
+      if (std::abs(dpar[0]) < 1.0e-7)
```
ianna: please, use:
```c++
#include "DataFormats/Math/interface/Rounding.h"
```
and replace lines from #197 to #215 with:
```c++
std::transform(dpar.begin(), dpar.end(), dpar.begin(),
                   [](double i) -> double { return cms_rounding::roundIfNear0(i); });
```

Geometry/CSCGeometryBuilder/src/CSCGeometryParsFromDD.cc
```diff
@@ -159,17 +183,56 @@ bool CSCGeometryParsFromDD::build(const DDCompactView* cview,
...
+      if (std::abs(dpar[6]) < 1.0e-7)
+        dpar[6] = 0;
+      if (std::abs(dpar[7]) < 1.0e-7)
+        dpar[7] = 0;
+      if (std::abs(dpar[8]) < 1.0e-7)
+        dpar[8] = 0;
+
     } else {
       dpar = fv.logicalPart().solid().parameters();
+      if (std::abs(dpar[0]) < 1.0e-7)
```
ianna: the same here

Geometry/CSCGeometryBuilder/src/CSCGeometryParsFromDD.cc
```diff
@@ -239,9 +302,41 @@ bool CSCGeometryParsFromDD::build(const DDCompactView* cview,
 
         // detid is for ME11 and that's what we're using for ME1b in the software
 
+        if (std::abs(gtran[0]) < 1.0e-7)
```
ianna: and again, the same here

Geometry/CSCGeometryBuilder/src/CSCGeometryParsFromDD.cc
```diff
@@ -263,6 +358,39 @@ bool CSCGeometryParsFromDD::build(const DDCompactView* cview,
             uparvals.begin() + noOfAnonParams + 1, uparvals.begin() + (2 * noOfAnonParams) + 2, uparvals.begin() + 1);
 
         CSCDetId detid1a = CSCDetId(jendcap, 1, 4, jchamber, 0);  // reset to ME1A
+        if (std::abs(gtran[0]) < 1.0e-7)
```
ianna: please, use lambda as advised above

Geometry/CSCGeometryBuilder/src/CSCGeometryParsFromDD.cc
```diff
@@ -284,9 +411,41 @@ bool CSCGeometryParsFromDD::build(const DDCompactView* cview,
         }
 
       } else {
+        if (std::abs(gtran[0]) < 1.0e-7)
```
ianna: and here

Geometry/CSCGeometryBuilder/src/CSCGeometryParsFromDD.cc
```diff
@@ -301,3 +460,366 @@ bool CSCGeometryParsFromDD::build(const DDCompactView* cview,
...
+    CSCWireGroupPackage wg;
+    uparvals.clear();
+
+    // if the specs are made no need to get all this crap!
+    int chamberType = CSCChamberSpecs::whatChamberType(jstation, jring);
+
+    size_t ct = 0;
+    bool chSpecsAlreadyExist = false;
+
+    for (; ct < rdp.pChamberType.size(); ++ct) {
```
ianna: this loop doesn't do anything

Geometry/CSCGeometryBuilder/src/CSCGeometryParsFromDD.cc
```diff
@@ -301,3 +460,366 @@ bool CSCGeometryParsFromDD::build(const DDCompactView* cview,
...
+           ++it) {
+        uparvals.emplace_back(*it);
+      }
+      for (CSCWireGroupPackage::Container::const_iterator it = wg.consecutiveGroups.begin();
+           it != wg.consecutiveGroups.end();
+           ++it) {
+        uparvals.emplace_back(*it);
+      }
+
+      /** end crap **/
```
ianna: hmm...?

Geometry/CSCGeometryBuilder/src/CSCGeometryParsFromDD.cc
```diff
@@ -301,3 +460,366 @@ bool CSCGeometryParsFromDD::build(const DDCompactView* cview,
...
+    int jstation = detid.station();
+    int jring = detid.ring();
+    int jchamber = detid.chamber();
+    int jlayer = detid.layer();
+
+    // Package up the wire group info as it's decoded
+    CSCWireGroupPackage wg;
+    uparvals.clear();
+
+    // if the specs are made no need to get all this crap!
```
ianna: Hmmm... please, use more descriptive comments

Geometry/CSCGeometryBuilder/src/CSCGeometryParsFromDD.cc
```diff
@@ -301,3 +460,366 @@ bool CSCGeometryParsFromDD::build(const DDCompactView* cview,
...
+
+    fpar.clear();
+
+    std::string my_title(fv.solid()->GetTitle());
+
+    if (my_title == "Subtraction") {
+      cms::DDSolid mysolid(fv.solid());
+      auto solidA = mysolid.solidA();
+      std::vector<double> dpar = solidA.dimensions();
+      if (std::abs(dpar[0]) < 1.0e-7)
```
ianna: please, use lambda

Geometry/CSCGeometryBuilder/src/CSCGeometryParsFromDD.cc
```diff
@@ -301,3 +460,366 @@ bool CSCGeometryParsFromDD::build(const DDCompactView* cview,
...
+      if (std::abs(dpar[8]) < 1.0e-7)
+        dpar[8] = 0;
+
+      fpar.emplace_back((dpar[1]));
+      fpar.emplace_back((dpar[2]));
+      fpar.emplace_back((dpar[3]));
+      fpar.emplace_back((dpar[4]));
+    } else {
+      dpar = fv.parameters();
+      if (std::abs(dpar[0]) < 1.0e-7)
```
ianna: lambda, please

Geometry/CSCGeometryBuilder/src/CSCGeometryParsFromDD.cc
```diff
@@ -301,3 +460,366 @@ bool CSCGeometryParsFromDD::build(const DDCompactView* cview,
...
+      for (size_t i = 0; i < wg.consecutiveGroups.size(); i++) {
+      }
+    } else {
+      LogTrace(myName) << myName << " wg.numberOfGroups == 0 ";
+    }
+
+    if (jlayer == 0) {  // Can only build chambers if we're filtering them
+
+      if (jstation == 1 && jring == 1) {
+        if (std::abs(gtran[0]) < 1.0e-7)
```
ianna: lambda, please

Geometry/CSCGeometryBuilder/src/CSCGeometryParsFromDD.cc
```diff
@@ -301,3 +460,366 @@ bool CSCGeometryParsFromDD::build(const DDCompactView* cview,
...
+        if (noOfAnonParams == 0) {
+          noOfAnonParams = kNoOfAnonParams;
+        }  // in case it wasn't seen
+
+        // copy ME1a params from back to the front
+        std::copy(
+            uparvals.begin() + noOfAnonParams + 1, uparvals.begin() + (2 * noOfAnonParams) + 2, uparvals.begin() + 1);
+
+        CSCDetId detid1a = CSCDetId(jendcap, 1, 4, jchamber, 0);  // reset to ME1A
+        if (std::abs(gtran[0]) < 1.0e-7)
```
ianna: lambda, please

Geometry/CSCGeometryBuilder/src/CSCGeometryParsFromDD.cc
```diff
@@ -301,3 +460,366 @@ bool CSCGeometryParsFromDD::build(const DDCompactView* cview,
...
+          LogTrace(myName) << myName << " ct < rdp.pChamberType.size() && rdp.pChamberType[ct] == chtypeA ";
+        } else {
+          rdp.pChamberType.emplace_back(chtypeA);
+          rdp.pUserParOffset.emplace_back(rdp.pfupars.size());
+          rdp.pUserParSize.emplace_back(uparvals.size());
+          std::copy(uparvals.begin(), uparvals.end(), std::back_inserter(rdp.pfupars));
+        }
+
+      } else {
+        if (std::abs(gtran[0]) < 1.0e-7)
```
ianna: lambda, please

Geometry/MuonNumbering/interface/DD4hep_CSCNumberingScheme.h
```diff
@@ -0,0 +1,43 @@
...
+    int getDetId() const { return detId; }
+
+  private:
+    const int get(const char*, const MuonConstants&) const;
+    void initMe(const MuonConstants& muonConstants);
+    void setDetId(int idnew) { detId = idnew; }
+
+    int chamberIndex(int, int, int, int) const;
+
+    int theRegionLevel;
```
ianna: please, use consistent naming conventions. You've used `name_` everywhere else. Here you've started with `theName`. Please, choose one or another and modify the code.

Geometry/MuonNumbering/src/CSCNumberingScheme.cc
```diff
@@ -4,7 +4,7 @@
 #include "Geometry/MuonNumbering/interface/MuonDDDConstants.h"
 #include "FWCore/MessageLogger/interface/MessageLogger.h"
 
-//#define LOCAL_DEBUG
+#define LOCAL_DEBUG
```
ianna: please, remove

ianna (changes requested):

Geometry/CSCGeometryBuilder/plugins/CSCGeometryESModule.cc
slomeo: Hi Yana, I made the same for GEM migration. It is needed to use Untracked Parameter for dd4hep in order to have no crashes during runTheMatrix tests

Geometry/CSCGeometryBuilder/plugins/CSCGeometryESModule.cc
slomeo: Done!

Geometry/CSCGeometryBuilder/plugins/CSCGeometryESModule.h
slomeo: Done

Geometry/CSCGeometryBuilder/plugins/CSCGeometryESModule.h
slomeo: Done

Geometry/CSCGeometryBuilder/plugins/CSCGeometryValidate.cc
slomeo: Done

Geometry/CSCGeometryBuilder/plugins/CSCGeometryValidate.cc
slomeo: Done

Geometry/CSCGeometryBuilder/plugins/CSCGeometryValidate.cc
slomeo: They are necessary for: "if (almost_equal(val1, val2, tolerance_))" and "name.c_str()" (in THD1)

Geometry/CSCGeometryBuilder/src/CSCGeometryBuilderFromDDD.cc
slomeo: Done

Geometry/CSCGeometryBuilder/src/CSCGeometryParsFromDD.cc
slomeo: Done

Geometry/CSCGeometryBuilder/src/CSCGeometryParsFromDD.cc
slomeo: Done

Geometry/CSCGeometryBuilder/src/CSCGeometryParsFromDD.cc
slomeo: Done

Geometry/CSCGeometryBuilder/src/CSCGeometryParsFromDD.cc
slomeo: Done

Geometry/CSCGeometryBuilder/src/CSCGeometryParsFromDD.cc
slomeo: Done

Geometry/CSCGeometryBuilder/src/CSCGeometryParsFromDD.cc
slomeo: Done

Geometry/CSCGeometryBuilder/src/CSCGeometryParsFromDD.cc
slomeo: It is present in the DD side of the code too. There is an "if" command and  a "break" command inside the loop. I don't know if it is right to delete these lines

Geometry/CSCGeometryBuilder/src/CSCGeometryParsFromDD.cc
slomeo: This a comment is taken from the DD side of the code. Maybe is it better to leave it?

Geometry/CSCGeometryBuilder/src/CSCGeometryParsFromDD.cc
slomeo: This a comment is taken from the DD side of the code. Maybe is it better to leave it?

Geometry/MuonNumbering/src/CSCNumberingScheme.cc
slomeo: For GEM migration (#28796), when I commented this line there was a problem (see mail of @fabiocos 11/02/2020)

Geometry/MuonNumbering/interface/DD4hep_CSCNumberingScheme.h
slomeo: I followed the convention I found in CSCNumberingScheme.cc. I think it is a kind of continuity with the DD format originally developed

ptcox:
I agree with Yana and suggest you just replace 'crap' by 'stuff' in the 
5 places it occurs. Clearly a bad day for Mike Case (who wrote the DD 
interface) :)

Tim

Geometry/CSCGeometryBuilder/src/CSCGeometryParsFromDD.cc
slomeo: Done

Geometry/CSCGeometryBuilder/src/CSCGeometryParsFromDD.cc
slomeo: Done

Geometry/CSCGeometryBuilder/src/CSCGeometryParsFromDD.cc
slomeo: Done

Geometry/CSCGeometryBuilder/src/CSCGeometryParsFromDD.cc
slomeo: Done

Geometry/CSCGeometryBuilder/src/CSCGeometryParsFromDD.cc
slomeo: Done

ptcox:
I guess I don't understand the way the code is supposed to work. It 
looks like ct should be incremented in the loop and then the code
   if (ct < rdp.pChamberType.size() && rdp.pChamberType[ct] == 
chamberType) {
       chSpecsAlreadyExist = true;
     } else {
...
get executed. But right now it clearly doesn't.

Tim

Geometry/MuonNumbering/src/CSCNumberingScheme.cc
slomeo: I'm sorry, I misunderstood the comment. Now there is only #define LOCAL_DEBUG

slomeo:
done

slomeo:
Done

> I agree with Yana and suggest you just replace 'crap' by 'stuff' in the 5 places it occurs. >Clearly a bad day for Mike Case (who wrote the DD interface) :) Tim
> […](#)

ptcox:
I agree, this do-nothing loop appears in the original code too. In fact 
I guess what happens is that ct comes out of the loop with value size() 
-1. Then most of the time the following 'if' fails and the 'else' block 
just gets executed a lot more times than it needs to be.
Looks like a long-standing bug that didn't break anything, but caused 
more code execution than is strictly required?

Tim

slomeo:
Do you want that I try to fix this loop? Any suggestions?

> I agree, this do-nothing loop appears in the original code too. In fact I guess what happens is that ct comes out of the loop with value size() -1. Then most of the time the following 'if' fails and the 'else' block just gets executed a lot more times than it needs to be. Looks like a long-standing bug that didn't break anything, but caused more code execution than is strictly required? Tim

slomeo:
Done (i.e a list of new files and modified files) in the PR description
 
> @slomeo Please edit the PR description to state what parts of the CSC DD4hep migration are completed in this PR. If some migration steps are still to be done, that would be helpful to mention in the description.

ptcox:
I was not the author of this code, ever. The DD interface was written by 
Mike Case. I wrote the version before the DDD, and the DD version is 
cryptic to me :(

I suggest you leave it for now. It looks like it should be 'easy' to 
move the end of that currently dead-loop to wherever it should be. But I 
don't understand the details well enough at the moment to say where 
exactly that is! Likewise for the ME1/1A code at the end. The same 
'dead' loop is there too. But does it only go through there once anyway? 
Otherwise it would seem chtypeA would get added multiple times.  I'd 
have to spend time reading the code to try to understand. I'll look at 
it, but I can't right this minute.

Regards, Tim

ianna:
> I agree with Yana and suggest you just replace 'crap' by 'stuff' in the 5 places it occurs. Clearly a bad day for Mike Case (who wrote the DD interface) :) Tim
> […](#)

Oh... we ought to rewrite this code if this is Mike's "stuff". Clearly, it was a temporary fix nearly 20 years ago.  I'd think, we can do better as the comment suggests. That's why CSC builder was so different and difficult to port :-)

Geometry/CSCGeometryBuilder/plugins/CSCGeometryESModule.cc
makortel: Strictly speaking tracked vs. untracked should not lead to crashes (or, crashes imply that something else is wrong). Where the failures perhaps complaining a missing parameter? In that case the best solution would be to implement `fillDescriptions()` to generated and validate the configuration fragment.

slomeo: It is not a crash, but test failure 
(i.e ----- Begin Fatal Exception 05-Mar-2020 18:40:27 CET-----------------------
   [1] Constructing ESProducer: class=CSCGeometryESModule label=''
Exception Message:
MissingParameter: Parameter 'useDD4hep' not found.
----- End Fatal Exception -------------------------------------------------)


 > Strictly speaking tracked vs. untracked should not lead to crashes (or, crashes imply >that something else is wrong). Where the failures perhaps complaining a missing >parameter? In that case the best solution would be to implement `fillDescriptions()` to >generated and validate the configuration fragment.

makortel: Thanks, so the complaint is about missing parameter. The best solutions are to either implement `fillDescriptions()` (preferred), or propagate the added parameter to all definitions.

cvuosalo:
@slomeo Sorry, I didn't say clearly what I wanted. The list of files isn't needed, since that is already part of the PR. What I wanted to know is what parts of the migration this PR addresses. It looks like it covers Reco / Sim Numbering and Validation. It does not fix CSC overlaps nor migrate Sensitive Detectors, so those steps are still to be done.

It would be helpful if the description said:
"This PR completes Reco / Sim Numbering and Validation for CSC DD4hep migration. Some CSC migration tasks are still to be done."

Of course, if you think more work needs to be done for numbering and validation, you should modify the statement accordingly.

> Done (i.e a list of new files and modified files) in the PR description
> 
> > @slomeo Please edit the PR description to state what parts of the CSC DD4hep migration are completed in this PR. If some migration steps are still to be done, that would be helpful to mention in the description.

Geometry/CSCGeometryBuilder/plugins/CSCGeometryESModule.cc
slomeo: When you write: "...or propagate the added parameter to all definitions." do you mean to set tUntrackedParameter all the parameters? If it so I think this solution is quite good (and quick) 
> Thanks, so the complaint is about missing parameter. The best solutions are to either >implement `fillDescriptions()` (preferred), or propagate the added parameter to all >definitions.

makortel: > When you write: "...or propagate the added parameter to all definitions." do you mean to set tUntrackedParameter all the parameters? If it so I think this solution is quite good (and quick)

No, all parameters affecting results/physics should be tracked. I meant to add the parameter in all places where `cms.ESProducer("CSCGeometryESModule", ...)` is defined. By quick look there does not seem to be that many, but it is in the HLT menu, which would need a customize function in https://github.com/cms-sw/cmssw/blob/master/HLTrigger/Configuration/python/customizeHLTforCMSSW.py

slomeo:
Ok @cvuosalo , this PR covers Sim / Numbering / Validation
I don't edit Reco files
It doesn't cover Sensitive Detectors.
There are not overlaps for CSC (I solved Muon Overlaps in old PRs related to Scenario 2021 and 2026) 


 > @slomeo Sorry, I didn't say clearly what I wanted. The list of files isn't needed, since that is already part of the PR. What I wanted to know is what parts of the migration this PR addresses. It looks like it covers Reco / Sim Numbering and Validation. It does not fix CSC overlaps nor migrate Sensitive Detectors, so those steps are still to be done.
> 
> It would be helpful if the description said:
> "This PR completes Reco / Sim Numbering and Validation for CSC DD4hep migration. Some CSC migration tasks are still to be done."
> 
> Of course, if you think more work needs to be done for numbering and validation, you should modify the statement accordingly.
> 
> > Done (i.e a list of new files and modified files) in the PR description
> > > @slomeo Please edit the PR description to state what parts of the CSC DD4hep migration are completed in this PR. If some migration steps are still to be done, that would be helpful to mention in the description.

Geometry/CSCGeometryBuilder/plugins/CSCGeometryESModule.cc
slomeo: I found 10 files (see attached picture). In all these .py files do I have to set the dd4hep parameter? Do I have to delete useDD4hep_{p.getUntrackedParameter<bool>("useDD4hep", false)} from CSCGeometryESModule.cc? Probably I'll see a compilation error..
Screenshot 2020-03-06 19 32 01

makortel: > In all these .py files do I have to set the dd4hep parameter? 

In the first three (under `Geometry/CSCGeometry*`) yes, the `useDD4hep` parameter would need to be added.

For the next 6 (under `HLTrigger/Configuration`) a customize function should be added in
https://github.com/cms-sw/cmssw/blob/master/HLTrigger/Configuration/python/customizeHLTforCMSSW.py
because the HLT menu dumps should not be edited directly.

The last one looks like it can be ignored (it only modifies an existing module, and the file may well be obsolete).

> Do I have to delete useDD4hep_{p.getUntrackedParameter("useDD4hep", false)} from CSCGeometryESModule.cc?

It should be changed to `useDD4hep_{p.getParameter("useDD4hep")}`.

ptcox:
I took a look at the original code, and I changed my mind. We were all 
misled by the for loop at first glance not appearing to DO anything. But 
it DOES. It identifies if the chamberSpecs object corresponding to the 
specific CSC station and ring being handled has already been 'built'.
-- The  for loop is required in order to check all elements of 
rdp.pChamberType.
-- If the chamberSpecs HAS been built already, the code  breaks out of 
the loop with 'ct' set to the corresponding index in  rdp.pChamberType .
-- If not, then the loop goes to the end.
-- The following  'if'  statement skips building the chamberSpecs again 
if it has already been built.
So I think there's nothing strictly wrong with the code as it is. 
Personally I don't like making use of what looks like in loop variable 
outside the loop. But I guess it is formally ok.
Conclusion: there's nothing wrong with that loop: it is doing something, 
and it looks like the something that should be done :)

Regards, Tim

slomeo:
A question for @civanch @cvuosalo @ianna @makortel @ptcox @fabiocos @mrodozov @kpedro88 (I'm sorry if I'm forgetting someone else): I understood by reading the posts of @makortel that's possible to fix in a formal (good) way the "problem" of a missing parameter for the runTheMatrix test (I fixed it by setting as Untracked the dd4hep parameter in ESProducer as I did for GEM migration) BUT could it not better to go ahead with the review of this PR?  Afterwards, when I'll have been finished with ME0 (my last step with migration for the Geometry of the Muon System), I could think about a refined review of something I skipped doing. I'm asking this for two reasons: 1) a dead line for migration 2) Our colleague Adinda de Vit is waiting for CSC migration in order to continue with her work about Tracker/muon alignment DD4HEP migration. Please let me know.

civanch:
@ianna , @cvuosalo ,

I am a bit lost: why the option should appear in too many files? Can we define one option "useDD4Hep", which propagated to all components? In any case, build of geometry is an action, which performed once in each sub-detector. Even if it is not possible avoiding check of this flag in several classes, the source of this flag (initial fragment) better be unique and the default should be "False" now. When migration will be done it should be "True". If flag is not defined it should be not exception but we should use the default "False". In this sense, "untracted" for this parameter seems to be a more correct solution. In the long term it will be removed.

cvuosalo:
I think the revision of the `useDD4hep` tracked/untracked parameter can wait for a later PR, unless @ianna or @makortel object.

cvuosalo:
@cmsbuild please test

makortel:
> I think the revision of the `useDD4hep` tracked/untracked parameter can wait for a later PR, unless @ianna or @makortel object.

That's fine.

I was not really advocating the use of tracked in this PR (even if I think the parameter should be tracked if non-default value would get used in production), but to describe the necessary steps if it were to made tracked.

Geometry/CSCGeometryBuilder/src/CSCGeometryParsFromDD.cc
cvuosalo: @slomeo Have you pushed all your changes?

Geometry/CSCGeometryBuilder/src/CSCGeometryParsFromDD.cc
```diff
@@ -301,3 +358,272 @@ bool CSCGeometryParsFromDD::build(const DDCompactView* cview,
...
+      std::transform(
+          dpar.begin(), dpar.end(), dpar.begin(), [](double i) -> double { return cms_rounding::roundIfNear0(i); });
+
+      fpar.emplace_back((dpar[0]));
+      fpar.emplace_back((dpar[1]));
+      fpar.emplace_back((dpar[2]));
+      fpar.emplace_back((dpar[3]));
+    }
+
+    gtran[0] = (float)1.0 * (fv.translation().X());
```
cvuosalo:
```suggestion
    gtran[0] = static_cast<float>(fv.translation().X());
```

cvuosalo: If you're trying to reduce the precision of the number, `static_cast` is faster. Otherwise, I don't understand why you would multiply by 1.0, other than to waste CPU.
I'm guessing simple assignment would be correct here.
Whatever change you make, it should be made to the following two lines also.

Geometry/CSCGeometryBuilder/src/CSCGeometryParsFromDD.cc
```diff
@@ -301,3 +358,272 @@ bool CSCGeometryParsFromDD::build(const DDCompactView* cview,
...
+    gtran[1] = (float)1.0 * (fv.translation().Y());
+    gtran[2] = (float)1.0 * (fv.translation().Z());
+
+    if ((-1.0e-10 < gtran[0]) && (gtran[0] < 1.0e-10))
+      gtran[0] = 0.0;
+    if ((-1.0e-10 < gtran[1]) && (gtran[1] < 1.0e-10))
+      gtran[1] = 0.0;
+    if ((-1.0e-10 < gtran[2]) && (gtran[2] < 1.0e-10))
+      gtran[2] = 0.0;
```
cvuosalo: You're already using roundIfNear0(). Use it on lines 514-519 with `1.e-10` as the second parameter.

Geometry/CSCGeometryBuilder/src/CSCGeometryParsFromDD.cc
```diff
@@ -301,3 +358,272 @@ bool CSCGeometryParsFromDD::build(const DDCompactView* cview,
...
+
+    fv.rotation().GetComponents(trm.begin(), trm.end());
+    size_t rotindex = 0;
+    for (size_t i = 0; i < 9; ++i) {
+      grmat[i] = (float)1.0 * trm[rotindex];
```
cvuosalo: Change the `1.0 *` as you decided in the previous case.

Geometry/CSCGeometryBuilder/test/python/validateCSCGeometry2_cfg.py
```diff
@@ -0,0 +1,53 @@
+import FWCore.ParameterSet.Config as cms
+
```
cvuosalo: Call this file `validateCSCGeometryOldDD_cfg.py` or something like that.

Geometry/CSCGeometryBuilder/test/python/validateCSCGeometry_cfg.py
```diff
@@ -0,0 +1,56 @@
+import FWCore.ParameterSet.Config as cms
+
```
cvuosalo: This file should be called `validateCSCGeometryDD4hep_cfg.py`.

Geometry/CSCGeometryBuilder/src/CSCGeometryParsFromDD.cc
slomeo: I missed something. Now it is fixed

Geometry/CSCGeometryBuilder/src/CSCGeometryParsFromDD.cc
slomeo: Done

Geometry/CSCGeometryBuilder/src/CSCGeometryParsFromDD.cc
slomeo: Done

Geometry/CSCGeometryBuilder/src/CSCGeometryParsFromDD.cc
slomeo: Done

Geometry/CSCGeometryBuilder/test/python/validateCSCGeometry_cfg.py
slomeo: Done

Geometry/CSCGeometryBuilder/test/python/validateCSCGeometry2_cfg.py
slomeo: Done

Geometry/CSCGeometryBuilder/plugins/CSCGeometryESModule.cc
fabiocos: please keep in mind that we have a process modifier https://cmssdt.cern.ch/lxr/source/Configuration/ProcessModifiers/python/dd4hep_cff.py that can be used to centralize settings change. This could help in tracking in a transparent way the set of needed modifications

davidlange6:
This is presumably not a parameter we expect to change frequently in the long term - its currently a expert developer switch which will go to always true once complete, no?

On Mar 11, 2020, at 9:35 AM, Fabio Cossutti <notifications@github.com<mailto:notifications@github.com>> wrote:


@fabiocos commented on this pull request.

In Geometry/CSCGeometryBuilder/plugins/CSCGeometryESModule.cc<https://github.com/cms-sw/cmssw/pull/29111#discussion_r390812287>:

>  using namespace edm;

 CSCGeometryESModule::CSCGeometryESModule(const edm::ParameterSet& p)
-    : alignmentsLabel_(p.getParameter<std::string>("alignmentsLabel")),
+    : useDDD_(p.getParameter<bool>("useDDD")),
+      useDD4hep_{p.getUntrackedParameter<bool>("useDD4hep", false)},


please keep in mind that we have a process modifier https://cmssdt.cern.ch/lxr/source/Configuration/ProcessModifiers/python/dd4hep_cff.py that can be used to centralize settings change. This could help in tracking in a transparent way the set of needed modifications

cvuosalo:
@cmsbuild please test

cvuosalo:
+1

fabiocos:
@davidlange6 in my view as soon as the migration is completed, also in terms of validation, the old DDD code should be progressively de-commissioned, so at the end of the process there will be no more special setting to be tracked, But I am afraid this is neither for tomorrow nor for April. So in the meanwhile the modifier may help in managing transparently one setup vs the other.

Geometry/CSCGeometryBuilder/plugins/CSCGeometryESModule.cc
silviodonato: @slomeo I think the easiest way is to set the default value by using `fillDescriptions()`, as suggested by @makortel

slomeo:
the issue of the parameter useDD4hep_ will be fixed in a dedicated PR, please merge #29111 in order to give the opportunity to Adinda De Wit &  Marco Musich keep on their job about Traker/Muon Alignment

silviodonato:
+1

PR #29127, opened 2020-03-05
Convert TrackingMonitor to be a DQMEDAnalyzer

Dr15Jones:
#### PR description:

There is no need to hand reset histograms on Lumi boundaries as the DQMStore now does that automatically for MonitorElements which are marked as per Lumi (which is the case here).
-Put TrackAnalyzer in a namespace to avoid 1 definition rule problems.

#### PR validation:

The code compiles. A visual inspection of the code did not point out any obvious reason this cannot be a stream module.


Dr15Jones:
please test

jfernan2:
+1

silviodonato:
+1

Dr15Jones:
This change has lead to segmentation faults in the IBs. I'm looking right now to see if I can do an easy fix.

Dr15Jones:
The problems are fixed in ##29143

PR #29000, opened 2020-02-19
Changing handling of color indexes and anti particle creation for pythia8 EGun

pasmuss:
#### PR description:
Changes for proper treatment of color (conservation and codes):
- Increasing color index after generating one particle (anti particle obtains the same color, just on the anticolor column)
- Adding a check that if only colored particles are generated, the AddAntiParticle option is used to make sure that all colors are balanced
- Only affecting pythia8 EGun

#### PR validation:
Ran some tests locally with a core CMSSW script:
- not set AddAntiParticle at all (does not start, option must be set)
- set option to False (runs but does not generate anything and throws errors)
- set option to True (as expected: anti particle gets same anti color as particle gets color, index is not the same for two different particles)

#### if this PR is a backport please specify the original PR and why you need to backport that PR:

Before submitting your pull requests, make sure you followed this checklist:
- verify that the PR is really intended for the chosen branch
- verify that changes follow [CMS Naming, Coding, And Style Rules](http://cms-sw.github.io/cms_coding_rules.html)
- verify that the PR passes the basic test procedure suggested in the [CMSSW PR instructions](https://cms-sw.github.io/PRWorkflow.html)

PR #29130, opened 2020-03-06
Backport to 9 2 8; only file that needs to be changed: GeneratorInterface/Pythia8Interface/plugins/Py8EGun.cc

pasmuss:

- Original PR #29000 
- Need backport to CMSSW_9_2_8 for MC production (release of needed campaign)
- Comments from original PU listed below (simply copy pasted)

#### PR description:

Changes for proper treatment of color (conservation and codes):

- Increasing color index after generating one particle (anti particle obtains the same color, just on the anticolor column)
- Adding a check that if only colored particles are generated, the AddAntiParticle option is used to make sure that all colors are balanced
- Only affecting pythia8 EGun

#### PR validation:

Ran some tests locally with a core CMSSW script:

- not set AddAntiParticle at all (does not start, option must be set)
- set option to False (runs but does not generate anything and throws errors)
- set option to True (as expected: anti particle gets same anti color as particle gets color, index is not the same for two different particles)


pasmuss:
Not sure whether this way of backporting is correct, sorry. Do I need to start from the branch I want to merge into in order to avoid all these conflicts?

PR #29141, opened 2020-03-06
Introduce SONIC core in CMSSW

kpedro88:
#### PR description:

This PR introduces the core classes for SONIC as a new package in the HeterogeneousCore subsystem. This code is ported from https://github.com/hls-fpga-machine-learning/SonicCMS. Unit tests are included (based on the Acquire tests for ExternalWork in `FWCore/Integration/test`).

Documentation is provided in the package README, as well as in a [presentation](https://indico.cern.ch/event/892069/) at the core software meeting.

This package will be supplemented by additional packages implementing concrete clients for different communication protocols as the needed externals become available. (A separate package for each communication protocol is foreseen in order to compartmentalize the different external dependencies.)

#### PR validation:

Unit tests run successfully.


makortel:
assign heterogeneous

(until the new packages get added to `categories_map.py`, and I'd like to first agree on the naming)

kpedro88:
please test

makortel:
(as I mentioned in person) I'm not super happy with the name "Sonic" because it is not very descriptive, but I don't have any better concise suggestions either, so to me it is good enough.

HeterogeneousCore/SonicCore/interface/SonicClientBase.h
```diff
@@ -0,0 +1,49 @@
...
+#include "FWCore/MessageLogger/interface/MessageLogger.h"
+
+#include <string>
+#include <chrono>
+#include <exception>
+
+class SonicClientBase {
+public:
+  //destructor
+  virtual ~SonicClientBase() {}
```
fwyzard:
```suggestion
  virtual ~SonicClientBase() = default;
```

HeterogeneousCore/SonicCore/interface/SonicClientBase.h
```diff
@@ -0,0 +1,49 @@
...
+
+  void setDebugName(const std::string& debugName) { debugName_ = debugName; }
+
+  //main operation
+  virtual void predict(edm::WaitingTaskWithArenaHolder holder) = 0;
```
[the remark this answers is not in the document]
fwyzard: in fact, if I understood the logic correctly
  - the framework calls `predict()`
  - the three specialisations (`Sync`, `Async` and `PseudoAsync`) offer different implementations of `predict()`, that eventually call `predictImpl()`
  - user modules are supposed to implement `predictImpl()`

If this is correct, I would suggest to
  - rename `predictImpl()` to `evaluate()` or similar
  - rename `predict()` to `dispatch()` or similar

HeterogeneousCore/SonicCore/interface/SonicClientTypes.h
```diff
@@ -0,0 +1,27 @@
...
+
+//this base class exists to limit the impact of dependent scope in derived classes
+template <typename InputT, typename OutputT = InputT>
+class SonicClientTypes {
+public:
+  //typedefs for outside accessibility
+  typedef InputT Input;
+  typedef OutputT Output;
+  //destructor
+  virtual ~SonicClientTypes() {}
```
fwyzard:
```suggestion
  virtual ~SonicClientTypes() = default;
```

HeterogeneousCore/SonicCore/interface/SonicEDProducer.h
```diff
@@ -0,0 +1,59 @@
...
+template <typename Client, typename... Capabilities>
+class SonicEDProducer : public edm::stream::EDProducer<edm::ExternalWork, Capabilities...> {
+public:
+  //typedefs to simplify usage
+  typedef typename Client::Input Input;
+  typedef typename Client::Output Output;
+  //constructor
+  SonicEDProducer(edm::ParameterSet const& cfg) : client_(cfg.getParameter<edm::ParameterSet>("Client")) {}
+  //destructor
+  virtual ~SonicEDProducer() {}
```
fwyzard:
```suggestion
  virtual ~SonicEDProducer() = default;
```

HeterogeneousCore/SonicCore/interface/SonicClientTypes.h
```diff
@@ -0,0 +1,27 @@
...
+
+  //accessors
+  Input& input() { return input_; }
+  const Input& input() const { return input_; }
+  void setInput(const Input& inp) { input_ = inp; }
```
fwyzard: this is never used ?

HeterogeneousCore/SonicCore/interface/SonicClientTypes.h
```diff
@@ -0,0 +1,27 @@
...
+public:
+  //typedefs for outside accessibility
+  typedef InputT Input;
+  typedef OutputT Output;
+  //destructor
+  virtual ~SonicClientTypes() {}
+
+  //accessors
+  Input& input() { return input_; }
+  const Input& input() const { return input_; }
```
fwyzard: this is not used, AFAICT; is there a use case for inspecting the input value in a `const` context ?

HeterogeneousCore/SonicCore/interface/SonicClientTypes.h
```diff
@@ -0,0 +1,27 @@
...
+  void setInput(const Input& inp) { input_ = inp; }
+  const Output& output() const { return output_; }
+
+protected:
+  Input input_;
```
fwyzard: I understand that these cannot be pointers because this class acts kind of like a buffer between the possibly asynchronous back-end, and the user code that produces the input (and could go out of scope before it is sent to the backend) and consumes the output (that it may not have allocated space for yet). Is it correct ?

HeterogeneousCore/SonicCore/interface/SonicClientBase.h
```diff
@@ -0,0 +1,49 @@
...
+  virtual void predictImpl() = 0;
+
+  void setStartTime() {
+    if (debugName_.empty())
+      return;
+    t0_ = std::chrono::high_resolution_clock::now();
+    setTime_ = true;
+  }
+
+  void finish(std::exception_ptr eptr = std::exception_ptr{}) {
```
fwyzard: If I understood the logic correctly, one thing that I find confusing (and thus do not like much) is that the `Sync` and `PseudoAsync` clients will automatically call `finish()`, while the `Async` client delegates the user to do so.

kpedro88: Yes, this is unfortunate but unavoidable. For a real-world example, see:
https://github.com/hls-fpga-machine-learning/SonicCMS/blob/v5.0.0/TensorRT/src/TRTClient.cc#L112

`finish()` has to be executed in the server's callback. There's no generic way to do it on the client side when using a truly asynchronous call.

fwyzard:
I think my suggestion was the opposite: `predict()` becomes `dispatch()` (the intermediate plugins implement this) and `predictImpl()` becomes `evaluate()` (the end users implement this)

HeterogeneousCore/SonicCore/interface/SonicClientTypes.h
kpedro88: Yes, that's correct. With the interface of `SonicEDProducer`, the user code just sets/gets these entries directly (passed by reference). In addition, in some cases these values might need to be accessed in other threads (PseudoAsync case).

kpedro88:
Ah, I misread. Will change it.

HeterogeneousCore/SonicCore/interface/SonicClientBase.h
fwyzard: Yes, I understand.

What about making the user call `finish()` also for the Sync and PseudoAsync cases ?

This would eliminate the confusion (and the specialisation in the dummy test).

kpedro88: That would be more consistent, but less user-friendly... to some extent I *want* to denote that true Async is a special case (some protocols don't support it, requires a callback, etc.). I'll think about it a bit more.

kpedro88:
please test

HeterogeneousCore/SonicCore/interface/SonicClientBase.h
kpedro88: I've implemented this change now, and added a note about it to the README.

kpedro88:
please test

fwyzard:
Looks reasonable to me.

@makortel , are there other improvements you would like to suggest ?

makortel:
I'm starting my review now. I think the package can be added to the `categories_map.py` now (under heterogeneous). @kpedro88 do you want to do that?

kpedro88:
see https://github.com/cms-sw/cms-bot/pull/1274

HeterogeneousCore/SonicCore/interface/SonicClientPseudoAsync.h
```diff
@@ -0,0 +1,76 @@
...
+#include <atomic>
+#include <exception>
+
+//pretend to be async + non-blocking by waiting for blocking calls to return in separate std::thread
+template <typename InputT, typename OutputT = InputT>
+class SonicClientPseudoAsync : public SonicClientBase, public SonicClientTypes<InputT, OutputT> {
+public:
+  //constructor
+  SonicClientPseudoAsync() : SonicClientBase(), SonicClientTypes<InputT, OutputT>(), hasCall_(false), stop_(false) {
+    thread_ = std::make_unique<std::thread>([this]() { waitForNext(); });
```
makortel: A thread per module x stream is fine for now, but I created an issue (#/29188) for a more general mechanism in case more use cases would surface.

makortel (commented):
Unless I missed something, the `SonicClientAsync` and `SonicClientSync` have identical implementations now. On the other hand, I do see value in keeping them as separate types so that the intended behavior is clear.

HeterogeneousCore/SonicCore/interface/SonicClientPseudoAsync.h
```diff
@@ -0,0 +1,76 @@
...
+  }
+  //destructor
+  virtual ~SonicClientPseudoAsync() {
```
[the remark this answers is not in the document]
fwyzard: `clang-tidy`should replace `virtual` with `override`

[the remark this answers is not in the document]
makortel: > `clang-tidy`should replace `virtual` with `override`

Right, but this PR is passing code checks so `clang-tidy` seems to be happy with the `virtual`?

HeterogeneousCore/SonicCore/interface/SonicClientPseudoAsync.h
```diff
@@ -0,0 +1,76 @@
...
+    {
+      std::lock_guard<std::mutex> guard(mutex_);
+      stop_ = true;
+    }
+    cond_.notify_one();
+    if (thread_) {
+      thread_->join();
```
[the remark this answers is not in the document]
fwyzard: Do you mean not calling `thread_join()`, or wrapping it in a `try/catch` block ?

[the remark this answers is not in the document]
makortel: I meant try-catch (sorry for being imprecise).

HeterogeneousCore/SonicCore/interface/SonicClientBase.h
```diff
@@ -0,0 +1,49 @@
...
+  virtual void dispatch(edm::WaitingTaskWithArenaHolder holder) = 0;
+
+protected:
+  virtual void evaluate() = 0;
+
+  void setStartTime() {
+    if (debugName_.empty())
+      return;
+    t0_ = std::chrono::high_resolution_clock::now();
+    setTime_ = true;
```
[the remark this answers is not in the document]
kpedro88: Nope, this can be simplified.

HeterogeneousCore/SonicCore/interface/SonicClientPseudoAsync.h
```diff
@@ -0,0 +1,76 @@
...
+public:
+  //constructor
+  SonicClientPseudoAsync() : SonicClientBase(), SonicClientTypes<InputT, OutputT>(), hasCall_(false), stop_(false) {
+    thread_ = std::make_unique<std::thread>([this]() { waitForNext(); });
+  }
+  //destructor
+  virtual ~SonicClientPseudoAsync() {
+    {
+      std::lock_guard<std::mutex> guard(mutex_);
+      stop_ = true;
```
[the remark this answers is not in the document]
kpedro88: At some point long ago, I was encountering a data race, so I moved every read and write inside a lock. This one shouldn't need it, though, so I'll remove it.

HeterogeneousCore/SonicCore/interface/SonicClientTypes.h
```diff
@@ -0,0 +1,25 @@
...
+  //destructor
+  virtual ~SonicClientTypes() = default;
+
+  //accessors
+  Input& input() { return input_; }
+  const Output& output() const { return output_; }
+
+protected:
+  Input input_;
+  Output output_;
```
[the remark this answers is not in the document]
kpedro88: In the future, we can consider if something like `swap()` or `shrink_to_fit()` would be worthwhile to free memory for certain cases.

HeterogeneousCore/SonicCore/test/DummyClient.h
```diff
@@ -0,0 +1,47 @@
...
+  }
+
+  //members
+  int factor_;
+  int wait_;
+};
+
+typedef DummyClient<SonicClientSync<int>> DummyClientSync;
+typedef DummyClient<SonicClientPseudoAsync<int>> DummyClientPseudoAsync;
+typedef DummyClient<SonicClientAsync<int>> DummyClientAsync;
```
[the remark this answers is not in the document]
kpedro88: Yes, related to your other observation (which I had overlooked after the last round of changes) that the base Async and Sync classes are now the same. Still, I prefer keeping the separate base classes, because derived versions won't be the same. In this case, the additional test is redundant, but maybe we'll come up with some clever generic async test in the future...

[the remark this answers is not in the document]
makortel: I would keep them separate as well (to exercise the compilation of all variations if nothing else).

kpedro88:
please test

makortel:
+heterogeneous

silviodonato:
please test for CMSSW_11_1_PY3_X

silviodonato:
please test for cc8_aarch64_gcc8

silviodonato:
+1

PR #29150, opened 2020-03-09
DQM: Sequence inspection tool and test

schneiml:
#### PR description:

This PR introduces a new tool that instruments `cmsDriver/cmsRun` to find out which modules are effectively loaded for each sequence. It is not strictly limited to DQM sequences, but focused on this usage for now. The method used is rather complicated and fragile, but it gives accurate results, unlike other methods I tried.

The information gathered about all sequences mentioned in `runTheMatrix` is stored into a SQLite database that can then be browsed interactively using a simple web-based frontend, or queried automatically.

This PR includes a unit test that builds the database and then queries for DQM legacy modules, and fails if any are found (it currently finds but ignores legacy `EDProducer`s and `EDFilter`s). The test runs multi-threaded and still takes about 10-20min to complete; I am not sure if this is ok for a unit test.

The tool can also be used to display the contents of a single sequence (see `--help`).

The tool works by
- Querying `runTheMatrix.py` for parameter combinations passed to `cmsDriver.py`.
- Executing `cmsDriver.py` once for each parameter combination, along with some fake options that should not affect which modules run but let `cmsDriver.py` succeed.
- Executing the resulting config files with `cmsRun` with the tracer service enabled. The jobs are configured for 0 events, so they finish rather quickly without actually reading any input, except for the HARVESTING jobs which tend to do things in `endJob` and eventually crash. This is fine though, the main purpose of actually running `cmsRun` is to get the list of loaded modules out of the tracer output.
- Dumping the configuration using `edmConfigDump` and parsing the output to resolve the module labels from the tracer into class names, and collect the precise config fragment for each instance.
- Running `edmPluginHelp` for each class to see what plugin type it is.

`cmsDriver` and `edmPluginHelp` are pretty slow, things are done multi-threaded to finish in reasonable time. If anybody has suggestions on how to avoid this huge detour and still produce reliable results, input is appreciated.

I may provide a an instance of the web interface hosted somewhere at some point, but for now, you need to run it yourself if you want to try it (Edit: test setup here: http://cmsswconfigexplore.cern.ch:8000/ (CERN network only), no sure for how long ti will keep running):

```
firefox http://localhost:8000 &
ssh -L 8000:localhost:8000 lxplus.cern.ch
mkdir -p /tmp/$USER
cd /tmp/$USER
cmsrel CMSSW_11_1_0_pre4
cd CMSSW_11_1_0_pre4/src
cmsenv
git cms-merge-topic schneiml:dqm-dqmdumpsequence
scram b
cmsswSequenceInfo.py --runTheMatrix
cmsswSequenceInfo.py --serve
```

#### PR validation:
This PR should not affect any production workflows. The tool seem to work.

The unit test might be very sensitive to all sorts of configuration problems.


schneiml:
please test

I assume this would run the new unit test? Let's see how it goes.

DQMOffline/Configuration/scripts/cmsswSequenceInfo.py
```diff
@@ -0,0 +1,532 @@
...
+import subprocess
+from collections import namedtuple
+from collections import defaultdict
+from multiprocessing.pool import ThreadPool
+
+Sequence = namedtuple("Sequence", ["seqname", "step", "era", "scenario", "mc", "data", "fast"])
+
+# We use two global thread pools, to avoid submitting from one Pool into itself.
+tp = ThreadPool()
+stp = ThreadPool()
```
schneiml: @smuzaffar @silviodonato We could make the number of threads configurable here quite easily for the unit tests.

DQMOffline/Configuration/test/BuildFile.xml
```diff
@@ -0,0 +1,5 @@
+<use   name="FWCore/Framework"/>
+<use   name="FWCore/Utilities"/>
+<bin file="TestRunner.cpp" name="TestDQMOfflineConfiguration">
+  <flags TEST_RUNNER_ARGS=" /bin/bash DQMOffline/Configuration/test runtests.sh"/>
```
smuzaffar: you can delete DQMOffline/Configuration/test/TestRunner.cpp and use the following to run the test. 
```
<test name="TestDQMOfflineConfiguration" command="runtests.sh --threads 1"/>
```
add `--threads` option to  cmsswSequenceInfo.py so that scram runtests always use 1 thread. Or if possible to break this tests in many then you can do something like

```
<test name="TestDQMOfflineConfiguration1" command="runtests.sh --threads 1 --workflows 1,2,3"/>
<test name="TestDQMOfflineConfiguration2" command="runtests.sh --threads 1 --workflows 4,5,6"/>
<test name="TestDQMOfflineConfiguration3" command="runtests.sh --threads 1 --workflows 7,8"/>
<test name="TestDQMOfflineConfiguration" command="runtests-final.sh">
  <flags PRE_TEST="TestDQMOfflineConfiguration1"/>
  <flags PRE_TEST="TestDQMOfflineConfiguration2"/>
  <flags PRE_TEST="TestDQMOfflineConfiguration3"/>
</test>
```
in this case scram can run TestDQMOfflineConfiguration1,2,3 in parallel.

schneiml: @smuzaffar thanks for the hint, I copied that pattern from other tests... (which probably also should not use it)

It should be pretty easy to break the test down into `DQM` and `VALIDATION`, more would get messy. I'll try that tomorrow.

schneiml:
please test

I split the test in two pieces and run each single threaded now. I have not run that to completion yet, let's see how it goes.

smuzaffar:
please test

schneiml:
It looks like the test timed out? Also it appears that the division into DQM and VALIDATION is not very even and DQM is a much bigger chunk. 

I guess then the next option is to actually divide the sequences onto multiple tests...

smuzaffar:
@schneiml , how about something like 

```
<test name="DumpDQMOfflineConfigurationWorkflows" command="runtests-dump.sh N" />
```
which can find all the workflows and divide them in N parts (write in to some files with N lines in it)

and
```
<test name="TestDQMOfflineConfiguration1" command="runtests.sh 1">
  <flags PRE_TEST="DumpDQMOfflineConfigurationWorkflows"/>
</test>
<test name="TestDQMOfflineConfiguration2" command="runtests.sh 2">
  <flags PRE_TEST="DumpDQMOfflineConfigurationWorkflows"/>
</test>
<test name="TestDQMOfflineConfiguration3" command="runtests.sh 3">
  <flags PRE_TEST="DumpDQMOfflineConfigurationWorkflows"/>
</test>
.....
<test name="TestDQMOfflineConfigurationN" command="runtests.sh N">
  <flags PRE_TEST="DumpDQMOfflineConfigurationWorkflows"/>
</test>
``` 
where runtests.sh  reads the Nth line to get the workflow number and process it. If runtests.sh finds an empty line then just exit 0

schneiml:
@smuzaffar is there any reason to use the `PRE_TEST`? I can just run all the batches completely independently (just test-running it). The only issue is that new batches need to be added if the number of sequences to be checked increases, but I added a test that will fail if there are more sequences than are covered by the batches, so that should be fine.

smuzaffar:
@schneiml , PRE_TEST is needed to make sure that DumpDQMOfflineConfigurationWorkflows runs before running any TestDQMOfflineConfigurationN

schneiml:
This version now does 5 chunks of up to 50 sequences each, and finished in 1h on `cmsdev32`. We could go even smaller, but there is increasing overhead since each job needs to run `edmPluginHelp` for ~all modules (inside the job it can be cached, but across jobs that would be complicated).

Now, the hard part is to have confidence that this will still actually trigger if there are legacy modules, but by construction it should be good.

schneiml:
please test

Now the new test should finish in time.

schneiml:
+1

Seems to work in the PR tests now, runs only a single thread per test job, and does actually fail when there are legacy modules (I tested that by adding "HARVESTING" to the list of steps to check, which contains legacy modules).

silviodonato:
+1

silviodonato:
@smuzaffar @mrodozov please note the next IB tests of CMSSW_11_1_X_2020-03-16-1100 will run the new unit test https://github.com/cms-sw/cmssw/blob/master/DQMOffline/Configuration/test/BuildFile.xml

PR #29151, opened 2020-03-09
New EDProducer emulating the Stub processing of the Phase 2 Outer Tracker DTC board

tschuh:
#### PR description:

New EDProducer emulating the Stub processing of the Phase 2 Outer Tracker DTC boards and it's Product. The Pruduct has been added to the EventContent and the Producer to the StandardSequence as other Track Trigger modules which are already present.

#### PR validation:

build of DataFormats/L1TrackTrigger and L1Trigger/L1TTrackerDTC only
run of L1Trigger/L1TTrackerDTC/test/test_cfg.py

#### Notes:

I did not template the DTCProducer, it can handle only
TTStub<edm::Ref<edm::DetSetVector<Phase2TrackerDigi>, Phase2TrackerDigi>>.
I don't understand why TTStub and TTCluster are templated in first place.


smuzaffar:
@tschuh , please update the title of this PR

smuzaffar:
please cleanup the history and remove the reference to these `nbproject/*` files

L1Trigger/L1TTrackerDTC/interface/DTC.h
```diff
@@ -0,0 +1,51 @@
...
+namespace L1TTrackerDTC {
+
+  class Settings;
+  class Module;
+  class Stub;
+
+  // representation of an outer tracker DTC board
+  class DTC {
+  private:
+    typedef std::deque<Stub*> Stubs;
```
kpedro88: the includes for these STL data formats should be present at the top of the file

L1Trigger/L1TTrackerDTC/interface/DTC.h
```diff
@@ -0,0 +1,51 @@
+#ifndef __L1TTrackerDTC_DTC_H__
+#define __L1TTrackerDTC_DTC_H__
+
+#include "L1Trigger/L1TTrackerDTC/plugins/ProducerFwd.h"
```
kpedro88: including from the plugins directory is not recommended

L1Trigger/L1TTrackerDTC/interface/Settings.h
```diff
@@ -0,0 +1,332 @@
...
+    const double outerRadius_;   // outer radius of outer tracker in cm
+    const double innerRadius_;   // inner radius of outer tracker in cm
+    const double maxPitch_;      // max strip/pixel pitch of outer tracker sensors in cm
+
+    // format specific parameter
+    const double maxEta_;        // cut on stub eta
+    const double minPt_;         // cut on stub pt, also defines region overlap shape in GeV
+    const double chosenRofPhi_;  // critical radius defining region overlap shape in cm
+    const int
+        numLayers_;  // max number of detector layer connected to one DTC (hybrid) / number of detector layers a reconstructbale particle may cross (tmtt)
```
kpedro88: better to put long comment on a separate to avoid awkward code format

L1Trigger/L1TTrackerDTC/interface/TTDTCConverter.h
```diff
@@ -0,0 +1,24 @@
...
+
+// converts TTDTC::Frame into GlobalPoint
+class TTDTCConverter {
+public:
+  TTDTCConverter(const edm::Run& iRun,
+                 const edm::EventSetup& iSetup,
+                 const std::string& processName = "",
+                 const std::string& productLabel = "L1TTrackerDTCProducer");
+
+  ~TTDTCConverter() { delete settings_; }
```
kpedro88: better to use a smart pointer

L1Trigger/L1TTrackerDTC/plugins/BuildFile.xml
```diff
@@ -0,0 +1,5 @@
+<use name="L1Trigger/L1TTrackerDTC"/>
+
+<flags CXXFLAGS="-g -Wno-unused-variable -Wno-misleading-indentation"/>
```
kpedro88: these flags aren't suitable for the official release

L1Trigger/L1TTrackerDTC/plugins/L1TTrackerDTCProducer.cc
```diff
@@ -0,0 +1,129 @@
...
+#include "L1Trigger/L1TTrackerDTC/interface/DTC.h"
+
+using namespace std;
+using namespace edm;
+
+namespace L1TTrackerDTC {
+
+  L1TTrackerDTCProducer::L1TTrackerDTCProducer(const ParameterSet& iConfig) {
+    // helper class to store configurations
+    settings_ = new Settings(iConfig);
```
kpedro88: better to use a smart pointer

L1Trigger/L1TTrackerDTC/plugins/L1TTrackerDTCProducer.h
```diff
@@ -0,0 +1,40 @@
+#ifndef __L1TTrackerDTC_L1TTRACKERDTCPRODUCER_H__
```
kpedro88: plugins are recommended not to have separate header files

L1Trigger/L1TTrackerDTC/plugins/ProducerFwd.h
```diff
@@ -0,0 +1,39 @@
+#ifndef __L1TTrackerDTC_RODUCERFWD_H__
```
kpedro88: it's better to make sure each class only includes the headers that it actually needs, rather than carrying around all possible headers used by the package
(and if you must do the latter, don't put the header file in the plugins directory)

L1Trigger/L1TTrackerDTC/python/Format_Hybrid_cfi.py
```diff
@@ -0,0 +1,46 @@
...
+
+    Label          = cms.string( "TTStubAlgorithm_official_Phase2TrackerDigi_" ), # 
+    Process        = cms.string( "HLT" ),                                         # empty string possible
+    BaseWindowSize = cms.double( .5 )                                             # precision of window sizes in pitch units
+
+  ),
+
+  #=== specific format parameter
+
+  ParamsFormat = cms.PSet (
```
kpedro88: how much do these values depend on the specific geometry version? Currently there are 7 versions of the phase 2 tracker (with another being proposed): https://github.com/cms-sw/cmssw/tree/master/Configuration/Geometry

L1Trigger/L1TTrackerDTC/python/Producer_Defaults_cfi.py
```diff
@@ -0,0 +1,62 @@
...
+    BaseCol               = cms.double(  1.     ), # precision of internal stub column in pitch units
+    BaseRow               = cms.double(   .5    ), # precision of internal stub row in pitch units
+    BendCut               = cms.double(  1.3125 ), # used stub bend uncertainty in pitch units
+    FreqLHC               = cms.double( 40.     )  # LHC bunch crossing rate in MHz
+  ),
+
+  #=== f/w constants
+
+  ParamsFW = cms.PSet (
+    SpeedOfLight = cms.double( 2.99792458       ), # in e8 m/s
```
kpedro88: this should be obtained as a constant in the code from CLHEP::c_light

L1Trigger/L1TTrackerDTC/python/Producer_Defaults_cfi.py
```diff
@@ -0,0 +1,62 @@
...
+    BField       = cms.double( 3.81120228767395 ), # in T
```
kpedro88: this should probably be obtained from the database (the magnetic field is specified as a database record)

L1Trigger/L1TTrackerDTC/python/Producer_Defaults_cfi.py
```diff
@@ -0,0 +1,62 @@
...
+    BFieldError  = cms.double(   1.e-6  ),         # accepted difference to EventSetup in T
+    OuterRadius  = cms.double( 112.7    ),         # outer radius of outer tracker in cm
```
kpedro88: dependent on geometry version?

L1Trigger/L1TTrackerDTC/python/Producer_cff.py
```diff
@@ -0,0 +1,19 @@
...
+# This describes the DTC Stub processing
+#---------------------------------------------------------------------------------------------------------
+
+#=== Import default values for all parameters & define EDProducer.
+
+from L1Trigger.L1TTrackerDTC.Producer_Defaults_cfi import L1TTrackerDTCProducer_params
+
+L1TTrackerDTCProducer = cms.EDProducer( 'L1TTrackerDTCProducer', L1TTrackerDTCProducer_params )
+
+if L1TTrackerDTCProducer_params.ParamsED.DataFormat == "Hybrid":
```
kpedro88: this is an unusual way to modify a producer. What are the use cases? (In this PR, it looks like DataFormat is only ever "Hybrid"...)

L1Trigger/L1TTrackerDTC/python/Producer_cff.py
```diff
@@ -0,0 +1,19 @@
...
+  from L1Trigger.L1TTrackerDTC.Format_Hybrid_cfi import L1TTrackerDTCFormat_params
+  L1TTrackerDTCProducer = cms.EDProducer('L1TTrackerDTCProducer', L1TTrackerDTCProducer_params, L1TTrackerDTCFormat_params )
+
+if L1TTrackerDTCProducer_params.ParamsED.DataFormat == "TMTT":
```
kpedro88: elif?

L1Trigger/L1TTrackerDTC/src/DTC.cc
```diff
@@ -0,0 +1,145 @@
...
+        region_(dtcId / settings->numDTCsPerRegion()),  // outer tracker detector region [0-8]
+        board_(dtcId % settings->numDTCsPerRegion()),   // outer tracker dtc id in region [0-23]
+        modules_(modules)                               // container of sensor modules connected to this DTC
+  {
+    stubs_.reserve(nStubs);  // container of dynamic allocated stubs on this DTC
+  }
+
+  DTC::~DTC() {
+    for (auto& stub : stubs_)
+      delete stub;
```
kpedro88: do all the stubs really need to be allocated dynamically? this implies some performance penalty

L1Trigger/L1TTrackerDTC/src/DTC.cc
```diff
@@ -0,0 +1,145 @@
...
+
+  // router step 1: merges stubs of all modules connected to one routing block into one stream
+  void DTC::merge(Stubss& inputs, Stubs& output) {
+    Stubss stacks(inputs.size());  // for each input one fifo
+
+    // clock accurate firmware emulation, each while trip describes one clock tick
+    while (!all_of(inputs.begin(), inputs.end(), [](const Stubs& channel) { return channel.empty(); }) or
+           !all_of(stacks.begin(), stacks.end(), [](const Stubs& channel) {
+             return channel.empty();
+           })) {  // work as long as needed to merge all stubs
```
kpedro88: again recommend putting comments on separate lines to avoid awkward breaks

L1Trigger/L1TTrackerDTC/src/Module.cc
```diff
@@ -0,0 +1,113 @@
...
+    if (!barrel_ && psModule)
+      type_ = SettingsHybrid::diskPS;
+    if (!barrel_ && !psModule)
+      type_ = SettingsHybrid::disk2S;
+
+    // getting windows size for bend encoding
+    const int type = trackerTopology->tobSide(detId);  // 1 for tilted-, 2 for tilted+, 3 for flat
+
+    int ladder = barrel_ ? trackerTopology->tobRod(detId) : trackerTopology->tidRing(detId);
+    if (barrel_ && type == 1)
```
kpedro88: magic number 1 should be named constant

L1Trigger/L1TTrackerDTC/src/Module.cc
```diff
@@ -0,0 +1,113 @@
...
+      ladder = 1 + numTiltedLayerRings.at(layerId_) -
+               ladder;  // Corrected ring number, bet 0 and barrelNTilt.at(layer), in ascending |z|
+
+    double windowSize =
+        barrel_ ? windowSizeBarrelLayers.at(layerId_) : windowSizeEndcapDisksRings.at(layerId_ - 10).at(ladder);
+    if (barrel_ && type != 3)
```
kpedro88: magic number 3 should be named constant

L1Trigger/L1TTrackerDTC/src/Module.cc
```diff
@@ -0,0 +1,113 @@
...
+    // getting windows size for bend encoding
+    const int type = trackerTopology->tobSide(detId);  // 1 for tilted-, 2 for tilted+, 3 for flat
+
+    int ladder = barrel_ ? trackerTopology->tobRod(detId) : trackerTopology->tidRing(detId);
+    if (barrel_ && type == 1)
+      ladder = 1 + numTiltedLayerRings.at(layerId_) -
+               ladder;  // Corrected ring number, bet 0 and barrelNTilt.at(layer), in ascending |z|
+
+    double windowSize =
+        barrel_ ? windowSizeBarrelLayers.at(layerId_) : windowSizeEndcapDisksRings.at(layerId_ - 10).at(ladder);
```
kpedro88: magic number 10 should be named constant

L1Trigger/L1TTrackerDTC/src/Module.cc
```diff
@@ -0,0 +1,113 @@
...
+    // getting bend encoding
+    bendEncoding_ =
+        psModule ? bendEncodingsPS.at(ws) : bendEncodings2S.at(ws);  // index = encoded bend, value = decoded bend
+
+    // encoding for 2S endcap radii
+    decodedR_ = type_ == SettingsHybrid::disk2S ? numColumns_ * (ladder - numRingsPS.at(layerId_ - 11)) : 0;
+
+    // r and z offsets
+    offsetR_ = barrel_ ? layerRs.at(layerId_ - 1) : 0.;
+    offsetZ_ = barrel_ ? 0. : diskZs.at(layerId_ - 11);
```
kpedro88: magic number 11 should be named constant

L1Trigger/L1TTrackerDTC/src/Settings.cc
```diff
@@ -0,0 +1,418 @@
...
+    maxCot_ = sinh(maxEta_);
+
+    rangeQoverPt_ = speedOfLight_ * bField_ / minPt_ / 1000.;
+
+    widthLayer_ = ceil(log2(numLayers_));
+
+    baseRegion_ = 2. * M_PI / numRegions_;
+
+    if (dataFormat_ == "TMTT")
+      tmtt_ = new SettingsTMTT(iConfig, this);
```
kpedro88: use smart pointer (or just allocate statically?)

L1Trigger/L1TTrackerDTC/src/Settings.cc
```diff
@@ -0,0 +1,418 @@
...
+
+    // read in cabling map
+    ESHandle<TrackerDetToDTCELinkCablingMap> cablingMapHandle;
+    iSetup.get<TrackerDetToDTCELinkCablingMapRcd>().get(cablingMapHandle);
+    const TrackerDetToDTCELinkCablingMap* cablingMap = cablingMapHandle.product();
+
+    // check magnetic field configuration
+    const double bField = magneticField->inTesla(GlobalPoint(0., 0., 0.)).z();
+    if (fabs(bField - bField_) > bFieldError_) {
+      cms::Exception exception("Configuration",
```
kpedro88: why does this need to be checked?

L1Trigger/L1TTrackerDTC/src/Settings.cc
```diff
@@ -0,0 +1,418 @@
...
+      exception.addContext("L1TrackerDTC::Settings::beginRun");
+      throw exception;
+    }
+
+    // convert TrackerDetToDTCELinkCablingMap
+    TTDTC ttDTC(numRegions_, numOverlappingRegions_, numDTCsPerRegion_);
+    vector<int> modIds(numDTCs_, 0);                         // module counter for each DTC
+    for (const DetId& detId : trackerGeometry_->detIds()) {  // all tracker modules
+
+      if (detId.subdetId() == 1 || detId.subdetId() == 2)  // skip pixel detector
```
kpedro88: use enums rather than magic numbers

L1Trigger/L1TTrackerDTC/src/Settings.cc
```diff
@@ -0,0 +1,418 @@
...
+
+    function<bool(double, double)> comp = [](const double& lhs, const double& rhs) {
+      return lhs > 0. ? (rhs > 0. ? lhs < rhs : true) : false;
+    };
+
+    const double rangeRT = 2. * max(fabs(settings->outerRadius_ - settings->chosenRofPhi_),
+                                    fabs(settings->innerRadius_ - settings->chosenRofPhi_));
+    const double rangePhi = settings->baseRegion_ + rangeRT * settings->rangeQoverPt_ / 2.;
+
+    basesZ_.reserve(4);
```
kpedro88: magic number 4 should be named constant

L1Trigger/L1TTrackerDTC/src/Settings.cc
```diff
@@ -0,0 +1,418 @@
...
+    layerIdEncodings_.reserve(settings->numDTCs_);
+    for (int dtcId = 0; dtcId < settings->numDTCs_; dtcId++) {
+      auto begin = next(settings->cablingMap_.begin(), dtcId * settings->numModulesPerDTC_);
+      auto end = next(begin, settings->numModulesPerDTC_);
+      auto last = find_if(begin, end, [](const DetId& detId) { return detId.null(); });
+      if (last < end)
+        end = last;
+
+      // assess layerIds connected to this DTC
+      set<int> layerIds;
```
kpedro88: `#include <set>` should be added at the top

L1Trigger/L1TTrackerDTC/src/Settings.cc
```diff
@@ -0,0 +1,418 @@
...
+      for (auto it = begin; it < end; it++)
+        layerIds.insert(it->subdetId() == StripSubdetector::TOB ? trackerTopology->layer(*it)
+                                                                : trackerTopology->tidWheel(*it) + 10);
```
kpedro88: magic number 10

L1Trigger/L1TTrackerDTC/src/Stub.cc
```diff
@@ -0,0 +1,255 @@
...
+  TTDTC::BV Stub::formatTMTT(const int& region) const {
+    SettingsTMTT* format = settings_->tmtt();
+    const int& layerM = module_->layerId_;
+
+    // convert unique layer id [1-6,11-15] into reduced layer id [0-6]
+    // a fiducial track may not cross more then 7 detector layers, for stubs from a given track the reduced layer id is actually unique
+    int layer(-1);
+    if (layerM == 1)
+      layer = 0;
+    if (layerM == 2)
```
kpedro88: else if

L1Trigger/L1TTrackerDTC/src/Stub.cc
```diff
@@ -0,0 +1,255 @@
...
+      if (phiT_.second < format->baseSector())
+        sectorsPhi.set(2);
+      else
+        sectorsPhi.set(3);
+      if (phiT_.first >= 0. && phiT_.first < format->baseSector())
+        sectorsPhi.set(2);
+    }
+
+    // assign stub to eta sectors within a processing region
+    pair<int, int> setcorEta({0, format->numSectorsEta() - 1});
```
kpedro88: `#include <utility>`

L1Trigger/L1TTrackerDTC/test/MCsamples/1110/RelVal/DisplacedMuPt10To30/noPU.txt
```diff
@@ -0,0 +1,5 @@
+/store/relval/CMSSW_11_1_0_pre1/RelValDisplacedMuPt10To30/GEN-SIM-DIGI-RAW/110X_mcRun4_realistic_v2_2026D49noPU-v1/20000/C23FF53E-4167-8746-B4AA-174B3AF0766D.root
```
kpedro88: please remove all of these test file lists from the history

kpedro88:
please test

DataFormats/L1TrackTrigger/BuildFile.xml
```diff
@@ -5,6 +5,7 @@
 <use   name="DataFormats/Phase2TrackerDigi"/>
 <use   name="DataFormats/SiStripDetId"/>
 <use   name="DataFormats/TrackerCommon"/>
+<use   name="SimDataFormats/TrackingAnalysis"/>
```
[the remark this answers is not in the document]
tschuh: I think it is needed since I added
`#include "SimTracker/TrackTriggerAssociation/interface/TTClusterAssociationMap.h"`
in TTTypes.h

DataFormats/L1TrackTrigger/interface/TTDTC.h
```diff
@@ -0,0 +1,69 @@
...
+
+private:
+  // converts DTC identifier (region[0-8], board[0-23], channel[0-1]) into allStreams_ index [0-431]
+  int index(const int& dtcRegion, const int& dtcBoard, const int& dtcChannel) const;
+
+  // converts TFP identifier (region[0-8], channel[0-47]) into allStreams_ index [0-431]
+  int index(const int& tfpRegion, const int& tfpChannel) const;
+
+private:
+  static const int numSlots_ = 12;  // number of ATCA slots per shelf
```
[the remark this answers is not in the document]
tschuh: done

DataFormats/L1TrackTrigger/interface/TTTypes.h
```diff
@@ -17,8 +17,23 @@
...
+
+typedef edm::Ref<TTStubDetSetVec, TTStub<Ref_Phase2TrackerDigi_> > TTStubRef;
+typedef edm::Ref<TTClusterDetSetVec, TTCluster<Ref_Phase2TrackerDigi_> > TTClusterRef;
+
+typedef edmNew::DetSet<TTStub<Ref_Phase2TrackerDigi_> > TTStubDetSet;
+
+typedef edm::Ptr<TrackingParticle> TTTPPtr;
+typedef TTClusterAssociationMap<Ref_Phase2TrackerDigi_> TTClusterAssMap;
+
+#include "DataFormats/L1TrackTrigger/interface/TTDTC.h"
```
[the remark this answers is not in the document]
tschuh: Here I have a Problem,  `TTDTC` uses some types defined in TTTypes and TTTypes includes `TTDTC`. Either I do not include `TTDTC` in TTTypes or I define the needed types inside `TTDTC` again? Advice is welcome.

L1Trigger/L1TTrackerDTC/README.md
```diff
@@ -0,0 +1,21 @@
+# Setup instructions to just run
```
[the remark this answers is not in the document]
tschuh: I removed it.

L1Trigger/L1TTrackerDTC/plugins/L1TTrackerDTCProducer.h
tschuh: I never saw an EDProducer without header file. Can you point me to an example?

L1Trigger/L1TTrackerDTC/plugins/ProducerFwd.h
tschuh: I removed the collection of headers

L1Trigger/L1TTrackerDTC/python/Format_Hybrid_cfi.py
tschuh: A lot of the configurations depending on geometry, topology, cabling map, magnetic field. This is an emulator delivering bit accurate results as the FPGA firmware. The firmware can't read in ESProducts, therefore I am using human readable configurations in both s/w and f/w.

L1Trigger/L1TTrackerDTC/python/Producer_cff.py
tschuh: There are different track finding algorithms. The default is "Hybrid", which should be used during RelVal sample production. Any private user code using "TMTT" has to rerun the DTC step.

DataFormats/L1TrackTrigger/BuildFile.xml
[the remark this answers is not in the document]
davidlange6: then the class needs to move out of DataFormats - as no SimTracker dependencies are allowed  in DataFormats.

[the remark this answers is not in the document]
tschuh: I see, I just found it handy to have all relevant Track Trigger types in one place. I will remove the SimTracker type from TTTypes.h

DataFormats/L1TrackTrigger/interface/TTBV.h
```diff
@@ -0,0 +1,255 @@
...
+  }
+
+  // conversion: to string
+  std::string str() const { return bs_.to_string().substr(S - size_, S); }
+
+  // conversion: range based to string
+  std::string str(const int& start, const int& end = 0) const { return this->str().substr(size_ - start, size_ - end); }
+
+  // conversion: to int
+  int val() const { return (signed_ && this->msb()) ? (int)bs_.to_ullong() - std::pow(2, size_) : bs_.to_ullong(); }
```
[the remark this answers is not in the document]
tschuh: I created a look up table.

DataFormats/L1TrackTrigger/interface/TTTypes.h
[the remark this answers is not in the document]
tschuh: I moved the TTDTC out of TTTypes.h

L1Trigger/L1TTrackerDTC/interface/DTC.h
tschuh: done

L1Trigger/L1TTrackerDTC/interface/Settings.h
tschuh: done

L1Trigger/L1TTrackerDTC/plugins/BuildFile.xml
tschuh: I removed them.

L1Trigger/L1TTrackerDTC/plugins/L1TTrackerDTCProducer.cc
tschuh: static allocated now

L1Trigger/L1TTrackerDTC/python/Producer_Defaults_cfi.py
tschuh: see comment on Format_Hybrid_cfi.py

L1Trigger/L1TTrackerDTC/python/Producer_Defaults_cfi.py
tschuh: see comment on Format_Hybrid_cfi.py

L1Trigger/L1TTrackerDTC/python/Producer_Defaults_cfi.py
tschuh: see comment on Format_Hybrid_cfi.py

L1Trigger/L1TTrackerDTC/python/Producer_cff.py
tschuh: done

L1Trigger/L1TTrackerDTC/src/DTC.cc
tschuh: static allocation now

L1Trigger/L1TTrackerDTC/src/DTC.cc
tschuh: done

L1Trigger/L1TTrackerDTC/src/Module.cc
tschuh: done

L1Trigger/L1TTrackerDTC/src/Module.cc
tschuh: done

L1Trigger/L1TTrackerDTC/src/Module.cc
tschuh: done

L1Trigger/L1TTrackerDTC/src/Settings.cc
tschuh: done

L1Trigger/L1TTrackerDTC/src/Settings.cc
tschuh: well it would be good to recognize when the firmware thinks the bfield would be significant smaller or bigger as that what cms thinks.

DataFormats/L1TrackTrigger/interface/TTDTC.h
```diff
@@ -0,0 +1,74 @@
...
+
+  // Access to product configurations
+  int numRegions() const;
+  int numDTCBoards() const;
+  int numDTCChannel() const;
+  int numTFPChannel() const;
+
+  // Access to one specific stream of TTStubRefs using DTC identifier (region[0-8], board[0-23], channel[0-1])
+  // dtcRegions aka detector regions are defined by tk layout
+  Stream& dtcStream(const int& dtcRegion, const int& dtcBoard, const int& dtcChannel);
```
makortel: This function should be `const` and return `const Stream&`.

DataFormats/L1TrackTrigger/interface/TTDTC.h
```diff
@@ -0,0 +1,74 @@
...
+
+  // all TFP identifier (region[0-8], channel[0-47])
+  std::vector<int> tfpRegions() const;
+  std::vector<int> tfpChannels() const;
+
+  // Access to one specific stream of TTStubRefs using TFP identifier (region[0-8], channel[0-47])
+  // tfpRegions aka processing regions are rotated by -0.5 region width w.r.t detector regions
+  Stream tfpStream(const int& tfpRegion, const int& tfpChannel) const;
```
makortel: Return a `const` reference? Also the parameters are more performant with just `int` instead of references (also below).

DataFormats/L1TrackTrigger/interface/TTDTC.h
tschuh: I was to lazy to write a set and get function, therefore the non const reference. Shall I write set and get separated instead?

makortel: Yes, separate functions are preferred (note 2.10 and 2.11 in http://cms-sw.github.io/cms_coding_rules.html). Note that currently this function can not be used to get (because it is not `const`).

L1Trigger/L1TTrackerDTC/plugins/L1TTrackerDTCProducer.h
kpedro88: https://github.com/cms-sw/cmssw/blob/master/RecoLocalCalo/HGCalRecProducers/plugins/HGCalRecHitProducer.cc

The class is declared and implemented in the cc file (since, as a plugin, it is not intended to be included in any other source files)

L1Trigger/L1TTrackerDTC/python/Producer_Defaults_cfi.py
kpedro88: c_light is used elsewere in the emulator, e.g. https://github.com/cms-sw/cmssw/blob/master/L1Trigger/TrackTrigger/src/StubPtConsistency.cc
this is a constant of nature and it should be obtained consistently

L1Trigger/L1TTrackerDTC/src/Settings.cc
kpedro88: I don't quite understand the reasoning. Is this something the firmware actually checks in real life? In the simulation, the magnetic field is exactly what is in the database record.

L1Trigger/TrackerDTC/src/Settings.cc
```diff
@@ -0,0 +1,431 @@
...
+    if (fabs(bField - bField_) > bFieldError_) {
+      cms::Exception exception("Configuration",
+                               "Magnetic Field from EventSetup (" + to_string(bField) + ") differs more then " +
+                                   to_string(bFieldError_) + " from configured value (" + to_string(bField_) + ").");
+      exception.addContext("L1TrackerDTC::Settings::beginRun");
+      throw exception;
+    }
+
+    // convert TrackerDetToDTCELinkCablingMap
+    enum SubDetId{A=1, B=2};
```
kpedro88: single letter names are discouraged, please use something more descriptive

L1Trigger/TrackerDTC/src/Settings.cc
```diff
@@ -0,0 +1,431 @@
...
+
+    function<bool(double, double)> comp = [](const double& lhs, const double& rhs) {
+      return lhs > 0. ? (rhs > 0. ? lhs < rhs : true) : false;
+    };
+
+    const double rangeRT = 2. * max(fabs(settings->outerRadius_ - settings->chosenRofPhi_),
+                                    fabs(settings->innerRadius_ - settings->chosenRofPhi_));
+    const double rangePhi = settings->baseRegion_ + rangeRT * settings->rangeQoverPt_ / 2.;
+
+    basesZ_.reserve(4);
```
kpedro88: still a magic number here

L1Trigger/TrackerDTC/src/Stub.cc
```diff
@@ -0,0 +1,256 @@
...
+      if (phiT_.second < format->baseSector())
+        sectorsPhi.set(2);
+      else
+        sectorsPhi.set(3);
+      if (phiT_.first >= 0. && phiT_.first < format->baseSector())
+        sectorsPhi.set(2);
+    }
+
+    // assign stub to eta sectors within a processing region
+    pair<int, int> setcorEta({0, format->numSectorsEta() - 1});
```
kpedro88: still missing `#include <utility>`

L1Trigger/L1TTrackerDTC/python/Producer_cff.py
kpedro88: It seems like this would be better accomplished with a customization function. Right now, the user would have to edit `Producer_Defaults_cfi.py`.

L1Trigger/L1TTrackerDTC/python/Format_Hybrid_cfi.py
kpedro88: Eventually the firmware will need to be consistent with the detector we build and install, right? While the geometry is constantly evolving, hardcoding a lot of parameters is not the best way to go.

L1Trigger/L1TTrackerDTC/python/Producer_Defaults_cfi.py
tschuh: The firmware, written in VHDL, has no access to any c/c++ libraries. The only consistent way I found is the way you see, gathering of hard coded constant in a configuration file in both worlds, f/w and s/w. I am open to better solutions but I can't see them in the moment.

L1Trigger/L1TTrackerDTC/python/Producer_cff.py
tschuh: I thought that is the purpose of configuration files. Could you please explain in more detail how you would restructure the code? I wanted to separate the two configurations as much as possible while sharing as much code as possible.

L1Trigger/L1TTrackerDTC/src/Settings.cc
tschuh: the firmware has no access to the database record, the bfield is hard coded there.
If the DB changes I would like recognize that. We could remove the check, but then we have to hope that we don't forget to update the f/w when DB changes.

L1Trigger/L1TTrackerDTC/python/Producer_Defaults_cfi.py
kpedro88: To me, the best approach is to use the value from the central source in the emulator, and copy that value into the firmware.

L1Trigger/L1TTrackerDTC/python/Producer_cff.py
kpedro88: * change this file so it only has what you want to be the default
* add a separate file called `customizeProducer_cff.py` (or whatever), with a function for each non-default case:
```python
def useHybrid(process):
    from L1Trigger.TrackerDTC.Producer_Defaults_cfi import TrackerDTCProducer_params
    from L1Trigger.TrackerDTC.Format_Hybrid_cfi import TrackerDTCFormat_params
    process.TrackerDTCProducer = cms.EDProducer('TrackerDTCProducer', TrackerDTCProducer_params, TrackerDTCFormat_params )
    return process
```

Then the user can just put `--customise L1Trigger.TrackerDTC.customizeProducer_cff.useHybrid` in their `cmsDriver.py` command, or import the function by hand if using an existing `cmsRun` config.

L1Trigger/TrackerDTC/src/TTDTCConverter.cc
```diff
@@ -0,0 +1,117 @@
...
+
+  if (!iConfig) {
+    cms::Exception exception("Configuration", "TrackerDTC config not found in process history.");
+    exception.addContext("L1TrackerDTC::Settings::beginRun");
+    exception.addAdditionalInfo("Used process name: " + processName + ", used product label: " + productLabel + ".");
+
+    throw exception;
+  }
+
+  settings_ = make_unique<Settings>(*iConfig);
```
makortel: Also here, getting parameters from other modules is not allowed.

makortel (changes requested):

L1Trigger/TrackerDTC/src/Settings.cc
```diff
@@ -0,0 +1,434 @@
...
+      if (processPset && processPset->exists(productLabel_))
+        ttStubAlgorithmPSet = &processPset->getParameterSet(productLabel_);
+    }
+
+    if (!ttStubAlgorithmPSet) {
+      cms::Exception exception("Configuration", "TTStub algo config not found in process history.");
+      exception.addContext("L1TrackerDTC::Settings::beginRun");
+
+      throw exception;
+    }
```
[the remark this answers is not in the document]
makortel: > A different solution needs to be found that propagates the parameters properly e.g. in the configuration.

Possible solutions include
* define parameters in one PSet from where they are propagated either directly or via `refToPSet_` (i.e. by label of a global PSet)
* encapsulate the parameters (or possibly even functionality) into an ESProduct

L1Trigger/L1TTrackerDTC/python/Producer_Defaults_cfi.py
tschuh: I found no way to copy the value to f/w in human readable form while ensuring equality. I would like to avoid specifying machine readable constants, since it makes it harder to compare.

L1Trigger/TrackerDTC/plugins/TrackerDTCProducer.cc
```diff
@@ -0,0 +1,176 @@
...
+    // empty DTC product
+    TTDTC product(settings_.numRegions(), settings_.numOverlappingRegions(), settings_.numDTCsPerRegion());
+
+    // read in stub collection
+    Handle<TTStubDetSetVec> handleTTStubDetSetVec;
+    iEvent.getByToken(getTokenTTStubDetSetVec_, handleTTStubDetSetVec);
+
+    // apply cabling map
+    auto acc = [](int& sum, const TTStubDetSet* module) { return sum += module ? module->size() : 0; };
+    static const vector<const TTStubDetSet*> nullDTC(settings_.numModulesPerDTC(), nullptr);
```
kpedro88: @makortel is this usage fully thread safe?

L1Trigger/TrackerDTC/src/SettingsHybrid.cc
```diff
@@ -0,0 +1,197 @@
...
+        layerRs_(paramsFormat_.getParameter<vector<double>>("LayerRs")),
+        diskZs_(paramsFormat_.getParameter<vector<double>>("DiskZs")),
+        disk2SRsSet_(paramsFormat_.getParameter<vector<ParameterSet>>("Disk2SRsSet")) {
+    disk2SRs_.reserve(disk2SRsSet_.size());
+    for (const auto& pSet : disk2SRsSet_)
+      disk2SRs_.emplace_back(pSet.getParameter<vector<double>>("Disk2SRs"));
+
+    auto comp = [](const double& lhs, const double& rhs) { return lhs > 0. ? (rhs > 0. ? lhs < rhs : true) : false; };
+
+    const double rangeRT = 2. * max(fabs(settings->outerRadius_ - settings->chosenRofPhi_),
```
kpedro88: use `std::abs` rather than `fabs`

kpedro88 (commented):
Follow-up review after various changes.

I am still concerned about hardcoding of geometry, field, and physical constants:
https://github.com/cms-sw/cmssw/pull/29151#discussion_r389753113
https://github.com/cms-sw/cmssw/pull/29151#discussion_r389753466
https://github.com/cms-sw/cmssw/pull/29151#discussion_r389758662

The issue of copying parameter sets also needs to be resolved:
https://github.com/cms-sw/cmssw/pull/29151/files#r392528634

L1Trigger/TrackerDTC/plugins/TrackerDTCProducer.cc
makortel: In principle it is if `settings_.numModulesPerDTC()` is a constant and does not change from one call to the next.

On the other hand, the `nullDTC` seems to be used only to initialized `ttDTCs` on the line below, and I can't really think of any practical benefit from `nullDTC` being `static const`. The line below could be just
```cpp
vector<vector<const TTStubDetSet*>> ttDTCs(settings_.numDTCs(), vector<const TTStubDetSet*>(settings_.numModulesPerDTC(), nullptr));
```

L1Trigger/TrackerDTC/python/Producer_Costomize_cff.py
```diff
@@ -0,0 +1,6 @@
+def useTMTT(process):
```
kpedro88: still a typo Costomize -> Customize

tschuh:
Regarding the ParameterSets from the ProcessHistory: I use them nowadays only to compare with the PSets from configured modules to detect misconfiguration. Is that a problem?

L1Trigger/TrackerDTC/python/Producer_Customize_cff.py
```diff
@@ -0,0 +1,6 @@
+def useTMTT(process):
+    from L1Trigger.TrackerDTC.Producer_Defaults_cfi import TrackerDTCProducer_params
+    from L1Trigger.TrackerDTC.Format_TMTT_cfi import TrackerDTCFormat_params
+    TrackerDTCProducer_params.ParamsED.DataFormat = "TMTT"
+    process.TrackerDTCProducer = cms.EDProducer('TrackerDTCProducer', TrackerDTCProducer_params, TrackerDTCFormat_params )
```
kpedro88: this needs to be changed to match `trackerDTC::Producer`

makortel:
> Regarding the ParameterSets from the ProcessHistory: I use them nowadays only to compare with the PSets from configured modules to detect misconfiguration. Is that a problem?

While we (=framework) don't like this usage of `pset::Registry()`, we don't have a better solution to suggest to cross check that a module is configured in the same way in two separate jobs. So the solution in this PR is tolerable. I created an issue https://github.com/cms-sw/cmssw/issues/29288 about it (mostly to not forget, don't expect a quick solution).

kpedro88:
please test

kpedro88:
please test

makortel:
> > Regarding the ParameterSets from the ProcessHistory: I use them nowadays only to compare with the PSets from configured modules to detect misconfiguration. Is that a problem?
> 
> While we (=framework) don't like this usage of `pset::Registry()`, we don't have a better solution to suggest to cross check that a module is configured in the same way in two separate jobs. So the solution in this PR is tolerable. I created an issue #29288 about it (mostly to not forget, don't expect a quick solution).

Let me add something I forgot yesterday. The current solution requires to pass the labels of the process and the module whose PSet hash is looked up for the check. Keeping these labels up to date has a maintenance cot (that may be acceptable, but I want to mention it explicitly).

kpedro88:
```
----- Begin Fatal Exception 25-Mar-2020 15:17:25 CET-----------------------
An exception of category 'TrackerDetToDTCELinkCablingMap has been asked to return a DTCELinkId associated to a DetId, but the latter is unknown to the map. ' occurred while
   [0] Processing  stream begin Run run: 1 stream: 0
   [1] Calling method for module trackerDTC::Producer/'TrackerDTCProducer'
Exception Message:
 DetId = 420234372
```

In addition to fixing whatever caused this error, please put the exception message in the message, not in the category.

mmusich:
> In addition to fixing whatever caused this error.

wf 20034.0 is for a T6 Tracker geometry which features a different OT than all the subsequent active phase-2 tracker geometries. 
Presumably (based on the name) the current DTC cabling map in Global Tag `TrackerDetToDTCELinkCablingMap__OT616_200_IT613__T14__OTOnly` is not OK for T6?

@luigicalligaris @emiglior 

>  please put the exception message in the message, not in the category.

The exception is called by the DTC cabling map itself: https://github.com/cms-sw/cmssw/blob/master/CondFormats/SiPhase2TrackerObjects/src/TrackerDetToDTCELinkCablingMap.cc#L16

tschuh:
@mmusich
> Presumably (based on the name) the current DTC cabling map in Global Tag `TrackerDetToDTCELinkCablingMap__OT616_200_IT613__T14__OTOnly` is not OK for T6?

the cabling map in the data base should be fine with T14-18 and not fine with T5-13. Can we exclude any tests with outdated tracker geometries? I have no idea if the DTC producer works with these anyway.

mmusich:
> Can we exclude any tests with outdated tracker geometries?

I would not recommend that. The most straightforward way to fix this, IMHO is to update the cabling map for the T6 case and adjust the DTC producer to run correctly in that case.

tschuh:
The DTC producer emulates my f/w. My f/w will not support multiple tracker geometries. From time to time (all two years roughly) we move from one to a new one. What I can do in the s/w is to detect if an unsupported geometry is being used and return without any product being produced. That would avoid any exceptions. Is there a way to throw a warning instead of an exception?

mmusich:
@tschuh 
I guess this calls for a customization of the wf which run outdated geometries. 
The DTC producer should simply not be run if it cannot give a meaningful answer. I think the exception in the DTC cabling map code is there for a reason....

tschuh:
> I guess this calls for a customization of the wf which run outdated geometries.
I guess that is what I meant with
> Can we exclude any tests with outdated tracker geometries?
Could one of you please make the needed customizations?

mmusich:
>  Can we exclude any tests with outdated tracker geometries?

No, that is not what I meant, unless there is a wide consensus from Tracker phase-2 simulation and Upgrade coordination to remove it from CMSSW. 
What I meant is that the wf for T6 should not run the DTC Producer (excluding it with a process modifier or by other means). @makortel @kpedro88 maybe can suggest a preferred solution?

kpedro88:
T6 is currently kept around in the MTD TDR baseline geometry. We don't have geometry-specific modifiers for the tracker at the moment, only for HGCal. If we were to add a phase2_tracker_T6 modifier, we would then need to add a Phase2C4T6 Era, combining all the necessary modifiers for the workflow. As you can imagine, the outer product of geometry-specific modifiers for all subdetectors becomes extremely large, so we prefer to avoid this practice unless absolutely necessary.

emiglior:
> > In addition to fixing whatever caused this error.
> 
> wf 20034.0 is for a T6 Tracker geometry which features a different OT than all the subsequent active phase-2 tracker geometries.
> Presumably (based on the name) the current DTC cabling map in Global Tag `TrackerDetToDTCELinkCablingMap__OT616_200_IT613__T14__OTOnly` is not OK for T6?
> 
> @luigicalligaris @emiglior
> 
adding @ghugo83 
number of OuterTracker Dets in T6 geometry is different from the number of Dets in T14 geometry

> > please put the exception message in the message, not in the category.
> 
> The exception is called by the DTC cabling map itself: https://github.com/cms-sw/cmssw/blob/master/CondFormats/SiPhase2TrackerObjects/src/TrackerDetToDTCELinkCablingMap.cc#L32

luigicalligaris:
Hello, I just saw this tread and I am trying to align with it. I confirm that the exception is there on purpose: if it is being triggered it's because the data being analyzed does not correspond to the loaded cabling map, and therefore the services the latter provides may return an incorrect answer.

If you are analyzing data which was generated with different geometry, you should specify a different DTC cabling map payload tag, corresponding to the geometry used for that data.

If I recall correctly, when we were disucssing about the calbing map it was decided to leave behing deprecated geometries, because the first release in which the DTC cabling map was included was a release that was not using those geometries anymore.

mmusich:
@luigicalligaris 

> If you are analyzing data which was generated with different geometry, you should specify a different DTC cabling map payload tag, corresponding to the geometry used for that data.

would you be able to produce a payload for T6 that we could test here? Crossing fingers that would solve the problem in the most painless way.

kpedro88:
Yes, adding an appropriate record in https://github.com/cms-sw/cmssw/blob/master/Configuration/AlCa/python/autoCondPhase2.py would be far preferable

luigicalligaris:
Ok, I checked out, compiled and ran the basic tests for the DTCCablingMapProducer on the latest snapshot CMSSW_11_1_X_2020-03-25-2300: it seems to be working properly.

Now, who can suggest which geometry should I produce a payoad for, specifically?
http://cms-tklayout.web.cern.ch/cms-tklayout/layouts-work/recent-layouts/

@tschuh , do you know which geometry you need? @ghugo83 has a suggestion?

tschuh:
The one which is there (OT616) is what I need. The wf of the PR test procedure needs additionally OT614 I guess.
As far as I can see there are only 3 different cabling maps for the OT: OT613 (for T5,T9,T10), OT614 (for T6,T7,T8,T11,T12,T13) and  OT616 (for T14,T15,T16,T17,T18).

emiglior:
OT614_200_IT404 is the tkLayout geometry corresponding to T6

http://cms-tklayout.web.cern.ch/cms-tklayout/layouts-work/cmssw-models/J_OT614_200_IT404/index.html

but I am not sure a cabling map was made available for this (old) geometry through the tkLayout  web site

ghugo83:
>adding @ghugo83
>number of OuterTracker Dets in T6 geometry is different from the number of Dets in T14 geometry

Yes absolutely, the T14 cabling map is designed for versions >=T14, and will not work on obsolete geometries.
tkLayout can only produce the latest cabling map on the latest OT: OT >=T14.

I could try to find back an old tkLayout mapping, generated at T6 time. 
An issue is that not only the geometry, but also the **mapping** itself are obsolete (we did reshuffle the mapping with feedbacks from bandwidth studies).

The OT cabling map should be stable now.
We are now doing minor tuning of the services channels arrangements, which will not affect the cabling map itself, in the sense of which module end up being connected to which DTC.

Are you sure you want to introduce to the framework an old mapping, which does not make sense from a design point of view? Will this not add complications and heaviness eventually?

Only the latest cabling map makes sense being studied.

luigicalligaris:
I took the csv for OT614_200_IT404_layer2_10G from @ghugo83 tklayout site intance, produced the payload and tested it to check whether it is stored with the right tag and can be read out. It seems to be ok.

@tschuh , you should now test it, specifying this as the tag for the payload you are retrieving:

`TrackerDetToDTCELinkCablingMap__OT614_200_IT404_layer2_10G__T6__OTOnly`

I think the following line in the PSet file should be enough to tel cmsRun to use the .db file:

```
process.load("CondCore.CondDB.CondDB_cfi")
process.CondDB.connect = 'sqlite_file:TrackerDetToDTCELinkCablingMap__OT614_200_IT404_layer2_10G__T6__OTOnly.db'
```

Let us know

luigicalligaris:
Link to the SQLite file here:

[https://cernbox.cern.ch/index.php/s/wPT4NRliMLpJczn](https://cernbox.cern.ch/index.php/s/wPT4NRliMLpJczn)

ghugo83:
Ha yes good news, I found back I had done a backport of the new **mapping** (ie, the mapping which integrates the latest DTC bandwidth studies) to the T6 geometry:
http://ghugo.web.cern.ch/ghugo/layouts/cabling/OT614_200_IT404_layer2_10G/cablingOuter.html 
At the bottom of the page, is the csv sent to @luigicalligaris for producing the .db .

And the new mapping on the latest geometry is at:
http://cms-tklayout.web.cern.ch/cms-tklayout/layouts-work/repository-git-dev/OT616_IT615/cablingOuter.html

tschuh:
I am not describing the run configuration.
@kpedro88 or @makortel or @mmusich
Could one of you please incorporate Luigis solution into the test wf?

mmusich:
@tschuh I'll take a look and let you know how to incorporate the changes (if the payload solves the issue).

mmusich:
so, I tested the payload provided by @luigicalligaris with these changes on top of this branch:
```diff
diff --git a/Configuration/AlCa/python/autoCondPhase2.py b/Configuration/AlCa/python/autoCondPhase2.py
index 877912dac93..797ad53c7d7 100644
--- a/Configuration/AlCa/python/autoCondPhase2.py
+++ b/Configuration/AlCa/python/autoCondPhase2.py
@@ -7,10 +7,15 @@ SiPixelSimLARecord        =   "SiPixelLorentzAngleSimRcd"
 SiPixelGenErrorRecord     =   "SiPixelGenErrorDBObjectRcd"       
 SiPixelTemplatesRecord    =   "SiPixelTemplateDBObjectRcd"       
 SiPixel2DTemplatesRecord  =   "SiPixel2DTemplateDBObjectRcd"     
+TrackerDTCCablingRecord   =   "TrackerDetToDTCELinkCablingMapRcd"
 
 #combines in a single dict of dict the tags defined below
 allTags={}
 
+allTags["DTCCabling"] = {
+    "T6" :  ( ','.join( [ 'TrackerDetToDTCELinkCablingMap__OT614_200_IT404_layer2_10G__T6__OTOnly',TrackerDTCCablingRecord,"sqlite_file:/tmp/musich/testDTC/CMSSW_11_1_X_2020-03-26-1100/src/TrackerDetToDTCELinkCablingMap__OT614_200_IT404_layer2_10G__T6__OTOnly.db", "", "2020-03-26 17:00:00.000"] ), ),  # cabling map provided for T6 geometry
+}
+
 allTags["LA"] = {
     'T6'  : ( ','.join( [ 'SiPixelLorentzAngle_phase2_T6_v2_mc'  ,SiPixelLARecord,connectionString, "", "2019-11-05 20:00:00.000"] ), ),  #uH = 0.106/T (TBPX), uH=0.0/T (TEPX+TFPX)
     'T14' : ( ','.join( [ 'SiPixelLorentzAngle_phase2_T14_v2_mc' ,SiPixelLARecord,connectionString, "", "2019-11-05 20:00:00.000"] ), ),  #uH = 0.106/T (TBPX), uH=0.0/T (TEPX+TFPX)
@@ -68,7 +73,7 @@ allTags["Template2Dden"] = {
 }
 
 # list of active tags to be replaced
-activeKeys = ["LA","LAWidth","SimLA","LAfromAlignment","GenError","Template"]
+activeKeys = ["DTCCabling","LA","LAWidth","SimLA","LAfromAlignment","GenError","Template"]
 
 # list of geometries supported
 activeDets = ["T6","T14","T15","T19"]
```

and step2 of wf. 20034.0 and indeed it runs to completion, so good news.
The less good news is that for some obscure reason at the moment the Database doesn't allow me to upload the payload (see: [log](https://cms-conddb-dev.cern.ch/cmsDbBrowser/logs/condition_uploader_logs/Prep/2020-03-19/2020-03-26/None/None/None/None/None/None/failure)). I'll investigate with AlCa/DB about that and come back to you...

mmusich:
The DB problem has been solved in the meanwhile.
@tschuh you can cherry-pick this commit https://github.com/cms-sw/cmssw/commit/f5f719efc4619a334878d87b0513c05987eafd8c here, or I can open a second ancillary PR which needs to be tested together with this one.

tschuh:
Many thanks @mmusich, I picked your commit. Additionally does the producer now cross check if certain configurations are supported and throws a warning if not.

kpedro88:
please test

kpedro88:
+upgrade

silviodonato:
+operations

silviodonato:
@rekovic @benkrikler @pohsun @tlampen @tocheng @christopheralanwest
do you have further comments?

boudoul:
Hello I do not think that the title is explicit for any of us, maybe at least changing it with 'New EDProducer emulating the Stub processing of the Phase 2 Outer Tracker DTC board' would be already much more explicit

rekovic:
+1

silviodonato:
alca:  @pohsun @tlampen @tocheng @christopheralanwest ?

tocheng:
+1

silviodonato:
code-checks

silviodonato:
+1

silviodonato:
A unit test is failing in the last IB and likely the reason is related to this PR.
Could you please have a look at https://github.com/cms-sw/cmssw/issues/29474?

PR #29153, opened 2020-03-09
Changing handling of color indexes and anti particle creation for pythia8 EGun (9_3_X backport)

pasmuss:
#### if this PR is a backport please specify the original PR and why you need to backport that PR
Backport for PR #29000 to CMSSW_9_3_17 for MC production (Fall17GS campaign). #29000 already approved and merged.

#### PR description (from original request):

Changes for proper treatment of color (conservation and codes):

- Increasing color index after generating one particle (anti particle obtains the same color, just on the anticolor column)
- Adding a check that if only colored particles are generated, the AddAntiParticle option is used to make sure that all colors are balanced
- Only affecting pythia8 EGun

#### PR validation (from original request):

Ran some tests locally with a core CMSSW script:

- not set AddAntiParticle at all (does not start, option must be set)
- set option to False (runs but does not generate anything and throws errors)
- set option to True (as expected: anti particle gets same anti color as particle gets color, index is not the same for two different particles)

### DIFF
--- a/GeneratorInterface/Pythia8Interface/plugins/Py8EGun.cc
+++ b/GeneratorInterface/Pythia8Interface/plugins/Py8EGun.cc
@@ -6,115 +6,120 @@
 
 namespace gen {
 
-class Py8EGun : public Py8GunBase {
-   
-   public:
-      
-      Py8EGun( edm::ParameterSet const& );
-      ~Py8EGun() {}
-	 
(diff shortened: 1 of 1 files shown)


qliphy:
please test

qliphy:
+1

silviodonato:
+1

PR #29171, opened 2020-03-10
removing HLTSeedL1LogicScalers

Sam-Harper:
#### PR description:
The DQM group alerted us to the fact this module no longer works in CMSSW_11_1_0_pre3. On closer inspection, it was found to be hardcoded to the legacy L1 not run since 2016 and thus did nothing.  Further inspection also implies that it hasnt been used since sometime around Run1 due to the paths it was monitoring. Given we've survived without it so far, I propose rather than fix it, we delete it.

Using the lxr, I deleted all references to it, even those that were commented out. 

#### PR validation:

It currently doesnt run so is unlikely to be in any validation workflow.


jfernan2:
please test

jfernan2:
+1

silviodonato:
+1

PR #29177, opened 2020-03-11
DQM: Blacklist some ProvInfo plots

schneiml:
#### PR description:

These cause false positives in the bin-to-bin comparison since `DQMDCSInfo` was replaced with `DQMProvInfo`. (#28829)

#### PR validation:
None. Should not affect any workflows. Effect will not be visible in the PR tests, since the comparison is run using the baseline IB, not the PR'ed code.


schneiml:
+1

silviodonato:
+1

PR #29184, opened 2020-03-11
DQM: Add README for DQMServices/Core.

schneiml:
#### PR description:

This PR adds a long README file to `DQMServices/Core` (and a smaller one to `DQMServices/Components`), based on an older README that I wrote for the "new DQMStore" project in 2019, the very extensive PR description from #28622 and some slides that will be presented at the upcoming PPD workshop. Due to the different sources, the structure is not perfect, but I think it covers all important aspects and, most importantly, matches the current implementation.

#### PR validation:
No code is affected. I did proof-read it once.


schneiml:
+1

silviodonato:
+1 
Thanks!!

PR #29185, opened 2020-03-11
Run3-sim58 Adding eval=true for a number of xml files

bsunanda:
#### PR description:

Adding the required eval="true" in some of the ConstantsSection to allow dd4hep to proceed

#### PR validation:

Use runTheMatrix.py for validation
 
Nothing special


bsunanda:
@cmsbuild Please test

cvuosalo:
@bsunanda In this PR, you are changing old XML files, like from 2015. We are not likely to run those scenarios with DD4hep. Also, how does adding `eval="true"` affect the comparison of geometry XML files and XML in the Conditions DB?

bsunanda:
eval="true" does not change any evaluation using DDD. The default is eval="true". But when one tries to use dd4hep, they create problem. It is absolutely safe to have this change.

cvuosalo:
+1

silviodonato:
+1

PR #29189, opened 2020-03-11
Enhance formatting of special geometry test

cvuosalo:
A special geometry test is used by experts to compare geometry descriptions in different formats, e.g. in a local database and in the Conditions database. This PR enhances the output of this test. First, floating-point values close to zero are rounded to zero to prevent minor fluctuations of very tiny values from showing up as differences. Second, each line is labeled with a sub-detector name so it is easier to identify differences between two geometry descriptions.

Further improvements in this test might be desirable. It is still vulnerable to re-arrangements of identical descriptions.

This test is not used in production or in unit tests.

#### PR validation:

Test output is now somewhat easier to read.


cvuosalo:
@cmsbuild please test

silviodonato:
Kind reminder to @ggovi

ggovi:
+1

silviodonato:
+1

PR #29193, opened 2020-03-11
Phase2-hgx233 Enable processing a full scenario

bsunanda:
#### PR description:

In dd4hep, the xml files are processed in order. So the scenario has to be selected accordingly. Also namespace is handled specially - change is needed to one of the plugins

#### PR validation:

Tested with some of the cfg's in test area. But it also needs some of the PR's which are waiting to be integrated to the IB

Nothing special


bsunanda:
@cmsbuild Please test

cvuosalo:
+1

bsunanda:
@cvuosalo @kpedro88 With this, #29185 and Ianna's PR which enables vectors to be fully evaluated, we can go through some phase2 scenario. This will enable debugging implementation for HGCal

kpedro88:
@bsunanda we should remember to propagate this to https://github.com/cms-sw/cmssw/blob/master/Configuration/Geometry/python/dict2026Geometry.py when the migration is complete

kpedro88:
+upgrade

silviodonato:
+1

PR #29213, opened 2020-03-16
MTD Sensitive Detector: make it independent on DDD

fabiocos:
#### PR description:

This PR removes the residual dependency of the ```MtdSD``` class on DDD, which was used just to print in a LogInfo message the numerical code assigned to Barrel or Endcap as SpecPar. At the same time the old ```FastTimerSD``` class, used in the very first simulation of MTD (no more available) is removed, and the corresponding production cuts and sensitive volume definitions in ```Geometry/HGCalSimData``` are deleted.

The residual dependency of ```SimG4CMS/Forward``` on DDD is concentrated in Zdc. But as far as I can see the ```ZdcShowerLibrary``` doe not really use the ```DDCompactView```  passed through the constructor, so this dependency does not look real and could be removed: @bsunanda @civanch do you agree?

#### PR validation:

Test wf 23234.0 (scenario D49) step 1 runs smoothly providing the identical same set of Sim and Calo hits, as shown by https://github.com/cms-sw/cmssw/blob/master/SimG4Core/Application/test/runSimHitCaloHitDumper_cfg.py


fabiocos:
please test

cvuosalo:
+1

fabiocos:
@civanch @cvuosalo as ```FastTimerSD``` is no more used nor available, I do not see the point to maintain it in the ```g4SimHits``` configuration. In my understanding this should not affect any backward compatibility.

fabiocos:
please test

in my private test nothing is changing as expected

kpedro88:
+upgrade

civanch:
I would agree to drop unused SD code.

civanch:
+1

silviodonato:
+1

PR #29221, opened 2020-03-17
use edm::createGlobalIdentifier instead of boost::uuids::random_generator

slava77:
This is a follow up to https://github.com/cms-sw/cmsdist/issues/5657

All 3 cases were using ```boost::uuids::random_generator``` to generate a random string.
Somewhat equivalent functionality is available as ```edm::createGlobalIdentifier```.

I only tested that the problematic case described in https://github.com/cms-sw/cmsdist/issues/5657 runs with this update


slava77:
@cmsbuild please test

pohsun:
+1

ggovi:
+1

silviodonato:
please test

silviodonato:
+1

PR #29228, opened 2020-03-18
Ecal laser cond tools

grasph:
#### PR description:

This PR insert a module that is used by ECAL to fill the condition database. The module is not used in this simulation or reconstruction workflow.

#### PR validation:

The code is used on regular basis. The compilation was tested specifically with CMSSW_10_6_9 and CMSS_11_0_1 on SLC7 before making this request.


grasph:
Why are they conflict on packages I've not touched ?

PR #29230, opened 2020-03-18
added module used to insert ECAL laser correction sets into the condi…

grasph:
…tion database

#### PR description:

This PR insert a module that is used by ECAL to fill the condition database. The module is not used in this simulation or reconstruction workflow.

#### PR validation:

The code is used on regular basis. The compilation was tested specifically with CMSSW_10_6_9 and CMSS_11_0_1 on SLC7 before making this request.

### DIFF
--- a/CalibCalorimetry/EcalTrivialCondModules/interface/EcalLaserCondTools.h
+++ b/CalibCalorimetry/EcalTrivialCondModules/interface/EcalLaserCondTools.h
@@ -1,0 +1,99 @@
+#ifndef EcaLaserCondTools_h
+#define EcaLaserCondTools_h
+
+/*
+ * $Id: EcalLaserCondTools.h,v 1.2 2010/06/14 10:45:16 pgras Exp $
+ */
+
+#include "FWCore/Framework/interface/Frameworkfwd.h"
+#include "FWCore/Framework/interface/EDAnalyzer.h"
+#include "FWCore/ParameterSet/interface/ParameterSet.h"
+#include "FWCore/ServiceRegistry/interface/Service.h"
+#include "CondCore/DBOutputService/interface/PoolDBOutputService.h"
+#include "CondFormats/EcalObjects/interface/EcalLaserAPDPNRatios.h"
+
+#include "DataFormats/DetId/interface/DetId.h"
+
+#include <stdio.h>
+#include <vector>
+#include <string>
+#include <time.h>
+/**
+ */
+class EcalLaserCondTools : public edm::EDAnalyzer {
+  //static fields
+
+  /** Number of extended laser monitoring regions
+   */
+  static const int nLmes = 92;
+
+  //methods
+public:
+  /** Constructor
+   * @param ps analyser configuration
+   */
+  EcalLaserCondTools(const edm::ParameterSet&);
+
+  /** Destructor
+   */
+  ~EcalLaserCondTools();
+
+  /** Called by CMSSW event loop
+   * @param evt the event
+   * @param es events setup
+   */
+  virtual void analyze(const edm::Event& evt, const edm::EventSetup& es);
+
+private:
+  static std::string toNth(int n);
+  static std::string timeToString(time_t t);
+  class CorrReader {
+  public:
+    CorrReader() : verb_(0) {}
+    virtual bool readTime(int& t1, int t2[nLmes], int& t3) { return false; }
+    virtual bool readPs(DetId& rawdetid, EcalLaserAPDPNRatios::EcalLaserAPDPNpair& corr) { return false; }
+    virtual ~CorrReader() {}
+    void setVerbosity(int verb) { verb_ = verb; }
+
+  protected:
+    int verb_;
+  };
+
+  class FileReader : public EcalLaserCondTools::CorrReader {
+  public:
+    FileReader(const std::vector<std::string>& fnames) : f_(0), fnames_(fnames), ifile_(-1), iline_(0) {}
+    virtual bool readTime(int& t1, int t2[nLmes], int& t3);
+    virtual bool readPs(DetId& rawdetid, EcalLaserAPDPNRatios::EcalLaserAPDPNpair& corr);
+    virtual ~FileReader() {}
+
+  private:
+    bool nextFile();
+    void trim();
+    FILE* f_;
+    std::vector<std::string> fnames_;
+    unsigned ifile_;
+    int iline_;
+  };
+
+private:
+  void fillDb(CorrReader& r);
+  void dbToAscii(const edm::EventSetup& es);
+  void processIov(CorrReader& r, int t1, int t2[nLmes], int t3);
+
+  //fields
+private:
+  FILE* fout_;
+  FILE* eventList_;
+  std::string eventListFileName_;
+  int verb_;
+  std::string mode_;
+  std::vector<std::string> fnames_;
+  edm::Service<cond::service::PoolDBOutputService> db_;
+  int skipIov_;
+  int nIovs_;
+  int fromTime_;
+  int toTime_;
+  FILE* ferr_;
+};
+
+#endif  //EcaLaserCondTools_h not defined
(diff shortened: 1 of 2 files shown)


grasph:
patch applied.

pohsun:
@grasph As the compilation is done on CMSSW_10_6_9, do you manage to backport?

grasph:
@pohsun the backport is not needed anymore.

grasph:
Sorry we need it for CMSSW_11_0_X and I didn't realize that master will go in CMSSW_11_1_X. Should I do a PR for CMSSW_11_0_X now ? It was not clear to me when I should to the PR for backport and if it was to wait for the merge in master.

pohsun:
@grasph You could make the backport PR now, but it will be reviewed and merged only after this PR is accepted. Please also mention backport is needed in this PR description.

pohsun:
please test

grasph:
Hello,

I don't understand the clang warnings about [line 37](https://github.com/grasph/cmssw/blob/448f4082912194cbdc2cd0c5d2ca5cf0053b7d7c/CalibCalorimetry/EcalTrivialCondModules/src/EcalLaserCondTools.cc#L373) and [375](https://github.com/grasph/cmssw/blob/448f4082912194cbdc2cd0c5d2ca5cf0053b7d7c/CalibCalorimetry/EcalTrivialCondModules/src/EcalLaserCondTools.cc#L375) of  EcalLaserConTools.cc.

Can some expert help with demystifying them ? 

I'm also not sure it is guaranteed that the compiler won't drop the intended calls to the constructors in these two lines.

Philippe.

CalibCalorimetry/EcalTrivialCondModules/src/EcalLaserCondTools.cc
```diff
@@ -0,0 +1,489 @@
...
+                                << " values,"
+                                << " raw id" << rawdetid << ": " << corr.p1 << ", " << corr.p2;
+  }
+  detid = rawdetid;
+  const int ECALID = 3;
+  if (detid.det() != ECALID)
+    throw cms::Exception("InvalidValue") << "Line " << iline_ << "of file " << fnames_[ifile_]
+                                         << " contains an invalid det ID (detector code is not ECAL!)\n";
+  if (detid.subdetId() == EcalBarrel)
+    EBDetId(detid);  //in order to throw an exception if detid is not valid
```
davidlange6: @grasph - your clang warnings are from this line, which doesn't seem like valid code.. What do you actually want to do here?

grasph: Thanks @davidlange6 to have look at this.

The EBDetId(int) ctor used to do some validity check on the detid and throw an exception in case it is invalid. The purpose of this line was to check we can build an EBDetId from detid. 

Anyway, looking at the code of EBDetId, the check was removed, I guess for performance reason, and therefore the two lines have no more effect. I will remove them.

davidlange6:
So calling a constructor means you should construct something..

eg

EcalDetId dummy(detid);

(Which then should be used) - perhaps a method in EcalDetId should be added to allow this check

The EBDetId(int) ctor used to do some validity check on the detid and throw an exception in case it is invalid. The purpose of this line was to check we can build an EBDetId from detid.

Anyway, looking at the code of EBDetId, the check was removed, I guess for performance reason, and therefore the two lines have no more effect. I will remove them.

grasph:
> So calling a constructor means you should construct something.. eg EcalDetId dummy(detid); (Which then should be used) - perhaps a method in EcalDetId should be added to allow this check 

I agree that having this method would have been the best , but it was simpler to handle it like this than asking for a modification in another package. Note, although not elegant, the code was valid, at least for c++-98 standard that states a constructor function can be explicitly called. I'm less knowledgeable of the newer standards.

I've modified the code and included the validity check. @fcouderc can you test the code on a file to check that the change I made didn't break anything ?

christopheralanwest:
please test

CalibCalorimetry/EcalTrivialCondModules/interface/EcalLaserCondTools.h
```diff
@@ -0,0 +1,93 @@
+#ifndef SRCONDACCESS_H
+#define SRCONDACCESS_H
```
pohsun:
```suggestion
#define EcalLaserCondTools_h
```

This name conflicts with [`SimCalorimetry/EcalSelectiveReadoutProducers/interface/EcalSRCondTools.h`](https://github.com/cms-sw/cmssw/blob/master/SimCalorimetry/EcalSelectiveReadoutProducers/interface/EcalSRCondTools.h#L1).

CalibCalorimetry/EcalTrivialCondModules/src/EcalLaserCondTools.cc
```diff
@@ -0,0 +1,497 @@
...
+  EcalLaserAPDPNRatios::EcalLaserTimeStamp t;
+  iovStart = uint64_t(t1) << 32;
+  for (size_t i = 0; i < nLmes; ++i) {
+    t.t1 = edm::Timestamp(uint64_t(t1) << 32);
+    t.t2 = edm::Timestamp(uint64_t(t2[i]) << 32);
+    t.t3 = edm::Timestamp(uint64_t(t3) << 32);
+    corrSet->setTime(i, t);
+  }
+
+  const int ncrystals = 75848;
```
pohsun: Constants which could be estimated at compile period should be defined with `const expr`.
Please apply the same fix to other constants.

CalibCalorimetry/EcalTrivialCondModules/src/EcalLaserCondTools.cc
```diff
@@ -0,0 +1,497 @@
...
+  size_t s = 0;
+  while ((i = fgetc(f_)) != 'T' && i != 'L' && i >= 0)
+    getline(&buf, &s, f_);
+  if (buf)
+    free(buf);
+  buf = 0;
+
+  if (i == 'L') {  //last record put 3 consecutive times starting from end of prev. IOV
+    t1 = t3;
+    for (int i = 0; i < 92; ++i)
```
pohsun: I guess this 92 come from `EcalLaserCondTools::fillDb::nLmes`. Then please use a single constant throughout the class.

pohsun (changes requested):
In general, it's suggested to go through [the CMS naming rules](http://cms-sw.github.io/cms_coding_rules.html).

CalibCalorimetry/EcalTrivialCondModules/src/EcalLaserCondTools.cc
grasph: Why ? It won't make any difference. The compiler will produce exactly the same code as there is no difficulty in this piece of code to detect that it can be estimated at compile time.

pohsun: This rule is clearly shown as [point 7.5](http://cms-sw.github.io/cms_coding_rules.html#7----design-and-coding-guidelines-1). I tend to fix them given the cost is low.
However, I don't insist on the patch.

CalibCalorimetry/EcalTrivialCondModules/interface/EcalLaserCondTools.h
pohsun: @grasph Please fix also
`#ifndef` at L.1 and the comment at the last line `#endif // EcalLaserCondTools_h not defined`

CalibCalorimetry/EcalTrivialCondModules/src/EcalLaserCondTools.cc
grasph: Rule applied.

Martin-Grunewald:
-1
rebase needed

perrotta:
-1
Rebase needed

grasph:
which revision I should rebase on ? I did rebase on official-cmssw/master before the push.

fwyzard:
hi @grasph ,
this should work, and squash all your commits in a single one:
```
git checkout ecal-laser-cond-tools-2
git reset d5d1a36b4dba60266358ded257a90c73266f5b12
git add CalibCalorimetry/EcalTrivialCondModules/
git commit -m'Add the modules used to insert ECAL laser correction sets into the conditions database'
# note down the commit hash, in my case it was 63e4acc5fe9e
git reset --hard official-cmssw/CMSSW_11_1_X
# replace this with the has you just noted
git cherry-pick 63e4acc5fe9e
git push -f my-cmssw ecal-laser-cond-tools-2
```

grasph:
hi Andrea,

thanks for the recipe.  I've applied it and I am now crossing my fingers.

Philippe.

christopheralanwest:
please test

pohsun:
+1

silviodonato:
+1

PR #29235, opened 2020-03-18
HCal: Fix calo sample precision data from SiPM hits

lwang046:
#### PR description:

A small fix to the hcal simulated calo sample precision data. Each precise bin represents charge in 0.5 ns instead of 1 ns. 
This change will affect the precise pulse shape of the calosample which is used in TDC simulation, thus we should expect changes in TDC related histograms.


kpedro88:
please test

civanch:
@lwang046 , can you, please, make a reference to the talk or to a plot in PR description. Have you any explanation for 126 differences in regression?

abdoulline:
@civanch 
Vladimir,  those DQM differences come solely from TDC-related plots for 2018/2021/2023/2026 wf's.  That's the exactly what the fix is supposed to do.     
We've had our internal investigation and tracked the issue with precision samples charge discrepancy (wrt regular CaloSamples) to the issue, which is now fixed by this PR.

civanch:
+1

silviodonato:
+1

PR #29236, opened 2020-03-18
Run3-sim50 Define transition region for QGSP_FTFP_BERT using parameter settings

bsunanda:
#### PR description:

Set the transition region among the 3 models in QGSP_FTFP_BERT using parameters defined in the cfg file

#### PR validation:

Use standard runTheMatrix script

Nothing special


bsunanda:
@cmsbuild Please test

SimG4Core/Application/python/g4SimHits_cfi.py
```diff
@@ -150,6 +150,9 @@
         FlagBERT    = cms.untracked.bool(False),
         EminFTFP    = cms.double(3.), # in GeV
         EmaxBERT    = cms.double(6.), # in GeV
+        EminQGSP    = cms.double(12.), # in GeV
```
civanch: @bsunanda , I do not in favor for addition of too many these parameter. From one hand, not a big deal_ from other ' these parameters are used a bit differently for 10.4 and 10.6

SimG4Core/PhysicsLists/plugins/QGSPCMS_FTFP_BERT.cc
```diff
@@ -18,9 +19,18 @@ QGSPCMS_FTFP_BERT::QGSPCMS_FTFP_BERT(const edm::ParameterSet& p) : PhysicsList(p
   bool hadPhys = p.getUntrackedParameter<bool>("HadPhysics", true);
   bool tracking = p.getParameter<bool>("TrackingCut");
   double timeLimit = p.getParameter<double>("MaxTrackTime") * CLHEP::ns;
+  double minFTFP = p.getParameter<double>("EminFTFP") * CLHEP::GeV;
+  double maxBERT = p.getParameter<double>("EmaxBERT") * CLHEP::GeV;
+  double minQGSP = p.getParameter<double>("EminQGSP") * CLHEP::GeV;
```
civanch: @bsunanda , eminQGSP is fixed in any way to 12 GeV, addition of it differently inside CMSSW is possible but only open extra freedom.

SimG4Core/PhysicsLists/plugins/QGSPCMS_FTFP_BERT.cc
```diff
@@ -18,9 +19,18 @@ QGSPCMS_FTFP_BERT::QGSPCMS_FTFP_BERT(const edm::ParameterSet& p) : PhysicsList(p
   bool hadPhys = p.getUntrackedParameter<bool>("HadPhysics", true);
   bool tracking = p.getParameter<bool>("TrackingCut");
   double timeLimit = p.getParameter<double>("MaxTrackTime") * CLHEP::ns;
+  double minFTFP = p.getParameter<double>("EminFTFP") * CLHEP::GeV;
+  double maxBERT = p.getParameter<double>("EmaxBERT") * CLHEP::GeV;
```
civanch: maxFTFP = 25 GeV in all PhysLists - we should not add this parameter - extra level of freedom without a possibility to control.

civanch:
Hi Sunanda, I do not agree with some of solutions you propose in this PR. Ideally, we should wait until 10.6 will be integrated into CMSSW and then use proper interfaces coherent with 10.6. At the same time, you make a substantial edition, so better to integrate all this, I will try to clean it up in the next PR.

civanch:
+1

bsunanda:
@civanch yes i had only 10.6 in mind. please change once we shift to 10.6

silviodonato:
+1

PR #29246, opened 2020-03-19
Add ONNXruntime python test

mrodozov:
#### PR description:
Add a test for onnxruntime python interface we added here:
https://github.com/cms-sw/cmsdist/pull/5668
Taken from this examples: 
https://github.com/microsoft/onnxruntime/tree/master/docs/python/examples
uses example model from onnxruntime datasets folder 
#### PR validation:

The test runs

### DIFF
--- a/PhysicsTools/ONNXRuntime/test/BuildFile.xml
+++ b/PhysicsTools/ONNXRuntime/test/BuildFile.xml
@@ -1,9 +1,9 @@
-<use name="py3-onnxruntime" />
-<test name="test_py_onnxruntime" command="plot_load_and_predict.py"/>
+<test name="test_py_onnxruntime" command="plot_load_and_predict.py">
+  <use name="onnxruntime"/>
+</test>
 <bin name="testONNXRuntime" file="testRunner.cpp, testONNXRuntime.cc">
     <use name="boost_filesystem" />
     <use name="cppunit" />
-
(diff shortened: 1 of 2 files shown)


mrodozov:
please test with https://github.com/cms-sw/cmsdist/pull/5668

smuzaffar:
hold
we need to fix onnxruntime python for aarch64

slava77:
> hold
> we need to fix onnxruntime python for aarch64

@smuzaffar is this hold still active ?

smuzaffar:
unhold

slava77:
+1

for https://github.com/cms-sw/cmssw/pull/29246	ee1d704 
- looks as described
- the set up seems somewhat questionable to me, because we are adding a unit test that's supposed to be testing an external itself rather than anything in CMSSW. Shouldn't this be a part of a test of the external?

smuzaffar:
> the set up seems somewhat questionable to me, because we are adding a unit test that's supposed to be testing an external itself rather than anything in CMSSW. Shouldn't this be a part of a test of the external?

@slava77 , you are right but having unit test in cmssw also makes sure that environment of the tool is properly setup within cmssw.

silviodonato:
merge
@santocch this PR is adding a unit test

santocch:
+1

PR #29248, opened 2020-03-20
Fix infrequent missing LuminosityBlock

Dr15Jones:
#### PR description:

Under very heavy load where a thread is paused at just the right part of the function, the LuminosityBlock cache would be empty when the begin LuminosityBlock transition starts. The fix was to not have a local variable holding the std::shared_ptr<> to the LuminosityBlockPrincipal. 
Additionally, removed the unnecessary LuminosityBlockProcessingStatus arguments from streamEndLumiAsync.

#### PR validation:

I was able to modify one of the framework tests to be able to reliably fail. After this fix, it no longer fails.


Dr15Jones:
please test

Dr15Jones:
The comparison failures appear to be from the HGCal pull request.

Dr15Jones:
Please test

silviodonato:
+1

PR #29254, opened 2020-03-20
PUPPI MET in RECO

ahinzmann:
#### PR description:

This PR computes puppiNoLep weights and puppi MET in RECO and stores the puppiNoLep weight as a value map. PUPPI MET is clustered based on standard PF candidates together with the value map of weights. A new boolean "isWeighted" is added to the reco::PFMet interface to indicate whether weights were applied during jet clustering. The packed candidates pick up the value map such that puppi does not need to be re-computed in MiniAOD.

The same was done for the puppi and PUPPI jets in #28936.

Other changes:
- Produce puppiForMET directly in PuppiProducer instead of a two step procedure puppiNoLep->puppiForMET and rename it to puppiNoLep, consistent with the name in PackedCandidates
- Implement fillDescriptions for PFMetProducer
- Rename PFMET_cfi to pfMet_cfi everywhere (for consistency with the module name)
- Fix input jet collection of pfMetEI
- Remove pfMetEI from RecoMET_EventContent (however, leave EITopPAG_EventContent_cff untouched)
- Resolve issue #29264
- By moving to the same input PF candidate collection for puppi MET and puppi jets (with just different weights), PUPPI MET significance and unclustered pT computation then don't rely on dR matching as implemented in #29331 anymore.
- Store ref to PF candidate inside pat muons (just like done for all other PAT objects). PUPPI MET significance and unclustered pT computation then don't rely on dR matching as implemented in #29385 anymore.
- Fix particle multiplicity computation in xy-correction for PUPPI MET. Previously, particles with a weight of 0 could also count in the multiplicity as a function of which the xy-corrections are computed. Since xy-correction are not needed/recommended for PUPPI MET, this fix is not backported.
- For HI workflows, raise candidate pT threshold for chsMet and trkMet to 999. consistent with pfMet.

#### PR validation:

runTheMatrix -l limited

No backport foreseen.


slava77:
@ahinzmann 
will this need to be tested together with #28936 ?
Please clarify.
Thank you.

ahinzmann:
This can be tested standalone. Once #28936 is merged in CMSSW, I can do the merge here as well. I'm still finishing one round of local testing. Once they finish, it would be good to test here.

ahinzmann:
local test ran successful. Could this version be tested?

slava77:
@cmsbuild please test

slava77:
@cmsbuild please test

ahinzmann:
testRecoMETMETProducers fixed
differences in PUPPI MET significance fixed
10024.0 and 136.8311 ran successfully

perrotta:
@cmsbuild please test

jfernan2:
+1
Small changes for some DQM plots.

ahinzmann:
Not yet tested locally after merging with master

slava77:
> Not yet tested locally after merging with master

please post a message when it's ready/tested.

Intermediate commits do not have to be pushed to the remove individually if the work is not done

perrotta:
@ahinzmann , the [last commit](https://github.com/cms-sw/cmssw/pull/29254/commits/1d010b8227800ee30ea10e31a6b6320e00efb893) is a bug fix for the issue reported in https://github.com/cms-sw/cmssw/pull/28936#issuecomment-603732511

Shouldn't/couldn't it be submitted separately with a dedicated PR?

ahinzmann:
> @ahinzmann , the [last commit](https://github.com/cms-sw/cmssw/pull/29254/commits/1d010b8227800ee30ea10e31a6b6320e00efb893) is a bug fix for the issue reported in [#28936 (comment)](https://github.com/cms-sw/cmssw/pull/28936#issuecomment-603732511)
> 
> Shouldn't/couldn't it be submitted separately with a dedicated PR?

#29296

ahinzmann:
Local tests ran successfully. Ready to test here.

slava77:
@cmsbuild please test

ahinzmann:
ok, what remains to be understood are the differences in
1. unclustered energy uncertainties (index 10+11)
https://cmssdt.cern.ch/SDT/jenkins-artifacts/baseLineComparisons/CMSSW_11_1_X_2020-03-25-1100+e7c2d8/35638/validateJR/all_OldVSNew_RunJetHT2016EreMINIAODwf136p7611/all_OldVSNew_RunJetHT2016EreMINIAODwf136p7611c_patMETs_slimmedMETsPuppi__PAT_obj_0__uncertainties__10__dsumEt89.png
2. xy-corrections (index 34,5,6,7):
https://cmssdt.cern.ch/SDT/jenkins-artifacts/baseLineComparisons/CMSSW_11_1_X_2020-03-25-1100+e7c2d8/35638/validateJR/all_OldVSNew_RunJetHT2016EreMINIAODwf136p7611/all_OldVSNew_RunJetHT2016EreMINIAODwf136p7611c_patMETs_slimmedMETsPuppi__PAT_obj_0__corrections__3__dpx110.png
The rest works now. I'll keep digging where this comes from.

ahinzmann:
> ok, what remains to be understood are the differences in
> 
>     1. unclustered energy uncertainties (index 10+11)
>        https://cmssdt.cern.ch/SDT/jenkins-artifacts/baseLineComparisons/CMSSW_11_1_X_2020-03-25-1100+e7c2d8/35638/validateJR/all_OldVSNew_RunJetHT2016EreMINIAODwf136p7611/all_OldVSNew_RunJetHT2016EreMINIAODwf136p7611c_patMETs_slimmedMETsPuppi__PAT_obj_0__uncertainties__10__dsumEt89.png
> 
>     2. xy-corrections (index 34,5,6,7):
>        https://cmssdt.cern.ch/SDT/jenkins-artifacts/baseLineComparisons/CMSSW_11_1_X_2020-03-25-1100+e7c2d8/35638/validateJR/all_OldVSNew_RunJetHT2016EreMINIAODwf136p7611/all_OldVSNew_RunJetHT2016EreMINIAODwf136p7611c_patMETs_slimmedMETsPuppi__PAT_obj_0__corrections__3__dpx110.png
>        The rest works now. I'll keep digging where this comes from.

Both features are understood. They turn out to be fixes, see PR description for details. A backport of these fixes is also provided in #29330.
Please test again.

perrotta:
@cmsbuild please test

slava77:
> Both features are understood. They turn out to be fixes, see PR description for details. A backport of these fixes is also provided in #29330.

I'm trying to understand if it's worth making a separate bugfix PR first, so that we have a more clear backport match.
@ahinzmann please comment if this is feasible/practical on your side.
Thanks

ahinzmann:
> > Both features are understood. They turn out to be fixes, see PR description for details. A backport of these fixes is also provided in #29330.
> 
> I'm trying to understand if it's worth making a separate bugfix PR first, so that we have a more clear backport match.
> @ahinzmann please comment if this is feasible/practical on your side.
> Thanks

I would have liked to do that, but unfortunately it is hard to "break" PUPPI MET in the new implementation to reproduce the old behavior, without much of code overhead. So I believe it is unavoidable to fix it in here.
One option would be to implement the temporary fix (with dR) I submitted for 10_6_X also in the master and then reverting it again here and the do the real fix (PUPPI in RECO), but I don't see a merit in this way.

slava77:
> I would have liked to do that, but unfortunately it is hard to "break" PUPPI MET in the new implementation to reproduce the old behavior, without much of code overhead. So I believe it is unavoidable to fix it in here.

I meant a  fix to be applied as is (no need for a config switch). On top of that then I'd expect this PR to become a more clear migration rather than a mix of migration + fixes, which requires more effort.

> One option would be to implement the temporary fix (with dR) I submitted for 10_6_X also in the master and then reverting it again here and do the real fix (PUPPI in RECO), but I don't see a merit in this way.

ahinzmann:
> > I would have liked to do that, but unfortunately it is hard to "break" PUPPI MET in the new implementation to reproduce the old behavior, without much of code overhead. So I believe it is unavoidable to fix it in here.
> 
> I meant a fix to be applied as is (no need for a config switch). On top of that then I'd expect this PR to become a more clear migration rather than a mix of migration + fixes, which requires more effort.
> 
> > One option would be to implement the temporary fix (with dR) I submitted for 10_6_X also in the master and then reverting it again here and do the real fix (PUPPI in RECO), but I don't see a merit in this way.

Ok, I've submitted the temporary fix to the master. We can now test the real fix (PUPPI in RECO) #29954 against the temporary fix in (dR-matching) in #29331, such that remaining differences are purely technical. When #29331 is merged, this PR becomes purely technical.

slava77:
@cmsbuild please test

ahinzmann:
Looks like the test didn't pick up the IB of last night yet.

perrotta:
test parameters
pull_request =

perrotta:
@cmsbuild please test
(Let retry now. I don't know if I was able to reset the test parameters with the previous command: it shouldn't matter, anyhow, since that additional PR is already present in the latest IB)

slava77:
> Reco comparison results: 228 differences found in the comparisons

the most recent tests included #29331  in the baseline.
Are the remaining changes expected?

ahinzmann:
> > Reco comparison results: 228 differences found in the comparisons
> 
> the most recent tests included #29331 in the baseline.
> Are the remaining changes expected?

The changes in MET corrections are due to the xy-correction modification and expected.
The small remaining change in MET unclustered uncertainty is not expected. It's smaller than relevant for physics, but I'm investigating where it originates (currently checking electron, muon, tau and photon  candidates substraction  from unclustered).

ahinzmann:
#29385 resolves 28 of 100 diffs observed in PUPPI MET unclustered energy uncertaintiy seen here. Investigating the remaining cases.

slava77:
@cmsbuild please test

slava77:
code-checks

@smuzaffar 
is there a problem with the bot or jenkins ?
The last "code-checks" started by the bot didn't run

smuzaffar:
@slava77 , yes the cmsdev machine where it was running the job has no disk space left thatis why job was failing.

slava77:
@cmsbuild please test

jfernan2:
@ahinzmann there are tiny changes in METunc for JetMET/ METValidation/slimmedMETsPuppi in wf 25202 and 10024, I don't know if they are still expected:
https://tinyurl.com/ya3vfq26
https://tinyurl.com/y8xxopy3
Thanks

slava77:
> Reco comparison results: 141 differences found in the comparisons

- most of these are changes in XY corrections (expected)
- HI wf 140.56 is set to write AOD, it doesn't have pfMetEI anymore (expected)
- 158.0 miniAOD has some changes in corrections 10 Chs and 11 Trk both in puppi and regular MET 
e.g.
all_mini_OldVSNew_HydjetQB12ppRECOin2018wf158p0c_patMETs_slimmedMETsPuppi__RECO_obj_0__corrections__11__dpx127
all_mini_OldVSNew_HydjetQB12ppRECOin2018wf158p0c_patMETs_slimmedMETs__RECO_obj_0__corrections__11__dpx67
are these expected?

slava77:
> @ahinzmann there are tiny changes in METunc for JetMET/ METValidation/slimmedMETsPuppi in wf 25202 and 10024, I don't know if they are still expected:
> https://tinyurl.com/ya3vfq26
> https://tinyurl.com/y8xxopy3
> Thanks

it's a bit surprising that these do not show up in reco comparisons; I checked in the jenkins output files explicitly and uncertainties_ are the same for Muon/Electron Up/Down.

ahinzmann:
@jfernan2. Changes in electron and muon uncertainties are not expected. I wasn't able to reproduce the diff locally with 25202 looking at the root files containing MiniAOD. So I don't understand how this can show up in DQM that simply reads the MiniAOD variables. Maybe some numerical precision issue.

I got rid of dR-matching as fall-back for ref-matching in the MET tools for all objects, by actually saving the PF muon ref inside PAT muons (just like its done in all other PAT objects and also in all slimmedObjects).

@slava77. I didn't investiate 158.0 yet. I changed back the alias of pfMet to what it used to be, but I assume this doesn't actually affect the numbers.

slava77:
@cmsbuild please test

ahinzmann:
> > Reco comparison results: 141 differences found in the comparisons
> 
>     * most of these are changes in XY corrections (expected)
> 
>     * HI wf 140.56 is set to write AOD, it doesn't have pfMetEI anymore (expected)
> 
>     * 158.0 miniAOD has some changes in corrections 10 Chs and 11 Trk both in puppi and regular MET
>       e.g.
>       all_mini_OldVSNew_HydjetQB12ppRECOin2018wf158p0c_patMETs_slimmedMETsPuppi__RECO_obj_0__corrections__11__dpx127
>       all_mini_OldVSNew_HydjetQB12ppRECOin2018wf158p0c_patMETs_slimmedMETs__RECO_obj_0__corrections__11__dpx67
>       are these expected?

The reason is the following. The candidate pT threshold for pfMet in HI workflows is set to 999. (I understand they are not really needed from @FHead). However for CHS-MET and TRK-MET the threshold was overwritten to 0. With this PR, the miniAOD_tools and runMETCorrectionsAndUncertainties apply the 999. also consistently for CHS-MET and TRK-MET.

slava77:
> The reason is the following. The candidate pT threshold for pfMet in HI workflows is set to 999. (I understand they are not really needed from @FHead). However for CHS-MET and TRK-MET the threshold was overwritten to 0. With this PR, the miniAOD_tools and runMETCorrectionsAndUncertainties apply the 999. also consistently for CHS-MET and TRK-MET.

Thanks. So, this is also expected.

The only thing remaining is the change in the DQM Electron/Muon Up/Down uncertainties.
Perhaps adding printouts in the Validation/RecoMET/plugins/METTester.cc may help to debug.

ahinzmann:
> > The reason is the following. The candidate pT threshold for pfMet in HI workflows is set to 999. (I understand they are not really needed from @FHead). However for CHS-MET and TRK-MET the threshold was overwritten to 0. With this PR, the miniAOD_tools and runMETCorrectionsAndUncertainties apply the 999. also consistently for CHS-MET and TRK-MET.
> 
> Thanks. So, this is also expected.
> 
> The only thing remaining is the change in the DQM Electron/Muon Up/Down uncertainties.
> Perhaps adding printouts in the Validation/RecoMET/plugins/METTester.cc may help to debug.

Well the difference between what is stored in the root file and what is in the histogram is that the difference to nominal MET is computed:
https://github.com/cms-sw/cmssw/blob/master/Validation/RecoMET/plugins/METTester.cc#L441
And some adding and transforming to a Vector2:
https://github.com/cms-sw/cmssw/blob/703f4fc1732e0410acad217a4acc70c739a8741e/DataFormats/PatCandidates/src/MET.cc#L246

jfernan2:
+1

santocch:
+1

ahinzmann:
> > > The reason is the following. The candidate pT threshold for pfMet in HI workflows is set to 999. (I understand they are not really needed from @FHead). However for CHS-MET and TRK-MET the threshold was overwritten to 0. With this PR, the miniAOD_tools and runMETCorrectionsAndUncertainties apply the 999. also consistently for CHS-MET and TRK-MET.
> > 
> > 
> > Thanks. So, this is also expected.
> > The only thing remaining is the change in the DQM Electron/Muon Up/Down uncertainties.
> > Perhaps adding printouts in the Validation/RecoMET/plugins/METTester.cc may help to debug.
> 
> Well the difference between what is stored in the root file and what is in the histogram is that the difference to nominal MET is computed:
> https://github.com/cms-sw/cmssw/blob/master/Validation/RecoMET/plugins/METTester.cc#L441
> And some adding and transforming to a Vector2:
> 
> https://github.com/cms-sw/cmssw/blob/703f4fc1732e0410acad217a4acc70c739a8741e/DataFormats/PatCandidates/src/MET.cc#L246
> 
> So I would think this is maybe just numerical differences.

METPackedUncertainties are converted from float16 precision:
https://github.com/cms-sw/cmssw/blob/master/DataFormats/PatCandidates/src/MET.cc#L431
So the observed differences of 10^-4 may be due to that.

RecoMET/METProducers/src/PFMETProducer.cc
```diff
@@ -95,12 +120,50 @@ namespace cms {
...
+    desc.add<double>("globalThreshold", 0.);
+    desc.add<std::string>("alias", "pfMet");
+    desc.addOptional<edm::InputTag>("srcJets");
+    desc.addOptional<std::vector<edm::InputTag>>("srcLeptons");
+    desc.addOptional<std::string>("srcJetSF");
+    desc.addOptional<std::string>("srcJetResPt");
+    desc.addOptional<std::string>("srcJetResPhi");
+    desc.addOptional<edm::InputTag>("srcRho");
+    edm::ParameterSetDescription params;
+    params.setAllowAnything();
```
slava77, line 155: it seems like the format is rather fixed. Are these always required (just with a swap of data vs MC)?
If so, it seems better to define these without defaults so that user code is forced to use them

RecoMET/METProducers/src/PFMETProducer.cc
```diff
@@ -95,12 +120,50 @@ namespace cms {
...
 
     return cov;
   }
 
+  void PFMETProducer::fillDescriptions(edm::ConfigurationDescriptions& descriptions) {
+    edm::ParameterSetDescription desc;
+    desc.add<edm::InputTag>("src", edm::InputTag("particleFlow"));
+    desc.add<bool>("calculateSignificance", false);
+    desc.add<double>("globalThreshold", 0.);
+    desc.add<std::string>("alias", "pfMet");
```
slava77, line 147: this is a rather questionable default, considering how many variants are cloned from this module and that all of them run in the same job in most cases.
```"@module_label"``` seems more appropriate

File: RecoMET/METProducers/src/PFMETProducer.cc (hunk 1 above)
Comment (ahinzmann) [reply RecoMET/METProducers/src/PFMETProducer.cc:155]: the params are needed when calculateSignificance=True. What do you suggest to add/change?

Comment (slava77) [reply RecoMET/METProducers/src/PFMETProducer.cc:155]: there is a "switch" option in the PSet validation https://twiki.cern.ch/twiki/bin/view/CMSPublic/SWGuideConfigurationValidationAndHelp#A_Switch

**Comment by ahinzmann** [comment]
don't test yet

File: RecoMET/METProducers/src/PFMETProducer.cc (hunk 1 above)
Comment (ahinzmann) [reply RecoMET/METProducers/src/PFMETProducer.cc:138]: hmm, with a switch, they are required to be present when calculateSignficicance is true, but the plugin doesn't accept them given when calculateSignficance is false.
how do I implement "required if true, optional if false"?

Comment (slava77) [reply RecoMET/METProducers/src/PFMETProducer.cc:138]: I thought that the logic is "required if true; not needed otherwise".
It sounds like the calculateSignficicance is not controlling anything and the parameter is needed anyways.
My earlier comment was about the setAllowAnything. Perhaps you can anyways define the structure (drop setAllowAnything) and leave it optional?

Comment (slava77) [reply RecoMET/METProducers/src/PFMETProducer.cc:138]: I thought that the logic is "required if true; not needed otherwise".
It sounds like the calculateSignficicance is not controlling anything and the parameter is needed anyways.
My earlier comment was about the setAllowAnything. Perhaps you can anyways define the structure (drop setAllowAnything) and leave it optional?

Comment (ahinzmann) [reply RecoMET/METProducers/src/PFMETProducer.cc:138]: if calculateSignficance: srcJets, ..., parameters are required.
if not calculateSignficance: srcJets, ..., parameters are not needed, but there is also no reason to forbit them.
I don't know how to implement this

Comment (ahinzmann) [reply RecoMET/METProducers/src/PFMETProducer.cc:138]: ok, so not use switch and go back to addOptional. understood.

Comment (ahinzmann) [reply RecoMET/METProducers/src/PFMETProducer.cc:145]: params.setAllowAnything(); on the other hand is not used directly in this module, but in another cc file. can look into this later.

**Comment by slava77** [comment]
> don't test yet

please clarify explicitly when it is ready
Thanks

**Comment by ahinzmann** [comment]
it's testable again. I just didn't get to replace the setAllowAnything.

**Comment by slava77** [comment]
@cmsbuild please test

File: PhysicsTools/PatAlgos/python/slimming/applySubstructure_cff.py [hunk 3]
```diff
@@ -13,13 +13,21 @@ def applySubstructure( process, postfix="" ) :
 
 
     # Configure the RECO jets
-    from RecoJets.JetProducers.ak4PFJets_cfi import ak4PFJetsPuppi
     from RecoJets.JetProducers.ak8PFJets_cfi import ak8PFJetsPuppi, ak8PFJetsPuppiSoftDrop, ak8PFJetsPuppiConstituents
+    setattr(process,'ak8PFJetsPuppi'+postfix,ak8PFJetsPuppi.clone())
+    setattr(process,'ak8PFJetsPuppiConstituents'+postfix, ak8PFJetsPuppiConstituents.clone(cut = cms.string('pt > 170.0 && abs(rapidity()) < 2.4') ))
+    setattr(process,'ak8PFJetsPuppiSoftDrop'+postfix, ak8PFJetsPuppiSoftDrop.clone( src = cms.InputTag('ak8PFJetsPuppiConstituents'+postfix, 'constituents') ))
```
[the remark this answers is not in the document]
slava77: not addressed yet

PhysicsTools/PatAlgos/python/slimming/miniAOD_tools.py
```diff
@@ -198,13 +198,9 @@ def miniAOD_customizeCommon(process):
...
 
-    process.pfMetCHS = cms.EDProducer("PFMETProducer",
-                                      src = cms.InputTag("CHSCands"),
-                                      alias = cms.string('pfMet'),
-                                      globalThreshold = cms.double(0.0),
-                                      calculateSignificance = cms.bool(False),
-                                      )
-    task.add(process.pfMetCHS)    
+    from RecoMET.METProducers.pfMet_cfi import pfMet
+    process.pfMetCHS = pfMet.clone(src = cms.InputTag('CHSCands'))
```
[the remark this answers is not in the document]
slava77: not addressed yet

PhysicsTools/PatAlgos/python/slimming/miniAOD_tools.py
```diff
@@ -222,13 +218,7 @@ def miniAOD_customizeCommon(process):
...
     task.add(process.TrkCands)
 
-    process.pfMetTrk = cms.EDProducer("PFMETProducer",
-                                      src = cms.InputTag("TrkCands"),
-                                      alias = cms.string('pfMet'),
-                                      globalThreshold = cms.double(0.0),
-                                      calculateSignificance = cms.bool(False),
-                                      )
-
+    process.pfMetTrk = pfMet.clone(src = cms.InputTag('TrkCands'))
```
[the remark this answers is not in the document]
slava77: not addressed yet

PhysicsTools/PatAlgos/python/slimming/miniAOD_tools.py
```diff
@@ -375,6 +365,17 @@ def miniAOD_customizeCommon(process):
         process.makePatTausTask, _makePatTausTaskWithTauReReco
         )
     
+    # Adding puppi jets
+    process.load('CommonTools.PileupAlgos.Puppi_cff')
+    process.load('RecoJets.JetProducers.ak4PFJets_cfi')
+    from Configuration.Eras.Modifier_run2_miniAOD_80XLegacy_cff import run2_miniAOD_80XLegacy
+    from Configuration.Eras.Modifier_run2_miniAOD_94XFall17_cff import run2_miniAOD_94XFall17
+    from Configuration.Eras.Modifier_pA_2016_cff import pA_2016
+    from Configuration.Eras.Modifier_pp_on_AA_2018_cff import pp_on_AA_2018
```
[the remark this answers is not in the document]
slava77: not addressed yet

PhysicsTools/PatAlgos/python/slimming/miniAOD_tools.py
```diff
@@ -426,11 +427,14 @@ def miniAOD_customizeCommon(process):
     run2_miniAOD_94XFall17.toModify(process.updatedPatJetsTransientCorrectedSlimmedDeepFlavour, addTagInfos = False )
     
     ## puppi met
-    from PhysicsTools.PatAlgos.slimming.puppiForMET_cff import makePuppies
-    makePuppies( process );
-
+    process.load('RecoMET.METProducers.pfMetPuppi_cfi')
+    from Configuration.Eras.Modifier_run2_miniAOD_80XLegacy_cff import run2_miniAOD_80XLegacy
+    from Configuration.Eras.Modifier_run2_miniAOD_94XFall17_cff import run2_miniAOD_94XFall17
```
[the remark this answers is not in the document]
slava77: not addressed yet

PhysicsTools/PatUtils/python/tools/runMETCorrectionsAndUncertainties.py
```diff
@@ -1675,13 +1682,8 @@ def miniAODConfigurationPre(self, process, patMetModuleSequence, pfCandCollectio
...
         addToProcessAndTask("pfCHS", pfCHS, process, task)
-        pfMetCHS = cms.EDProducer("PFMETProducer",
-                                  src = cms.InputTag('pfCHS'),
-                                  alias = cms.string('pfMet'),
-                                  globalThreshold = cms.double(0.0),
-                                  calculateSignificance = cms.bool(False),
-                                  )            
-
+        from RecoMET.METProducers.pfMet_cfi import pfMet
+        pfMetCHS = pfMet.clone(src = cms.InputTag('pfCHS'))
```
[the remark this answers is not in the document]
slava77: not addressed yet

PhysicsTools/PatUtils/python/tools/runMETCorrectionsAndUncertainties.py
```diff
@@ -1698,13 +1700,7 @@ def miniAODConfigurationPre(self, process, patMetModuleSequence, pfCandCollectio
...
         pfTrk = cms.EDFilter("CandPtrSelector", src = cms.InputTag("packedPFCandidates"), cut = cms.string("fromPV(0) > 0 && charge()!=0"))
         addToProcessAndTask("pfTrk", pfTrk, process, task)
-        pfMetTrk = cms.EDProducer("PFMETProducer",
-                                  src = cms.InputTag('pfTrk'),
-                                  alias = cms.string('pfMet'),
-                                  globalThreshold = cms.double(0.0),
-                                  calculateSignificance = cms.bool(False),
-                                  )            
-
+        pfMetTrk = pfMet.clone(src = cms.InputTag('pfTrk'))
```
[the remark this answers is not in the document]
slava77: not addressed yet

RecoMET/METAlgorithms/src/METAlgo.cc
```diff
@@ -13,12 +13,16 @@
...
 
 //____________________________________________________________________________||
-CommonMETData METAlgo::run(const edm::View<reco::Candidate>& candidates, double globalThreshold) {
+CommonMETData METAlgo::run(const edm::View<reco::Candidate>& candidates,
+                           double globalThreshold,
+                           edm::ValueMap<float> const* weights) {
   math::XYZTLorentzVector p4;
-  for (auto cand = candidates.begin(); cand != candidates.end(); ++cand) {
-    if (!(cand->et() > globalThreshold))
+  for (auto candPtr = candidates.ptrs().begin(); candPtr != candidates.ptrs().end(); ++candPtr) {
```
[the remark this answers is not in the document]
slava77: not addressed yet

RecoMET/METAlgorithms/src/METAlgo.cc
```diff
@@ -33,10 +37,12 @@ CommonMETData METAlgo::run(const edm::View<reco::Candidate>& candidates, double
   ret.met = met.Pt();
 
   double et = 0.0;
-  for (auto cand = candidates.begin(); cand != candidates.end(); ++cand) {
-    if (!(cand->et() > globalThreshold))
+  for (auto candPtr = candidates.ptrs().begin(); candPtr != candidates.ptrs().end(); ++candPtr) {
```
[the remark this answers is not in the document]
slava77: not addressed yet

RecoMET/METAlgorithms/src/PFSpecificAlgo.cc
```diff
@@ -24,12 +24,13 @@ SpecificPFMETData PFSpecificAlgo::run(const edm::View<reco::Candidate>& pfCands)
   double type6Et = 0.0;
   double type7Et = 0.0;
 
-  for (edm::View<reco::Candidate>::const_iterator iPfCand = pfCands.begin(); iPfCand != pfCands.end(); ++iPfCand) {
-    const reco::Candidate* pfCand = dynamic_cast<const reco::Candidate*>(&(*iPfCand));
+  for (auto pfCandPtr = pfCands.ptrs().begin(); pfCandPtr != pfCands.ptrs().end(); ++pfCandPtr) {
```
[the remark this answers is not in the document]
slava77: not addressed yet

RecoMET/METProducers/python/pfChMet_cfi.py
```diff
@@ -10,12 +11,9 @@
     )
 
 ##____________________________________________________________________________||
-pfChMet = cms.EDProducer(
-    "PFMETProducer",
+pfChMet = pfMet.clone(
     src = cms.InputTag("particleFlowForChargedMET"),
     alias = cms.string('pfChMet'),
```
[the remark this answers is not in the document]
slava77: not addressed yet

slava77 (commented):
here is a bunch of technical comments, mostly repeats from the last iteration of review

RecoMET/METProducers/src/PFMETProducer.cc
slava77: > params.setAllowAnything(); on the other hand is not used directly in this module, but in another cc file. can look into this another time.

it's a fair point that the outside "plugin" should provide this description (to be inserted here). I'm fine for a later PR, but if updates are coming, please add a "FIXME" here in the comments.
Thank you.

ahinzmann:
Sorry, I somehow missed half of the comments (maybe hidden in github history). Ready to test now.

perrotta:
@cmsbuild please test

jfernan2:
+1

slava77:
> By moving to the same input PF candidate collection for puppi MET and puppi jets (with just different weights), PUPPI MET significance and unclustered pT computation then don't rely on dR matching as implemented in #29331 anymore.

was this supposed to refer (also) to #29385?
apparently now (after #29385 intergation; at least as early as in 8b895b9) there is a new diff appearing in comparisons for slimmedMETsPuppi in metSumPtUnclustered.
in the re-miniAOD wf 136.7611
all_OldVSNew_RunJetHT2016EreMINIAODwf136p7611c_log10patMETs_slimmedMETsPuppi__PAT_obj_metSumPtUnclustered
In a manual scan of the two files from the last diff in jenkins out of 100 events: 
- 99 have a relative difference in metSumPtUnclustered of below 1E-8 (OK)
- 1 event (run 277069 event 35877918) has it at about 0.02 relative or 5.9 GeV absolute with a new value of 373.5 GeV

slava77:
in view of #29264 I see that in wf 11634.21 we still have ak8PFJetsPuppiSoftDropMass repeated both in the reco and in the PAT steps with this PR.
This is much better than the baseline having puppi, ak[48]PFJetsPuppi and ak8PFJetsPuppi{Constituents,SoftDrop,SoftDropMass} repeated.

PhysicsTools/PatAlgos/python/slimming/applySubstructure_cff.py
```diff
@@ -13,13 +13,21 @@ def applySubstructure( process, postfix="" ) :
...
+    setattr(process,'ak8PFJetsPuppiConstituents'+postfix, ak8PFJetsPuppiConstituents.clone(cut = cms.string('pt > 170.0 && abs(rapidity()) < 2.4') ))
+    setattr(process,'ak8PFJetsPuppiSoftDrop'+postfix, ak8PFJetsPuppiSoftDrop.clone( src = 'ak8PFJetsPuppiConstituents'+postfix+':constituents' ))
+    from Configuration.Eras.Modifier_run2_miniAOD_80XLegacy_cff import run2_miniAOD_80XLegacy
+    from Configuration.Eras.Modifier_run2_miniAOD_94XFall17_cff import run2_miniAOD_94XFall17
+    from Configuration.Eras.Modifier_pA_2016_cff import pA_2016
+    from Configuration.Eras.Modifier_pp_on_AA_2018_cff import pp_on_AA_2018
+    _rerun_puppijets_task = task.copy()
+    _rerun_puppijets_task.add(getattr(process,'ak8PFJetsPuppi'+postfix),
+                              getattr(process,'ak8PFJetsPuppiConstituents'+postfix),
+                              getattr(process,'ak8PFJetsPuppiSoftDrop'+postfix))
```
slava77, line 27: Shouldn't all this ```_rerun_puppijets_task``` be done only for the empty postfix?

Also, related to https://github.com/cms-sw/cmssw/pull/29254#issuecomment-620316191, down below on L157 there is an insertion of ```ak8PFJetsPuppiSoftDropMass```, which is available in the new puppi-in-reco setup.
I think that it should be added to this ```_rerun_puppijets_task``` and removed from L157.

slava77: > Shouldn't all this `_rerun_puppijets_task` be done only for the empty postfix?

I mean the era-driven part of it.
While for the non-empty post-fix, the modules better be added by default

slava77:
About differences in DQM Electron/Muon uncertainties https://github.com/cms-sw/cmssw/pull/29254#issuecomment-615235645
The uncertainties should be the same at run time and in storage because their ```set``` methods have pack()->unpack() cycle applied.
However, the MET itself as a reco candidate is double in memory, but it's streamed as Double32 (float on disk).

A few other DQM plots show differences in MET profiles mostly below float precision value. I think that these are actually in agreement with the uncertainty plots which just happen to split zero to +0/-0 slightly differently and show more visible bin migration effect.
A mitigation to this in METTester would be to add one more bin to have 0 in the center of a bin instead of in between two bins.

ahinzmann:
> > By moving to the same input PF candidate collection for puppi MET and puppi jets (with just different weights), PUPPI MET significance and unclustered pT computation then don't rely on dR matching as implemented in #29331 anymore.
> 
> was this supposed to refer (also) to #29385?
> apparently now (after #29385 intergation; at least as early as in [8b895b9](https://github.com/cms-sw/cmssw/commit/8b895b98458c755686225e24f1deab0c79413fb4)) there is a new diff appearing in comparisons for slimmedMETsPuppi in metSumPtUnclustered.
> in the re-miniAOD wf 136.7611
> all_OldVSNew_RunJetHT2016EreMINIAODwf136p7611c_log10patMETs_slimmedMETsPuppi__PAT_obj_metSumPtUnclustered
> In a manual scan of the two files from the last diff in jenkins out of 100 events:
> 
>     * 99 have a relative difference in metSumPtUnclustered of below 1E-8 (OK)
> 
>     * 1 event (run 277069 event 35877918) has it at about 0.02 relative or 5.9 GeV absolute with a new value of 373.5 GeV

I tracked it down to this change:
https://github.com/cms-sw/cmssw/pull/29254/files#diff-2917681d29a8d5cc61ead8bb9d6c4b9aR618

In this event a muon identified via PF reference matching (enabled in this PR) fails the dR matching. So the dR matching missed to remove one PF muon from PUPPI unclustered energy without this PR. I think it's not really worth fixing this in backports by tuning the dR criterion that was used in backports as the change in the quantity relevant for physics, i.e. MET significance changes from 1.47392 to 1.47931 due to this.

slava77:
@cmsbuild please test

jfernan2:
+1

slava77:
+1

for https://github.com/cms-sw/cmssw/pull/29254	d6e1687
- jenkins tests pass and comparisons with the baseline show differences in expected places in puppi MET related variables (see earlier discussion)

resolves #29264

silviodonato:
Do you have comments @santocch ?

slava77:
@silviodonato 
please include this in pre7

silviodonato:
merge

silviodonato:
We are getting some errors in the IB. Please look at https://github.com/cms-sw/cmssw/issues/29744

santocch:
+1

PhysicsTools/PatUtils/python/tools/runMETCorrectionsAndUncertainties.py
```diff
@@ -818,21 +824,16 @@ def getMETUncertainties(self, process, metType, metModName, electronCollection,
...
             #electron projection ==
             pfCandsNoJetsNoEle = cms.EDProducer("CandPtrProjector", 
                                                 src = cms.InputTag("pfCandsNoJets"+postfix),
                                                 veto = copy.copy(electronCollection),
-                                                useDeltaRforFootprint = cms.bool(False),
                                                 )
             if not self.getvalue("onMiniAOD"):
-              pfCandsNoJetsNoEle.useDeltaRforFootprint = True
               pfCandsNoJetsNoEle.veto = "pfeGammaToCandidate:electrons"
             addToProcessAndTask("pfCandsNoJetsNoEle"+postfix, pfCandsNoJetsNoEle, process, task)
```
slava77, line 838: I was debugging  differences in UL re-miniAOD between 106X and 112X, and was puzzled by the somewhat small but numerous changes in slimmedMETsNoHF UnclusteredEnUp/Down.
If I understood correctly (although I did not add any debug printouts to confirm), the problem is that the "NoHF" is based on ```noHFCands```, which are copies (not Ptrs) of the original PF cands https://github.com/cms-sw/cmssw/blob/3402e097caee10a4a7205dc67a80108939870f74/CommonTools/ParticleFlow/interface/GenericPFCandidateSelectorDefinition.h#L31-L34
So, the CandPtrProjector can not match the vetoing electron's sourceCandidatePtr (the original particleFlow) to the ```noHFCands``` (another check by the cand's sourceCandidatePtr may suffice though).
So, either we need an extension in CandPtrProjector to also check for the main collection sourceCandidatePtr or instead have ```useDeltaRforFootprint = True```

I'll re-post this as an issue.

PR #29252, opened 2020-03-20
EMTF configures from Global Tag conditions in MC for 2016 only

abrinke1:
We have unfortunately uncovered another error in the EMTF emulation for 2016, which had been masked by the previous one. [1]  As it turns out the emulator was not accessing the 2016 Global Tag conditions for most of the algorithm settings, creating a mix of 2016 and 2018 logic which is internally inconsistent.  Efe measured the efficiency in the latest RelVal samples (see attached slides), and due to this bug it is 10% lower in the negative endcap than in the positive.

Thankfully the fix is quite simple, as you can see in this pull request.  Efe's plots confirm that the efficiency looks good when re-running the newest RelVals with this patch applied, and is consistent with the efficiency in 2016 data. [2]

For historical background on the likely cause for this bug, between 2016 and 2017 we completely overhauled the EMTF emulator, in part to handle the new RPC inputs.  Unfortunately both the CondDB code and payloads for EMTF at that point were in disarray (very long story), and accessing the conditions payload was crashing the MC RelVals.  At that point (if my hazy memory serves) we decided to configure by firmware version only for data, and by "fake conditions" for MC, with 3 different "fake conditions" python config files.  Sometime subsequently the "fake conditions" for each year went away, so the MC emulation (again, for some but not all of the algorithm settings) used the default settings, instead of the time-dependent firmware version settings.  This worked fine when the "default" settings were also the "current" settings (which was true when the 2017 and 2018 MC were produced), but fails for 2016. re-emulation.  Long story short, our workflows had never been fully validated for legacy MC, and now we're finding the bugs - for which we apologize.

Let us know if you have any other questions.
Best regards,Andrew, Efe, Olivier, and Benjamin

[1] https://github.com/cms-sw/cmssw/pull/29080
[2] https://twiki.cern.ch/twiki/bin/view/CMSPublic/L1TMuonPerformanceICHEP16

[EMTF_UL16MC_validation_19.03.2020.pdf](https://github.com/cms-sw/cmssw/files/4359945/EMTF_UL16MC_validation_19.03.2020.pdf)

PR #29260, opened 2020-03-20
EMTF configures from Global Tag conditions in MC as well as data

abrinke1:
"master" version of https://github.com/cms-sw/cmssw/pull/29252
Intended for all future versions of EMTF emulator.

We have unfortunately uncovered another error in the EMTF emulation for 2016, which had been masked by the previous one. [1]  As it turns out the emulator was not accessing the 2016 or 2017 Global Tag conditions for most of the algorithm settings, creating a mix of 2016 or 2017 and 2018 logic which is internally inconsistent.  Efe measured the efficiency in the latest 2016 RelVal samples (see attached slides), and due to this bug it is 10% lower in the negative endcap than in the positive.

Thankfully the fix is quite simple, as you can see in this pull request. Efe's plots confirm that the efficiency looks good when re-running the newest RelVals with this patch applied, and is consistent with the efficiency in 2016 data. [2]

For historical background on the likely cause for this bug, between 2016 and 2017 we completely overhauled the EMTF emulator, in part to handle the new RPC inputs.  Unfortunately both the CondDB code and payloads for EMTF at that point were in disarray (very long story), and accessing the conditions payload was crashing the MC RelVals.  At that point (if my hazy memory serves) we decided to configure by firmware version only for data, and by "fake conditions" for MC, with 3 different "fake conditions" python config files.  Sometime subsequently the "fake conditions" for each year went away, so the MC emulation (again, for some but not all of the algorithm settings) used the default settings, instead of the time-dependent firmware version settings.  This worked fine when the "default" settings were also the "current" settings (which was true when the 2017 and 2018 MC were produced), but fails for 2016. re-emulation.  Long story short, our workflows had never been fully validated for legacy MC, and now we're finding the bugs - for which we apologize.

[1] #29080
[2] https://twiki.cern.ch/twiki/bin/view/CMSPublic/L1TMuonPerformanceICHEP16
[EMTF_UL16MC_validation_19.03.2020.pdf](https://github.com/cms-sw/cmssw/files/4361572/EMTF_UL16MC_validation_19.03.2020.pdf)

### DIFF
--- a/L1Trigger/L1TMuonEndCap/src/TrackFinder.cc
+++ b/L1Trigger/L1TMuonEndCap/src/TrackFinder.cc
@@ -189,7 +189,7 @@ void TrackFinder::process(const edm::Event& iEvent,
                      (sector - emtf::MIN_TRIGSECTOR);
 
       // Run-dependent configure. This overwrites many of the configurables passed by the python config file.
-      if (iEvent.isRealData() && fwConfig_) {
+      if (fwConfig_) {
         sector_processors_.at(es).configure_by_fw_version(condition_helper_.get_fw_version());
       }
 


rekovic:
please test

silviodonato:
@rekovic do you have any objection about merging #29260 and  #29262?

silviodonato:
urgent
see #29252

silviodonato:
merge
in order to have this PR in CMSSW_11_1_X_2020-03-23-1100 tests

silviodonato:
@abrinke1 are the changes in `250202.181 TTbar_13UP18+PREMIXUP18_PU25+DIGIPRMXLOCALUP18_PU25+RECOPRMXUP18_PU25+HARVESTUP18_PU25 ` expected?

rekovic:
> @rekovic do you have any objection about merging #29260 and #29252?

The PR is simple and straight forward, so I am OK with it.  
Let's see the comment of  @abrinke1 regarding the changes in 250202.181

abrinke1:
Hi @silviodonato , @rekovic ,

From what I can see, all these changes are to the L1T muon plots, and the distributions change only slightly. This is expected, as the EMTF algorithm has changed, but it is not a dramatic change.

https://cmssdt.cern.ch/SDT/jenkins-artifacts/baseLineComparisons/CMSSW_11_1_X_2020-03-20-1100+36ec38/35572/250202.181_TTbar_13UP18+TTbar_13UP18+PREMIXUP18_PU25+DIGIPRMXLOCALUP18_PU25+RECOPRMXUP18_PU25+HARVESTUP18_PU25/L1TEMU_L1TdeStage2uGMT_intermediate_muons_EMTF_pos.html

https://cmssdt.cern.ch/SDT/jenkins-artifacts/baseLineComparisons/CMSSW_11_1_X_2020-03-20-1100+36ec38/35572/250202.181_TTbar_13UP18+TTbar_13UP18+PREMIXUP18_PU25+DIGIPRMXLOCALUP18_PU25+RECOPRMXUP18_PU25+HARVESTUP18_PU25/L1T_L1TStage2uGMT_EMTFInput.html

PR #29262, opened 2020-03-21
EMTF March 2020 emulator update

jiafulow:
#### PR description:

This PR **adds/updates the interface to all the Phase-2 trigger primitives**. This is the first PR from a series of EMTF emulator updates planned for 2020.

The EMTF emulator needs to be upgraded to handle all the Phase-2 trigger primitives, including GEM, ME0, and iRPC. It's also nice to be able to handle DT/TwinMux. Majority of the work has been done during the development for Phase-2 EMTF++, and used for L1T Phase-2 Upgrade TDR studies. These changes are now being ported to the official CMSSW. For Run 3 specifically, the GE1/1 interface is needed to continue the rest of the EMTF-GEM integration.

There are a lot of changes because what's in the existing EMTF emulator related to GEM/ME0/iRPC is not correct/updated. Also, unused codes related to converting track trigger stubs into muon trigger primitives have been removed.

Notes:

- The definition of `Subsector()` for EMTFHit converted from RPC hit has changed. It is now consistent with the CSC definition (1 or 2 in ME1, 0 in ME2,3,4). The old definition is no longer stored, but can be recovered via `Subsector_RPC()`.
- GEM input is now `GEMPadDigiClusterCollection`.
- ME0 input is now `ME0TriggerDigiCollection`.
- Regarding input: access to `simTwinMuxDigis`, `simDtTriggerPrimitiveDigis`, `rpcRecHits`, `me0TriggerConvertedPseudoDigis` has been added, but none of them are turned on by default.
- Regarding output: no change. While the interface has been implemented, the rest of the track finder logic has not been updated to use the new trigger primitives.

Additional notes:
- No dependence on other PRs or externals.
  - This PR does not interfere with the recent EMTF PRs by Andrew that are related to UL2016 emulation bugs.
- The plan for EMTF emulator update is discussed [here]( https://indico.cern.ch/event/890597/contributions/3756065/subcontributions/300461/attachments/1991401/3321960/2020-02-20_emtf_emulator_v2.pdf).

#### PR validation:

From `runTheMatrix.py -l limited -i all -j10`:

```
18 17 16 10 4 1 1 1 1 tests passed, 16 0 0 0 0 0 0 0 0 failed
(all the errors are due to DAS ERROR - not sure what's wrong?)
```

From `runTheMatrix.py -w upgrade -l 20434.0,20504.0 -j10`:

```
2 2 2 2 tests passed, 0 0 0 0 failed
```

Considering this is a big PR, there might be bugs. But we will try to detect and fix them as soon as possible.

---
Notifications: @abrinke1 @eyigitba @rekovic


silviodonato:
please test

jfernan2:
@jiafulow there are changes in wf 1330.o related to L1T, can you please check if they are expected?
I indeed wonder why only this 1330.0 MC dataset shows differences.
https://cmssdt.cern.ch/SDT/jenkins-artifacts/baseLineComparisons/CMSSW_11_1_X_2020-03-22-2300+075404/35583/dqm-histo-comparison-summary.html

jiafulow:
@jfernan2 it seems one EMTF muon from RelValZMM_13 (MC, Run2_2016 era) has changed its pT value by 5.5 GeV (or 11 unit in HW pT)

ugmtMuonPt
ugmtMuonhwPt

And this has a cascading effect on other 1330.0 DQM plots: ugmtMuonPhiAtVtx, ugmtMuonhwPhiAtVtx, ugmtMuMuInvMass, ugmtMuMuInvMassAtVtx; as well as all the 2D plots that depend on one of these quantities.

The expectation is that all the output remain the same. I'm wondering if this could be due to some numerical instability? Given all the other DQM plots show perfect agreement, I'm not entirely sure how important is this single event? But if it's really important, I can try to print out all these events to check.

jfernan2:
Thanks @jiafulow !! As long as it is understood, I am fine with it. I let the Muon experts to understand if this is relevant for the emulator. I am glad to hear that DQM is able to spot such a tiny effects.

jfernan2:
+1

dildick:
@jiafulow for the GEM input, you'll want to use "simMuonGEMPadDigiClusters" which is of type GEMPadDigiCluster instead of "simMuonGEMPadDigis" which is of type GEMPadDigi.

jiafulow:
@dildick Yes, you are right. I fixed it. Thanks!

jfernan2:
@jiafulow it seems this mutation from this muon is not related to your PR but something happening around in all PRs, see for example:
https://github.com/cms-sw/cmssw/pull/29275
Workflows 1330.0 and 25202

I wonder if you or somebody else from L1T could have a look...

> 
> 
> @jfernan2 it seems one EMTF muon from RelValZMM_13 (MC, Run2_2016 era) has changed its pT value by 5.5 GeV (or 11 unit in HW pT)
> 
> ugmtMuonPt
> ugmtMuonhwPt
> 
> And this has a cascading effect on other 1330.0 DQM plots: ugmtMuonPhiAtVtx, ugmtMuonhwPhiAtVtx, ugmtMuMuInvMass, ugmtMuMuInvMassAtVtx; as well as all the 2D plots that depend on one of these quantities.
> 
> The expectation is that all the output remain the same. I'm wondering if this could be due to some numerical instability? Given all the other DQM plots show perfect agreement, I'm not entirely sure how important is this single event? But if it's really important, I can try to print out all these events to check.

jfernan2:
please test

jfernan2:
@jiafulow  Now there seems to be other mutations in the output of 1330, if you could please have a look and make sure they are decoupled to this PR, it is highly appreciated. Thanks.
https://cmssdt.cern.ch/SDT/jenkins-artifacts/baseLineComparisons/CMSSW_11_1_X_2020-03-23-2300+075404/35608/dqm-histo-comparison-summary.html

jiafulow:
@jfernan2 Very strange indeed. My suspicion is that it's due to a recent PR related to EMTF that is supposed to fix the MC samples with era Run2_2016. Both 1330.0 and 25202.0 are MC Run2_2016. I'll take a look and try to figure out what's going on.

jfernan2:
Thanks @jiafulow !

silviodonato:
yes, we still have fake and random errors both in wf 25202.0 and 1330.0
https://github.com/cms-sw/cmssw/issues/29237

silviodonato:
Do you have further comments?
l1:  @benkrikler @rekovic

jfernan2:
I was waiting for further investigations from @jiafulow

silviodonato:
kind reminder @benkrikler @rekovic

silviodonato:
> I was waiting for further investigations from @jiafulow

I thought the error was due to https://github.com/cms-sw/cmssw/issues/29237 , I let @jiafulow  to confirm this

jiafulow:
I'm very sorry for the delay, I'll look into it this week.

jiafulow:
I have identified the issue, and posted the info on https://github.com/cms-sw/cmssw/issues/29237 .

Technically, it has nothing to do with this PR, but other recent PRs related to EMTF.

jfernan2:
+1

rekovic:
to discuss with @jiafulow

rekovic:
+1

silviodonato:
assign upgrade

rekovic:
Hi @silviodonato @kpedro88 
This PR is Phase-1 oriented as it concerns the current Phase-1 EMTF emulator.  It is not the Phase-2 EMTF hardware emulator which will eventually come as a separate package.  This is the adjustment to the current Phase-1 emulator to be able to read input TPs from some Phase-2 detectors, one of them being GEMs which are going to be used in Run-3. This is only a preparation step needed to adjust the current emulator be able to consume GEMs for Run-3.  It will remain backward compatible to Run-2 data and MC.

kpedro88:
@rekovic that doesn't seem quite in line with the PR description, which mentions Phase 2 and TDR studies several times

rekovic:
@kpedro88 
This package `L1Trigger/L1TMuonEndCap` will remain there and `L1TMuonEndCapTrackProducer`is to serve for emulation of the Muon Track Finder algorithm running in Phase-1 hardware for Run2.  We will use the same hardware in Run3 and this  producer will be adjusted for the consumption of GEMs, but will however use the same type of pattern recognition algorithm.

The new hardware used in Phase-2 will support a different algorithm which will be emulated by a different code (not yet ready for integration).  I would like to keep the two separate and avoid possibilities of disrupting development of Phase-2 or endangering backward compatibilities.

What is written in this PR description only has to do with the interface, namely ability to read the TPs.
Till the Phase-2 algorithm is emulation is integrated, the Phase-1 producer of standalone muon-tracks will be used for building of Phase2 correlated object (TrackerTrack + Muon). 

This was discussed with @jiafulow.

jiafulow:
@kpedro88 
I'm used to call these "Phase-2" trigger primitives, although GE1/1 has been installed during this shutdown and GE2/1 will be installed before Phase-2. So, the GEM TPs are more like Phase-1.5. In any case, as Vladimir explained, this PR allows the EMTF to receive the new TPs (GEM, iRPC, ME0), but the rest of the algorithm has not been updated to use them yet. That will happen in the future PRs, and development for the Phase-2 algorithm will be kept separated from `L1Trigger/L1TMuonEndCap` to avoid confusion.

L1Trigger/L1TMuonEndCap/src/PrimitiveSelection.cc
```diff
@@ -380,7 +421,89 @@ void PrimitiveSelection::process(GEMTag tag,
...
+    if (selected_me0 >= 0) {
+      assert(selected_me0 < NUM_GEM_CHAMBERS);
+      selected_me0_map[selected_me0].push_back(*tp_it);
+    }
+  }
+
+  // Apply truncation
+  bool apply_truncation = true;
+  if (apply_truncation) {
+    std::map<int, TriggerPrimitiveCollection>::iterator map_tp_it = selected_me0_map.begin();
```
kpedro88, line 444: can use `auto` here

L1Trigger/L1TMuonEndCap/src/PrimitiveSelection.cc
```diff
@@ -380,7 +421,89 @@ void PrimitiveSelection::process(GEMTag tag,
...
+    std::map<int, TriggerPrimitiveCollection>::iterator map_tp_end = selected_me0_map.end();
+
+    for (; map_tp_it != map_tp_end; ++map_tp_it) {
+      //int selected = map_tp_it->first;
```
kpedro88, line 448: delete commented-out code

L1Trigger/L1TMuonEndCap/src/PrimitiveSelection.cc
```diff
@@ -380,7 +421,89 @@ void PrimitiveSelection::process(GEMTag tag,
...
+      TriggerPrimitiveCollection& tmp_primitives = map_tp_it->second;  // pass by reference
+
+      // Keep the first 20 clusters
+      if (tmp_primitives.size() > 20)
```
kpedro88, line 452: magic number should be named constant

L1Trigger/L1TMuonEndCap/src/PrimitiveSelection.cc
```diff
@@ -380,7 +421,89 @@ void PrimitiveSelection::process(GEMTag tag,
...
+    }
+  }
+
+  // Duplicate DT muon primitives
+  if (duplicateTheta_) {
+    std::map<int, TriggerPrimitiveCollection>::iterator map_tp_it = selected_dt_map.begin();
+    std::map<int, TriggerPrimitiveCollection>::iterator map_tp_end = selected_dt_map.end();
+
+    for (; map_tp_it != map_tp_end; ++map_tp_it) {
+      //int selected = map_tp_it->first;
```
kpedro88, line 482: delete commented-out code

L1Trigger/L1TMuonEndCap/src/PrimitiveSelection.cc
```diff
@@ -524,85 +657,40 @@ int PrimitiveSelection::select_csc(const TriggerPrimitive& muon_primitive) const
...
+    if (endcap_ == 1 && sector_ == 1 && bx_ == -3) {  // do assertion checks only once
+      assert_no_abort(emtf::MIN_ENDCAP <= tp_endcap && tp_endcap <= emtf::MAX_ENDCAP);
+      assert_no_abort(emtf::MIN_TRIGSECTOR <= tp_sector && tp_sector <= emtf::MAX_TRIGSECTOR);
+      assert_no_abort(1 <= tp_station && tp_station <= 4);
+      assert_no_abort(1 <= tp_csc_ID && tp_csc_ID <= 9);
+      assert_no_abort(tp_data.strip < max_strip);
+      assert_no_abort(tp_data.keywire < max_wire);
+      assert_no_abort(tp_data.valid == true);
+      assert_no_abort(tp_data.pattern <= 10);
+      //assert_no_abort(tp_data.quality > 0);
```
kpedro88, line 673: delete commented-out code

L1Trigger/L1TMuonEndCap/src/PrimitiveSelection.cc
```diff
@@ -683,85 +785,87 @@ int PrimitiveSelection::select_rpc(const TriggerPrimitive& muon_primitive) const
...
+
+    if (endcap_ == 1 && sector_ == 1 && bx_ == -3) {  // do assertion checks only once
+      assert_no_abort(tp_region != 0);
+      assert_no_abort(emtf::MIN_ENDCAP <= tp_endcap && tp_endcap <= emtf::MAX_ENDCAP);
+      assert_no_abort(emtf::MIN_TRIGSECTOR <= tp_sector && tp_sector <= emtf::MAX_TRIGSECTOR);
+      assert_no_abort(1 <= tp_subsector && tp_subsector <= 6);
+      assert_no_abort(1 <= tp_station && tp_station <= 4);
+      assert_no_abort((!is_irpc && 2 <= tp_ring && tp_ring <= 3) || (is_irpc && tp_ring == 1));
+      assert_no_abort((!is_irpc && 1 <= tp_roll && tp_roll <= 3) || (is_irpc && 1 <= tp_roll && tp_roll <= 5));
+      //assert_no_abort((!is_irpc && (tp_CPPF || (1 <= tp_strip && tp_strip <= 32))) || (is_irpc && 1 <= tp_strip && tp_strip <= 96));
```
kpedro88, line 798: delete commented-out code

L1Trigger/L1TMuonEndCap/src/PrimitiveSelection.cc
```diff
@@ -813,127 +928,318 @@ int PrimitiveSelection::select_gem(const TriggerPrimitive& muon_primitive) const
...
+
+  if (muon_primitive.subsystem() == TriggerPrimitive::kME0) {
+    const ME0DetId& tp_detId = muon_primitive.detId<ME0DetId>();
+    const ME0Data& tp_data = muon_primitive.getME0Data();
+
+    int tp_region = tp_detId.region();  // 0 for Barrel, +/-1 for +/- Endcap
+    int tp_endcap = (tp_region == -1) ? 2 : tp_region;
+    int tp_station = tp_detId.station();
+    int tp_ring = 1;  // tp_detId.ring() does not exist
+    //int tp_roll      = tp_detId.roll();
```
kpedro88, line 1025: delete commented-out code

L1Trigger/L1TMuonEndCap/src/PrimitiveSelection.cc
```diff
@@ -813,127 +928,318 @@ int PrimitiveSelection::select_gem(const TriggerPrimitive& muon_primitive) const
...
+  // This means that in sector 1, CSC chambers cover 15 to 75 deg, but ME0
+  // chambers cover 10 to 70 deg. 5 deg (1/4 of chamber) needs to be added
+  // to cover 70 to 75 deg, and 5 deg needs to be subtracted from 10 to 15 deg.
+  auto get_other_neighbor = [](int sector) { return (sector == 6) ? 1 : sector + 1; };
+
+  bool add5deg = false;
+  bool sub5deg = false;
+  if (includeNeighbor_) {
+    if ((endcap_ == tp_endcap) && (get_other_neighbor(sector_) == tp_sector)) {
+      if (tp_csc_ID == 1 && tp_endcap == 1 && tp_pad >= (767 - 192)) {  // higher 1/4 of chamber
```
kpedro88, line 1071: magic numbers should be named constants

L1Trigger/L1TMuonEndCap/src/PrimitiveSelection.cc
```diff
@@ -813,127 +928,318 @@ int PrimitiveSelection::select_gem(const TriggerPrimitive& muon_primitive) const
...
+    if (tp_station == 4) {
+      if (tp_sector == 13)
+        tp_sector = 4;
+      else if (tp_sector == 14)
+        tp_sector = 10;
     }
+
+    int tp_bx = tp_data.bx;
+    int tp_phi = tp_data.radialAngle;
+    //int tp_phiB      = tp_data.bendingAngle;
```
kpedro88, line 1175: delete commented-out code

L1Trigger/L1TMuonEndCap/src/TrackFinder.cc
```diff
@@ -156,14 +169,30 @@ void TrackFinder::process(const edm::Event& iEvent,
...
-    collector.extractPrimitives(CSCTag(), iEvent, tokenCSC_, muon_primitives);
-  if (useRPC_ && useCPPF_)
-    collector.extractPrimitives(CPPFTag(), iEvent, tokenCPPF_, muon_primitives);
-  else if (useRPC_)
-    collector.extractPrimitives(RPCTag(), iEvent, tokenRPC_, muon_primitives);
-  if (useGEM_)
-    collector.extractPrimitives(GEMTag(), iEvent, tokenGEM_, muon_primitives);
+  if (useCSC_) {
+    collector.extractPrimitives(emtf::CSCTag(), &geometry_translator_, iEvent, tokenCSC_, muon_primitives);
+    //collector.extractPrimitives(emtf::CSCTag(), &geometry_translator_, iEvent, tokenCSC_, tokenCSCComparator_, muon_primitives);
```
kpedro88, line 174: delete commented-out code

L1Trigger/L1TMuonEndCap/src/TrackTools.cc
```diff
@@ -80,4 +80,90 @@ namespace emtf {
     }
   }
 
+  // Use CSC trigger sector definitions
+  // Copied from DataFormats/MuonDetId/src/CSCDetId.cc
+  int get_trigger_sector(int ring, int station, int chamber) {
+    int result = 0;
+    if (station > 1 && ring > 1) {
+      result = ((static_cast<unsigned>(chamber - 3) & 0x7f) / 6) + 1;  // ch 3-8->1, 9-14->2, ... 1,2 -> 6
```
kpedro88, line 88: magic numbers should be named constants

kpedro88:
@rekovic @jiafulow some of the code rule/style issues in my review seem preexisting/beyond the scope of this PR, so they can be addressed in a subsequent PR. However, I identified at least one bug and some other potential issues with how the PR will work in production, so please take a look.

L1Trigger/L1TMuonEndCap/src/PrimitiveSelection.cc
```diff
@@ -203,12 +203,24 @@ void PrimitiveSelection::process(RPCTag tag,
...
+        // Skip cluster size cut if primitives are from CPPF emulator or EMTF unpacker (already clustered)
+        if (x.getRPCData().isCPPF)
+          return false;
+
         int sz = x.getRPCData().strip_hi - x.getRPCData().strip_low + 1;
+
+        const RPCDetId& tp_detId = x.detId<RPCDetId>();
+        int tp_station = tp_detId.station();
+        int tp_ring = tp_detId.ring();
+        const bool is_irpc = (tp_station == 3 || tp_station == 4) && (tp_ring == 1);
```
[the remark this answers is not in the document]
rekovic: There probably not much point here in introducing named constants to represent the four muon stations 1, 2, 3, 4, or muon rings.  In this case it should be clear from the code.

rekovic:
Thanks @kpedro88 .
Commented-out code can be removed indeed even now. 

Thanks for the catch of a wrong size array
`  const char subsystem_names[][4] = {"DT", "CSC", "RPC", "GEM", "ME0"};`

Some magic numbers should indeed be turned into named constants. In some cases I would say there might not be much point in doing so.

kpedro88:
@rekovic there are several reasons to avoid magic numbers in almost all circumstances. using named constants not only makes the code more readable and understandable for future maintainers by labeling numbers by their meaning, but also it enforces consistency. if a number is changed in one place, should the same number be changed somewhere else? how do i know if that 2 is the same as this other 2? this is *always* a future problem, not a present one, which is why the policy should always be followed whether or not a problem is "foreseen".

jiafulow:
Regarding `const char subsystem_names[][4] = {"DT", "CSC", "RPC", "GEM", "ME0"};`
In this case, the compiler expands `[][4]` to `[5][4]`, which is the necessary size for 5 elements that have maximum 3 characters plus 1 `\0' character. It is absolutely correct.

I have no idea that commented out codes are against the CMS rules. Sometimes, codes are commented out for now, but might need to be uncommented in the future. Sometimes codes are commented out to leave as a trace, so that if bugs are found in the future, it's easier to debug by checking what the previous version was. Deleting them just makes life harder for the few developers who have to work with the codes. The codes are almost always work in progress, rather than a finished product. It sounds crazy to me that this is instituted as a rule.

Regarding the magic numbers, most of the cryptic ones are already explained by the comments next to them. The trivial ones are clear from the context. There's no reason to assume any future human maintainers wouldn't be able to figure that out.

abrinke1:
Hi @kpedro88 ,

I'm all for clarity, but surely when the "magic" number is just the ID number, you don't increase the clarity by creating a new named constant?  For example, in the code above "tp_csc_ID == 1" clearly indicates that the CSC ID must equal 1.  This would not be made any clearer by defining some new constant:
int CSC_ID_1 = 1;
and then changing the condition to be:
tp_csc_ID == CSC_ID_1;
In fact, it would make the code more confusing.  There's no "translation" of CSC ID = 1; the CSC chamber with ID = 1 is just the CSC chamber with ID = 1.  I suppose you could add a geometrical description ("the chamber in the bottom left of each sector"), but that would be incredibly tedious.

Cheers,
@abrinke1

kpedro88:
@jiafulow 

> I have no idea that commented out codes are against the CMS rules. Sometimes, codes are commented out for now, but might need to be uncommented in the future. Sometimes codes are commented out to leave as a trace, so that if bugs are found in the future, it's easier to debug by checking what the previous version was. Deleting them just makes life harder for the few developers who have to work with the codes. The codes are almost always work in progress, rather than a finished product. It sounds crazy to me that this is instituted as a rule.

If there is a particular reason that a commented-out line of code should be kept, this should be indicated in a separate comment above that line of code. Typically, this is a temporary situation, and therefore the conditions for uncommenting that line of code should be clearly stated.

Otherwise, if a line of code is no longer needed, it is still preserved in the git history (and previous releases) for anyone who needs to look back at it.

CMSSW has 100s of contributors, with people frequently entering and leaving the collaboration. Leaving a file littered with commented-out lines of code makes it significantly harder to tell what the code is supposed to do. Once code is in the official repository, maintaining it is everyone's responsibility, not just the original developer's. Therefore, a higher standard must be met.

@jiafulow and @abrinke1:
> Regarding the magic numbers, most of the cryptic ones are already explained by the comments next to them. The trivial ones are clear from the context. There's no reason to assume any future human maintainers wouldn't be able to figure that out.

Making human maintainers "figure things out" that computers can do automatically is bad practice.

> In fact, it would make the code more confusing. There's no "translation" of CSC ID = 1; the CSC chamber with ID = 1 is just the CSC chamber with ID = 1. I suppose you could add a geometrical description ("the chamber in the bottom left of each sector"), but that would be incredibly tedious.

What if, for some reason, a future muon subdetector geometry swaps what is currently considered chamber 1 with chamber 2? Do you want to go through thousands of lines of code and figure out which "1"s need to be swapped with which "2"s? This is a somewhat contrived example, but there are plenty of non-contrived examples just in this PR.

abrinke1:
"What if, for some reason, a future muon subdetector geometry swaps what is currently considered chamber 1 with chamber 2? Do you want to go through thousands of lines of code and figure out which "1"s need to be swapped with which "2"s?"

My answer would be: "absolutely, yes."  Because there is no central CMSSW object linking physical chambers to their numerical IDs, and if we decided to create one now, we'd have to go through *all* the existing L1T and CSC code and update it to use this chamber-map object.  If we don't create such an object, then any change to the mapping means we have to go through all the code and find all of the new constants we created and change them by hand anyway.  It saves 0 time, and adds 0 clarity compared to just using the ID number.

silviodonato:
In https://cms-sw.github.io/cms_coding_rules.html you can find the "CMSSW rules".
About magic numbers, it is a generic C++ code style rules. In https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#enum1-prefer-enumerations-over-macros 
you can find a trivial example
```
constexpr int first_month = 1;
constexpr int last_month = 12;
for (int m = first_month; m <= last_month; ++m) 
    cout << month[m] << '\n';
```
is more understandable than
```
for (int m = 1; m <= 12; ++m)
    cout << month[m] << '\n';
```
The code has exactly the same performance (last_month is replaced with `12` at compile time).
Therefore, we suggest to use `constexpr` constants instead of numbers and comments.
Clearly, this is not a huge problem, but we should try to improve the code quality of CMSSW. 

About 
```
const bool is_irpc = (tp_station == 3 || tp_station == 4) && (tp_ring == 1);
```
I think it is rather clear, but the following code is more understandable for people who are not L1T experts
```
constexpr auto RE3_1 =  std::pair<int, int>(3, 1) //station, ring
constexpr auto RE4_1 =  std::pair<int, int>(4, 1) //station, ring
[...]
const auto module = std::pair<int, int>(tp_station, tp_ring)
const bool is_irpc = ( module == RE3_1 ) ||( module == RE4_1 )
```
Anyhow, you can keep the 1-line version, if you strongly prefer it.

About
```
result = ((static_cast<unsigned>(chamber - 3) & 0x7f) / 6) + 1;  // ch 3-8->1, 9-14->2, ... 1,2 -> 6
```
you definitely need to use `constexpr` variable for `0x7f`and all other numbers.

Regarding the commented code, I confirm the policy of deleting commented lines.
We can retrieve any version of the old code from GitHub. 
Commented line should be used only for temporary changes.
You are right, this rule does not appear in "CMS Naming, Coding, And Style Rules". We are going to update them.

jiafulow:
Let me be clear. No one in our group who is responsible for developing/maintaining this particular package would have trouble understanding

```
for (int m = 1; m <= 12; ++m)
    cout << month[m] << '\n';
```

or

```
const bool is_irpc = (tp_station == 3 || tp_station == 4) && (tp_ring == 1);
```

The hypothetical scenario where a maintainer would struggle to understand them doesn't exist. Not now, not in the future. 

I'm sorry I simply don't agree that the following code is better:

```
constexpr auto RE3_1 =  std::pair<int, int>(3, 1) //station, ring
constexpr auto RE4_1 =  std::pair<int, int>(4, 1) //station, ring
[...]
const auto module = std::pair<int, int>(tp_station, tp_ring)
const bool is_irpc = ( module == RE3_1 ) ||( module == RE4_1 )
```

It's introducing throwaway variables, and it's harder to read than a one-liner. Everyone has different opinions, I've stated my opinions, and I'm not going to argue further on what is 'better' or higher code quality. In the end, our group is the one that takes responsibilities of the output of this particular package. As the developer, I have my own opinions and I am not going to be force-fed bullshit opinions. My job is to submit a PR that does what the L1 group asked of us. If the PR is not doing what is described, I'm happy to fix it. But I'm not entertaining pointless code review comments.

Regarding `result = ((static_cast<unsigned>(chamber - 3) & 0x7f) / 6) + 1;  // ch 3-8->1, 9-14->2, ... 1,2 -> 6`
it was taken from https://github.com/cms-sw/cmssw/blob/master/DataFormats/MuonDetId/src/CSCDetId.cc#L11-L20
If you are really unhappy with the magic numbers, please fix that piece of code and then I'll copy from the updated code.

kpedro88:
> But I'm not entertaining pointless code review comments.

Please note that no developer or group is entitled to have their code included in the official repository. Your group may "take responsibility" for the code in an informal way, but once it is included in the official repository, it is everyone's responsibility - if it breaks, it must be fixed.

I deliberately included this statement in my PR review to indicate the relative importance of comments: "some of the code rule/style issues in my review seem preexisting/beyond the scope of this PR, so they can be addressed in a subsequent PR." But I am also happy to reject the PR outright if the developers feel they don't need to follow the rules at all.

jiafulow:
>  it is everyone's responsibility - if it breaks, it must be fixed.

Except that every time the EMTF emulator breaks, it was up to us to fix it.

If you feel you have the power over all the codes that we contributed to, and do not want to address any concerns that your rules are bad and sometimes even pointless, then go ahead and do what you want to do. There's no reason for me to fix anything that is not broken.

kpedro88:
-1
developers are belligerent and do not understand how software management works in a large collaboration
this behavior should not be rewarded

rekovic:
Coming back to this thread.

@kpedro88 
> @rekovic there are several reasons to avoid magic numbers in almost all circumstances. using named constants not only makes the code more readable and understandable for future maintainers by labeling numbers by their meaning, but also it enforces consistency. if a number is changed in one place, should the same number be changed somewhere else? how do i know if that 2 is the same as this other 2? this is always a future problem, not a present one, which is why the policy should always be followed whether or not a problem is "foreseen".

I agree. For the sake of consistency using constants is preferred.  In some cases this could severely jeopardizes the readability, but in the long run it pays off.

rekovic:
Hi @jiafulow 
The requests of @kpedro88 should be addressed.  
- if indeed needed to keep commented out lines perhaps add description.  This all can be cleaned up in the subsequent PR's part of this series when the debugging is finished. 
- where possible please use named constants

silviodonato:
This PR  has conflicts, it needs to be rebased.
CMSSW_11_1_0_pre7 will be close tomorrow.

silviodonato:
moved to  #29767

PR #29269, opened 2020-03-22
Run3-sim61 dd4hep version of the forward algorithms

bsunanda:
#### PR description:

Provide dd4hep versions of the forward algorithms

#### PR validation:

Tested with cfg files within Geometry/ForwardCommonData

Nothing special

### DIFF
--- a/Geometry/ForwardCommonData/plugins/dd4hep/DDBHMAngular.cc
+++ b/Geometry/ForwardCommonData/plugins/dd4hep/DDBHMAngular.cc
@@ -62,8 +62,8 @@ static long algorithm(dd4hep::Detector& /* description */,
 
     parent.placeVolume(child, jj + 1, dd4hep::Transform3D(rot, tran));
 #ifdef EDM_ML_DEBUG
-    edm::LogVerbatim("ForwardGeom") << "DDBHMAngular test: " << child.name() << " number " << jj + 1 << " positioned in "
-                                    << parent.name() << " at " << tran << " with " << rot;
+    edm::LogVerbatim("ForwardGeom") << "DDBHMAngular test: " << child.name() << " number " << jj + 1
+                                    << " positioned in " << parent.name() << " at " << tran << " with " << rot;
 #endif
   }
   return 1;
--- a/Geometry/ForwardCommonData/plugins/dd4hep/DDBHMAngular.cc
+++ b/Geometry/ForwardCommonData/plugins/dd4hep/DDBHMAngular.cc
@@ -4,7 +4,7 @@
 #include "FWCore/MessageLogger/interface/MessageLogger.h"
 #include "DD4hep/DetFactoryHelper.h"
 
-#define EDM_ML_DEBUG
+//#define EDM_ML_DEBUG
 using namespace geant_units::operators;
 
 static long algorithm(dd4hep::Detector& /* description */,
--- a/Geometry/ForwardCommonData/plugins/dd4hep/DDBHMAngular.cc
+++ b/Geometry/ForwardCommonData/plugins/dd4hep/DDBHMAngular.cc
@@ -34,25 +34,24 @@ static long algorithm(dd4hep::Detector& /* description */,
   if (rotstr != "NULL") {
     rot = ns.rotation(rotMat);
   }
+  static const int startpos = 16;
   static const double fac1 = 0.5;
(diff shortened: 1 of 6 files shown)


bsunanda:
@cmsbuild Please test

bsunanda:
@cmsbuild Please test

Geometry/ForwardCommonData/plugins/dd4hep/DDBHMAngular.cc
```diff
@@ -0,0 +1,73 @@
+#include "DataFormats/Math/interface/GeantUnits.h"
+#include "DetectorDescription/Core/interface/DDSplit.h"
+#include "DetectorDescription/DDCMS/interface/DDPlugins.h"
+#include "FWCore/MessageLogger/interface/MessageLogger.h"
+#include "DD4hep/DetFactoryHelper.h"
+
+#define EDM_ML_DEBUG
```
cvuosalo:
```suggestion
// #define EDM_ML_DEBUG
```

Geometry/ForwardCommonData/plugins/dd4hep/DDBHMAngular.cc
```diff
@@ -0,0 +1,73 @@
...
+      driverX = rr * cos(0.5 * dphi);
+    } else if (jj == 17) {
+      driverX = rr * cos(14.5 * dphi);
+      driverY = -sqrt(rr * rr - driverX * driverX);
+    } else if (jj == 18) {
+      driverX = rr * cos(0.5 * dphi);
+      driverY = -sqrt(rr * rr - driverX * driverX);
+    } else if (jj == 19) {
+      driverX = rr * cos(1.5 * dphi);
+      driverY = -sqrt(rr * rr - driverX * driverX);
```
cvuosalo: Several magic numbers here that should be replaced with constants: 16, 15.5, 14.5, 1.5.

Geometry/ForwardCommonData/plugins/dd4hep/DDTotemAngular.cc
```diff
@@ -0,0 +1,62 @@
...
+  edm::LogVerbatim("ForwardGeom") << "DDTotemAngular debug: Parent " << parent.name() << "\tChild " << child.name();
+#endif
+
+  double phi = startAngle;
+  int copyNo = startCopyNo;
+
+  for (int ii = 0; ii < n; ii++) {
+    double phitmp = phi;
+    if (phitmp >= 2._pi)
+      phitmp -= 2._pi;
```
cvuosalo: We have a class for this purpose:
```
#include "DataFormats/GeometryVector/interface/Phi.h"
Geom::Phi0To2pi<double> phitmp = phi;
```
`phitmp` would automatically be kept in the range 0-2pi, and the case of <-2pi is taken care of, which line 41 misses.

bsunanda:
@cmsbuild Please test

Geometry/ForwardCommonData/plugins/dd4hep/DDBHMAngular.cc
```diff
@@ -0,0 +1,71 @@
...
+  if (rotstr != "NULL") {
+    rot = ns.rotation(rotMat);
+  }
+  static const double fac1 = 0.5;
+  static const double fac2 = 1.5;
+  static const double fac3 = 14.5;
+  static const double fac4 = 15.5;
+  for (int jj = 0; jj < units; jj++) {
+    double driverX(0), driverY(0), driverZ(0);
+    if (jj < 16) {
```
cvuosalo: It would be nice to replace the 16 with a constant. I'm not sure what the 16 means. Then line 49 would be something like `else if (jj == startPos + 1)` if `startPos` was the name of the constant. Then lines 52 and 55 would be similar with +2 and +3.

Geometry/ForwardCommonData/plugins/dd4hep/DDBHMAngular.cc
cvuosalo: `fac1`, `fac2`, etc. are not very illuminating about what they are. I suppose we don't know what these values mean: 14.5, 15.5?

bsunanda:
@cmsbuild Please test

cvuosalo:
+1

silviodonato:
+1

PR #29276, opened 2020-03-23
keeping MET sumPtUnclustered (backport of #28370)

danbarto:
#### PR description:

The sum of the unclustered energy is one of the inputs to MET significance, but is currently not stored in miniAOD, making recomputation of the significance and the MET covariance matrix difficult. With this PR, the calculated sum of the pTs of the unclustered particles is stored, and can be accessed with metSumPtUnclustered(), similar to metSignificance().

#### PR validation:

All tests have been run, and were passed.

backport of #28370, needed for nanoAODv7

### DIFF
--- a/DataFormats/PatCandidates/interface/MET.h
+++ b/DataFormats/PatCandidates/interface/MET.h
@@ -75,10 +75,10 @@ namespace pat {
       void setMETSignificance(const double& metSig);
       // get the MET significance
       double metSignificance() const;
-      // set the MET metSumPtUnclustered for MET Significance
-      void setMETSumPtUnclustered(const double& metSumPtUnclustered);
-      // get the MET metSumPtUnclustered
-      double metSumPtUnclustered() const;
+      // set the MET sumPtUnclustered for MET Significance
+      void setMETSumPtUnclustered(const double& sumPtUnclustered);
+      // get the MET sumPtUnclustered
+      double sumPtUnclustered() const;
 
       // ---- methods for uncorrected MET ----
       // Methods not yet defined
@@ -254,8 +254,8 @@ namespace pat {
 
       // MET significance
       double metSig_;
-      // MET metSumPtUnclustered for MET Significance
-      double metSumPtUnclustered_;
+      // MET sumPtUnclustered for MET Significance
+      double sumPtUnclustered_;
       
       const PackedMETUncertainty findMETTotalShift(MET::METCorrectionLevel cor, MET::METUncertainty shift) const;
    
--- a/DataFormats/PatCandidates/interface/MET.h
+++ b/DataFormats/PatCandidates/interface/MET.h
@@ -78,7 +78,7 @@ namespace pat {
       // set the MET sumPtUnclustered for MET Significance
       void setMETSumPtUnclustered(const double& sumPtUnclustered);
       // get the MET sumPtUnclustered
-      double sumPtUnclustered() const;
+      double metSumPtUnclustered() const;
 
       // ---- methods for uncorrected MET ----
       // Methods not yet defined
--- a/DataFormats/PatCandidates/interface/MET.h
+++ b/DataFormats/PatCandidates/interface/MET.h
@@ -75,6 +75,10 @@ namespace pat {
       void setMETSignificance(const double& metSig);
       // get the MET significance
       double metSignificance() const;
+      // set the MET metSumPtUnclustered for MET Significance
+      void setMETSumPtUnclustered(const double& metSumPtUnclustered);
+      // get the MET metSumPtUnclustered
+      double metSumPtUnclustered() const;
 
       // ---- methods for uncorrected MET ----
       // Methods not yet defined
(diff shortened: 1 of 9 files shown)


slava77:
@cmsbuild please test

peruzzim:
thank you really much @danbarto for providing this backport

DataFormats/PatCandidates/src/classes_def_objects.xml
```diff
@@ -271,9 +271,10 @@
   <ioread sourceClass = "pat::Jet" version="[1-]" targetClass="pat::Jet" source="" target="daughtersTemp_">
     <![CDATA[daughtersTemp_.reset();]]>
   </ioread>
-  <class name="pat::MET"  ClassVersion="15">
-    <version ClassVersion="15" checksum="428901429"/>
+  <class name="pat::MET"  ClassVersion="16">
+    <version ClassVersion="16" checksum="2969659150"/>
```
perrotta: @Dr15Jones @slava77 (also for my own understanding): is it correct that the same ```ClassVersion``` (here "16") can get two different ```checksum```'s in different releases (here 10_2 vs 11_1)?

RecoMET/METAlgorithms/interface/METSignificance.h
```diff
@@ -44,7 +44,8 @@ namespace metsig {
 					  JME::JetResolution & resPtObj,
 					  JME::JetResolution & resPhiObj,
 					  JME::JetResolutionScaleFactor & resSFObj,
-					  bool isRealData);
+					  bool isRealData,
+            double& sumPtUnclustered);
```
perrotta: fix indentation

RecoMET/METAlgorithms/src/METSignificance.cc
```diff
@@ -54,7 +54,8 @@ metsig::METSignificance::getCovariance(const edm::View<reco::Jet>& jets,
 				       JME::JetResolution& resPtObj,
 				       JME::JetResolution& resPhiObj,
 				       JME::JetResolutionScaleFactor& resSFObj,
-				       bool isRealData) {
+				       bool isRealData,
+               double& sumPtUnclustered) {
```
perrotta: fix indentation

DataFormats/PatCandidates/interface/MET.h
```diff
@@ -250,6 +254,8 @@ namespace pat {
 
       // MET significance
       double metSig_;
+      // MET metSumPtUnclustered for MET Significance
+      double metSumPtUnclustered_;
```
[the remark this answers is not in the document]
danbarto: Sorry, my mistake. I'll fix that.

DataFormats/PatCandidates/src/classes_def_objects.xml
Dr15Jones: Once a checksum has been assigned a version number it should not be changed. The only exception is when it is deemed that version was never used for any useful jobs.

slava77: - changes of existing versions in a release are not allowed
- same versions in different releases with different checksums are not allowed

perrotta: Thank you @Dr15Jones and @slava77 

I assume then that, once ```metSumPtUnclustered_``` is renamed as ```sumPtUnclustered_```. as in the master, the very same structure in terms of class members and methods will be there for this backport and the master. Therefore the same version "16" and checksum as the master can be also backported.

slava77: if for whatever reason we have to increment a version number here, we should make sure to add it to the master as well

perrotta: > if for whatever reason we have to increment a version number here, we should make sure to add it to the master as well

Once class members and methods are sinchronized as suggested [above](https://github.com/cms-sw/cmssw/pull/29276#discussion_r397214919), the same class version can be considered for both releases.

The only difference in the code itself between the master and the 10_2 backport will come from the application of the clang format updates in the master, and not yet in 10_2.

That is not enough to trigger a version number change. (And even if so, then probably both
```
DataFormats/PatCandidates/interface/MET.h
DataFormats/PatCandidates/src/MET.cc
```
can be directly copied from the version in the master, and at that point they will be absolutely identical in the two CMSSW releases).

danbarto: After fixing the inconsistencies in MET.h/MET.cc the version is now consistent with the master. I'll update the PR soon.

perrotta:
@cmsbuild please test

perrotta:
backport of #28370

perrotta:
+1
- "double sumPtUnclustered" added to the MET DataFormat
- Jenkins tests pass and show no other differences than this additional double added to the dataformat
- Since only one additional double is added to the pat::MET, while all other class members stay untouched, I think this can fulfil  the requirements for a backport. The algorithm to read from it will need to be enabled with an era modifier.

perrotta:
+1
- "double sumPtUnclustered" added to the MET DataFormat
- Jenkins tests pass and show no other differences than this additional double added to the dataformat; changes observable in the output of automatic tests originate from #29294, which was run together in those tests
- Since only one additional double is added to the pat::MET, while all other class members stay untouched, I think this can fulfil the requirements for a backport. The algorithm to read from it will need to be enabled with an era modifier.

silviodonato:
merge
@santocch

santocch:
+1

PR #29268, opened 2020-03-22
Run3-sim60 Define the new Totem T2 for Run3

bsunanda:
#### PR description:

Define the geometry of the new Totem T2 detector defined for the Run3 period. The dd4hep version is in #29269

#### PR validation:

Tested with cfg files within Geometry/ForwardCommonData/test

#### if this PR is a backport please specify the original PR and why you need to backport that PR:

Nothing special

PR #29281, opened 2020-03-24
Run3-sim62 Add hits for the new Totem T2 detector

bsunanda:
#### PR description:

Add a new hit collection for the newly proposed Totem T2 detector for the Run3 application

#### PR validation:

Tested with runTheMatrix.py and the new hit collection is tested with #29268 and a cfg file in SimG4CMS/Forward

Nothing special


bsunanda:
@cmsbuild Please test

SimG4CMS/Forward/src/TotemT2ScintSD.cc
```diff
@@ -0,0 +1,84 @@
...
+#include "G4VProcess.hh"
+#include "G4ios.hh"
+#include "G4Cerenkov.hh"
+#include "G4ParticleTable.hh"
+#include "CLHEP/Units/GlobalSystemOfUnits.h"
+#include "CLHEP/Units/GlobalPhysicalConstants.h"
+#include "Randomize.hh"
+#include "G4Poisson.hh"
+
+#define EDM_ML_DEBUG
```
cvuosalo:
```suggestion
// #define EDM_ML_DEBUG
```

cvuosalo:
@bsunanda You might want to put "Totem" in the PR title to inform the uninitiated.

cvuosalo:
+1

civanch:
@bsunanda , for  me it is strange, that Geometry/ForwardSimData/data/totemsensT2/2021/v2/totemsensT2.xml is inside "2021/v2". Why not directly inside "2021"? An agreement as I understand it is that inside 2021 we have the baseline for the Run-3, when the geometry will be frozen each new update is inside "v1", "v2"... 

Second moment: it is not possible to overwrite cuts for the world volume here. If for a sub-detector we need a special set of cuts we create a new region. I would propose making Totem2 region for this purpose.

bsunanda:
There was a file called Geometry/ForwardSimData/data/totemsensT2/2021/v1/totemsensT2.xml - which was indeed a file for Run2 and I moved it to Geometry/ForwardSimData/data/totemsensT2/2018 now. This detector was used in Run2 and is decommissioned for Run3. I have defined a new cut region "TotemT2Region" for which cuts are defined

bsunanda:
@cmsbuild Please test

SimG4CMS/Forward/src/TotemT2ScintSD.cc
bsunanda: Done

cvuosalo:
+1

civanch:
+1

silviodonato:
+1

PR #29053, opened 2020-02-27
Remove remaining unneeded dependencies from Reco subsystems

guitargeek:
#### PR description:

Another quick round of automatic BuildFile.xml cleaning to improve the modularity of CMSSW.

Now, all of the Reco* Build files don't use dependencies anymore which are not included in the source files.

#### PR validation:

CMSSW compiles. It was checked that all newly added dependencies are actually required by the package with `git grep`.

PR #29295, opened 2020-03-25
Remove unneeded dependencies from plugins, test and bin in Reco subsystems

guitargeek:
#### PR description:

After https://github.com/cms-sw/cmssw/pull/29053 finished cleaning the BuildFiles in the libraries, this PR now covers all the other subdirectories like `bin`, `plugins` and `test`. This is a bit easier because no other build targets depend of these.

Little particularity: the `Geometry/CaloEventSetup` library had no BuildFile before because it was headers only, but now I added a `BuildFile.xml` to specify the dependencies of these header files, so they are correctly propagated. Otherwise, the build targets that require `Geometry/CaloEventSetup` have to make the corresponding `use` statements even if they don't use the libraries themselves. Furthermore, a few includes that were clearly out of place were removed.

#### PR validation:


perrotta:
please test

jfernan2:
+1
DQM Changes in 25202 due https://github.com/cms-sw/cmssw/issues/29237

cvuosalo:
+1

kpedro88:
+upgrade

Martin-Grunewald:
+1

santocch:
+1

civanch:
+1

pohsun:
+1

slava77:
@guitargeek 
do you have the methods/tools leading to this PR scripted somewhere?
It would be nice to pass these to the release QA

slava77:
+1

for https://github.com/cms-sw/cmssw/pull/29295	18803c7 
- technical
- jenkins tests pass and comparisons show no (relevant) differences

guitargeek:
Hi @slava77, thanks of signing off!

My script is here, I originally made it when I attempted to build parts of CMSSW on my laptop:
https://github.com/guitargeek/PKGBUILDs/blob/master/cmssw/clean_buildfiles.py
It's really nothing special, but it runs fast and does the job. It just removes `use` tags that don't have matches in any header file. It can happen that there are missing dependencies after the removal process, because some build targets got their dependencies indirectly via unnecessary use tags in other packages that the script removed. One has to add these back manually.

Of course one could also have a script that adds a `use` tag for every CMSSW package that is included in the packages code. On one hand, that makes the BuildFiles blow quite a bit and I don't know if that's desired. It's certainly harder to review. On the other hand, everything could be automatized then I guess.

guitargeek:
Sorry I forgot to add that even thought the script is reasonably fast to run for a few packages/subsystems at once, it's still very inefficient because it does a git grep for each `use` tag it encounters. That's actually terrible... Probably it's better to only loop over the code once and dump all the includes into a json file or something (like I do here: https://gist.github.com/guitargeek/ae0449726b5ac1a4bc05319138e9e652) before checking with the BuildFiles.

guitargeek:
Hi @silviodonato, can you please merge this? I think the `db-pending` tag is not really used anymore for at least half a year or something. And I'm pretty scared that this PR has to be rebased if we wait longer.

silviodonato:
merge
@guitargeek thanks for the reminder
@ggovi this is BuildFile cleaning

PR #29304, opened 2020-03-25
Remove unused SeedFilter class in EgammaElectronAlgos to simplify ElectronSeedProducer

guitargeek:
#### PR description:

Going over the Egamma reco code to work on the EGM paper I checked out once again the `ElectronSeedProducer`. This time I noticed that the option `preFilteredSeeds` is `false` and it apparently has been for the long time. In this situation, the producer does not use the `SeedFilter` class from the EgammaElectronAlgos library. I therefore suggest to remove the usage of this class and the class itself, so the `ElectronSeedProducer` can live with much less configuration parameters. Like this the plugin is easier to understand and I also hope that this way, it gets easier apparent in the future how this plugin can potentially be unified or replaced with the `ElectronNHitSeedProducer` which is used at the HLT.

Furthermore, I merged the `.h` and `.cc` files of some other producers in the electron reconstruction chain.

This PR does not collide with my two PRs from earlier today (https://github.com/cms-sw/cmssw/pull/29303, https://github.com/cms-sw/cmssw/pull/29295):

#### PR validation:

CMSSW compiles, matrix tests pass.

No backport intended.


slava77:
@cmsbuild please test

perrotta:
@afiqaize @SohamBhattacharya please explicitely confirm here that the removal proposed in this PR is  approved by the EGamma POG, and that the removed functionalities are not expected to have any possible usage in the future.

perrotta:
@guitargeek please revert the  inclusion of *Fwd.h headers in the corresponding complete DataFormats headers, as it was also already pointed out in #29375

jainshilpi:
> @afiqaize @SohamBhattacharya please explicitely confirm here that the removal proposed in this PR is approved by the EGamma POG, and that the removed functionalities are not expected to have any possible usage in the future.

As always, thanks Jonas. I talked to Sam about it and he mentioned we can even switch to the Nhit version of the seed producer completely and hence removing any dependency on ElectronSeedProducer class. May be it the right time to do it?

guitargeek:
Yep, indeed that would be the ultimate goal as I stated in the description: cleaning up the old seeding code like in this PR will help with the migration (no matter if it's done by me or someone else) because it gets easier to compare the seed producers. Good to know that also Sam was thinking of moving to the Nhit version for both reco and HLT.

guitargeek:
> @guitargeek please revert the inclusion of *Fwd.h headers in the corresponding complete DataFormats headers (and update the PR description accordingly), as it was also already pointed out in #29375

No worries, I did it. I will probably come back with the DataFormats changes in another PR because I'm still convinced it's the way to go :)

perrotta:
please test

perrotta:
> > @guitargeek please revert the inclusion of *Fwd.h headers in the corresponding complete DataFormats headers (and update the PR description accordingly), as it was also already pointed out in #29375
> 
> No worries, I did it. I will probably come back with the DataFormats changes in another PR because I'm still convinced it's the way to go :)

Thank you Jonas
Please also update the PR description accordingly

guitargeek:
ok, done

guitargeek:
Something seems to have gone wrong with the bot

slava77:
@cmsbuild please test

RecoEgamma/EgammaElectronAlgos/src/ElectronSeedGenerator.cc
```diff
@@ -148,8 +148,8 @@ ElectronSeedGenerator::ElectronSeedGenerator(const edm::ParameterSet &pset, cons
       // new B/F configurables
       deltaPhi2B_(pset.getParameter<double>("DeltaPhi2B")),
       deltaPhi2F_(pset.getParameter<double>("DeltaPhi2F")),
-      phiMin2B_(pset.getParameter<double>("PhiMin2B")),
-      phiMin2F_(pset.getParameter<double>("PhiMin2F")),
+      phiMin2B_(-pset.getParameter<double>("PhiMax2B")),
```
perrotta, line 151: Are these boundaries always supposed to be symmetric?
(Same question for the other parameter changes below)

RecoEgamma/EgammaElectronProducers/plugins/ElectronSeedProducer.cc
```diff
@@ -16,11 +16,10 @@
 //
 //
 
+#include "DataFormats/EgammaReco/interface/SuperCluster.h"
```
perrotta, line 19: SuperClusterFwd.h ?

guitargeek: Well we still don't agree about what to do in these kind of situations! Maybe we can wait to see what happens with https://github.com/cms-sw/cmssw/pull/29410 so we have a precedential case to refer to? I still don't think that forward headers should be included in `.cc` files because every forward header *should* be included in the corresponding full header, which necessarily also has to be included in the `.cc` file for the forward declarations and `typedef`s to make sense.

RecoEgamma/EgammaElectronAlgos/src/ElectronSeedGenerator.cc
guitargeek: I think there is really an extremely small chance that we will ever consider electrons and positrons to be anything more than charge-inverted in the CMS reconstruction, so I really don't get why it was at all possible to configure this in a non-symmetric way. I think this was a clear case of over-configurabilty. Maybe I can argue that this does not matter so much at this point anyway, because as @jainshilpi said, it it very likely that the `ElectronSeedProducer` will be hopefully replaced by the `ElectronNHitSeedProducer` in the near future anyway? That's why I don't have a strong preference, but leaving it as it is in the PR right now is of course easier.

RecoEgamma/EgammaElectronProducers/plugins/ElectronSeedProducer.cc
perrotta: As I already told you, I agree on the principle (provided we can get rid of the recursive inclusions that @Dr15Jones  pointed out in the previous discussion).

In this particular case, however, you have in this class some types (e.g. ```reco::SuperClusterCollection```) which were only defined in  SuperClusterFwd.h. Therefore, the header which defines them must be included here.

Again in this particular case the inclusion is granted anyhow because SuperClusterFwd.h is already included by (at least) RecoEgamma/EgammaElectronAlgos/interface/ElectronUtilities.h, RecoEgamma/EgammaElectronAlgos/interface/ElectronSeedGenerator.h, RecoEgamma/EgammaElectronAlgos/interface/ElectronHcalHelper.h. Still an explicit inclusion of whatever is needed here should be allowed based on the addition of relevant include's, unless I misunderstood your point.

RecoEgamma/EgammaElectronAlgos/src/ElectronSeedGenerator.cc
perrotta: It makes sense, just wanted to be sure there were not possible use cases for the contrary, since it was originally coded as such,
Taking also into account the intention of moving eventually to ElectronNHitSeedProducer, ok for moving towards the simpler version.

RecoEgamma/EgammaElectronProducers/plugins/ElectronSeedProducer.cc
perrotta: On the other hand, this is nothing new brough in by this PR, and there are several other such cases in the classes touched by this PR. Let postpone the discussion to #29410, then

perrotta:
+1
- Removal of the SeedFilter class (and other simplifications in the ElectronSeedProducer) are implemented as planned
- That removal is [endorsed](https://github.com/cms-sw/cmssw/pull/29304#issuecomment-610309039 )by the EGamma POG
- No differences are visible in jenkis outputs, as it should

guitargeek:
Hi @chayanit, I just deleted some commented out lines here that don't make sense anymore after changing the way the `ElectronSeedProducer` is configured, so it should be trivial to check and sign. Thanks!

chayanit:
+1

silviodonato:
+1

PR #29316, opened 2020-03-26
DQM: Python DQMIO IO library

schneiml:
#### PR description:
This PR adds a Python library that can be used to read DQMIO data from individual files or remote datasets. The focus is on reading from full datasets without reading them from disk entirely.

So far there are no usages, but this code might be useful for detector studies on the per-lumi saved data from the UL rereco, Machine Learning studies, and potentially an alternative DQMGUI backend.

#### PR validation:
I gave the code a bit of testing but there might still be problems. Now and then ROOT just dies; multi-threading is hard after all. 

Probably we should not merge that until some users confirm it is useful.

@gbenelli voiced some interest in this.

### DIFF
--- a/DQMServices/FwkIO/python/DQMIO.py
+++ b/DQMServices/FwkIO/python/DQMIO.py
@@ -344,11 +344,16 @@ class DQMIOReader:
     dataset read from DAS. Operations are internally multi-threaded.
     All state is kept in a SQLite database, caches are managed by `DQMIOFileIO`.
     """
-    def __init__(self, dbname="", nthreads=32, pooltype=ThreadPoolExecutor):
+    def __init__(self, dbname="", nthreads=32, pooltype=ThreadPoolExecutor, progress=True):
         self.db = sqlite3.connect(dbname)
         self.db.executescript(DBSCHEMA)
         self.pool = pooltype(nthreads)
         self.io = DQMIOFileIO()
+        if progress:
+            self.readentrymes = self.readentrymes_slow
+        else:
+            self.readentrymes = self.readentrymes_fast
+
           
     def importdatasets(self, datasetpattern):
         """
@@ -453,7 +458,9 @@ class DQMIOReader:
         entries = self.db.execute(query, (datasetname, run, lumi))
         return [IndexEntry(*row) for row in entries]
 
-    def readentrymes(self, entries, fullnames):
+    # there are two of these, one with progress indication, one without.
+    # one of them is selected in __init__.
+    def readentrymes_slow(self, entries, fullnames):
         items = []
         def append(_, thing): items.append(thing)
         tasks = [(None, self.pool.submit(self.io.locateme, e, fullname)) for e in entries for fullname in fullnames]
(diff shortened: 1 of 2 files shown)


silviodonato:
Shall we expect any update? Is this PR ready?

schneiml:
@silviodonato still waiting for any actual use cases. Though I have used it for a bit now and have a few more fixes, I will apply those next week and we can maybe continue to merge it.

schneiml:
+1

I added other changes I had locally.
Just ran the example on a recent IB and it worked flawlessly.

schneiml:
uh -- Is there a way to use Python3 features in CMSSW today?

I know we provide `python3` in the normal builds, this code won't run there (because it needs ROOT and apparently one cannot have ROOT support Python2 and Python3 at the same time?), but it should at least compile?

@smuzaffar maybe.

silviodonato:
please test for CMSSW_11_1_PY3_X
I think we can select explicitly `python3` only in stand-alone scipts (eg. https://github.com/cms-sw/cmssw/blob/master/DQMOffline/Configuration/scripts/cmsswSequenceInfo.py#L1)
but we cannot use python3 in the standard "python" code.  @smuzaffar correct me if I am wrong.

davidlange6:
@silviodonato  what do you mean by "standard python code"? CMSSW supports python3 just as python2 except for externals that don't yet support python3 (some generators), cases where you have to pick either python2 or python3 (ROOT, boost python), and codes that have no unit tests and thus haven't been found to be broken in python3

davidlange6:
The ROOT in the CMSSW master doesn’t support compiling multiple python versions, no. Instead you have the python3 builds as well as the CMSSW build against the ROOT master branch, which does support multiple python versions

smuzaffar:
@schneiml , can you add a simple unit test for this new library? 
Looks like this only works with python3 ( https://github.com/cms-sw/cmssw/pull/29316/files#diff-3cddf755090de9d1e3fbf80f37ae2f01R6 ) ... right? If yes then of course it will not work in any IB other than PY3 Ibs.

schneiml:
@smuzaffar I can turn the example in the readme into a test. The problem is data availability, This is all about reading remote DQMIO data which we usually don't keep on disk for very long.

About Python3: How to proceed there? Even if this works in Py3 IBs (that is how I developed it after all), I think it would be a bad idea to have code that does not compile in a normal release... One could make the code work with Py2 as well, but I think I have better things to do, given we don't really need this right now.

Can we just wait for a Python3-by-default release? When is that planned?

davidlange6:
What is the blocker that was hard to do in a python2 compatible way? I guess a py3-as-default cmssw release is quite some time away (and mostly out of the control of the cmssw developers) - the current ROOT master gets released by ROOT in a few months, so thats likely the shorter term solution.

On Apr 24, 2020, at 12:37 PM, Marcel Schneider <notifications@github.com<mailto:notifications@github.com>> wrote:



@smuzaffar<https://github.com/smuzaffar> I can turn the example in the readme into a test. The problem is data availability, This is all about reading remote DQMIO data which we usually don't keep on disk for very long.

About Python3: How to proceed there? Even if this works in Py3 IBs (that is how I developed it after all), I think it would be a bad idea to have code that does not compile in a normal release... One could make the code work with Py2 as well, but I think I have better things to do, given we don't really need this right now.

Can we just wait for a Python3-by-default release? When is that planned?

schneiml:
@davidlange6 nothing, Except I don't really have time right now and there is also no real need to have this merged. I'll close it for now and maybe come back later.

PR #29318, opened 2020-03-27
Add EventsPerLumi for relvals workflows and prod-like concurrentLS relvals

srimanob:
#### PR description:
This PR is to add an option to set EventsPerLumi in relvals. Normally, GEN relvals will set EventsPerLumi = EventsPerJobs. Setting can be done in 2 ways:
1. propagate from `process.source.numberEventsInLuminosityBlock` in relvals workflow, or
2. parameter of runTheMatrix, `runTheMatrix.py --numberEventsInLuminosityBlock` (overwrite "1." if both options exist)

In addition, 2 additional relvals (1302.181 and 1304.181) are added. They are prod-like 2018 production with concurrentLumis enable.

RelVals: [Link](https://dmytro.web.cern.ch/dmytro/cmsprodmon/workflows.php?campaign=CMSSW_11_1_0_pre4__ml_PHATTEST-1585255954)
+ srimanob_RVCMSSW_11_1_0_pre4ProdTTbar_13UP18ml__ml_PHATTEST_200326_215358_6428 (concurrentLumis enable with new EventsPerLumi setting )
+ srimanob_RVCMSSW_11_1_0_pre4ProdTTbar_13UP18__ml_PHATTEST_200326_215402_5324 (normal relvals workflows, with new EventsPerLumi setting )

#### PR validation:

`runTheMatrix.py --what standard -l 1302.18,1304.18,1302.181,1304.181 -t 4 --wm init`
==> only concurrentLumis workflows have EventsPerLumi = 5 (as set in cmsDriver)

`runTheMatrix.py --what standard -l 1302.18,1304.18,1302.181,1304.181 --numberEventsInLuminosityBlock 20 -t 4 --wm init`
==> All workflows have EventsPerLumi = 20

-

### DIFF
--- a/Configuration/PyReleaseValidation/python/MatrixInjector.py
+++ b/Configuration/PyReleaseValidation/python/MatrixInjector.py
@@ -5,6 +5,7 @@ import os
 import copy
 import multiprocessing
 import time
+import re
 
 def performInjectionOptionTest(opt):
     if opt.show:
@@ -361,12 +362,10 @@ class MatrixInjector(object):
                                     chainDict['nowmTasklist'][-1]['EventsPerLumi'] = ns[1]
                                     #overwrite EventsPerLumi if numberEventsInLuminosityBlock is set in cmsDriver
                                     if 'numberEventsInLuminosityBlock' in s[2][index]:
-                                        nEventsInLuminosityBlock = (((s[2][index].replace(" ","").replace("numberEventsInLuminosityBlock=cms.untracked.uint32("," ")).split()[1]).split(')'))[0]
-                                        if nEventsInLuminosityBlock.isdigit() and int(nEventsInLuminosityBlock) < ns[1]:
-                                            #print('Overwrite EventsPerLumi using cmsDriver customisation')
-                                            chainDict['nowmTasklist'][-1]['EventsPerLumi'] = nEventsInLuminosityBlock
+                                        nEventsInLuminosityBlock = re.findall('process.source.numberEventsInLuminosityBlock=cms.untracked.uint32\(([ 0-9 ]*)\)', s[2][index],re.DOTALL)
+                                        if nEventsInLuminosityBlock[-1].isdigit() and int(nEventsInLuminosityBlock[-1]) < ns[1]:
+                                            chainDict['nowmTasklist'][-1]['EventsPerLumi'] = int(nEventsInLuminosityBlock[-1])
                                     if(self.numberEventsInLuminosityBlock > 0 and self.numberEventsInLuminosityBlock < ns[1]):
-                                        #print('Overwrite EventsPerLumi (again) using runTheMatrix parameter')
                                         chainDict['nowmTasklist'][-1]['EventsPerLumi'] = self.numberEventsInLuminosityBlock
                                 if 'FASTSIM' in s[2][index] or '--fast' in s[2][index]:
                                     thisLabel+='_FastSim'
--- a/Configuration/PyReleaseValidation/python/MatrixInjector.py
+++ b/Configuration/PyReleaseValidation/python/MatrixInjector.py
@@ -53,6 +53,7 @@ class MatrixInjector(object):
         self.keep = opt.keep
         self.memoryOffset = opt.memoryOffset
         self.memPerCore = opt.memPerCore
+        self.numberEventsInLuminosityBlock = opt.numberEventsInLuminosityBlock
         self.batchName = ''
         self.batchTime = str(int(time.time()))
         if(opt.batchName):
@@ -128,6 +129,7 @@ class MatrixInjector(object):
             "GlobalTag": None,
             "SplittingAlgo"  : "EventBased",             #Splitting Algorithm
             "EventsPerJob" : None,                       #Size of jobs in terms of splitting algorithm
+            "EventsPerLumi" : None,
             "RequestNumEvents" : None,                      #Total number of events to generate
             "Seeding" : "AutomaticSeeding",                          #Random seeding method
             "PrimaryDataset" : None,                          #Primary Dataset to be created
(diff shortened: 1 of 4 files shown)


srimanob:
@makortel @Dr15Jones @silviodonato @chayanit
Do you think it's useful if we try to cross LS in every GEN-related relvals, not only in IB? I don't see the cons yet.

Dr15Jones:
Doing more comprehensive testing of jobs with multiple LuminosityBlocks makes sense to me. The down side, is without concurrent lumis (and even with but with module's that inhibit efficient use of concurrent lumis) the multi-thread efficiency of the jobs will descrease. Say we use 4 threads and 5 events per lumi. So on average each thread only processes 1 event and then has to wait for all other threads to finish (because of synchronization on the lumi boundary).

srimanob:
Hi @Dr15Jones 
Thanks for the comment. Normally, relvals run 50 events per job, we do like 10 events/lumi or 25/lumi by default. For IB, it will pick up from cmsDriver directly, 5 events/lumi.

chayanit:
please test workflow 1302.18,1304.18,1302.181,1304.181

Configuration/PyReleaseValidation/python/MatrixInjector.py
```diff
@@ -356,6 +358,16 @@ def prepare(self,mReader, directories, mode='init'):
                                     ns=map(int,arg[arg.index('--relval')+1].split(','))
                                     chainDict['nowmTasklist'][-1]['RequestNumEvents'] = ns[0]
                                     chainDict['nowmTasklist'][-1]['EventsPerJob'] = ns[1]
+                                    chainDict['nowmTasklist'][-1]['EventsPerLumi'] = ns[1]
+                                    #overwrite EventsPerLumi if numberEventsInLuminosityBlock is set in cmsDriver
+                                    if 'numberEventsInLuminosityBlock' in s[2][index]:
+                                        nEventsInLuminosityBlock = (((s[2][index].replace(" ","").replace("numberEventsInLuminosityBlock=cms.untracked.uint32("," ")).split()[1]).split(')'))[0]
```
kpedro88: I'm worried that this is a somewhat fragile construct. For example, if `numberEventsInLuminosityBlock` is set on one line with the type `cms.untracked.uint32()`, and then changed on a subsequent line (when specifying the type is not necessary), this won't pick up the final value.

Also, this currently returns a string; should it be converted to an int?

Proposal:
```
nEventsInLuminosityBlock = eval(s[2][index].split("=")[-1])
if isinstance(nEventsInLuminosityBlock,cms.uint32): nEventsInLuminosityBlock = nEventsInLuminosityBlock.value()
```

This gets the RHS of the expression as its actual type (and then gets the value if it's a cms type). Still not foolproof, would be better to `exec()` the whole config and extract the value, but maybe good enough as a starting point.

Configuration/PyReleaseValidation/python/MatrixInjector.py
```diff
@@ -356,6 +358,16 @@ def prepare(self,mReader, directories, mode='init'):
...
+                                        if nEventsInLuminosityBlock.isdigit() and int(nEventsInLuminosityBlock) < ns[1]:
+                                            #print('Overwrite EventsPerLumi using cmsDriver customisation')
```
kpedro88: delete commented-out code

Configuration/PyReleaseValidation/python/MatrixInjector.py
```diff
@@ -356,6 +358,16 @@ def prepare(self,mReader, directories, mode='init'):
...
+                                            chainDict['nowmTasklist'][-1]['EventsPerLumi'] = nEventsInLuminosityBlock
+                                    if(self.numberEventsInLuminosityBlock > 0 and self.numberEventsInLuminosityBlock < ns[1]):
+                                        #print('Overwrite EventsPerLumi (again) using runTheMatrix parameter')
```
kpedro88: delete commented-out code

Configuration/PyReleaseValidation/python/MatrixInjector.py
srimanob: @kpedro88 
Thanks. Maybe the first question is do we need this first? For example, if it's useful to split all GEN relvals anyways, having the default parameter set, it should be enough. I think this setting is there, just to make LS crossing in IB. We don't set it in any regular relvals.

kpedro88: @srimanob I think the code in my proposal is a bit simpler/easier to parse (eval + only one split vs. two replaces, two splits, multiple indices). Also, I foresee some relvals testing RandomizedParameters eventually, which would also require this parameter, and it may not be provided in exactly the same way.

srimanob: Hi,
Oh, I mean I can change, I have no objection in code improvement. When I say "do we need this first?", I mean the setting using this method, instead of using parameter defined in runTheMatrix directly, also in this PR.

Dr15Jones: Why even use a CMS specific type at all? The parameter is never used by the C++ code, correct?

kpedro88: @Dr15Jones I'm not sure what you mean; `numberEventsInLuminosityBlock` is definitely used in the C++ code: https://github.com/cms-sw/cmssw/search?l=C%2B%2B&q=numberEventsInLuminosityBlock

Dr15Jones: My bad.

srimanob: Not sure how the proposal works?
s[2][index] is the full cmsDriver (*), not just
`process.source.numberEventsInLuminosityBlock=cms.untracked.uint32(5)`

Do I miss understand anything? Thx.

(*)
`cmsDriver.py TTbar_13TeV_TuneCUETP8M1_cfi --conditions auto:phase1_2018_realistic -n 10 --customise_commands "process.source.numberEventsInLuminosityBlock=cms.untracked.uint32(5)" --era Run2_2018 --relval 9000,50 -s GEN,SIM --nStreams 4 --datatier GEN-SIM --beamspot Realistic25ns13TeVEarly2018Collision --eventcontent RAWSIM --geometry DB:Extended --nConcurrentLumis 2 --io ProdTTbar_13UP18ml.io --python ProdTTbar_13UP18ml.py --no_exec --fileout file:step1.root`

kpedro88: Ah, I misunderstood; I thought that variable contained lines from a config file, not a cmsDriver command. In that case:
```
import shlex
lumiCmds = [cmd for cmd in shlex.split(s[2][index]) if 'numberEventsInLuminosityBlock' in cmd]
if len(lumiCmds)>0:
    nEventsInLuminosityBlock = eval(lumiCmds[-1].split("=")[-1])
    if isinstance(nEventsInLuminosityBlock,cms.uint32): nEventsInLuminosityBlock = nEventsInLuminosityBlock.value()
```

Here I use `shlex.split()` to separate the command-line arguments (best practice), and then use the last argument that contains the desired variable to extract the value.

srimanob: Is `eval(lumiCmds[-1].split("=")[-1])` really work with statement like`cms.untracked.uint32(5)` ?
It seems to me, eval tries to evaluate the string `cms.untracked.uint32(5)` then lead to 
`NameError: name 'cms' is not defined`

Is it not straight forward to go with regex instead, and check condition of output? This comes without any loops.
```
ss = re.findall('process.source.numberEventsInLuminosityBlock=cms.untracked.uint32\(([ 0-9 ]*)\)', s,re.DOTALL)
if(ss[-1].isDigit):
   nEventsInLuminosityBlock = int(ss[-1])
```
As prefer, I have no strong opinion. Maybe some guide on how to make eval works with this case.

kpedro88: This works fine for me in a python terminal:
```
x = eval("cms.untracked.uint32(5)")
print x
```

But the regex solution also looks reasonable to me.

srimanob: Done, thx.

chayanit:
please test workflow 1302.18,1304.18,1302.181,1304.181

kpedro88:
tests appear to be stuck?

kpedro88:
test parameters
workflows = 1302.18,1304.18,1302.181,1304.181

kpedro88:
please test

kpedro88:
+upgrade

silviodonato:
Do you have comments  @chayanit @pgunnell @wajidalikhan ?

chayanit:
+1

silviodonato:
+1

Issue #29314, opened 2020-03-26
Handling summary histograms in DQMEDAnalyzers


Dr15Jones:
Many of the modules that are presently DQMOneLumiEDAnalyzer just need a way to fill 'summary' histograms at global end Lumi transition. A way to support this in a concurrent manner would be to have those module use a Lumi cache which holds a pointer to the MonitorElement from the base stream:: type module. Then during the call to the globalEndLuminosityBlockProducer, the base class would call a member function (e.g. `finish()`) of the Lumi cache object to do what ever is needed to do the summary.

The same could be done to handle Run summary histograms.

Dr15Jones:
assign core, dqm

schneiml:
@Dr15Jones Examples?

For summaries derived from other MEs: this should simply all be done in harvesting.

For summaries derived from local variables: this can be handled correctly if we simply expose the `endLumi` to the stream modules. The modules need to do things correctly (`Fill`, not `setBinContent`), but it should work. Actually, this is exactly what we did in 2014.

There are more cases, see here: https://indico.cern.ch/event/870369/#5-dqmstore-leftovers-migration

Dr15Jones:
Here
https://github.com/cms-sw/cmssw/blob/7bf830bbb5f57be6673227056342a5adf8c88567/DQMOffline/JetMET/src/METAnalyzer.cc#L1382

and pretty much any use of `setBinContent` from within an DQMEDAnalyzer (which includes DQMOneEDAnalyzer and such).

schneiml:
@Dr15Jones but in `METAnalyzer`, ah, but that is `endRun`. `METAnalyzer` is only `DQMOneLumiEDAnalyzer` recently to get a plot per-lumi saved, it does not care about lumi transitions otherwise. 

Actually, it was a normal `edm::stream` `DQMEDAnalyzer` after this PR https://github.com/cms-sw/cmssw/pull/7656 before we started messing around with one modules.

In general I do agree that we should work towards completely banning `getBinContent` and maybe also `setBinContent` from `dqm::reco::MonitorElement`.

jfernan2:
+1

PR #29321, opened 2020-03-27
DQM: Allow per-lumi MEs in DQMOneEDAnalyzer

schneiml:
#### PR description:

After the discussion started by @Dr15Jones in #29314 , I remembered that we still had the option to do the lumi-switching in the `DQMStore` on a per-event basis. This allows modules like `METAnalyzer` to book per-lumi histograms but remain `DQMOneEDAnalyzer` and yet not block concurrent lumis.

The cost is higher overhead. The `enterLumi/leaveLumi` calls should be fast in the common case (take a lock, do a map lookup, do nothing), but with many lumi MEs (for example when per-lumi saving by default like in the UL ReReco) all the pointer bending could cause some overhead. However, even that might still be negligible, or could be optimized a bit more if it is a problem.

#### PR validation:

I tested a workflow with `METAnalyzer` and it seems to come out well. The PR tests should catch eventual differences.

I also updated the unit tests to excercise this feature and it seems fine. However, I don't observe any concurrent lumis in the unit tests, even though they are enabled and there should not be any blocking modules. It would be nice to see some test case where events actually alternate between lumis.

### DIFF
--- a/DQMOffline/JetMET/interface/METAnalyzer.h
+++ b/DQMOffline/JetMET/interface/METAnalyzer.h
@@ -86,7 +86,7 @@
 #include <map>
 #include <string>
 
-class METAnalyzer : public DQMOneLumiEDAnalyzer<> {
+class METAnalyzer : public DQMOneEDAnalyzer<> {
 public:
   /// Constructor
   METAnalyzer(const edm::ParameterSet&);
--- a/DQMServices/Core/README.md
+++ b/DQMServices/Core/README.md
@@ -24,7 +24,7 @@ DQM code runs as CMSSW plugins. There are two main types of plugins: *Analyzers*
 
 There are six supported types of DQM modules:
 - `DQMEDAnalyzer`, based on `edm::stream::EDProducer`. Used for the majority of histogram filling in RECO jobs.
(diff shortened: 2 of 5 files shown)


schneiml:
please test

There should not be any differences, and that should actually sort of validate the change.

schneiml:
please test

Maybe now.

schneiml:
Unless anybody objects (@Dr15Jones , @makortel ?) I think this is a viable solution. Paying special attention to the `DQMOneEDAnalyzer`s in RelVals is of course mandated, esp. if concurrent lumis are actually used (I don't think the DQM unit tests that enable concurrent lumis actually managed to produce some events where the one modules went back-and forth between lumis).

Otherwise, I think we can go ahead here.

schneiml:
please test

Nothing changes, but the pervious comparison is gone and things may have changed in master.

schneiml:
The changes in 1330 are unexpected. They were also present in the previous tests and are not in #29438 (which is clean otherwise). So, this needs to be investigated.

schneiml:
It turns out the changes here are a well known problem.

So, this looks all fine to me.

schneiml:
+1

silviodonato:
+1

PR #29329, opened 2020-03-27
Fix HGCal_disableNoise customization function

rovere:
#### PR description:

The logic with which the parameters are parsed requires the presence of
many parameters that were not present in the customization function.

#### PR validation:

Tested on workflow 23234, with and without the customization function. The code works and the Noise is turned off with the customization active.


civanch:
please test

civanch:
+1

kpedro88:
+upgrade

silviodonato:
+1

PR #29330, opened 2020-03-28
Fix MET unclustered energy computation [10_6_X]

ahinzmann:
#### PR description:

1. Fix MET significance and unclustered pT computation for PUPPI MET. Previously these lines failed for the candidates inside jets as puppi MET and puppi jets are clustered from different particle collections with different particle momenta:
https://github.com/cms-sw/cmssw/blob/760c98128277dcab5dc4d2887c5c814cb0401740/RecoMET/METAlgorithms/src/METSignificance.cc#L95-L99
Particles from jets with puppi!=puppiForMET weights were not removed or falsely kept from unclustered energy.

2. Fix MET unclusted pT uncertainty computation for PUPPI MET. Previously this line failed for the candidates inside jets as puppi MET and puppi jets are clustered from different particle collections:
https://github.com/cms-sw/cmssw/blob/760c98128277dcab5dc4d2887c5c814cb0401740/PhysicsTools/PatUtils/python/tools/runMETCorrectionsAndUncertainties.py#L791
Particles from jets were not removed from unclustered energy.

3. Replace dE-matching by dR-matching when computing unclustered pT and MET signficance for PF MET and PUPPI MET. Some low pT particles randomly passed the dE matching even though they did not point in the same direction, leading to changes up to 25%.

In this PR for UL-reMiniAOD, the issue is fixed by matching candidates based on dR instead of Ptrs. In the corresponding PR in the master #29254, the issue is solved by using the same particle flow collections for puppi jets and puppi MET and separate weight valuemaps.

#### PR validation:

On 100 events of 136.8311, it was checked that this PR obtains the same values of metSumPtUnlustered, metSignficance and unclustered energy uncertainties as the corresponding PT in the master #29254.

#### if this PR is a backport please specify the original PR and why you need to backport that PR:

This is a backport of the fix in #29331.

PR #29331, opened 2020-03-28
Fix MET unclustered energy computation

ahinzmann:
#### PR description:

1. Fix MET significance and unclustered pT computation for PUPPI MET. Previously these lines failed for the candidates inside jets as puppi MET and puppi jets are clustered from different particle collections with different particle momenta:
https://github.com/cms-sw/cmssw/blob/760c98128277dcab5dc4d2887c5c814cb0401740/RecoMET/METAlgorithms/src/METSignificance.cc#L95-L99
Particles from jets with puppi!=puppiForMET weights were not removed or falsely kept from unclustered energy.

2. Fix MET unclusted pT uncertainty computation for PUPPI MET. Previously this line failed for the candidates inside jets as puppi MET and puppi jets are clustered from different particle collections:
https://github.com/cms-sw/cmssw/blob/760c98128277dcab5dc4d2887c5c814cb0401740/PhysicsTools/PatUtils/python/tools/runMETCorrectionsAndUncertainties.py#L791
Particles from jets were not removed from unclustered energy.

3. Replace dE-matching by dR-matching when computing unclustered pT and MET signficance for PF MET and PUPPI MET. Some low pT particles randomly passed the dE matching even though they did not point in the same direction, leading to changes up to 25%.

In this PR, the issue is fixed by matching candidates based on dR instead of Ptrs. In the following PR #29254, the issue is solved by using the same particle flow collections for puppi jets and puppi MET and separate weight valuemaps.

#### PR validation:

On 100 events of 136.8311, it was checked that this PR obtains the same values of metSumPtUnlustered, metSignficance and unclustered energy uncertainties as the following PR #29254.

Backport to 10_6 for UL-reminiaod in #29330.

### DIFF
--- a/CommonTools/CandAlgos/plugins/CandPtrProjector.cc
+++ b/CommonTools/CandAlgos/plugins/CandPtrProjector.cc
@@ -66,7 +66,7 @@ void CandPtrProjector::fillDescriptions(edm::ConfigurationDescriptions& descript
   desc.add<edm::InputTag>("src");
   desc.add<edm::InputTag>("veto");
   desc.add<bool>("useDeltaRforFootprint", false);
-  descriptions.add("CandPtrProjector", desc);
+  descriptions.addWithDefaultLabel(desc);
 }
 
 DEFINE_FWK_MODULE(CandPtrProjector);
--- a/CommonTools/CandAlgos/plugins/CandPtrProjector.cc
+++ b/CommonTools/CandAlgos/plugins/CandPtrProjector.cc
@@ -7,6 +7,7 @@
 #include "DataFormats/Common/interface/RefToBaseVector.h"
 #include "DataFormats/Candidate/interface/Candidate.h"
 #include "DataFormats/Candidate/interface/CandidateFwd.h"
+#include "DataFormats/Math/interface/deltaR.h"
 
 class CandPtrProjector : public edm::global::EDProducer<> {
 public:
@@ -16,11 +17,14 @@ public:
 private:
   edm::EDGetTokenT<edm::View<reco::Candidate>> candSrcToken_;
   edm::EDGetTokenT<edm::View<reco::Candidate>> vetoSrcToken_;
+  bool useDeltaRforFootprint_;
 };
 
 CandPtrProjector::CandPtrProjector(edm::ParameterSet const& iConfig)
     : candSrcToken_{consumes<edm::View<reco::Candidate>>(iConfig.getParameter<edm::InputTag>("src"))},
       vetoSrcToken_{consumes<edm::View<reco::Candidate>>(iConfig.getParameter<edm::InputTag>("veto"))} {
+  useDeltaRforFootprint_ =
+      iConfig.exists("useDeltaRforFootprint") ? iConfig.getParameter<bool>("useDeltaRforFootprint") : false;
   produces<edm::PtrVector<reco::Candidate>>();
 }
 
@@ -43,7 +47,15 @@ void CandPtrProjector::produce(edm::StreamID, edm::Event& iEvent, edm::EventSetu
   for (size_t i{}; i < cands->size(); ++i) {
     auto const c = cands->ptrAt(i);
     if (vetoedPtrs.find(c) == vetoedPtrs.cend()) {
-      result->push_back(c);
+      bool addcand = true;
+      if (useDeltaRforFootprint_)
+        for (const auto& it : vetoedPtrs)
+          if (reco::deltaR2(it->p4(), c->p4()) < 0.00000025) {
+            addcand = false;
+            break;
+          }
+      if (addcand)
+        result->push_back(c);
     }
   }
   iEvent.put(std::move(result));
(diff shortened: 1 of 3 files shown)


ahinzmann:
Can we postpone the fillDescriptions implementation to a followup PR, to wrap up #29254?

CommonTools/CandAlgos/plugins/CandPtrProjector.cc
```diff
@@ -16,11 +17,14 @@ class CandPtrProjector : public edm::global::EDProducer<> {
...
   edm::EDGetTokenT<edm::View<reco::Candidate>> candSrcToken_;
   edm::EDGetTokenT<edm::View<reco::Candidate>> vetoSrcToken_;
+  bool useDeltaRforFootprint_;
 };
 
 CandPtrProjector::CandPtrProjector(edm::ParameterSet const& iConfig)
     : candSrcToken_{consumes<edm::View<reco::Candidate>>(iConfig.getParameter<edm::InputTag>("src"))},
       vetoSrcToken_{consumes<edm::View<reco::Candidate>>(iConfig.getParameter<edm::InputTag>("veto"))} {
+  useDeltaRforFootprint_ =
+      iConfig.exists("useDeltaRforFootprint") ? iConfig.getParameter<bool>("useDeltaRforFootprint") : false;
```
slava77: I suppose that this config parameter is still useful/needed here (regardless of the PUPPI MET developments).
Please add this as a required parameter.

PhysicsTools/PatUtils/python/tools/runMETCorrectionsAndUncertainties.py
```diff
@@ -786,6 +786,8 @@ def getMETUncertainties(self, process, metType, metModName, electronCollection,
                                            src = pfCandCollection, 
                                            veto = jetCollection
                                            )
+            if self._parameters["Puppi"].value:
+              pfCandsNoJets.useDeltaRforFootprint = cms.bool(True)
```
slava77: once useDeltaRforFootprint is added as required, ```cms.bool``` can be dropped

slava77:
@cmsbuild please test

slava77:
@cmsbuild please test

CommonTools/CandAlgos/plugins/CandPtrProjector.cc
```diff
@@ -43,10 +47,26 @@ void CandPtrProjector::produce(edm::StreamID, edm::Event& iEvent, edm::EventSetu
...
   }
   iEvent.put(std::move(result));
 }
 
+void CandPtrProjector::fillDescriptions(edm::ConfigurationDescriptions& descriptions) {
+  edm::ParameterSetDescription desc;
+  desc.add<edm::InputTag>("src");
+  desc.add<edm::InputTag>("veto");
+  desc.add<bool>("useDeltaRforFootprint", false);
+  descriptions.add("CandPtrProjector", desc);
```
slava77: default naming is to start with lower case.
```descriptions.addWithDefaultLabel(desc);``` will do it for you

slava77:
@cmsbuild please test

slava77:
the effect on the unclustered sumPt is apparently quite large
e.g. from 1325.7 (ttbar with PU)
wf1325 7_sumPtUnclustered

ahinzmann:
> the effect on the unclustered sumPt is apparently quite large
> e.g. from 1325.7 (ttbar with PU)
> wf1325 7_sumPtUnclustered

Indeed. Before this fix, all constituents of jets with pT>15 GeV were wrongly added to the unclustered energy for PUPPI MET.

slava77:
> Indeed. Before this fix, all constituents of jets with pT>15 GeV were wrongly added to the unclustered energy for PUPPI MET.

This may sound like the expectation for the fix is a reduction in the unclustered sumpt, but the plot shows that the sumpt after the fix is increased.

slava77:
IIUC, the change in CandPtrProjector.cc leads to changes in the MET uncertainty like the following in wf 136.7611
all_OldVSNew_RunJetHT2016EreMINIAODwf136p7611c_patMETs_slimmedMETsPuppi__PAT_obj_0__uncertainties__10__dpx88

While the change in METSignificance.cc , switching from delta on Et2 to deltaR, apparently leads to the increase in sumpt

slava77:
+1

for https://github.com/cms-sw/cmssw/pull/29331 5fee5c8 
- jenkins tests pass and comparisons with the baseline show differences in pat::MET both in the computed metSumPtUnclustered (typically increasing) and in the MET uncertainties related to the unclustered energy (typically decreasing) due to different kinds of bugfixes

ahinzmann:
> > Indeed. Before this fix, all constituents of jets with pT>15 GeV were wrongly added to the unclustered energy for PUPPI MET.
> 
> This may sound like the expectation for the fix is a reduction in the unclustered sumpt, but the plot shows that the sumpt after the fix is increased.

In the case of METSignficance (rather than CandPtrProjector), there are two competing effects:
1. Particles that were not removed because dE matching failed, but dR matching now removes them.
2. Particles that were wrongly removed by dE matching, even though they did not match in dR.

As a representative example, in the first event of 136.8311:
- For PUPPI MET there are 6 particles with a sumPt of 18.4 GeV of type 1. and 6 particles with a sumPt of 1.6 GeV of type 2. with an average dR of 4.3 to the particle that was dE-matched, resulting overall in a slightly lower sumPtUnclustered with dR matching compared to dE matching.
- For PF MET there are 0 particles of type 1. and 523 particles of type 2. with a sumPt of 264 GeV with an average dR of 3.7 to the particle that was dE-matched, resulting in a significantly higher sumPtUnclustered with dR matching compared to dE matching.

slava77:
@silviodonato 
please consider this for merging.
This is a prerequisite for #29254

santocch:
+1

silviodonato:
+1

PR #29333, opened 2020-03-29
pixel mc gain calibration scheme: new GTs for MC Run3, modified Clusterizer conditions for Run3 

dkotlins:
#### PR description:

Modify the scheme of the pixel gain calibration. Instead of applying the vcal calibration in the 
clusterizer include it already in the gain calibration payload.  Like this the vcal calibration can 
be easily adjusted for each new gain calibration.
This is the first step in which we only change the MC gain calibrations. It should apply to any future Run3 simulations. Data from Run1/2 and older, none Run3 MC, will not be affected.

To be applied for Run 3 MC & Data. No code changes, the only changes are:
1. new GTs for Run3 MC which include the new gain calibration values,
2. changes in the clusterizer parameters for Run 3, for the VCAL calibration use slope=1 and offset =0.

  - what changes are expected in the output if any,
No changes expected to the output 
  - what other PRs or externals it depends upon if any,
There are no other PRs, but the change must be in sync with a GT change, this is why the autoCond.py file is also modified.

  - link to any additional material useful to provide a documentation for this PR (slides, JIRA tickets, related pull requestes, hypernews, TWiki or Indico pages)  
https://indico.cern.ch/event/879470/contributions/3796405/attachments/2009273/3356603/pix_off_25_3_gain_calibration_mc.pdf

No backport required.

The changes have been verified using Run3 simulations.

### DIFF
--- a/Configuration/AlCa/python/autoCond.py
+++ b/Configuration/AlCa/python/autoCond.py
@@ -65,19 +65,19 @@ autoCond = {
     # GlobalTag for MC production (cosmics) with realistic conditions for full Phase1 2018 detector,  Strip tracker in DECO mode
     'phase1_2018_cosmics'      :   '110X_upgrade2018cosmics_realistic_deco_v6',
     # GlobalTag for MC production (cosmics) with realistic conditions for full Phase1 2018 detector,  Strip tracker in PEAK mode
-    'phase1_2018_cosmics_peak' :   '110X_upgrade2018cosmics_realistic_peak_v7',
+    'phase1_2018_cosmics_peak' :   '110X_upgrade2018cosmics_realistic_peak_v7',   
     # GlobalTag for MC production with perfectly aligned and calibrated detector for Phase1 2021
-    'phase1_2021_design'       : '111X_mcRun3_2021_design_v3', # GT containing design conditions for Phase1 2021
+    'phase1_2021_design'       : '111X_mcRun3_2021_design_newPixelGain_v1', # GT containing design conditions for Phase1 2021
     # GlobalTag for MC production with realistic conditions for Phase1 2021
-    'phase1_2021_realistic'    : '111X_mcRun3_2021_realistic_v3', # GT containing realistic conditions for Phase1 2021
+    'phase1_2021_realistic'    : '111X_mcRun3_2021_realistic_newPixelGain_v1', # GT containing realistic conditions for Phase1 2021
     # GlobalTag for MC production (cosmics) with realistic conditions for Phase1 2021,  Strip tracker in DECO mode
-    'phase1_2021_cosmics'      : '111X_mcRun3_2021cosmics_realistic_deco_v3',
+    'phase1_2021_cosmics'      : '111X_mcRun3_2021cosmics_realistic_deco_newPixelGain_v1',
     # GlobalTag for MC production with realistic conditions for Phase1 2021 detector for Heavy Ion
-    'phase1_2021_realistic_hi' :  '111X_mcRun3_2021_realistic_HI_v4',
+    'phase1_2021_realistic_hi' :  '111X_mcRun3_2021_realistic_HI_newPixelGain_v1',
     # GlobalTag for MC production with realistic conditions for Phase1 2023
-    'phase1_2023_realistic'    : '111X_mcRun3_2023_realistic_v3', # GT containing realistic conditions for Phase1 2023
+    'phase1_2023_realistic'    : '111X_mcRun3_2023_realistic_newPixelGain_v1', # GT containing realistic conditions for Phase1 2023
     # GlobalTag for MC production with realistic conditions for Phase1 2024
-    'phase1_2024_realistic'    : '111X_mcRun3_2024_realistic_v3', # GT containing realistic conditions for Phase1 2024
+    'phase1_2024_realistic'    : '111X_mcRun3_2024_realistic_new_PixelGain_v1', # GT containing realistic conditions for Phase1 2024
     # GlobalTag for MC production with realistic conditions for Phase2
     'phase2_realistic'         : '110X_mcRun4_realistic_v3'
 }
--- a/Configuration/AlCa/python/autoCond.py
+++ b/Configuration/AlCa/python/autoCond.py
@@ -77,7 +77,7 @@ autoCond = {
     # GlobalTag for MC production with realistic conditions for Phase1 2023
     'phase1_2023_realistic'    : '111X_mcRun3_2023_realistic_newPixelGain_v1', # GT containing realistic conditions for Phase1 2023
     # GlobalTag for MC production with realistic conditions for Phase1 2024
-    'phase1_2024_realistic'    : '111X_mcRun3_2024_realistic_new_PixelGain_v1', # GT containing realistic conditions for Phase1 2024
+    'phase1_2024_realistic'    : '111X_mcRun3_2024_realistic_newPixelGain_v1', # GT containing realistic conditions for Phase1 2024
     # GlobalTag for MC production with realistic conditions for Phase2
     'phase2_realistic'         : '110X_mcRun4_realistic_v3'
 }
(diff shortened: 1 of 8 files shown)


mmusich:
@dkotlins this PR has code style and quality issues. Can you please resolve them as suggested at https://github.com/cms-sw/cmssw/pull/29333#issuecomment-605903814? 
In alternative you can run `scram b code-format` and `scram b code-checks` in your local development area and then push the changes.
Otherwise the PR integration tests cannot be performed.

slava77:
@cmsbuild please test

mmusich (commented):
@dkotlins I have few comments, mostly stylistic.

Configuration/AlCa/python/autoCond.py
```diff
@@ -65,19 +65,19 @@
...
+    'phase1_2021_cosmics'      : '111X_mcRun3_2021cosmics_realistic_deco_newPixelGain_v1',
     # GlobalTag for MC production with realistic conditions for Phase1 2021 detector for Heavy Ion
-    'phase1_2021_realistic_hi' :  '111X_mcRun3_2021_realistic_HI_v4',
+    'phase1_2021_realistic_hi' :  '111X_mcRun3_2021_realistic_HI_newPixelGain_v1',
     # GlobalTag for MC production with realistic conditions for Phase1 2023
-    'phase1_2023_realistic'    : '111X_mcRun3_2023_realistic_v3', # GT containing realistic conditions for Phase1 2023
+    'phase1_2023_realistic'    : '111X_mcRun3_2023_realistic_newPixelGain_v1', # GT containing realistic conditions for Phase1 2023
     # GlobalTag for MC production with realistic conditions for Phase1 2024
-    'phase1_2024_realistic'    : '111X_mcRun3_2024_realistic_v3', # GT containing realistic conditions for Phase1 2024
+    'phase1_2024_realistic'    : '111X_mcRun3_2024_realistic_newPixelGain_v1', # GT containing realistic conditions for Phase1 2024
```
mmusich: @christopheralanwest @tlampen @tocheng given that these Global Tags are meant to be integrated in release rather than for testing, shouldn't they named according to the normal naming conventions?

dkotlins:
Concerning the Era=Run3 modification, I thought that it was necessary to define it again.
This constants might be anyway modified for Run3. 
I will remove it now.
D.

dkotlins:
Here I am not guilty. These names were provided by Tongguang Cheng.
I was just trying to make a consistent PR, one which works, but after all it is not my 
task to find good names for GTs.

I propose to cancel this Pull request. 
I will make a new one which will include just the pixel change and I will specify 
that it has to be accompanied by a corresponding new GT.
The two will have to be synchronised, otherwise it will not work.
D.

mmusich:
> I propose to cancel this Pull request.
I will make a new one which will include just the pixel change and I will specify
that it has to be accompanied by a corresponding new GT.
The two will have to be synchronised, otherwise it will not work.

I'd just rather wait for AlCa/DB to clarify if they plan to provide better named GTs or they plan to sign the update as is.

tocheng:
I will give GTs official names and you can update them with a commit. Thanks.

tocheng:
@dkotlins @mmusich 
Please use this commit
https://github.com/tocheng/cmssw/commit/1f941d72f1b7d06d226f80328b0c2dfbb901b0ae

Thanks.

dkotlins:
I am not sure what should I do.
Can I use your commit somehow or should I just retype your names to my commit?
D.

mmusich:
@dkotlins Danek, for you it is probably faster just to retype the new names in your local area and then push again. Please try to address also the other comments.
Thanks
M.

christopheralanwest:
please test

perrotta:
@dkotlins (@mmusich) please edit the title and the description of this PR so that it can reflect what is done here now that you removed all the CondTools related part:
- new GTs
- skip the vcal calibration in the clusterizer, by setting "slope = 1 & offfset = 0." in the vcal

Please also check the differences in the Run3 outputs, and verify whether they correspond to what expected

dkotlins:
**PR: Description**

New scheme for the pixel gain calibration.
To be applied for Run 3 MC & Data. No code changes.
- new GTs for Run3 MC which include the new gain calibration values,
- changes in the clusterizer parameters for Run 3, for the VCAL calibration use slope=1 and offset =0.

dkotlins:
I have changed the title.
I do not know how to modify the initial PR description, so I just added a comment.

I could not find any relevant changes coming from this PR. 
There are all sorts of ECAL/HCAL differences but I do not see how this is 
related to pixels.
The changes in the pixel cluster charge are tiny, below any statistical significance.

D.

perrotta:
pr

In the top right of the PR description panel you can access the commands for it. One of them is "edit", which is only allowed to the owner of the PR

perrotta:
There are several pixel related differences in jenkins outputs, which I assume due to the modified GT.
I imagine that those differences propagate with some tiny effect to the reconstructed quantities, but I think it is worth to validate/confirm in some larger sample that the low-stat jenkins tests outputs.

By the way, for my own knowledge, could you please point out were the calibrations taken from the DB replaces in run3 workflows the ones set by the gain/offset parameters, which are set here to 0/1 respectively?

dkotlins:
I am not sure that I understand the question.
Until now the gain calibration was taken from the DB and than corrected for the vcal calibration
in the clusterizer using parameters taken from the configuration.
Now the vcal calibration constants are already inlcuded in the gain calibration in the DB.
So the constant taken form the DB do not need it. All the calibrations are included in the DN payload.

This is explained in slides, link below, there are also plots there with higher statistics showing that the 
two approaches give the same answers:
https://indico.cern.ch/event/879470/contributions/3796405/attachments/2009273/3356603/pix_off_25_3_gain_calibration_mc.pdf

The results cannot be exactly the same because of random generators using for smearing around the average values 
and probably because of binning effects . 
The 2nd one is because if:
 - you take a distribution and bin it, 
  - take the same distribution, rescale it, bin it, rescale back,
I believe the 2 will have some numerical differences.

D.

perrotta:
@dkotlins sorry for getting late on it, as I was busy with other things.

My question starts from 
"For this to work one has to skip the vcal calibration in the clusterizer.
No code changes are need, it is enough to set the vcal to
slope = 1 & offfset = 0."

Questions:
- Do I understand correctly that with such a "neutralization" of slope and offset in the python configurations those values are taken from the DB, while otherwise the values written in the python configuration are taken instead?
- Could you please point me the place in the code where such a switch to the GT given those values for slope and offset is actually implemented?
- Why the same procedure of having gains and offsets in the DB is not implemented also for Run1/Run2? Is it just because you consider them fozen, and you don't consider there can be further calibrations that can modify them?

mmusich:
@perrotta taking the liberty to provide some answers, @dkotlins can of course complement.

 >  Do I understand correctly that with such a "neutralization" of slope and offset in the python configurations those values are taken from the DB, while otherwise the values written in the python configuration are taken instead?

The per pixel gains are still taken from DB, only now the values don't get multipied and offset because that operation is already done *within* the numbers stored in DB.

>    Could you please point me the place in the code where such a switch to the GT given those values for slope and offset is actually implemented?

There is no such switch, the gain payloads for Run3 in the DB are just populated differently, taking into account that slope is 1 and offset is 0, so the "new" gains get multiplied by a factor ~ 50 see plot below (red is "old" gain scheme, blue is "new gain scheme).
The net effect is that the calibration remains ~ the same (with the caveat that Danek reported at https://github.com/cms-sw/cmssw/pull/29333#issuecomment-610206141)

image
 
>    Why the same procedure of having gains and offsets in the DB is not implemented also for Run1/Run2? Is it just because you consider them fozen, and you don't consider there can be further calibrations that can modify them?

To avoid redoing dozens of payloads to keep physics the same... 
The run2 / run3 boundary is an excellent moment to do this, because in this way we can just switch to the new scheme for the future payloads (all of them, of course, yet to be derived).

perrotta:
+1
- For reco, only the SiPixelClusterizer_cfi.py configuration gets modified for Run3, where "neutral" values are assigned to slope and offset so that the pixel mc gain remains only defined by the payloads in the GT
- These new gains for Run3 are somehow modified with respect to those derived from the previous formulation. The differences observed in the jenkins outputs are minor, but probably a bit larger than possible numerical rounding effects, and they propagate to higher level quantities as reduced calo recHits, electrons, jets, met, etc.
  - AlCa will take care of verifying those new gains in the modified GT payloads while signing
  - A clearer picture can be obtained with the usual pre-release validations: I look forward this PR can enter the next 11_X pre-release, so that its effects can be thoroughly inspected

mmusich:
Can AlCa please review?
@tlampen @tocheng @christopheralanwest

tocheng:
+1

Sorry for being late.

silviodonato:
+1
the differences in DQM plots are explained in https://github.com/cms-sw/cmssw/pull/29333#issuecomment-610206141

PR #29343, opened 2020-03-30
DQM: fix some autoDQM.py harvesting sequences

schneiml:
#### PR description:
It turns out that `ConfigBuilder` treats failing to find a HARVESTING sequence not as an error, and in fact we have some non-existing sequences in `autoDQM.py`. One of them is `DQMHarvestDCS`, which I forgot to remove in #28829.

But also `DQMNone`, which is used to signify that there is nothing to do (`''` could be understood as default sequence, as usual in `cmsDriver`) is not actually defined in harvesting. So I added that to all the main config files.

Finally, this PR makes `ConfigBuilder` fail on undefined harvesting sequences, like it does for other sequences. 

#### PR validation:

Functionality should not be affected, but misconfigured jobs may crash now. I expect some PR and even IB failures untilwe catch all of them.

### DIFF
--- a/Configuration/Applications/python/ConfigBuilder.py
+++ b/Configuration/Applications/python/ConfigBuilder.py
@@ -2000,6 +2000,8 @@ class ConfigBuilder(object):
         for name in harvestingList:
             if not name in harvestingConfig.__dict__:
                 print(name,"is not a possible harvesting type. Available are",harvestingConfig.__dict__.keys())
+                # trigger hard error, like for other sequence types
+                getattr(self.process, name)
                 continue
             harvestingstream = getattr(harvestingConfig,name)
             if isinstance(harvestingstream,cms.Path):
--- a/DQMOffline/Configuration/python/DQMOfflineCosmics_SecondStep_cff.py
+++ b/DQMOffline/Configuration/python/DQMOfflineCosmics_SecondStep_cff.py
@@ -13,6 +13,8 @@ from DQM.RPCMonitorClient.RPCTier0Client_cff import *
 from DQM.CSCMonitorModule.csc_dqm_offlineclient_cosmics_cff import *
 from DQMServices.Components.DQMFEDIntegrityClient_cff import *
 
+DQMNone = cms.Sequence()
+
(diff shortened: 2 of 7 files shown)


schneiml:
please test

This has some chance of blowing up due to more undefined harvesting sequences. Let's see.

schneiml:
+1 

looks good so far, though be aware of potential failures due to misconfigured harvesting in the IB tests.

silviodonato:
+operations

silviodonato:
+1

silviodonato:
@schneiml we are getting errors in CMSSW_11_1_X_2020-04-02-1100 in wf 140.0, 145.0, 150.0 .
https://cmssdt.cern.ch/SDT/cgi-bin/buildlogs/raw/slc7_amd64_gcc820/CMSSW_11_1_X_2020-04-02-1100/pyRelValMatrixLogs/run/145.0_HydjetQ_B12_5020GeV_2015+HydjetQ_B12_5020GeV_2015+DIGIHI2015+RECOHI2015+HARVESTHI2015/step4_HydjetQ_B12_5020GeV_2015+HydjetQ_B12_5020GeV_2015+DIGIHI2015+RECOHI2015+HARVESTHI2015.log

Do you think it is related to this PR? Tomorrow I will check it better.

schneiml:
@silviodonato yep, that is related. `validationHarvestingNoHLT` (in your link) does not exist and should be removed from all configs that use it.

silviodonato:
Do you mean we have to remove `validationHarvestingNoHLT+` from HARVESTHI2011, HARVESTHI2015, HARVESTHI2018 in
https://github.com/cms-sw/cmssw/blob/master/Configuration/PyReleaseValidation/python/relval_steps.py
?

silviodonato:
`validationHarvestingNoHLT` is defined here https://github.com/cms-sw/cmssw/blob/master/Configuration/StandardSequences/python/Harvesting_cff.py#L26

silviodonato:
https://github.com/cms-sw/cmssw/pull/29393 should fix the error

PR #29351, opened 2020-03-31
DQM: Fix fallback in DQMRootSource

schneiml:
#### PR description:
A small bugfix to make opening remote files work in `DQMRootSource`, that is, in HARVESTING jobs.

#### PR validation:

This code path is not used in production setups; all production harvesting jobs download the input files locally first. It rarely worked in the past, now it works again (in my private tests). No effect on output expected.


schneiml:
+1

should not affect anything.

silviodonato:
+1

PR #29355, opened 2020-03-31
Additional Tracker DQM residual plots

dmeuser:
#### PR description:

We added plots to the TkDQM showing the mean residual and rms of the residual on module level separating:
 - inner/outer modules for BPIX and FPIX
 - positibe/negative side for FPIX

The plots are filled from information already available in the TProfile2D plots.

#### PR validation:

The changes have been tested by running `runTheMatrix.py -l 10024`, which runs without any failed steps and produces the desired additional plots.

This PR is not a backport.

cc:
@sroychow @arossi83

### DIFF
--- a/DQM/SiPixelPhase1Summary/src/SiPixelPhase1Summary.cc
+++ b/DQM/SiPixelPhase1Summary/src/SiPixelPhase1Summary.cc
@@ -143,44 +143,100 @@ void SiPixelPhase1Summary::bookSummaries(DQMStore::IBooker& iBooker) {
   for (unsigned int i = 0; i < xAxisLabelsReduced_.size(); i++) {
     deadROCSummary->setBinLabel(i + 1, xAxisLabelsReduced_[i]);
   }
-  
+
   //New residual plots for the PXBarrel separated by inner and outer modules per layer
   iBooker.setCurrentFolder("PixelPhase1/Tracks/PXBarrel");
-  for (std::string layer:{"1","2","3","4"}){
-    residuals_["residual_mean_x_Inner_PXLayer_"+layer] = iBooker.book1D("residual_mean_x_Inner_PXLayer_"+layer, "Mean Track Residuals X Inner Modules for Layer "+layer+";mean(x_rec-x_pred)[cm]", 100, -0.1, 0.1);
-    residuals_["residual_mean_x_Outer_PXLayer_"+layer] = iBooker.book1D("residual_mean_x_Outer_PXLayer_"+layer, "Mean Track Residuals X Outer Modules for Layer "+layer+";mean(x_rec-x_pred)[cm]", 100, -0.1, 0.1);
-    residuals_["residual_mean_y_Inner_PXLayer_"+layer] = iBooker.book1D("residual_mean_y_Inner_PXLayer_"+layer, "Mean Track Residuals Y Inner Modules for Layer "+layer+";mean(x_rec-x_pred)[cm]", 100, -0.1, 0.1);
-    residuals_["residual_mean_y_Outer_PXLayer_"+layer] = iBooker.book1D("residual_mean_y_Outer_PXLayer_"+layer, "Mean Track Residuals Y Outer Modules for Layer "+layer+";mean(x_rec-x_pred)[cm]", 100, -0.1, 0.1);
-    
-    residuals_["residual_rms_x_Inner_PXLayer_"+layer] = iBooker.book1D("residual_rms_x_Inner_PXLayer_"+layer, "RMS of Track Residuals X Inner Modules for Layer "+layer+";rms(x_rec-x_pred)[cm]", 100, -0.1, 0.1);
-    residuals_["residual_rms_x_Outer_PXLayer_"+layer] = iBooker.book1D("residual_rms_x_Outer_PXLayer_"+layer, "RMS of Track Residuals X Outer Modules for Layer "+layer+";rms(x_rec-x_pred)[cm]", 100, -0.1, 0.1);
-    residuals_["residual_rms_y_Inner_PXLayer_"+layer] = iBooker.book1D("residual_rms_y_Inner_PXLayer_"+layer, "RMS of Track Residuals Y Inner Modules for Layer "+layer+";rms(x_rec-x_pred)[cm]", 100, -0.1, 0.1);
-    residuals_["residual_rms_y_Outer_PXLayer_"+layer] = iBooker.book1D("residual_rms_y_Outer_PXLayer_"+layer, "RMS of Track Residuals Y Outer Modules for Layer "+layer+";rms(x_rec-x_pred)[cm]", 100, -0.1, 0.1);
+  for (std::string layer : {"1", "2", "3", "4"}) {
+    residuals_["residual_mean_x_Inner_PXLayer_" + layer] =
+        iBooker.book1D("residual_mean_x_Inner_PXLayer_" + layer,
+                       "Mean Track Residuals X Inner Modules for Layer " + layer + ";mean(x_rec-x_pred)[cm]",
+                       100,
+                       -0.1,
+                       0.1);
+    residuals_["residual_mean_x_Outer_PXLayer_" + layer] =
+        iBooker.book1D("residual_mean_x_Outer_PXLayer_" + layer,
+                       "Mean Track Residuals X Outer Modules for Layer " + layer + ";mean(x_rec-x_pred)[cm]",
+                       100,
+                       -0.1,
+                       0.1);
+    residuals_["residual_mean_y_Inner_PXLayer_" + layer] =
+        iBooker.book1D("residual_mean_y_Inner_PXLayer_" + layer,
+                       "Mean Track Residuals Y Inner Modules for Layer " + layer + ";mean(x_rec-x_pred)[cm]",
+                       100,
+                       -0.1,
+                       0.1);
+    residuals_["residual_mean_y_Outer_PXLayer_" + layer] =
+        iBooker.book1D("residual_mean_y_Outer_PXLayer_" + layer,
+                       "Mean Track Residuals Y Outer Modules for Layer " + layer + ";mean(x_rec-x_pred)[cm]",
+                       100,
+                       -0.1,
+                       0.1);
+
+    residuals_["residual_rms_x_Inner_PXLayer_" + layer] =
+        iBooker.book1D("residual_rms_x_Inner_PXLayer_" + layer,
+                       "RMS of Track Residuals X Inner Modules for Layer " + layer + ";rms(x_rec-x_pred)[cm]",
+                       100,
+                       -0.1,
+                       0.1);
+    residuals_["residual_rms_x_Outer_PXLayer_" + layer] =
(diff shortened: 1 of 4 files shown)


jfernan2:
@dmeuser  could you please:
- add the Subsystem name (Tracker?) to the PR title for a proper identification?
- can you please add yourself with your git user name to the appropiate e-group?
https://twiki.cern.ch/twiki/bin/viewauth/CMS/DQMContacts#Tracker_SiStrips_and_Pixels
Thanks

DQM/SiPixelPhase1Summary/src/SiPixelPhase1Summary.cc
```diff
@@ -144,6 +144,100 @@ void SiPixelPhase1Summary::bookSummaries(DQMStore::IBooker& iBooker) {
...
+                       0.1);
+    residuals_["residual_mean_x_Outer_PXLayer_" + layer] =
+        iBooker.book1D("residual_mean_x_Outer_PXLayer_" + layer,
+                       "Mean Track Residuals X Outer Modules for Layer " + layer + ";mean(x_rec-x_pred)[cm]",
+                       100,
+                       -0.1,
+                       0.1);
+    residuals_["residual_mean_y_Inner_PXLayer_" + layer] =
+        iBooker.book1D("residual_mean_y_Inner_PXLayer_" + layer,
+                       "Mean Track Residuals Y Inner Modules for Layer " + layer + ";mean(x_rec-x_pred)[cm]",
```
jfernan2: Could you please change the description of this plot (and others below) to: 
Mean Track Residuals Y Inner Modules for Layer " + layer + ";mean(y_rec-y_pred)[cm

jfernan2:
please test

jfernan2:
@dmeuser  there are build errors

jfernan2:
please test

jfernan2:
please test

DQM/SiPixelPhase1Summary/src/SiPixelPhase1Summary.cc
```diff
@@ -144,6 +144,100 @@ void SiPixelPhase1Summary::bookSummaries(DQMStore::IBooker& iBooker) {
...
+                       "Mean Track Residuals Y Outer Modules for Layer " + layer + ";mean(y_rec-y_pred)[cm]",
+                       100,
+                       -0.1,
+                       0.1);
+
+    residuals_["residual_rms_x_Inner_PXLayer_" + layer] =
+        iBooker.book1D("residual_rms_x_Inner_PXLayer_" + layer,
+                       "RMS of Track Residuals X Inner Modules for Layer " + layer + ";rms(x_rec-x_pred)[cm]",
+                       100,
+                       -0.1,
```
mmusich, line 179: @dmeuser, why the RMS are binned in the negatives? It's a defined positive quantity, see plots at: https://tinyurl.com/qsb3yst

dmeuser: @mmusich, you are right, this makes no sense. I will change it and make a new commit?

mmusich: thx!

jfernan2:
please test

jfernan2:
@dmeuser  is there any technical reason to have null residuals for cosmic Run2 workflows like 7.3?
https://tinyurl.com/w3mk4gt
Thanks

dmeuser:
@jfernan2 the new plots are based on the TProfile2D  `PixelPhase1/Tracks/PXBarrel/residual_x_per_SignedModule_per_SignedLadder_PXLayer_*` and `PixelPhase1/Tracks/PXFoward/residual_x_per_PXDisk_per_SignedBladePanel_PXRing_*`. Are those produced for the workflow you mentioned? I do not find them in the file from your link.

jfernan2:
@dmuser The link just shows the differences added by this PR w.r.t. baseline.
The full output DQM root file from workflow 7.3 is on:
https://cmssdt.cern.ch/SDT/jenkins-artifacts/pull-request-integration/PR-841e65/5480/runTheMatrix-results/7.3_CosmicsSPLoose_UP18+CosmicsSPLoose_UP18+DIGICOS_UP18+RECOCOS_UP18+ALCACOS_UP18+HARVESTCOS_UP18/
I don't see any complaint message in the logs.
If those TProfile2Ds are not there, you should add some prevention to the class or disable it for those workflows where it should not run

mmusich:
@jfernan2 these plots are empty, simply because of the acceptance of the pixel detector in cosmics...

jfernan2:
Thanks for the info @mmusich 
Then I wonder what does this DQM sequence in cosmics for Pixel:
https://github.com/cms-sw/cmssw/blob/02d4198c0b6615287fd88e9a8ff650aea994412e/DQM/SiPixelMonitorTrack/python/SiPixelMonitorTrack_Cosmics_cfi.py#L4
which is run here. 
I am fine with the plots, just trying to understand their origin and purpouse

mmusich:
@jfernan2 in a normal cosmics run with thousands of evnts we have many pixel hits. On the other hand on a relval with 10 events it's very unlikely to have one...

mmusich:
Well, it's a combination of the two... but fine.

mmusich:
@dmeuser on the other hand @jfernan2 has a good point. If the original plot is empty, the DMR plot should rather not be filled than show a 0. This is misleading in terms of physical interpretation.

dmeuser:
@mmusich sounds reasonable. I will change to code to only fill the new histogram  if the entry for a given module in the TProfile2D is not empty.

jfernan2:
please test

DQM/SiPixelPhase1Summary/src/SiPixelPhase1Summary.cc
```diff
@@ -443,29 +460,31 @@ void SiPixelPhase1Summary::fillSummaries(DQMStore::IBooker& iBooker, DQMStore::I
           continue;  //Middle bins of y axis is empty
         if (i == (me_x->getNbinsY() / 2) + 1)
           continue;
+        if (me_x->getBinEntries(j + (me_x->getNbinsX() + 2) * i) < 30)  //Fill only if number of hits is above 30
+          continue;
```
jfernan2, line 464: Is this cut expected to not create the residual histograms in the case of reduced occupancy/low acceptance?
I am asking since last test still shows histograms with null residual for Cosmic wf 7.3
https://tinyurl.com/sm5fsuu

arossi83:
@jfernan2 the cut  will avoid to fill the histogram added by this PR if the number of hits in a single module is <30. Looking at the link you pointed out this works, histograms are in place but they are not filled since there are few hits in each module.

jfernan2:
I see @arossi83 but creating empty histograms is far from efficient, couldn't this be improved somehow?

DQM/SiPixelPhase1Summary/src/SiPixelPhase1Summary.cc
mmusich: @dmeuser, in the interest of readability, can this "30" be a named constant rather than a magic number?

mmusich:
@jfernan2 the histogram is empty only because that's tested with a relval with few events. Rest assured that in operations this histogram won't be empty.

DQM/SiPixelPhase1Summary/src/SiPixelPhase1Summary.cc
mmusich: @schneiml @jfernan2, may I ask if there is a method of `MonitorElement` to get the bin index of a 2D histogram to avoid this kludgy `binx+(TOTbinX+2)*biny`, something like https://root.cern.ch/doc/master/classTH2.html#a9d47a53d1b2499475542b70a183c5c1c?

DQM/SiPixelPhase1Summary/src/SiPixelPhase1Summary.cc
```diff
@@ -283,6 +383,113 @@ void SiPixelPhase1Summary::fillSummaries(DQMStore::IBooker& iBooker, DQMStore::I
...
+        iGetter.get("PixelPhase1/Tracks/PXBarrel/residual_y_per_SignedModule_per_SignedLadder_PXLayer_" + layer);
+
+    bool isLayer_4 = layer == "4";  //Order of inner and outer ladders is reversed for Layer 4
+
+    for (int i = 1; i <= me_x->getNbinsY(); i++) {
+      if (i == (me_x->getNbinsY() / 2 + 1))
+        continue;  //Middle bin of y axis is empty
+
+      if (i <= me_x->getNbinsY() / 2) {
+        if ((i % 2 == 0 && !isLayer_4) || (i % 2 == 1 && isLayer_4)) {
```
mmusich, line 401: @dmeuser I guess it will be very difficult for someone reading this code in 5 years to understand why we do this here. Can an ancillary variable be added here like it is done at https://github.com/cms-sw/cmssw/blob/master/DQM/SiPixelPhase1Track/plugins/SiPixelPhase1TrackClusters.cc#L181-L184 to make it clear the ladder order in BPix L4 is swapped?

DQM/SiPixelPhase1Summary/src/SiPixelPhase1Summary.cc
schneiml: @mmusich @dmeuser It seems we have `getBinContent` and `getBinError` in 2D, but not `getBinEntries`. Feel free to add an overload here: https://github.com/cms-sw/cmssw/blob/master/DQMServices/Core/src/MonitorElement.cc#L643

dmeuser: @schneiml I am currently trying to add the method to the MonitorElement. Compiling the code does work fine, but when running the Harvesting I always get something like this:

schneiml: There is clearly some level of memory corruption there. Not much else that I can tell without the code.

jfernan2:
please test

DQM/SiPixelPhase1Summary/src/SiPixelPhase1Summary.cc
```diff
@@ -385,22 +385,24 @@ void SiPixelPhase1Summary::fillSummaries(DQMStore::IBooker& iBooker, DQMStore::I
 
   //Fill additional residuals plots
   //PXBarrel
+
+  int minHits = 30;  //Miniminal number of hits needed for module to be filled in histograms
```
mmusich, line 389: this could be `constexpr`

dmeuser: `static constexpr int` ?

DQM/SiPixelPhase1Summary/src/SiPixelPhase1Summary.cc
dmeuser: I did not take the dependencies into account, my fault.

jfernan2:
please test
Triggering again the test to get the last commit into the check

jfernan2:
+1

silviodonato:
+1

PR #29363, opened 2020-04-01
Support of condition updates per-lumisection

ggovi:
#### PR description:

Adding support of condition updates with lumisection granularity. Targeted for HLT workflow. 
The updates will be forced by an 'artificial' fixed lenght of the IOV, determined by configuration and set when the since value is identified for a given target event time.

#### PR validation:

A full test workflow for this case has been added. The existing unit tests are running.

### DIFF
--- a/CondCore/ESSources/interface/DataProxy.h
+++ b/CondCore/ESSources/interface/DataProxy.h
@@ -105,7 +105,7 @@ namespace cond {
     void loadTag(std::string const& tag, boost::posix_time::ptime const& snapshotTime);
     void reload();
 
-    ValidityInterval setIntervalFor(Time_t target);
+    ValidityInterval setIntervalFor(Time_t target, Time_t defaultIovSize);
     TimeType timeType() const { return m_iovProxy.tagInfo().timeType; }
 
   private:
--- a/CondCore/ESSources/plugins/CondDBESSource.cc
+++ b/CondCore/ESSources/plugins/CondDBESSource.cc
@@ -347,14 +347,17 @@ void CondDBESSource::setIntervalFor(const EventSetupRecordKey& iKey,
   }
   //}
   cond::Time_t lastTime = m_lastRun;
-  auto iR = m_refreshTimeForRecord.find(recordname);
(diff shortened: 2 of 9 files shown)


ggovi:
@wddgit can you please have a look? Feedback wellcome

ggovi:
please test

wddgit:
I reviewed it. It mostly looks good to me. Some comments below.

One thing I noticed is that this could cause multiple Sessions to be alive for the same connection string. Prior to that there was at most one per Run and that meant one in existence at a time because the Framework does not execute Runs concurrently yet. As far as I know that is OK, but I do not know much about what is going on internally in the Session objects. The PayloadProxy is holding a shared_ptr, so there should not be any lifetime issues with the Session objects. And the mutex's will still prevent concurrency issues as before. Probably all is OK.

When refreshThisRecord is true, then  the various policies seem to be ignored. For example, REFRESH_ALWAYS will be ignored even if it is true. If that is the behavior you want, then that is also OK. It is a reasonable behavior.

When m_policy equals REFRESH_EACH_RUN or RECONNECT_EACH_RUN, then it seems to follow the logic for refreshThisRecord even if refreshThisRecord is false. So for example if in the input the run number decreased, then the code would not set doRefresh true even if REFRESH_EACH_RUN were true. Seems strange.

```
if (lastTime - lastRecordRun >= minDiffTime) {
```
Maybe instead:
```
if ((refreshThisRecord &&lastTime - lastRecordRun >= minDiffTime) ||
     (!refreshThisRecord && lastTime != lastRecordRun)) {
```
or something similar to that. That is all I noticed in the time I had available to look at it.

ggovi:
@wddgit Thanks, useful comment. However, both the REFRESH_EACH_RUN or RECONNECT_EACH_RUN have been introduced for HLT ( the only use case to my knowledge ), where the run sequence will be only increasing anyhow. 
To be 'on the safe side' I have anyway relaxed the condition for them.

ggovi:
please test

ggovi:
please test

ggovi:
+1

pohsun:
+1

silviodonato:
+1

PR #29366, opened 2020-04-01
Solve all overlaps detected in Phase I Tracker

ghugo83:
Solve all overlaps detected in Phase I Tracker:
- **TEC petals with optical connectors**:
75704087-b0cfc180-5c86-11ea-8446-11731be4b2b3
- **TEC services channel with grounding axis**:
channel_axgrounding

In general, I would think it is good practice to avoid having volumes 'sharp' at the same boundary in the XML description, since with default meshing / rounding errors, they can be seen / actually become overlaps.
Directly using an epsilon in the XML description in this kind of situations should solve a lot of headaches, at no cost for the geo and Material Budget.

**PR tests:**
0 overlap with G4 AND Fireworks overlap-checking tools, for both Reference and Migrated 2021 Scenario.


cvuosalo:
@cmsbuild please test

civanch:
@ghugo83 , this may be a useful fix if an overlap is detected. At the same time, making gaps between volumes is not a good practice, because it is obviously slow down the simulation - all particles will make an extra step between such volumes. CPU in Geant4 is always proportional to number of steps independently on how long they are.

ghugo83:
@civanch Thanks a lot for your comment.
Though for solving overlaps, it is needed here. 

In general, speaking of the number of steps, why are the cables modeled in such fine detail? Would phi sections of cylinder, averaging out all services contributions, not suffice? This would instead reduce by several orders of magnitude the number of steps.

civanch:
@ghugo83 , a general approach in CMS, ATLAS, and some others is to describe as accurate as possible real geometry. If simulation is slow due overdoing this in some sub-detector, than a possible short-cut is developed. It is difficult to do in the inverse order.

civanch:
@cvuosalo , @vargasa , can you agree with this fix?

cvuosalo:
+1

silviodonato:
+1

PR #29372, opened 2020-04-01
DQM: Added unit test support for most online DQM clients

andrius-k:
#### PR description:

Added a dedicated input source for the unit tests (`unittestinputsource_cfi`) that selects only last 100 events of 1st and 2nd lumisections to make sure that the tests can run fast yet encountering a lumi transition. Each client had to be updated to make use of the newly added input source.

Most clients are tested however, some require more thinking due to reading non-standard streams or GT issues.

#### PR validation:

PR was validated locally.

### DIFF
--- a/DQM/Integration/python/clients/sistrip_dqm_sourceclient-live_cfg.py
+++ b/DQM/Integration/python/clients/sistrip_dqm_sourceclient-live_cfg.py
@@ -1,7 +1,7 @@
 from __future__ import print_function
 import FWCore.ParameterSet.Config as cms
 
-
+import sys
 from Configuration.Eras.Era_Run2_2018_pp_on_AA_cff import Run2_2018_pp_on_AA
 process = cms.Process("SiStrpDQMLive", Run2_2018_pp_on_AA)
 
@@ -15,6 +15,12 @@ process.MessageLogger = cms.Service("MessageLogger",
 )
 
 live=True
+unitTest=False
+
+if 'unitTest=True' in sys.argv:
+    live=False
+    unitTest=True
+
 # uncomment for running on lxplus
 #live=False
 offlineTesting=not live
@@ -24,7 +30,9 @@ offlineTesting=not live
 # Event Source
 #-----------------------------
 # for live online DQM in P5
-if (live):
+if (unitTest):
+    process.load("DQM.Integration.config.unittestinputsource_cfi")
+elif (live):
     process.load("DQM.Integration.config.inputsource_cfi")
 # for testing in lxplus
 elif(offlineTesting):
--- a/DQM/Integration/python/clients/beam_dqm_sourceclient-live_cfg.py
+++ b/DQM/Integration/python/clients/beam_dqm_sourceclient-live_cfg.py
@@ -3,6 +3,7 @@ import FWCore.ParameterSet.Config as cms
 
 #from Configuration.Eras.Era_Run2_2018_cff import Run2_2018
 #process = cms.Process("BeamMonitor", Run2_2018) FIXME
+import sys
 from Configuration.Eras.Era_Run2_2018_pp_on_AA_cff import Run2_2018_pp_on_AA
 process = cms.Process("BeamMonitor", Run2_2018_pp_on_AA)
 
@@ -17,10 +18,17 @@ process.MessageLogger = cms.Service("MessageLogger",
 
 # switch
 live = True # FIXME
+unitTest = False
+
+if 'unitTest=True' in sys.argv:
+    live=False
+    unitTest=True
 
 #---------------
 # Input sources
-if (live):
+if unitTest:
+    process.load("DQM.Integration.config.unittestinputsource_cfi")
+elif live:
     process.load("DQM.Integration.config.inputsource_cfi")
 else:
     process.load("DQM.Integration.config.fileinputsource_cfi")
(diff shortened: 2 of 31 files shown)


andrius-k:
please test

andrius-k:
please test

andrius-k:
please test

andrius-k:
please test

DQM/Integration/python/clients/sistrip_dqm_sourceclient-live_cfg.py
```diff
@@ -1,7 +1,7 @@
 from __future__ import print_function
 import FWCore.ParameterSet.Config as cms
 
-
+import sys
 from Configuration.Eras.Era_Run2_2018_pp_on_AA_cff import Run2_2018_pp_on_AA
 process = cms.Process("SiStrpDQMLive", Run2_2018_pp_on_AA)
```
mmusich, line 6: Hi @andrius-k I would just like to point out that as you set a 2020 run as input for the unit tests at:
https://github.com/cms-sw/cmssw/pull/29372/files#diff-6e50ae85a1069d2862274e07c8ba9d81R28

```bash
Run_number  Start_time                  End_time                    Start_IOV            End_IOV               
----------  --------------------------  --------------------------  -------------------  -------------------  
334393      2020-01-22 21:52:41.170000  2020-01-23 06:35:04.284000  6784888519006355456  6785023136166313984   
```

the cmssw era in Strips, Pixel (and possibly many others) online clients is not configured correctly to cope with that run, so I can't vouch for the sensibility of what comes out of the unit test.
@sroychow @arossi83

andrius-k: Hi @mmusich, this run and dataset combination is coming from the only `runTheMatrix` workflow of 2020, so I thought clients would be configured to run on the recent data. Which data would you suggest to use for the test to be more valuable?

mmusich: @andrius-k given the nature of the online client, I agree that using the latest data is the right move.  I'd rather migrate the clients to the run-3 era (though maybe each subsystem should take care of that). For Tracker we can do it in a separate PR.

andrius-k: @mmusich but all tested clients were running in the Online during the MWGR that this data was taken at so I'm not sure I understand what migration do you have in mind. In my opinion, Online clients should be optimised to be used for processing the most recent data and we're planning to change the input data to these tests regularly to keep them up to date.

mmusich: @andrius-k the MWGR was taken with CMSSW_11_0_X release cycle. This PR targets the master branch. The underlying code has changed in the meanwhile.

arossi83: @andrius-k if I remember correctly (@fioriNTU can correct me) in the past Online clients were tested directly at P5 in a parallel test system, sometimes we had PR which had been never merged in any release but they were used in Online (so they were merged manually at P5) for short time period (i.e special runs or HI runs). Unit test for Online client here are completely new and we need to take care of keeping updated the Online client on the base of the unit test configuration, i.e. Era is an example as pointed out by @mmusich but also other things can affect unit test output, this is completely fine with me but we need to understand how to manage this, if each subdet need to be updated his own Client in order to be aligned with the unit test or if this could be done centrally (as was in the past at P5, the DQM core person responsible for online keep all the client running with the correct era)

andrius-k: Ah I see what you mean @mmusich and indeed you have a point. However, given that according to the back-port policy, generally speaking, everything that is running in the online is in the master too, I think that testing the master is is the best approach. After all, the tests are supposed to verify the future not the past.

In a situation where any recent data is already too old for a fast changing client, the test can be manually turned off for a specific client, or changes can be force merged. Keep in mind that in this situation we can't verify the client in the DQM playback system as well, so commissioning it is a bit risky anyways.

fioriNTU: @andrius-k actually the point thet @mmusich raise is a good one, we have to think better about how to setup this kind of tests. I have no solution at the moment.

mmusich: > However, given that according to the back-port policy, generally speaking, everything that is running in the online is in the master too.

unfortunately as @arossi83 said, that's not always the case. Specifically here, we made some changes in 11.1.X to the code underlying the online strips and pixel client which, when tested on a 2020 run, make sense only if tested with the Run3 era, which we'll surely use at some point. That's all what I want to say about it.

andrius-k:
please test

andrius-k:
please test with #29484

andrius-k:
please test with #29484

andrius-k:
please test with #29493

andrius-k:
please abort

andrius-k:
please test with #29493,#29484

andrius-k:
please test with #29493,#29484,#29506

andrius-k:
ping

smuzaffar:
please test with #29493,#29484,#29506
lts retry

silviodonato:
+1

silviodonato:
The new DQM tests are failing in PY3 (see #29532)

PR #29374, opened 2020-04-01
Restrict timestamps in PFCandidates to reliable cases

bendavid:
Cleanup the timestamps in the PFCandidates in two ways:

1.  Apply track-MTD quality matching criteria, using a cut on the MVA>0.5 corresponding to what was added in https://github.com/cms-sw/cmssw/pull/28815
If this criteria is not satisfied, the PFCandidate is not assigned a time (ie treated as though no timing information is available.)

2.  Don't propagate timestamps from PFClusters to neutrals in the SimPFProducer (used for the realistic simcluster-based HGCal reconstruction) since this information is not reliable/consistently filled.  (Timing for neutral PFCandidates in HGCal is being revisited as part of migration to TICL-based reconstruction for PF.)

These changes should simplify downstream use of MTD timing information by POGs for phase 2 offline and HLT.

Since the PFCandidate timestamps are not used downstream yet in any of the standard workflows, the expected change is only in the PFCandidate and PackedCandidate times and timeErrors, with a larger fraction of candidates having no timestamp assigned (and timeError = -1.0)

@hatakeyamak @pmeridian @parbol @rovere @felicepantaleo

### DIFF
--- a/RecoParticleFlow/PFSimProducer/plugins/SimPFProducer.cc
+++ b/RecoParticleFlow/PFSimProducer/plugins/SimPFProducer.cc
@@ -68,7 +68,7 @@ private:
   const bool useTiming_;
   const bool useTimingQuality_;
   const double timingQualityThreshold_;
-  
+
   // inputs
   const edm::EDGetTokenT<edm::View<reco::PFRecTrack>> pfRecTracks_;
   const edm::EDGetTokenT<edm::View<reco::Track>> tracks_;
@@ -113,8 +113,8 @@ SimPFProducer::SimPFProducer(const edm::ParameterSet& conf)
                              ? consumes<edm::ValueMap<float>>(conf.getParameter<edm::InputTag>("trackTimeErrorMap"))
                              : edm::EDGetTokenT<edm::ValueMap<float>>()),
       srcTrackTimeQuality_(useTimingQuality_
-                             ? consumes<edm::ValueMap<float>>(conf.getParameter<edm::InputTag>("trackTimeQualityMap"))
-                             : edm::EDGetTokenT<edm::ValueMap<float>>()),
+                               ? consumes<edm::ValueMap<float>>(conf.getParameter<edm::InputTag>("trackTimeQualityMap"))
+                               : edm::EDGetTokenT<edm::ValueMap<float>>()),
       srcGsfTrackTime_(useTiming_
                            ? consumes<edm::ValueMap<float>>(conf.getParameter<edm::InputTag>("gsfTrackTimeValueMap"))
                            : edm::EDGetTokenT<edm::ValueMap<float>>()),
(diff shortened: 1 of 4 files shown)


slava77:
@cmsbuild please test

perrotta:
@cmsbuild please test

perrotta:
Test restarted, because of the several differences in the logs with respect to the baseline. 

They were due to messages **in the baseline logs** like
```
%MSG-e ExcessiveTime:  SiPixelClusterProducer:siPixelClustersPreSplitting  02-Apr-2020 01:28:07 CEST Run: 1 Event: 5
ExcessiveTime: Module used 1174.37 seconds of time which exceeds the error threshold configured in the Timing Service of 600 seconds.
%MSG
```

Those messages seemed restricted to the baseline used for comparisons in the first tests (```CMSSW_11_1_X_2020-04-01-2300/slc7_amd64_gcc820```), so let retry testing with a different baseline, where they should hopefully have disappeared

perrotta:
assign upgrade

kpedro88:
+upgrade

pfcand time
increase at 0, decrease at -30 (actually -100), as expected from PR description

RecoParticleFlow/PFSimProducer/plugins/SimPFProducer.cc
```diff
@@ -99,6 +101,8 @@ SimPFProducer::SimPFProducer(const edm::ParameterSet& conf)
       neutralEMThreshold_(conf.getParameter<double>("neutralEMThreshold")),
       neutralHADThreshold_(conf.getParameter<double>("neutralHADThreshold")),
       useTiming_(conf.existsAs<edm::InputTag>("trackTimeValueMap")),
+      useTimingQuality_(conf.existsAs<edm::InputTag>("trackTimeQualityMap")),
```
perrotta, line 104: Please add a fillDescriptions(...) method for this producer, so that you can get rid of this esistAs check (among other things...)

perrotta: Indeed, if so some modification for the  ```useTiming_``` and ```useTimingQuality_``` bool's should be allowed.
But, by the way: is it even foreseen that such "trackTimeValueMap" and "trackTimeQualityMap" can not exist? Couldn't we give them for granted, and in that case those two bool's are not even needed any more?

bendavid: They don't exist in the "fastsim" implementation with smeared sim times, and there are still eras defined for this (though I'm not sure if they're still actively used anywhere).

perrotta: @bendavid I agree that this PR is not the place where to fix the issue of missing fillDescription in that producer, taking also into account that those functionalities are missing in all producers in RecoParticleFlow/PFSimProducer.

Still, I believe that all producers in that package should profit of the Configuration validation. Issue #29449 was created to keep note of it.

perrotta:
+1
- The code changes implement the cleanup of time stamps in PFCandidates from MTD as described in the PR desription
- Jenkins tests only show changes in Phase2 related workflows, and those changes goes in the expected direction (with only timestamps affected and no propagations on other derived quantities, as expected)

silviodonato:
+1

PR #29044, opened 2020-02-26
Adding parameters needed for PUjetID 94X and 102X (backport to 102X)

alefisico:
#### PR description:

This is a backport from #28984 meant for nanoAODv7.

FYI: @ahinzmann @lathomas @camclean @peruzzim @nurfikri89

PR #28984, opened 2020-02-18
Adding parameters needed for PUjetID 94X and 102X v2

alefisico:
#### PR description:

 - Including the files and parameters needed for the latest PUjetID training for 94X and 102X samples.
 - Once merged, we need to backport this PR to the corresponding CMSSW release for the latest nanoAOD.
 - There is also a PR for the new weights for PUJetID cms-data/RecoJets-JetProducers#10 (comment)

This PR superseeds #28827.
Expected changes with this PR: https://indico.cern.ch/event/860457/#17-pujetid-training-and-valida 

FYI: @ahinzmann @lathomas  @camclean

PR #29376, opened 2020-04-02
fix miniaod_devel modifier use for pileupJetId.algos

slava77:
a problem was observed when testing re-mini and nano workflows with run2_miniAOD_devel: there  is no support in process modifiers ```toReplaceWith``` for a VPset.
This PR fixes the problem.

Tested on 136.8311 and 1325.7 modified to also use run2_miniAOD_devel

This is a fix to the backport of #28984 made in #29044


slava77:
@cmsbuild please test

slava77:
+1

for https://github.com/cms-sw/cmssw/pull/29376	6c362e7 
- since this affects only a setup with ```run2_miniAOD_devel``` modifier, the fix is tested locally
- regular jenkins tests pass and comparisons show no differences

silviodonato:
+1

PR #29395, opened 2020-04-04
added unit test to check for clang-tidy changes

smuzaffar:
This PR runs clang-tidy and make sure the generated yaml files format is not changed. This could happen with new version of llvm. As it happened with llvm 9.0.1 but our build system did not noticed it. This broke clang-tidy in CMSSW_11_1_X (with LLVM 9). https://github.com/cms-sw/cmsdist/pull/5721 fixes the build rules to work with llvm 9.0.1 but we think it would be help help top have a unit test which makes sure that we catch such issues at the time of llvm integration.

Local build and test run successfully.


smuzaffar:
please test

smuzaffar:
please test

smuzaffar:
+core
comparison failures has nothing to do with this PR which only adds an extra unit test

silviodonato:
+1

PR #29399, opened 2020-04-06
vavilov parameters kappa fix

OzAmram:
A small PR to fix the issue reported [here](https://github.com/cms-sw/cmssw/issues/29251). The issue was caused by rounding differences between float and doubles. Namely that the value of kappa when on the edge of its allowed range as a float (0.01f) was less than the double value (0.01) used to check for errors. This fix simply adds a small amount to Kappa if it is exactly on the edge. 

Tested by following the procedure outlined [here](https://github.com/cms-sw/cmssw/issues/29251#issuecomment-607706773) and checking that this does indeed fix the error.

### DIFF
--- a/CondFormats/SiPixelTransient/src/SiPixelTemplate.cc
+++ b/CondFormats/SiPixelTransient/src/SiPixelTemplate.cc
@@ -4056,9 +4056,11 @@ void SiPixelTemplate::vavilov_pars(double& mpv, double& sigma, double& kappa)
 
   // Copy to parameter list
 
-  //avoid rounding difference between floats and doubles causing issues later
-  if (kappavav_ == 0.01f)
+  //Avoid rounding difference between floats and doubles causing issues later
+  if (kappavav_ <= 0.01f) {
+    LOGERROR("SiPixelTemplate") << "Vavilov kappa value is " << kappavav_ << " changing it to be above 0.01" << ENDL;
     kappavav_ = 0.01f + 0.0000001f;
+  }
 
   mpv = (double)mpvvav_;
   sigma = (double)sigmavav_;
--- a/CondFormats/SiPixelTransient/src/SiPixelTemplate.cc
+++ b/CondFormats/SiPixelTransient/src/SiPixelTemplate.cc
@@ -4057,7 +4057,8 @@ void SiPixelTemplate::vavilov_pars(double& mpv, double& sigma, double& kappa)
   // Copy to parameter list
 
   //avoid rounding difference between floats and doubles causing issues later
-  if(kappavav_ == 0.01f) kappavav_ = 0.01f + 0.0000001f;
+  if (kappavav_ == 0.01f)
+    kappavav_ = 0.01f + 0.0000001f;
 
   mpv = (double)mpvvav_;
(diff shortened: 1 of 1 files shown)


slava77:

this should be addressed before full PR tests can start

OzAmram:
Oops sorry I forgot to run the code format before pushing. Should be ok now

slava77:
@cmsbuild please test workflow 13034.0

perrotta:
(RFC, or maybe just a suggestion)

I am normally a bit scared when looking for exact equalities of floating point like quantities. Wouldn't it be safer in this case to lower bound the ```double``` passed out by the method, e.g.

```C++
    constexpr double kappamin = 0.01+ 0.0000001;
    kappa = std:max((double)kappavav_, kappamin);
```
or even (trusting the fact that the passed parameter is not manipulated any more bofore asserting on it in the caller):
```C++
    kappa = std::max((double)kappavav_, 0.01);
```

OzAmram:
I guess the question is whether we still want to 'check' for whether something went wrong in the assert or not. If we just set to be >=0.01 then we won't throw an error in the case where there actually is something wrong with the templates and kappa is < 0.01. 

My thought with the equality is that we only want to 'fix' the case where the error is being thrown due to the float vs double rounding (which only happens when its 0.01f) but still let the error happen in other cases.

slava77:
> I guess the question is whether we still want to 'check' for whether something went wrong in the assert or not.

that was my interpretation of the logic as well.
OTOH, shouldn't there be a way to test the template quality when it's created or with some unit test instead of run time? So that there is no need to throw after all.

OzAmram:
Yes we have test we run on the templates when we make them but it is hard to test every corner of phase space. Overall I think its very unlikely this particular issue could arise because kappa is constrained to >= 0.01 when we make the templates. So I would be ok with something similar to what @perrotta proposed above with a suitable warning.

perrotta:
> Yes we have test we run on the templates when we make them but it is hard to test every corner of phase space. Overall I think its very unlikely this particular issue could arise because kappa is constrained to >= 0.01 when we make the templates. So I would be ok with something similar to what @perrotta proposed above with a suitable warning.

@OzAmram please go ahead with it, warning included

OzAmram:
Ok I just pushed a new commit

perrotta:
please test

perrotta:
Still not clear to me why you do not apply the boundary limitation at L4067 to the returned ```double kappa```, instead than to the intermediate ```float kappavav_```. Applying it to the returned ```double``` value would avoid the remote possibility of non continuos returned value in the (indeed negligibly tiny) interval between ```0.01f ``` and ```0.01f + 0.0000001f ```.

This is indoubtly a ultra-minor point: once tests succeed this PR will be considered ready to be merged by reco.

perrotta:
+1
- It fixes the rare issue posted in #29251 by bounding the values of the kappa vavilov parameter to their physically allowed values (they could have been exceeded by rounding approximations in the float to double conversion)
- Jenkins tests pass and show no differences in reco outputs

ggovi:
+1

silviodonato:
+1

PR #29387, opened 2020-04-03
[DD4hep] Expose Full Geo History on Request

ianna:
#### PR description:

* As discussed with @fabiocos, a numbering scheme may need to use extra information encoded in a node name. 
* The new `geoHistory()` implementation gives direct access to the nodes in a reverse order:
```
Tracker[1]/CMSE[1]/OCMS[1]/
[1]/[1]/[1]/
0/0/0/
```
Here is an example on how to extract such information without the need to parse the path to the node:

```c++
cms::DDFilter filter("TrackingMaterialGroup", {i.data(), i.size()});
cms::DDFilteredView fv(*cpv, filter);
bool firstChild = fv.firstChild();
const std::vector<const cms::Node*> nodes = fv.geoHistory();
for(const auto& n : nodes) {
  log << n->GetVolume()->GetName() << "[" << n->GetNumber() << "]/";
 }
const auto nodeCopyNums = fv.copyNumbers();
for(const auto& c : nodeCopyNums) {
  log << "[" << c << "]/";
}
const auto navPos = fv.navPos();
for(const auto& p : navPos) {
  log << p << "/";
}
```

#### if this PR is a backport please specify the original PR and why you need to backport that PR:

Before submitting your pull requests, make sure you followed this checklist:
- verify that the PR is really intended for the chosen branch
- verify that changes follow [CMS Naming, Coding, And Style Rules](http://cms-sw.github.io/cms_coding_rules.html)
- verify that the PR passes the basic test procedure suggested in the [CMSSW PR instructions](https://cms-sw.github.io/PRWorkflow.html)

PR #29407, opened 2020-04-07
MTD Geometry: DD4hep migration, preliminary update and cleaning of ideal geometry test

fabiocos:
#### PR description:

This PR updates DD4hep scenarios D50 and D53 according to the most recent updates, and refresh the ```MTDCommonData``` tests. With the opportunity a few constants for the scenario D53 are added to ```ETLDetId``` and ```ETLNumberingScheme```, and the MtdSD test verbosity is improved. While preparing the migration of ```MTDNumberingBuilder``` (to come in a next PR), I realized that even the old basic DD4hep geometry test had occasional problems with the printed strings, but no sign of issue with the computed DetIds or positions (which makes me think this is mostly a printout issue, although I do not fully understand it).
I have refreshed and simplified the test to avoid useless code duplication and multiple output streams, getting the same results.

#### PR validation:

The standalone ```MTDCommonData``` tests run and produce consistent output.


fabiocos:
@cvuosalo @ianna this PR (and the next one) would benefit from #29387 for the ```geoHistory``` computation

fabiocos:
please test

cvuosalo:
+1

kpedro88:
+upgrade

civanch:
+1

@fabiocos , since some moment, we are trying use LogVerbatim instead of LogInfo, because in the last variant log files are polluted by date and time at each printout, which make a problem to use diff/tkdiff

fabiocos:
@civanch in general I agree about ```LogVerbatim``` vs ```LogInfo```. Anyway 1) several statements are under the debug flag ```EDM_ML_DEBUG``` ,  so no affecting normally anything, and the test code is never run in default workflows, but always in standalone tests where I do want to have the output. As you see I have tried to avoid multiple streams, i.e. both ```MessageLogger``` and external files as in the DDD tests, because I notice some issues with printing names coming from DD4hep that I am not fully understanding, but I am afraid might be made worse by multiple streams. And in that case I use different labels of messages to collect them in files I can directly compare with the old ones, I have a python script to parse MessageLogger output that can selectively do this (may be I should make it available).

silviodonato:

fabiocos:
@silviodonato the problems in the IB need to be solved, before a meaningful test for this PR can be made

silviodonato:
+1

`23434.1001_TTbar_14TeV+RecoFullGlobalPU_TestOldDigi_2026D49PU+HARVESTFullGlobalPU_2026D49PU Step0-PASSED Step1-PASSED  - time date Thu Apr  9 23:29:03 2020-date Thu Apr  9 22:30:05 2020; exit: 0 0`

PR #29410, opened 2020-04-07
Don't include forward headers in cc files in RecoParticleFlow/PFProducer package

guitargeek:
#### PR description:

These are just some changes which were originally included in https://github.com/cms-sw/cmssw/pull/29375 (**PFEGammaProducer to global producer**), but they were deemed too controversial to be secondary changes in a PR with a different primary goal.

Now I'm coming back to this argument that there should never be the need to include forward headers (these `*Fwd.h` files in the DataFormates for example) in `.cc` sources; trying to set a precedent. I'll attept to elaborate the point once more:

1. Besides forward declarations, these forward headers usually define a few `typedef`s such that the DataFormats can be used efficiently in other header files without including the full interface
2. However, if you include a DataFormat in a `.cc` source, you do so with the intention of actually using it's interface, while still relying on the `typedef`s. So right now it's a very common pattern to include both the forward header and and full header in `.cc` sources.
3. For brevity and to ensure they `typedef`s are always consistent, the full headers should include the forward headers themselves.
4. In that case, the `.cc` sources should never have to include forward headers anymore.

Any thoughts on this? What are the arguments in favour of strictly separating the header files? Thanks!

#### PR validation:

CMSSW compiles.

No backport intended.


slava77:
@cmsbuild please test

perrotta:
@Dr15Jones @makortel @smuzaffar FYI

perrotta:
please test
(to remove spurious differences from the comparisons)

perrotta:
+1
- Technical cleaning
- Jenkins tests pass
- Possible circular includes are avoided

silviodonato:
+1

PR #29440, opened 2020-04-09
Run3-gem34 Add a GE21 version with 16 eta partitions

bsunanda:
#### PR description:

Add a GE21 version with 16 eta partitions

#### PR validation:

Tested a scenario by visual inspection

Nothing special


dildick:
@bsunanda Can you change https://github.com/cms-sw/cmssw/blob/master/DataFormats/MuonDetId/interface/GEMDetId.h#L33 to ```static constexpr int32_t maxRollId = 16;```? That way we can consider eta partition 0 to be the chamber as in GE1/1 and ME0. Thanks.

bsunanda:
@dildick I am doing that and adding to this PR

bsunanda:
@cmsbuild Please test

Geometry/MuonCommonData/python/testExtendedGeometry2026D54XML_cfi.py
```diff
@@ -0,0 +1,129 @@
...
+        'Geometry/MuonCommonData/data/mb4/2015/v2/mb4.xml',
+        'Geometry/MuonCommonData/data/mb4Shield/2021/v1/mb4Shield.xml',
+        'Geometry/MuonCommonData/data/muonYoke/2021/v2/muonYoke.xml',
+        'Geometry/MuonCommonData/data/csc/2021/v1/csc.xml',
+        'Geometry/MuonCommonData/data/mfshield/2017/v1/mfshield.xml',
+        'Geometry/MuonCommonData/data/mf/2026/v2/mf.xml',
+        'Geometry/MuonCommonData/data/rpcf/2026/v2/rpcf.xml',
+        'Geometry/MuonCommonData/data/gemf/TDR_BaseLine/gemf.xml',
+        'Geometry/MuonCommonData/data/gem11/TDR_BaseLine/gem11.xml',
+        'Geometry/MuonCommonData/data/gem21/TDR_Eta21/gem21.xml',
```
jshlee, line 81: i think TDR_Eta21 is suppose to be TDR_Eta16

bsunanda: Thanks Jason. Corrected

bsunanda:
@cmsbuild Please test

jshlee:
@bsunanda - I'm seeing roll 17...
region 1 ring 1 station 2 layer 2 chamber 10 roll 17
I think we have a problem in gemDetId

jshlee:
> @bsunanda - I'm seeing roll 17...
> region 1 ring 1 station 2 layer 2 chamber 10 roll 17
> I think we have a problem in gemDetId

@dildick @watson-ij 
I think this is from the trigger. Sven, could you take a look at this.
CSCTriggerPrimitivesProducer/'simCscTriggerPrimitiveDigis'
Exception Message:
GEMDetId ctor: Invalid parameters:  region 1 ring 1 station 2 layer 2 chamber 10 roll 17

dildick:
@jshlee Can you copy the entire exception message? Can't really tell where it is coming from.

bsunanda:
@dildick @jshlee @watson-ij I just checked the DetId code. I could not find it is caused by the DetId code

jshlee:
> @dildick @jshlee @watson-ij I just checked the DetId code. I could not find it is caused by the DetId code

@dildick @bsunanda yep, I suspect it is happening when GEMDetId is constructed somewhere in CSCMotherboardME21GEM.

dildick:
@jshlee So... this geometry will not work with the GE2/1-ME2/1 trigger. The coincidence pad producer may work, but all ILT lookup tables were made for an 8-eta partition geometry. Until that is updated, the code cannot be properly tested. Also the pad producer for GE2/1 needs to be modified.

jshlee:
> @jshlee So... this geometry will not work with the GE2/1-ME2/1 trigger. The coincidence pad producer may work, but all ILT lookup tables were made for an 8-eta partition geometry. Until that is updated, the code cannot be properly tested. Also the pad producer for GE2/1 needs to be modified.

ok, we can turn gem trigger for this geometry to begin with.
it happens in
https://github.com/cms-sw/cmssw/blob/master/L1Trigger/CSCTriggerPrimitives/src/GEMCoPadProcessor.cc#L44

dildick:
In addition to the trigger, also the validation modules will either crash or produce nonsense.

dildick:
And because we have to support the 8 eta partition GE2/1 geometry for the HLT TDR, I think it would be best to enable this geometry only in a specific Phase2 development era.

cvuosalo:
+1

civanch:
+1

silviodonato:
+1

Issue #29406, opened 2020-04-07
Prepare Run-3 geometry


civanch:
Geometry of Run-3 slightly different from Run-2: Castor is removed, Totem-2 is added, YE3 width and position is changed, few overlaps are fixed in tracker and muon system. Run-3 XML and code should be completed, after geometry DB should be updated. The target release CMSSW_11_1_0.

civanch:
geometry+

civanch:
+geometry

civanch:
assign geometry

civanch:
assign alcadb

civanch:
assign aldb

civanch:
assign alca

christopheralanwest:
What is the status of this issue? Are Run 3 geometry DB payloads available for AlCa to include in the Run 3 global tags?

cvuosalo:
@christopheralanwest We have been hoping to include the PPS XML revisions in the Run 3 geometry payloads. Unfortunately, PPS development ran into various problems. Now PR #31383 seems close to getting merged, so I'm waiting for it before creating the Run 3 geometry DB payloads. Your review and approval of #31383 would be appreciated.
Is time running out to include new geometry payloads in the Run 3 Global Tags?

christopheralanwest:
Run 3 geometry payloads have been updated in the recently merged PR #31529. Since there are some remaining issues (PPS and zero-material scenario), I will not close this issue yet.

tvami:
Hi @cvuosalo this issue is ready to be closed, right? This was even before the DD4hep changes, so I think we are already one step ahead then what this issue reports. Thanks!

cvuosalo:
+1

cvuosalo:
@tvami Yes, the Run 3 geometry has been completed. This issue can be closed.

tvami:
+alca

PR #29441, opened 2020-04-09
Phase1 Tracker: Added rails, Shrinked outermost shield, Added TBPX inter-flanges support tubs

ghugo83:
This fixes a mismatch between 2018 data and CMSSW Phase 1 Tracker description, as described at: https://indico.cern.ch/event/886623/contributions/3743154/attachments/1994728/3327417/2020.02.27_MaterialBudget_kropiv.pdf

**From data (2018D Single Mu PD):**
data

**From framework geometry:**
mc

What was changed, with info from Mechanics team:
- Added rails and updated support: both around Barrel and Forward sections.
- Shrinked Al-Kevlar outermost shield.
- Also realized that the 8 inter-flange tubs were missing (they are used to insure the support junction between the 2 TBPX flanges), hence added them.

**New Phase 1 Tracker geometry in framework:**
_(XY) section at Z=0:_
XY
_Profile around Barrel:_
profileright

NB 1: Obvisouly, 'Phase 0' Tracker not modified. 
There is also a version of Phase 2 Outer Tracker which is plugged on a Phase 1 Inner Tracker. I had done that description at a time where we did not have any Phase 2 Inner Tracker design / description yet. This description is kept for reference only, hence was of course not modified either.

NB 2: I have not done nor looked at the rest of the Phase 1 Tracker description, so discard any responsibility from it :)

NB 3: Do not have info on brackets, inserts, screws and pins, not added here.

**PR tests:**
No G4 error with new shapes + Checked there is still 0 overlap on both reference and migrated Phase 1 Tracker geometry.


civanch:
please test

civanch:
Problem is not connected with this PR

cvuosalo:
This PR changes the Phase 1 tracker XML. Does this change affect Run 2? 
Actually, I think you need to create `v1` directories to put the changes into.

civanch:
@ghugo83 , I agree with Carl. We should not edit Run-2 XML but create a new one. Please, move updated file to "v1". @cvuosalo , in which directory this v1 should be?

civanch:
@ghugo83 , @cvuosalo , @ianna,  I think we need sort out in which subdirectory this fix should go and the same for already merged #29366 .

Geometry/TrackerCommonData/data/PhaseI/pixbarmaterial.xml
```diff
@@ -131,6 +131,12 @@
       <rMaterial name="materials:CFK"/>
     </MaterialFraction>
   </CompositeMaterial>
+
+  <CompositeMaterial name="InterFlangesSuppTub_CFK" density="1.575*g/cm3" symbol=" " method="mixture by weight">
```
ianna, line 135: @ghugo83 - this new material is identical to the one above. Does it make sense to create an extra one?
@civanch - FYI

Geometry/TrackerCommonData/data/PhaseI/pixbar.xml
```diff
@@ -930,7 +1096,7 @@
    <Translation x="[zero]" y="[zero]" z="-[PixelBarrelPipe_FlangeToST_Z]" />
  </PosPart>
 
- 
+ <!-- CONNECTORS --> 
  <!--
```
ianna, line 1100: @ghugo83 - I'd suggest to remove commented out parts.

ghugo83:
@civanch Yes I discussed with @cvuosalo privately, because I created v1 but was not sure on the convention for Run 3, which scenarios to include it in.
As discussed it will be in:
Geometry/CMSCommonData/python/cmsExtendedGeometry2021XML_cfi.py
DetectorDescription/DDCMS/data/cms-geometry-2021.xml
DetectorDescription/DDCMS/data/cms-tracker-2021.xml

Geometry/TrackerCommonData/data/PhaseI/pixbarmaterial.xml
ghugo83: Hi @ianna 
Thanks for your feedback.

Here I followed the convention which was already there in pixbar.xml, ie that apart from materials:Air, all materials are defined in pixbarmaterial.xml, and trackermaterial.xml.

Then I followed an existing convention in Phase 1 apparently (see pixbarmaterarial.xml for example): the materials are defined separately when it is not obvious, from a Mechanical point of view, that they have same composition.
In my understanding, it is made like that, so that if later on, the density / composition is different, it is trivial to update it in pixbarmaterial.xml

For example here, InterFlangesSuppTub (tubes linking the 2 flanges) do not compulsory have the same CFK as the CFK used for the flange disks.
It is then trivial to update.
In that file, evth follows that logic, for example Flange1_CFK, Flange2_CFK, Flange3_CFK are actually identical, and could have been InnerRadiusFlange_CFK.

Geometry/TrackerCommonData/data/PhaseI/pixbar.xml
ghugo83: @ianna 
As you want, I found it nice to add a few comments, I find it makes it much easier to avoid errors and know which section of the detector is being described :/

Geometry/TrackerCommonData/data/PhaseI/pixbarmaterial.xml
civanch: Of course, not - existing material should be reused.

ghugo83: Ok, but there are a lot of materials following this convention in this file. 
Anyway I removed it, because this is a non v1 MB file and now the LogicalSection is defined in v1, so was not worth the hassle in any case.

civanch: @ghugo83 , I reply without reading of your reply. To be more precise: if it is assumed that density of similar materials may be different we can create different materials for different flanges. This should be a requirement from tracker group, which plan making material budget studies and tuning. In most cases, nobody will change geometry and we should maximally reuse existing material list.

civanch:
As discussed it will be in:
Geometry/CMSCommonData/python/cmsExtendedGeometry2021XML_cfi.py
DetectorDescription/DDCMS/data/cms-geometry-2021.xml
DetectorDescription/DDCMS/data/cms-tracker-2021.xml

@ghugo83 , I confirm that.

The problem we are discussing : two files in this PR should in the directory ../Phase-1/v1/
Previous versions of these files should remain unchanged inside ../Phase-1/

I am asking Carl and Ianna to confirm that or propose better places. Also we need to check   #29366 and make the same for files which were edited and not created as a new.

ghugo83:
@civanch For #29366  it was only overlap fix, no diff in the geo itself, so not sure it is really worth the hassle.

This PR: Am adding TFPX rails as well now on it in the meantime.

mmusich:
@ghugo83 
sorry to jump in,

> This PR: Am adding TFPX rails as well now on it in the meantime.

Can you share a picture of the FPIX rails as well once you have them?

cvuosalo:
@ghugo83, @civanch Yes, I agree with Vladimir. It would be best to also do a follow-up to #29366 that puts the new versions in `v1` directories.

ghugo83:
> Can you share a picture of the FPIX rails as well once you have them?

Yes I have added them now, please find a couple of pics below.
NB: The holes are colored in black here, but they are made of Air.

xy
cote
cote2

civanch:
please test

cvuosalo:
+1

kpedro88:
+upgrade

silviodonato:
+1

mmusich:
Dear all,
I am wondering now that this is merged, who's going to propagate these changes (also the ones in https://github.com/cms-sw/cmssw/pull/29366)  to the DB geometry objects actually consumed in sample production? 
This is somehow related to issue https://github.com/cms-sw/cmssw/issues/29406 @civanch

cvuosalo:
@mmusich We aim to complete all geometry XML changes for Run 3 in CMSSW 11_1. After all the changes are complete, then I would create the new geometry DB payloads.
You mention a sample production. Do you know when the sample production is planned?

mmusich:
@cvuosalo if we had these changes in DB right now, we would gladly ask production of samples to validate the change against data. As it is missing it's a showstopper.

cvuosalo:
@mmusich How many validation campaigns would you like to run on this series of geometry changes? We have more changes to the geometry XML coming soon. I was assuming that there would be only one validation of the new geometry after all the changes were completed. Also, we are still doing our own internal checking of these changes.

mmusich:
@cvuosalo this change is main one regarding the Tracker geometry and it is also the only one we can actually validate beforehand by comparison with Run2 data, so in that sense is a bit "special" as it is more of a bug-fix than an update. I guess we can wait for the availability of 11_1_0 assuming it will contain the right DB tags in Global Tag.

civanch:
@cvuosalo , @mmusich , we need to complete XML for Run3. beam-pipe and pps are two main components , which are not yet included into XML, there are also questions to materials description. For many reasons, including what you said, we need to make DB for Run3 asap, not waiting to the end of 11_1_0 cycle.

PR #29453, opened 2020-04-10
improve exception reporting in HLTTauRefProducer

slava77:
The change is to raise whyFailed if tau ID container provenance is not available.
This way we will get an edm::Exception instead of a segfault

this will reclassify #29448 from a crash to an error


slava77:
@cmsbuild please test workflow 136.7803,136.7801

slava77:
e.g. in 136.7801 the error now is 
```
   [2] Calling method for module HLTTauRefProducer/'TauRefProducer'
Exception Message:
Looking for type: edm::ValueMap<reco::SingleTauDiscriminatorContainer>
Looking for module label: hpsPFTauBasicDiscriminators
```
this is instead of a segfault; this was the goal of this PR.

slava77:
@cmsbuild please test

now to check that the standard short matrix is OK

jfernan2:
+dqm

silviodonato:
+1

PR #29205, opened 2020-03-14
Run-3 CLCT and LCT data formats (CCLUT-3)

dildick:
#### PR description:

This pull requests implements the CLCT and LCT data formats for Run-3 data taking. The Run-3 CSC local trigger will be notably different than then Run-1/2 trigger. CSC stubs will have a higher resolution in position (x2 to x4) and bending (x3). This is achieved by implementing a 12-bit comparator code data word (CCLUT algorithm). It is set by default to -1 for Run-1/2 CLCTs. For Run-3 it takes a value between 0 and 2^12 - 1. The halfstrip number of CLCT and LCT is given 2 extra bits, one for the 1/4 strip value and one for the 1/8 strip value. In addition, I added a new `ComparatorContainer` object to keep track of which comparators were used to build a CLCT. 

This PR is a follow-up of https://github.com/cms-sw/cmssw/pull/28846. But the code is written such that the original functions retain their original meaning for the default arguments. The actual CCLUT algorithm will be implemented in a separate PR.

More information can be found here: 
https://indico.cern.ch/event/881406/contributions/3713334/attachments/1973349/3283422/CSC_NewPatterns_SD_20200121.pdf

#### PR validation:

Tested with WF 22034.0.

#### if this PR is a backport please specify the original PR and why you need to backport that PR:

N/A

@barvic @tahuang1991 @ptcox

PR #29233, opened 2020-03-18
Store comparator hits in the CLCT (CCLUT-4)

dildick:
#### PR description:

This PR saves the comparator hits that were used to build a CLCT in a ComparatorContainer. (Similar to how a CSCSegment has access to the CSCRecHit2Ds.) The relevant change is in function `patternFinding`, which called for both the trigger and the pre-trigger.  This function now also identifies all hits that were part of a particular pattern. An intermediate pattern map keeps track of all comparator hits, for each key half-strip and for each pattern Id. Ultimately, only the comparator hits for the best pattern are stored in the CLCT. 

No changes to the actual trigger logic have been made. No changes are expected in the output. 

#### PR validation:

Tested with WF 22034.0.

#### if this PR is a backport please specify the original PR and why you need to backport that PR:

N/A

@tahuang1991

PR #29456, opened 2020-04-10
Run-3 ALCT and LCT data formats

dildick:
#### PR description:

This PR extends the ALCT and LCT data formats for Run-3. Detailed information can be found in this note: https://gitlab.cern.ch/tdr/notes/DN-20-016/blob/master/temp/DN-20-016_temp.pdf

Edited Summary:

**ALCT changes** (section 3.1 in DN-20-016)
- 3 highest bits in BXN[4:0], BXN[4:2], will be reassigned for high-multiplicity triggering
- a new high-multiplicity trigger member `hmt_` was added
- a new member `version_` that indicates Run3 or Legacy was added

**CLCT changes**
- a new member `version_` that indicates Run3 or Legacy was added

**LCT changes** (section 3.2 in DN-20-016)
- 2 bits are reallocated to encode the 1/4 and 1/8 strip bit (see also #29205 )
- 4-bit quality becomes a 3-bit for ME1/1, or 2-bit for non-ME1/1
- 4-bit pattern ID is reinterpreted as 4-bit bending value
- 2-bit (3-bit) high-multiplicity trigger word for each ME1/1 (non-ME1/1) chamber
- a new high-multiplicity trigger member `hmt` was added
- a new member `version_` that indicates Run3 or Legacy was added

#### PR validation:

Tested with WF 22034.0

N/A

### DIFF
--- a/DataFormats/CSCDigi/interface/CSCALCTDigi.h
+++ b/DataFormats/CSCDigi/interface/CSCALCTDigi.h
@@ -36,14 +36,14 @@ public:
   void setValid(const int valid) { valid_ = valid; }
 
   /// return quality of a pattern
-  int getQuality() const { return quality_; }
+  uint16_t getQuality() const { return quality_; }
 
   /// set quality
   void setQuality(const int quality) { quality_ = quality; }
 
   /// return Accelerator bit
   /// 1-Accelerator pattern, 0-CollisionA or CollisionB pattern
-  int getAccelerator() const { return accel_; }
+  uint16_t getAccelerator() const { return accel_; }
 
   /// set accelerator bit
   void setAccelerator(const int accelerator) { accel_ = accelerator; }
@@ -51,31 +51,31 @@ public:
   /// return Collision Pattern B bit
   /// 1-CollisionB pattern (accel_ = 0),
   /// 0-CollisionA pattern (accel_ = 0)
-  int getCollisionB() const { return patternb_; }
+  uint16_t getCollisionB() const { return patternb_; }
 
   /// set Collision Pattern B bit
   void setCollisionB(const int collision) { patternb_ = collision; }
 
   /// return key wire group
-  int getKeyWG() const { return keywire_; }
+  uint16_t getKeyWG() const { return keywire_; }
 
   /// set key wire group
   void setKeyWG(const int keyWG) { keywire_ = keyWG; }
 
   /// return BX - five low bits of BXN counter tagged by the ALCT
-  int getBX() const { return bx_; }
+  uint16_t getBX() const { return bx_; }
 
   /// set BX
   void setBX(const int BX) { bx_ = BX; }
 
   /// return track number (1,2)
-  int getTrknmb() const { return trknmb_; }
+  uint16_t getTrknmb() const { return trknmb_; }
 
   /// Set track number (1,2) after sorting ALCTs.
   void setTrknmb(const uint16_t number) { trknmb_ = number; }
 
   /// return 12-bit full BX.
-  int getFullBX() const { return fullbx_; }
+  uint16_t getFullBX() const { return fullbx_; }
 
   /// Set 12-bit full BX.
   void setFullBX(const uint16_t fullbx) { fullbx_ = fullbx; }
--- a/DataFormats/CSCDigi/interface/CSCALCTDigi.h
+++ b/DataFormats/CSCDigi/interface/CSCALCTDigi.h
@@ -109,7 +109,7 @@ public:
   /// Distinguish Run-1/2 from Run-3
   bool isRun3() const { return version_ == Version::Run3; }
 
-  void setRun3(bool isRun3);
+  void setRun3(const bool isRun3);
 
 private:
   uint16_t valid_;
--- a/DataFormats/CSCDigi/interface/CSCALCTDigi.h
+++ b/DataFormats/CSCDigi/interface/CSCALCTDigi.h
@@ -14,6 +14,8 @@
 
 class CSCALCTDigi {
 public:
+  enum class Version { Legacy = 0, Run3 };
+
   /// Constructors
   CSCALCTDigi(const int valid,
               const int quality,
@@ -22,7 +24,8 @@ public:
               const int keywire,
               const int bx,
               const int trknmb = 0,
-              const int hmt = 0);
+              const int hmt = 0,
+              const Version version = Version::Legacy);
   /// default
   CSCALCTDigi();
 
@@ -120,8 +123,6 @@ private:
   /// Note: In DN-20-016, 3 bits are allocated for HMT in the
   /// ALCT board. These bits are copied into the ALCT digi in CMSSW
   uint16_t hmt_;
-
-  enum class Version { Legacy = 0, Run3 };
 
   Version version_;
 };
--- a/DataFormats/CSCDigi/interface/CSCALCTDigi.h
+++ b/DataFormats/CSCDigi/interface/CSCALCTDigi.h
@@ -14,8 +14,6 @@
 
 class CSCALCTDigi {
 public:
-  enum ALCTHMT { kIsRun3Mask = 0x1, kHMTMask = 0x7, kIsRun3Shift = 14, kHMTShift = 0 };
-
   /// Constructors
   CSCALCTDigi(const int valid,
               const int quality,
@@ -83,10 +81,10 @@ public:
   void setFullBX(const uint16_t fullbx) { fullbx_ = fullbx; }
 
   /// return the high multiplicity bits
-  int getHMT() const;
+  int getHMT() const { return hmt_; }
 
   /// set the high multiplicity bits
-  void setHMT(const int hmt);
+  void setHMT(const int hmt) { hmt_ = hmt; }
 
   /// True if the first ALCT has a larger quality, or if it has the same
   /// quality but a larger wire group.
@@ -105,9 +103,9 @@ public:
   void setWireGroup(unsigned int wiregroup) { keywire_ = wiregroup; }
 
   /// Distinguish Run-1/2 from Run-3
-  bool isRun3() const;
+  bool isRun3() const { return version_ == Version::Run3; }
 
-  void setRun3(bool isRun3);
+  void setRun3(bool isRun3) { version_ = Version::Run3; }
 
 private:
   uint16_t valid_;
@@ -122,6 +120,10 @@ private:
   /// Note: In DN-20-016, 3 bits are allocated for HMT in the
   /// ALCT board. These bits are copied into the ALCT digi in CMSSW
   uint16_t hmt_;
+
+  enum class Version { Legacy = 0, Run3 };
+
+  Version version_;
 };
 
 std::ostream& operator<<(std::ostream& o, const CSCALCTDigi& digi);
--- a/DataFormats/CSCDigi/interface/CSCALCTDigi.h
+++ b/DataFormats/CSCDigi/interface/CSCALCTDigi.h
@@ -81,10 +81,10 @@ public:
   void setFullBX(const uint16_t fullbx) { fullbx_ = fullbx; }
 
   /// return the high multiplicity bits
-  uint16_t getHMT() const { return ( (version_ == Version::Run3 ) ? hmt_ : -1 ); }
+  uint16_t getHMT() const { return ((version_ == Version::Run3) ? hmt_ : -1); }
 
   /// set the high multiplicity bits
-  void setHMT(const int hmt) {version_ == Version::Run3 ? hmt_ = hmt : -1; }
+  void setHMT(const int hmt) { version_ == Version::Run3 ? hmt_ = hmt : -1; }
 
   /// True if the first ALCT has a larger quality, or if it has the same
   /// quality but a larger wire group.
@@ -105,7 +105,7 @@ public:
   /// Distinguish Run-1/2 from Run-3
   bool isRun3() const { return version_ == Version::Run3; }
 
-  void setRun3(bool isRun3) { isRun3 ?  version_ = Version::Run3 : Version::Legacy; }
+  void setRun3(bool isRun3) { isRun3 ? version_ = Version::Run3 : Version::Legacy; }
 
 private:
   uint16_t valid_;
--- a/DataFormats/CSCDigi/interface/CSCALCTDigi.h
+++ b/DataFormats/CSCDigi/interface/CSCALCTDigi.h
@@ -81,10 +81,10 @@ public:
   void setFullBX(const uint16_t fullbx) { fullbx_ = fullbx; }
 
   /// return the high multiplicity bits
-  int getHMT() const { return hmt_; }
+  uint16_t getHMT() const { return ( (version_ == Version::Run3 ) ? hmt_ : -1 ); }
 
   /// set the high multiplicity bits
-  void setHMT(const int hmt) { hmt_ = hmt; }
+  void setHMT(const int hmt) {version_ == Version::Run3 ? hmt_ = hmt : -1; }
 
   /// True if the first ALCT has a larger quality, or if it has the same
   /// quality but a larger wire group.
@@ -105,7 +105,7 @@ public:
   /// Distinguish Run-1/2 from Run-3
   bool isRun3() const { return version_ == Version::Run3; }
 
-  void setRun3(bool isRun3) { version_ = Version::Run3; }
+  void setRun3(bool isRun3) { isRun3 ?  version_ = Version::Run3 : Version::Legacy; }
 
 private:
   uint16_t valid_;
--- a/DataFormats/CSCDigi/interface/CSCALCTDigi.h
+++ b/DataFormats/CSCDigi/interface/CSCALCTDigi.h
@@ -14,6 +14,8 @@
 
 class CSCALCTDigi {
 public:
+  enum ALCTHMT { kIsRun3Mask = 0x1, kHMTMask = 0x7, kIsRun3Shift = 14, kHMTShift = 0 };
+
   /// Constructors
   CSCALCTDigi(const int valid,
               const int quality,
@@ -81,10 +83,10 @@ public:
   void setFullBX(const uint16_t fullbx) { fullbx_ = fullbx; }
 
   /// return the high multiplicity bits
-  int getHMT() const { return hmt_; }
+  int getHMT() const;
 
   /// set the high multiplicity bits
-  void setHMT(const int hmt) { hmt_ = hmt; }
+  void setHMT(const int hmt);
 
   /// True if the first ALCT has a larger quality, or if it has the same
   /// quality but a larger wire group.
@@ -101,6 +103,11 @@ public:
 
   /// set wiregroup number
   void setWireGroup(unsigned int wiregroup) { keywire_ = wiregroup; }
+
+  /// Distinguish Run-1/2 from Run-3
+  bool isRun3() const;
+
+  void setRun3(bool isRun3);
 
 private:
   uint16_t valid_;
--- a/DataFormats/CSCDigi/interface/CSCALCTDigi.h
+++ b/DataFormats/CSCDigi/interface/CSCALCTDigi.h
@@ -11,20 +11,21 @@
 
 #include <cstdint>
 #include <iosfwd>
+#include <limits>
 
 class CSCALCTDigi {
 public:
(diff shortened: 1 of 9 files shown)


dildick:
@civanch I think I still need to add a member in order to distinguish Run-1/2 from Run-3 ALCTs and LCTs. I was able to do this for CLCTs in #29233 with the comparator code value. For ALCTs and LCTs, there is no obvious choice with the existing members. Some CSC folks suggested to add a flag `isRun3`. Can you offer a recommendation?

civanch:
please test

civanch:
@dildick , addition of an extra flag seems to be a working solution. However, it is not the first case when in a sub-detector we have to modify data structure, let us discuss with experts what is optimal in the case of a minor change as you have.

dildick:
@silviodonato Do you have a suggestion I can implement?

civanch:
@makortel , @bsunanda , can you, please, comment what is optimal solution for a producer if data format slightly changed for Run-3?

silviodonato:
meanwhile, @dildick, could you confirm that this PR has no effects on Phase-2 digis?

silviodonato:
If I understand correctly, 3-bits data are stored in a 16-bits variable (`hmt_`). A possible idea is to use part of this variable (eg. `hmt_%(2^3)`) to tag if `hmt_` is filled (Run-3) or empty (Run-1, Run-2). Anyway, I let @makortel , @bsunanda comment on this.

bsunanda:
I believe the optimal solution is that in memory the format is the new one. So if one reads in the object produced with an old version, it should be translated into the new format. I believe this is allowed by ROOT and handled in classes_def.xml. Once HcalDetId format was changed and the corresponding classes_def.xml invokes a method within the class which changes the format. Chris or Matti should give the correct advice.

dildick:
@silviodonato We plan to use the Run-3 objects also in Phase-2 workflows. It will thus have an effect, intentionally. We will have higher resolution in position and bending, for CLCTs and LCTs. For Run-3 we have identified a few bits for high multiplicity triggering. It makes sense to use them in Phase-2 workflows as well. That way we can make projections of the CMS physics reach at the HL-LHC.

I like the idea of using the `hmt_` bits to distinguish Run-1/2 from Run-3. But in that case, I would rather use bit `hmt_[15]`, than bit `hmt_[3]` - just in case more bits become available for high multiplicity triggering. I would also add a function `isRun3` which specifically looks at `hmt_[15]`. The `getHMT()` (`setHMT()`) function can then be updated to return (set) only `hmt_[2:0]`. I guess the `DataFormats/CSCDigi/src/classes_def.xml` file would need a target?

dildick:
My last commit is a proposal to utilize bit 15 of the `hmt` field to indicate Run-3 for ALCT and LCT.

silviodonato:
@dildick , you proposal looks ok to me. 
About my question of the Phase-2 digi, I was worried about the re-reco of the HLT TDR.
We are going to run a re-reco in the future 11_1_0 using RAW/digis produced in 11_0_0, so we need to keep this compatibility. As Sunanda wrote, the new format should be able to read the old data format, filling the new variables with the default value. You may also double check that the old `CSCDigi` are readable properly after this PR.
We encountered a similar issue in #28579.
cc: @trtomei @fwyzard

fwyzard:
From a sanity point of view, I'd rather suggest using an additional `bool` or `enum` variable to identify the "version" of the data (Run 2 vs Run 3 vs Run 4 ...).

From the backwards compatibility point of view: do the Phase 2 samples store the RAW or the DIGIs for the CSC ?
If we store the RAW, how will the unpacker behave when finding RAW data without the additional information ?
Either way, how will the local reconstruction cope with Phase 2 data that doesn't have the additional information ?

makortel:
Just repeating to check that I understood. The `CSCALCTDigi` and `CSCCorrelatedLCTDigi` gain a new member (`uint16_t hmt_`). The same classes will be used for data from <= Run 2 and >= Run 3, but the interpretation of the other member fields can be different in the two modes. Right?

The ROOT schema evolution should take care to initialize the new `hmt_` to 0 (same for any new member if a separate variable is used) when reading old files. If all of the "old files" would be "<= Run 2", I think it should work. If the "old files" contain also ">= Run 3", then it gets more complicated and I don't have any good ideas for that at the moment.

dildick:
@makortel " The CSCALCTDigi and CSCCorrelatedLCTDigi gain a new member (uint16_t hmt_). The same classes will be used for data from <= Run 2 and >= Run 3, but the interpretation of the other member fields can be different in the two modes. " - That is correct.

@silviodonato @fwyzard Concerning the re-reco for the HLT TDR. I can give you a bit more information so we can work out a strategy. The Run-3 algorithms, which assign the high multiplicity bits in the ALCT and LCT, and the comparator code in the CLCT (https://github.com/cms-sw/cmssw/pull/29205), are not enabled yet. And they probably won't be for a while. The CSC packer and unpacker are still the same. They won't be reading the hmt bits for a while either.

fwyzard:
@dildick Thanks for the clarification.
If the new data member is not filled (by the unpacker) or used (by the local reco) in the final 11.1.0 release, then the addition of the new member shouldn't cause any compatibility problems with the 11.0.0 MC samples.

Regarding the strategy, I still suggest to use a separate `version_` (or whatever name is preferred) variable rather than one bit of `hmt_` member.

trtomei:
> Regarding the strategy, I still suggest to use a separate version_ (or whatever name is preferred) variable rather than one bit of hmt_ member.

For the record, I also think that this is an excellent idea. It is better to be explicit with versioning -- in five years chances are that the memory of using a bit to mark a version is both inadequate (what if we need more versions) and badly documented.

dildick:
@fwyzard Yes, we will not enable any of the CSC upgrades by the final 11.1.0 version. It could easily take another 6 months before the code is run in standard productions. After all, it requires a synchronization between CSC, EMTF, OMTF, packer/unpacker, DQM, validation etc. So we will be making the upgrade algorithms available in the CSC trigger, but will not be enabling them yet. 

Regarding the versioning member, you suggested an enum, correct? Can you point me to an example data format?

fwyzard:
> Regarding the versioning member, you suggested an enum, correct? Can you point me to an example data format?

I don't know by heart... I think something like this should work:
```c++
class CSCALCTDigi {

  ...

public:
  bool isRun3() const {
    return version_ == Version::Run3;
  }

  void setRun3(bool isRun3) {
    version_ = Version::Run3;
  }

private:
  enum class Version {
    Legacy = 0,
    Run3
  };

  Version version_;
};

**Comment by dildick** [comment]
@fwyzard Awesome. I'll add it to the ALCT, CLCT and LCT. 

@jiafulow @abrinke1 Would this be clear enough for the EMTF?

**Comment by dildick** [comment]
So this branch has the version in each object as requested. I had to rewrite the history a few times to get a clean branch. Let me know if anything else is missing.

File: DataFormats/CSCDigi/interface/CSCCorrelatedLCTDigi.h [hunk 1]
```diff
@@ -133,6 +141,14 @@ class CSCCorrelatedLCTDigi {
   /// set cscID
   void setCSCID(unsigned int c) { cscID = c; }
 
+  /// set high-multiplicity bits
+  void setHMT(const unsigned int h) { hmt = h; }
+
+  /// Distinguish Run-1/2 from Run-3
+  bool isRun3() const { return version_ == Version::Run3; }
+
+  void setRun3(bool isRun3) { version_ = Version::Run3; }
```
abrinke1, line 150: What does the bool "isRun3" do here?  It looks like "version_" is set to Run3 whether "isRun3" is true or false.

DataFormats/CSCDigi/interface/CSCALCTDigi.h
```diff
@@ -104,6 +116,14 @@ class CSCALCTDigi {
...
   uint16_t trknmb_;
   uint16_t fullbx_;
+  /// Run-3 introduces high-multiplicity bits for CSCs.
+  /// Note: In DN-20-016, 3 bits are allocated for HMT in the
+  /// ALCT board. These bits are copied into the ALCT digi in CMSSW
+  uint16_t hmt_;
+
+  enum class Version { Legacy = 0, Run3 };
+
+  Version version_;
```
abrinke1, line 126: What is the default value of "version_", and where is this set?

DataFormats/CSCDigi/interface/CSCCorrelatedLCTDigi.h
```diff
@@ -78,6 +79,7 @@ class CSCCorrelatedLCTDigi {
   int getBX() const { return bx; }
 
   /// return CLCT pattern number (in use again Feb 2011)
+  /// This function should not be used for Run-3
```
abrinke1, line 82: In that case, IMO it would be better to replace the line below with something like this:
int getCLCTPattern() const { return ( (version_ == Version::0) ? (pattern & 0xF) : -1 ); }

dildick: Yes, good point.

DataFormats/CSCDigi/interface/CSCCorrelatedLCTDigi.h
fwyzard: yeah, sorry about that, that come from my hasty suggestion...

fwyzard, line 150:
```suggestion
  void setRun3(bool isRun3) { version_ = isRun3 ? Version::Run3 : Version::Legacy; }
```

DataFormats/CSCDigi/interface/CSCALCTDigi.h
```diff
@@ -95,6 +102,11 @@ class CSCALCTDigi {
   /// set wiregroup number
   void setWireGroup(unsigned int wiregroup) { keywire_ = wiregroup; }
 
+  /// Distinguish Run-1/2 from Run-3
+  bool isRun3() const { return version_ == Version::Run3; }
+
+  void setRun3(bool isRun3) { version_ = Version::Run3; }
```
fwyzard, line 108:
```suggestion
  void setRun3(bool isRun3) { version_ = isRun3 ? Version::Run3 : Version::Legacy; }
```

DataFormats/CSCDigi/interface/CSCCLCTDigi.h
```diff
@@ -142,6 +142,11 @@ class CSCCLCTDigi {
   /// Print content of digi.
   void print() const;
 
+  /// Distinguish Run-1/2 from Run-3
+  bool isRun3() const { return version_ == Version::Run3; }
+
+  void setRun3(bool isRun3) { version_ = Version::Run3; }
```
fwyzard, line 148:
```suggestion
  void setRun3(bool isRun3) { version_ = isRun3 ? Version::Run3 : Version::Legacy; }
```

DataFormats/CSCDigi/interface/CSCALCTDigi.h
dildick: The default is now "Legacy" it appears.

fwyzard: Uninitialized.

dildick:
Hi @fwyzard, I'm not sure how to interpret this ```void setRun3(bool isRun3) { version_ = isRun3 ? Version::Run3 : Version::Legacy; }```, but sure I'll use the version check for the getPattern function, the set/getHMT, and for the set/getValid functions.

dildick:
@fwyzard Never mind my last statement. I'll incorporate it.

dildick:
Alright, I think I have a good version now. Can someone run the tests please?

DataFormats/CSCDigi/interface/CSCALCTDigi.h
```diff
@@ -35,50 +39,56 @@ class CSCALCTDigi {
...
 
   /// return 12-bit full BX.
-  int getFullBX() const { return fullbx_; }
+  uint16_t getFullBX() const { return fullbx_; }
 
   /// Set 12-bit full BX.
   void setFullBX(const uint16_t fullbx) { fullbx_ = fullbx; }
 
+  /// return the high multiplicity bits
+  uint16_t getHMT() const { return ((version_ == Version::Run3) ? hmt_ : -1); }
```
abrinke1, line 87: Probably not a great idea to return -1 as an unsigned int; maybe better to use 0?  (Same for the setHMT below.)

DataFormats/CSCDigi/interface/CSCALCTDigi.h
```diff
@@ -35,50 +39,56 @@ class CSCALCTDigi {
...
+
+  /// set the high multiplicity bits
+  void setHMT(const int hmt) { version_ == Version::Run3 ? hmt_ = hmt : -1; }
```
abrinke1, line 90: I think the logic needs to be changed slightly as well, from:
version_ == Version::Run3 ? hmt_ = hmt : -1;
to:
hmt_ = (version_ == Version::Run3 ? hmt : 0);
But it's possible your original if/else logic is valid: I'm not an expert.

DataFormats/CSCDigi/interface/CSCALCTDigi.h
dildick: Here, 0 means "no high multiplicity event in chamber", which makes sense for Run-3, but not for legacy data since the HMT was not enabled then. Hence the "-1" for legacy

abrinke1: Right, but the function will not return "-1", it will return some large positive number (I think 2^16 - 1, so 65535), since the function returns a uint16_t.  Maybe this is OK, but 65535 is a less obvious "Legacy / Invalid" flag than 0, or 999, or something like that.

dildick: "it will return some large positive number" - Exactly. This is what I had in mind.

makortel: `std::numeric_limits<uint16_t>::max()` would be a bit more clear about that.

DataFormats/CSCDigi/interface/CSCALCTDigi.h
civanch: I would agree, that "hmt_ = (version_..." is better

DataFormats/CSCDigi/interface/CSCALCTDigi.h
civanch: I would not use -1, @makortel proposal is the most classical

DataFormats/CSCDigi/interface/CSCALCTDigi.h
```diff
@@ -95,6 +105,11 @@ class CSCALCTDigi {
   /// set wiregroup number
   void setWireGroup(unsigned int wiregroup) { keywire_ = wiregroup; }
 
+  /// Distinguish Run-1/2 from Run-3
+  bool isRun3() const { return version_ == Version::Run3; }
+
+  void setRun3(bool isRun3) { isRun3 ? version_ = Version::Run3 : Version::Legacy; }
```
fwyzard, line 111: this is wrong; it is the equivalent of
```c++
if (isRun3) {
  (void) version_ = Version::Run3;
} else {
  (void) Version::Legacy;
}
```
while what you want is probably
```c++
if (isRun3) {
  version_ = Version::Run3;
} else {
  version_ = Version::Legacy;
}
```
or, on one line
```c++
version_ = isRun3 ? Version::Run3 : Version::Legacy;
```

So
```suggestion
  void setRun3(bool isRun3) { version_ = isRun3 ? Version::Run3 : Version::Legacy; }
```

DataFormats/CSCDigi/interface/CSCALCTDigi.h
fwyzard: > Here, 0 means "no high multiplicity event in chamber", which makes sense for Run-3, but not for legacy data since the HMT was not enabled then.

So, in a sense, there could not be any high multiplicity events, and 0 could make sense ?
But I guess it depends on how he value will be used in the client code.

dildick: Yes, it is very likely that a chamber had 1 or 2 LCTs, without any high multiplicity hits (extra LCTs, or an unusual large number of single hits). In that case HMT = 0.

DataFormats/CSCDigi/interface/CSCALCTDigi.h
dildick: Thanks. I'll fix it.

DataFormats/CSCDigi/interface/CSCALCTDigi.h
dildick: @civanch @abrinke1 Thanks for the suggestion. I changed it to ``` hmt_ = (version_ == Version::Run3) ? hmt : std::numeric_limits<uint16_t>::max();```

DataFormats/CSCDigi/interface/CSCALCTDigi.h
dildick: @fwyzard @abrinke1 I set the version as ```Version::Legacy;``` by default in the ALCT/CLCT/LCT constructor.

civanch:
please test

perrotta:
please test

dildick:
Glad to see that the unit test passed this time. Other than that, the comparisons look good; no failures in the CSC system that I can tell.

DataFormats/CSCDigi/src/CSCCLCTDigi.cc
```diff
@@ -81,7 +84,7 @@ void CSCCLCTDigi::clear() {
   }
 }
 
-int CSCCLCTDigi::getKeyStrip(int n) const {
+uint16_t CSCCLCTDigi::getKeyStrip(const uint16_t n) const {
   // 10-bit case for strip data word
   if (compCode_ != -1 and n == 8) {
```
abrinke1, line 89: This is for Run 3 only, I guess?  Just to confirm: the "compCode" has never been used in Run 2, i.e. it was unused in firmware, and was always set to -1 in software?  (See getCompCode() function above.)

abrinke1:
Looks like there were 0 changes in any of the DQM histograms, including L1T:
https://cmssdt.cern.ch/SDT/jenkins-artifacts/baseLineComparisons/CMSSW_11_1_X_2020-04-16-2300+4c06e6/35939/1330.0_ZMM_13+ZMM_13+DIGIUP15+RECOUP15_L1TMuDQM+HARVESTUP15_L1TMuDQM+NANOUP15/

DataFormats/CSCDigi/src/CSCCLCTDigi.cc
dildick: Yes, correct. I suppose we can update this function so that for Run-1/2 it always returns the half-strip, whatever the argument.

abrinke1: I think that's already what it does, no?  See below, "// 8-bit case for strip data word (all other cases)"

dildick: Yes, although we could strengthen it with an `isRun3()` check.

dildick:
Hi, are there more comments on this PR?

perrotta:
+1
- Reconstruction code is only marginally touched by this PR: minor reco changes in EventFilter/CSCRawToDigi/src/CSCDigiToRaw.cc, justified by the change from ```int``` to ```uint16_t``` in the return value of two functions defined elsewhere
- Jenkins tests pass and show no differences in reco outputs

civanch:
+1

silviodonato:
@rekovic do you have any comments?

dildick:
Hi, can this PR be integrated?

silviodonato:
Reminder @rekovic  (CMSSW_11_1_0_pre7 is scheduled for tomorrow)

rekovic:
@dildick 
Please make a follow-up PR with changing the name of getter function to follow the naming rules 2.11 (http://cms-sw.github.io/cms_coding_rules.html)

+1

rekovic:
+1

silviodonato:
+1

dildick:
@rekovic I see. I'll need to remove the "get" from each getter function... Is there a chance this introduces an issue with reading old MC/data with newer code?

PR #29467, opened 2020-04-13
Update PR #28417, add era modifier to TauRefProducer to read old data format from existing files

swozniewski:
#### PR description:

Addressing issue #29448 
Given workflows read root files with old tauID format which requires an era modifier to adapt the configuration accordingly which is done in this PR.

#### PR validation:

Ran 136.7801 successfully

### DIFF
--- a/Configuration/PyReleaseValidation/python/relval_steps.py
+++ b/Configuration/PyReleaseValidation/python/relval_steps.py
@@ -2676,6 +2676,7 @@ steps['DQMHLTonAOD_2017']={
     '--datatier':'DQMIO',
     '--era':'Run2_2017',
     '--fileout':'DQMHLTonAOD.root',
+    '--procModifiers':'tau_readOldDiscriminatorFormat',
     }
 steps['DQMHLTonAODextra_2017'] = merge([ {'-s':'DQM:offlineHLTSourceOnAODextra'}, steps['DQMHLTonAOD_2017'] ])
 
@@ -2690,6 +2691,7 @@ steps['DQMHLTonRAWAOD_2017']={
     '--secondfilein':'filelist:step1_dasparentquery.log',
     '--customise_commands':'"process.HLTAnalyzerEndpath.remove(process.hltL1TGlobalSummary)"',
     '--fileout':'DQMHLTonAOD.root',
+    '--procModifiers':'tau_readOldDiscriminatorFormat',
     }
 
 steps['HARVESTDQMHLTonAOD_2017'] = merge([ {'--filein':'file:DQMHLTonAOD.root','-s':'HARVESTING:hltOfflineDQMClient'}, steps['HARVEST2017'] ]) ### Harvesting step for the DQM-only workflow
--- a/Configuration/ProcessModifiers/python/tau_readOldDiscriminatorFormat_cff.py
+++ b/Configuration/ProcessModifiers/python/tau_readOldDiscriminatorFormat_cff.py
@@ -1,0 +1,6 @@
+import FWCore.ParameterSet.Config as cms
+
+# This modifier is used to adapt input configurations of tau producers if input files with old tau ID format are used.
+
+tau_readOldDiscriminatorFormat =  cms.Modifier()
+
--- a/DQMOffline/Trigger/python/HLTTauDQMOffline_cfi.py
+++ b/DQMOffline/Trigger/python/HLTTauDQMOffline_cfi.py
@@ -1,4 +1,5 @@
 import FWCore.ParameterSet.Config as cms
+from Configuration.ProcessModifiers.tau_readOldDiscriminatorFormat_cff import tau_readOldDiscriminatorFormat
 
 hltTauDQMofflineProcess = "HLT"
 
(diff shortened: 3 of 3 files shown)


Configuration/Eras/python/Era_Run2_2017_cff.py
```diff
@@ -22,8 +22,9 @@
 from Configuration.Eras.Modifier_run2_jme_2017_cff import run2_jme_2017
 from Configuration.Eras.Modifier_run2_jme_2016_cff import run2_jme_2016
 from Configuration.Eras.Modifier_strips_vfp30_2016_cff import strips_vfp30_2016
+from Configuration.Eras.Modifier_run2_tauIDformat_cff import run2_tauIDformat
 
 Run2_2017 = cms.ModifierChain(Run2_2016.copyAndExclude([run2_muon_2016, run2_HLTconditions_2016, run2_ECAL_2016, run2_egamma_2016,pixel_2016,run2_jme_2016, strips_vfp30_2016]),
                               phase1Pixel, run2_ECAL_2017, run2_HF_2017, run2_HCAL_2017, run2_HE_2017, run2_HEPlan1_2017, 
-                              trackingPhase1, run2_GEM_2017, stage2L1Trigger_2017, run2_HLTconditions_2017, run2_muon_2017,run2_egamma_2017, ctpps_2017, run2_jme_2017)
+                              trackingPhase1, run2_GEM_2017, stage2L1Trigger_2017, run2_HLTconditions_2017, run2_muon_2017,run2_egamma_2017, ctpps_2017, run2_jme_2017, run2_tauIDformat)
```
slava77, line 29: I think that this modifier should be added to the workflow, not to the era.
IIUC, this current update will break other workflows that actually rerun tau reco.

Configuration/Eras/python/Modifier_run2_tauIDformat_cff.py
```diff
@@ -0,0 +1,4 @@
+import FWCore.ParameterSet.Config as cms
+
+run2_tauIDformat =  cms.Modifier()
```
slava77, line 3: why does this have run2_ prefix? IIRC, #28417 added tau container-based ID for all data/simulation eras/ranges.

I was actually expecting that the modifier will add the sequence/task to rerun tau IDs rather than insert support to read the old IDs, but for this case it seems like your solution is less invasive.

I propose to change the name of the modifier to ```tau_readOldDiscriminator ``` or something similar

jfernan2:
please test

Configuration/Eras/python/Era_Run2_2017_cff.py
swozniewski: Ok, could you point me to the configuration of these workflows?

Configuration/Eras/python/Modifier_run2_tauIDformat_cff.py
swozniewski: My reasoning was that files produced during run2 contain the old format (which could be called run2_tauIDformat), but I'm fine with calling it `tau_readOldDiscriminator` which better reflects what it is doing. However, one might think that this means to switch to a different ID rather than just an old format. Would `tau_readOldDiscriminatorFormat` also suitable or is it too long?

Configuration/Eras/python/Era_Run2_2017_cff.py
slava77: I think that the step definition is here https://github.com/cms-sw/cmssw/blob/959f56649b489b3c33a15bd502bef3df4f31ea78/Configuration/PyReleaseValidation/python/relval_steps.py#L2671-L2679

Configuration/Eras/python/Modifier_run2_tauIDformat_cff.py
slava77: "during run2" is just a calendar period, there are AOD files produced for Run-1, Run-3 and Phase-2 during that period. It would be rather confusing to use run2_ prefix for a modifier for these cases as well if it becomes necessary.

Configuration/Eras/python/Era_Run2_2017_cff.py
swozniewski: Thanks! I'll implement it as process modifier then

Configuration/Eras/python/Modifier_run2_tauIDformat_cff.py
swozniewski: true

swozniewski:
@slava77 I switched to a process modifier and added it via the `--procModifiers` option to the workflow you pointed at. However, I don't see this option in the printout of the cmsDriver command when running the 136.7801 workflow. (scram has been run) Do you have an idea what is missing?

Configuration/PyReleaseValidation/python/relval_steps.py
```diff
@@ -2690,6 +2690,7 @@ def gen2021HiMix(fragment,howMuch):
     '--secondfilein':'filelist:step1_dasparentquery.log',
     '--customise_commands':'"process.HLTAnalyzerEndpath.remove(process.hltL1TGlobalSummary)"',
     '--fileout':'DQMHLTonAOD.root',
+    '--procModifiers':'tau_readOldDiscriminatorFormat',
```
slava77, line 2693: IIUC, the problem shows up in DQMHLTonAOD_2017 as well (used in 136.7801).
This will address only 136.7803

slava77:
> @slava77 I switched to a process modifier and added it via the `--procModifiers` option to the workflow you pointed at. However, I don't see this option in the printout of the cmsDriver command when running the 136.7801 workflow. (scram has been run) Do you have an idea what is missing?

I commented inline; it looks like you edited a line for a different workflow

swozniewski:
Ah I see. I thought, they share this part. Will quickly test and update

slava77:
@cmsbuild please test 136.7801,136.7803

slava77:
@cmsbuild please test workflow 136.7801,136.7803

silviodonato:
urgent

silviodonato:
+operations

jfernan2:
@swozniewski The change in DQMOffline/Trigger/python/HLTTauDQMOffline_cfi.py is producing a  lost in the entries for the denominator taken to calculate efficiencies in HLT / TAU / PFTaus / folder
Is this understood? Thanks

slava77:
> @swozniewski The change in DQMOffline/Trigger/python/HLTTauDQMOffline_cfi.py is producing a lost in the entries for the denominator taken to calculate efficiencies in HLT / TAU / PFTaus / folder
> Is this understood? Thanks

In the comparisons: DQMHistoTests: Total failures: 1
Did you look at the older/buggy diff?

swozniewski:
Does your question refer to the workflows fixed in this PR or to the original PR? Because here, we use the same inputs now as before and there should be no difference. If the new data format is used, there may be small changes due to the usage of ByLooseCombinedIsolationDBSumPtCorr3Hits because its configuration was unified with other cut based IDs. But the change should only be very small. So it depends on how large the loss is that you observe.

kpedro88:
+upgrade

jfernan2:
+dqm
Sorry, it seems I looked to the comparison from 18h ago, not the last one 14h ago. Apologies

chayanit:
+1

silviodonato:
+1

Issue #29474, opened 2020-04-15
SLHCUpgradeSimulations/Geometry unit test failure 


silviodonato:
There is a failing unit test in `SLHCUpgradeSimulations/Geometry` since `CMSSW_11_1_X_2020-04-14-2300`, after the integration of 
```
===== Test "testPhase2PixelNtuple" ====
argument 0: testPhase2PixelNtuple
argument 1: /bin/bash
argument 2: SLHCUpgradeSimulations/Geometry/test
argument 3: phase2_digi_reco_pixelntuple.sh
shell is: /bin/bash
Current directory is: /data/cmsbld/jenkins/workspace/ib-run-qa/CMSSW_11_1_X_2020-04-14-2300
topdir is: /data/cmsbld/jenkins/workspace/ib-run-qa/CMSSW_11_1_X_2020-04-14-2300
testdir is: /data/cmsbld/jenkins/workspace/ib-run-qa/CMSSW_11_1_X_2020-04-14-2300/src/SLHCUpgradeSimulations/Geometry/test
tmpdir is: /data/cmsbld/jenkins/workspace/ib-run-qa/CMSSW_11_1_X_2020-04-14-2300/tmp/cc8_amd64_gcc8
testbin is: /data/cmsbld/jenkins/workspace/ib-run-qa/CMSSW_11_1_X_2020-04-14-2300/test/cc8_amd64_gcc8
Running script: /data/cmsbld/jenkins/workspace/ib-run-qa/CMSSW_11_1_X_2020-04-14-2300/src/SLHCUpgradeSimulations/Geometry/test/phase2_digi_reco_pixelntuple.sh
15-Apr-2020 04:00:49 CEST  Initiating request to open file root://eoscms.cern.ch//eos/cms/store/user/cmsbuild/store/relval/CMSSW_10_6_0_patch2/RelValSingleMuPt10/GEN-SIM/106X_upgrade2023_realistic_v3_2023D41noPU-v1/10000/7377ED92-245C-CC4D-9F05-25ABB5522A08.root
15-Apr-2020 04:00:56 CEST  Successfully opened file root://eoscms.cern.ch//eos/cms/store/user/cmsbuild/store/relval/CMSSW_10_6_0_patch2/RelValSingleMuPt10/GEN-SIM/106X_upgrade2023_realistic_v3_2023D41noPU-v1/10000/7377ED92-245C-CC4D-9F05-25ABB5522A08.root
[...]

PR #29476, opened 2020-04-15
Fix a TrackerDTCProducer input tag in SLHCUpgradeSimulations/Geometry unit test

silviodonato:
This is a trivial way to solve #29474.
I would prefer the L1T experts fix `L1Trigger/TrackerDTC/src/Settings.cc` to avoid this kind of error.
Please note that special InputTag such as `@currentProcess` does not work because of `Settings::checkConfiguration()` in  `L1Trigger/TrackerDTC/src/Settings.cc`

#### PR validation:
`cmsRun SLHCUpgradeSimulations/Geometry/test/phase2_digi_reco_pixelntuple_cfg.py`


silviodonato:
another PR will arrive from @tschuh see #29474

PR #29482, opened 2020-04-15
Phase2-hgx235 Rearrange the files to enable both DDD and DD4Hep to work

bsunanda:
#### PR description:

Rearrange the files to enable both DDD and DD4Hep to work

#### PR validation:

Use standard runTheMatrix.py

Nothing special


bsunanda:
@cmsbuild Please test

cvuosalo:
@bsunanda Will this PR fix the `zHGCalEE1` error Sergio encountered?

cvuosalo:
@bsunanda Why does the `hfnose1` directory have:
```
v0
v1
v1a
v2
v3
```
and this PR adds `v1a`? Shouldn't you add new directories in numeric sequence?

bsunanda:
I don't think it will fix any thing. I do not see the error which Sergio sees. Of course I did not try sim step - but I access the parameters defined in hgcalCons in all Phase2 scenario.

I named v1a which is v1 corrected while v3 is v2 corrected. May be I should name them v1c and v2c. Please suggest

cvuosalo:
@bsunanda Are you saying there are two parallel lines of development?
v0 -> v1 -> v1a
and
v0 -> v1-> v2 -> v3
So both v1a and v3 would be in use?

bsunanda:
@cvuosalo You are absolutely right: v0->v1->v1a and v0->v1->v2->v3 so 2 versions will be there for the ones used: v1a and V3

cvuosalo:
@bsunanda I don't think the versioning scheme foresaw parallel lines of development. How about creating two directories with explanatory names in the `hfnose` directory, one for the v1 -> v1a line and one for the v2 -> v3 line? These two directories would then contain simple v1, v2, etc. directories. In this way, the parallel lines of development would be reflected in the directory structure.

bsunanda:
@cvuosalo v0->v1->v1a is not a development line. This is created just to compare the v1 version to work with dd4hep. The development cycle will be v1->v2->v3. It will most likely soon become v4. They are not parallel. v1a is made just to have the older version alive with dd4hep and this version is widely used for the studies at the moment.

cvuosalo:
+1

kpedro88:
+upgrade
no changes observed in any phase 2 workflows

silviodonato:
+1

PR #29492, opened 2020-04-16
Decrease begin Run startup time for HLT

Dr15Jones:
#### PR description:

Running igprof on an example HLT configuration uncovered that the vast amount of time spent in the begin Run transition was all the ShmStreamConsumer OutputModules calculating the ParameterSetBlobs to be stored in the files. This pull request adds the option to run ParameterSetBlobProducer at begin Run to create the ParameterSetBlobs once and then have all the OutputModules use that information. This decreased the time spent making the blobs from 147s in the orginal job to 4s.

#### PR validation:

The test configuration I was using ran fine both with and without ParameterSetBlobProducer added.

### DIFF
--- a/DataFormats/Common/src/classes.h
+++ b/DataFormats/Common/src/classes.h
@@ -30,7 +30,10 @@
 #include "DataFormats/Common/interface/ThinnedAssociation.h"
 #include "DataFormats/Common/interface/SecondaryEventIDAndFileInfo.h"
 #include "DataFormats/Provenance/interface/EventAuxiliary.h"
+#include "DataFormats/Provenance/interface/ParameterSetBlob.h"
+#include "DataFormats/Provenance/interface/ParameterSetID.h"
 #include "FWCore/MessageLogger/interface/ErrorSummaryEntry.h"
 #include "DataFormats/Common/interface/RandomNumberGeneratorState.h"
 
 #include <vector>
+#include <map>
--- a/DataFormats/Common/src/classes_def.xml
+++ b/DataFormats/Common/src/classes_def.xml
@@ -190,5 +190,5 @@
  <class name="edm::RandomNumberGeneratorState" ClassVersion="3">
   <version ClassVersion="3" checksum="2237762164"/>
  </class>
-
+ <class name="edm::Wrapper<std::map<edm::ParameterSetID,edm::ParameterSetBlob>>"/>
 </lcgdict>
--- a/DataFormats/Provenance/src/classes_def.xml
+++ b/DataFormats/Provenance/src/classes_def.xml
@@ -156,6 +156,7 @@
  <class name="std::vector<std::vector<edm::ParameterSetID > >"/>
  <class name="std::map<edm::ParameterSetID,edm::ParameterSetBlob>"/>
  <class name="std::pair<edm::ParameterSetID,edm::ParameterSetBlob>"/>
+ <class name="std::pair<const edm::ParameterSetID,edm::ParameterSetBlob>"/>
  <class name="edm::ProcessConfigurationID"/>
  <class name="std::vector<edm::ProcessConfigurationID>"/>
  <class name="edm::ParentageID"/>
--- a/EventFilter/Utilities/interface/EvFOutputModule.h
+++ b/EventFilter/Utilities/interface/EvFOutputModule.h
@@ -5,6 +5,7 @@
 #include "FWCore/Framework/interface/one/OutputModule.h"
 #include "IOPool/Streamer/interface/StreamerOutputModuleCommon.h"
 #include "FWCore/Utilities/interface/EDGetToken.h"
+#include "DataFormats/Streamer/interface/StreamedProducts.h"
 
 #include "EventFilter/Utilities/interface/JsonMonitorable.h"
 #include "EventFilter/Utilities/interface/FastMonitor.h"
@@ -93,6 +94,7 @@ namespace evf {
     edm::ParameterSet const& ps_;
     std::string streamLabel_;
     edm::EDGetTokenT<edm::TriggerResults> trToken_;
+    edm::EDGetTokenT<edm::SendJobHeader::ParameterSetMap> psetToken_;
 
     evf::FastMonitoringService* fms_;
 
--- a/EventFilter/Utilities/src/EvFOutputModule.cc
+++ b/EventFilter/Utilities/src/EvFOutputModule.cc
@@ -109,7 +109,9 @@ namespace evf {
         EvFOutputModuleType(ps),
         ps_(ps),
         streamLabel_(ps.getParameter<std::string>("@module_label")),
-        trToken_(consumes<edm::TriggerResults>(edm::InputTag("TriggerResults"))) {
+        trToken_(consumes<edm::TriggerResults>(edm::InputTag("TriggerResults"))),
+        psetToken_(consumes<edm::SendJobHeader::ParameterSetMap, edm::InRun>(
+            ps.getUntrackedParameter<edm::InputTag>("psetMap"))) {
     //replace hltOutoputA with stream if the HLT menu uses this convention
     std::string testPrefix = "hltOutput";
     if (streamLabel_.find(testPrefix) == 0)
@@ -138,6 +140,8 @@ namespace evf {
     edm::ParameterSetDescription desc;
     edm::StreamerOutputModuleCommon::fillDescription(desc);
     EvFOutputModuleType::fillDescription(desc);
+    desc.addUntracked<edm::InputTag>("psetMap", {"psetMap"})
+        ->setComment("Optionally allow the map of ParameterSets to be calculated externally.");
     descriptions.addDefault(desc);
   }
 
@@ -153,13 +157,16 @@ namespace evf {
     uint32 preamble_adler32 = 1;
     edm::BranchIDLists const* bidlPtr = branchIDLists();
 
+    auto psetMapHandle = run.getHandle(psetToken_);
+
     std::unique_ptr<InitMsgBuilder> init_message =
         jsonWriter_->streamerCommon_.serializeRegistry(*jsonWriter_->streamerCommon_.getSerializerBuffer(),
                                                        *bidlPtr,
                                                        *thinnedAssociationsHelper(),
                                                        OutputModule::processName(),
                                                        description().moduleLabel(),
-                                                       moduleDescription().mainParameterSetID());
+                                                       moduleDescription().mainParameterSetID(),
+                                                       psetMapHandle.isValid() ? psetMapHandle.product() : nullptr);
 
     //Let us turn it into a View
     InitMsgView view(init_message->startAddress());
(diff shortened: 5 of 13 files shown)


Dr15Jones:
please test

Dr15Jones:
@fwyzard FYI I found this from the configuration you sent me.

fwyzard:
thanks, looks like an interesting improvement !

Dr15Jones:
In order to turn on the feature, I added the following to the test configuration I was using
```python
process.psetMap = cms.EDProducer("ParameterSetBlobProducer")
process.PhysicsMuonsOutput.associate(cms.Task(process.psetMap))
```

I.e. I added the EDProducer as a Task to one of the EndPaths, that made it available to all of the OutputModules.

mommsen:
This sounds like a great improvement. Thanks for looking into this.
@smorovic, @emeschi, this could be of interest for you, too.

emeschi:
Yes, very interesting. Do I get it right that the whole configuration is stored as a blob in every output file ?
Is this really necessary ?
If the numbers quote by   @Dr15Jones<https://github.com/Dr15Jones> are for a realistic HLT menu we should re-evaluate the actual startup time and see if we can say something new about the amount of input buffer needed in the HLT…
To be followed up


On 16 Apr 2020, at 17:08, Remi Mommsen <notifications@github.com<mailto:notifications@github.com>> wrote:



This sounds like a great improvement. Thanks for looking into this.
@smorovic<https://github.com/smorovic>, @emeschi<https://github.com/emeschi>, this could be of interest for you, too.

smorovic:
Thanks for looking into it.
@emeschi, if I understand well, it is run product so it isn't written in every file or event (just INI file).

I can look at intialization time on HLT machine(s). @fwyzard, could you please point to this HLT python configuration? Is it similar to collision physics menu (i.e. something with similar number of modules, paths and output modules as we could have in Run 3)?

emeschi:
Ah yes you’re right @smorovic. I had completely forgotten the init file.
Okay so the solution proposed is to only serialise the pest once and then have each output module get it from the run, right @Dr15Jones<https://github.com/Dr15Jones> ?
This means we had roughly 147/4-1~36 output modules in this config doing the same thing again and again…
That I guess sounds about right as the number of streams…

Dr15Jones:
The relevant bit of the igprof log of the original version of the code is
```
[17]       59.6     154.72       0.00 / 154.72     edm::StreamerOutputModuleBase::beginRun(edm::RunForOutput const&)
           58.0  .........     150.65 / 150.65       edm::StreamerOutputModuleCommon::serializeRegistry(SerializeDataBuffer&, std::vector<std::vector<unsigned int, std::allocator<unsigned int> >, std::allocator<std::vector<unsigned int, std::allocator<unsigned int> > > > const&, edm::ThinnedAssociationsHelper const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, edm::Hash<1> const&) [18]

[18]       58.0     150.65       0.00 / 150.65     edm::StreamerOutputModuleCommon::serializeRegistry(SerializeDataBuffer&, std::vector<std::vector<unsigned int, std::allocator<unsigned int> >, std::allocator<std::vector<unsigned int, std::allocator<unsigned int> > > > const&, edm::ThinnedAssociationsHelper const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, edm::Hash<1> const&)
           57.9  .........     150.37 / 150.37       edm::StreamSerializer::serializeRegistry(SerializeDataBuffer&, std::vector<std::vector<unsigned int, std::allocator<unsigned int> >, std::allocator<std::vector<unsigned int, std::allocator<unsigned int> > > > const&, edm::ThinnedAssociationsHelper const&) [19]

[19]       57.9     150.37       0.00 / 150.37     edm::StreamSerializer::serializeRegistry(SerializeDataBuffer&, std::vector<std::vector<unsigned int, std::allocator<unsigned int> >, std::allocator<std::vector<unsigned int, std::allocator<unsigned int> > > > const&, edm::ThinnedAssociationsHelper const&)
           56.8  .........     147.46 / 147.46       edm::pset::Registry::fillMap(std::map<edm::Hash<1>, edm::ParameterSetBlob, std::less<edm::Hash<1> >, std::allocator<std::pair<edm::Hash<1> const, edm::ParameterSetBlob> > >&) const [20]

```

Dr15Jones:
> This means we had roughly 147/4-1~36 output modules in this config doing the same thing again and again…

Doing
```python
>>> len(process.outputModules)
36
```

shows your math is exactly right :).

mommsen:
I think we should do a test with this change under realistic conditions asap. Not sure how to do it, though. @fwyzard, @smorovic, any suggestions?

If the speedup is indeed confirmed in HLT (which needs to load the conditions, too), it could save us quite a bit of money otherwise spent on DIMMs for the RAMdisk to buffer data during the CMSSW startup on the HLT farm.

fwyzard:
> @fwyzard, could you please point to this HLT python configuration? Is it similar to collision physics menu (i.e. something with similar number of modules, paths and output modules as we could have in Run 3)?

@smorovic yes, it's a "frozen" version of an HLT menu from 2018, adapted to run with CMSSW 11.1.x (though I later realised that DQM and JSON monitoring is missing).
Here is the [hlt.py](https://fwyzard.web.cern.ch/fwyzard/hlt_timing/hlt.py) and a [data.tar.xz](https://fwyzard.web.cern.ch/fwyzard/hlt_timing/data.tar.xz) archive with the input in FED raw data format.

fwyzard:
> I think we should do a test with this change under realistic conditions asap. Not sure how to do it, though.

I can run a realistic test once 11.1.0-pre6 is available (to use as a baseline).

By realistic I mean
  - 2018 HLT menu and L1-selected data at pileup ~50
  - using `.raw` input and `.dat` output
  - with 16 jobs with 4 threads each, running in parallel on a hilton-like machine

mommsen:
One more point: the content of the blob (i.e. the init file) depends on the products which are written by the output module, IIRC. Is this taken into account with these changes @Dr15Jones ? It also implies that not all output modules can reuse the same blob.

Dr15Jones:
> One more point: the content of the blob (i.e. the init file) depends on the products which are written by the output module, IIRC. Is this taken into account with these changes @Dr15Jones ? It also implies that not all output modules can reuse the same blob.

The part I did was the part they all use, the storage of the ParameterSets.

smorovic:
@fwyzard thanks for the files.
We should install 11_1_0_pre6 in Hilton and BU appliance that I use for tests as soon as it's available. It was being prepared this morning so we can expect it soon.
I can run tests with scripts in BU appliance. It is equivalent to Hilton tests so we can compare (but I expect the same conclusion). In appliance test we can see init period from the microstate plot. Event sample won't matter since this all comes before first event.

makortel:
+core

mommsen:
+1

Martin-Grunewald:
@Dr15Jones 
Since ConfDB is not yet task-aware, how should this new producer be included in a menu?
Adding it to each EndPath with an OutputModule?

fwyzard:
I would add it to the `HLTriggerFirstPath`, together with `hltGetConditions` and `hltGetRaw` .

Dr15Jones:
> I would add it to the HLTriggerFirstPath, together with hltGetConditions and hltGetRaw .

I think Andrea's suggestion is a good one. It actually doesn't much matter since during Run and LuminosityBlock transitions modules are run in data dependency order, not in strict Path order.

perrotta:
+1
- It adds the option to run ParameterSetBlobProducer at begin Run to create the ParameterSetBlobs once, and then have all the OutputModules use that information: this may speed up beginRun startup time for HLT: that option is not enabled by default
- Reco only affected by the needed modifications (new parameter added) in EventFilter/Utilities/EvFOutputModule
- Jenkins tests pass

fwyzard:
On a job with 4 threads I see a reduction in the startup time (measured from the `%MSG-i ThreadStreamSetup
[diff.txt](https://github.com/cms-sw/cmssw/files/4493310/diff.txt)
 pre-events` message to the `Begin processing the 1st record` message) from `106 ± 2 s` to `62 ± 2 s`.

The diff to the hlt configuration is attached.

@Dr15Jones do you think this is coherent with your results ?

Dr15Jones:
> do you think this is coherent with your results ?

There are two things happening
1. when using the old code under multi-threading several output modules could run concurrently at begin Run and just doing the exact same work so the biggest benefit happens at 1 thread in a job (since everything is serialized and the doing the same work over and over hurts the most). When using multiple threads, if the framework can find more work to do (based on data dependencies) then you can get a bigger 'win'.
1. conditions access can become the dominant timing for begin Run (I saw large variations in timing because of different response time from the frontier servers).

The work I'm trying to do for allowing EventSetup modules to run concurrently should also allow the framework to do better scheduling at begin Run.

fwyzard:
OK, I'll try again
  - with a single threaded job
  - with a multi-threaded job
  - with a single threaded job, without any actual events and conditions data
  - with a multi-threaded job, without any actual events and conditions data

smorovic:
I get similar result with BU-FU appliance. 
With12 FUs, 4-streams/thread processes (1 thread per hyperthread). I did two tries each case.

116 s, 113 s (no blob producer module)
73 s, 78 s (module added)

Average improvement: 39 seconds.
Note:  one HLT machine took longer than others, but only in "no blob producer" case. I ignored it for the result.

As far as I know, squid cache expiration is 30 seconds in HLT, and repeated attempts don't happen within that time frame.

fwyzard:
> As far as I know, squid cache expiration is 30 seconds in HLT, and repeated attempts don't happen within that time frame.

Is that only for the data with a short lifetime (i.e. IOVs) or also for the actual immutable payloads ?

Dr15Jones:
Another thing to keep in mind is when accessing data from the EventSetup the module takes a lock. The framework doesn't know about the lock so in multi-threading the framework can schedule multiple modules who all want the EventSetup lock so no progress gets made on some of the threads while waiting for the lock. (This is again why I'm working on running EventSetup modules concurrently and why we are adding 'consumes' to modules for EventSetup products.) The TimeReport does say how much time was spent waiting on the EventSetup lock.

smorovic:
@fwyzard  good question. It was discussed in context of lumi-based conditions and Dave Dykstra mentioned it, but did not specify if there is different setting for IOVs and payloads.

fwyzard:
For a single threaded job the time drops from `150 ± 20s (best: 138s)` to `63 ± 1s (best: 62s)`, with a gain of `-87 ± 20s (best: 76s)`.

For a multi threaded job the time drops from `107 ± 3s (best: 103s)` to `62 ± 2s (best: 60s)`, with a gain of `-45 ± 3s (best: 43s)`.

So, the startup time with the new approach is independent of using multiple threads, while the original approach had some benefit from it (as expected following the explanation by Chris).

In any case, the improvement is impressive :-)

silviodonato:
+1
wow! Impressive improvement!

PR #29461, opened 2020-04-12
Re-add SiPixelPhase1TrackResidualsAnalyzer to HI scenario

mmusich:
resolves https://github.com/cms-sw/cmssw/issues/29460

#### PR description:

Title says it all.

#### PR validation:

Tested wf 150.0: `runTheMatrix.py -l 150.0 -t 4 -j 8` in `CMSSW_11_1_X_2020-04-12-0000` with this branch which now runs to completion.

#### if this PR is a backport please specify the original PR and why you need to backport that PR:

This PR is not a backport and no backport is meant.
Attn: @dmeuser @arossi83 @sroychow

PR #20496, opened 2017-09-13
Fix hi phase1

fioriNTU:
This PR should fix a recent crash in the HI validation workflow originating by the fact that the DQM step1 was using the Phase0 code while the harvesting was using Phase1.

Now the era mechanism is correctly used for the HI. 

@prebello @fabozzi @dmitrijus

PR #29495, opened 2020-04-16
re-add also SiPixelPhase1RecHitsAnalyzer to DQM offline for HI scenario

mmusich:
#### PR description:

While reviewing the code for PR #29461 I noticed also that some other DQM modules (`SiPixelPhase1RecHitsAnalyzer`) were excluded from being run without apparent reason.
I re-add it here, together with the harvester for the residuals.

#### PR validation:

Standard limited matrix integration tests passed + 
```
runTheMatrix.py -l 150.0 -t 4 -j 8
```

This is not a backport.


jfernan2:
please test

jfernan2:
+1

silviodonato:
Let me add in cc @fioriNTU who added those lines long time ago with #20496.

fioriNTU:
> Let me add in cc @fioriNTU who added those lines long time ago with #20496.

@silviodonato and @mmusich, sorry I should have commented before, but I simply forgot. I do not remember why I excluded the residuals plugin form HI, sorry. For the RecHit plugin, I did it, because probably at that time the code was doing exactly the same things done by the TrackCluster plugin, but my memory is not that strong in general :-)
Anyway, in that PR, I added for the first time the HI processing sequence in PixelPhase1, which was not available yet. So, probably I had some weird crash for the residuals plugin, and I simply decided that was enough without. Then, as far as I remember, for the last HI run we have used the pp configuration, so probablly nobody touched that code anymore because is obsolete in production (but maybe not in Validation). Hope this helps.

mmusich:
Indeed the only wf which uses this is 150.0. This is not meant for production but as long as this scenario is supported in release should be monitored.

silviodonato:
+1

PR #29513, opened 2020-04-19
decrease event size by storing PFCandidates with half precision float…

jsalfeld:
… point format

#### PR description:

As mentioned in a TSG on Jan. 15th:
https://indico.cern.ch/event/878772/contributions/3703563/attachments/1969842/3276766/ScoutingGroup.pdf

The PFScouting event size is reduced by 40% when PFCandidate 4-vectors are stored in Hal precision floating point format. I used the libminifloat class in CMSSW and convert back to float32 (the type returned by float32to16 is uint16_t) and let the compression reduce the information being stored.

#### PR validation:

I was running with and without the change and compared the output, and it looks the same. The per event size of the corresponding collection is indeed reduced after checking via edmEventSize.

File outputScoutingPF.root Events 10
**float16:** ScoutingParticles_hltScoutingPFPacker__HLT2018. 24530.5 7804.4
**float32:** ScoutingParticles_hltScoutingPFPacker__HLT2018. 24530.5 12444.5

### DIFF
--- a/HLTrigger/JetMET/plugins/HLTScoutingPFProducer.cc
+++ b/HLTrigger/JetMET/plugins/HLTScoutingPFProducer.cc
@@ -62,7 +62,7 @@ private:
   const double pfJetEtaCut;
   const double pfCandidatePtCut;
   const double pfCandidateEtaCut;
-  const bool storeWithHalfPrecisionFloatingPoint;
+  const int mantissaPrecision;
 
   const bool doJetTags;
   const bool doCandidates;
@@ -84,7 +84,7 @@ HLTScoutingPFProducer::HLTScoutingPFProducer(const edm::ParameterSet &iConfig)
       pfJetEtaCut(iConfig.getParameter<double>("pfJetEtaCut")),
       pfCandidatePtCut(iConfig.getParameter<double>("pfCandidatePtCut")),
       pfCandidateEtaCut(iConfig.getParameter<double>("pfCandidateEtaCut")),
-      storeWithHalfPrecisionFloatingPoint(iConfig.getParameter<bool>("storeWithHalfPrecisionFloatingPoint")),
+      mantissaPrecision(iConfig.getParameter<int>("mantissaPrecision")),
       doJetTags(iConfig.getParameter<bool>("doJetTags")),
       doCandidates(iConfig.getParameter<bool>("doCandidates")),
       doMet(iConfig.getParameter<bool>("doMet")) {
@@ -156,16 +156,12 @@ void HLTScoutingPFProducer::produce(edm::StreamID sid, edm::Event &iEvent, edm::
           ++index_counter;
         }
 
-        if (storeWithHalfPrecisionFloatingPoint) {
-          outPFCandidates->emplace_back(MiniFloatConverter::float16to32(MiniFloatConverter::float32to16(cand.pt())),
-                                        MiniFloatConverter::float16to32(MiniFloatConverter::float32to16(cand.eta())),
-                                        MiniFloatConverter::float16to32(MiniFloatConverter::float32to16(cand.phi())),
-                                        MiniFloatConverter::float16to32(MiniFloatConverter::float32to16(cand.mass())),
-                                        cand.pdgId(),
-                                        vertex_index);
-        } else {
-          outPFCandidates->emplace_back(cand.pt(), cand.eta(), cand.phi(), cand.mass(), cand.pdgId(), vertex_index);
-        }
+        outPFCandidates->emplace_back(MiniFloatConverter::reduceMantissaToNbitsRounding(cand.pt(), mantissaPrecision),
+                                      MiniFloatConverter::reduceMantissaToNbitsRounding(cand.eta(), mantissaPrecision),
+                                      MiniFloatConverter::reduceMantissaToNbitsRounding(cand.phi(), mantissaPrecision),
+                                      MiniFloatConverter::reduceMantissaToNbitsRounding(cand.mass(), mantissaPrecision),
+                                      cand.pdgId(),
+                                      vertex_index);
       }
     }
   }
@@ -267,7 +263,7 @@ void HLTScoutingPFProducer::fillDescriptions(edm::ConfigurationDescriptions &des
   desc.add<double>("pfJetEtaCut", 3.0);
   desc.add<double>("pfCandidatePtCut", 0.6);
   desc.add<double>("pfCandidateEtaCut", 5.0);
-  desc.add<bool>("storeWithHalfPrecisionFloatingPoint", false);
+  desc.add<int>("mantissaPrecision", 23);
   desc.add<bool>("doJetTags", true);
   desc.add<bool>("doCandidates", true);
   desc.add<bool>("doMet", true);
--- a/HLTrigger/JetMET/plugins/HLTScoutingPFProducer.cc
+++ b/HLTrigger/JetMET/plugins/HLTScoutingPFProducer.cc
@@ -39,6 +39,8 @@ Description: Producer for ScoutingPFJets from reco::PFJet objects, ScoutingVerte
 
 #include "DataFormats/Math/interface/deltaR.h"
 
+#include "DataFormats/Math/interface/libminifloat.h"
+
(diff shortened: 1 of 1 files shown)


Martin-Grunewald:
please test

HLTrigger/JetMET/plugins/HLTScoutingPFProducer.cc
```diff
@@ -151,7 +155,17 @@ void HLTScoutingPFProducer::produce(edm::StreamID sid, edm::Event &iEvent, edm::
             break;
           ++index_counter;
         }
-        outPFCandidates->emplace_back(cand.pt(), cand.eta(), cand.phi(), cand.mass(), cand.pdgId(), vertex_index);
+
+        if (storeWithHalfPrecisionFloatingPoint) {
+          outPFCandidates->emplace_back(MiniFloatConverter::float16to32(MiniFloatConverter::float32to16(cand.pt())),
+                                        MiniFloatConverter::float16to32(MiniFloatConverter::float32to16(cand.eta())),
+                                        MiniFloatConverter::float16to32(MiniFloatConverter::float32to16(cand.phi())),
+                                        MiniFloatConverter::float16to32(MiniFloatConverter::float32to16(cand.mass())),
```
fwyzard, line 163: is there no way to just truncate the precision, without converting to 16 bits and back to 32 bits ?

VinInn: of course there is: it is called bfloat16 (instead of halffloat)
https://en.wikipedia.org/wiki/Bfloat16_floating-point_format

BUT it really has low precision

bfloat16 is what it is mostly used for AI.

On Intel machines (AVX2 and more) one could use intrinsics to perform the conversion to/from halffloat (worth only if contiguous as in a SOA)

fwyzard: Mhm,  no, that doesn't seem like a good candidate here.

I just meant to ask if there is way to do `MiniFloatConverter::float16to32(MiniFloatConverter::float32to16(x))` in a single call.

VinInn: under the assumption that the first call did not over/underflow it is trivial yes
(well, one must properly round, not truncate. still pretty trivial).

but at this point I do not understand what is going on, I was thinking data were stored in hf16 (aka int16), not in truncated float32
(ok I missed "let the compression reduce the information being stored.")

fwyzard: > but at this point I do not understand what is going on, I was thinking data were stored in hf16 (aka int16), not in truncated float32

that was my assumption as well

on the other hand, storing the truncated precision back in a float32 has the advantage of not affecting the data format, and apparently the compression is still good enough to reduce the size by ~37%

VinInn:
what about this
```
#include<iostream>
#include<iomanip>
#include<cmath>
#include<limits>
#include<cstdio>
#include<cstring>

template<typename T>
void print(T x) {
 std::cout<< std::hexfloat << x <<' '<<  std::scientific << std::setprecision(8) << x << ' ' <<  std::defaultfloat << x << std::endl;
}



template<int N>
float roundit(float x) {
  static_assert(N<23);
  constexpr auto shift = 23-N;
  constexpr uint32_t mask = 1<<(shift-1);

  uint32_t i; memcpy(&i, &x, sizeof(x));

  i += (i>=0) ? mask : -mask;
  memcpy(&i, &x, sizeof(x));
  i>>=shift; i<<=shift;
  memcpy(&x, &i, sizeof(x));

  return x;
}



int main() {


  print(roundit<8>(M_PI));
  print(roundit<12>(M_PI));
  print(roundit<16>(M_PI));
  print(roundit<20>(M_PI));
  print(roundit<22>(M_PI));

  print(roundit<8>(-M_PI));
  print(roundit<12>(-M_PI));
  print(roundit<16>(-M_PI));
  print(roundit<20>(-M_PI));
  print(roundit<22>(-M_PI));


  return 0;
}
```

fwyzard:
[`DataFormats/Math/interface/libminifloat.h`](https://github.com/cms-sw/cmssw/blob/master/DataFormats/Math/interface/libminifloat.h) seems to do a bit more checks for rounding, denormals, etc.

VinInn:
denormals better to flush to zero (still my little code will handle them)
one could check for inf and NaN... (in my code inf may become NaN)

VinInn:
I made a PR with my little code (including inf/NaN protection): one can choose the number of bits at will. (actually it can be changed in a runtime parameters instead of a template at the cost of some additional instructions.

jsalfeld:
Thanks a lot for the suggestions, a couple of comments:

1.) preferring precision over range (bfloat16 vs float16) would probably be more suitable, if I understand correctly up to 2048 GeV the float16 precision is ~<0.1%
2.) One reason why the per event size reduction of the PFCandidate collection is less than 50%, is because of the fact that "vertex_index" and "pdgId" of PFCandidates stay unchanged. Also please not that I was only running over 10 events
3.) I checked that one could indeed also change float to float16 in DataFormats/Scouting/interface/ScoutingParticle.h, which would reduce the event size. In principle we could also implement a second "ScoutingParticle.h" format with float16
4.) I would currently think it is beneficial to store the actual values instead of uint16_t, which would need to be converted at a later stage I believe
5.) the roundit function above and in PR #29518 looks nicer than what I do above with the libminifloat.h class. If you prefer it also, please let me know.

Please let me know about your preferences/suggestions and I will adopt

jsalfeld:
ok if I understand #29518 correctly I should use ReduceMantissaToNbitsRounding
for example https://github.com/cms-sw/cmssw/blob/master/DataFormats/Math/interface/libminifloat.h#L107

fwyzard:
> 1.) preferring precision over range (bfloat16 vs float16) would probably be more suitable, if I understand correctly up to 2048 GeV the float16 precision is ~<0.1%

I think it's the opposite: `bfloat16` has larger range and smaller precision (7+1 bit), while `float16` has smaller range and larger precision (10+1 bit):
  - with `bfloat16` you can store 2000 GeV with a granularity of 8 GeV;
  - with `float16` you can store 2000 GeV with a granularity of 1 GeV.

>  4.) I would currently think it is beneficial to store the actual values instead of uint16_t, which would need to be converted at a later stage I believe

We should use whatever is more compact after compression.
The dataformat itself can take care of doing the conversion on the fly when reading.

fwyzard:
> ok if I understand #29518 correctly I should use `ReduceMantissaToNbitsRounding`

Intersesting.
Could you check that using that gives the same results (both in terms of physics and compression) as what you are doing now ?

VinInn:
Yes, try ReduceMantissaToNbitsRounding with 10 bits: should be binary identical to what you are doing now.

VinInn:
BTW: I do not understand the need of such complex code as in ReduceMantissaToNbitsRounding.
Mine is just based on trivial integer arithmetic (as I said just two instructions).
Maybe the author/mantiners of ReduceMantissaToNbitsRounding should give a look

silviodonato:
Status: waiting for an update from @jsalfeld (moving to `reduceMantissaToNbitsRounding'`)

silviodonato:
I created an issue ( #29564) where you can discuss about the possible improvement of `ReduceMantissaToNbitsRounding`

jsalfeld:
sorry for the delay, I switched to MiniFloatConverter::reduceMantissaToNbitsRounding. And checked that the output matches the one from "MiniFloatConverter::float16to32(MiniFloatConverter::float32to16())" with 10 bits mantissa via Events->Scan(). The size is similar:
File outputScoutingPF.root Events 120
**MantissaRounding**: ScoutingParticles_hltScoutingPFPacker__HLT2018. 25165.4 7747.8
**ConversionReConversion**: ScoutingParticles_hltScoutingPFPacker__HLT2018. 25165.4 7487.35

I this is good then we can adjust the precision as we want.

silviodonato:
please test

Martin-Grunewald:
+1

silviodonato:
+1

PR #29518, opened 2020-04-20
add a fast and simple rounding function with given precision

VinInn:
As discussed in #29513 it would be useful to have a fast and simple function to round a floating point number to a given precision of N bits.

here is it!

compiler will simplify it in just 2 integer instructions
https://godbolt.org/z/rpBkVS

### DIFF
--- a/DataFormats/Math/interface/roundAtPrecision.h
+++ b/DataFormats/Math/interface/roundAtPrecision.h
@@ -1,0 +1,34 @@
+#ifndef DataFormatsMathroundAtPrecision_H
+#define DataFormatsMathroundAtPrecision_H
+
+#include "FWCore/Utilities/interface/isFinite.h"
+#include<cstring>
+#include<cstdint>
+
+// round with N significant bits
+template<int N>
+float roundAtPrecision(float x) {
+
+  if (edm::isNotFinite(x)) return x;
+
+  static_assert(N<23);
+  constexpr auto shift = 23-N;
+  constexpr uint32_t mask = 1<<(shift-1);
+
+  uint32_t i; memcpy(&i, &x, sizeof(x));
+
+  i += mask;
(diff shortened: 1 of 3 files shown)


DataFormats/Math/interface/roundAtPrecision.h
```diff
@@ -0,0 +1,34 @@
...
+#define DataFormatsMathroundAtPrecision_H
+
+#include "FWCore/Utilities/interface/isFinite.h"
+#include <cstring>
+#include <cstdint>
+
+// round with N significant bits
+template <int N>
+float roundAtPrecision(float x) {
+  if (edm::isNotFinite(x))
```
slava77, line 11: other than this, what is the performance difference wrt reduceMantissaToNbits in libminifloat?
I'd rather not proliferate. If this is better and the isNotFinite is the only difference, perhaps it's better to update reduceMantissaToNbits

VinInn:
duplicate of reduceMantissaToNbits

DataFormats/Math/interface/roundAtPrecision.h
VinInn: reduceMantissaToNbits just truncate.
reduceMantissaToNbitsRounding most probably is numerically equivalent even if the code looks more complex

slava77: but isn't this code actually a truncation?
I see the unit test ```   assert(roundAtPrecision<0>(2.9) == 2.f);``` (round(2.9) = 3)

VinInn: It is a binary rounding
with only exponent (0 bits) you have to choose between 2 and 4
so 2.9 rounds to 2 and pi to 4...

PR #29520, opened 2020-04-20
L1Trigger/TrackerDTC Debugged and Analyzer added.

tschuh:
#### PR description:

I added an EDAnalyzer to study the DTC performance and fixed some bugs I have introduced during initial PR review. The analyzer is not part of the standard sequences.


kpedro88:
please test

L1Trigger/TrackerDTC/interface/Settings.h
```diff
@@ -342,8 +379,10 @@ namespace trackerDTC {
 
     // derived event setup
 
-    // index = track trigger module id [0-15551], value = det id
-    std::vector<::DetId> cablingMap_;
+    // value = track trigger module id [0-15551], key = det id
+    std::map<::DetId, int> cablingMap_;
```
kpedro88, line 383: can this be unordered_map?

L1Trigger/TrackerDTC/interface/Settings.h
```diff
@@ -119,9 +117,26 @@ namespace trackerDTC {
 
     double bField() const { return bField_; }
 
+    // Analzyer
+
+    bool useMCTruth() const { return useMCTruth_; }
+    edm::InputTag inputTagTTClusterAssMap() const { return inputTagTTClusterAssMap_; }
+    std::string producerLabel() const { return producerLabel_; }
```
kpedro88, line 124: in general it's better to return complex objects by const ref, then the code that calls the accessor can decide if it wants to make a copy or keep a reference

L1Trigger/TrackerDTC/interface/TTDTCConverter.h
```diff
@@ -31,10 +31,12 @@ GlobalPoint TTDTCConverter(trackerDTC::Settings* settings, const TTDTC::Frame& f
     const std::vector<double>& diskZs = format->diskZs();
     const std::vector<double>& layerRs = format->layerRs();
 
-    const DetId detId = frame.first->getDetId();
+    const DetId& detId = frame.first->getDetId() + settings->offsetDetIdDSV();
```
kpedro88, line 34: not sure a ref makes sense here; you could do `const DetId detId(frame.first->getDetId() + settings->offsetDetIdDSV());` if you're concerned about avoiding an unnecessary copy

L1Trigger/TrackerDTC/plugins/Analyzer.cc
```diff
@@ -0,0 +1,508 @@
...
+  constexpr initializer_list<Efficiency> AllEfficiency = {Phi0, Pt, InvPt, D0, Z0, Eta};
+  constexpr auto NameEfficiency = {"Phi0", "Pt", "InvPt", "D0", "Z0", "Eta"};
+  inline string name(Efficiency e) { return string(*(NameEfficiency.begin() + e)); }
+
+  /*! \class  trackerDTC::Analyzer
+   *  \brief  Class to analyze hardware like structured TTStub Collection used by Track Trigger emulators
+   *  \author Thomas Schuh
+   *  \date   2020, Apr
+   */
+  class Analyzer : public one::EDAnalyzer<one::WatchRuns, one::SharedResources> {
```
kpedro88, line 76: this seems like it should migrate toward the validation subsystem eventually (as always, sooner is better than later)

tschuh: I will get in contact with the person making DQM plots in our group to asses if, what and how. Will come back when we have a plan but I cannot promise that we will do that quick enough for 11.1
Is it ok if that lives for the short term here?

kpedro88: Yes, we've allowed it before (on a temporary basis)

kpedro88:
please test

kpedro88:
+upgrade

silviodonato:
@rekovic do you have any comments?

tschuh:
I added two little conversion functions to the TTDTC product requested by downstream user. I hope that is ok to this in this PR while we wait.

silviodonato:
please test

kpedro88:
+upgrade

rekovic:
+1

silviodonato:
+1

PR #29389, opened 2020-04-03
Update of the lostTrack collection for MUO

sscruz:
#### PR description:

MUO POG need to store all the generalTracks with minimal kinematic cuts in order to perform efficiency measurements, as shown in these slides: https://indico.cern.ch/event/865993/contributions/3702398/. During the presentation it was suggested to enlarge the lostTrack collection to include the missing tracks, and we have found this alternative does not increase significantly the dataformat size, and is easier to use. 

This PR includes:
+ All tracks with pt > 2 that are not PF candidates are stored in the lostTrack collection.
+ lostTracks that are the innerTrack of a muon are assigned with a pdgId of +/- 13.
+ Module to cross-clean pfCandidates, lostTracks and slimmedMuons collections

https://github.com/cms-sw/cmssw/issues/27889

#### PR validation:
This PR has been validated with 136.8311

@dntaylor @gkaratha @peruzzim

PR #29529, opened 2020-04-21
Update of the lostTrack collection for MUO (back-port of #29389)

sscruz:
#### PR description:

Backport from #29389, update of the lostTrack collection for MUO 
https://github.com/cms-sw/cmssw/issues/27889
#### PR validation:

This PR has been validated with 136.8311

This PR is back-port from #29389, needs to be in 10_6_X for the UL re-miniAOD

### DIFF
--- a/PhysicsTools/PatAlgos/plugins/PATLostTracks.cc
+++ b/PhysicsTools/PatAlgos/plugins/PATLostTracks.cc
@@ -1,5 +1,6 @@
 #include <string>
 
+
 #include "DataFormats/Candidate/interface/Candidate.h"
 #include "DataFormats/ParticleFlowCandidate/interface/PFCandidate.h"
 #include "DataFormats/ParticleFlowCandidate/interface/PFCandidateFwd.h"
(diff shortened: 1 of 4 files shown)


perrotta:
@sscruz 
- Please edit the title of the PR with something more meaningful
- The last commit, which includes also the adjustments for the code checks, makes quite complicate to compare with the previous version of the code: is it necessary? Are there plans for further updates that suggest you to make identically the same the code in 10_6 and in the master?
  - If not, I would rather revert that commit
- Backports in a production release must satisfy the "no change" rule: the new ```lostTrack``` features should be made configurable and only switched on by the ```run2_miniAOD_devel``` modifier

perrotta:
please test

perrotta:
The output of the automatic tests confirms that there are no changes in the reco outputs with the default setting, as it should.

Going a bit further, let me notice that this PR adds:
- A "pass-through" cut to enlarge the sample of ```lostTracks```
- A muon id to those ```lostTracks```, that was missing before (only pions and electrons were considered)

Instead of having a single ```bool``` that steers both changes at once, I would rather customize differently:
- define the default value for ```passThroughCut``` as ```cms.string("false")```, which has to be reported to ```cms.string("pt>2")``` with the ```run2_miniAOD_devel``` modifier
- add a bool switch to be only applied at [line 250 of PATLostTracks.cc](https://github.com/cms-sw/cmssw/pull/29529/commits/47ecd7a34db8c385e1284a402db75b3e2452c324#diff-4cf30fc26fd4c278f2c4db56c61456d2R250); you can call it ```allowMuonId``` (e.g.), and set it "true" with the ```run2_miniAOD_devel``` modifier

As such, one can control separately the size of the ```lostTracks``` and the muonId in them.
What do you think?

sscruz:
> The output of the automatic tests confirms that there are no changes in the reco outputs with the default setting, as it should.
> 
> Going a bit further, let me notice that this PR adds:
> 
>     * A "pass-through" cut to enlarge the sample of `lostTracks`
> 
>     * A muon id to those `lostTracks`, that was missing before (only pions and electrons were considered)
> 
> 
> Instead of having a single `bool` that steers both changes at once, I would rather customize differently:
> 
>     * define the default value for `passThroughCut` as `cms.string("false")`, which has to be reported to `cms.string("pt>2")` with the `run2_miniAOD_devel` modifier
> 
>     * add a bool switch to be only applied at [line 250 of PATLostTracks.cc](https://github.com/cms-sw/cmssw/pull/29529/commits/47ecd7a34db8c385e1284a402db75b3e2452c324#diff-4cf30fc26fd4c278f2c4db56c61456d2R250); you can call it `allowMuonId` (e.g.), and set it "true" with the `run2_miniAOD_devel` modifier
> 
> 
> As such, one can control separately the size of the `lostTracks` and the muonId in them.
> What do you think?

sure, sounds good, i will push it in a bit

perrotta:
please test

perrotta:
backport of #29389

perrotta:
+1
- lostTracks collection is updated by enlarging the sample with looser cuts, and adding a muonID to those lostTracks
- That collection is not modified in the default config, being the update only enabled for the ```run2_miniAOD_devel``` modifier
- Size increase was evaluated for the master version of this PR and deemed acceptable
- Jenkins tests pass and show no differences with the default config, as it should
- Modifications to the lostTracks and isolatedTracks show up as expected when the ```run2_miniAOD_devel``` modifier is enabled

santocch:
+1

silviodonato:
+1

Issue #29532, opened 2020-04-22
DQM/Integration unit test failures in PY3


silviodonato:
We are getting 24 unit test failures in PY3 in DQM/Integration since CMSSW_11_1_PY3_X_2020-04-20-2300.
See https://cmssdt.cern.ch/SDT/cgi-bin/buildlogs/raw/slc7_amd64_gcc820/CMSSW_11_1_PY3_X_2020-04-20-2300/unitTestLogs/DQM/Integration
Most likely, the error is originated from #29372.

```
===== Test "TestDQMOnlineClient-scal_dqm_sourceclient" ====
+ [[ 1 -eq 0 ]]
+ [[ -z '' ]]
+ LOCAL_TEST_DIR=.
+ [[ -z '' ]]
+ CLIENTS_DIR=./src/DQM/Integration/python/clients
+ mkdir -p ./upload
+ cmsRun ./src/DQM/Integration/python/clients/scal_dqm_sourceclient-live_cfg.py unitTest=True
Querying DAS for files...
the query is file run=334393 dataset=/ExpressCosmics/Commissioning2019-Express-v1/FEVT lumi=1
DAS succeeded after 1 attempts 0
found files:  ['/store/express/Commissioning2019/ExpressCosmics/FEVT/Express-v1/000/334/393/00000/D0F052ED-9CA5-F547-BA73-2AA370D51AE8.root']
edmFileUtil --catalog file:/cvmfs/cms-ib.cern.ch/SITECONF/local/PhEDEx/storage.xml?protocol=xrootd --events /store/express/Commissioning2019/ExpressCosmics/FEVT/Express-v1/000/334/393/00000/D0F052ED-9CA5-F547-BA73-2AA370D51AE8.root | tail -n +9 | head -n -5 | awk '{ print $3 }'
----- Begin Fatal Exception 22-Apr-2020 03:46:48 CEST-----------------------
An exception of category 'ConfigFileReadError' occurred while
   [0] Processing the python configuration file named ./src/DQM/Integration/python/clients/scal_dqm_sourceclient-live_cfg.py
Exception Message:
 unknown python problem occurred.
TypeError: a bytes-like object is required, not 'str'

At:
  /cvmfs/cms-ib.cern.ch/nweek-02625/slc7_amd64_gcc820/cms/cmssw/CMSSW_11_1_PY3_X_2020-04-21-2300/python/DQM/Integration/config/unittestinputsource_cfi.py(89): &lt;module&gt;
[...]

PR #29533, opened 2020-04-22
Fix DQM/Integration unit test in PY3

silviodonato:
This PR fix the DQM/Integration unit tests in PY3 by forcing the conversion of bytes to `utf-8` string.
See #29532

Validated by running the failing unit tests both in PY2 and PY3.


silviodonato:
please close

PR #29545, opened 2020-04-23
Improve Validation for unphysical CaloParticles

rovere:
In the special case of CaloParticles shot parallel to the beamline,
all the checks and selection on the caloParticles collection are turned
off via a configuration parameter, so that the usual plots of
efficiencies, fake, merge and duplicate rate for both layerClusters and
tracksters can be produced.

#### PR validation:

`runTheMatrix.py -l limited`

### DIFF
--- a/Validation/HGCalValidation/plugins/HGCalValidator.cc
+++ b/Validation/HGCalValidation/plugins/HGCalValidator.cc
@@ -11,6 +11,7 @@ HGCalValidator::HGCalValidator(const edm::ParameterSet& pset)
       label_mcl(pset.getParameter<std::vector<edm::InputTag>>("label_mcl")),
       SaveGeneralInfo_(pset.getUntrackedParameter<bool>("SaveGeneralInfo")),
       doCaloParticlePlots_(pset.getUntrackedParameter<bool>("doCaloParticlePlots")),
+      doCaloParticleSelection_(pset.getUntrackedParameter<bool>("doCaloParticleSelection")),
       dolayerclustersPlots_(pset.getUntrackedParameter<bool>("dolayerclustersPlots")),
       domulticlustersPlots_(pset.getUntrackedParameter<bool>("domulticlustersPlots")),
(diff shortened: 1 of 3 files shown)


rovere:
please test

jfernan2:
@mrovere from the PR description and the actual code change I understand no changes are expected, as the tests confirm. If you can confirm it is appreciated. Thanks

rovere:
> @mrovere from the PR description and the actual code change I understand no changes are expected, as the tests confirm. If you can confirm it is appreciated. Thanks

Yes, absolutely no regression is expected. The parameter has been added and set to a value(`True`) that preserves the same logic as before. The value of the parameter has to be changed only for the exceptional case of "particles produced parallel to the beamline", which is never the case in any MC workflow. I decided to add it so that, for specially produced samples, the code can still produce Eff/Fake/Merge/Duplicate rate plots **provided the user manually alter the value from `True` to `False`**.

jfernan2:
+1

Validation/HGCalValidation/plugins/HGCalValidator.cc
```diff
@@ -146,7 +147,7 @@ void HGCalValidator::cpParametersAndSelection(const Histograms& histograms,
   for (auto const caloParticle : cPeff) {
     int id = caloParticle.pdgId();
 
-    if (cpSelector(caloParticle, simVertices)) {
+    if (!doCaloParticleSelection_ || (doCaloParticleSelection_ && cpSelector(caloParticle, simVertices))) {
```
silviodonato, line 150: Isn't better a simpler `     if (!doCaloParticleSelection_ || cpSelector(caloParticle, simVertices)) {` ?

rovere: I guess the compiler will optimize it this way.
Can I propose to couple this with a coming PR from Leonardo that will restructure a great portion of this code/class?

silviodonato: ok, no problem

silviodonato:
+1

PR #29550, opened 2020-04-23
Creating merger call with duplicate removal.

laurenhay:
#### PR description:

Adding a merger class that checks for duplicates. This will be used by JME workflows.

#### PR validation:

These are new files; they effect no existing workflows. These files were tested on the NanoAODJMAR workflows and are working as intended.

N/A

Also adding @rappoccio .

### DIFF
--- a/CommonTools/CandAlgos/plugins/CandFwdPtrMerger.cc
+++ b/CommonTools/CandAlgos/plugins/CandFwdPtrMerger.cc
@@ -1,0 +1,23 @@
+/* \class CandFwdPtrMerger
+ * 
+ * Producer of merged Candidate forward pointer collection 
+ *
+ * \author: Lauren Hay
+ *
+ */
+#include "FWCore/Framework/interface/MakerMacros.h"
+#include "CommonTools/UtilAlgos/interface/Merger.h"
+#include "DataFormats/Candidate/interface/Candidate.h"
+#include "DataFormats/PatCandidates/interface/PackedCandidate.h"
+#include "DataFormats/PatCandidates/interface/PackedGenParticle.h"
+#include "DataFormats/Common/interface/FwdPtr.h"
+
+
+typedef Merger<std::vector<edm::FwdPtr<reco::Candidate>>> CandFwdPtrMerger;
+typedef Merger<std::vector<edm::FwdPtr<pat::PackedCandidate>>> PackedCandidateFwdPtrMerger;
+typedef Merger<std::vector<edm::FwdPtr<pat::PackedGenParticle>>> PackedGenParticleFwdPtrMerger;
+
+
+DEFINE_FWK_MODULE( CandFwdPtrMerger );
+DEFINE_FWK_MODULE( PackedCandidateFwdPtrMerger );
+DEFINE_FWK_MODULE( PackedGenParticleFwdPtrMerger );
--- a/CommonTools/CandAlgos/plugins/CandFwdPtrMerger.cc
+++ b/CommonTools/CandAlgos/plugins/CandFwdPtrMerger.cc
@@ -6,15 +6,15 @@
  *
  */
 #include "FWCore/Framework/interface/MakerMacros.h"
-#include "CommonTools/UtilAlgos/interface/Merger.h"
+#include "CommonTools/UtilAlgos/interface/UniqueMerger.h"
 #include "DataFormats/Candidate/interface/Candidate.h"
 #include "DataFormats/PatCandidates/interface/PackedCandidate.h"
 #include "DataFormats/PatCandidates/interface/PackedGenParticle.h"
 #include "DataFormats/Common/interface/FwdPtr.h"
 
-typedef Merger<std::vector<edm::FwdPtr<reco::Candidate>>> CandFwdPtrMerger;
-typedef Merger<std::vector<edm::FwdPtr<pat::PackedCandidate>>> PackedCandidateFwdPtrMerger;
-typedef Merger<std::vector<edm::FwdPtr<pat::PackedGenParticle>>> PackedGenParticleFwdPtrMerger;
+typedef UniqueMerger<std::vector<edm::FwdPtr<reco::Candidate>>> CandFwdPtrMerger;
+typedef UniqueMerger<std::vector<edm::FwdPtr<pat::PackedCandidate>>> PackedCandidateFwdPtrMerger;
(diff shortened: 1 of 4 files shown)


perrotta:
please test

perrotta:
Please @laurenhay share here some instruction for testing this code
Also some more info in the PR description about possible use cases, or a link to a possible presentation at a JME meeting (e.g.), could be useful.

rappoccio:
Sorry for the delay, were digging out an appropriate global tag. We added a test workflow for CMSSW_11_1_X:

```
cmsrel CMSSW_11_1_X_2020-04-24-1100
cd CMSSW_11_1_X_2020-04-24-1100/src
git cms-init
git cms-merge-topic 29550
git clone git@github.com:cms-jet/NanoAODJMAR.git PhysicsTools/NanoAODJMAR
scram b -j 10
cd PhysicsTools/NanoAODJMAR/test
cmsRun nano111x_on_2020_mc_NANO.py
```

There are expected to be two additional branches to NANOAOD, `GenJetCands` and `JetPFCands`, which contain relevant information for JMAR and BTV.

perrotta:
The SetMerger producer is an almost identical clone of the already existing Merger one. 
Perhaps all that copy/paste could be avoided by simply adding an additional configuration parameter (e.g. ```removeDuplicates```) which is set false by default, and therefore won't need to modify the callers of the current Merger plugin, and configured appropriately for your case. Do you see any counterindication with it?

rappoccio:
We thought of that and tried to implement it at first, but it's rather challenging. For this to be implementable, it must have an `operator==` concept. In particular, things like `reco::Candidate` can't do that. So, we had to add at least one more template decorator, which was a dummy operation by default, and it rapidly got uglier from there. At the end of the day, we thought it better to avoid further template craziness and instead just use a specific class for the `std::set` functionality.

slava77:
it looks like  ```if (checkUnique_ && find(c,coll.begin()...``` is what's needed, where  checkUnique_ is a config parameter.
The SetMerger also has a property of reordering everything: is this an obviously needed property?

rappoccio:
@slava77 Right, `std::find` requires an `operator==`, which is not present in all class instances. We could in principle do an `std::unordered_set` which does the same thing as you're suggesting, but accessing `Ptrs` later will be faster if it is sorted already. Insertion for `std::set` is only `log(n)`.

perrotta:
Is something like this too naive?

```C++
  set_type coll_set;
  std::unique_ptr<OutputCollection> coll(new OutputCollection);
  for (typename vtoken::const_iterator s = srcToken_.begin(); s != srcToken_.end(); ++s) {
    edm::Handle<InputCollection> h;
    evt.getByToken(*s, h);
    for (typename InputCollection::const_iterator c = h->begin(); c != h->end(); ++c) {
      if (checkUnique_) {
        coll_set.emplace(P::clone(*c));
      }
      else {
        coll->push_back(P::clone(*c));
      }
    }
  }
  if (checkUnique_) {
    std::copy(coll_set.begin(), coll_set.end(), coll->begin());
  }
  evt.put(std::move(coll));
```

rappoccio:
@perrotta Yeah, we tried that too :). The `coll_set.emplace(P::clone(*c));` also requires `operator==`.

perrotta:
> @perrotta Yeah, we tried that too :). The `coll_set.emplace(P::clone(*c));` also requires `operator==`.

Still I don't understand why the solution proposed in https://github.com/cms-sw/cmssw/pull/29550#issuecomment-622299991 is different from what originally proposed in this PR: if ```checkUnique_``` the code behaves as in this PR, if ```!checkUnique_``` the code behaves as in the curremt ```Merger``` producer. Or not?

rappoccio:
The code behaves the same way only if there is a defined comparator so that the `std::set` can function. 

https://en.cppreference.com/w/cpp/container/set

```
template<
    class Key,
    class Compare = std::less<Key>,
    class Allocator = std::allocator<Key>
> class set;
```

We don't have a concept of `std::less` for `reco::Candidate` so it doesn't compile.

perrotta:
Ok, thank you @rappoccio : got it!

After discussing this with Slava, while we are convinced that a templated solution could be probably found, we agreed that given the limited extension of the copy/pasted code it could still be acceptable having two different mergers as you are proposing.

However:
- Please add some comment in the code itself that explains where this code is taken from, and which are the differences wrt the copied one
- Is the ordering really needed? Is the ordering provided by std::set for this container of pointers the same as the one that would be used in the sorting [later on](https://github.com/cms-sw/cmssw/pull/29550#issuecomment-621999277)? If so, it is fine keeping it here: but please add another line of comment in the code that explains the advantages of having it at this level.
- About the name of the new class, ```SetMerger``` makes think about a setter method, rather than a "merger of sets".  Perhaps ```MergerWithoutDuplicates``` (or something better that I'm sure you will be able to find) could be a more intuitive name for this class.

perrotta:
@laurenhay @rappoccio : any news?

rappoccio:
@perrotta sorry, last week was finals so both Lauren and I were busy. We'll get to this during the next day or so.

laurenhay:
- Changed named from "SetMerger" to "UniqueMerger" to imply the unique nature of the OutputCollection type.
- Added the following comment: 
` Merges an arbitrary number of collections                                                                                                                                                                
  into a single collection, without duplicates.                                                                                                                                                            
  This class template differs from Merger.h in that it uses a set instead of std::vector.                                                                                                                  
  This requires the OutputCollection type to be sortable, which allows us to search for elements downstream efficiently. `

perrotta:
please test

perrotta:
+1
- A new merged is cloned starting from the original Merger producer: it allows merging without duplicates and provides sorted collections, for future usage in some JME workflows
- Tested working with the script provided
- Jenkins tests pass and show no differences with the baseline, as expected

silviodonato:
merge
@slava77 @perrotta Can't we speedup https://github.com/cms-sw/cmssw/blob/master/CommonTools/UtilAlgos/interface/Merger.h#L61 by adding something like `coll->reserve( sumOfSizeInputCollection )`?
FYI @santocch

slava77:
> merge
> @slava77 @perrotta Can't we speedup https://github.com/cms-sw/cmssw/blob/master/CommonTools/UtilAlgos/interface/Merger.h#L61 by adding something like `coll->reserve( sumOfSizeInputCollection )`?
> FYI @santocch

yes, calling a reserve will speed it up. I don't have a good guess for the total speedup, fractionally it can be about a factor of more than 2, but I think that this is sub-ms per call for even rather large collections.

santocch:
+1

PR #29556, opened 2020-04-24
Magnetic Field optimization

namapane:
#### PR description:

- Minor optimizations discussed in #28180 (some computations moved from MF query calls to constructor)
- move if(debug) couts to MessageLogger
- some minor cleanup of deprecated, obsoleted or unused stuff.

#### PR validation:

Regression-tested for identical values against reference files for all maps/currents/eras.

Optimizations found to sum up to a 0.5% speedup for non-cached MF calls, according to callgrind.


MagneticField/VolumeBasedEngine/src/MagGeometry.cc
```diff
@@ -49,25 +46,49 @@ MagGeometry::MagGeometry(int geomVersion,
...
+      barrelZ1 = 633.890f;
+      barrelZ2 = 662.010f;
+      break;
+    case 2:  // versions 71212, 90322
+      barrelRsq1 = 172.400f * 172.400f;
+      barrelRsq2 = 308.755f * 308.755f;
+      barrelZ0 = 350.000f;
+      barrelZ1 = 633.290f;
+      barrelZ2 = 661.010f;
+      break;
```
slava77, line 90: this is apparently relocating some magic numbers from what was earlier in ```MagGeometry::inBarrel```.
There we had ```// FIXME: Get these dimensions from the builder.```
Is this available from the builder?
Minimally, it's still worth to keep the FIXME comment.

namapane: I put the comment back,
I wrote the FIXME a dozen years ago in the hope I would have figured out a reasonable way to detect the dimensions automatically or to label volumes appropriately in the geometry. I haven't since. A sufficiently generic solution would require adding specifications to the geometry and some logic, which I don't think is worth given that these numbers are not expected to change (last change was 8 years ago and none is in sight in the future)

slava77:

just in case it was missed

slava77:
@cmsbuild please test

slava77:
> Reco comparison results: 6 differences found in the comparisons

in 136.731 
	https://cmssdt.cern.ch/SDT/jenkins-artifacts/pull-request-integration/PR-c491c8/5857/runTheMatrix-results/136.731_RunSinglePh2016B+RunSinglePh2016B+HLTDR2_2016+RECODR2_2016reHLT_skimSinglePh_HIPM+HARVESTDR2/step3_RunSinglePh2016B+RunSinglePh2016B+HLTDR2_2016+RECODR2_2016reHLT_skimSinglePh_HIPM+HARVESTDR2.log
in 136.874
https://cmssdt.cern.ch/SDT/jenkins-artifacts/pull-request-integration/PR-c491c8/5857/runTheMatrix-results/136.874_RunEGamma2018C+RunEGamma2018C+HLTDR2_2018+RECODR2_2018reHLT_skimEGamma_Offline_L1TEgDQM+HARVEST2018_L1TEgDQM/step3_RunEGamma2018C+RunEGamma2018C+HLTDR2_2018+RECODR2_2018reHLT_skimEGamma_Offline_L1TEgDQM+HARVEST2018_L1TEgDQM.log

these are sort of similar 
```
	%MSG-w MagneticField:  CosmicMuonProducer:cosmicMuons 25-Apr-2020 16:37:44 CEST  Run: 274199 Event: 40038054
	MagGeometry::fieldInTesla: failed to find volume for  (306.475,37.2876,658.755) 

```

```
	%MSG-w MagneticField:  MuonIdProducer:muons1stStep 25-Apr-2020 16:53:09 CEST  Run: 319450 Event: 106717766
	MagGeometry::fieldInTesla: failed to find volume for  (59.804,-161.695,625.049) 
```
both are apparently related to changes in this PR

MagneticField/VolumeBasedEngine/interface/MagGeometry.h
```diff
@@ -75,5 +72,12 @@ class MagGeometry {
 
   bool cacheLastVolume;
   int geometryVersion;
+
+  // boundaries of internal barrel-endcap volume separation
+  float barrelRsq1;
+  float barrelRsq2;
+  float barrelZ0;
+  float barrelZ1;
+  float barrelZ2;
```
slava77, line 81: (most) data members have prefix "the" in this class, which is good for distinguishing from local variables.
Please update for consistency.

MagneticField/VolumeBasedEngine/src/MagGeometry.cc
```diff
@@ -199,15 +217,8 @@ MagVolume const* MagGeometry::findVolume(const GlobalPoint& gp, double tolerance
...
-    return (Z < 350. || (R > 172.4 && Z < 633.29) || (R > 308.735 && Z < 662.01));
-  } else if (geometryVersion >= 90812) {  // FIXME no longer supported
-    return (Z < 350. || (R > 172.4 && Z < 633.89) || (R > 308.755 && Z < 662.01));
-  } else {  // versions 71212, 90322
-    return (Z < 350. || (R > 172.4 && Z < 633.29) || (R > 308.755 && Z < 661.01));
-  }
+  float aZ = fabs(gp.z());
+  float aRsq = gp.perp2();
+
+  return ((aZ < barrelZ0) || (aZ < barrelZ1 && aRsq > barrelRsq1) || (aZ < barrelZ2 && aRsq > barrelRsq2));
```
slava77, line 223: the warnings mentioned in https://github.com/cms-sw/cmssw/pull/29556#issuecomment-620715989 are at 
r,z: (308.73498, 658.755) in the 2016 wf and (172.40009, 625.049) in the 2018 wf.
IIUC, we are hitting rounding issues.

namapane: Indeed, the boundary values were doubles before, and I set them to float here. 
If I change boundaries back to doubless the problem goes away. But before claiming this is a fix I think I then need to do some more checks also for the change of making computations in perp2 instead than in perp. 
This is apparently not caught by regression which, I suppose, does not hits these points (standard regression is done comparing the output of 1M random points with a reference file)

namapane: > Reco comparison results: 6 differences found in the comparisons

Slava, apologies for the dumb question, how do I find the other 4?

slava77: this was a lazy copy-paste of the summary of diffs.
The differences are in two workflows as described in https://github.com/cms-sw/cmssw/pull/29556#issuecomment-620715989
there is nothing else, at least in reco steps.

namapane: OK, these two are fixed by simply changing the constants back to doubles, but I clearly need to extend the regression until it catches these (and possible other) numerical side effects before proceeding. WIll do that tomorrow.

MagneticField/VolumeBasedEngine/interface/MagGeometry.h
namapane: done

MagneticField/VolumeBasedEngine/src/MagGeometry.cc
namapane: OK I committed a fix, this is very simple but comes with some updates to test and a re-organization of MessageLogger categories that was necessary to be able to check whether these optimizations could also induce numerical problems elsewhere.
PS: I am porting this fix and all updates to #29561 as well.

perrotta:
@cmsbuild please test

slava77:
> You potentially added 184151 lines to the logs

this should be cleaned up

e.g.  https://cmssdt.cern.ch/SDT/jenkins-artifacts/pull-request-integration/PR-c491c8/5926/runTheMatrix-results/136.731_RunSinglePh2016B+RunSinglePh2016B+HLTDR2_2016+RECODR2_2016reHLT_skimSinglePh_HIPM+HARVESTDR2/step3_RunSinglePh2016B+RunSinglePh2016B+HLTDR2_2016+RECODR2_2016reHLT_skimSinglePh_HIPM+HARVESTDR2.log

```
%MSG-w MagneticField:   SeedCreatorFromRegionConsecutiveHitsEDProducer:initialStepSeedsPreSplitting  30-Apr-2020 12:51:45 CEST Run: 274199 Event: 39389360
 Point  (343.499,-53.995,-5.20496)  is outside the validity region of ParabolicParametrizedMagneticField
```

slava77:
or more obviously in a point where it's supposed to work
```
%MSG-w MagneticField:   SeedCreatorFromRegionConsecutiveHitsEDProducer:initialStepSeeds  30-Apr-2020 12:52:06 CEST Run: 274199 Event: 39797590
 Point  (85.2733,-89.756,5.55544)  is outside the validity region of ParabolicParametrizedMagneticField
```

slava77:
> or more obviously in a point where it's supposed to work
> 
> ```
> %MSG-w MagneticField:   SeedCreatorFromRegionConsecutiveHitsEDProducer:initialStepSeeds  30-Apr-2020 12:52:06 CEST Run: 274199 Event: 39797590
>  Point  (85.2733,-89.756,5.55544)  is outside the validity region of ParabolicParametrizedMagneticField
> ```

ah, actually I'm wrong, this is r=123, while the field is defined in r < 115.
So, the text is correct, but it was a LogDebug before and now was made a LogWarning

namapane:
I can put it back to LogDebug. However, callers should rcheck why they are calling a fast parametrization outside its region of validity, since they will be getting (oviously) an incorrect result.
It may be that trajectory states are extrapolated just a little too far before starting the backward fit, and in this case there should be no problem. 
But  (343.499,-53.995,-5.20496) seems to me REALLY too far away. R=343 is inside the coil. That parametrization will return complete nonsense at that location.Why a seed creator would ever have to propagate to the coil? It looks like a bug. 
So after all, it should have probably been a LogWarning from the beginning.

namapane:
I have the impression that this fast parametrization (which was provided upon request of tracking people for fast field query where modest accuracy is acceptable) is accidentally misused, e.g. by percolating it to places where the full map should be used instead. 
I think that this "incident" could be taken as a chance to check if anything like this, or other algorithmic misuse of the fast parametrization, is actually happening. 
But as I said, I have no problem in turning a blind eye and silencing the warning; just let me know.

slava77:
> I think that this "incident" could be taken as a chance to check if anything like this, or other algorithmic misuse of the fast parametrization, is actually happening.
> But as I said, I have no problem in turning a blind eye and silencing the warning; just let me know.

in this case "initialStepSeeds" has no good reason to be anywhere outside even the pixel tracker. So, the large values are most likely just random bad cases of some position computation. 
I think that it's fine to stay with LogDebug.
```::isDefined``` is available in the public interface and clients can use it if they care.

slava77:
@cmsbuild please test

slava77:
+1

for https://github.com/cms-sw/cmssw/pull/29556	8850003 

silviodonato:
+1

PR #29557, opened 2020-04-24
Tau decay hook

smrenna:
[WIP] A new UserHook in Pythia8 that forces the decay of taus at the process level.  At this point, one should be able to select only tau decays to leptons, but that has not been implemented yet.  It did not work properly with ResonanceDecayFilter, even though the taus are clearly decayed.


slava77:
-1

wrong branch or repo

smrenna:
All I want to do is make a minor change to GeneratorInterface/Pythia8Interface.
How do I do that when my fork of cmssw is way out of synch with the current developments in other parts that aren’t important to this change?

Steve

kpedro88:
@smrenna it's best to check out a new CMSSW release (like CMSSW_11_1_0_pre6), then add the package you want to modify. It will pull from the official repository, thereby getting the latest developments.
```
cmsrel CMSSW_11_1_0_pre6
cd CMSSW_11_1_0_pre6/src
cmsenv
git cms-addpkg GeneratorInterface/Pythia8Interface
git checkout -b tauDecayHook2
[make changes, commit changes]
git push my-cmssw tauDecayHook2
```

smrenna:
That was what I originally tried, but it did not work at all.
It complained about permission.

kpedro88:
Perhaps the ssh key for your GitHub account was missing?

smrenna:
Hi, Kevin,

Setting the key worked, but what happens when you get put onto a different cms lpc node?
Is the key associated with the cmslpc account, or the particular machine on which you do the commit?

Steve

kpedro88:
The filesystem on cmslpc is shared among all nodes, so you should be able to use your key on any of them.

smrenna:
Second attempt in tauDecayHook2.

PR #29561, opened 2020-04-25
Tentative implementation of TLS cache for MF

namapane:
#### PR description:

Tentative impementation of per-thread MF volume cache as suggested by @VinInn in the discussion of #28180 (https://github.com/cms-sw/cmssw/pull/28180#issuecomment-618174565), intended to allow further testing.

#### PR validation:

Regression-tested for identical values against reference files for all maps/currents/eras.

Also verified this in a multithread job spanning across multiple IOVs to verify that the cache gets invalidated when necessary as expected.

### DIFF
--- a/MagneticField/VolumeBasedEngine/src/MagGeometry.cc
+++ b/MagneticField/VolumeBasedEngine/src/MagGeometry.cc
@@ -21,6 +21,9 @@ using namespace std;
 using namespace edm;
 
 namespace {
+  // A thread-local cache is accepted because MagneticField is used by several other ESProducts
+  // via member variable, MagneticField and the other ESProducts are widely used, and migrating
+  // all the uses of all those was deemed to have very high cost.
   std::atomic<int> instanceCounter(0);
   thread_local int localInstance = 0;
   thread_local MagVolume const* lastVolume = nullptr;
--- a/MagneticField/VolumeBasedEngine/src/MagGeometry.cc
+++ b/MagneticField/VolumeBasedEngine/src/MagGeometry.cc
@@ -97,7 +97,8 @@ MagGeometry::MagGeometry(int geomVersion,
       break;
   }
 
-  LogTrace("MagGeometry_cache") << "*** In MagGeometry ctor: me_=" << me_ << " instanceCounter=" << instanceCounter << endl;
+  LogTrace("MagGeometry_cache") << "*** In MagGeometry ctor: me_=" << me_ << " instanceCounter=" << instanceCounter
+                                << endl;
 }
 
 MagGeometry::~MagGeometry() {
@@ -181,7 +182,7 @@ MagVolume const* MagGeometry::findVolume(const GlobalPoint& gp, double tolerance
   // Clear volume cache if this is a new instance
   if (me_ != localInstance) {
     LogTrace("MagGeometry_cache") << "*** In MagGeometry::findVolume resetting cache: me=" << me_
-                            << " localInstance=" << localInstance << endl;
+                                  << " localInstance=" << localInstance << endl;
     localInstance = me_;
     lastVolume = nullptr;
   }
--- a/MagneticField/VolumeBasedEngine/src/MagGeometry.cc
+++ b/MagneticField/VolumeBasedEngine/src/MagGeometry.cc
@@ -20,6 +20,12 @@
 using namespace std;
 using namespace edm;
 
+namespace {
+  std::atomic<int> instanceCounter(0);
+  thread_local int localInstance = 0;
+  thread_local MagVolume const* lastVolume = nullptr;
+}  // namespace
+
 MagGeometry::MagGeometry(int geomVersion,
                          const std::vector<MagBLayer*>& tbl,
                          const std::vector<MagESector*>& tes,
@@ -36,7 +42,7 @@ MagGeometry::MagGeometry(int geomVersion,
                          const std::vector<MagESector const*>& tes,
                          const std::vector<MagVolume6Faces const*>& tbv,
                          const std::vector<MagVolume6Faces const*>& tev)
-    : lastVolume(nullptr),
+    : me_(++instanceCounter),
       theBLayers(tbl),
       theESectors(tes),
       theBVolumes(tbv),
@@ -90,6 +96,8 @@ MagGeometry::MagGeometry(int geomVersion,
       theBarrelZ2 = 661.010;
       break;
   }
+
+  LogTrace("MagGeometry_cache") << "*** In MagGeometry ctor: me_=" << me_ << " instanceCounter=" << instanceCounter << endl;
 }
 
 MagGeometry::~MagGeometry() {
@@ -170,10 +178,16 @@ MagVolume const* MagGeometry::findVolume1(const GlobalPoint& gp, double toleranc
 
 // Use hierarchical structure for fast lookup.
 MagVolume const* MagGeometry::findVolume(const GlobalPoint& gp, double tolerance) const {
-  // Check volume cache
-  auto lastVolumeCheck = lastVolume.load(std::memory_order_acquire);
-  if (lastVolumeCheck != nullptr && lastVolumeCheck->inside(gp)) {
-    return lastVolumeCheck;
+  // Clear volume cache if this is a new instance
+  if (me_ != localInstance) {
+    LogTrace("MagGeometry_cache") << "*** In MagGeometry::findVolume resetting cache: me=" << me_
+                            << " localInstance=" << localInstance << endl;
+    localInstance = me_;
+    lastVolume = nullptr;
+  }
+
+  if (lastVolume != nullptr && lastVolume->inside(gp)) {
+    return lastVolume;
   }
 
   MagVolume const* result = nullptr;
@@ -211,7 +225,7 @@ MagVolume const* MagGeometry::findVolume(const GlobalPoint& gp, double tolerance
   }
 
   if (cacheLastVolume)
-    lastVolume.store(result, std::memory_order_release);
+    lastVolume = result;
 
   return result;
 }
--- a/MagneticField/Engine/test/regression.py
+++ b/MagneticField/Engine/test/regression.py
@@ -1,6 +1,7 @@
 
 import FWCore.ParameterSet.Config as cms
 import FWCore.ParameterSet.VarParsing as VarParsing
+import sys
 
 options = VarParsing.VarParsing()
 
@@ -20,7 +21,7 @@ options.register('current',
                  18000, #default value
                  VarParsing.VarParsing.multiplicity.singleton,
                  VarParsing.VarParsing.varType.float,
-                 "Magnet current (nominal values: 18164=3.8T; 16730=3.5T; 14340=3T; 9500=2T")
+                 "Magnet current (nominal values: 18164=3.8T; 16730=3.5T; 14340=3T; 9500=2T; -1=loop in different IOV")
 
 options.parseArguments()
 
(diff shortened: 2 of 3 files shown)


MagneticField/VolumeBasedEngine/src/MagGeometry.cc
```diff
@@ -15,14 +15,17 @@
...
-
 #include <iostream>
 
 using namespace std;
 using namespace edm;
 
+namespace {
+  std::atomic<int> instanceCounter(0);
+  thread_local int localInstance = 0;
+  thread_local MagVolume const* lastVolume = nullptr;
```
slava77, line 26: considering thread_local is restrictive in our environment and should be avoided,
how would this differ from using e.g. ```std::map<std::thread::id, MagVolume const*> lastVolume;``` and then accessing it via ```lastVolume[std::this_thread::get_id()]```? 
the cost of navigating the map is likely still much smaller than calling the volume inTesla.
A slightly faster version could use a predetermined array size with the id conversion to int of the array size.

@makortel @VinInn @Dr15Jones

makortel: Every access to an `std::map` shared across threads would need to be protected e.g. with a `mutex`. I would rather not add synchronization points.

slava77: unordered_map with a large enough reserve wouldn't need a sync, or would it?

slava77: I'd be happy to stay with thread_local, however if this is acceptable, it should be acceptable in all packages where it's reasonably needed.
Can we go for this precedent?

namapane: In the meanwhile I finally managed test the cache re-set on several threads running in parallel (besides specifiyng process.options.numberOfThreads and numberOfStreams I also had to turn my analyzer into a stream analyzer; thanks to @VinInn for the tip). 

So I verified that the cache is re-set correctly in each of the active threads at each IOV transition.

@slava77 : I'm not the one who can answer your last question, but my 2 cents: I do understand your concern about setting a precedent. However, MF is a fundamental service for pretty much anything in SIM, RECO and HLT, not some obscure analysis package. 
If it sets a precedent, I think it also sets a pretty high standard on what is to be considered reasonable...

Regarding the (unordered) map: nice idea for what I am concerned, but I don't have enough experience to tell whether it's better than the TLS one. If experts agree it is, I can try implementing it as well.

civanch:
@namapane , effect on CPU for SIM step is less than expected. To be sure, I have done the test for MinBias and ttbar events:
image

civanch:
There is a spead-up for 5% for both type of events for 16 threads. For MinBias events scaling is not good, for ttbar is better. This plot is different from one of @amadio shown in #28180. The difference may be connected with different Geant4 or different way of caching.

slava77:
assign core

slava77:
@cmsbuild please test

slava77:
> There is a spead-up for 5% for both type of events for 16 threads. For MinBias events scaling is not good, for ttbar is better. This plot is different from one of @amadio shown in #28180. The difference may be connected with different Geant4 or different way of caching.

which machine did you use?
was the test based on the same command as in https://github.com/cms-sw/cmssw/pull/28180#issue-328345939 ?
```
cmsDriver.py MinBias_13TeV_cfi --era Run2_2018 -s GEN,SIM --pileup=NoPileUp
--conditions auto:phase1_2018_realistic --geometry DB:Extended
--eventcontent=RAWSIM --datatier GEN-SIM --dirout=./ -n $nevents --nThreads $nthreads
--fileout file:scaling.root --mc --no_exec cmsRun MinBias_13TeV_cfi_GEN_SIM.py
```

namapane:
> This plot is different from one of @amadio shown in #28180.

Actually the plot in #28280 was showing a significant improvement only starting from 24 threads, it was rather small which 16 threads.

MagneticField/VolumeBasedEngine/src/MagGeometry.cc
makortel: > unordered_map with a large enough reserve wouldn't need a sync, or would it?

It would, standard does not guarantee anything between synchronization between threads without the user doing it.

> the cost of navigating the map is likely still much smaller than calling the volume inTesla.

Maybe, but it would slow down the fast path, i.e. workloads with large cache hit ratios would have an impact.

> If it sets a precedent, I think it also sets a pretty high standard on what is to be considered reasonable...

I agree, if we are setting a precedent here, we should be really careful.

slava77: > > the cost of navigating the map is likely still much smaller than calling the volume inTesla.
> 
> Maybe, but it would slow down the fast path, i.e. workloads with large cache hit ratios would have an impact.

Just to clarify my earlier comment, I understand that selecting a correct cache pointer will be more costly. My point is that even the fastest path that has the cache immediately, still has to call the interpolation math ("the volume inTesla"), which is likely much more expensive than any reasonable cache retrieval from non-thread_local mapping/array using thread id.

slava77:
the minbias throughput scaling drop from 12 to 16 looks in some disagreement with the plot in #28280  
compared to the still nice linearity in the ttbar, it was either running on a different machine with 12 physical cores or there was something else running on the node.

namapane:
> the minbias throughput scaling drop from 12 to 16 looks in some disagreement with the plot in #28280
> compared to the still nice linearity in the ttbar, it was either running on a different machine with 12 physical cores or there was something else running on the node.

I agree, but still, at 12 threads the improvement was +3.65% according to the table; what I mean is that it does not seem inconsistent with a 5% improvements observed here by @civanch with 16 threads.
To see a significant improvement we should probably check at least 24 and 32 threads here as well, if I understand @amadio's numbers correctly.

MagneticField/VolumeBasedEngine/src/MagGeometry.cc
namapane: The interpolation is quite efficient; it is reduced to one coordinate transformation + a linear interpolation within a regular grid + a RF transformation of the result. I can try timing the overhead of your suggestion in a cache-hit scenario - but I'd save the effort if the sync issue is considered to be a showstopper...

makortel: > but I'd save the effort if the sync issue is considered to be a showstopper...

In general we should avoid adding synchronization points that the framework is unaware of (they will decrease CPU efficiency).

slava77: can the framework provide us with a thread id -> int index map object (best if the range of int is small and can be used as an array index) that would not require synchronization on the client side?
This way the framework will be aware of how it's filled and we'll just have an array of cache pointers (possibly atomic) to read from in the MF code instead of a thread_local.

makortel: I replied on the general "per-thread" question in https://github.com/cms-sw/cmssw/pull/28180#issuecomment-620174366.

An array with per-thread indexing would technically work, but for a good performance one has to give each thread a separate cache line to prevent false sharing (so a simple 'std::array<MagVolume const*, N>`  would not be sufficient). How would you decide the array size? Strictly speaking we do not guarantee to keep the same threads though a job. And allowing resize leads to the need to lock/synchronize (there may be ways to do an update without locking with a price of complicated code).

slava77: for the array I was thinking (if a fixed thread indexing is not practical) to  keep it ```std::array<std::atomic<MagVolume const*>, N>``` and do ```(int)thread_index%N```
where N can be a reasonable max (like 256)

civanch:
I use the same PC without hyperthreading as @amadio. PC from openlab: olhswep04, it has 32 cores. I will extend the plot.

civanch:
New cpu plot: 
image

MagneticField/VolumeBasedEngine/src/MagGeometry.cc
VinInn: WAIT. the framework should guarantee NOT to exceed the number of "processing" threads in the configuration. I do not think that the framework should be free to add a couple of threads to overcome a temporary shortage. If the framework wants to have such a freedom it should reserve  those thread in advance.

I proposed  somewhere above a possible implementation of all this.
Essentially the framework should provide a global function that return a struct with all info (process-status?)  that may be of interest during processing (event,setup, module etc as pointers or just identifiers) including an "index" that is per thread and per stream and it is sequential and has a known maximum (returned by another global function or stored in the struct above). Internally will use a thread_local and I suggest to locate  thise functions in the "main" program (statically linked)

civanch:
I suspect, that in Geant4 10.6 number of calls to MF is less than in 10.4p03 by 10-20%.

namapane:
Thanks @civanch . It is clearly puzzling, but I find it hard to draw conclusions: we would expect to recover the improvement shown by @amadio , but numbers here are different (much lower throughput) even before the fix, and also the scaling behavior is different even in the no-fix case.
If you used the same machine that could maybe be because you are running on a different kind of events? In any case, if we can't even reproduce the before-the-fix scaling shown by @amadio it is hard to tell if the apparently lower gain here is due to the different implementation of caching, or anything else. 
IMO @amadio 's results are a good reference because he did a detailed profiling. A possibility would be to re-test the scaling of his implementation (#28180), so that we have a proper comparison.

namapane:
> I suspect, that in Geant4 10.6 number of calls to MF is less than in 10.4p03 by 10-20%.

That could indeed dilute the improvement, but I find it puzzling that the before-the-fix throughput has become much lower (was eg 4.471 evt/sec @32 threads, it is ~3.1 in your plot), if you use the same machine and run the same kind of events.

amadio:
Is the time being measured the same? I measured only the time from the beginning of the event loop to the end of the simulation to calculate events/s. If initialization time is not excluded, that could explain the lower throughput here.

Dr15Jones:
@VinInn wrote
> WAIT. the framework should guarantee NOT to exceed the number of "processing" threads in the configuration. I do not think that the framework should be free to add a couple of threads to overcome a temporary shortage. If the framework wants to have such a freedom it should reserve those thread in advance.

The number of threads used to run TBB tasks are set from the configuration. TBB does reserve the right to be able to shutdown and startup different threads while staying within that user requested bounds. (In fact there are cases where if you specify 1 thread to TBB it will start a second one.) Then there are all the other threads used by 3rd party tools, xrootd, TensorFlow's use of Eigen, CUDA, etc. Each of those would be assigned an thread id by C++ so just calling std::thread_id would not give a safe index into a limited sized container.

VinInn:
@Dr15Jones , sure. This is why I think that thread_id is not a good/useful index.
On the other hand I expect the Framework to have control of "its' own threads and streams and I am sure that, before calling any module methods, the framework can
1) set a pointer in its own unique thread_local (statically bound to "main")
2) provide a unique index that is guaranteed to be between 0 and "The number of threads used to run TBB tasks".  (I can design many simple/slow mechanisms to do that even locally)
Normal users should not care of any other threads: I do not expect the mag field to be called from xrootd or tensorflow.

Dr15Jones:
> for the array I was thinking (if a fixed thread indexing is not practical) to keep it std::array<std::atomic<MagVolume const*>, N> and do (int)thread_index%N
where N can be a reasonable max (like 256)

The fixed max is not a safe value. I've run jobs on KNL using 256 TBB threads (which meant there were more than 256 threads running in the job). I also ran test jobs using the prototype framework using 3000 threads (to prove scaling). So we do not know what is actually a reasonable upper bound on number of threads that will hold for the next 10+ years.

If the modulo ever causes two threads to share the same memory location, even the use of an std::atomic is not enough to make it safe. The problem is there are two values which must be tracked
1. the memory location being cached
1. the id of the underlying object for which you are doing the caching

The suggested design does
1. test to see if the last id used by the entry in the array matches the id of the object requesting the cache
1. if it does use the cached value
1. if it doesn't find the new cached value and updae the id

The problem is if two threads share the same memory location, the first thread could check the id see it was what was cached and then the thread could be paused. The second thread comes in sees the id does not match, update both the cache and the id. Then the first thread unpauses and uses the wrong cache value.

slava77:
@civanch 
what is the baseline/reference in your tests?
Is it #29556 or the IB?
I think that we need to decouple the technical optimization of #29556 from the thread_local cache handling.

@namapane 
do you have some more MF-dominated multi-threaded tests to quantify the performance changes?

civanch:
@slava77, I use one of the recent IB and standard WF GEN-SIM. Number of events 400 (may be lower than necessary for 32 threads). CPU is measured via TimeMemoryInfo.py Throughput is plotted.

namapane:
> I think that we need to decouple the technical optimization of #29556 from the thread_local cache handling.

Well keep in mind that this optimization accounts for a 0.5% speedup for MF calls (ie over the cost of  ::InTesla according to callgrind). This would be diluted in any real-time application. So I doubt this can make a visible difference.
 
> @namapane
> do you have some more MF-dominated multi-threaded tests to quantify the performance changes?

I can prepare a set of tests , but I have no access to a 32-core machine suitable for profiling. I checked olhswep04, but I cannot login there.

Dr15Jones:
@civanch 
> @slava77, I use one of the recent IB and standard WF GEN-SIM. Number of events 400 (may be lower than necessary for 32 threads). CPU is measured via TimeMemoryInfo.py Throughput is plotted.

When I do such performance measurements I always scale the number of events with the number of threads. I also reuse the same events in order to guarantee that all jobs are doing the same work. E.g. I might have 100 events which I replicate N times in the same input file to process. Then if a job is going to use m threads I have it process m*100 events. I have found that seems to give me the most consistent results.

civanch:
@Dr15Jones , all, I agree, that low statistics bias results. After some checks it become clear that we have only two differences with @amadio : Geant4 version and statistics. He used statistics 1024 events and my plot made from 400 events. This is a problem for 32 threads and MinBias events, not a problem for ttbar. I have redone the plot using 1000 events and results are different:

image

This PR improve CPU for few % even for 8 thread. It is a clear effect beyond of systematic uncertainty of the method.

amadio:
The other thing I have done is to pin the threads up to 16 to the first socket, as CMSSW doesn't do that automatically. This alone increases performance compared with not pinning. Is TiMemory showing throughput based only on the event loop?

namapane:
In any case it the improvement is now quite evident. Why don't we proceed with integrating this implementation, so that we profit from the improvement while alternatives are discussed and possibly implemented.
Given that #29556 has been merged, should I close this PR (which I started from the same improvements of #29556) and make a new one with only the new cache mechanism?

civanch:
@namapane , I agree.

slava77:
>  Given that #29556 has been merged, should I close this PR (which I started from the same improvements of #29556) and make a new one with only the new cache mechanism?

you can just rebase; this way the conversation with measurements will be a bit closer to the implementation.

namapane:
> you can just rebase; this way the conversation with measurements will be a bit closer to the implementation.

Uhm, I tried this way:
- created a new area based on CMSSW_11_1_X_2020-05-01-2300, (which includes #29556)
-` git cms-addpkg MagneticField; git fetch my-cmssw MF_TLScache:MF_TLScache; git branch MF_TLScache`
-git rebase CMSSW_11_1_X

At this point my local  branch contains only the relevant commits, as expected. So it seems fine.
But when I try to push it I get the message:
```
> git push my-cmssw MF_TLScache
To github.com:namapane/cmssw.git
 ! [rejected]                MF_TLScache -> MF_TLScache (non-fast-forward)
error: failed to push some refs to 'git@github.com:namapane/cmssw.git'
hint: Updates were rejected because the tip of your current branch is behind
hint: its remote counterpart. Integrate the remote changes (e.g.
hint: 'git pull ...') before pushing again.
hint: See the 'Note about fast-forwards' in 'git push --help' for details.
```

Can you please advise on the correct way to perform such a rebase?

civanch:
@namapane , my personal expérience: always faster to create a new empty working area, add needed packages, and copy needed files by hand from old area. 

Concerning your case  likely "-git rebase CMSSW_11_1_X" mess up things, because you need rebase to the master.

slava77:
> At this point my local branch contains only the relevant commits, as expected. So it seems fine.
> But when I try to push it I get the message:

post-rebase updates should be force-pushed (push --force).

namapane:
> post-rebase updates should be force-pushed (push --force).

Thanks, indeed using the Force it went smoothly :-)

slava77:
@cmsbuild please test

silviodonato:
After this long discussion, do you want to include this change in 11_1_0_pre7? Do you need more time?
core:  @Dr15Jones @makortel @smuzaffar
reconstruction:  @perrotta @slava77

slava77:
this is a rather technical update, I do not see a need to be tied to pre7 schedule.

civanch:
I agree with Slava, there is no urgency, we only need to converge.

slava77:
+1

for https://github.com/cms-sw/cmssw/pull/29561	744dc50
    - accepted now based on https://github.com/cms-sw/cmssw/pull/29631#issuecomment-625425556
- jenkins tests pass and comparisons with the baseline show no relevant differences

MagneticField/VolumeBasedEngine/src/MagGeometry.cc
```diff
@@ -20,6 +20,12 @@
 using namespace std;
 using namespace edm;
 
+namespace {
+  std::atomic<int> instanceCounter(0);
```
makortel, line 24: @namapane Could you add a comment along
```suggestion
  // A thread-local cache is accepted because MagneticField is used by several other ESProducts
  // via member variable, MagneticField and the other ESProducts are widely used, and migrating
  // all the uses of all those was deemed to have very high cost.
  std::atomic<int> instanceCounter(0);
```
to document to rationale?

makortel:
@cmsbuild, please test

Thanks!

makortel:
+1

slava77:
+1

for https://github.com/cms-sw/cmssw/pull/29561	df4ab9b 
- added comments in the code since the last reco signoff

silviodonato:
+1

PR #29566, opened 2020-04-27
Phase2-hgx238 Add some missing methods to a few geometry classes used for HGCal

bsunanda:
#### PR description:

Add some missing methods to a few geometry classes used for HGCal

#### PR validation:

Tested with runTheMatrix.py

Nothing special

### DIFF
--- a/Geometry/CaloGeometry/src/FlatHexagon.cc
+++ b/Geometry/CaloGeometry/src/FlatHexagon.cc
@@ -18,6 +18,8 @@ typedef HepGeom::Vector3D<double> DVec3D;
 typedef HepGeom::Plane3D<double> DPlane3D;
 typedef HepGeom::Point3D<double> DPt3D;
 
+static const float tolmin = 1.e-12;
+
 //----------------------------------------------------------------------
 
 FlatHexagon::FlatHexagon()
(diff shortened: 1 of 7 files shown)


Geometry/CaloGeometry/src/FlatHexagon.cc
```diff
@@ -96,6 +96,18 @@ GlobalPoint FlatHexagon::getPosition(const Pt3D& local) const {
   return GlobalPoint(glb.x(), glb.y(), glb.z());
 }
 
+float FlatHexagon::etaSpan() const {
+  float eta1 = -std::log(std::tan(0.5 * std::atan((m_global.perp() + param()[k_R]) / std::abs(m_global.z()))));
```
kpedro88, line 100: is it possible to cache these values (and the ones in phiSpan below) so the calculation doesn't have to be repeated every time the accessor is called?

Dr15Jones: Thread safe (and efficient) caching is a tricky business.

kpedro88: This is a pretty simple class, though. the dEta and dPhi values can just be calculated and stored whenever m_global is changed (only in constructors or setPosition() accessor)

bsunanda: It is used in HGCalGeometry and m_global is changed when the CaloCellGeometry pointer is asked. In most of the geometry applications, this function is not required. It should be heavy to calculate this every time it is loaded. Most likely this is required in debugging or in some analysis code. It is better to calculate them only when it is needed.

bsunanda:
@cmsbuild Please test

kpedro88:
+upgrade

Geometry/CaloGeometry/src/FlatTrd.cc
```diff
@@ -96,6 +96,18 @@ GlobalPoint FlatTrd::getPosition(const Pt3D& local) const {
...
 
+float FlatTrd::etaSpan() const {
+  float eta1 = -std::log(std::tan(0.5 * std::atan((m_global.perp() + param()[k_dY1]) / std::abs(m_global.z()))));
+  float eta2 = -std::log(std::tan(0.5 * std::atan((m_global.perp() - param()[k_dY1]) / std::abs(m_global.z()))));
+  float dEta = std::abs(eta1 - eta2);
+  return dEta;
+}
+
+float FlatTrd::phiSpan() const {
+  float dPhi = 2.0 * std::atan(0.5 * (param()[k_dX1] + param()[k_dX2]) / m_global.perp());
```
cvuosalo, line 107: For all four versions of `etaSpan` and `phiSpan`, please prevent segmentation faults and nonsense values. Please check:
```
assert(param() != nullptr);
```
and check that `m_global.z()` and `m_global.perp()` are not zero before dividing by them.

bsunanda: Done

bsunanda:
@cmsbuild Please test

cvuosalo:
+1

kpedro88:
+upgrade

Geometry/CaloGeometry/src/FlatHexagon.cc
```diff
@@ -18,6 +18,8 @@ typedef HepGeom::Vector3D<double> DVec3D;
 typedef HepGeom::Plane3D<double> DPlane3D;
 typedef HepGeom::Point3D<double> DPt3D;
 
+static const float tolmin = 1.e-12;
```
silviodonato, line 21: It is suggested to use `constexpr` for all constant values that can be evaluated at compile time. https://cms-sw.github.io/cms_coding_rules.html

silviodonato:
+1
@bsunanda I added a minor comment, you can implement it later

PR #29581, opened 2020-04-29
[LLVM10] PhysicsTools fix clang warnings

mrodozov:
#### PR description:

Fix warnings in PhysicsTools, 
please see the  PhysicsTools/PatUtils one the others are trivial
the warning about it is here:
https://cmssdt.cern.ch/SDT/cgi-bin/buildlogs/slc7_amd64_gcc820/CMSSW_11_1_CLANG_X_2020-04-27-2300/PhysicsTools/PatUtils

#### PR validation:

Builds without warnings with the CLANG IB

### DIFF
--- a/PhysicsTools/PatUtils/plugins/ShiftedPFCandidateProducerForPFNoPUMEt.cc
+++ b/PhysicsTools/PatUtils/plugins/ShiftedPFCandidateProducerForPFNoPUMEt.cc
@@ -74,11 +74,13 @@ void ShiftedPFCandidateProducerForPFNoPUMEt::produce(edm::Event& evt, const edm:
     const reco::PFJet* jet_matched = nullptr;
     for (auto jet : selectedJets) {
       for (auto jetc : jet->getPFConstituents()) {
-        if (jet_matched)
-          break;
-        if (deltaR2(originalPFCandidate->p4(), jetc->p4()) < dR2Match)
+        if (deltaR2(originalPFCandidate->p4(), jetc->p4()) < dR2Match) {
           jet_matched = jet;
+          break;
+        }
       }
+      if (jet_matched)
+        break;
     }
 
     double shift = 0.;
--- a/PhysicsTools/PatUtils/plugins/ShiftedPFCandidateProducerForPFNoPUMEt.cc
+++ b/PhysicsTools/PatUtils/plugins/ShiftedPFCandidateProducerForPFNoPUMEt.cc
@@ -72,12 +72,12 @@ void ShiftedPFCandidateProducerForPFNoPUMEt::produce(edm::Event& evt, const edm:
        originalPFCandidate != originalPFCandidates->end();
        ++originalPFCandidate) {
     const reco::PFJet* jet_matched = nullptr;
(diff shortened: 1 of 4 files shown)


mrodozov:
please test for CMSSW_11_1_CLANG_X_2020-04-27-2300

mrodozov:
please test for CMSSW_11_1_CLANG_X

slava77:
@cmsbuild please test

for clean comparisons

PhysicsTools/PatUtils/plugins/ShiftedPFCandidateProducerForPFNoPUMEt.cc
```diff
@@ -72,12 +72,12 @@ void ShiftedPFCandidateProducerForPFNoPUMEt::produce(edm::Event& evt, const edm:
...
     const reco::PFJet* jet_matched = nullptr;
-    for (std::vector<const reco::PFJet*>::iterator jet = selectedJets.begin(); jet != selectedJets.end(); ++jet) {
-      for (std::vector<reco::PFCandidatePtr>::const_iterator jetConstituent = (*jet)->getPFConstituents().begin();
-           jetConstituent != (*jet)->getPFConstituents().end() && jet_matched == nullptr;
-           ++jetConstituent) {
-        if (deltaR2(originalPFCandidate->p4(), (*jetConstituent)->p4()) < dR2Match)
-          jet_matched = (*jet);
+    for (auto jet : selectedJets) {
+      for (auto jetc : jet->getPFConstituents()) {
+        if (jet_matched)
```
perrotta, line 77: It could be even simpler if you ```break``` inside an if block started with L79 below

On the other hand, there could be some difference between the fixed code and the original one if there are two PFCandidates that are within ```dr2Match```. I understand that this is rather unlikely, and if there are two such candidates nobody can say that taking the first one (as with your fix) is worst than taking the last one (as in the original code). Therefore I'd be tempted to stick on your proposed faster solution.

smuzaffar: @perrotta , looks like the original code is also taking the first match (due to ` && jet_matched == nullptr` condition in the inner loop) .  @mrodozov I think something like the following will simplify and improve the code

```
    for (auto jet : selectedJets) {
      for (auto jetc : jet->getPFConstituents()) {
        if (deltaR2(originalPFCandidate->p4(), jetc->p4()) < dR2Match) {
          jet_matched = jet;
          break;
        }
      }
      if (jet_matched)
        break;
    }
```

there is no need to loop over all jets if match is already found ... what do you think @perrotta ?

perrotta: Correct.
Yes, that's even better: please @mrodozov  go ahead with it

mrodozov:
please test

perrotta:
+1
- Technical
- Jenkins tests pass

santocch:
+1

silviodonato:
merge
@peruzzim please let me know if you have any objections

PR #29590, opened 2020-04-29
Run3-gem36 Adding an object MuonGeometryConstants to EventSetup which can give similar functionality as MuonDDDConstants

bsunanda:
#### PR description:

Creates an object (MuonGeometryConstants) from DDCompactView of DD4Hep cms::DDCompactView and attaches to EventSetup. This can then be used in all applications (Simulation/Alignment/ ..) without invoking DDDCompactView

#### PR validation:

Tested with built in cfg files (the implementation is complete for DDD)

Nothing special


bsunanda:
@cmsbuild Please test

cvuosalo:
@bsunanda Don't we use "DDD" for the old DD? Isn't it confusing for `MuonDDDParameters` and `MuonDDDConstants` to be used for both the old DD and DD4hep?
Maybe they should be named generically, like `MuonParameters` and `MuonConstants`.

bsunanda:
@cvuosalo MuonDDDConstants is used in many parts of the code. That is why I left it like that - later on Sergio has to rename MuonDDDParameters as MuonDDDConstants and alter a vey few files - can we discuss about this among ourselves. I have no problems - I would like to name it as MuonGeometryParameters if Sergio agrees to edit the classes in all parts of the code

cvuosalo:
@bsunanda The new names created by this PR should not contain "DDD" if they are to be used also with DD4hep. We don't have to rename `MuonDDDConstants`. We can just add the statement
```
using MuonDDDConstants  = MuonGeometryParameters;
```
to `​Geometry/​MuonNumbering/​interface/​MuonDDDConstants.h` when `MuonGeometryParameters` has been fully implemented.
I am suggesting we keep `MuonDDDConstants` as an alias since it is in so many files, but the rest of the names should not contain "DDD".

bsunanda:
@cvuosalo Let me rename it as MuonGeometryConstants

bsunanda:
@cmsbuild Please test

bsunanda:
@cvuosalo I thought "using" is forbidden in the .h files according to the coding rules.

cvuosalo:
@bsunanda I think the "using namespace" rule never made it into the official coding rules. By practice we prohibit a statement like `using namespace std;` in a header file. However, the `using` statement I suggested is equivalent to `typedef`, and that is widely used in our header files.

bsunanda:
@cmsbuild Please test

bsunanda:
@cmsbuild Please test

bsunanda:
@cvuosalo @ianna I do not understand the error mentioned. It is referring to a header file which is in a plugins and it is referred as in interface. Could you help me in removing the error

cvuosalo:
+1

silviodonato:
+1

Issue #29569, opened 2020-04-28
LLVM 10 warning for SimCalorimetry/EcalZeroSuppressionAlgos


smuzaffar:
Hi,
  @civanch and @mdhildreth ,  In latest CLANG IBs we have LLVM 10 which reveal a compilation warning for `SimCalorimetry/EcalZeroSuppressionAlgos/interface/TrivialAmplitudeAlgo.icc`  [a]. This looks like a bug to me, I think the statement should have `&&` instead of `||` here?

[a]
https://cmssdt.cern.ch/SDT/cgi-bin/buildlogs/slc7_amd64_gcc820/CMSSW_11_1_CLANG_X_2020-04-27-2300/SimCalorimetry/EcalZeroSuppressionProducers

```
SimCalorimetry/EcalZeroSuppressionAlgos/interface/TrivialAmplitudeAlgo.icc:34:19: warning: overlapping comparisons always evaluate to true [-Wtautological-overlap-compare]
     if (gain >= 1 || gain <= 3) {
```

smuzaffar:
assign simulation

smuzaffar:
I am in process of fixing other llvm warnings for simulations, so if you can suggest a proper fix then I can include that in my too.

civanch:
@fabiocos , I suspect it is your code. It is likely correct even if the new compiler complains. May be "if" statement better rewrite in the way, which will be accepted by the compiler?

civanch:
@fabiocos , I suspect it is your code. It is likely correct even if the new compiler complains. May be "if" statement better rewrite in the way, which will be accepted by the compiler?

davidlange6:
if (gain >= 1 || gain <= 3) could be replaced by if (true)

civanch:
There are two variants:

if (gain >= 1 && gain <= 3)  or if (gain <= 3)

davidlange6:
not sure what you mean....the compiler is just telling you that

if (gain >= 1 || gain <= 3) {
      Erec = Erec + theWeights[sample] * (frame.sample(sample).adc()) * theGainFactors[gain];
    } else {
      edm::LogWarning("EcalAmplitudeError") << "Wrong gain in frame " << sample << " \n" << frame;
    }

can become

Erec = Erec + theWeights[sample] * (frame.sample(sample).adc()) * theGainFactors[gain];

if its coded correctly - which as @smuzaffar suggests it presumably is not.

fabiocos:
[...]

PR #29597, opened 2020-04-30
ECAL Digitization: fix logical bug in protection against unwanted gainId values in trivial zero suppression

fabiocos:
#### PR description:

This PR addresses the issue #29569 . A logical protection against unwated values of gain identifier is incorrect in the ```TrivialAmplitudeAlgo``` class. As far as I can see the fix is simply replacing ```||``` with ```&&```, i.e. the gain id must be between 1 and 3 included. This class is very old, and practically unused in production workflows as far as I can see (the ECAL selective readout mechanism is used). 

#### PR validation:

Code compiles, wf 11634.0 runs without apparent effect.


fabiocos:
please test

civanch:
+1

silviodonato:
+1

PR #29611, opened 2020-05-01
Extend ConfigBuilder to properly handle special import handling

Dr15Jones:
#### PR description:

The cms.Process object's dumpPython command allows specifying additional 'import' commands that are needed. The ConfigBuilder now includes those when creating a configuration.

#### PR validation:

Using a generator configuration fragment containing the use of an ExternalGeneratorFilter with cmsDriver.py now works where it was failing before.


Dr15Jones:
please test

silviodonato:
+1

PR #29617, opened 2020-05-02
[HBHE] Always use 8 pulses in MAHI without 1-pulse to 8-pulse switch for Run3 and beyond

jaehyeok:
MAHI first runs with 1 pulse that corresponds to SOI, and then moves to the fit with 8 pulses if the chi2 of the 1-pulse fit exceeds 15. The original intention of 1-pulse fit was to reconstruct high energy hits because the contribution from OOTPU is negligible, so using only 1 pulse was thought to be enough. 

But, this had effects on low energy hits as well. For low energy hits the 8-pulse fit is more appropriate because the relative contribution from the OOTPU is large. However, in many cases 1-pulse fit was used instead of 8-pulse fit because the relative uncertainty of the fit is large, and this allows 1-pulse fit to have small chi2. 

We studied the possibility of using 8-pulse fit only, without the 1-pulse to 8-pulse switch [1]. In the slides you can see that 8-pulse fit describes digi better than 1-pulse fit in both high and low energy regimes, so we want to start MAHI with 8 pulses. Same conclusion can be drawn from the study with PU [2].

This change is expected to impact on both data and MC. 

This change is to be effective for Run3 and beyond only.

It passed `runTheMatrix.py -l limited --ibeos`.

[1] https://indico.cern.ch/event/913567/contributions/3841987/attachments/2029001/3395305/Mahi_Chi2Switch_for-Run3_29thApril.pdf
[2] 
https://indico.cern.ch/event/911693/contributions/3857751/attachments/2035263/3407350/Simulation_Run3_PU.pdf


slava77:
@cmsbuild please test

slava77:
@jaehyeok 
did I understand correctly that the study is based on 50 MC events with no PU for Run-3?
- this doesn't seem to have a good coverage
    - without PU noise has a larger fraction of the total number of hits
    - the number of events seems too small
    - the dynamic range in ttbar is still too small to make conclusions for up to the max O(TeV) energy
- shouldn't behavior with PU be checked?
- what about Run-2 (phase-0) detector? IIUC, noise there is larger and noPU 8p is likely to reduce response to the intime pulse

mariadalfonso:
> did I understand correctly that the study is based on 50 MC events with no PU for Run-3?
>

HCAL  is not sensitive to the bunch spacing or soi as ECAL in configuration.
In fact no difference is seen for us between the epsilonPU and PU=0.
[q: where the 50ns is assumed other than the bunchspace = cms.int32(450) ?

> 
> * this doesn't seem to have a good coverage
>   
>   * without PU noise has a larger fraction of the total number of hits
>   * the number of events seems too small
>   * the dynamic range in ttbar is still too small to make conclusions for up to the max O(TeV) energy
> * shouldn't behavior with PU be checked?

the current relVal show indeed that in presence of PU~200 is biased toward 1 pulse only

TTbar PU 2021 RelVal
`https://cms-cpt-software.web.cern.ch/cms-cpt-software/General/Validation/SVSuite/HCAL/1110pre4_PU_110X_mcRun3_2021_realistic_v8-v1_vs_1110pre3_PU_110X_mcRun3_2021_realistic_v8-v1_RelVal_PileUp/TTbar/RecHits/HE_StatusWord.gif`
  - more than ~40% is multipulses

TTabr PU200 2026D49 
https://cms-cpt-software.web.cern.ch/cms-cpt-[software/General/Validation/SVSuite/HCAL/1110pre5_PU25ns_110X_mcRun4_realistic_v3_2026D49PU200-v1_vs_1110pre4_PU25ns_110X_mcRun4_realistic_v3_2026D49PU200-v1_RelVal_PileUp/TTbar/RecHits/HB_StatusWord.gif](url)
  - only ~10% is multiupulses

> * what about Run-2 (phase-0) detector? IIUC, noise there is larger and noPU 8p is likely to reduce response to the intime pulse

For Phase0 MC we should keep the old scheme:
for the HPD detector we do not have a good pulse shape and the multi-fit at high energy give a bad result. The switch in fact was designed for the M2 and we kept that to wait the all-SIPM detector (for which the pulse shape is measured directly with a phase scan).

abdoulline:
@jaehyeok   Jae, 
(1) I've been asking to provide a link to (the ensemble of the presentations at) the meeting
(2) would be good to have a branch discussed first within HCAL before submitting it
(Mari has mentioned above that it's for >=Run3, well We did have HE Phase1 in 2018...) 
I've stumbled on this PR accidentally (not receiving any notification...

slava77:
> > did I understand correctly that the study is based on 50 MC events with no PU for Run-3?
> 
> HCAL is not sensitive to the bunch spacing or soi as ECAL in configuration.
> In fact no difference is seen for us between the epsilonPU and PU=0.
> [q: where the 50ns is assumed other than the bunchspace = cms.int32(450) ?

I didn't get this comment, as it was not a response to "did you test only on 50 events without PU?" 
[I do not expect any difference between PU=0 and epsilonPU due to the way HCAL reco is set up. The difference in ECAL between PU=0 and PU=epsilon is purely from the way how reco is set up, not from how the PU is simulated.]

slava77:
> Maria has mentioned above that this changes is for >=Run3
> (well, we did have Phase1 HE in 2018, which is a special case, but we may not want to opt for going back to 2018...)

some relevant modifiers will need to be used. 
We have run3_HB and run2_HE_2017 for detector-specific behavior settings. However, IIUC, the MAHI settings do not distinguish HB from HE, unless they are coupled in the code via ```hasTimeInfo```.
From the description it sounds like this switch can be made for SIPMs and should not be made for HPDs. So, for a more targeted update some changes in the code will be needed.
A more covering  way, this way applicable only to Run3  is to use ```(run3_HB & run2_HE_2017).toModify(mahiParameters, chiSqSwitch = -1)``` in HBHEMahiParameters_cfi

mariadalfonso:
> > Maria has mentioned above that this changes is for >=Run3
> > (well, we did have Phase1 HE in 2018, which is a special case, but we may not want to opt for going back to 2018...)
> 
> some relevant modifiers will need to be used.
> We have run3_HB and run2_HE_2017 for detector-specific behavior settings. However, IIUC, the MAHI settings do not distinguish HB from HE, unless they are coupled in the code via `hasTimeInfo`.
> From the description it sounds like this switch can be made for SIPMs and should not be made for HPDs. So, for a more targeted update some changes in the code will be needed.
> A more covering way, this way applicable only to Run3 is to use `(run3_HB & run2_HE_2017).toModify(mahiParameters, chiSqSwitch = -1)` in HBHEMahiParameters_cfi

Wouldn't be better to keep the Run2 as it is. Otherwise we (HCAL, PF, jetMET, muon/ele iso or H/E wp) need to recalibrate all HEP17 and HE18 as well (just as exercise, afaik there is no physics need now). It's important to do for Run3 and Phase2 because the PU increase.

slava77:
> Wouldn't be better to keep the Run2 as it is. Otherwise we (HCAL, PF, jetMET, muon/ele iso or H/E wp) need to recalibrate all HEP17 and HE18 as well (just as exercise, afaik there is no physics need now). It's important to do for Run3 and Phase2 because the PU increase.

on one hand, calibrations for the past (not expected to be in production) data are often outdated. The strategy for reco developments is to keep the best algorithmic settings for all detector variants including the past. 

The comment on calibration sounds like this change requires a related update in the calibrations. Will you be able to deliver these on the time scale of closing 11_1  so that this is included in the upcoming rereco for the TDR or should we consider coupled deployment?
This switch change is just in the configuration level and can be done at job submission if some samples are needed for calibration. Once the calibration is available, we could then have the reco config PR together with the updates in the GT. Else, there is danger of introducing inconsistencies in other developments.

abdoulline:
@slava77 
NB: run2_HE_2017  modifier doesn't act separately on HE chiSqSwitch (no separation between HB and HE for this parameter) .
For run3 alone it's simple (as I've suggested in HCAL internal thread): 

from Configuration.Eras.Modifier_run3_common_cff import run3_common
run3_common.toModify(hbheprereco, algorithm = dict(chiSqSwitch = -1.))

slava77:
>  run3_common.toModify(hbheprereco, algorithm = dict(chiSqSwitch = -1.))

this is fine and should work OK, although we were doing this targeted by subdetector when specific to the detector type and here it seems to be the case.
Still, wouldn't you need to test/show that this approach works on real detector (HE) data?

abdoulline:
Slava, I agree in principle - better to use detector-specific modifier (than a era-common), but we don't have it for HCAL in run3 (can be added, in principle),  exists only for run3_HB ( Modifier_run3_HB_cff.py which can be used for both HB and HE specifically for chiSqSwitch, but it would be somewhat misleading...) 
HB and HE - specific modifiers in most of the cases used wherever HE and HE-specific cases can be disentangled (specific parameters exist). 

As to demonstrating that his PR is working (it's certainly working per se, the question is rather how small is the difference) on 2018 HE data - this is a good point...
 
BTW, I'm afraid that Maria has (above) misinterpreted 2021 TTbar PU fraction of 8p, which is 50%(80%) _per_ HB(HE) RecHit, as estimated in my presentation at the aforementioned meeting
 https://indico.cern.ch/event/913567/

slava77:
the tests here have comparisons for Run2. 
E.g. for 2018 wf 136.874
In HE we have  3% fewer hits with non-zero energy and the energy spectrum is roughly unchanged outside of the 0.1-1 GeV range
all_OldVSNew_RunEGamma2018Cwf136p874c_log10HBHERecHitsSorted_hbhereco__reRECO_obj_obj_energy13

In HB we have about 1% fewer hits, but the net effect seems to be an increase in the hit energy up to 2-3 GeV
all_OldVSNew_RunEGamma2018Cwf136p874c_log10HBHERecHitsSorted_hbhereco__reRECO_obj_obj_energy3
I thought that doing 8p vs 1p should decrease the fitted amplitude.
Well, that's at least if the pulse is described well, but it's apparently not the case.

From the first plot, do I understand correctly that we do not have a significant need for recalibration of response for E>1-2 GeV for the phase-1 detector?

abdoulline:
Slava,  thank you for 2018 data plots, which is a good indication (HE one) of what we may expect in 2021 with this PR - quite a small overall effect  
NB: I believe HCAL DPG has a strong preference in favor of Run3-only update.

If we're talking about HCAL IsoTrack calibration for 2021 (MC),  we expect several-% changes after moving from (1+8p) to 8p. Somewhat smaller than what's shown in slide#8: 
https://indico.cern.ch/event/913567/contributions/3845442/attachments/2028912/3395109/S.A._PSG_Apr29_2020.pdf
It would need IsoTrack re-calibration.  Well, we're doing it quite often (along local reco development)...

I have a small tech.question: do you have a strong preference  to **not** use "run3_common" in this case, but to appropriately add "run2_HCAL"  instead? - It doesn't need munch and l can instruct Jae how to do this. 
(Anyway, this PR may need some other modification in addition).

slava77:
> I have a small tech.question: do you have a strong preference **not** to use "run3_common" modifier in this case, but to appropriately add **"run3_HCAL"** instead? - It doesn't need much and l can instruct Jae how to do this.

My variant was/is ```(run3_HB & run2_HE_2017).toModify```
the motivation is to keep this coupled to the appropriate detector combination used in particular in the digitizers.

mariadalfonso:
> > I have a small tech.question: do you have a strong preference **not** to use "run3_common" modifier in this case, but to appropriately add **"run3_HCAL"** instead? - It doesn't need much and l can instruct Jae how to do this.
> 
> My variant was/is `(run3_HB & run2_HE_2017).toModify`
> the motivation is to keep this coupled to the appropriate detector combination used in particular in the digitizers.

This will not work for HLT, as eras are not defined for HLT.
each menu is for one year, so if the desired to to go back and change the Run2 for future legacy of a legacy, then we need to get to the hasTimeInfo().

o for offline: we likely do not use the 11_1 or later for Run2 physics, so if there is a legacy of a legacy win 2026 for sure we will need to re-calibrate all (but we re-calibrate only if there is a physics need).
o for HLT: there is the need to use the 2018 data for the menu studies of the Run3.

slava77:
> This will not work for HLT, as eras are not defined for HLT. ...

So, it sounds like this PR will need to be redone.
However, the tests and the discussion were rather useful.

abdoulline:
@slava77  I didn't pay enough attention to your previous suggestion this morning. That's why run3_HCAL has apeared...
So would it be OK for >=Run3 modification:  to put at the end of HBHEPhase1Reconstructor_cfi.py 
to make it in principal config , where appropriate modifiers are aready present (not to modify lower-level MAHI config):

(run3_HB & run2_HE_2017).toModify(hbheprereco, algorithm = dict(chiSqSwitch = -1.))

slava77:
> So would it be OK for >=Run3 modification: to put at the end of HBHEPhase1Reconstructor_cfi.py
> to make it in principal config , where appropriate modifiers are aready present (not to modify lower-level MAHI config):
> 
> (run3_HB & run2_HE_2017).toModify(hbheprereco, algorithm = dict(chiSqSwitch = -1.))

what is the motivation for this exception?
The normal rule is to apply modifiers in the same file where the parameters are defined.
Is the same "mahiParameters" used somewhere else for run3 configs where a different value is needed?

abdoulline:
I wanted to "move" re-definition to the  HBHEPhase1Reconstructor_cfi.py because:
(1)  it requires 1 line of modification, not 3 as for HBHEMahiParameters_cfi.py
(2) the modification is more "visible" in the former. 
But if you think it's important:
"The normal rule is to apply modifiers in the same file where the parameters are defined."
there is no problem (below I list the lines explicitly for Jae)  at the bottom of  HBHEMahiParameters_cfi.py to add : 

from Configuration.Eras.Modifier_run2_HE_2017_cff import run2_HE_2017
from Configuration.Eras.Modifier_run3_HB_cff import run3_HB
#--- >= Run3 modification: 
(run3_HB & run2_HE_2017).toModify(mahiParameters, chiSqSwitch = -1)

jaehyeok:
I have updated this PR with 

> from Configuration.Eras.Modifier_run2_HE_2017_cff import run2_HE_2017
from Configuration.Eras.Modifier_run3_HB_cff import run3_HB
#--- >= Run3 modification:
> (run3_HB& run2_HE_2017).toModify(mahiParameters, chiSqSwitch = -1.0)

In addition, I’ve updated the HLT customization so that `chiSqSwitch=-1.0` and `applyLegacyHBMCorrection=False` can be set by a customization function. Note that `applyLegacyHBMCorrection=False` is from PR26177, but was not propagated to HLT customization when the PR was submitted. So, we take this time as an opportunity to do it.

HLTrigger/Configuration/python/customizeHLTforCMSSW.py
```diff
@@ -164,6 +170,13 @@ def customiseFor2017DtUnpacking(process):
 
     return process
 
+# Use 8p fit (PR29617) and don't apply HB- correction (PR26177)
+def customiseFor29617(process):
+    for producer in producers_by_type(process, "HBHEPhase1Reconstructor"):
+        producer.algorithm.chiSqSwitch = cms.double(-1.0)
+        producer.algorithm.applyLegacyHBMCorrection = cms.bool( False )
+    return process
+
```
Martin-Grunewald, line 179: This subroutine should be called (see end of the file).

intrepid42:
Hi, just let me point out that the RecHit chi2 is nicely reduced, in agreement with the studies presented on [29 April](https://indico.cern.ch/event/913567/). Plots are from [here](https://cmssdt.cern.ch/SDT/jenkins-artifacts/baseLineComparisons/CMSSW_11_1_X_2020-05-01-2300+a7f53d/36105/validateJR/all_OldVSNew_RunEGamma2018Cwf136p874/), as the energy plots sent by Slava earlier.

**HB:**
image
**!HB:**
image

slava77:
> Hi, just let me point out that the RecHit chi2 is nicely reduced, in agreement with the studies presented on [29 April](https://indico.cern.ch/event/913567/). Plots are from [here](https://cmssdt.cern.ch/SDT/jenkins-artifacts/baseLineComparisons/CMSSW_11_1_X_2020-05-01-2300+a7f53d/36105/validateJR/all_OldVSNew_RunEGamma2018Cwf136p874/), as the energy plots sent by Slava earlier.

I'm trying to understand if a reduction is an indication of improvement in the fit or rather a change in the underlying number of degrees of freedom.
Please remind me what is the expected distribution for chi2 with 8 pulse and 1p in an ideal case.
IIUC, with 8 TS inputs, a fit for 8 pulses has no chi2 and it's apparently via the algorithm which turns off pulses in iterations that we are getting some degrees of freedom.
So, isn't 1p  expected to have a chi2 around 7 (7 dof: 8TS in, one amplitude out) and then the 8p is probably to be around 1 by construction?

slava77:
@cmsbuild please test

HLTrigger/Configuration/python/customizeHLTforCMSSW.py
jaehyeok: Thanks for comment. It's now called in `customizeHLTforCMSSW`.

jaehyeok:
@slava77 

In MAHI chi2 is just an indicator of goodness of fit and is expected to be generally smaller in the 8p fit than the 1p fit due to increased ndof as you mentioned. The best metric we have is the pulse display (comparison between digis and fitted pulses) and we've seen that the 8p fit fits digis better than the 1p fit in both low and high energy regions [1]. High stats test is running to check the performance in even higher energy regime. We will let you know the results when ready. 

[1] https://indico.cern.ch/event/913567/contributions/3841987/attachments/2029001/3395305/Mahi_Chi2Switch_for-Run3_29thApril.pdf

slava77:
> In MAHI chi2 is just an indicator of goodness of fit and is expected to be generally smaller in the 8p fit than the 1p fit due to increased ndof as you mentioned. The best metric we have is the pulse display (comparison between digis and fitted pulses) and we've seen that the 8p fit fits digis better than the 1p fit in both low and high energy regions [1]. High stats test is running to check the performance in even higher energy regime. We will let you know the results when ready.

Thanks, still if all 8 pulses are actually in the fit using 8 inputs from timeslices, the resulting display will always have a perfect agreement unless there is a bug.

I see that the pulse displays show a different number of total pulses fitted. Does this correspond to the choice of non-zero amplitude after the NNLS minimization?
Is this variable saved somewhere in the aux? it seems to have a significant importance for the quality of the fit (mainly because it is best connected with the number of degrees of freedom).

slava77:
@cmsbuild please test

silviodonato:
please rebase the PR to a more recent CMSSW release

jaehyeok:
Rebased to the most recent CMSSW release `CMSSW_11_1_X_2020-05-05-1100`

slava77:
@cmsbuild please test

fwyzard:
@jaehyeok
> We studied the possibility of using 8-pulse fit only, without the 1-pulse to 8-pulse switch [1]. In the slides you can see that 8-pulse fit describes digi better than 1-pulse fit in both high and low energy regimes, so we want to start MAHI with 8 pulses.

What is the additional "cost" in terms of cpu time from always running 8-pulse fit ?

mariadalfonso:
@fwyzard 

If we go from the 1+8 pulses with chi2 switch to directly 8 pulses fit, the  hbherereco is slower only of 6% .
Test done on fu-c2a02-37-03 with run321177 2018 data on top of 11_1_0_pre4.

fwyzard:
thanks

slava77:
@jaehyeok 
please update the PR description (or even the title) to cover that this is only for run3+.
Thank you.

jaehyeok:
@slava 
the title and the description now say that this is for Run3+.

HLTrigger/Configuration/python/customizeHLTforCMSSW.py
```diff
@@ -183,11 +189,19 @@ def customiseFor29512(process):
 
     return process
 
+# Use 8p fit (PR29617) and don't apply HB- correction (PR26177)
+def customiseFor29617(process):
+    for producer in producers_by_type(process, "HBHEPhase1Reconstructor"):
+        producer.algorithm.chiSqSwitch = cms.double(-1.0)
+        producer.algorithm.applyLegacyHBMCorrection = cms.bool( False )
+    return process
```
slava77, line 197: I was puzzled at first by the change in the plots in the HI Run2 workflow 
https://cmsweb.cern.ch/dqm/dev/start?runnr=1;dataset%3D/RelVal_wf158_0_pr/CMSSW_11_1_X-PRcmssw_29617-6107/DQMIO;referenceshow%3Dall;referencenorm=False;referenceobj1%3Dother::/RelVal_wf158_0_base/CMSSW_11_1_X-PRcmssw_29617-6107/DQMIO::;sampletype%3Doffline_relval;workspace%3DEverything;

The changes in this file are not here to patch the config to keep the HLT running.
Instead  these are to change its default behavior. Isn't this a wrong way to do it? 
If the goal is to change existing parameters in the existing tables, shouldn't there be a JIRA ticket instead?

slava77:
for https://github.com/cms-sw/cmssw/pull/29617	36fcd78 
- the change is as discussed earlier in the thread and is expected to affect only Run3 and later reco
- jenkins tests pass and in reco comparisons show changes only in Run-3 and phase-2 workflows, which start from hbheprereco and propagate downstream
    - there is also a change in the Run2 HI wf 158.0, which is apparently not expected (assuming changes should really be only in Run3 and that the wf 158.0 is not buggy by itself)

in 2021 wf without PU 11634
HB energy (above 1MeV) is little changed
all_OldVSNew_TTbar14TeV2021wf11634p0c_log10HBHERecHitsSorted_hbhereco__RECO_obj_obj_energy3
and so is the HE
all_OldVSNew_TTbar14TeV2021wf11634p0c_log10HBHERecHitsSorted_hbhereco__RECO_obj_obj_energy13


However, in 2023 wf without PU 12434:
while the HB energy (above 1MeV) is little changed
all_OldVSNew_TTbar14TeV2023wf12434p0c_log10HBHERecHitsSorted_hbhereco__RECO_obj_obj_energy3
the HE has some more hits above threshold
all_OldVSNew_TTbar14TeV2023wf12434p0c_log10HBHERecHitsSorted_hbhereco__RECO_obj_obj_energy13

Is the HE more noisy in 2023 and now the plot is dominated by noise which is not expected to be fitted well (there is a 7 fold increase of hits in HE compared to the 2021 setup) or is there some other problem in the 2023 setup?

abdoulline:
@slava77 

Slava, I'm also puzzled about wf 158.  There is one single histo changed...
GlobalRecHits/Hcal is not HCAL DQM domain, may be they're missing appropriate era customization?   

But there are also muon stuff differences (wf 1330.0):
 https://cmsweb.cern.ch/dqm/dev/session/V6SxPS

abdoulline:
For 2023 wrt 2021 we have quite high Dark Current (DC) noise, which easily goes above unchanged (from 2021)  ZeroSuppression (ZS) and HE occupancy becomes very high even wo PU.

abdoulline:
NB: I've edmConfigRump-ed step2 of wf 158.0  and it shows (as expected):  
chiSqSwitch = cms.double(15.0)

slava77:
> NB: I've edmConfigRump-ed step2 & step3 of wf 158.0 and both show (as expected) in all the instances of producer:
> chiSqSwitch = cms.double(15.0)

Well, now I doubt your setup, because I can understand the change in step2 config to be from the HLT config change.
I was puzzled "at first" about 158.0. 
The lack of differences is explained by the fact that this workflow has non-trivial HLT in step2, but step3 has no HLT in the validation/DQM ```VALIDATION:@standardValidationNoHLT+@miniAODValidation,DQM:@standardDQMFakeHLT+@miniAODDQM```
and the only place to pick it up is in the global hit validation which reads all collections by type.
IIUC, this workflow in the jenkins short tests is the only workflow with non-trivial HLT in step2 for run2.

slava77:
> HE: for 2023 wrt 2021 we have quite a high Dark Current (DC) noise (+ Gains increase), which easily goes above unchanged (from 2021) ZeroSuppression (ZS) threshold and HE occupancy becomes very high even wo PU.
> HB: actual 2023 condition are similar to 2021 (at the moment), means no aging applied yet.

Do I understand that the more significant changes are expected  in 2023?

slava77:
> Well, now I doubt your setup, because I can understand the change in step2 config to be from the HLT config change.

to be specific, in CMSSW_11_1_X_2020-05-05-1100, comparing the IB with that including this PR
after dumping step2\*py from 158.0 I see in a diff
```
... in  process.hltHbhereco = cms.EDProducer("HBHEPhase1Reconstructor",
+        applyLegacyHBMCorrection = cms.bool(False),
-        chiSqSwitch = cms.double(15.0),
+        chiSqSwitch = cms.double(-1.0),
```

abdoulline:
> Well, now I doubt your setup, because I can understand the change in step2 config to be from the HLT config change.

Ah... Indeed, I was too fast... I didn't use this PR (with HLT customization...) , but my version of test area without HLT :(

abdoulline:
> > HE: for 2023 wrt 2021 we have quite a high Dark Current (DC) noise (+ Gains increase), which easily goes above unchanged (from 2021) ZeroSuppression (ZS) threshold and HE occupancy becomes very high even wo PU.
> > HB: actual 2023 condition are similar to 2021 (at the moment), means no aging applied yet.
> 
> Do I understand that the more significant changes are expected in 2023?

Not sure I understand your question correctly, so two answers: 
 (1) 2023 is supposedly more affected by this PR in HE (high DC noise, many fired readouts) than 2021;
(2) realistic 2023 (wrt actual 2023 situation/conditions) will eventually have: 
- quite higher DC noise in HB  (than now both in 2021 and 2023 GTs) but
-  smaller (than now in 2023 GT) in HE.

HLTrigger/Configuration/python/customizeHLTforCMSSW.py
mariadalfonso: @slava77 
* synchronizeHCALHLTofflineRun3on2018data and synchronizeHCALHLTofflineRun2 are function to let now play and testing in old data/menu
* customiseFor29617: is for the Run3 menu

For a modification in the confDB we usually propose a change in a customization function like this, then @Martin-Grunewald/TSG import the modification in the ConfDB in the Run3 menu and dump the menu the release. A JIRA is also needed to activate TSG. Will do.

slava77:
> > Do I understand that the more significant changes are expected in 2023?
> 
> Not sure I understand your question correctly, so two answers:
> (1) 2023 is supposedly more affected by this PR in HE (high DC noise, many fired readouts) than 2021;
> (2) realistic 2023 (wrt actual 2023 situation/conditions) will eventually have:
> 
>     * quite higher DC noise in HB  (than now both in 2021 and 2023 GTs) but
> 
>     * smaller (than now in 2023 GT) in HE.

I was just trying to understand the "sign" of the change from this PR: apparently (unrelated to this PR) there is more noise in HE; the question is if MAHI is expected to behave like this for the noise, to produce 10% more hits with non-zero energy by going from 1 pulse to 8 pulse in the fitting. If the noise were outside of SOI, then the 8pulse should lead to a reduced response, perhaps some slice display would help.

abdoulline:
Slava, specifically for 158.0 issue  - do you see any (HLT) workaround?

HLTrigger/Configuration/python/customizeHLTforCMSSW.py
slava77: > customiseFor29617: is for the Run3 menu

however, it apparently applies to the Run2 menu used in wf 158.0

slava77:
> Slava, specifically for 158.0 issue - do you see any (HLT) workaround?

eras are known during building the config. So, it should supposedly work to add the same ```(run3_HB & run2_HE_2017).toModify(process. ... ``` in ```customiseFor29617```

mariadalfonso:
> > Slava, specifically for 158.0 issue - do you see any (HLT) workaround?
> 
> eras are known during building the config. So, it should supposedly work to add the same `(run3_HB & run2_HE_2017).toModify(process. ... ` in `customiseFor29617`

HLT doesn't know about the era and the whole function is going to be removed as soon as the the modifications are in the confDB.

slava77:
> HLT doesn't know about the era

I suppose you meant here the actual HLT processing on the farm. For that indeed, the final change will be in the confDB.
The cmsDriver-based configs made by runTheMatrix know about the eras just fine
and I checked that it works in 158.0

slava77:
@jaehyeok 
there is a merge conflict here now with customizeHLTforCMSSW.py.
Please rebase.

For the reco signature I still need a confirmation that the behavior observed in 2023 wf comparisons does not require any further changes in this PR (and if any mitigation, it will be handled elsewhere, still with 8p approach). 
@jaehyeok @mariadalfonso @abdoulline please confirm.
Thank you.

jaehyeok:
@slava 

(1) The PR has been rebased to  ` CMSSW_11_1_X_2020-05-10-2300`.

(2) `(run3_HB & run2_HE_2017).toModify(process. ...` has been add to `customiseFor29617`.

(3) We confirm that "the behavior observed in 2023 wf comparisons does not require any further changes in this PR". The reason is explained in Salavat's comment (https://github.com/cms-sw/cmssw/pull/29617#issuecomment-625748661).

(4) We now have performed a study with PU [1] (500 events). Same conclusion can be drawn; 8p looks better than 1p fit in both high and low energy hits (you can look at the pulse displays in the backup).

[1] https://indico.cern.ch/event/911693/contributions/3857751/attachments/2035263/3407350/Simulation_Run3_PU.pdf

perrotta:
@cmsbuild please test

Martin-Grunewald:
-1
Sorry, that will not work for HLT! No eras. 

HLT menues (into which these customisations will eventuall be absorbed) are 'fully expanded' (Run 3),
also we need the explicit code changes in the customisation, not hidden inside an era modifier.

Any workflows left dangling should then switch to a fake menu.

abdoulline:
> -1
> Sorry, that will not work for HLT! No eras.
> 
> HLT menues (into which these customisations will eventuall be absorbed) are 'fully expanded' (Run 3),
> also we need the explicit code changes in the customisation, not hidden inside an era modifier.
> 
> Any workflows left dangling should then switch to a fake menu.

@Martin-Grunewald 
as HLT expert, could you, please, provide a "(1)(2)(3)"-kind of explicit recipe for non-HLT-experts on how to overcome this (like extensively discussed 158.0  wf) particular issue? 
So that we could move on with RECO update specifically for Run3+

abdoulline:
@slava77  @Martin-Grunewald 
to avoid a situation "between Scylla and Charybdis" (RECO and HLT)
may I assume that something like the following be a good compromise? - 
   
(1) revert HLT customization to the previous (no-era) simple customization intended solely for Run3+; 
(2) ignore 158.0 - like issue for now, as it's kind of transient one and we know what happens. 
Please, suggest.

HLTrigger/Configuration/python/customizeHLTforCMSSW.py
```diff
@@ -196,12 +202,23 @@ def customiseFor29658(process):
...
+
 # CMSSW version specific customizations
 def customizeHLTforCMSSW(process, menuType="GRun"):
 
     # add call to action function in proper order: newest last!
     # process = customiseFor12718(process)
     process = customiseFor29512(process)
     process = customiseFor29658(process)
+    process = customiseFor29617(process)
```
slava77, line 222: it seems like the menuType can be passed to the customiseForNNNN, which would apparently be the better solution in the context of not using eras

Martin-Grunewald: Remember, this customisation code is just temporary. Eventually, we'll move the customisations directly into the menu in ConfDB and re-generate the HLT menu dumps.
The development menus in CMSSW_11_1 now must correspond to run3.
Workflow 158.0 is using HLT:HIon, so either it is updated to/replaced by a Run3 workflow (it is MC after all), OR it should be modified to use the HLT fake2 menu.

Martin-Grunewald:
That would be fine with me...

slava77:
> @slava77 @Martin-Grunewald
> to avoid a situation "between Scylla and Charybdis" (RECO and HLT)
> may I assume that something like the following be a good compromise? -
> 
> (1) revert HLT customization to the previous (no-era) simple customization intended solely for Run3+;
> (2) ignore 158.0 - like issue for now, as it's kind of transient one and we know what happens.
> Please, suggest.

fine with me

jaehyeok:
@slava77 @Martin-Grunewald 

Thanks very much! I just reverted it to the (no-era) simple customization.

slava77:
@cmsbuild please test

slava77:
+1

for https://github.com/cms-sw/cmssw/pull/29617	7677235 
- jenkins tests pass and comparisons with the baseline show differences as described in https://github.com/cms-sw/cmssw/pull/29617#issuecomment-625652875
    - after some iteration, the change in Run2 HLT wf 158.0 was considered acceptable (as temporary), despite the main target of this PR to be the Run-3 HCAL reco.

Martin-Grunewald:
+1

silviodonato:
+1

PR #29627, opened 2020-05-04
Restructured code of againstElectronDeadECAL tauID

mbluj:
#### PR description:
Restructured code of againstElectronDeadECAL discriminator to run on either AOD or miniAOD inputs:
- add a version of the discriminant to run on top of pat::Taus to be used with miniAOD data samples;
- use eta and phi of impact position of the leading PFChargedHadron on ECal surface instead of eta and phi of tau at the vertex (taking bending of track in the magnetic field and the zVertex into account); it can be disabled for backwards compatibility; 
- restructure the code in the way that both AOD and miniAOD version of the discriminant call the same algorithm to avoid code duplication;
- move code preforming extrapolation to the ECal surface from the againstElectronMVA algorithm to an auxiliary helper class which avoids code duplication;
- add output of the againstElectronDeadECAL discriminant produced at AOD (already present in RECO/AOD workflow) to `slimmedTaus` in MiniAOD

Expected some differences:
- small difference in output of PFRecoTauDiscriminationAgainstElectronDeadECAL (AOD) discriminant because of enabled extrapolation to the ECAL (version with the extrapolation disabled gives identical results as one before the changes in this PR);
- new tauID embedded into slimmedTaus in miniAOD.

In addition it is indented to backport the development to releases used currently to analyze data (102X, 106X). 
It is also intended to add the discriminant to nanoAOD, but it will go via nanoAOD repository (with link to foreseen backports).

#### PR validation:

`runTheMatrix.py -l limited -i all` passed successfully except a few failed due to problem with inputs (DAS_ERROR).


perrotta:
please test

slava77:
> In addition it is indented to backport the development to releases used currently to analyze data (102X, 106X).

will this backport allow for the default tau ID/reco to stay unchanged using appropriate configuration?

perrotta:
@mbluj A rebase is needed because of the modified BuildFile

perrotta:
Wrong rebase
Please follow http://cms-sw.github.io/tutorial-resolve-conflicts.html

mbluj:
Code restored to state before first rebase (it was plain `git rebase CMSSW_11_1_X` which caused problems) and then rebased with `git cms-rebase-topic cms-tau-pog:CMSSW_11_1_X_tau-pog_deadECAL -n CMSSW_11_1_X` which is hopefully fine.

slava77:
@cmsbuild please test

santocch:
+1

RecoTauTag/RecoTau/src/PositionAtECalEntrance.cc
```diff
@@ -0,0 +1,38 @@
...
+                  math::XYZTLorentzVector(particle->vertex().x(), particle->vertex().y(), particle->vertex().z(), 0.),
+                  particle->charge()),
+      0.,
+      0.,
+      bField_z_);
+  propagator.propagateToEcalEntrance(false);
+  reco::Candidate::Point position;
+  if (propagator.getSuccess() != 0) {
+    position = reco::Candidate::Point(
+        propagator.particle().vertex().x(), propagator.particle().vertex().y(), propagator.particle().vertex().z());
```
slava77, line 32: this looks like ```position = propagator.particle().vertex().Vect();```

RecoTauTag/RecoTau/src/PositionAtECalEntrance.cc
```diff
@@ -0,0 +1,38 @@
...
+void PositionAtECalEntrance::beginEvent(const edm::EventSetup& es) {
+  edm::ESHandle<MagneticField> bField;
+  es.get<IdealMagneticFieldRecord>().get(bField);
+  bField_z_ = bField->inTesla(GlobalPoint(0., 0., 0.)).z();
+}
+
+reco::Candidate::Point PositionAtECalEntrance::operator()(const reco::Candidate* particle, bool& success) const {
+  assert(bField_z_ != -1.);
+  BaseParticlePropagator propagator = BaseParticlePropagator(
+      RawParticle(math::XYZTLorentzVector(particle->px(), particle->py(), particle->pz(), particle->energy()),
```
slava77, line 22: the first ```math::XYZTLorentzVector``` construction looks like ```particle->p4()```

RecoTauTag/RecoTau/plugins/PATTauDiscriminationAgainstElectronDeadECAL.cc
```diff
@@ -0,0 +1,80 @@
...
+#include "FWCore/Framework/interface/EventSetup.h"
+#include <FWCore/ParameterSet/interface/ConfigurationDescriptions.h>
+#include <FWCore/ParameterSet/interface/ParameterSetDescription.h>
+#include "DataFormats/TauReco/interface/PFTau.h"
+#include "DataFormats/TauReco/interface/PFTauFwd.h"
+#include "DataFormats/TauReco/interface/PFTauDiscriminator.h"
+#include "RecoTauTag/RecoTau/interface/TauDiscriminationProducerBase.h"
+#include "RecoTauTag/RecoTau/interface/AntiElectronDeadECAL.h"
+
+class PATTauDiscriminationAgainstElectronDeadECAL : public PATTauDiscriminationProducerBase {
```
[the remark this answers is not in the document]
slava77: considering that most of the code in PFRecoTauDiscriminationAgainstElectronDeadECAL and this file is now rather "boilerplate", the need for a template is not very strong. Still, it would be better if the implementation comes out rather straightforward.

RecoTauTag/RecoTau/interface/PositionAtECalEntrance.h
```diff
@@ -0,0 +1,32 @@
...
+#include "DataFormats/Candidate/interface/Candidate.h"
+
+class PositionAtECalEntrance {
+public:
+  PositionAtECalEntrance();
+  ~PositionAtECalEntrance();
+
+  void beginEvent(const edm::EventSetup&);
+
+  reco::Candidate::Point operator()(const reco::Candidate* particle, bool& success) const;
```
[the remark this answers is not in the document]
mbluj: OK, noted and a comment added to the code.

RecoTauTag/RecoTau/python/PATTauDiscriminationAgainstElectronDeadECAL_cfi.py
```diff
@@ -0,0 +1,26 @@
+import FWCore.ParameterSet.Config as cms
+
+from RecoTauTag.RecoTau.TauDiscriminatorTools import noPrediscriminants
+
+patTauDiscriminationAgainstElectronDeadECAL = cms.EDProducer(
```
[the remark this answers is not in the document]
mbluj: Removed.

RecoTauTag/RecoTau/python/PFRecoTauDiscriminationAgainstElectronDeadECAL_cfi.py
```diff
@@ -1,4 +1,5 @@
 import FWCore.ParameterSet.Config as cms
+
 from RecoTauTag.RecoTau.TauDiscriminatorTools import requireLeadTrack
 
 pfRecoTauDiscriminationAgainstElectronDeadECAL = cms.EDProducer(
```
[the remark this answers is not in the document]
mbluj: Removed, but it should be noted that similar issue affects other cfi of tau modules. This will be a subject of some future cleaning campaign.

RecoTauTag/RecoTau/src/AntiElectronDeadECAL.cc
```diff
@@ -0,0 +1,170 @@
...
+    }
+  }
+}  // namespace
+
+void AntiElectronDeadECAL::updateBadTowers(const edm::EventSetup& es) {
+  // NOTE: modified version of SUSY CAF code
+  //         UserCode/SusyCAF/plugins/SusyCAF_EcalDeadChannels.cc
+  const uint32_t channelStatusId = es.get<EcalChannelStatusRcd>().cacheIdentifier();
+  const uint32_t caloGeometryId = es.get<CaloGeometryRecord>().cacheIdentifier();
+  const uint32_t idealGeometryId = es.get<IdealGeometryRecord>().cacheIdentifier();
```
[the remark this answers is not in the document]
mbluj: Done, but it will be welcome that you check this implementation. My test passed successfully, but I am not sure indeed conditions were uploaded when changed.

slava77:
@cmsbuild please test

RecoTauTag/RecoTau/plugins/TauDiscriminationAgainstElectronDeadECAL.cc
```diff
@@ -0,0 +1,106 @@
...
+std::string getTauTypeNameString<reco::PFTau>(bool capitalise) {
+  return capitalise ? "PFRecoTau" : "pfRecoTau";
+}
+template <>
+std::string getTauTypeNameString<pat::Tau>(bool capitalise) {
+  return capitalise ? "PATTau" : "patTau";
+}
+
+template <class TauType, class TauDiscriminator>
+class TauDiscriminationAgainstElectronDeadECALBase : public TauDiscriminationProducerBase<TauType, TauDiscriminator> {
```
slava77, line 42: Why ```*DeadECALBase``` is it supposed to be used as a base class (it wasn't in the earlier version of this PR)?
Please drop -> ```DeadECAL```.

RecoTauTag/RecoTau/plugins/TauDiscriminationAgainstElectronDeadECAL.cc
```diff
@@ -0,0 +1,106 @@
...
+        moduleLabel_(cfg.getParameter<std::string>("@module_label")),
+        verbosity_(cfg.getParameter<int>("verbosity")),
+        antiElectronDeadECAL_(cfg) {}
+  ~TauDiscriminationAgainstElectronDeadECALBase() override {}
+
+  void beginEvent(const edm::Event& evt, const edm::EventSetup& es) override { antiElectronDeadECAL_.beginEvent(es); }
+
+  double discriminate(const TauRef& tau) const override {
+    if (verbosity_) {
+      edm::LogPrint(getTauTypeNameString<TauType>(true) + "AgainstEleDeadECAL")
```
slava77, line 57: The base ```TauDiscriminationProducerBase.h``` has ```std::string getProducerString()``` (which should probably be moved inside the class as a static method).
It seems more practical to use the existing interface/design.

RecoTauTag/RecoTau/plugins/TauDiscriminationAgainstElectronDeadECAL.cc
```diff
@@ -0,0 +1,106 @@
...
+
+  desc.add<double>("dR", 0.08);
+  desc.add<unsigned int>("minStatus", 12);
+  desc.add<bool>("extrapolateToECalEntrance", true);
+  desc.add<int>("verbosity", 0);
+
+  TauDiscriminationProducerBase<TauType, TauDiscriminator>::fillProducerDescriptions(
+      desc);  // inherited from the base-class
+
+  descriptions.add(getTauTypeNameString<TauType>(false) + "DiscriminationAgainstElectronDeadECAL", desc);  //base
```
slava77, line 97: doesn't ```addWithDefaultLabel```  work here?
(I can imagine that the capitalization may be incorrect; but please check the default method if it works already.)

RecoTauTag/RecoTau/plugins/TauDiscriminationAgainstElectronDeadECAL.cc
mbluj: In the earlier version it inherited from specialized versions (for `reco::PFTau` or `pat::Tau`) of template class `TauDiscriminationProducerBase<TauType, TauDiscriminator>` for a generic tau discriminant. Now the template is for `TauDiscriminationAgainstElectronDeadECAL` for any TauType, so one more sensible renaming (if mandatory) will be removal of `Base`.

RecoTauTag/RecoTau/plugins/TauDiscriminationAgainstElectronDeadECAL.cc
mbluj: Well, `std::string getProducerString()` and `getTauTypeNameString(bool)` do similar but not identical thing and I am not sure if they can be really unified w/o adding some logic to them. The helper function(s) can be moved to become static methods of the `TauDiscriminationProducerBase.h` class, but I do not see a big gain from it.

RecoTauTag/RecoTau/plugins/TauDiscriminationAgainstElectronDeadECAL.cc
mbluj: We would like to keep uniform capitalisation over cfi's of tau modules. In addition, `std::string getTauTypeNameString(bool)` is used to provide TauType dependent string allowing to keep log printouts in coherent form with other tau modules (and previous version of this one). The latter is not critical, but nice feature.

RecoTauTag/RecoTau/plugins/TauDiscriminationAgainstElectronDeadECAL.cc
slava77: From the feedback, there seems to be a good case to move ```getTauTypeNameString``` to TauDiscriminationProducerBase.h

RecoTauTag/RecoTau/plugins/TauDiscriminationAgainstElectronDeadECAL.cc
slava77: >  so one more sensible renaming (if mandatory) will be removal of `Base`.

that's what I meant, actually, but ended up typing in a bit confusing way. 
(we'll have ```TauDiscriminationAgainstElectronDeadECAL```)

RecoTauTag/RecoTau/plugins/TauDiscriminationAgainstElectronDeadECAL.cc
slava77: OK

mbluj: Checked that `addWithDefaultLabel` indeed works here as wanted, so adopted.

RecoTauTag/RecoTau/plugins/TauDiscriminationAgainstElectronDeadECAL.cc
mbluj: After some thinking I realized that making `std::string getProducerString()` a method of template `TauDiscriminationProducerBase` class requires partial specialization of the class which is not possible. Do I miss something?
Moving of  `getTauTypeNameString()` helper to `TauDiscriminationProducerBase.h` header file (or alternatively usage of another generic helper from the header) leads to errors like `invalid declaration of '::getTauTypeNameString<TauType>'`... It could be that it is something trivial, but I am not able to solve it. Help will be appreciated.

slava77: ah, right, it's less straightforward.
Since there are just two types to care about, it can be implemented in the method, as in the following example
```c++
template <typename T, typename T1> class A {
public:
  static std::string a() {
    if (std::is_same<T, int>::value) return "int";
    if (std::is_same<T, bool>::value) return "bool";
    throw std::logic_error("BAD");
  }
```

mbluj: OK, thank you. It can be done like this.

slava77:
@cmsbuild please test

RecoTauTag/RecoTau/plugins/TauDiscriminationAgainstElectronDeadECAL.cc
```diff
@@ -21,51 +20,36 @@
...
 
   void beginEvent(const edm::Event& evt, const edm::EventSetup& es) override { antiElectronDeadECAL_.beginEvent(es); }
 
   double discriminate(const TauRef& tau) const override {
     if (verbosity_) {
-      edm::LogPrint(getTauTypeNameString<TauType>(true) + "AgainstEleDeadECAL")
-          << "<" + getTauTypeNameString<TauType>(true) + "AgainstElectronDeadECAL::discriminate>:";
-      edm::LogPrint(getTauTypeNameString<TauType>(true) + "AgainstEleDeadECAL") << " moduleLabel = " << moduleLabel_;
-      edm::LogPrint(getTauTypeNameString<TauType>(true) + "AgainstEleDeadECAL")
+      edm::LogPrint(TauDiscriminationProducerBase<TauType, TauDiscriminator>::getTauTypeString() + "AgainstEleDeadECAL")
```
slava77, line 39: I'm a bit surprised to see the prefix.
I was expecting that ```getTauTypeString()``` alone should work

slava77: ```using TauDiscriminationProducerBase<TauType, TauDiscriminator>::getTauTypeString;```  somewhere in the scope of ```TauDiscriminationAgainstElectronDeadECAL``` should allow to avoid the lengthy prefix

slava77:
+1

for https://github.com/cms-sw/cmssw/pull/29627	407f38c 
    - addressing https://github.com/cms-sw/cmssw/pull/29627#pullrequestreview-414862823 would be nice, but is somewhat acceptable
- jenkins tests pass and comparisons with the baseline show differences in the tau ID variables: 
    - the reco comparisons are full of false-positive diffs due to reordering/reshuffling in the ```tauIDs_``` vectors; the only real change here is a one new ID variable
    - DQM plots show no differences, mostly as expected (existing IDs are not changed, even though small changes in ```PFRecoTauDiscriminationAgainstElectronDeadECAL``` could be anticipated)

santocch:
+1

RecoTauTag/RecoTau/plugins/TauDiscriminationAgainstElectronDeadECAL.cc
mbluj: No it doesn't work alone without the prefix (compiler doesn't know actual type of the method), `using TauDiscriminationProducerBase<TauType, TauDiscriminator>::getTauTypeString;` doesn't help, too. The trick can be use of `this->getTauTypeString()`, which is more compact, so I adopt this.

slava77:
@cmsbuild please test

slava77:
+1

for https://github.com/cms-sw/cmssw/pull/29627	84762b7 
- syntax improvements since the last reco signoff 
    - @mbluj thank you for the update

santocch:
+1

silviodonato:
+1

mbluj:
Just to be sure: this PR is merged to master, but in time of its review main development series have been switched from 11_1_X to 11_2_X, so the question is if it will appear in both release series or a "backport" to 11_1_X is needed? I am asking as at least now comparison on my development branch with CMSSW_11_1_X shows differences.

perrotta:
> Just to be sure: this PR is merged to master, but in time of its review main development series have been switched from 11_1_X to 11_2_X, so the question is if it will appear in both release series or a "backport" to 11_1_X is needed? I am asking as at least now comparison on my development branch with CMSSW_11_1_X shows differences.

Master is now 11_2_X, and pull requests are (only) merged in it.
If you think this PR is eligible for entering the bug_fix 11_1_0_pre9 pre-release, please prepare a backport pull request. The same git branch should still be mergeable also in 11_1_X.

mbluj:
> 
> 
> > Just to be sure: this PR is merged to master, but in time of its review main development series have been switched from 11_1_X to 11_2_X, so the question is if it will appear in both release series or a "backport" to 11_1_X is needed? I am asking as at least now comparison on my development branch with CMSSW_11_1_X shows differences.
> 
> Master is now 11_2_X, and pull requests are (only) merged in it.
> If you think this PR is eligible for entering the bug_fix 11_1_0_pre9 pre-release, please prepare a backport pull request. The same git branch should still be mergeable also in 11_1_X.

Thanks Andrea! I'm going to prepare the PR to 11_1_X

PR #29634, opened 2020-05-04
Handle Depth 7-only Signal in HcalTriggerPrimitiveAlgo

JHiltbrand:
#### PR description:

This PR addresses the issue described in JIRA ticket [CMSHCALTRG-9](https://its.cern.ch/jira/browse/CMSHCALTRG-9). When a signal is only in depth 7 of HCAL TT26, TT27, or TT28, the corresponding `HcalTriggerPrimitiveDigi` is not fully formed.

#### PR validation:

To understand this fix, the `HcalTriggerPrimitiveAlgo` code was run before and after and these depth 7 special cases were investigated. Example printout is provided below:

Prefix
```
EVENT 7227 | (HcalTrigTower v0: 27,37) | size() 0 | presamples() 0
```
Here, since the TP is not fully formed, its `size()` and `presamples()` members are left initialized at 0.

Postfix
```
EVENT 7227 | (HcalTrigTower v0: 27,37) | size() 4 | presamples() 2
  Value=0, FG=0
  Value=0, FG=0
  Value=0, FG=0
  Value=0, FG=0
```

After the fix, when the TP is fully constructed, it has valid `size()` and `presamples()` values. Likewise, there are non-zero number of `HcalTriggerPrimitiveSamples`.

Only one additional `static const` variable is added to the code.

### DIFF
--- a/SimCalorimetry/HcalTrigPrimAlgos/interface/HcalTriggerPrimitiveAlgo.h
+++ b/SimCalorimetry/HcalTrigPrimAlgos/interface/HcalTriggerPrimitiveAlgo.h
@@ -95,7 +95,7 @@ private:
   bool validUpgradeFG(const HcalTrigTowerDetId& id, int depth) const;
   bool validChannel(const QIE10DataFrame& digi, int ts) const;
   bool needLegacyFG(const HcalTrigTowerDetId& id) const;
-  bool needUpgradeFG(const HcalTrigTowerDetId& id, int depth) const;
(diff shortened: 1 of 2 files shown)


silviodonato:
please test

silviodonato:
please test

silviodonato:
Do you have any comments @rekovic ?

rekovic:
+1

silviodonato:
+1

PR #29620, opened 2020-05-02
Add ECAL-only local reconstruction workflow

fwyzard:
#### PR description:

Implement:
  - ECAL-only RAW-to-DIGI task and sequence
  - ECAL-only (no PF) reconsruction task and sequence
  - ECAL-only (no PF) offline DQM sequences
  - ECAL-only validation sequences

Add two new runTheMatrix.py modifiers:
  - .511 for ECAL-only reconstruction
  - .512 for ECAL-only reconstruction with GPU modifier

#### PR validation:
```
$ runTheMatrix.py -j4 -l 10824.511,10824.512

...

10824.511_TTbar_13+TTbar_13TeV_TuneCUETP8M1_2018_GenSimFull+DigiFull_2018+RecoFullPatatrack_ECALOnlyCPU_2018+HARVESTFullPatatrack_ECALOnlyCPU_2018 Step0-PASSED Step1-PASSED Step2-PASSED Step3-PASSED - time date Mon May 4 13:01:40 2020-date Mon May 4 12:56:37 2020; exit: 0 0 0 0
10824.512_TTbar_13+TTbar_13TeV_TuneCUETP8M1_2018_GenSimFull+DigiFull_2018+RecoFullPatatrack_ECALOnlyGPU_2018+HARVESTFullPatatrack_ECALOnlyGPU_2018 Step0-PASSED Step1-PASSED Step2-PASSED Step3-PASSED - time date Mon May 4 13:01:40 2020-date Mon May 4 12:56:37 2020; exit: 0 0 0 0
2 2 2 2 tests passed, 0 0 0 0 failed
```

PR #29635, opened 2020-05-05
Extend the ECAL-only workflows with more PF and DQM modules

fwyzard:
### Update the default configuration of PFECALSuperClusterProducer

Update the default parameters provided by fillDescriptions() to those
used in the particleFlowSuperClusterECALMustache module.

### Split particleFlowSuperClusterECAL_cfi into a cff and multiple cfi files

Each cfi file defines a single module and its customisations:
  - particleFlowSuperClusterECALMustache
    (autogenerated by fillDescriptions)
  - particleFlowSuperClusterECALBox
  - particleFlowSuperClusterECALOnly
    (a clone of Mustache with various modifiers - see below)

The cff files includes the three cfi files.

particleFlowSuperClusteringSequence_cff is modified to use the cff file,
in order to make all modules available downstream.

### Add an ECAL-only PF supercluster module

Add an ECAL-only configuration based on the default (Mustache) configuration,
using the "online" regression instead of "offline_v2", without the
requirement of tracks and vertices, and with updated
EB/EE/ES PF cluster thresholds.

Add ECAL-only PF modules to the ECAL-only reconstruction sequence

### Rework ECAL-only DQM

Add an ECAL-only copy of ecalMonitorTask that uses the
ECAL-only PF superclusters from "particleFlowSuperClusterECALOnly" and
disables the "doExtra" plots to avoid the dependency on L1 trigger
objects.

Rework the ECAL-only DQM sequence to use this module, instead of
skipping the monitoring of the superclusters altogether.


fwyzard:
Follow up to  #29620.

fwyzard:
@cmsbuild, please test

RecoEcal/EgammaClusterProducers/python/particleFlowSuperClusterECALHLT_cfi.py
```diff
@@ -0,0 +1,77 @@
+import FWCore.ParameterSet.Config as cms
+
+particleFlowSuperClusterECALHLT = cms.EDProducer("PFECALSuperClusterProducer",
```
slava77, line 3: does this really make it suitable for HLT or is this really a reco without vtx?
For the latter, perhaps NoVtx is a better suffix than HLT.

I see that many parameters repeat particleFlowSuperClusterECALMustache. Some explicit clone-replace seems more appropriate.

fwyzard: Yes, this is the same configuration used at HLT, except for the input collections (`offlineBeamspot` vs `hltOnlineBeamspot`, `ecalRecHit` vs `hltEcalRecHit`, *etc*.), see for example https://github.com/cms-sw/cmssw/blob/acf9623a7368ca0bf1200b7411291b4dbf9c2483/HLTrigger/Configuration/python/HLT_GRun_cff.py#L13521-L13565.

> I see that many parameters repeat particleFlowSuperClusterECALMustache. Some explicit clone-replace seems more appropriate.

I tend to agree - I left it fully expanded since this is what Box and Mustache did.
If ECAL people don't have a strong preference, I'll change it to
```python
from RecoEcal.EgammaClusterProducers.particleFlowSuperClusterECALMustache_cfi import particleFlowSuperClusterECALMustache as _particleFlowSuperClusterECALMustache

# "Mustache" clustering
particleFlowSuperClusterECALHLT = _particleFlowSuperClusterECALMustache.clone()

# HLT regression setup
particleFlowSuperClusterECALHLT.regressionConfig = _particleFlowSuperClusterECALMustache.regressionConfig.clone(
    isHLT = cms.bool(True),
    eRecHitThreshold = cms.double(1.),
    regressionKeyEB  = cms.string('pfscecal_EBCorrection_online'),
    uncertaintyKeyEB = cms.string('pfscecal_EBUncertainty_online'),
    regressionKeyEE  = cms.string('pfscecal_EECorrection_online'),
    uncertaintyKeyEE = cms.string('pfscecal_EEUncertainty_online'),
    vertexCollection = cms.InputTag()
)
```

amassiro:
@rchatter

fwyzard:
@cmsbuild, please test

RecoEcal/EgammaClusterProducers/python/particleFlowSuperClusterECALHLT_cfi.py
rchatter: Hi @fwyzard,
We have no objections from the ECAL sides. The fully expanded forms looks good to us.

Thanks,
Rajdeep

fwyzard:
@cmsbuild, please test

fwyzard:
The reconstruction sequences implemented here result in these modules and dependencies:

dependency

which includes
  - hybrid clusters
  - multi5x5 clusters
  - some trigger primitive modules

My goal was something leaner, closer to

reduced

@argiro  @rchatter @slava77 @perrotta 
do you prefer that I try to streamline the ECAL-only configuration to get down to this, or keep it as is and define the leaner one as a further customisation ?

slava77:
I'm not well familiar with the graphs. Please clarify how they are made. Is this runtime or config-time? particleCluster*HLT is not in the reconstruction sequence.

I'm not sure I understand why in the first case the particleCluster*HLT has no arrow pointed to  but is still present in the graph, while in the same case RECO*Output points to the digis directly, while it goes via particleCluster*HLT in the second case.

Is the bottomline that more modules than necessary are running or that more than needed are configured, or that they claim that they depend on more things than reality?
If some modules are not used and can be dropped, I'm quite interested in the cleanup/improvement.

fwyzard:
The graphs are made adding
```python
process.DependencyGraph = cms.Service("DependencyGraph")
```
to the configuration, and rendered with
```bash
dot dependency.gv -Tpng -o dependency.png
```

The arrows indicate the dependencies declared at construction time by the `consumes` mechanism.

A module can be present without any arrows pointing into it if it is listed explicitly in a Path or EndPath (or if there is a bug in the dependency tracking of the service :-) )
The reason `particlefFowSuperClusterECALHLT` does not have an arrow pointing to it in the plot above is that the output modules do not include it.
Should they include it ?

In the full ECAL-only workflow, its output is consumed by the DQM/validation modules, but I didn;t include them here.

Back to my original question:
  - currently the ECAL-only workflow runs all the modules in the top graph
  - I would like to run only those in the bottom graph

Can I simplify the new ECAL-only workflow to run only the smaller set of modules ?
Or should I keep the workflow as it is because people think it can be useful, and define a second, leaner one for the Patatrack case ?

slava77:
> Back to my original question:
> 
>     * currently the ECAL-only workflow runs all the modules in the top graph
> 
>     * I would like to run only those in the bottom graph
> 
> 
> Can I simplify the new ECAL-only workflow to run only the smaller set of modules ?
> Or should I keep the workflow as it is because people think it can be useful, and define a second, leaner one for the Patatrack case ?

Do I understand correctly that you are proposing to move 
- hybrid clusters
- multi5x5 clusters
- some trigger primitive modules

to separate tasks which will still run in full ecalLocalRecoTask but will not be running in the leaner task which would be more useful for running of just particlefFowSuperClusterECALHLT ?

fwyzard:
> Do I understand correctly that you are proposing to move
> 
>     * hybrid clusters
> 
>     * multi5x5 clusters
> 
>     * some trigger primitive modules
> 
> 
> to separate tasks which will still run in full ecalLocalRecoTask but will not be running in the leaner task which would be more useful for running of just particlefFowSuperClusterECALHLT ?

Something like that.

I would probably define a new Task with only the modules that are relevant here, include that Task in the main ECAL reconstruction Task (to avoid duplication), and use it explicitly in the ECAL-only workflow instead of the full one.

fwyzard:
I can also look into defining 4 separate tasks for TPs, hybrid, multi5x5, and default, but it may be more complicated.

slava77:
Do I understand correctly that even now there is no dependency issue at run time (after module construction, during event processing)?
The framework was already not running modules which are not consumed. More recently these modules were also deleted and pruned from the dependency graphs so that they do not affect execution.
@makortel please clarify if I got the recent status right.

rchatter:
> The reconstruction sequences implemented here result in these modules and dependencies:
> 
> dependency
> 
> which includes
> 
>     * hybrid clusters
> 
>     * multi5x5 clusters
> 
>     * some trigger primitive modules
> 
> 
> My goal was something leaner, closer to
> 
> reduced
> 
> @argiro @rchatter @slava77 @perrotta
> do you prefer that I try to streamline the ECAL-only configuration to get down to this, or keep it as is and define the leaner one as a further customisation ?

Hi @fwyzard ,
The graph looks really nice and thanks for the instructions on how to produce it. 

In ECAL the most commonly used CMSSW sequence is what we use during calibration/re-calibration with new or updated conditions. Please refer to the attached slides for the key steps in this process. The full set of steps used for calibration use the CMSSW modules listed on slide 3. This corresponds to your top more detailed  graph. The distinct steps involved here are listed on slide 1 and the modules for these steps on slide 2.  Your second more simplified graph includes the steps ECAL local reco. + "basic" clustering. Given that there is a frequent use case for the more detailed work flow I feel its best to keep the more detailed one as the default and define the leaner one as a further customization and if possible a further leaner one with only the ECAL local reco. modules.

thanks,
Rajdeep
[ECAL_Calibration_workflow.pdf](https://github.com/cms-sw/cmssw/files/4588369/ECAL_Calibration_workflow.pdf)

makortel:
> The framework was already not running modules which are not consumed. More recently these modules were also deleted and pruned from the dependency graphs so that they do not affect execution.
> @makortel please clarify if I got the recent status right.

The PR for deleting non-consumed modules (#29553) is still open (and I'm not sure I want to push it to 11_1). The statement of framework not running Event transitions for unscheduled producers whose products are not consumed by anything else that runs is correct.

fwyzard:
@slava77 what do you mean ?
> Do I understand correctly that even now there is no dependency issue at run time (after module construction, during event processing)?
The framework was already not running modules which are not consumed. More recently these modules were also deleted and pruned from the dependency graphs so that they do not affect execution.

If you refer to the hybrid and multi5x5 modules, they are run in the standard reconstruction sequence; here is an excerpt from the TrigReport of the `10824.0` workflow (the full one is attached, [TrigReport.txt](https://github.com/cms-sw/cmssw/files/4589792/TrigReport.txt)):
```
TrigReport ---------- Module Summary ------------
TrigReport    Visited   Executed     Passed     Failed      Error Name
...
TrigReport         10         10         10          0          0 multi5x5BasicClustersCleaned
TrigReport         10         10         10          0          0 multi5x5BasicClustersUncleaned
TrigReport         20         10         10          0          0 multi5x5PreshowerClusterShape
TrigReport         40         10         10          0          0 multi5x5SuperClusters
TrigReport         10         10         10          0          0 multi5x5SuperClustersCleaned
TrigReport         10         10         10          0          0 multi5x5SuperClustersUncleaned
TrigReport         30         10         10          0          0 multi5x5SuperClustersWithPreshower
...
```

fwyzard:
> In ECAL the most commonly used CMSSW sequence is what we use during calibration/re-calibration with new or updated conditions. Please refer to the attached slides for the key steps in this process. The full set of steps used for calibration use the CMSSW modules listed on slide 3. This corresponds to your top more detailed graph. The distinct steps involved here are listed on slide 1 and the modules for these steps on slide 2. Your second more simplified graph includes the steps ECAL local reco. + "basic" clustering. Given that there is a frequent use case for the more detailed work flow I feel its best to keep the more detailed one as the default and define the leaner one as a further customization and if possible a further leaner one with only the ECAL local reco. modules.

hi @rchatter,
thanks for the details, probably too many for me here :-)

OK - I will keep the ECAL-only workflow as defined by this PR, and produce a slimmed one for RECO-only Patatrack step.

Only two caveats:
  - I will use the `.511`, `.512`, *etc*. modifiers for the Patatrack version, since I've already allocated that block for the Pixel, ECAL, HCAL, *etc*. workflows on CPU and GPU - we'll need to figure out a different modifier for the "ECAL-only for calibration" workflow;
  - the "ECAL-only" workflow uses the "no vertex" or "HLT" variant of the `particleFlowSuperClusterECAL` module (to avoid building all tracks and vertices); I expect you do want to use the default version for the calibration workflow ?

.A

slava77:
perhaps I misunderstood the setup and reasoning for streamlining/reorganization.

I thought that the concern was that ```RECO:reconstruction_ecalOnly``` has extras like ```multi5x5*``` constructed which makes them appear in the dependency graph.
My response was that, while they are not consumed they will not be running, hence the reorganization targets only the construction time issues, which do not affect event processing.

However,
If ```multi5x5*``` are running with ```RECO:reconstruction_ecalOnly```, then there is a problem to be fixed.
If they are running in the default ```RECO``` and the claim is that they are actually not consumed by anything, then there is a problem to be fixed.

fwyzard:
> If `multi5x5*` are running with `RECO:reconstruction_ecalOnly`, then there is a problem to be fixed.

Apparently ECAL prefers it that way ...

> If they are running in the default `RECO` and the claim is that they are actually not consumed by anything, then there is a problem to be fixed.

They are running in the default `RECO` because they *are* consumed, at the very least by the output module.
I don't know if they are consumed by anything else - but I'd say that's beyond the scope of this PR ?

I will update this PR so that
  - there is a sequence that runs only the "default" part of the ECAL reconstruction, and only that
  - the `.511` and `.512` workflows use that sequence

In the Patatrack branch I will
  - implement the use of the `gpu` modifier for the available ECAL modules
  - make the `.512` workflow use the `gpu` modifier to run on a GPU

later
  - add a `.513` workflow that uses the `SwitchProducer` to choose between CPU and GPU at runtime
  - add a `.514` workflow that runs both branches of the `SwitchProducer`

even later
  - add to the validation of the `.514` workflow the module that compares the ECAL bjects between GPU and CPU (at least as much as it is available)

rchatter:
> > In ECAL the most commonly used CMSSW sequence is what we use during calibration/re-calibration with new or updated conditions. Please refer to the attached slides for the key steps in this process. The full set of steps used for calibration use the CMSSW modules listed on slide 3. This corresponds to your top more detailed graph. The distinct steps involved here are listed on slide 1 and the modules for these steps on slide 2. Your second more simplified graph includes the steps ECAL local reco. + "basic" clustering. Given that there is a frequent use case for the more detailed work flow I feel its best to keep the more detailed one as the default and define the leaner one as a further customization and if possible a further leaner one with only the ECAL local reco. modules.
> 
> hi @rchatter,
> thanks for the details, probably too many for me here :-)
> 
> OK - I will keep the ECAL-only workflow as defined by this PR, and produce a slimmed one for RECO-only Patatrack step.
> 
> Only two caveats:
> 
>     * I will use the `.511`, `.512`, _etc_. modifiers for the Patatrack version, since I've already allocated that block for the Pixel, ECAL, HCAL, _etc_. workflows on CPU and GPU - we'll need to figure out a different modifier for the "ECAL-only for calibration" workflow;
 
Sounds good.

>     * the "ECAL-only" workflow uses the "no vertex" or "HLT" variant of the `particleFlowSuperClusterECAL` module (to avoid building all tracks and vertices); I expect you do want to use the default version for the calibration workflow ?

That is correct. Just to ensure we have no misunderstanding I am pasting the settings we use for the said module:
##########################
>>> process.particleFlowSuperClusterECAL
cms.EDProducer("PFECALSuperClusterProducer",
    BeamSpot = cms.InputTag("offlineBeamSpot"),
    ClusteringType = cms.string('Mustache'),
    ESAssociation = cms.InputTag("particleFlowClusterECAL"),
    EnergyWeight = cms.string('Raw'),
    PFBasicClusterCollectionBarrel = cms.string('particleFlowBasicClusterECALBarrel'),
    PFBasicClusterCollectionEndcap = cms.string('particleFlowBasicClusterECALEndcap'),
    PFBasicClusterCollectionPreshower = cms.string('particleFlowBasicClusterECALPreshower'),
    PFClusters = cms.InputTag("particleFlowClusterECAL"),
    PFSuperClusterCollectionBarrel = cms.string('particleFlowSuperClusterECALBarrel'),
    PFSuperClusterCollectionEndcap = cms.string('particleFlowSuperClusterECALEndcap'),
    PFSuperClusterCollectionEndcapWithPreshower = cms.string('particleFlowSuperClusterECALEndcapWithPreshower'),
    applyCrackCorrections = cms.bool(False),
    doSatelliteClusterMerge = cms.bool(False),
    dropUnseedable = cms.bool(False),
    etawidth_SuperClusterBarrel = cms.double(0.04),
    etawidth_SuperClusterEndcap = cms.double(0.04),
    phiwidth_SuperClusterBarrel = cms.double(0.6),
    phiwidth_SuperClusterEndcap = cms.double(0.6),
    regressionConfig = cms.PSet(
        applySigmaIetaIphiBug = cms.bool(False),
        ecalRecHitsEB = cms.InputTag("ecalRecHit","EcalRecHitsEB"),
        ecalRecHitsEE = cms.InputTag("ecalRecHit","EcalRecHitsEE"),
        isHLT = cms.bool(False),
        regressionKeyEB = cms.string('pfscecal_EBCorrection_offline_v2'),
        regressionKeyEE = cms.string('pfscecal_EECorrection_offline_v2'),
        uncertaintyKeyEB = cms.string('pfscecal_EBUncertainty_offline_v2'),
        uncertaintyKeyEE = cms.string('pfscecal_EEUncertainty_offline_v2'),
        vertexCollection = cms.InputTag("offlinePrimaryVertices")
    ),
    satelliteClusterSeedThreshold = cms.double(50.0),
    satelliteMajorityFraction = cms.double(0.5),
    seedThresholdIsET = cms.bool(True),
    thresh_PFClusterBarrel = cms.double(0.0),
    thresh_PFClusterES = cms.double(0.0),
    thresh_PFClusterEndcap = cms.double(0.0),
    thresh_PFClusterSeedBarrel = cms.double(1.0),
    thresh_PFClusterSeedEndcap = cms.double(1.0),
    thresh_SCEt = cms.double(4.0),
    useDynamicDPhiWindow = cms.bool(True),
    useRegression = cms.bool(True),
    verbose = cms.untracked.bool(False)
)
###############################################

rchatter:
Hi @slava77 ,

> However,
> If `multi5x5*` are running with `RECO:reconstruction_ecalOnly`, then there is a problem to be fixed.
> If they are running in the default `RECO` and the claim is that they are actually not consumed by anything, then there is a problem to be fixed.

I couldn't quite follow the precise problem you are referring to here.

Rajdeep

perrotta:
> Hi @slava77 ,
> 
> > However,
> > If `multi5x5*` are running with `RECO:reconstruction_ecalOnly`, then there is a problem to be fixed.
> > If they are running in the default `RECO` and the claim is that they are actually not consumed by anything, then there is a problem to be fixed.
> 
> I couldn't quite follow the precise problem you are referring to here.
> 
> Rajdeep

The problem would have been a module that was run even if nobody consumed it. If at least the output module consumes it, then there is no problem. As @fwyzard says, it is probably beyond the scope of this PR to review the event content and verify whether whatever is saved there is actually needed.

fwyzard:
Here is the graph corresponding to the latest ECAL-only workflow, e.g. `10824.511`:
ecalonly

The modules from the raw2digi and reco steps are highlighted in blue.

FYI the graph was generated with
```
from FWCore.ParameterSet.Utilities import moduleLabelsInSequences
process.DependencyGraph = cms.Service("DependencyGraph",
    highlightModules = cms.untracked.vstring(moduleLabelsInSequences(process.raw2digi_step, process.reconstruction_step))
)
```

fwyzard:
However I haven't been able to find a simply way to run the full ECAL reconstruction, including the original `particleFlowSuperClusterECAL` module that requires full tracks and vertices - at least not without duplicating a lot of the reconstruction configuration.

fwyzard:
@cmsbuild, please test

fwyzard:
@perrotta, @slava77, do the latest version look good to you ?

fwyzard:
@andrius-k, @kmaeshima, @schneiml, @jfernan2, @fioriNTU, any obections from the DQM side ?

fwyzard:
@civanch, @mdhildreth could you sing for Simulation (for `Validation/Configuration`) ?

jfernan2:
+dqm

slava77 (commented):
I suppose the comments about the dead parameters do not need to be addressed in this PR, but could be a good reference for a follow up 
@rchatter please take a note and make a plan to switch the ECAL clustering configs to start from the auto-generated configs from the fillDescriptions in PFECALSuperClusterProducer.cc
and cleanup dead configs at that point as well

RecoEcal/EgammaClusterProducers/python/particleFlowSuperClusterECALMustache_cfi.py
```diff
@@ -0,0 +1,76 @@
...
+
+    PFBasicClusterCollectionBarrel = cms.string("particleFlowBasicClusterECALBarrel"),
+    PFSuperClusterCollectionBarrel = cms.string("particleFlowSuperClusterECALBarrel"),
+    PFBasicClusterCollectionEndcap = cms.string("particleFlowBasicClusterECALEndcap"),
+    PFSuperClusterCollectionEndcap = cms.string("particleFlowSuperClusterECALEndcap"),
+    PFBasicClusterCollectionPreshower = cms.string("particleFlowBasicClusterECALPreshower"),
+    PFSuperClusterCollectionEndcapWithPreshower = cms.string("particleFlowSuperClusterECALEndcapWithPreshower"),
+
+    # use preshower ?
+    use_preshower = cms.bool(True),
```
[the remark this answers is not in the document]
fwyzard: @slava77, do you mean that it was present explicitly in `particleFlowSuperClusterECALBox` and only implicitly in `particleFlowSuperClusterECALMustache` (via `fillDescriptions()`) but not explicitly ?

RecoEcal/EgammaClusterProducers/python/particleFlowSuperClusterECALBox_cfi.py
```diff
@@ -0,0 +1,76 @@
...
+
+    # window width in ECAL
+    phiwidth_SuperClusterBarrel = cms.double(0.28),
+    etawidth_SuperClusterBarrel = cms.double(0.04),
+
+    phiwidth_SuperClusterEndcap = cms.double(0.28),
+    etawidth_SuperClusterEndcap = cms.double(0.04),
+
+    # threshold in preshower
+    thresh_PFClusterES = cms.double(0.),
```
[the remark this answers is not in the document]
fwyzard: @slava77, do you mean that it was present only implicitly in `particleFlowSuperClusterECALBox` (via `fillDescriptions()`) but not explicitly ?

RecoEcal/EgammaClusterProducers/python/particleFlowSuperClusterECALMustache_cfi.py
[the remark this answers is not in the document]
slava77: right, by "was not present" I meant the content in particleFlowSuperClusterECAL_cfi before this PR.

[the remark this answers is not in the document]
fwyzard: OK, I assume that if I have to fix that, you'd rather I import the autogenerated cfi and only set the parameters that are different from the default ?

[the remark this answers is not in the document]
slava77: no need to do anything about these in this PR, as noted in the review "comments about the dead parameters do not need to be addressed in this PR, but could be a good reference for a follow up"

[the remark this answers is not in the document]
rchatter: Hi @slava77,
 "please take a note and make a plan to switch the ECAL clustering configs to start from the auto-generated configs from the fillDescriptions in PFECALSuperClusterProducer.cc
and cleanup dead configs at that point as well"

I take note and will follow up.

Thanks,
Rajdeep

fwyzard:
With the latest commit:
  - move the `ecalDigis` and `ecalMultiFitUncalibRecHit` modules to separate Tasks, so that each of them can be overridden with a group of modules when running on GPU
  - implement an `ecalOnlyLocalRecoTask` based on the `ecalLocalRecoTask`, without the trigger primitive-related modules

fwyzard:
A simplified `RecoEcal/EgammaClusterProducers/python/particleFlowSuperClusterECALMustache_cfi.py` would look something like
```python

from RecoEcal.EgammaClusterProducers.particleFlowSuperClusterECALMustache_cfi import particleFlowSuperClusterECALMustache as _particleFlowSuperClusterECALMustache

# update the regression configuration to the "offline_v2" tags
_particleFlowSuperClusterECALMustache_regressionConfig = _particleFlowSuperClusterECALMustache.regressionConfig.clone(
  regressionKeyEB  = 'pfscecal_EBCorrection_offline_v2',
  regressionKeyEE  = 'pfscecal_EECorrection_offline_v2',
  uncertaintyKeyEB = 'pfscecal_EBUncertainty_offline_v2',
  uncertaintyKeyEE = 'pfscecal_EEUncertainty_offline_v2',
)

particleFlowSuperClusterECALMustache = _particleFlowSuperClusterECALMustache.clone(
  regressionConfig = _particleFlowSuperClusterECALMustache_regressionConfig,
)
```

Shall I update it and the other particleFlowSuperClusterECAL cfi files in this way ?

fwyzard:
@cmsbuild, please test

slava77:
> A simplified `RecoEcal/EgammaClusterProducers/python/particleFlowSuperClusterECALMustache_cfi.py` would look something like
> 
> ```python
> from RecoEcal.EgammaClusterProducers.particleFlowSuperClusterECALMustache_cfi 
> ```
> 
> Shall I update it and the other particleFlowSuperClusterECAL cfi files in this way ?

I'm confused by the identical file names in the name and the imported file.
I expect that a moduleName_cfi made by fillDescriptions will not match in name with another hand-edited moduleName_cfi . Some disambiguation will be needed (e.g fillDescriptions making moduleNameDefault_cfi instead, if it can not be used directly). 
But perhaps there was a typo.

RecoLocalCalo/Configuration/python/ecalLocalRecoSequence_cff.py
```diff
@@ -23,20 +23,29 @@
...
 
-ecalRecHitTask        = cms.Task(ecalRecHit,
-                                         ecalCompactTrigPrim,
-                                         ecalTPSkim,
-                                         ecalPreshowerRecHit)
+ecalRecHitTask = cms.Task(
+    ecalRecHit,
+    ecalCompactTrigPrim,
+    ecalTPSkim,
+    ecalPreshowerRecHit)
```
slava77, line 34: wouldn't it be more straightforward to define ```ecalRecHitNoTPTask```  with ```ecalRecHit, ecalPreshowerRecHit``` so that it can be directly used in both ecalRecHitTask and ecalOnlyLocalRecoTask?

RecoEcal/EgammaClusterProducers/python/particleFlowSuperClusterECALOnly_cfi.py
```diff
@@ -0,0 +1,22 @@
...
+    uncertaintyKeyEB = cms.string('pfscecal_EBUncertainty_online'),
+    regressionKeyEE  = cms.string('pfscecal_EECorrection_online'),
+    uncertaintyKeyEE = cms.string('pfscecal_EEUncertainty_online'),
+    vertexCollection = cms.InputTag('')
+)
+
+# ECAL-only (no primary vertices) thresholds
+particleFlowSuperClusterECALOnly.thresh_PFClusterBarrel = cms.double(0.5)
+particleFlowSuperClusterECALOnly.thresh_PFClusterEndcap = cms.double(0.5)
+particleFlowSuperClusterECALOnly.thresh_PFClusterES = cms.double(0.5)
```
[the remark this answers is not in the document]
slava77: NB: this is still not addressed.
I suspect this is under discussion in the main thread  in https://github.com/cms-sw/cmssw/pull/29635#issuecomment-626035943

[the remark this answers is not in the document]
fwyzard: I can move the changes inside `clone()`.

Is there any actual difference between doing
```python
a = b.clone()
a.parameter = 'value'
```
and
```python
a = b.clone(
  parameter = 'value'
)
```
?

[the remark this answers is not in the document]
fwyzard: > please drop type specifications for all parameters which already exist.

is there a way to do that for `InputTag`s ?
how would I do the equivalent of `ecalRecHitsEB = cms.InputTag('ecalRecHit','EcalRecHitsEB')` ?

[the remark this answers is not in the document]
makortel: > how would I do the equivalent of `ecalRecHitsEB = cms.InputTag('ecalRecHit','EcalRecHitsEB')` ?

`ecalRecHitsEB = 'ecalRecHit:EcalRecHitsEB'`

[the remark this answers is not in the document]
makortel: > Is there any actual difference between doing

There is no functional difference, but overrides inside `clone()` need less typing, the name `a` needs to be specified only once, and the interface is similar `toModify()`.

fwyzard:
> I'm confused by the identical file names in the name and the imported file.

Right, I missed that.
Then I guess the easiest solution would be to update the content of the autogenerated cfi file so that it can be used directly for the Mustache case: the only difference seems to be the regression tags

RecoEcal/EgammaClusterProducers/python/particleFlowSuperClusterECALBox_cfi.py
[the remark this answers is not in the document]
fwyzard: Since `particleFlowSuperClusterECALBox` was almost fully defined anyway, I prefer to list here all parameters and their values.

I agree it can be simplified importing particleFlowSuperClusterECALMustache_cfi and setting only the different parameters, but I would leave it to a subsequent PR.

fwyzard:
@cmsbuild, please test

fwyzard:
@cmsbuild, please test workflow 10824.511

fwyzard:
I think I addressed all comments relevant for this PR ?

jfernan2:
+1

RecoEcal/EgammaClusterProducers/python/particleFlowSuperClusterECALOnly_cfi.py
```diff
@@ -0,0 +1,21 @@
...
+        regressionKeyEB  = cms.string('pfscecal_EBCorrection_online'),
+        uncertaintyKeyEB = cms.string('pfscecal_EBUncertainty_online'),
+        regressionKeyEE  = cms.string('pfscecal_EECorrection_online'),
+        uncertaintyKeyEE = cms.string('pfscecal_EEUncertainty_online'),
+        vertexCollection = cms.InputTag(''),
+    ),
+    # ECAL-only (no primary vertices) thresholds
+    thresh_PFClusterBarrel = cms.double(0.5),
+    thresh_PFClusterEndcap = cms.double(0.5),
+    thresh_PFClusterES = cms.double(0.5),
```
slava77, line 20: please drop type specifications for all parameters which already exist.

in the above it will be ```thresh_PFClusterES = 0.5``` and so on

RecoEcal/EgammaClusterProducers/python/particleFlowSuperClusterECALOnly_cfi.py
```diff
@@ -0,0 +1,21 @@
+import FWCore.ParameterSet.Config as cms
+
+from RecoEcal.EgammaClusterProducers.particleFlowSuperClusterECALMustache_cfi import particleFlowSuperClusterECALMustache as _particleFlowSuperClusterECALMustache
+
+# "Mustache" clustering
+particleFlowSuperClusterECALOnly = _particleFlowSuperClusterECALMustache.clone(
+    # ECAL-only (no primary vertices) regression setup
+    regressionConfig = _particleFlowSuperClusterECALMustache.regressionConfig.clone(
```
slava77, line 8: a more common/cleaner syntax here is like the following
```python
regressionConfig = dict(
    isHLT = True
...
```

fwyzard: would the `dict` get merged with the existing parameters, or overwrite the full `PSet` ?

makortel: `dict` overrides only the parameters that are given as arguments (similar to `clone()` itself).

RecoEcal/EgammaClusterProducers/python/particleFlowSuperClusterECALOnly_cfi.py
fwyzard: Right, sorry, I forgot to fix that.

I don't know if I can make further changes to this PR as I'm now using the same branch in the Patatrack developments.
If you consider this a strict requirement, I'll see what I can do.

RecoEcal/EgammaClusterProducers/python/particleFlowSuperClusterECALOnly_cfi.py
fwyzard: Thanks.

Completely unrelated: could I suggest allowing to build an `InputTag` from a `tuple`, i.e. on top of
```python
parameter = cms.InputTag("label", "instance")
parameter = "label:instance"
```
also
```python
parameter = ("label", "instance")
```
?

makortel: > Completely unrelated: could I suggest allowing to build an `InputTag` from a `tuple`

Could you open an issue about it?

RecoEcal/EgammaClusterProducers/python/particleFlowSuperClusterECALOnly_cfi.py
slava77: I'm not sure about all the implications wrt using this branch in Patatrack developments.
I can suggest a solution to open a replacement/followup PR based on this PR changes + the pending changes from this review on a branch different from this PR topic branch.

fwyzard: Honestly, we always come to the point where I believe the expectations you have from other people are not in line with the amount of work they are willing to do.

I believe I have done more changes and cleaning up in this PR as I was supposed to do, and the rest should be handled by the ECAL DPG reco contacts and/or by the RECO conveners.

So what I am going to do is to make this last fix, and expect the PR to be signed by RECO immediately afterwards.

fwyzard:
@cmsbuild, please test workflow 10824.511

fwyzard:
@slava77 please sign

fwyzard:
@jfernan2 sorry for all the updates following the requests from the reco conveners :-(
could you sign again ?
I promise this is the last time :-)

fwyzard:
On Mon, 11 May 2020 at 19:12, Slava Krutelyov <notifications@github.com>
wrote:

> I'm not sure about all the implications wrt using this branch in Patatrack
> developments.
>

It means that I am developing, integrating and validating various workflows
at the same time - which in turn requires merging multiple independent PR
from multiple repositories for each test.
Making changes to one PR require rebuilding the whole merge history, which
in turns require rebuilding an almost full release, and re-running all
validation checks.

Simply because you don't like the way in which some parameters are defined
- something that has no impact whatsoever on the efficiency, performance,
readability, or maintainability of the code.

I can suggest a solution to open a replacement/followup PR based on this PR
> changes + the pending changes from this review on a branch different from
> this PR topic branch.
>

That does not help with getting *this* PR merged.

.A

jfernan2:
+1

slava77:
+1

for https://github.com/cms-sw/cmssw/pull/29635	6abcdc2
- jenkins tests pass and comparisons with the baseline show no relevant differences; the ecalOnly wf 10824.511 runs

civanch:
+1

silviodonato:
+operations

silviodonato:
+1

PR #29632, opened 2020-05-04
Run3-gem36 Use MuonGeometryConstants rather than MuonDDDConstants

bsunanda:
#### PR description:

Use the object MuonGeometryConstants in all the downstream codes rather than MuonDDDConstants

#### PR validation:

Use runTheMatrix.py to validate the tests

#### if this PR is a backport please specify the original PR and why you need to backport that PR:

Nothing special

PR #29647, opened 2020-05-05
MTD geometry: porting of MTDNumberingBuilder to DD4hep

fabiocos:
#### PR description:

This PR builds on top of #29599 and previous developments a porting of ```Geometry/MTDNumberingBuilder``` to DD4hep. In order to accomplish this task, additional ```SpecPars``` have been added in ```mtdStructureTopology``` to specify the list of volumes to be considered in the loop to build subdetectors, layers and modules. The names of BTL Layer1 and ETL discs have been extended with the addition of the ```Timing``` suffix to distinguish them from the tracker subvolumes, and resolve an ambiguity that DDCMS at present seems unable to handle (as reported to @ianna). The DDD code has been updated accordingly.

Non negligible memory issues have been noticed with the ```navPos()``` method of DDCMS in this porting, and its use has been dropped from everywhere, and it is not used to fill ```GeometricTimingDet```, as it looks like this information is not anyway used afterwards (this was carbon-coped from tracker). They can be seen just adding the print of ```fv->navPos().size() ``` in the new test added in ```DDCMS``` for the ```cms::DDFilteredView```.  

The porting to DD4hep tries to:

- use the FilteredView philosophy based on SpecPar selection, avoiding a full loop  based on ```fv.next(0)```;

- keep the loop on the entire geometry inside a single method, avoiding nested calls to different levels as done in the Tracker geometry;

- circumvent the nested filling of ```GeometricTimingDet``` containers by using intermediate temporary ```subdet``` and ```layer``` vectors. For the time being the bare pointer original structure of the containers has been kept to minimize simultaneous changes and disruption.

#### PR validation:

The test configuration in ```Geometry/MTDNumberingBuilder``` runs on both scenario D50 and D53, providing identical lists of detIds. The DDD version is unchanged compared to the base reference. Differences in rotation matrices are due to the sign of terms that are null up to the 6th decimal digit. The calculation of ```Phi``` looks different for the top level volumes, but I believe this might be an artefact of translation vectors along the z direction. No difference is observed for modules. 

The ```Geometry/MTDGeometryBuilder``` tests have been updated, they give the same result as before for DDD, there is a factor of 10 to be fixed in next developments for DD4hep.


fabiocos:
@smuzaffar hard to notice any visible difference in the proposed patch...

fabiocos:
@silviodonato @kpedro88 as you notice this PR embeds #29599. If that can be merged as it is (it should IMO) I will rebase this one, otherwise better to close that and continue here in case.

fabiocos:
please test

Geometry/MTDNumberingBuilder/plugins/DDDCmsMTDConstruction.cc
```diff
@@ -158,13 +162,36 @@ std::unique_ptr<GeometricTimingDet> DDDCmsMTDConstruction::construct(const DDCom
...
   }
 
+  // Add layers to subdetectors:
+  // first BTL (one layer only)
+
+  subdet[0]->addComponent(layer[0]);
+
+  // then ETL (number of layers depend on preTDR design or not)
+
+  if (layer.size() == 3) {
```
kpedro88, line 179: magic numbers should be named constants

Geometry/MTDNumberingBuilder/plugins/DDDCmsMTDConstruction.cc
```diff
@@ -177,3 +204,142 @@ std::unique_ptr<GeometricTimingDet> DDDCmsMTDConstruction::construct(const DDCom
...
+  }
+
+  // Add layers to subdetectors:
+  // first BTL (one layer only)
+
+  subdet[0]->addComponent(layer[0]);
+
+  // then ETL (number of layers depend on preTDR design or not)
+
+  if (layer.size() == 3) {
```
kpedro88, line 318: magic numbers also used here
maybe indicates the need for a geometry version tool/function?

Geometry/MTDNumberingBuilder/src/CmsMTDStringToEnum.cc
```diff
@@ -10,6 +10,8 @@ CmsMTDStringToEnum::Impl::Impl() {
                                                                                   GeometricTimingDet::BTL));
   _map.insert(
       std::pair<std::string, GeometricTimingDet::GeometricTimingEnumType>("Layer1", GeometricTimingDet::BTLLayer));
+  _map.insert(std::pair<std::string, GeometricTimingDet::GeometricTimingEnumType>("Layer1Timing",
```
kpedro88, line 13: preferable to use emplace, but beyond scope of this PR

fabiocos:
hold

to be rebased, and I will add an update of test formatting, with better rounding to compare seamlessly

fabiocos:
rebased onto CMSSW_11_1_X_2020-05-06-2300

fabiocos:
With the update of rounding in test output (@icosivi this will affect also your tests) the only residual differences in geometry dumps are related to:

- list of siblings (now dummy for DD4hep, as far as I can see this is used nowhere);

- name of ETL module in D50, as the old (DDD) choice is the wafer, while for DD4hep I want to fully use the ETLNumberingScheme and I use Timingactive. Due to the definition of these tow volumes (identical, the latter child of the former) this implies no change at all in the built GeometricTimingDets apart for the name. I could instruct the code to hide this in case, although I doubt that this has any practical consequence, apart for visual comparison of geometric dumps.

fabiocos:
please test

fabiocos:
As far as I can see, the DQM differences have nothing to do with this PR.

fabiocos:
unhold

kpedro88:
+upgrade

cvuosalo:
@fabiocos There are several files that have "DDD" in their names that you have added DD4hep functionality in this PR. Can you remove the "DDD" from the file names and make the names more generic? "DDD" indicates the old DD, so it is confusing if "DDD" files contain migrated code.

fabiocos:
@cvuosalo as far as I can see, the only two cases are ```DDDCmsMTDConstruction``` and ```ExtractStringFromDDD``` in ```Geometry/MTDNumberingBuilder```. There were the original names, cloned from ```Geometry/TrackerNumberingBuilder``` where they have also not been renamed after the migration. Please notice that the latter is exactly a clone, made just to avoid an annoying dependence of the MTD geometry on the Tracker one. 

If you prefer, I can easily turn the DDD string into DD, I doubt this makes any effect in the long term but ok

fabiocos:
please test

fabiocos:
This latest update completes the migration, adapting the main geometry configurations to the use of ```_cff``` fragments instead of ```_cfi``` for ```mtdNumberingGeometry``` and ```mtdParameters```. This is irrelevant for DDD, but allows the needed customization for DD4hep outside of the test configurations. I have adapted also those for DDD just to avoid confusion.

This should be also propagated in tests in RecoMTD/DetLayers (@parbol I'll do it when working on that) and Validation/MtdValidation (@casarsa @gsorrentino18 so far you use D49 as default, my changes are relevant for D50 and D53 only).

I see that this PR overlaps with #29632 by @bsunanda , although there should be no conflict for what I can see. As that is bigger and taking more time, I'm in favour of pushing that forward, and then mine (to be rebased if needed, but I do not see why it should).

fabiocos:
please test

cvuosalo:
@fabiocos Thanks. Some day next year (or later) I will delete all the old DD code. It would be nice to just delete all "DDD" files. I know it will not be that easy, but trying to mark clearly old DD, DD4hep, and shared files will help me later.

kpedro88:
+upgrade

bsunanda:
@fabiocos @silviodonato Could you delete the thermal screen from V4 version of ETL geometry

fabiocos:
@bsunanda this is done in a separate PR, please check #29800

fabiocos:
@silviodonato ok, I see that after the integration of #29632 there are conflicts to be fixed, let me check and restart again tests

silviodonato:
> @silviodonato ok, I see that after the integration of #29632 there are conflicts to be fixed, let me check and restart again tests

Thanks!

fabiocos:
@silviodonato the fix was quite trivial, @bsunanda just added another fragment into D50 and D53 configurations, but the position was close to my change and they entered into conflict. The fix should be completely transparent.

fabiocos:
Unfortunately this PR will miss next IB, I hope we may converge by next one this evening

fabiocos:
please test

fabiocos:
The reported failure in CMSSW code rules is unrelated to this PR, DQM differences are unrelated as well

kpedro88:
+upgrade

silviodonato:
merge

cvuosalo:
+1

PR #8872, opened 2015-04-24
Update to simplified ecal pfcluster corrections on top of new Global Tags for PF corrections plus fixes for Run2 MC production

diguida:
Move to updated version of ecal pfcluster corrections with reduced number of input variables, plus some additional pt binning and slightly different parameterization to improve the behaviour at very low energies. Originally in #8818 

New Global Tags including updated set of PF corrections for E-gamma and hadrons, and a couple of fixes for Run2 MC production. Originally in #8869

PR #8818, opened 2015-04-22
update to simplified ecal pfcluster corrections

bendavid:
Move to updated version of ecal pfcluster corrections with reduced number of input variables, plus some additional pt binning and slightly different parameterization to improve the behaviour at very low energies.

Having some problems reading the conditions from the db at the moment, so the branch currently is pointing to an sqlite file on my public afs area (clearly the minimum to merge the pull request would be to have the esprefer working from FrontierProd, if not updated global tags)

PR #7558, opened 2015-02-04
Ecal PFCluster corrections (74x)

bendavid:
Adapt PFCluster energy correction code to use new regression weights in the conditions database (current implementation is using ESPrefer until this is available in a GT)

The PFAlgo also needed to be modified since it was still using hardcoded inlined corrections (taking the old corrections) instead of taking the corrected energy filled by the more modular correction scheme which @lgray added a while back.

The DataFormat change to CaloCluster adds just an additional float to store the uncertainty associated with the corrected energy (which is a new feature of the regression-based corrections which is not currently used anywhere, but intended to be used as input to the higher level electron and photon corrections)

Two caveats:
1) The dataformat change introduces some potential issues with checksum and class version divergences and incompatibility once we try to backport this to 73x.  See https://hypernews.cern.ch/HyperNews/CMS/get/edmFramework/3433/1.html

2) I am still running some internal closure tests against the standalone validation code for the regression training (to make sure there are no trivial mistakes like variables being defined in a way which doesn't match the training, or passed in the wrong order to the BDT's in the CMSSW class).  Assuming that these checks will be positive, physics checks of this pull request can already start in principle.  (but would maybe hold off merging it until the outcome of those tests is confirmed)

PR #29658, opened 2020-05-05
small cleanup of CorrectedECALPFClusterProducer::fillDescriptions

missirol:
#### PR description:

It seems that a couple of parameters specified in the `fillDescriptions` method of `CorrectedECALPFClusterProducer` are not really being used by the producer.

The two parameters are `algoName` and `verticesLabel`, and they are currently part of a PSet used to configure an instance of `PFClusterEMEnergyCorrector`.

As far as I can see, the primary vertices were used in previous versions of `PFClusterEMEnergyCorrector` (they were introduced in #7558), but then were later removed when the BDT training was updated (in #8872, which was a follow-up of #8818).

For the `algoName` parameter, in my understanding this was needed when `PFClusterEMEnergyCorrector` was inheriting from `PFClusterEnergyCorrectorBase`, but that is not the case anymore (since #7558).

The only change in `Reconstruction_hiPF_cff` is the removal of one line, which was modifying the value of one of these (unused) parameters.

No changes are expected.

#### PR validation:

Passed tests with `runTheMatrix.py -l limited -i all`.

Not a backport.

### DIFF
--- a/HLTrigger/Configuration/python/customizeHLTforCMSSW.py
+++ b/HLTrigger/Configuration/python/customizeHLTforCMSSW.py
@@ -183,11 +183,25 @@ def customiseFor29512(process):
 
     return process
 
+def customiseFor29XXX(process):
+    """small cleanup of CorrectedECALPFClusterProducer::fillDescriptions
+    Removes two unused parameters of CorrectedECALPFClusterProducer
+    PR: https://github.com/cms-sw/cmssw/pull/29XXX
+    """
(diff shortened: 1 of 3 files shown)


slava77:
@cmsbuild please test

Martin-Grunewald:
+1

perrotta:
please test
(in order to clean up the comparison)

perrotta:
+1
- Technical: indeed the two parameters seems not to be used anywhere

silviodonato:
+1

PR #29741, opened 2020-05-05
Fix crashing bug  in reco when run with XML geometry

cvuosalo:
When reco for the Extended2021 scenario was run with geometry from XML instead of the DB, it was crashing due to missing GEM super chambers. The validation code was referencing GEM super chambers even if they didn't exist, thus causing the crash.

This bug fix adds protection in the validation code to make sure that null or empty objects are not dereferenced.

The issue of the missing GEM super chambers requires further investigation. This PR is just a first step that allows reco to run to completion without crashing. The next step is to examine the plots derived from reco to try to understand what is missing from the GEMs.

#### PR validation:

The 11634.0_TTbar_14TeV+TTbar_14TeV_TuneCP5_2021 reco step was tested using XML geometry and a second time with DB geometry, and it ran successfully.

No backport is needed.

### DIFF
--- a/Validation/MuonGEMDigis/plugins/GEMCoPadDigiValidation.cc
+++ b/Validation/MuonGEMDigis/plugins/GEMCoPadDigiValidation.cc
@@ -165,7 +165,7 @@ void GEMCoPadDigiValidation::analyze(const edm::Event& event, const edm::EventSe
         lpArray[ptCounter] = etaPartition->centreOfPad(padArray[ptCounter]);
       }                     // end for
       if (ptCounter < 2) {  // Broke out of "for" loop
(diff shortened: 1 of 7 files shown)


cvuosalo:
type bugfix

cvuosalo:
@cmsbuild please test

cvuosalo:
@ianna FYI. I found out that GEM super chambers are missing because reco was crashing when I ran it using XML geometry. This PR prevents the crash but does not address the missing super chambers.

ianna:
@dildick - FYI

dildick:
@jshlee Do you know why GEM superchambers are missing? Is it picking up the slice test configuration?

jfernan2:
+1
Only the random L1T known changes

silviodonato:
+1

PR #29746, opened 2020-05-06
Update eventcontent structure for RecoHiEgamma

jeongeun:
#### PR description:  
Update event content definitions to explicitly have RECO to be a superset of AOD and for FEVT to be a superset of RECO. 1 file changed : 
+ `RecoHI/HiEgammaAlgos/python/RecoHiEgamma_EventContent_cff.py`

(The previous tasks for RecoHI event content is [PR#29576](https://github.com/cms-sw/cmssw/pull/29576) )

#### PR validation:
Event Content comparison check was also done and there is no change with these updates.
Tested in CMSSW_11_1_X, the basic test all passed in the [CMSSW PR instructions](https://cms-sw.github.io/PRWorkflow.html).

PR #29576, opened 2020-04-28
Update eventcontent structure in RecoHI packages

jeongeun:
#### PR description:  
Update event content definitions to explicitly have RECO to be a superset of AOD and for FEVT to be a superset of RECO. 5 files changed : 
+ `RecoHI/Configuration/python/RecoHI_EventContent_cff.py`
+ `RecoHI/HiCentralityAlgos/python/RecoHiCentrality_EventContent_cff.py `
+ `RecoHI/HiEvtPlaneAlgos/python/RecoHiEvtPlane_EventContent_cff.py` 
+ `RecoHI/HiMuonAlgos/python/RecoHiMuon_EventContent_cff.py `
+ `RecoHI/HiTracking/python/RecoHiTracker_EventContent_cff.py `

FYI, `RecoHI/HiEgammaAlgos/python/RecoHiEgamma_EventContent_cff.py`, `RecoHI/HiJetAlgos/python/RecoHiJets_EventContent_cff.py` files are work in progress and they will be updated in new PRs later (maybe separately).

(The previous tasks are [PR#29025](https://github.com/cms-sw/cmssw/pull/29025), [PR#29158](https://github.com/cms-sw/cmssw/pull/29158), [PR#29225](https://github.com/cms-sw/cmssw/pull/29225), [PR#29299](https://github.com/cms-sw/cmssw/pull/29299), [PR#29470](https://github.com/cms-sw/cmssw/pull/29470), [PR#29517](https://github.com/cms-sw/cmssw/pull/29517),  [PR#29535](https://github.com/cms-sw/cmssw/pull/29535), and  [PR#29571](https://github.com/cms-sw/cmssw/pull/29571) )

#### PR validation:
Event Content comparison check was also done and there is no change with these updates.
Tested in CMSSW_11_1_X, the basic test all passed in the [CMSSW PR instructions](https://cms-sw.github.io/PRWorkflow.html).

PR #29752, opened 2020-05-06
Update eventcontent structure for RecoHiJets

jeongeun:
#### PR description:  
Update event content definitions to explicitly have RECO to be a superset of AOD and for FEVT to be a superset of RECO. 1 file changed : 
+ `RecoHI/HiJetAlgos/python/RecoHiJets_EventContent_cff.py`

(The previous tasks for RecoHI event content : [PR#29576](https://github.com/cms-sw/cmssw/pull/29576) and [PR#29746](https://github.com/cms-sw/cmssw/pull/29746). )

#### PR validation:
Event Content comparison check was also done and there is no change with these updates.
Tested in CMSSW_11_1_X, the basic test all passed in the [CMSSW PR instructions](https://cms-sw.github.io/PRWorkflow.html).

### DIFF
--- a/RecoHI/HiJetAlgos/python/RecoHiJets_EventContent_cff.py
+++ b/RecoHI/HiJetAlgos/python/RecoHiJets_EventContent_cff.py
@@ -1,45 +1,37 @@
 import FWCore.ParameterSet.Config as cms
 
-#Full Event content ---- temporary
-RecoHiJetsRECO = cms.PSet(
-    outputCommands = cms.untracked.vstring( 'keep *_akPu3CaloJets_*_*',
-                                            'keep *_akPu4CaloJets_*_*', 
-                                            'keep *_akPu5CaloJets_*_*', 
-                                            'keep *_iterativeConePu5CaloJets_*_*', 
-                                            'keep *_akPu3PFJets_*_*',
-                                            'keep *_akPu4PFJets_*_*',
-                                            'keep *_akPu5PFJets_*_*',
-                                            'keep *_akCs3PFJets_*_*',
-                                            'keep *_akCs4PFJets_*_*',
-                                            'keep *_kt4PFJetsForRho_*_*',
-                                            'keep *_*HiGenJets_*_*',
-                                            'keep *_PFTowers_*_*',
-                                            'keep *_hiFJRhoProducer_*_*'
-
-                                           )
-    )
-
-RecoHiJetsFEVT = cms.PSet(
-    outputCommands = cms.untracked.vstring('keep *_*CaloJets_*_*',
-                                           'keep *_*PFJets_*_*',
-                                           'keep *_*HiGenJets_*_*',
-                                           'keep *_*PFTowers_*_*',
-                                           'keep *_*hiFJRhoProducer_*_*'
-                                           )
-    )
-
+# AOD content
 RecoHiJetsAOD = cms.PSet(
-    outputCommands = cms.untracked.vstring('keep *_*CaloJets_*_*',
-                                           'keep *_*PFJets_*_*',
-                                           'keep *_*HiGenJets_*_*',
-                                           'keep *_*PFTowers_*_*',
-                                           'keep *_*hiFJRhoProducer_*_*',
-                                           'keep CaloTowersSorted_towerMaker_*_*',
-                                           'drop recoCandidatesOwned_caloTowers_*_*',
-                                           'keep recoPFCandidates_particleFlowTmp_*_*'
-                                           )
-    )
+    outputCommands = cms.untracked.vstring(
+        'keep *_ak4CaloJets_*_*',
+        'keep *_akPu4CaloJets_*_*', 
+        'keep *_ak4PFJets_*_*',
+        'keep *_akPu3PFJets_*_*',
+        'keep *_akPu4PFJets_*_*',
+        'keep *_akCs3PFJets_*_*',
+        'keep *_akCs4PFJets_*_*',
+        'keep *_PFTowers_*_*',
+        'keep *_hiFJRhoProducer_*_*',
+        'keep CaloTowersSorted_towerMaker_*_*',
+        'keep recoPFCandidates_particleFlowTmp_*_*')
+)
 
+# RECO content
+RecoHiJetsRECO = cms.PSet(
+    outputCommands = cms.untracked.vstring( 
+        'keep *_akPu3CaloJets_*_*', 
+        'keep *_akPu5CaloJets_*_*', 
+        'keep *_iterativeConePu5CaloJets_*_*', 
+        'keep *_akPu5PFJets_*_*',
+        'keep recoCandidatesOwned_caloTowers_*_*',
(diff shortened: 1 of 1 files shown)


perrotta:
please test

RecoHI/HiJetAlgos/python/RecoHiJets_EventContent_cff.py
```diff
@@ -1,45 +1,37 @@
...
-
-                                           )
-    )
-
+    outputCommands = cms.untracked.vstring( 
+        'keep *_akPu3CaloJets_*_*', 
+        'keep *_akPu5CaloJets_*_*', 
+        'keep *_iterativeConePu5CaloJets_*_*', 
+        'keep *_akPu5PFJets_*_*',
+        'keep recoCandidatesOwned_caloTowers_*_*',
```
perrotta, line 26: This (or a similar) line was not present in the original event content

perrotta:
@mandrenguyen , the rationale of this restructuration of the reco-related event contents is:
- Explicitely make the AOD event content a subset of the RECO event content, which in its turn must be a subset of the FEVT event content
- Avoid using wildcards as much as possible, so that developers will be forced to indicate explicitely all products they want to have saved in a given event content, and only those ones

While doing so, it sometime becomes evident that some cleaning and reorganization is needed for a few packages. I think that The RecoHiJets package is one of them, as the AOD/RECO/FEVT event contents seem to be filled somehow independently one from each other. Some advice would be welcome in this case.

For example, looking at the AOD event content now in the release I would say that the rationale is:
- keep all CaloJets, PFJets, HiGenJets found in the Event, and add the few additional products explicitely listed in the event content

Relaxing the request of not using wild cards (which I'd think it can be lifted here if we say that "we want to keep **all**  CaloJets, PFJets, HiGenJets whatever produced) I would code it as

```PYTHON
RecoHiJetsAOD = cms.PSet(	
    outputCommands = cms.untracked.vstring(
                                           'keep *_*CaloJets_*_*',	
                                           'keep *_*PFJets_*_*',	
                                           'keep *_*HiGenJets_*_*',	
                                           'keep *_PFTowers_*_*',	
                                           'keep *_hiFJRhoProducer_*_*',	
                                           'keep CaloTowersSorted_towerMaker_*_*',	
                                           'keep recoPFCandidates_particleFlowTmp_*_*'	
                                           )	
    )
```

If so, then only the following should be added to the RECO event content:
```PYTHON
RecoHiJetsRECO = cms.PSet(	
       'keep *_kt4PFJetsForRho_*_*'
    )
```

And the FEVT event content can be kept identical to the RECO one.

Alternatively, we could be even more strictly remain linked to the spirit of this migration and explicitely list all CaloJets, PFJets, HiGenJets products we want to save, Since RECO and AOD outputs are typically produced in the same step, and since originally the intention was to save all those products in the AOD content, such a list cannot be different for the AOD and RECO case: therefore, if we want to be explicit, the list which is now (i.e. in current CMSSW) in the RECO event content should be also taken for the AOD content.

@mandrenguyen , could you please advise on which strategy would you consider more appropriate for these event contents?

mandrenguyen:
@perrotta  Note that we rarely use anything, but AOD.
Sticking to the scheme that AOD is a subset of RECO is a subset of AOD, then:


RecoHiJetsFEVT = cms.PSet(
    outputCommands = cms.untracked.vstring('keep *_*CaloJets_*_*',
                                           'keep *_*PFJets_*_*',
                                           'keep *_*HiGenJets_*_*',
                                           'keep *_*PFTowers_*_*',
                                           'keep *_*hiFJRhoProducer_*_*',
                                           'keep CaloTowersSorted_towerMaker_*_*',
                                           'drop recoCandidatesOwned_caloTowers_*_*',
                                           'keep recoPFCandidates_particleFlowTmp_*_*'
                                           )
    )

RecoHiJetsAOD = cms.PSet(
    outputCommands = cms.untracked.vstring(
 'drop recoCandidatesOwned_caloTowers_*_*',
)
)

Presumably, the wildcard in the first slot can be made explicit for all cases.

perrotta:
@mandrenguyen , thank you for you message, but maybe I was not clear enough: sorry for that.

The idea is:
- We build the AOD event conten
- We add products on top of it to build the RECO event content
- We add products on top of the RECO event content to build the FEVT one

To relax the wild cards would be great, and exactly in the spirit of this transition. If so, all CaloJets, PFJets, HiGenJets available in the Event should get expanded already in the AOD event content, and nothing else should be added in the RECO/FEVT ones.

Therefore, the following should be moved from the RECO (exclusive) content to the AOD content, which is the base content to be used in all cases:
```PYTHON
        'keep *_akPu3CaloJets_*_
        'keep *_akPu5CaloJets_*_*', 
        'keep *_iterativeConePu5CaloJets_*_*', 
        'keep *_akPu5PFJets_*_*',
```

If that list is exhaustive, I would suggest @jeongeun to proceed as such

perrotta:
And please @jeongeun don't forget the  ```*HiGenJets```, since they were still in use for the "old style" HI reconstruction (is it correct @mandrenguyen ?)

mandrenguyen:
@perrotta Yes, that's correct

mandrenguyen:
@perotta Ok, I'm following now. 
Yes, I agree, the collections you listed should be preserved in AOD to keep the current behavior.

RecoHI/HiJetAlgos/python/RecoHiJets_EventContent_cff.py
jeongeun: In original event content, this caloTower not in AOD.

Based on the hi workflow (140.53/158.0) RECO output's event content list, the caloTowers exist.
`edm::OwnVector<reco::Candidate,edm::ClonePolicy<reco::Candidate> >    "caloTowers"                ""                "reRECO"`

So I keep this caloTower in RECO and FEVT in this file just in case.

But If you don't need this caloTower then I will remove this line.
@mandrenguyen :  could you please confirm this ?

jeongeun:
Thanks a lot! @perrotta @mandrenguyen 
I will fix it and as mentioned in above, (https://github.com/cms-sw/cmssw/pull/29752#discussion_r421428842)
@mandrenguyen : please let me know if the caloTower doesn't need in RECO/FEVT. 
Thanks again.

perrotta:
please test

perrotta:
The following contents still remains with a wildcard since AOD:
```PYTHON
        'keep *_*HiGenJets_*_*',
        'keep *_*PFTowers_*_*',
        'keep *_*hiFJRhoProducer_*_*',
 ```

In any case, this reproduces exactly what was there in the previous form of that event content.
Removing that wild card and expanding to the actual products can be deferred to a forthcoming PR to be submitted by (or in coordination with) the HI group contact.

perrotta:
+1
- Event contents refactored according to the plans
- Whenever possible, wildcards have been also removed: still a few of them could be attempted to get expanded in a future update
- Jenkins tests pass

silviodonato:
+1

PR #29526, opened 2020-04-20
[RecoEgamma/EgammaTools] Updates to EnergyScaleCorrection class for scales and smearings in UL

jainshilpi:
This PR takes care of the updated txt file format which stores the scales and smearing corrections. 
Updates to following codes have been done:

(1)  EnergyScaleCorrection.cc/h - As Sam suggested, the code can now determine the txt file format and reads accordingly
(2) calibratedEgammas_cff.py - now the default txt file to be read is the UL one. For the last re-reco of 2017, run2_miniAOD_94XFall17 is used
(3) Small change to PhotonEnergyCalibrator/ElectronEnergyCalibrator have been done to avoid problems at the boundary. 

The related txt files can be found here: https://github.com/jainshilpi/EgammaAnalysis-ElectronTools/tree/UL2017SSV2/ScalesSmearings (*24Feb*)

runMatrix passed all the tests.

PR #29756, opened 2020-05-07
run2 re-miniAOD setup based on UL inputs

slava77:
- ```run2_miniAOD_UL``` process modifier is introduced for UL reminiAOD needs. 
    - note that using proc modifiers has been preferable over eras since a while; I took the chance to switch to the new way
    - I expect that one modifier should be enough for all years; a year-specific config modifications should be possible with standard modifiers in combination, e.g. as ```(ctpps_2016 & run2_miniAOD_UL).toModify```
        - if this will not be enough e.g. for 2016 we can add ```run2_miniAOD_UL16extra```, but hopefully this can be avoided
- only minimal updates to miniAOD parameters compared to the defaults are made to have the workflow running
    - rerun container discriminators as in run2_miniAOD_94XFall17 @mbluj @swozniewski please check and confirm if this needs a simple change (a more complex one would better be done by you in a follow up PR)
    - drop trying to add pixelClusterTagInfos as in other run2 reminis because this collection is not available in 106X AODs (added in 11_X)
    - PUPPI re-addition will be needed after a fixed version of #29254 is available
    - I did not add the photon/electron correctors, assuming that these will be added by EGM @jainshilpi either in #29526 
- tentative data and MC workflows are set up for 4 periods of Run2 (2 for 2016, 2017, and 2018)
    - the data workflow numbering takes the existing "legacy" remini as "10" as a reference (even though 0 is implicit due to the numeric nature) and assigns this UL to be "11": e.g. 136.8311 (= 136.83110) -> 136.83111. This minimizes conflicts with some nanoAOD workflows and uses a common pattern
    - the MC workflow numbering starts from 1325.5 as a base (this corresponds to the old 2016 MC remini) and appends "160" (1325.516) and "161" (1325.5161) for the 2016 pre and post-VFP, respectively; similarly "17" and "18" are used for the 2017 and 2018 UL workflows. This was an attempt to establish a somewhat intuitive pattern and avoid conflicts with existing wfs.
- the input datasets are not available at CERN for most of the cases, a request will need to be made (somehow, for data the cmsDriver is not generating a site=T2_CH_CERN requirement and the workflows can run out of the box).
    - for ```TTbar_13_reminiaod2017UL_INPUT``` I couldn't find an AODSIM relval with PU and use a RECO input instead
- 136.88811, 2018D UL remini wf is added to the short matrix

A summary of limited validation for 1325.517 (MC 2017) by comparing the original 10_6_X MINIAOD with remini in CMSSW_11_1_X_2020-05-04-1100 using FWLite: the differences are apparently in recognizable places where changes were introduced since 10_6_X
- PUPPI updates are visible, including bugfixes in the MET uncertainties and corrections
- muon beta variables are filled
- tau ID variables are hard to compare with a plain branch diff due to reshuffling of the IDs introduced in 11_1_X

@chayanit @pgunnell @wajidalikhan please let me know if you can make PhedEx request for the datasets proposed here in the tests or if an alternative can be transferred (I can make an update to the PR then)

### DIFF
--- a/Configuration/ProcessModifiers/python/run2_miniAOD_UL_cff.py
+++ b/Configuration/ProcessModifiers/python/run2_miniAOD_UL_cff.py
@@ -1,0 +1,6 @@
+import FWCore.ParameterSet.Config as cms
+
+# This modifier is for common settings to run miniAOD on top of 
+# ultra-legacy (during LS2) Run-2 AOD
+
+run2_miniAOD_UL =  cms.Modifier()
--- a/Configuration/PyReleaseValidation/python/WorkFlowRunner.py
+++ b/Configuration/PyReleaseValidation/python/WorkFlowRunner.py
@@ -120,16 +120,18 @@ class WorkFlowRunner(Thread):
                 # If the das output is not there or it's empty, consider it an
                 # issue of this step, not of the next one.
                 dasOutputPath = join(self.wfDir, 'step%d_dasquery.log'%(istep,))
-                if not exists(dasOutputPath):
-                  retStep = 1
-                  dasOutput = None
-                else:
-                  # We consider only the files which have at least one logical filename
-                  # in it. This is because sometimes das fails and still prints out junk.
-                  dasOutput = [l for l in open(dasOutputPath).read().split("\n") if l.startswith("/")]
-                if not dasOutput:
-                  retStep = 1
-                  isInputOk = False
+                # Check created das output in no-dryRun mode only
+                if not self.dryRun:
+                    if not exists(dasOutputPath):
+                        retStep = 1
+                        dasOutput = None
+                    else:
+                        # We consider only the files which have at least one logical filename
+                        # in it. This is because sometimes das fails and still prints out junk.
+                        dasOutput = [l for l in open(dasOutputPath).read().split("\n") if l.startswith("/")]
+                    if not dasOutput:
+                        retStep = 1
+                        isInputOk = False
                  
                 inFile = 'filelist:' + basename(dasOutputPath)
                 print("---")
--- a/Configuration/PyReleaseValidation/python/relval_standard.py
+++ b/Configuration/PyReleaseValidation/python/relval_standard.py
@@ -580,6 +580,12 @@ workflows[1325.4] = ['', ['TTbar_13','DIGIUP15','RECOUP15_HIPM','HARVESTUP15']]
 workflows[1325.5] = ['', ['ProdZEE_13_reminiaodINPUT','REMINIAOD_mc2016','HARVESTDR2_REMINIAOD_mc2016']]
 # reminiaod wf on 94X MC
 workflows[1325.51] = ['', ['TTbar_13_94XreminiaodINPUT','REMINIAOD_mc2017','HARVESTUP17_REMINIAOD_mc2017']]
+# reminiaod on UL MC
+workflows[1325.5160] = ['', ['TTbar_13_reminiaod2016UL_preVFP_INPUT','REMINIAOD_mc2016UL_preVFP','HARVESTDR2_REMINIAOD_mc2016UL_preVFP']]
+workflows[1325.5161] = ['', ['TTbar_13_reminiaod2016UL_postVFP_INPUT','REMINIAOD_mc2016UL_postVFP','HARVESTDR2_REMINIAOD_mc2016UL_postVFP']]
+workflows[1325.5170] = ['', ['TTbar_13_reminiaod2017UL_INPUT','REMINIAOD_mc2017UL','HARVESTUP17_REMINIAOD_mc2017UL']]
+workflows[1325.5180] = ['', ['TTbar_13_reminiaod2018UL_INPUT','REMINIAOD_mc2018UL','HARVESTUP18_REMINIAOD_mc2018UL']]
+
 
 # nanoaod wf without intermediate EDM,  starting from existing MINIAOD inputs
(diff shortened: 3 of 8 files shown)


slava77:
@cmsbuild please test

slava77:
@smuzaffar @mrodozov 
what is the strategy of introducing new input files for new workflows in the IB infrastructure?
Can you trigger transfers while the PR is in progress or does it happen after the integration?
Please let me know.
Thank you.

slava77:
assign xpog
in case there are comments/concerns on the strategy and the setup.

slava77:
+1

for https://github.com/cms-sw/cmssw/pull/29756	eac0dc2 
- tests passed OK, including the new workflow

kpedro88:
+upgrade

mbluj:
>     * rerun container discriminators as in run2_miniAOD_94XFall17 @mbluj @swozniewski please check and confirm if this needs a simple change (a more complex one would better be done by you in a follow up PR)

I took a look on this only now. Rerunig tau discriminators as in `run2_miniAOD_94XFall17` is correct to deal with different tau discriminator data format in 10_6_X UL AOD input samples compared to one expected by patTaus in 11_1_X. But, rerun discriminators will be as in release used for reMiniAOD rather than ones in input. This is not necessarily wrong (I think Tau POG might like it), but I wonder if it is what was expected. Regardless, I think that there is a way to read also discriminators in old (pre 11_1_X) format.

slava77:
> This is not necessarily wrong (I think Tau POG might like it), but I wonder if it is what was expected. Regardless, I think that there is a way to read also discriminators in old (pre 11_1_X) format.

Thank you for checking.
If Tau POG will like it, I would then accept this as expected. 😄 
In the context of 10_6_X, this will be a bit of a waste of CPU, but that's a rather small fraction.

For 11_X, rerunning may be more forward looking so that we do not commit ourselves to support reading the old discriminators in the future.

swozniewski:
Reading old data format is indeed possible. We recently used this functionality to fix two of the matrix workflows. Slava might remember. But I agree with Michal that rerunning IDs in the latest version might be preferred. If the decision depends on the ID, a hybrid solution could also be possible as the switch of reading old or new format is just a matter of python configuration.

mbluj:
> 
> 
> > This is not necessarily wrong (I think Tau POG might like it), but I wonder if it is what was expected. Regardless, I think that there is a way to read also discriminators in old (pre 11_1_X) format.
> 
> Thank you for checking.
> If Tau POG will like it, I would then accept this as expected. smile
> In the context of 10_6_X, this will be a bit of a waste of CPU, but that's a rather small fraction.
> 
> For 11_X, rerunning may be more forward looking so that we do not commit ourselves to support reading the old discriminators in the future.

I agree with the above arguments. We are checking with conveners.

One more issue: it will be needed to adjust also NanoAOD with the new modifier. But I think it is anyway beyond the scope of this PR, isn't it?

slava77:
> One more issue: it will be needed to adjust also NanoAOD with the new modifier. But I think it is anyway beyond the scope of this PR, isn't it?

I'm assuming that it can be done in a follow up.
I added xpog to the list of signatures earlier. So, if some updates are needed in this PR from xpog review, I can update.

silviodonato:
Do you have any comments ?
xpog:  @fgolf @peruzzim
analysis:  @santocch

chayanit:
+1

mbluj:
> 
> 
> > This is not necessarily wrong (I think Tau POG might like it), but I wonder if it is what was expected. Regardless, I think that there is a way to read also discriminators in old (pre 11_1_X) format.
> 
> Thank you for checking.
> If Tau POG will like it, I would then accept this as expected. smile
> In the context of 10_6_X, this will be a bit of a waste of CPU, but that's a rather small fraction.
> 
> For 11_X, rerunning may be more forward looking so that we do not commit ourselves to support reading the old discriminators in the future.

It is just a confirmation from conveners (@roger-wolf, @rmanzoni) that Tau POG likes it.

silviodonato:
do you have any comments?
analysis:  @santocch
xpog:  @fgolf @peruzzim

peruzzim:
+xpog

yes, for NanoAOD there will be a follow up in a separate pull request, that will come once the updates for MiniAOD are integrated

silviodonato:
+operations

silviodonato:
merge
@slava77 @smuzaffar @mrodozov have you copied the missing input files at CERN?
https://github.com/cms-sw/cmssw/pull/29756#issuecomment-625238866

slava77:
>     * PUPPI re-addition will be needed after a fixed version of #29254 is available

ah. So, the tests were not redone while the PUPPI PR was merged. This broke the tests in the IB

ahinzmann:
The relevant lines to change for these workflow are likely these:
https://github.com/cms-sw/cmssw/pull/29254/files#diff-76dede28e7751d7bb6c5c9542de4155cR430
https://github.com/cms-sw/cmssw/pull/29254/files#diff-416ad16d53535660648a4a4bb98636d6R33
https://github.com/cms-sw/cmssw/pull/29254/files#diff-76dede28e7751d7bb6c5c9542de4155cR374
@slava77 if you could take care that would be awesome

santocch:
+1

chayanit:
@slava77 sorry I just saw your comment about PheDEX. I think it should be fine even the input is not at CERN. We can switch these lines to search at other sites. 
https://github.com/slava77/cmssw/blob/eac0dc26d24ad80f85d4b4bf638ce1b931feac0f/Configuration/PyReleaseValidation/python/MatrixUtil.py#L180-L185

PR #29759, opened 2020-05-07
Strip hybrid zero suppression: do less work if the APV is already zero-suppressed

pieterdavid:
#### PR description:

Changed the implementation of the zero-suppression for hybrid strip tracker data, where most of the APVs are zero-suppressed in the FED, and some detected as having a non-flat or shifted baseline, and fully read out. When the software zero-suppression processes these data, the digis for APVs that were already zero-suppressed can simply be copied, but the current implementation would still recalculate the common mode (and ignore the result from that).
The time gain for a 2021 HI scenario is difficult to estimate (it depends on occupancy and the number of APVs that is not zero-suppressed by the FEDs), but should be significant, about an order of magnitude less time spent in the ZS (on the 2015 VR file I used for validation a factor 14).
With the changes above it didn't really make sense anymore to have different "convert hybrid to raw digis" and two suppressHybridData methods, so I made it one that can handle the calls from both the ZS module and the regional unpacker; I also removed the option to write the raw digis out again from the ZS for hybrid inputs (it was not used anywhere, and keeping this would have made the rest more complicated and inefficient).

#### PR validation:

No changes are expected. I did a digi-to-digi comparison by adding a flag to switch between the old and new implementation (removed only in the last commit), on 2018 repacked data (as in the HI benchmarks), [config](https://github.com/pieterdavid/SiStripLocalRecoDebugging/blob/master/test/test_hybridZSSpeedup_2018.py), and 2015 VR data with hybrid emulation, [config](https://github.com/pieterdavid/SiStripLocalRecoDebugging/blob/master/test/test_hybridZSSpeedup_2015VR.py); I also checked that the digis are the same between the full sequence and the regional unpacker (which calls the zero-suppression in a slightly different way, I think it may have been wrong before, but was not used).

Could this be tested with workflow 140.55? I think that's the only one that uses this code.

CC: @icali @CesarBernardes

### DIFF
--- a/RecoLocalTracker/SiStripZeroSuppression/plugins/SiStripZeroSuppression.cc
+++ b/RecoLocalTracker/SiStripZeroSuppression/plugins/SiStripZeroSuppression.cc
@@ -19,8 +19,7 @@ SiStripZeroSuppression::SiStripZeroSuppression(edm::ParameterSet const& conf)
       produceCalculatedBaseline(conf.getParameter<bool>("produceCalculatedBaseline")),
       produceBaselinePoints(conf.getParameter<bool>("produceBaselinePoints")),
       storeInZScollBadAPV(conf.getParameter<bool>("storeInZScollBadAPV")),
-      produceHybridFormat(conf.getParameter<bool>("produceHybridFormat")),
-      fasterHybridZS(conf.getUntrackedParameter<bool>("fasterHybridZS", false)) {
+      produceHybridFormat(conf.getParameter<bool>("produceHybridFormat")) {
   for (const auto& inputTag : conf.getParameter<std::vector<edm::InputTag>>("RawDigiProducersList")) {
     const auto& tagName = inputTag.instance();
     produces<edm::DetSetVector<SiStripDigi>>(tagName);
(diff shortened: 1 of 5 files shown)


perrotta:
please test

mmusich:
@perrotta 

> Could this be tested with workflow 140.55? I think that's the only one that uses this code.

Thanks!

perrotta:
please abort
(I should read everything before launching tests...)

perrotta:
please test workflow 140.55

RecoLocalTracker/SiStripZeroSuppression/plugins/SiStripZeroSuppression.cc
```diff
@@ -61,9 +61,13 @@ SiStripZeroSuppression::SiStripZeroSuppression(edm::ParameterSet const& conf)
     throw cms::Exception("Invalid option") << "When producing data in the hybrid format, the APV restorer must be "
                                               "configured with APVInspectMode='HybridEmulation'";
 
+  if ((!hybridInputs.empty()) && produceRawDigis) {
+    edm::LogWarning("SiStripZeroSuppression") << "Raw digis will not be saved for hybrid inputs";
```
slava77, line 65: is it worth to throw here instead?
Given that this is a stream module, IIUC, we'll get nThread counts of these warnings in the log file.
If instead of exception this condition is OK and this message is only useful for experts, perhaps a LogInfo is better?

pieterdavid: It's OK as long as the product isn't used (and if it is, there will be a ProductNotFound exception), so I made it a LogInfo indeed (in any case this option would only be used by experts).

perrotta:
please test workflow 140.55

slava77:
+1

for https://github.com/cms-sw/cmssw/pull/29759	4abf09d 
- local test with wf 140.55 shows about a factor of 2 speedup in the siStripZeroSuppression module time (from 1.5 s/ev to 0.8 s/ev); based on the PR description this is probably still within the expected range

silviodonato:
+1

PR #29760, opened 2020-05-07
Cellular Automaton: Add hit-compatibility parameters tunable by layer triplets

kjpena:
#### PR description:

This PR adds classes needed for tuning the hit-compatibility cuts of the Cellular Automaton by seeding layer triplets, following up on the comments received for:
https://github.com/cms-sw/cmssw/pull/27955

More details in:
https://indico.cern.ch/event/845939/contributions/3552658/attachments/1904226/3144492/CACuts_KPena_2019-09-09.pdf
https://indico.cern.ch/event/905963/contributions/3811694/attachments/2022955/3414974/CACuts_KPena_2020-04-20.pdf

#### PR validation:
Workflow 10824.1 was used to check the performance in Phase1. 
A tracking-only workflow with D49 geometry was used to check the performance in Phase2. 

Tracking validation plots are available here: 
https://cern.ch/karla.pena/trackingTasks/CACuts/CMSSW_11_1_0_pre6/

Request for comments from @felicepantaleo, @mtosi, @VinInn, @JanFSchulte, @makortel


mtosi:
@cmsbuild, please test

RecoPixelVertexing/PixelTriplets/src/CACut.h
```diff
@@ -0,0 +1,169 @@
...
+  void setCutValuesByInnerLayerIds(CAGraph &caLayers) {  //TODO: Is "const {" needed here?
+
+    for (auto &thisTriplet : valuesByLayerIds_) {
+      if (thisTriplet.hasValueByInnerLayerId)
+        continue;
+      auto it = std::find(thisTriplet.layerIds.begin(), thisTriplet.layerIds.end(), -1);
+      if (it != thisTriplet.layerIds.end())
+        continue;
+
+      std::cout << "(" << thisTriplet.layerIds[0] << ", " << thisTriplet.layerIds[1] << ", " << thisTriplet.layerIds[2]
```
perrotta, line 84: Please remove this cout (temporary debug?) or replace it with a LogWarning or similar thread safe messaging

RecoPixelVertexing/PixelTriplets/src/CACut.h
```diff
@@ -0,0 +1,169 @@
...
+                << "), " << thisTriplet.cutValue << std::endl;
+
+      bool foundOuterDoublet = false;
+
+      for (auto &thisOuterDoublet : valuesByInnerLayerIds_) {
+        if (thisOuterDoublet.outerDoubletIds[0] == thisTriplet.layerIds[1] &&
+            thisOuterDoublet.outerDoubletIds[1] == thisTriplet.layerIds[2]) {
+          thisOuterDoublet.innerLayerIds.emplace_back(thisTriplet.layerIds[0]);
+          thisOuterDoublet.cutValues.emplace_back(thisTriplet.cutValue);
+          foundOuterDoublet = true;
```
perrotta, line 94: Maybe you can break here?

RecoPixelVertexing/PixelTriplets/src/CACut.h
```diff
@@ -0,0 +1,169 @@
...
+
+      // Cut
+      thisCACut.cutValue = thisTriplet.cutValue;
+      thisCACut.hasValueByInnerLayerId = false;
+    }
+
+    setCutValuesByInnerLayerIds(caLayers);
+  }
+
+  void setCutValuesByInnerLayerIds(CAGraph &caLayers) {  //TODO: Is "const {" needed here?
```
perrotta, line 75: Where is ```caLayers``` needed in this method?

RecoPixelVertexing/PixelTriplets/src/CACut.h
kjpena: Removed, thanks!

RecoPixelVertexing/PixelTriplets/src/CACut.h
kjpena: Done!

RecoPixelVertexing/PixelTriplets/src/CACut.h
kjpena: Indeed, it's not. Removed.

perrotta:
please test

RecoPixelVertexing/PixelTriplets/src/CACut.h
```diff
@@ -0,0 +1,175 @@
...
+
+  class CAValueByLayerIds {
+  public:
+    std::vector<int> layerIds;
+    float cutValue;
+    bool hasValueByInnerLayerId;
+  };
+
+public:
+  bool useCACuts;
```
perrotta, line 165: Trailing underscore for local variables, as you already to for the private ones.
By the way: why do those two need to be public?

RecoPixelVertexing/PixelTriplets/src/CACut.h
```diff
@@ -0,0 +1,175 @@
...
+      valuesByTripletNames_.emplace_back();
+      auto &thisCACut = valuesByTripletNames_.back();
+
+      thisCACut.tripletName = thisTriplet.getParameter<std::string>("seedingLayers");
+      thisCACut.cutValue = thisTriplet.getParameter<double>("cut");
+    }
+  }
+
+  void setCutValuesByLayerIds(CAGraph &caLayers) {
+    if (!useCACuts || foundAllLayerIds)
```
perrotta, line 44: This check shouldnt be needed:
- ```useCACuts``` is prevented being "false" from line 24
- ```foundAllLayerIds``` is initialized as "false" in the constructor

RecoPixelVertexing/PixelTriplets/src/CACut.h
perrotta: However, ```useCACuts``` and ```foundAllLayerIds``` are  only set here, but never used: what are they expected to be used for? Shouldn't they just get removed?

RecoPixelVertexing/PixelTriplets/src/CACut.h
```diff
@@ -0,0 +1,175 @@
...
+        return thisCut;
+      }
+    }
+
+    CAValuesByInnerLayerIds emptyCutsByInnerLayer(defaultCut_);
+    return emptyCutsByInnerLayer;
+  }
+
+private:
+  class CAValueByTripletName {
```
perrotta, line 151: As simple as they are, ```CAValueByTripletName``` and ```CAValueByLayerIds``` could simply be C++ ```struct```'s

RecoPixelVertexing/PixelTriplets/src/CACut.h
```diff
@@ -0,0 +1,175 @@
...
+    float at(int layerId) const {
+      for (size_t thisLayer = 0; thisLayer < innerLayerIds.size(); thisLayer++) {
+        if (innerLayerIds.at(thisLayer) == layerId)
+          return cutValues.at(thisLayer);
+      }
+
+      return defaultCut_;
+    }
+
+    std::vector<int> outerDoubletIds;
```
perrotta, line 131: Trailing underscor for local variables

RecoPixelVertexing/PixelTriplets/src/CACut.h
```diff
@@ -0,0 +1,175 @@
...
+        newOuterDoublet.outerDoubletIds.emplace_back(thisTriplet.layerIds[2]);
+        newOuterDoublet.innerLayerIds.emplace_back(thisTriplet.layerIds[0]);
+        newOuterDoublet.cutValues.emplace_back(thisTriplet.cutValue);
+      }
+
+      thisTriplet.hasValueByInnerLayerId = true;
+    }
+  }
+
+  class CAValuesByInnerLayerIds {
```
perrotta, line 118: Maybe this can also be more conveniently a C++ struct

RecoPixelVertexing/PixelTriplets/src/CACut.h
```diff
@@ -0,0 +1,175 @@
...
+#include "FWCore/ParameterSet/interface/ParameterSetDescription.h"
+#include "CAGraph.h"
+
+class CACut {
+public:
+  explicit CACut(const double defaultCut, const std::vector<edm::ParameterSet> &tripletCuts)
+      : useCACuts(true), foundAllLayerIds(false), defaultCut_(defaultCut) {
+    if (tripletCuts.size() == 1 && tripletCuts[0].getParameter<double>("cut") == -1.) {
+      useCACuts = false;
+      //TODO Uncomment the Following warning after tuning
```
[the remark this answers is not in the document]
kjpena: Since there will be no tuning for Run2, this would indeed produce unnecessary noise. The idea was to have a warning when the current default cut is used instead of the tuned values e.g. in case of future changes to the configurations or CA code. Would it make sense to keep it as a LogDebug instead or should I remove it altogether?

RecoPixelVertexing/PixelTriplets/src/CACut.h
kjpena: They are used in setCutValuesByLayerIds to skip the layer name -> layer id mapping when not needed. Would you suggest to do this differently?

perrotta:
You are right @kipena: my comment about the two bool's not being used  
was wrong.


kjpena <notifications@github.com> ha scritto:

perrotta:
@kjpena please let us know when you plan to reply to the review comments

perrotta:
@kjpena please let us know if you intend to follow up with this PR. Otherwise, we will (temporarily) reject it for reco while waiting for something happening here.

kjpena:
> @kjpena please let us know if you intend to follow up with this PR. Otherwise, we will (temporarily) reject it for reco while waiting for something happening here.

@perrotta Sorry for the late reply. Yes, I will follow up on the review comments and push the changes in the coming days

RecoPixelVertexing/PixelTriplets/src/CACut.h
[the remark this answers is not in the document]
kjpena: How about keeping it as a LogDebug instead? Or should I remove it altogether?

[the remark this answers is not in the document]
kjpena: And should the warning on line 71 also be a LogDebug then?

RecoPixelVertexing/PixelTriplets/src/CACut.h
kjpena: Do we agree that this check should stay, to avoid the layer name -> layer id mapping when it's not needed?

RecoPixelVertexing/PixelTriplets/src/CACut.h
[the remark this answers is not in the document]
perrotta: If they need to be "warning", i.e. if they will happen rarely and they are worth being notified, they can stay as such. Otherwise you can move them to LogDebug.
The one at L71 seems really needing a warning, according to my understanding. Unless it can be fired also in "normal" situations, in which case it doesn't
In any case, staying commented out doesn't serve any purpose.

RecoPixelVertexing/PixelTriplets/src/CACut.h
perrotta: This comment was wrong, indeed: not to be taken into account

RecoPixelVertexing/PixelTriplets/src/CACut.h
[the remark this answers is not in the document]
kjpena: Since there will be no tuned values for Run2, this warning would indeed produce unnecessary noise. I can move it to LogDebug, to be used when working on the tune itself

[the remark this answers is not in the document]
kjpena: About the warning on L71, maybe @makortel or @mtosi can comment? The CAGraph is updated for every event, and includes only layers with hits. This means the warning would also be fired for events where there was a detector layer with no hits.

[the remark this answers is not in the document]
makortel: I would imagine events without hits on a pixel layer to be rare, but I would say such an event would nevertheless be a good event and not an indication of a mistake in the configuration (just my 2 cents).

Thinking more broadly, the current need to interpret the layer combinations of `SeedingLayersEDProducer` is suboptimal (and I'm guilty as charged for that). Possible ways to check the compatibility of the configurations could be to pass the layer combinations as a Run product (or eventually `ProcessBlock`), or via EventSetup. I think either would be beyond the scope of this PR.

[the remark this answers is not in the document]
kjpena: Thanks @makortel! Would you have any suggestions on how to improve either this particular check of the layer combinations or the warning itself? Should we make it clear that one should check the layers in the configuration, but the warning could also be fired for rare events where one of the detector layers had no hits?

perrotta:
@cmsbuild please test

RecoPixelVertexing/PixelTriplets/src/CAGraph.h
```diff
@@ -42,6 +48,14 @@ struct CALayerPair {
 };
 
 struct CAGraph {
+  int getLayerId(const std::string &layerName) {
+    for (auto thisLayer : theLayers) {
```
perrotta, line 52: ```const auto &```

perrotta (commented):
Sorry, I missed this in my previous review
After it, I think I'm done

RecoPixelVertexing/PixelTriplets/src/CAGraph.h
kjpena: No problem! Just to be sure, you mean for `this Layer`, right?

perrotta:
I am looking at the performance figures pointed out in the PR description, in particular the timing:
https://kjpena.web.cern.ch/kjpena/trackingTasks/CACuts/CMSSW_11_1_0_pre6/
Were those performance mesured with the current PR, or with previous #27955? Since they are done with pre6 I imagine it is this one, but let me ask...

RecoPixelVertexing/PixelTriplets/src/CAGraph.h
perrotta: Yes, thank you!

kjpena:
> I am looking at the performance figures pointed out in the PR description, in particular the timing:
> https://kjpena.web.cern.ch/kjpena/trackingTasks/CACuts/CMSSW_11_1_0_pre6/
> Were those performance mesured with the current PR, or with previous #27955? Since they are done with pre6 I imagine it is this one, but let me ask...

This one

perrotta:
@cmsbuild please test

kjpena:
Sorry, seems like I pushed too fast. I will check and push again

perrotta:
Uhm, perhaps it is enough to remove the ```const``` from my previous suggestion...

RecoPixelVertexing/PixelTriplets/src/CAGraph.h
```diff
@@ -6,14 +6,19 @@
 #include <vector>
 
 struct CALayer {
-  CALayer(const std::string &layerName, std::size_t numberOfHits) : theName(layerName) {
+  CALayer(const std::string &layerName, const int seqNum, std::size_t numberOfHits)
+      : theName(layerName), theSeqNum(seqNum) {
     isOuterHitOfCell.resize(numberOfHits);
   }
 
   bool operator==(const std::string &otherString) { return otherString == theName; }
```
kjpena, line 14: We can keep the changes from the previous suggestion but would need to add `const` after the function declaration here. The == operator is used in other parts of the CA code, but it seems to work properly also after making this change (testing with workflow 10824.1)

perrotta: Whatever it works. Removing the const looks safer to me, but if you have already verified that modifying the == operator works as well I let it to you to decide.

perrotta:
please test

perrotta:
+1
- Hit-compatibility cuts of the Cellular Automaton are made configurable by layer triples, not uniform as before
- The default config maintained so far reproduces the previous uniform settings: therefore no differences are expected in the outputs, and none are observed in the automatic jenkins tests
- Timing measured for the new setup does not differ relevantly from the ones taken by the version of the modules currently in release

silviodonato:
+1

PR #29763, opened 2020-05-07
Initialize SpatialRange to avoid compilation warnings on PPC64LE

smuzaffar:
#### PR description:

In Power IBs, we have few compilation warnings [a]. This could happen if there are no hits and https://cmssdt.cern.ch/lxr/source/RecoPPS/Local/interface/CTPPSTimingTrackRecognition.h#0198 return the un-initialized  SpatialRange. This PR propose to initialize  SpatialRange with default values.

[a]
```
RecoPPS/Local/src/TotemTimingTrackRecognition.cc: In member function 'virtual int TotemTimingTrackRecognition::produceTracks(edm::DetSet<TotemTimingLocalTrack>&)':
  RecoPPS/Local/src/TotemTimingTrackRecognition.cc:66:96: warning: 'result.CTPPSTimingTrackRecognition<TotemTimingLocalTrack, TotemTimingRecHit>::SpatialRange::zEnd' may be used uninitialized in this function [-Wmaybe-uninitialized]
          math::XYZPoint positionSigma(xTrack.x0Sigma(), yTrack.y0Sigma(), 0.5f * (hitRange.zEnd - hitRange.zBegin));
                                                                                 ~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~
  RecoPPS/Local/src/TotemTimingTrackRecognition.cc:66:96: warning: 'result.CTPPSTimingTrackRecognition<TotemTimingLocalTrack, TotemTimingRecHit>::SpatialRange::zBegin' may be used uninitialized in this function [-Wmaybe-uninitialized]
 In file included from RecoPPS/Local/interface/TotemTimingTrackRecognition.h:18,
                 from RecoPPS/Local/src/TotemTimingTrackRecognition.cc:11:
  RecoPPS/Local/interface/CTPPSTimingTrackRecognition.h:131:15: warning: 'result.CTPPSTimingTrackRecognition<TotemTimingLocalTrack, TotemTimingRecHit>::SpatialRange::yEnd' may be used uninitialized in this function [-Wmaybe-uninitialized]
    const float profileRangeEnd = param.rangeEnd + profileRangeMargin;
               ^~~~~~~~~~~~~~~
 ```
#### PR validation:

Local compilation with both normal amd64 and ppc64le IBs is successful (without any warnings).

### DIFF
--- a/RecoPPS/Local/interface/CTPPSTimingTrackRecognition.h
+++ b/RecoPPS/Local/interface/CTPPSTimingTrackRecognition.h
@@ -197,7 +197,7 @@ template <class TRACK_TYPE, class HIT_TYPE>
 inline typename CTPPSTimingTrackRecognition<TRACK_TYPE, HIT_TYPE>::SpatialRange
 CTPPSTimingTrackRecognition<TRACK_TYPE, HIT_TYPE>::getHitSpatialRange(const HitVector& hits) {
   bool initialized = false;
-  SpatialRange result;
+  SpatialRange result = {};
 
   for (const auto& hit : hits) {
     const float xBegin = hit.x() - 0.5f * hit.xWidth(), xEnd = hit.x() + 0.5f * hit.xWidth();


smuzaffar:
please teest

perrotta:
+1
- Technical: it initializes SpatialRange with default values, which will be the returned ones when there are no hits
- No changes expected: ```getHitSpatialRange(hits)``` is called in two places, and in both cases there is a check of the ```hits``` vector not being empty **after** that call:
   - ```RecoPPS/Local/src/CTPPSDiamondTrackRecognition.cc```
   - ```RecoPPS/Local/src/TotemTimingTrackRecognition.cc```
   - Therefore, even with this fix no changes are expected in outputs
- Jenkins tests pass and show no differences, as it should

perrotta:
@fabferro : while this fix should remain, wouldn't be more convenient in the two calls pointed out above to check for the non-emptiness of the vecor of hits **before** calling ```getHitSpatialRange``` and ```producePartialTracks``` (instead than after them), so that useless computations can be avoided?

fabferro:
> 
> 
> @fabferro : while this fix should remain, wouldn't it be more convenient in the two calls pointed out above to check for the non-emptiness of the vecor of hits **before** calling `getHitSpatialRange` and `producePartialTracks` (instead than after them), so that useless computations can be avoided?

@perrotta that's a good question, I actually don't know if there's a reason why it's not been implemented already. I'd forward the question to the developers of the PPS timing code @forthommel @jan-kaspar

silviodonato:
+1

forthommel:
> @fabferro : while this fix should remain, wouldn't it be more convenient in the two calls pointed out above to check for the non-emptiness of the vecor of hits **before** calling `getHitSpatialRange` and `producePartialTracks` (instead than after them), so that useless computations can be avoided?

Indeed, thanks @perrotta! I will tackle this in a follow-up PR.

PR #29764, opened 2020-05-07
First version of DeepMET

steggema:
#### PR description:

This PR introduces a producer for DeepMET, a deep-learning-based missing pT estimator. The producer creates a new MET collection, and a test configuration is included. The plan is to create a separate PR for possible inclusion in central sequences, e.g. for the upcoming ReMiniAOD campaign.

The tensorflow models for this PR are proposed in a separate PR to cms-data/RecoMET-METPUSubtraction https://github.com/cms-data/RecoMET-METPUSubtraction/pull/5

There are different trainings for different years/conditions (2016, 2018, phase 2), and for 2018 and phase 2 also non-response-corrected trainings. 

Presentations:
https://indico.cern.ch/event/912067/contributions/3835851/ (most recent update)
https://indico.cern.ch/event/883809/contributions/3733818/ (CMS week JetMET meeting)
https://indico.cern.ch/event/854654/contributions/3594579/ (first presentation in MET meeting)

Note that an alternative implementation would be to store additional weights for each PFCandidate and calculate MET and possibly jets in a subsequent step. However, we prefer to leave this option to future studies given that we have not checked the performance on jets (and assume some non-trivial effects) and that this would lead to an increase in complexity of the integration and the additional storage required, in particular if we want to have different METs (e.g. response and non-response-corrected) for a single campaigns.

#### PR validation:

The code has been validated by running it in large-scale checks with simulated and data events to evaluate the performance of the algorithm. A test configuration is included. We have not run any memory or timing tests but we suspect that it runs fast and consumes little memory given the models are small compared to most other tensorflow models.

CPU and memory reports can be found under the following links, obtained on 1000 events from the 136.8311_RunJetHT2017F workflow:
https://steggema.web.cern.ch/steggema/cgi-bin/igprof-navigator/deepmet_cpu_reminiaod
https://steggema.web.cern.ch/steggema/cgi-bin/igprof-navigator/deepmet_mem_reminiaod (note that DeepMET does not seem to appear here, supposedly because it's not in the top 1000, see the text file linked below)

Text dumps are also available here: /afs/cern.ch/user/s/steggema/public/DeepMETIntegration/

-

@intrepid42 @yongbinfeng

### DIFF
--- a/PhysicsTools/PatAlgos/python/slimming/MicroEventContent_cff.py
+++ b/PhysicsTools/PatAlgos/python/slimming/MicroEventContent_cff.py
@@ -20,8 +20,6 @@ MicroEventContent = cms.PSet(
         'keep *_slimmedMETs_*_*',
         'keep *_slimmedMETsNoHF_*_*',
         'keep *_slimmedMETsPuppi_*_*',
-        'keep *_slimmedMETsDeep_*_*',
-        'keep *_slimmedMETsDeepResp_*_*',
         'keep *_slimmedSecondaryVertices_*_*',
         'keep *_slimmedLambdaVertices_*_*',
         'keep *_slimmedKshortVertices_*_*',
--- a/PhysicsTools/PatAlgos/python/slimming/MicroEventContent_cff.py
+++ b/PhysicsTools/PatAlgos/python/slimming/MicroEventContent_cff.py
@@ -20,6 +20,8 @@ MicroEventContent = cms.PSet(
         'keep *_slimmedMETs_*_*',
         'keep *_slimmedMETsNoHF_*_*',
         'keep *_slimmedMETsPuppi_*_*',
+        'keep *_slimmedMETsDeep_*_*',
+        'keep *_slimmedMETsDeepResp_*_*',
         'keep *_slimmedSecondaryVertices_*_*',
         'keep *_slimmedLambdaVertices_*_*',
         'keep *_slimmedKshortVertices_*_*',
--- a/PhysicsTools/PatAlgos/python/slimming/miniAOD_tools.py
+++ b/PhysicsTools/PatAlgos/python/slimming/miniAOD_tools.py
@@ -454,26 +454,25 @@ def miniAOD_customizeCommon(process):
 
     process.load('RecoMET.METPUSubtraction.deepMETProducer_cfi')
 
-
-    addToProcessAndTask('slimmedMETsDeep', process.deepMETProducer.clone(), process, task)
-    addToProcessAndTask('slimmedMETsDeepResp', process.deepMETProducer.clone(), process, task)
-    process.slimmedMETsDeepResp.graph_path = 'RecoMET/METPUSubtraction/data/deepmet/deepmet_resp_v1_2018.pb'
+    addToProcessAndTask('deepMETsResolutionTune', process.deepMETProducer.clone(), process, task)
+    addToProcessAndTask('deepMETsResponseTune', process.deepMETProducer.clone(), process, task)
+    process.deepMETsResponseTune.graph_path = 'RecoMET/METPUSubtraction/data/deepmet/deepmet_resp_v1_2018.pb'
 
     from Configuration.Eras.Modifier_phase2_common_cff import phase2_common
     phase2_common.toModify(
-        process.slimmedMETsDeep,
+        process.deepMETsResolutionTune,
         max_n_pf=12500,
         graph_path="RecoMET/METPUSubtraction/data/deepmet/deepmet_v1_phase2.pb"
     )
     phase2_common.toModify(
-        process.slimmedMETsDeepResp,
+        process.deepMETsResponseTune,
         max_n_pf=12500,
         graph_path="RecoMET/METPUSubtraction/data/deepmet/deepmet_resp_v1_phase2.pb"
     )
 
     from Configuration.Eras.Modifier_run2_jme_2016_cff import run2_jme_2016
     run2_jme_2016.toModify(
-        process.slimmedMETsDeepResp,
+        process.deepMETsResponseTune,
         graph_path="RecoMET/METPUSubtraction/data/deepmet/deepmet_resp_v1_2016.pb"
     )
     # add DetIdAssociatorRecords to EventSetup (for isolatedTracks)
--- a/PhysicsTools/PatAlgos/python/slimming/miniAOD_tools.py
+++ b/PhysicsTools/PatAlgos/python/slimming/miniAOD_tools.py
@@ -462,19 +462,19 @@ def miniAOD_customizeCommon(process):
     from Configuration.Eras.Modifier_phase2_common_cff import phase2_common
     phase2_common.toModify(
         process.slimmedMETsDeep,
-        max_n_pf=cms.uint32(12500),
-        graph_path=cms.string("RecoMET/METPUSubtraction/data/deepmet/deepmet_v1_phase2.pb")
+        max_n_pf=12500,
+        graph_path="RecoMET/METPUSubtraction/data/deepmet/deepmet_v1_phase2.pb"
     )
     phase2_common.toModify(
         process.slimmedMETsDeepResp,
-        max_n_pf=cms.uint32(12500),
-        graph_path=cms.string("RecoMET/METPUSubtraction/data/deepmet/deepmet_resp_v1_phase2.pb")
+        max_n_pf=12500,
+        graph_path="RecoMET/METPUSubtraction/data/deepmet/deepmet_resp_v1_phase2.pb"
     )
 
     from Configuration.Eras.Modifier_run2_jme_2016_cff import run2_jme_2016
     run2_jme_2016.toModify(
         process.slimmedMETsDeepResp,
-        graph_path=cms.string("RecoMET/METPUSubtraction/data/deepmet/deepmet_resp_v1_2016.pb")
+        graph_path="RecoMET/METPUSubtraction/data/deepmet/deepmet_resp_v1_2016.pb"
     )
     # add DetIdAssociatorRecords to EventSetup (for isolatedTracks)
     process.load("TrackingTools.TrackAssociator.DetIdAssociatorESProducer_cff")
--- a/PhysicsTools/PatAlgos/python/slimming/miniAOD_tools.py
+++ b/PhysicsTools/PatAlgos/python/slimming/miniAOD_tools.py
@@ -177,6 +177,8 @@ def miniAOD_customizeCommon(process):
 
     process.load('PhysicsTools.PatAlgos.slimming.slimmedMETs_cfi')
     task.add(process.slimmedMETs)
+    process.slimmedMETs.addDeepMETs = True
+
     addToProcessAndTask('slimmedMETsNoHF', process.slimmedMETs.clone(), process, task)
     process.slimmedMETsNoHF.src = cms.InputTag("patMETsNoHF")
     process.slimmedMETsNoHF.rawVariation =  cms.InputTag("patPFMetNoHF")
--- a/PhysicsTools/PatAlgos/python/slimming/miniAOD_tools.py
+++ b/PhysicsTools/PatAlgos/python/slimming/miniAOD_tools.py
@@ -452,6 +452,30 @@ def miniAOD_customizeCommon(process):
     process.slimmedMETsPuppi.tXYUncForT01Smear = cms.InputTag("patPFMetT0pcT1SmearTxyPuppi")
     del process.slimmedMETsPuppi.caloMET
 
+    process.load('RecoMET.METPUSubtraction.deepMETProducer_cfi')
+
+
+    addToProcessAndTask('slimmedMETsDeep', process.deepMETProducer.clone(), process, task)
+    addToProcessAndTask('slimmedMETsDeepResp', process.deepMETProducer.clone(), process, task)
+    process.slimmedMETsDeepResp.graph_path = 'RecoMET/METPUSubtraction/data/deepmet/deepmet_resp_v1_2018.pb'
+
+    from Configuration.Eras.Modifier_phase2_common_cff import phase2_common
+    phase2_common.toModify(
+        process.slimmedMETsDeep,
+        max_n_pf=cms.uint32(12500),
+        graph_path=cms.string("RecoMET/METPUSubtraction/data/deepmet/deepmet_v1_phase2.pb")
+    )
+    phase2_common.toModify(
+        process.slimmedMETsDeepResp,
+        max_n_pf=cms.uint32(12500),
+        graph_path=cms.string("RecoMET/METPUSubtraction/data/deepmet/deepmet_resp_v1_phase2.pb")
+    )
+
+    from Configuration.Eras.Modifier_run2_jme_2016_cff import run2_jme_2016
+    run2_jme_2016.toModify(
+        process.slimmedMETsDeepResp,
+        graph_path=cms.string("RecoMET/METPUSubtraction/data/deepmet/deepmet_resp_v1_2016.pb")
+    )
     # add DetIdAssociatorRecords to EventSetup (for isolatedTracks)
     process.load("TrackingTools.TrackAssociator.DetIdAssociatorESProducer_cff")
 
--- a/RecoMET/METPUSubtraction/plugins/DeepMETProducer.cc
+++ b/RecoMET/METPUSubtraction/plugins/DeepMETProducer.cc
@@ -1,0 +1,163 @@
+#include "FWCore/Framework/interface/Frameworkfwd.h"
+#include "FWCore/Framework/interface/stream/EDProducer.h"
+#include "FWCore/Framework/interface/Event.h"
+#include "FWCore/Framework/interface/MakerMacros.h"
+
+#include "DataFormats/PatCandidates/interface/MET.h"
+#include "DataFormats/PatCandidates/interface/PackedCandidate.h"
+
+#include "PhysicsTools/TensorFlow/interface/TensorFlow.h"
+
+struct DeepMETCache {
+  std::atomic<tensorflow::GraphDef*> graph_def;
+};
+
+class DeepMETProducer : public edm::stream::EDProducer<edm::GlobalCache<DeepMETCache> > {
+public:
+  explicit DeepMETProducer(const edm::ParameterSet&, const DeepMETCache*);
+  void produce(edm::Event& event, const edm::EventSetup& setup) override;
+  static void fillDescriptions(edm::ConfigurationDescriptions& descriptions);
+
+  // static methods for handling the global cache
+  static std::unique_ptr<DeepMETCache> initializeGlobalCache(const edm::ParameterSet&);
+  static void globalEndJob(DeepMETCache*);
+private:
+  const edm::EDGetTokenT<std::vector<pat::PackedCandidate> > pf_token_;
+  float norm_;
+  bool ignore_leptons_;
+  unsigned int max_n_pf_;
+
+  tensorflow::Session* session_;
+
+  std::unordered_map<int, int32_t> charge_embedding_;
+  std::unordered_map<int, int32_t> pdg_id_embedding_;
+};
+
+
+namespace {
+  float divide_and_rm_outlier(float val, float norm) {
+    float ret_val = val/norm;
+    if (ret_val > 1e6 || ret_val < -1e6)
+      return 0.;
+    return ret_val;
+  }
+}
+
+DeepMETProducer::DeepMETProducer(const edm::ParameterSet& cfg, const DeepMETCache* cache) :
+  pf_token_(consumes<std::vector<pat::PackedCandidate> >(cfg.getParameter<edm::InputTag>("pf_src"))),
+  norm_(cfg.getParameter<double>("norm_factor")), 
+  ignore_leptons_(cfg.getParameter<bool>("ignore_leptons")),
+  max_n_pf_(cfg.getParameter<unsigned int>("max_n_pf")) {
+    session_ = tensorflow::createSession(cache->graph_def);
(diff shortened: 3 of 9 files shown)


slava77:
please provide some estimates of CPU time and memory use.

Other than just enabling the code in mini/nanoAOD, what else needs to happen (more developments, retraining)?
If nothing else, I think it would be better to add this now at least to the miniAOD step.

steggema:
> please provide some estimates of CPU time and memory use.

Will be added.

> 
> Other than just enabling the code in mini/nanoAOD, what else needs to happen (more developments, retraining)?
> If nothing else, I think it would be better to add this now at least to the miniAOD step.

We don't need and foresee any additional developments or training at the moment (except for uncertainty estimates, which are however a completely different matter both content- and code-wise), so we'll add this to the MiniAOD step if you prefer to have this included in this PR.

Given other commitments, I expect an update by sometime next week.

slava77:
> Given other commitments, I expect an update by sometime next week.

OK
Please note

the PR tests can not proceed without this addressed

slava77:
test parameters
- pull_request = cms-data/RecoMET-METPUSubtraction#5

slava77:
@cmsbuild please test

slava77:
@cmsbuild please test

steggema:
I've updated the PR description with CPU and memory measurements. Please let me know if I should run on other samples or different tests. A copy of the text added to the PR description follows: 

CPU and memory reports can be found under the following links, obtained on 1000 events from the 136.8311_RunJetHT2017F workflow:
https://steggema.web.cern.ch/steggema/cgi-bin/igprof-navigator/deepmet_cpu_reminiaod
https://steggema.web.cern.ch/steggema/cgi-bin/igprof-navigator/deepmet_mem_reminiaod (note that DeepMET does not seem to appear here, supposedly because it's not in the top 1000, see the text file linked below)

Text dumps are also available here: /afs/cern.ch/user/s/steggema/public/DeepMETIntegration/

santocch:
+1

PhysicsTools/PatAlgos/python/slimming/miniAOD_tools.py
```diff
@@ -452,6 +452,30 @@ def miniAOD_customizeCommon(process):
...
+
+    addToProcessAndTask('slimmedMETsDeep', process.deepMETProducer.clone(), process, task)
+    addToProcessAndTask('slimmedMETsDeepResp', process.deepMETProducer.clone(), process, task)
+    process.slimmedMETsDeepResp.graph_path = 'RecoMET/METPUSubtraction/data/deepmet/deepmet_resp_v1_2018.pb'
+
+    from Configuration.Eras.Modifier_phase2_common_cff import phase2_common
+    phase2_common.toModify(
+        process.slimmedMETsDeep,
+        max_n_pf=cms.uint32(12500),
+        graph_path=cms.string("RecoMET/METPUSubtraction/data/deepmet/deepmet_v1_phase2.pb")
```
slava77, line 466: please drop type specifications for all parameters which already exist.

```suggestion
        max_n_pf=12500,
        graph_path="RecoMET/METPUSubtraction/data/deepmet/deepmet_v1_phase2.pb"
```

same for the rest of this file

RecoMET/METPUSubtraction/plugins/DeepMETProducer.cc
```diff
@@ -0,0 +1,162 @@
...
+
+  tensorflow::Session* session_;
+
+  std::unordered_map<int, int32_t> charge_embedding_;
+  std::unordered_map<int, int32_t> pdg_id_embedding_;
+};
+
+namespace {
+  float divide_and_rm_outlier(float val, float norm) {
+    float ret_val = val / norm;
```
slava77, line 39: it looks like ```norm``` is the same for all calls.
Better compute an inverse of it and do a product here, which is faster. 
Perhaps "scale" will be more intuitive in that case.

RecoMET/METPUSubtraction/plugins/DeepMETProducer.cc
```diff
@@ -0,0 +1,162 @@
...
+    : pf_token_(consumes<std::vector<pat::PackedCandidate> >(cfg.getParameter<edm::InputTag>("pf_src"))),
+      norm_(cfg.getParameter<double>("norm_factor")),
+      ignore_leptons_(cfg.getParameter<bool>("ignore_leptons")),
+      max_n_pf_(cfg.getParameter<unsigned int>("max_n_pf")) {
+  session_ = tensorflow::createSession(cache->graph_def);
+  produces<pat::METCollection>();
+  charge_embedding_ = {{-1, 0}, {0, 1}, {1, 2}};
+  pdg_id_embedding_ = {
+      {-211, 0}, {-13, 1}, {-11, 2}, {0, 3}, {1, 4}, {2, 5}, {11, 6}, {13, 7}, {22, 8}, {130, 9}, {211, 10}};
+  ;
```
slava77, line 56: what is this ```;``` for ?

RecoMET/METPUSubtraction/plugins/DeepMETProducer.cc
```diff
@@ -0,0 +1,162 @@
...
+
+  // static methods for handling the global cache
+  static std::unique_ptr<DeepMETCache> initializeGlobalCache(const edm::ParameterSet&);
+  static void globalEndJob(DeepMETCache*);
+
+private:
+  const edm::EDGetTokenT<std::vector<pat::PackedCandidate> > pf_token_;
+  float norm_;
+  bool ignore_leptons_;
+  unsigned int max_n_pf_;
```
slava77, line 29: can be const

RecoMET/METPUSubtraction/plugins/DeepMETProducer.cc
```diff
@@ -0,0 +1,162 @@
...
+
+  tensorflow::Session* session_;
+
+  std::unordered_map<int, int32_t> charge_embedding_;
+  std::unordered_map<int, int32_t> pdg_id_embedding_;
```
slava77, line 34: these can be static const

RecoMET/METPUSubtraction/plugins/DeepMETProducer.cc
```diff
@@ -0,0 +1,162 @@
...
+    return ret_val;
+  }
+}  // namespace
+
+DeepMETProducer::DeepMETProducer(const edm::ParameterSet& cfg, const DeepMETCache* cache)
+    : pf_token_(consumes<std::vector<pat::PackedCandidate> >(cfg.getParameter<edm::InputTag>("pf_src"))),
+      norm_(cfg.getParameter<double>("norm_factor")),
+      ignore_leptons_(cfg.getParameter<bool>("ignore_leptons")),
+      max_n_pf_(cfg.getParameter<unsigned int>("max_n_pf")) {
+  session_ = tensorflow::createSession(cache->graph_def);
```
slava77, line 51: can be moved to the initializer list (before ```{```)

RecoMET/METPUSubtraction/plugins/DeepMETProducer.cc
```diff
@@ -0,0 +1,162 @@
...
+  tensorflow::Tensor input(tensorflow::DT_FLOAT, shape);
+  tensorflow::Tensor input_cat0(tensorflow::DT_FLOAT, cat_shape);
+  tensorflow::Tensor input_cat1(tensorflow::DT_FLOAT, cat_shape);
+  tensorflow::Tensor input_cat2(tensorflow::DT_FLOAT, cat_shape);
+
+  tensorflow::TensorShape out_shape({1, 2});
+  tensorflow::Tensor output(tensorflow::DT_FLOAT, out_shape);
+
+  tensorflow::NamedTensorList input_list = {
+      {"input", input}, {"input_cat0", input_cat0}, {"input_cat1", input_cat1}, {"input_cat2", input_cat2}};
```
slava77, line 75: static const?
Please check other values if can be made const, if not constexpr.

RecoMET/METPUSubtraction/plugins/DeepMETProducer.cc
```diff
@@ -0,0 +1,162 @@
...
+  pdg_id_embedding_ = {
+      {-211, 0}, {-13, 1}, {-11, 2}, {0, 3}, {1, 4}, {2, 5}, {11, 6}, {13, 7}, {22, 8}, {130, 9}, {211, 10}};
+  ;
+}
+
+void DeepMETProducer::produce(edm::Event& event, const edm::EventSetup& setup) {
+  edm::Handle<std::vector<pat::PackedCandidate> > pf_h;
+  event.getByToken(pf_token_, pf_h);
+
+  // PF keys [b'PF_dxy', b'PF_dz', b'PF_eta', b'PF_mass', b'PF_pt', b'PF_puppiWeight', b'PF_px', b'PF_py']
```
slava77, line 63: perhaps move this to the place where the 8 values are filled. It's more directly useful there

RecoMET/METPUSubtraction/plugins/DeepMETProducer.cc
```diff
@@ -0,0 +1,162 @@
...
+  produces<pat::METCollection>();
+  charge_embedding_ = {{-1, 0}, {0, 1}, {1, 2}};
+  pdg_id_embedding_ = {
+      {-211, 0}, {-13, 1}, {-11, 2}, {0, 3}, {1, 4}, {2, 5}, {11, 6}, {13, 7}, {22, 8}, {130, 9}, {211, 10}};
+  ;
+}
+
+void DeepMETProducer::produce(edm::Event& event, const edm::EventSetup& setup) {
+  edm::Handle<std::vector<pat::PackedCandidate> > pf_h;
+  event.getByToken(pf_token_, pf_h);
```
slava77, line 61:
```suggestion
  auto const& pfs = event.get(pf_token_);
```
is shorter

RecoMET/METPUSubtraction/plugins/DeepMETProducer.cc
```diff
@@ -0,0 +1,162 @@
...
+
+  // The DNN directly estimates the missing px and py, not the recoil
+  float px = outputs[0].tensor<float, 2>()(0, 0) * norm_;
+  float py = outputs[0].tensor<float, 2>()(0, 1) * norm_;
+
+  px -= px_leptons;
+  py -= py_leptons;
+
+  auto pf_mets = std::make_unique<pat::METCollection>();
+  reco::LeafCandidate::LorentzVector p4(px, py, 0., std::sqrt(px * px + py * py));
```
slava77, line 130:
```suggestion
  reco::Candidate::LorentzVector p4(px, py, 0., std::hypot(px, py));
```

RecoMET/METPUSubtraction/plugins/DeepMETProducer.cc
```diff
@@ -0,0 +1,162 @@
...
+  const reco::Candidate::Point vtx(0.0, 0.0, 0.0);
+  pf_mets->emplace_back(reco::MET(p4, vtx));
```
slava77, line 132: IIRC Point is default-filled with zero, this may even work:
```suggestion
  pf_mets->emplace_back(reco::MET(p4, {}));
```

PhysicsTools/PatAlgos/python/slimming/miniAOD_tools.py
steggema: Done.

RecoMET/METPUSubtraction/plugins/DeepMETProducer.cc
steggema: Done (using scale and renaming the function)

RecoMET/METPUSubtraction/plugins/DeepMETProducer.cc
steggema: Removed.

RecoMET/METPUSubtraction/plugins/DeepMETProducer.cc
steggema: Made const.

RecoMET/METPUSubtraction/plugins/DeepMETProducer.cc
steggema: Made static const and moved to inline initialisation.

RecoMET/METPUSubtraction/plugins/DeepMETProducer.cc
steggema: Done.

RecoMET/METPUSubtraction/plugins/DeepMETProducer.cc
steggema: Couldn't be static const because the tensors were defined and constructed in the local scope. I've now moved the tensors to the class level, with the advantage that this avoids reconstructing them in each event but will keep memory occupied (so I'm actually not sure it if's better or not).

RecoMET/METPUSubtraction/plugins/DeepMETProducer.cc
steggema: Done.

RecoMET/METPUSubtraction/plugins/DeepMETProducer.cc
steggema: Done.

RecoMET/METPUSubtraction/plugins/DeepMETProducer.cc
steggema: Done (supposedly zero-filled according to ROOT doc)

perrotta:
please  test

RecoMET/METPUSubtraction/plugins/DeepMETProducer.cc
slava77: ah, sorry, I was looking apparently at it and seeing only the format definition and missed the obvious second argument in each case, which are per-event.
Local defs should be fine as well as data members.

slava77:
The two kinds of METs are produced in this PR as ```reco::MET``` objects (not as ```pat::MET```). Considering the content (only px and py), ```reco::MET``` is more practical than the pat variant. 
Still, considering that the implementation really does not return anything more than px,py, would it be more practical to insert the DeepMETs  in slimmedMETsPuppi or even slimmedMETs in the same way how Calo and CHS METs are done?
OTOH, it seems like a longer term plan may be to have a fully fledged MET based on particle weights. So, perhaps a more complete object is still more meaningful.

@ahinzmann @lathomas was this discussed?

steggema:
> The two kinds of METs are produced in this PR as `reco::MET` objects (not as `pat::MET`). Considering the content (only px and py), `reco::MET` is more practical than the pat variant.

Just a quick correction (?): The producer currently yields pat::METs, https://github.com/cms-sw/cmssw/pull/29764/files#diff-63cabc35b8b22ca018a8e84322481671R58, but changing it to reco::MET would of course be a straightforward change. I have no preference and simply picked pat::MET such that the MiniAOD content looks more uniform.

PhysicsTools/PatAlgos/python/slimming/MicroEventContent_cff.py
```diff
@@ -20,6 +20,8 @@
         'keep *_slimmedMETs_*_*',
         'keep *_slimmedMETsNoHF_*_*',
         'keep *_slimmedMETsPuppi_*_*',
+        'keep *_slimmedMETsDeep_*_*',
+        'keep *_slimmedMETsDeepResp_*_*',
```
slava77, line 24: all other ```slimmedMET*``` objects have some more subtle structure and come from ```PATMETSlimmer```.
There is nothing similarly "slimmed" in the DeepMET case.
I'm thinking ```deepMETs``` for the former.
The "Resp" apparently stands for response-corrected. Is it the particle-level response or some more general MET response? I'm thinking that a more descriptive (perhaps longer) name would be useful.
Considering that "slimmed" can be omitted here, the longer replacement of "Resp" may not change the total length of the product.

To fit the MET object framework, the "Resp" case may need to enter as a correction to the ```deepMETs``` (that's unless I'm overloading the meaning of "correction" here).

slava77:
> > The two kinds of METs are produced in this PR as `reco::MET` objects (not as `pat::MET`). Considering the content (only px and py), `reco::MET` is more practical than the pat variant.
> 
> Just a quick correction (?): The producer currently yields pat::METs, https://github.com/cms-sw/cmssw/pull/29764/files#diff-63cabc35b8b22ca018a8e84322481671R58, but changing it to reco::MET would of course be a straightforward change. I have no preference and simply picked pat::MET such that the MiniAOD content looks more uniform.

ah, you are right, I was thinking of ```   pf_mets->emplace_back(reco::MET(p4, {}));```

PhysicsTools/PatAlgos/python/slimming/MicroEventContent_cff.py
steggema: > all other `slimmedMET*` objects have some more subtle structure and come from `PATMETSlimmer`.
> There is nothing similarly "slimmed" in the DeepMET case.
> I'm thinking `deepMETs` for the former.

Sounds good to me.

> The "Resp" apparently stands for response-corrected. Is it the particle-level response or some more general MET response? I'm thinking that a more descriptive (perhaps longer) name would be useful.
> Considering that "slimmed" can be omitted here, the longer replacement of "Resp" may not change the total length of the product.
> 
> To fit the MET object framework, the "Resp" case may need to enter as a correction to the `deepMETs` (that's unless I'm overloading the meaning of "correction" here).

"Response-corrected" is actually a bit of a misnomer: The two DeepMET versions have a different loss. The first version uses mean-squared-error loss, and has more optimal variance in the bias-variance trade-off and tends to be better for searches in MET resolution tails. The second version has an additional term in the loss yielding a better response and is such better-suited for analyses where unity response matters, in particular at low recoil. There is hence no correction applied. Would something like "deepMETsUnityResp" or "deepMETsUnityResponse" be more clear? (though this would technically only be true beyond some threshold)

slava77: I see. So, Resp is not really a modifier of the former. Perhaps a more balanced naming is better than.
Still a bit offhand, how about ```deepMETsResponseTune``` and ```deepMETsResolutionTune``` or even  shorter ```deepMETsForResponse``` and ```deepMETsForResolution``` ... or ```deepMETsMinTails``` and ```deepMETsMinBias```?

slava77:
>  Still, considering that the implementation really does not return anything more than px,py, would it be more practical to insert the DeepMETs in slimmedMETsPuppi or even slimmedMETs in the same way how Calo and CHS METs are done?

I had in mind this
https://github.com/cms-sw/cmssw/blob/424ad43856c2c739d702c240187401b7b08566ea/PhysicsTools/PatAlgos/plugins/PATMETSlimmer.cc#L80-L82

steggema:
> > Still, considering that the implementation really does not return anything more than px,py, would it be more practical to insert the DeepMETs in slimmedMETsPuppi or even slimmedMETs in the same way how Calo and CHS METs are done?
> 
> I had in mind this
> 
> https://github.com/cms-sw/cmssw/blob/424ad43856c2c739d702c240187401b7b08566ea/PhysicsTools/PatAlgos/plugins/PATMETSlimmer.cc#L80-L82

I've updated using this solution. Note that the PATMETSlimmer is apparently also always re-run in NanoAOD, so I switched adding the DeepMET variants off by default and only add them now in regular MiniAOD processing after a quick exchange with @peruzzim . Adding the DeepMET px and py values to NanoAOD will require a bit of fiddling with the according sequence in Nano.

slava77:
@cmsbuild please test

slava77:
> > > Still, considering that the implementation really does not return anything more than px,py, would it be more practical to insert the DeepMETs in slimmedMETsPuppi or even slimmedMETs in the same way how Calo and CHS METs are done?
> > 
> > 
> > I had in mind this
> > https://github.com/cms-sw/cmssw/blob/424ad43856c2c739d702c240187401b7b08566ea/PhysicsTools/PatAlgos/plugins/PATMETSlimmer.cc#L80-L82
> 
> I've updated using this solution. Note that the PATMETSlimmer is apparently also always re-run in NanoAOD, so I switched adding the DeepMET variants off by default and only add them now in regular MiniAOD processing after a quick exchange with @peruzzim . Adding the DeepMET px and py values to NanoAOD will require a bit of fiddling with the according sequence in Nano.

@ahinzmann @lathomas please clarify if this works OK for JME.
Thank you.

ahinzmann:
> > > > Still, considering that the implementation really does not return anything more than px,py, would it be more practical to insert the DeepMETs in slimmedMETsPuppi or even slimmedMETs in the same way how Calo and CHS METs are done?
> > > 
> > > 
> > > I had in mind this
> > > https://github.com/cms-sw/cmssw/blob/424ad43856c2c739d702c240187401b7b08566ea/PhysicsTools/PatAlgos/plugins/PATMETSlimmer.cc#L80-L82
> > 
> > 
> > I've updated using this solution. Note that the PATMETSlimmer is apparently also always re-run in NanoAOD, so I switched adding the DeepMET variants off by default and only add them now in regular MiniAOD processing after a quick exchange with @peruzzim . Adding the DeepMET px and py values to NanoAOD will require a bit of fiddling with the according sequence in Nano.
> 
> @ahinzmann @lathomas please clarify if this works OK for JME.
> Thank you.

Yes, that's fine. As long as the fiddling with the sequence for JME-extended-Nano is doable based on this PR for MiniAOD that's fine.

slava77:
+1

for https://github.com/cms-sw/cmssw/pull/29764 74e15da
- jenkins tests pass and comparisons with the baseline show differences only in the miniAOD workflows in the "corrections" of the ```slimmedMETs``` and ```slimmedMETsPuppi``` (similar to the way of storing trk, chs, and calo METs)

silviodonato:
merge
@santocch

santocch:
+1

Issue #29264, opened 2020-03-21
cleanup miniAOD workflow after moving PUPPI to reco


slava77:
PR #28936 moved PUPPI and some related jet modules to RECO, however puppi, ak4PFJetsPuppi and a few more modules are repeatedly running both in RECO and in PAT (when configured to run separately). Some simple tests show that most of it can be avoided https://github.com/cms-sw/cmssw/pull/28936#issuecomment-601484360.

By comparing the RECO and PAT steps in wf 11634.21 (production-like wf),  I see the following modules repeated, and related to PUPPI
- puppi
- ak4PFJetsPuppi, ak8PFJetsPuppi
- ak8PFJetsPuppiConstituents, ak8PFJetsPuppiSoftDrop, ak8PFJetsPuppiSoftDropMass

These should eventually be running only once to resolve/address this github issue.

Current matrix tests either run PAT step on newly produced AOD/RECO inputs or have specific process modifiers to run on old inputs (```run2_miniAOD_94XFall17``` and ```run2_miniAOD_80XLegacy```). For the existing workflows, we can then explicitly add puppi and related to miniAOD. For the possible future tests or workflows, the proposal is to add ```rerun_puppi``` modifier.

slava77:
assign reconstruction

slava77:
@ahinzmann

slava77:
+1
via #29254

PR #29748, opened 2020-05-06
Revert "PUPPI MET in RECO"

silviodonato:
See https://github.com/cms-sw/cmssw/issues/29744
We are getting errors related to #29254 in many workflows.
In addition, the errors seems hardly reproducible, and then we risk to crash the relval production.
So I'm going to revert temporarly #29254 before the night IB in order to build CMSSW_11_1_0_pre7
https://cmssdt.cern.ch/SDT/html/cmssdt-ib/#/ib/CMSSW_11_1_X

Issue #29744, opened 2020-05-06
Exception: Original collection and shifted collection are of different size in ShiftedParticleMETcorrInputProducer


silviodonato:
Several workflows (25211.18, 25403.0, 250205.0, 250208.18) are failing with the error
```
----- Begin Fatal Exception 06-May-2020 05:07:53 CEST-----------------------
An exception of category 'InvalidInput' occurred while
   [0] Processing  Event run: 1 lumi: 1 event: 8 stream: 3
   [1] Prefetching for module METTester/'pfPuppiMetAnalyzerMiniAOD'
   [2] Prefetching for module PATMETSlimmer/'slimmedMETsPuppi'
   [3] Prefetching for module CorrectedPATMETProducer/'patPFMetT1SmearJetResUpPuppi'
   [4] Calling method for module ShiftedParticleMETcorrInputProducer/'shiftedPatMETCorrSmearedJetResUpPuppi'
Exception Message:
Original collection and shifted collection are of different size in ShiftedParticleMETcorrInputProducer
----- End Fatal Exception -------------------------------------------------
```
Most likely, the error is related to https://github.com/cms-sw/cmssw/pull/29254 .

https://cmssdt.cern.ch/SDT/cgi-bin/logreader/slc7_amd64_gcc820/CMSSW_11_1_X_2020-05-05-2300/pyRelValMatrixLogs/run/25211.18_GluGluHToZZTo4L_M125_Pow_py8_Evt_13UP18+GluGluHToZZTo4L_M125_Pow_py8_Evt_13UP18+DIGIUP18_PU25+RECOUP18_PU25+HARVESTUP18_PU25/step3_GluGluHToZZTo4L_M125_Pow_py8_Evt_13UP18+GluGluHToZZTo4L_M125_Pow_py8_Evt_13UP18+DIGIUP18_PU25+RECOUP18_PU25+HARVESTUP18_PU25.log#/721-721

silviodonato:
assign reconstruction

silviodonato:
@ahinzmann

silviodonato:
Please note that in  `slc7_amd64_gcc900` fewer and different workflow are failing (`25212.18`). This sounds weird. We see the same effect also in  #29743.

ahinzmann:
Hi, sorry, I'm out of (home-)office this week, will look into it next week.

silviodonato:
I'm afraid we have to revert  #29743 from 11_1_0_pre7.
[...]

PR #29768, opened 2020-05-07
PUPPI MET in RECO

silviodonato:
Clone of #29254. #29254 was merged and reverted  with #29748 because of issue #29744 

#### PR description:

This PR computes puppiNoLep weights and puppi MET in RECO and stores the puppiNoLep weight as a value map. PUPPI MET is clustered based on standard PF candidates together with the value map of weights. A new boolean "isWeighted" is added to the reco::PFMet interface to indicate whether weights were applied during jet clustering. The packed candidates pick up the value map such that puppi does not need to be re-computed in MiniAOD.

The same was done for the puppi and PUPPI jets in #28936.

Other changes:
- Produce puppiForMET directly in PuppiProducer instead of a two step procedure puppiNoLep->puppiForMET and rename it to puppiNoLep, consistent with the name in PackedCandidates
- Implement fillDescriptions for PFMetProducer
- Rename PFMET_cfi to pfMet_cfi everywhere (for consistency with the module name)
- Fix input jet collection of pfMetEI
- Remove pfMetEI from RecoMET_EventContent (however, leave EITopPAG_EventContent_cff untouched)
- Resolve issue #29264
- By moving to the same input PF candidate collection for puppi MET and puppi jets (with just different weights), PUPPI MET significance and unclustered pT computation then don't rely on dR matching as implemented in #29331 anymore.
- Store ref to PF candidate inside pat muons (just like done for all other PAT objects). PUPPI MET significance and unclustered pT computation then don't rely on dR matching as implemented in #29385 anymore.
- Fix particle multiplicity computation in xy-correction for PUPPI MET. Previously, particles with a weight of 0 could also count in the multiplicity as a function of which the xy-corrections are computed. Since xy-correction are not needed/recommended for PUPPI MET, this fix is not backported.
- For HI workflows, raise candidate pT threshold for chsMet and trkMet to 999. consistent with pfMet.

#### PR validation:

runTheMatrix -l limited

No backport foreseen.


silviodonato:
please test workflow 25202.0, 25213.0, 250206.182, 25403.18, 250203.18, 250402.18, 11025.0, 11634.7, 25403.18, 250402.18, 11834.0, 25211.18, 25403.0, 250205.18
hoping at least one of them will catch the error

ahinzmann:
I think I understand the issue from this error.
For MET uncertainty due to jet energy resolution, jet energies are randomly smeared and apparently this may potentially change the number of jets above threshold.
Will look into it next week in detail.

silviodonato:
> I think I understand the issue from this error.
> For MET uncertainty due to jet energy resolution, jet energies are randomly smeared and apparently this may potentially change the number of jets above threshold.
> Will look into it next week in detail.

Very good. Then, we can close this PR.

PR #29778, opened 2020-05-08
[DD4hep] Bugfix for DDFilteredView::navPos

ianna:
#### PR description:

* Check if there is an iterator

@fabiocos - FYI

Hopefully it fixes the problem you've reported.

### DIFF
--- a/DetectorDescription/DDCMS/src/DDFilteredView.cc
+++ b/DetectorDescription/DDCMS/src/DDFilteredView.cc
@@ -547,10 +547,13 @@ std::string_view DDFilteredView::getString(const std::string& key) const {
 }
 
 DDFilteredView::nav_type DDFilteredView::navPos() const {
-  Int_t level = it_.back().GetLevel();
-  nav_type pos(level);
-  for (Int_t i = 1; i <= level; ++i)
-    pos[i] = it_.back().GetIndex(i);
+  nav_type pos;
+
+  if (not it_.empty()) {
+    int level = it_.back().GetLevel();
+    for (int i = 1; i <= level; ++i)
+      pos.emplace_back(it_.back().GetIndex(i));
+  }
 
   return pos;
 }


fabiocos:
please test

@ianna thanks! I'll do a test to whether that is enough to prevent the crashes I've observed

cvuosalo:
With this change, the caller needs to remember to check whether the returned `nav_type` vector is empty. How about an interface like:
```
bool DDFilteredView::navPos(DDFilteredView::nav_type &pos) const;
```
`navPos` would return false if `pos` ends up empty. This interface would remind the user that `pos` could be empty. I've just been fixing crashing bugs where developers assumed incorrectly that vectors were not empty, so I think it is good to make the interface remind the user of this danger.

fabiocos:
@ianna thanks, with this fix if I print ```navPos().size()``` in https://cmssdt.cern.ch/lxr/source/DetectorDescription/DDCMS/plugins/test/DDTestFilteredView.cc#0046  for the MTD geometry I do not get any more a crash. 

Concerning the comment of @cvuosalo I am not sure about the general sue of this method. In my case I was mostly checking its size, so this should be safe against your new implementation. Of course the real issue is why this vector could be empty, and if there are cases in the code where this is unexpected but it happens anyway

ianna:
> With this change, the caller needs to remember to check whether the returned `nav_type` vector is empty. How about an interface like:
> 
> ```
> bool DDFilteredView::navPos(DDFilteredView::nav_type &pos) const;
> ```
> 
> `navPos` would return false if `pos` ends up empty. This interface would remind the user that `pos` could be empty. I've just been fixing crashing bugs where developers assumed incorrectly that vectors were not empty, so I think it is good to make the interface remind the user of this danger.

I do not see an advantage in creating an additional variable to check a standard container state. In fact, I'd be promoting a range-based `for` loop iteration instead of using an `operator[]` where no bounds checking is performed. Since C++11 there is a mandatory elision of copy and move operations, so there is no advantage in passing a reference here either.

IMHO, current version:
```c++
auto myPos = navPos();
for(auto const& it : myPos) {
  check(it);
}
```
is cleaner then proposed:
```c++
DDFilteredView::nav_type myPos;
if(navPos(myPos)) {
  for(int i = 0; i < myPos.size(); i++) {
    check(navPos[i]);
  }
}

cvuosalo:
+1

cvuosalo:
@ianna OK, the interface is fine as-is. The next time `DDFilteredView.h` is edited, it would be good to mention in the comment for `navPos` that it could return an empty vector, just to remind the user not to blindly dereference it.

silviodonato:
+1

PR #29779, opened 2020-05-08
[LLVM10] Fix compilation warnings for RECO package

smuzaffar:
#### PR description:

Fixed loop variable copy warnings. Looks like we need a copy anyway as `hLasermon` returns `edm::DataFrame` and we need to convert it to `QIE10DataFrame`

#### PR validation:


slava77:
@cmsbuild please test

slava77:
+1

for https://github.com/cms-sw/cmssw/pull/29779	ff70528 
- technical
- there are no relevant differences in comparisons

silviodonato:
+1

PR #29781, opened 2020-05-08
Handle PseudoTraps in DD4hep MF geometry builder

namapane:
#### PR description:

Some legacy MF maps still use an old geometry that includes a few PseudoTrap shapes; notably, the  2T, which is in production and whose geometry is loaded into the DB. 
PseudoTraps are no longer supported in DD4hep, and for this reason these MF maps could not be created with the DD4hep-based MF builders.
This PR fixes this. It works by retrieving the vector of PseudoTrap parameters from the DDFilteredView without creating a DD4Hep shape object.

Recent and future geometries do not use PseudoTraps, so it is expected that it will be possible to drop support for it completely once all maps in production are updated to a recent geometry at some point in the future.

#### PR validation:
Checked for proper construction of the 2T geometry with DD4Hep against DDD, and regression-tested all maps in production against reference values.

### DIFF
--- a/MagneticField/GeomBuilder/src/DD4hep_volumeHandle.cc
+++ b/MagneticField/GeomBuilder/src/DD4hep_volumeHandle.cc
@@ -53,6 +53,10 @@ volumeHandle::volumeHandle(const cms::DDFilteredView &fv, bool expand2Pi, bool d
     case DDSolidShape::ddtubs:
       buildTubs();
       break;
+    case DDSolidShape::ddpseudotrap: {
+      vector<double> d = solid.volume().volume().solid().dimensions();
+      buildPseudoTrap(d[0], d[1], d[2], d[3], d[4], d[5], d[6]);
+    } break;
     case DDSolidShape::ddtrunctubs:
       buildTruncTubs();
       break;
@@ -148,11 +152,10 @@ void volumeHandle::referencePlane(const cms::DDFilteredView &fv) {
     // See comments above for the conventions for orientation.
     LocalVector globalZdir(0., 0., 1.);  // Local direction of the axis along global Z
 
-    /* Preserve in case pseudotrap is needed again
     if (theShape == DDSolidShape::ddpseudotrap) {
       globalZdir = LocalVector(0., 1., 0.);
     }
-    */
+
     if (refPlane->toGlobal(globalZdir).z() < 0.) {
       globalZdir = -globalZdir;
     }
@@ -198,4 +201,5 @@ using namespace cms::dd;
 #include "buildTrap.icc"
 #include "buildTubs.icc"
 #include "buildCons.icc"
+#include "buildPseudoTrap.icc"
 #include "buildTruncTubs.icc"
--- a/MagneticField/GeomBuilder/src/DD4hep_volumeHandle.h
+++ b/MagneticField/GeomBuilder/src/DD4hep_volumeHandle.h
@@ -47,6 +47,9 @@ namespace magneticfield {
     void buildTubs();
     // Build the surfaces for a ddcons shape
     void buildCons();
+    // Build the surfaces for a ddpseudotrap. This is not a supported
+    // shape in DD4hep; it is handled here to cope with legacy geometries.
+    void buildPseudoTrap(double x1, double x2, double y1, double y2, double halfZ, double radius, bool atMinusZ);
     // Build the surfaces for a ddtrunctubs shape
     void buildTruncTubs();
 
--- a/MagneticField/GeomBuilder/src/MagGeoBuilderFromDDD.cc
+++ b/MagneticField/GeomBuilder/src/MagGeoBuilderFromDDD.cc
@@ -186,7 +186,7 @@ void MagGeoBuilderFromDDD::build(const DDCompactView& cpva) {
     //       }
     //     }
 
-    volumeHandle* v = new volumeHandle(fv, expand);
+    volumeHandle* v = new volumeHandle(fv, expand, debug);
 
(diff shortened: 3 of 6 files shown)


namapane:
@cvuosalo , you may want have a look and possibly comment.

slava77:
@cmsbuild please test

MagneticField/GeomBuilder/src/DD4hep_volumeHandle.cc
```diff
@@ -53,6 +53,10 @@ volumeHandle::volumeHandle(const cms::DDFilteredView &fv, bool expand2Pi, bool d
     case DDSolidShape::ddtubs:
       buildTubs();
       break;
+    case DDSolidShape::ddpseudotrap: {
+      vector<double> d = solid.volume().volume().solid().dimensions();
```
cvuosalo, line 57:
```suggestion
      vector<double> d = solid.parameters();
```

cvuosalo: Let's avoid internal DD4hep implementation details by using the `parameters` method.

cvuosalo:
@namapane What is the future of these legacy maps that use pseduotrapezoids? Are they needed to support Run 1 or 2? We are planning to nominally support Runs 1-2 with DD4hep going forward, so will we have to continue supporting these legacy maps indefinitely?

MagneticField/GeomBuilder/src/DD4hep_volumeHandle.cc
namapane: `solid.volume().volume().solid().dimensions() `is exactly the workaround I had to figure out to avoid the crash that otherwise happens when asking parameters() on a PseudoTrap.
For the records, I've debugged the crash to happen after this if evaluates to true:
https://github.com/cms-sw/cmssw/blob/master/DetectorDescription/DDCMS/src/DDFilteredView.cc#L452

in fact, for some reason beyond my understanding, currVol->GetShape()->IsA() turns out to be TGeoCompositeShape::Class() for a PseudoTrap. Then the code to handle a composite object is executed, and this leads to something unexpected. 
I verified that there would be no crash if that if() evaluated to false, but to me there is no way to add a protection without adding some specific condition for PseudoTraps, which I guess we want to avoid here.

So I figured out that it is a better idea of putting a small workaround in MF, which is in a very closed scope, than to hack DDFilteredView for a very specific use case.

Of course if you know how to fix DDFilteredView so that it gives dimensions() for PseudoTraps without crashing, that would be welcome.
I was con

namapane:
> 
> 
> @namapane What is the future of these legacy maps that use pseduotrapezoids? Are they needed to support Run 1 or 2? We are planning to nominally support Runs 1-2 with DD4hep going forward, so will we have to continue supporting these legacy maps indefinitely?

This geometry is practically only used for the 2T map. That is not relevant for Run I since we have nothing useful collected at 2T (if anything at all); it is also very hypotetical for Run II since we hope we never need to collect any data at 2T. However, we can't exclude it will never be needed, and the map must be ready for such a scenario.

Updating the 2T map has never been a priority, because a scenario where need it for real is so desperate that the improvements of the updates were absolutely not worth the effort, nor could justify the additional amount of data to be added to externals. We decided to update 3T and 3.5T (besides 3.8T) and that was possibly already an overkill in terms of amount of data.

Now that keeping the old geometry is an issue, I think we will consider updating the 2T map as well the next time all other maps are updated. But such an update is not immediately needed so I don't expect it to come in a short timescale. In the meanwhile, I think that this fix should be just enough.

MagneticField/GeomBuilder/src/DD4hep_volumeHandle.cc
cvuosalo: We have two issues here. Line 57 is vulnerable to any internal change to DD4hep. I can tell you from experience that DD4hep changes with some frequency, even in its external interface, so I would worry that some time years from now, the DD4hep internals change to break this line, and we will have all forgotten about the special pseudotrapezoids and will have a hard time figuring out how to fix the problem.

cvuosalo: The second issue is that a `DDFilteredView` method should never crash, regardless of what shape is involved. Yana and I need to fix `parameters` to make it more robust.
@ianna FYI `DDFilteredView::parameters` crashes with a pseudotrapezoid.

cvuosalo: This PR can go ahead while we figure out how to deal with these issues. Use the version of line 57 that you found works.

cvuosalo:
+1

MagneticField/GeomBuilder/src/DD4hep_volumeHandle.cc
namapane: @cvuosalo, I fully agree. This fix is intended to keep legacy maps working while we have no better solution; it is intended that it might need to be adapted if the underlying interface changes.
A fix to DDFilteredView::parameters() would be very useful for this purpose. I spent some time on it but could not find an acceptable fix myself. I can produce a simple recipe to reproduce the crash if that can help others to work on it.

slava77:
+1

for https://github.com/cms-sw/cmssw/pull/29781	80edf9c 
- signing based mostly on visual inspection and review by @cvuosalo 
- jenkins tests pass and comparisons with the baseline show no relevant differences (this PR is technical wrt these tests)

silviodonato:
+1

PR #29783, opened 2020-05-08
Phase2 Updated Pixel Conditions and GenericCPE Seeding Errors Fix

OzAmram:
This is the PR intending to fix the loss in the phase 2 tracking performance (efficiency and fakerate) since the 11_1_0 and later releases.

We presented about the ongoing issues in several places, most recently at the Tracking POG meeting [here](https://indico.cern.ch/event/912823/contributions/3841072/attachments/2027324/3391939/phase2_det_angle_update_april26.pdf)

The PR contains updated template and genError conditions that fix a slight Lorentz Angle mismatch that was introduced in the last conditions.

It also introduces a new configuration option for the generic CPE which allows one to use the hardcoded errors when the CPE is called without local track angles (ie in the seeding step) the database values (genErrors) when called with local track angles. This option is turned on only for the phase 2 configuration of the generic CPE. 

There should be no changes in any phase 1 workflows. 

We have run the tracking validation comparing the performance of the current CMSSW, the performance before the introduction of the templates for phase 2 and this proposed fix. The plots are [here](http://cmantill.web.cern.ch/cmantill/phase2/PU200_ComparisonCMSSW_11_1_0_pre6/plots/). We see that this fix does indeed fix the efficiency loss and fakerate increase seen in the current release as compared to the 11_0_0 releases. 
There is still an increase in the chi^2 observed for eta > 1.5 or so as compared to the 11_0_0 releases. This has been traced back to the templates having a smaller error for FPIX local x than the hardcoded values, but looking at the pull distributions it seems this is actually the more correct error (RMS with template is ~1 and with hardcoded errors ~0.75). It is believed that the reason the chi^/2dof is larger than 1 is due to a known issue, that the material budget is underestimated, not the CPE's. 

@mmusich @tsusa @mtosi @emiglior @cmantill @tvami @pmaksim1 @JanFSchulte @jalimena @perrotta @slava77

### DIFF
--- a/RecoLocalTracker/SiPixelRecHits/src/PixelCPEGeneric.cc
+++ b/RecoLocalTracker/SiPixelRecHits/src/PixelCPEGeneric.cc
@@ -641,7 +641,7 @@ void PixelCPEGeneric::fillPSetDescription(edm::ParameterSetDescription& desc) {
   desc.add<double>("EdgeClusterErrorX", 50.0);
   desc.add<double>("EdgeClusterErrorY", 85.0);
   desc.add<bool>("inflate_errors", false);
-  desc.add<bool>("inflate_all_errors_no_trk_angle", true);
+  desc.add<bool>("inflate_all_errors_no_trk_angle", false);
   desc.add<bool>("NoTemplateErrorsWhenNoTrkAngles", false);
   desc.add<bool>("UseErrorsFromTemplates", true);
   desc.add<bool>("TruncatePixelCharge", true);
--- a/RecoLocalTracker/SiPixelRecHits/src/PixelCPEGeneric.cc
+++ b/RecoLocalTracker/SiPixelRecHits/src/PixelCPEGeneric.cc
@@ -59,15 +59,11 @@ PixelCPEGeneric::PixelCPEGeneric(edm::ParameterSet const& conf,
   TruncatePixelCharge_ = conf.getParameter<bool>("TruncatePixelCharge");
   IrradiationBiasCorrection_ = conf.getParameter<bool>("IrradiationBiasCorrection");
   DoCosmics_ = conf.getParameter<bool>("DoCosmics");
-  //LoadTemplatesFromDB_       = conf.getParameter<bool>("LoadTemplatesFromDB");
 
   // Upgrade means phase 2
-  isUpgrade_ = false;
-  if (conf.getParameter<bool>("Upgrade"))
-    isUpgrade_ = true;
+  isUpgrade_ = conf.getParameter<bool>("Upgrade");
 
-  // Select the position error source
-  // For upgrde and cosmics force the use simple errors
+  // For cosmics force the use of simple errors
   if ((DoCosmics_))
     UseErrorsFromTemplates_ = false;
 
--- a/RecoLocalTracker/SiPixelRecHits/src/PixelCPEGeneric.cc
+++ b/RecoLocalTracker/SiPixelRecHits/src/PixelCPEGeneric.cc
@@ -54,20 +54,21 @@ PixelCPEGeneric::PixelCPEGeneric(edm::ParameterSet const& conf,
   inflate_errors = conf.getParameter<bool>("inflate_errors");
   inflate_all_errors_no_trk_angle = conf.getParameter<bool>("inflate_all_errors_no_trk_angle");
 
+  NoTemplateErrorsWhenNoTrkAngles_ = conf.getParameter<bool>("NoTemplateErrorsWhenNoTrkAngles");
   UseErrorsFromTemplates_ = conf.getParameter<bool>("UseErrorsFromTemplates");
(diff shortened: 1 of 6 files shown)


perrotta:
please test

RecoLocalTracker/SiPixelRecHits/src/PixelCPEGeneric.cc
```diff
@@ -54,20 +54,21 @@ PixelCPEGeneric::PixelCPEGeneric(edm::ParameterSet const& conf,
...
+  NoTemplateErrorsWhenNoTrkAngles_ = conf.getParameter<bool>("NoTemplateErrorsWhenNoTrkAngles");
   UseErrorsFromTemplates_ = conf.getParameter<bool>("UseErrorsFromTemplates");
   TruncatePixelCharge_ = conf.getParameter<bool>("TruncatePixelCharge");
   IrradiationBiasCorrection_ = conf.getParameter<bool>("IrradiationBiasCorrection");
   DoCosmics_ = conf.getParameter<bool>("DoCosmics");
   //LoadTemplatesFromDB_       = conf.getParameter<bool>("LoadTemplatesFromDB");
 
-  // no clear what upgrade means, is it phase1, phase2? Probably delete.
+  // Upgrade means phase 2
   isUpgrade_ = false;
```
perrotta, line 65: default for ```isUpgrade_``` is already taken into account via the ```fillDescriptions``` method:
```C++
  isUpgrade_ = conf.getParameter<bool>("Upgrade");
```

RecoLocalTracker/SiPixelRecHits/src/PixelCPEGeneric.cc
```diff
@@ -54,20 +54,21 @@ PixelCPEGeneric::PixelCPEGeneric(edm::ParameterSet const& conf,
...
   if (conf.getParameter<bool>("Upgrade"))
     isUpgrade_ = true;
 
   // Select the position error source
   // For upgrde and cosmics force the use simple errors
```
perrotta, line 70: Since you changed the selections, please also modify the comment

RecoLocalTracker/SiPixelRecHits/src/PixelCPEGeneric.cc
```diff
@@ -54,20 +54,21 @@ PixelCPEGeneric::PixelCPEGeneric(edm::ParameterSet const& conf,
   inflate_errors = conf.getParameter<bool>("inflate_errors");
   inflate_all_errors_no_trk_angle = conf.getParameter<bool>("inflate_all_errors_no_trk_angle");
 
+  NoTemplateErrorsWhenNoTrkAngles_ = conf.getParameter<bool>("NoTemplateErrorsWhenNoTrkAngles");
   UseErrorsFromTemplates_ = conf.getParameter<bool>("UseErrorsFromTemplates");
   TruncatePixelCharge_ = conf.getParameter<bool>("TruncatePixelCharge");
   IrradiationBiasCorrection_ = conf.getParameter<bool>("IrradiationBiasCorrection");
   DoCosmics_ = conf.getParameter<bool>("DoCosmics");
   //LoadTemplatesFromDB_       = conf.getParameter<bool>("LoadTemplatesFromDB");
```
perrotta, line 62: Perhaps it is time now to remove this commented out line, as well as the corresponding commented out definition of ```LoadTemplatesFromDB_``` in the interface: that class member is already defined and set in ```PixelCPEBase```, from which this class derives

perrotta:
assign upgrade

RecoLocalTracker/SiPixelRecHits/src/PixelCPEGeneric.cc
OzAmram: done

RecoLocalTracker/SiPixelRecHits/src/PixelCPEGeneric.cc
OzAmram: done

RecoLocalTracker/SiPixelRecHits/src/PixelCPEGeneric.cc
OzAmram: done

perrotta:
please test

kpedro88:
+upgrade
comparison changes isolated to phase 2 workflows, as expected
minor fluctuations in tracks and downstream quantities:
log10 generalTracks pt

perrotta:
+1
- Code and conditions updates originate minor changes in tracks and related quantities for Phase2 worrkflows
- Jenkins tests pass and show no differences for non-Phase2 workflows

tlampen:
+1

silviodonato:
@OzAmram  please update the title of the pull request to something more descriptive

silviodonato:
+1

PR #29788, opened 2020-05-09
adapt tau validation for miniAOD-only workflows

slava77:
this PR was initially triggered by observation of numerous printouts in re-miniAOD workflows like
```
 Reference collection not found while running TauValidationMiniAOD.cc
```
This PR replaces cerr (not allowed for production/centrally running code) with a LogWarning.
It looks like the fix to the underlying issue is simple, to add the ```produceDenoms``` to the prevalidationMiniAOD.


slava77:
@cmsbuild please test

jfernan2:
+1

silviodonato:
+operations

silviodonato:
+1

PR #29622, opened 2020-05-04
Improvements for Payload inspector

ggovi:
#### PR description:

Payload Inspector Interface generalised to support combinations of iov-policies and tag multiplicity 

#### PR validation:

- All the System-specific unit tests executed; picture plot files compared with previous release after generation
- BasicPayload plots generated in stand-alone running the getPayloadData script
- Current branch deployed as the DEMO release  in the Payload Inspector application

PR #29789, opened 2020-05-09
[LLVM10] Fix compilation warnings from CondCore

smuzaffar:
#### PR description:

Integration of https://github.com/cms-sw/cmssw/pull/29622 introducs new warnings in CLANG IB [a]. This PR explicitly adds `using Baseclase::fill;` to avoid this warnings.

FYI, @ggovi , please suggest if there is any better/simple way to avoid such warning.

[a]
```
CondCore/AlignmentPlugins/plugins/TrackerAlignmentErrorExtended_PayloadInspector.cc:70:10: warning: '(anonymous namespace)::TrackerAlignmentErrorExtendedValue<AlignmentPI::YY>::fill' hides overloaded virtual function [-Woverloaded-virtual]
     bool fill(const std::vector<std::tuple<cond::Time_t, cond::Hash> >& iovs) override {
CondCore/Utilities/interface/PayloadInspector.h:724:12: note: hidden overloaded virtual function 'cond::payloadInspector::Histogram1D<AlignmentErrorsExtended, cond::payloadInspector::UNSPECIFIED_IOV>::fill' declared here: different number of parameters (0 vs 1)
      bool fill() override {
```
#### PR validation:

### DIFF
--- a/CondCore/AlignmentPlugins/plugins/TrackerAlignmentErrorExtended_PayloadInspector.cc
+++ b/CondCore/AlignmentPlugins/plugins/TrackerAlignmentErrorExtended_PayloadInspector.cc
@@ -67,6 +67,7 @@ namespace {
       Base::setSingleIov(true);
     }
 
+    using cond::payloadInspector::Histogram1D<AlignmentErrorsExtended>::fill;
     bool fill(const std::vector<std::tuple<cond::Time_t, cond::Hash> >& iovs) override {
       for (auto const& iov : iovs) {
         std::shared_ptr<AlignmentErrorsExtended> payload = Base::fetchPayload(std::get<1>(iov));
--- a/CondCore/AlignmentPlugins/plugins/TrackerSurfaceDeformations_PayloadInspector.cc
+++ b/CondCore/AlignmentPlugins/plugins/TrackerSurfaceDeformations_PayloadInspector.cc
@@ -52,6 +52,7 @@ namespace {
       Base::setSingleIov(true);
     }
 
+    using cond::payloadInspector::Histogram1D<AlignmentSurfaceDeformations>::fill;
     bool fill(const std::vector<std::tuple<cond::Time_t, cond::Hash> > &iovs) override {
       for (auto const &iov : iovs) {
         std::shared_ptr<AlignmentSurfaceDeformations> payload = Base::fetchPayload(std::get<1>(iov));
--- a/CondCore/EcalPlugins/plugins/EcalIntercalibConstantsMC_PayloadInspector.cc
+++ b/CondCore/EcalPlugins/plugins/EcalIntercalibConstantsMC_PayloadInspector.cc
@@ -56,6 +56,7 @@ namespace {
       Base::setSingleIov(true);
     }
 
+    using cond::payloadInspector::Histogram2D<EcalIntercalibConstantsMC>::fill;
     // Histogram2D::fill (virtual) needs be overridden - the implementation should use fillWithValue
     bool fill(const std::vector<std::tuple<cond::Time_t, cond::Hash> >& iovs) override {
       for (auto const& iov : iovs) {
@@ -115,6 +116,7 @@ namespace {
       Base::setSingleIov(true);
     }
 
+    using cond::payloadInspector::Histogram2D<EcalIntercalibConstantsMC>::fill;
     bool fill(const std::vector<std::tuple<cond::Time_t, cond::Hash> >& iovs) override {
       for (auto const& iov : iovs) {
         std::shared_ptr<EcalIntercalibConstantsMC> payload = Base::fetchPayload(std::get<1>(iov));
--- a/CondCore/EcalPlugins/plugins/EcalIntercalibConstants_PayloadInspector.cc
+++ b/CondCore/EcalPlugins/plugins/EcalIntercalibConstants_PayloadInspector.cc
@@ -56,6 +56,7 @@ namespace {
       Base::setSingleIov(true);
     }
 
+    using cond::payloadInspector::Histogram2D<EcalIntercalibConstants>::fill;
     // Histogram2D::fill (virtual) needs be overridden - the implementation should use fillWithValue
     bool fill(const std::vector<std::tuple<cond::Time_t, cond::Hash> >& iovs) override {
       for (auto const& iov : iovs) {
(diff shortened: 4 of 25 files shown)


smuzaffar:
please test

CondCore/AlignmentPlugins/plugins/TrackerAlignmentErrorExtended_PayloadInspector.cc
```diff
@@ -66,7 +66,7 @@ namespace {
               500.0) {
       Base::setSingleIov(true);
     }
-
+    using cond::payloadInspector::Histogram1D<AlignmentErrorsExtended>::fill;
```
mmusich, line 69: I don't understand the pattern here, why is it needed here and for instance not at:
https://github.com/cms-sw/cmssw/blob/f039d3d393d17765248bc5ec0502738bce223da7/CondCore/AlignmentPlugins/plugins/TrackerAlignmentErrorExtended_PayloadInspector.cc#L118

smuzaffar: No idea why compiler did not complain about few other classes in this file. May be because Histogram1D explicitly implements `bool fill()` ( https://github.com/cms-sw/cmssw/blob/master/CondCore/Utilities/interface/PayloadInspector.h#L724 ) while `PlotImage` does not ( https://github.com/cms-sw/cmssw/blob/master/CondCore/Utilities/interface/PayloadInspector.h#L829 )

mmusich: but e.g. this other plot in which you did add the `using ...` is inheriting from `PlotImage`:
https://github.com/cms-sw/cmssw/pull/29789/files#diff-25914482e3ee9c4f2fcefeac3d8dfd30R96.
Said in other way, I need to implement more of these plots, how do I decide where it is appropriate? How does one reproduce the warning?

smuzaffar: you can create clang base IB dev rea e.g. `scram p CMSSW_11_1_CLANG_X_2020-05-10-2300` and then checkout the packages and run `scram build`. I think the reason is that few classes in https://github.com/cms-sw/cmssw/blob/master/CondCore/Utilities/interface/PayloadInspector.h do not have 
```
virtual bool fill(const std::vector<std::tuple<cond::Time_t, cond::Hash>>& iovs) { return false; }
```
so may be proper fix would be to add these in base classes ?

ggovi: @mmusich 
The final goal of the changes triggering the CLANG compiler errors is to remove the fill(vector<...>) method in favour of the fill() method. That should be the guideline for new implementations. Concerning this PR, if we are happy I'd propose to approuve it.

smuzaffar: @mmusich , yes looks like in few places the `using ` for `PlotImage` not needed. I will update the PR

smuzaffar: PR is updated now to only fix those classes which are inherited from using `cond::payloadInspector::Histogram*`

mmusich: @smuzaffar, I had a deeper look to the base class and the method 
```
virtual bool fill(const std::vector<std::tuple<cond::Time_t, cond::Hash>>& iovs) { return false; }
```
has totally disappeared from `cond::payloadInspector::Histogram*`. 
So the proper fix would be to migrate all the derive classes to use instead 
https://github.com/cms-sw/cmssw/blob/29634922c859dc7a7cec988de55ee4ff48becb6c/CondCore/Utilities/interface/PayloadInspector.h#L724-L734

at the moment all of these plots are not functional !
See 
   * normal behaviour in [CMSSW_11_1_0_pre7](https://cms-conddb.cern.ch/cmsDbBrowser/payload_inspector/Prod/CMSSW_11_1_0_pre7/eyJwbG90cyI6MSwicGxvdDEiOnsidGFnIjoiVHJhY2tlckFsaWdubWVudEVycm9yRXh0ZW5kZWRfUnVuMlN0YXJ0dXBfdjBfbWMiLCJwbG90IjoicGxvdF9UcmFja2VyQWxpZ25tZW50RXJyb3JFeHRlbmRlZFhYVmFsdWUiLCJwbHVnaW4iOiJwbHVnaW5UcmFja2VyQWxpZ25tZW50RXJyb3JFeHRlbmRlZF9QYXlsb2FkSW5zcGVjdG9yIiwidHlwZSI6Ikhpc3RvMUQiLCJpb3ZzIjp7InN0YXJ0X2lvdiI6IjEiLCJlbmRfaW92IjoiMSJ9fX0=) (working)
image

   * not functional behaviour in the [latest IB](https://cms-conddb.cern.ch/cmsDbBrowser/payload_inspector/Prod/CMSSW_11_1_X_2020-05-11-2300/eyJwbG90cyI6MSwicGxvdDEiOnsidGFnIjoiVHJhY2tlckFsaWdubWVudEVycm9yRXh0ZW5kZWRfUnVuMlN0YXJ0dXBfdjBfbWMiLCJwbG90IjoicGxvdF9UcmFja2VyQWxpZ25tZW50RXJyb3JFeHRlbmRlZFhYVmFsdWUiLCJwbHVnaW4iOiJwbHVnaW5UcmFja2VyQWxpZ25tZW50RXJyb3JFeHRlbmRlZF9QYXlsb2FkSW5zcGVjdG9yIiwidHlwZSI6Ikhpc3RvMUQiLCJpb3ZzIjp7InN0YXJ0X2lvdiI6IjEiLCJlbmRfaW92IjoiMSJ9fX0=): 
image

@ggovi

smuzaffar: thanks @mmusich for looking in to this. Can you or @ggovi please provide a proper fix (in a new PR)?

mmusich: @smuzaffar yes, I'll prepare a fix.

mmusich: For the record I've prepared PR https://github.com/cms-sw/cmssw/pull/29817

smuzaffar:
closing in favor of https://github.com/cms-sw/cmssw/pull/29817

PR #29792, opened 2020-05-09
Have EcalSimParametersESModule only consume what is needed

Dr15Jones:
#### PR description:

At configuration time we know which of the two data products we will actually be consuming. Only call `consumesFrom` on that one. This avoids trying to do a prefetch on something we know we will not actually want.

#### PR validation:

The code compiles.


Dr15Jones:
please test

cvuosalo:
+1

cvuosalo:
@Dr15Jones Thanks for noticing this problem and fixing it.

silviodonato:
+1

PR #29797, opened 2020-05-11
Add TICLPFValidation sequence

rovere:
#### PR description:

Now that TICL is part of the Phase2 reconstruction sequence, we activate
the corresponding validation module to monitor the performance of the
PFCandidates produces by TICL.

#### PR validation:

`runTheMatrix.py -l limited`

### DIFF
--- a/Validation/Configuration/python/hgcalSimValid_cff.py
+++ b/Validation/Configuration/python/hgcalSimValid_cff.py
@@ -8,6 +8,9 @@ from Validation.HGCalValidation.hgcalHitValidation_cfi  import *
 from Validation.HGCalValidation.HGCalValidator_cfi import hgcalValidator
 from Validation.RecoParticleFlow.PFJetValidation_cff import pfJetValidation1 as _hgcalPFJetValidation
 
+from Validation.HGCalValidation.ticlPFValidation_cfi import ticlPFValidation
+hgcalTiclPFValidation = cms.Sequence(ticlPFValidation)
+
 hgcalValidatorSequence = cms.Sequence(hgcalValidator)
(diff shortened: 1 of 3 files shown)


rovere:
please test

jfernan2:
@rovere is it expected that folders 0,3,6,7 have empty histograms? I am not sure about the simple number naming, may I suggest a more meaningful name?

rovere:
@jfernan2 not on a regular basis, but depending on the sample and on the quality of our reconstruction, which is still heavily under development, it may happen.
More specifically:

* category 0 means undefined: 0 entries is good in this case.
* category 3 means muons: in this case, it could be that samples used had no muons. This is also a case in which, possibly, the reconstruction is failing.
* category 6 is h_HF, which is missing from the Phase2 simulation, yet it is present in the definition of a PFCandidate. In case it will be re-used for other purposes in future, we included that, too.

jfernan2:
Thanks @rovere 
Could these categories-numbering scheme be documented in the code somehow (if not already) for future coding references?

rovere:
Ciao @jfernan2 
those numbers are related to the `PFCandidate` internal representation and I'd not like to have them documented in `DQM` code. At the same time, since these numbers are coming from an `enum` type, I preferred to stick to numbers rather than hard-coding labels that could out-of-sync wrt their original meaning.
I'd say that the definitive documentation is [this code](https://github.com/cms-sw/cmssw/blob/master/DataFormats/ParticleFlowCandidate/interface/PFCandidate.h), which is already in release.

jfernan2:
+1

civanch:
+1

silviodonato:
+1

PR #29799, opened 2020-05-11
Fix features order

rovere:
#### PR description:

In recent meetings at the Upgrade HLT TDR, strange distribution of PFCandidates produced from TICL have been reported: [JetMET](https://indico.cern.ch/event/909498/contributions/3843607/attachments/2028099/3393510/200428_hltupg_jme.pdf) [BTV](https://indico.cern.ch/event/909499/contributions/3826040/subcontributions/304057/attachments/2032345/3401691/05_05_20_HLTUpgradeBTV.pdf).
After some debugging, this has been traced back to the way in which we pass features to ML model for PID and energy regression. This PR reshuffle them in the correct order. Local tests show no more dependency on phi.

#### PR validation:

Besides checking that the strange phi behaviour is gone:

`runTheMatrix.py -l limited`

### DIFF
--- a/RecoHGCal/TICL/plugins/PatternRecognitionbyCA.cc
+++ b/RecoHGCal/TICL/plugins/PatternRecognitionbyCA.cc
@@ -333,7 +333,7 @@ void PatternRecognitionbyCA::energyRegressionAndID(const std::vector<reco::CaloC
         float *features = &input.tensor<float, 4>()(i, j, seenClusters[j], 0);
 
         // fill features
-        *(features++) = float(cluster.energy()/float(trackster.vertex_multiplicity(k)));
+        *(features++) = float(cluster.energy() / float(trackster.vertex_multiplicity(k)));
         *(features++) = float(std::abs(cluster.eta()));
         *(features) = float(cluster.phi());
 
--- a/RecoHGCal/TICL/plugins/PatternRecognitionbyCA.cc
+++ b/RecoHGCal/TICL/plugins/PatternRecognitionbyCA.cc
@@ -333,7 +333,7 @@ void PatternRecognitionbyCA::energyRegressionAndID(const std::vector<reco::CaloC
         float *features = &input.tensor<float, 4>()(i, j, seenClusters[j], 0);
 
         // fill features
-        *(features++) = float(cluster.energy());
+        *(features++) = float(cluster.energy()/float(trackster.vertex_multiplicity(k)));
         *(features++) = float(std::abs(cluster.eta()));
         *(features) = float(cluster.phi());
 
--- a/RecoHGCal/TICL/plugins/PatternRecognitionbyCA.cc
+++ b/RecoHGCal/TICL/plugins/PatternRecognitionbyCA.cc
@@ -333,9 +333,9 @@ void PatternRecognitionbyCA::energyRegressionAndID(const std::vector<reco::CaloC
         float *features = &input.tensor<float, 4>()(i, j, seenClusters[j], 0);
 
         // fill features
+        *(features++) = float(cluster.energy());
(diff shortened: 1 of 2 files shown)


rovere:
please test

kpedro88:
Comparison differences are limited to Phase 2 workflows, as expected.

@rovere I notice a few potentially surprising changes. Please confirm if these are expected:
* overall slightly fewer objects reconstructed:
    trackster eta
* many fewer EM objects:
    tracksterEM energy
* corresponding increase in HAD objects:
    tracksterHAD energy
* narrower distribution of MIP regressed energy:
    tracksterMIP energy

rovere:
Ciao @kpedro88 thanks for checking. The problem is that the comparison, in this case, does not make too much sense, since the ML inference was wrong before and, as a consequence, all the results produced in terms of PID and energy regression were unreliable. 
Rephrased in another way: if POGs report mismatches and poor performances related to PID and energy regression, we will follow them up and investigate. We are also in the process of retraining the model using the most up-to-date TICL/trackster reconstruction that was not available at the time of the original training included in CMSSW.

RecoHGCal/TICL/plugins/PatternRecognitionbyCA.cc
```diff
@@ -333,9 +333,9 @@ void PatternRecognitionbyCA::energyRegressionAndID(const std::vector<reco::CaloC
         float *features = &input.tensor<float, 4>()(i, j, seenClusters[j], 0);
 
         // fill features
```
fwyzard, line 335: This "interface" is extremely error prone (as this bug just demonstrated).
Whoever approved this code for CMSSW clearly was not doing their job.

fwyzard: Looks like the original sin is #27917.

What are the plans to fix the use of anonymous, unordered list for passing parameters to tensorflow or other ML engines ?

kpedro88: @fwyzard this is a good point. Maybe the framework should add a data structure like `edm::featureMap` that enforces a schema and has interfaces to output in various ML formats.

(Though, this would only prevent one of the two problems addressed by this PR. No framework feature can specify whether or not energy should be divided by vertex multiplicity...)

fwyzard: > (Though, this would only prevent one of the two problems addressed by this PR. No framework feature can specify whether or not energy should be divided by vertex multiplicity...)

:-)

kpedro88:
@rovere I take the point that the "old" results are just wrong. However, can you remark if the "new" results seem correct?

rovere:
@kpedro88 I can confirm that using single-pion and single-electron guns into HGCAL the network is able to assign the proper ID to the reconstructed `Tracksters`. How this will scale on more complex events with PU is something we still have to verify.
The regressed energy is a little off, right now, but that's expected and due to the changes that went into the reconstruction since the last time the model has been trained. Users have been warned not to use it for the time being. A retraining of the model is needed and will come in the next weeks.

RecoHGCal/TICL/plugins/PatternRecognitionbyCA.cc
kpedro88: Opened #29818 to follow this

kpedro88:
+upgrade

perrotta:
+1
- Correct order of TensorFlow input features is restored
- Jenkins tests pass and show differences where expected
- A retraining of the model will be needed, but this is independent from this pull request

silviodonato:
+1

PR #29224, opened 2020-03-18
DQM: Fix broken plots by MuonTrackValidator

schneiml:
#### PR description:

We got a report that in the `11_1_0_pre4` relvals, some Moun Validation plots appear completely black in the DQMGUI. This is due to `NaN`s in the bin limits, which are in turn due to (apparently) not idempotent code in `BinLogX`, called in booking to change the binning of the `vs_pt` plots, which is a problem with the introduction of `edm::stream` based `DQMEDAnalyzer` in pre4.

This issue can be fixed by moving the call to `BinLogX` into a booking callback. This fixes `MuonTrackValidator`, there is another `BinLogX` in another module that might have the same issue.

The issue was not spotted in #28813 for multiple reasons:
- The problem only appears when running multiple streams; therefore not in the normal PR tests.
- The relevant plots are probably never filled in the PR/IB tests. _Basically_ not a problem, as this is a booking issue, but
- The bin-to-bin comparison tool only checks bin values, not axis data. @andrius-k promised an improvement to cover this in the future.

#### PR validation:
Tested multi-threaded using
```
runTheMatrix.py -l 11634.7 -t 8
```
The `NaN`s disappeared (for one random affected ME):
```
diff <(dqmdumpme.py 11634.7_*/DQM_*.root Muons/RecoMuonV/MuonTrack/tevMuons_picky/dzres_vs_pt) <(dqmdumpme.py base/11634.7_*/DQM_*.root Muons/RecoMuonV/MuonTrack/tevMuons_picky/dzres_vs_pt)
83c83
<         0., 1., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0.,
...
352,364c352,355
<         0., 0., 0., 0.]], dtype=float32), [(array([8.99999976e-01, 1.04997325e+00, 1.22493768e+00, 1.42905772e+00,
<        1.66719151e+00, 1.94500732e+00, 2.26911759e+00, 2.64723635e+00,
<        3.08836389e+00, 3.60300016e+00, 4.20339298e+00, 4.90383339e+00,
<        5.72099400e+00, 6.67432404e+00, 7.78651285e+00, 9.08403301e+00,
<        1.05977688e+01, 1.23637495e+01, 1.44240093e+01, 1.68275795e+01,
<        1.96316776e+01, 2.29030437e+01, 2.67195320e+01, 3.11719971e+01,
<        3.63664093e+01, 4.24264030e+01, 4.94962120e+01, 5.77441025e+01,
[...]

PR #14857, opened 2016-06-13
 add track hit eff monitoring vs inst. lumi

mtosi:
this PR is meant to add the missing monitoring of the hit efficiency vs inst luminosity
the inst luminosity used is an estimate from the pixel cluster

PR #29810, opened 2020-05-12
Replace LumiScaler with OnlineMetaData for Run3 in TrackingDQM

arossi83:
#### PR description:

inside the TrackingDQM some information about luminosity are retrieve via LumiScalers, with this PR the same info will be taken from OnlineMetaData in Run3 Era. The switch take place via Era modifiers.

#### PR validation:

Tested via standard WF:
runTheMatrix.sh -l 11634.0
runTheMatrix.sh -l 138.1

but also manually running the code over a 2018 run (325308) with Era Run3 condition in order to check the access to OnlineMetaData info:
1) cmsDrive step1 -s RAW2DIGI,L1Reco,RECO,DQM -n 100 --eventcontent DQM --datatier DQMIO --conditions 110X_dataRun3_Prompt_v3 --filein /store/express/Run2018E/ExpressPhysics/FEVT/Express-v1/000/325/308/00000/B2080A23-0CFB-6F47-8250-D18B7CC16B45.root --data --no_exec --python_filename=runFirst.py --era Run3
2) cmsDrive step3_DT1_1 -s HARVESTING:dqmHarvesting --filein file:step1_RAW2DIGI_L1Reco_RECO_DQM.root --filetype DQM --scenario pp --data --no_exec --conditions  110X_dataRun3_Prompt_v3 --python_filename=runSecond.py --era Run3
3) cmsRun runFirst.py
4) cmsRun runSecond.py

#### Not a Backport and no backport foreseen


jfernan2:
please test

arossi83:
@mmusich @sroychow FYI

jfernan2:
@arossi83 it seems that HitEffFromHitPatternVsSCALLUMI are having issues being filled with a black canvas, e.g. https://tinyurl.com/yd99w35p
In the past I have seen this problem when using logX, like in:
https://github.com/cms-sw/cmssw/pull/29224
but it does not seem to be your case so I guess it is related to the scalit/scallumi you are doing.
I will try to find out where the bug is originated in your code but you will probably be faster than me
Thanks

mmusich:
@jfernan2 
I don't understand your comment. If I look to the baseline of wf. 23234.0 (which is a phase-2 workflow), even the baseline plots are completely black: https://tinyurl.com/y7fdmxx7, what am I missing? 
BTW they are actually all black even in PRODUCTION relvals, see: https://tinyurl.com/y75njgz8.
It seems to be these plots are buggy to start with.
@mtosi

arossi83:
@jfernan2 I'm checking, some stuff with inf/Nan but I need to understand from where it comes from. In any cases this seems to happen in phase2 wf only. Could you confirm?

jfernan2:
@mmusich they were already in base, I dit not notice sorry. Once again Tracking plots not looked by anybody since a long time because nobody noticed... just spotted now because some changes are being introduced in this PR, my apologies.
@mtosi any chance you coudl fix this or should I remove the plots directly? Thanks

jfernan2:
> @jfernan2 I'm checking, some stuff with inf/Nan but I need to understand from where it comes from. In any cases this seems to happen in phase2 wf only. Could you confirm?

Thanks @arossi83 yes only phase2 wf seem to have this issue.
https://cmssdt.cern.ch/SDT/jenkins-artifacts/baseLineComparisons/CMSSW_11_1_X_2020-05-12-1100+a0acf4/36361/dqm-histo-comparison-summary.html
Thank you

mmusich:
>  Once again Tracking plots not looked by anybody since a long time because nobody noticed... 

@jfernan2 we checked these plots very regularly for Run2 data validation, as they are one of the key monitoring piece to check the pixel / strip dynamic inefficiency, and we still plan to make heavy usage of them for Run3, so they definitely should NOT be removed!

For Phase-2 I agree they are buggy and we never looked at them, because we are not simulating any PU-dependent inefficiency (yet), so in my opinion they should rather be debugged.

mtosi:
https://tinyurl.com/ybnggh3a
they are filled in relval for data, where the SCAL means something

fioriNTU:
@mtosi @mmusich @arossi83 (and @jfernan2 ), sorry if my comment is out of topic, I didn't check in details the code. However, as far as I know, SCAL (FED 735) will be removed from DAQ in Run3 and the luminosity info will be provided by FED 1022. I am not sure though if the code is transparent to this change. Can you confirm?

Also, for the issue of black plots, could you introduce a switch to turn off such plots in MC?

mmusich:
> sorry if my comment is out of topic, I didn't check in details the code. However, as far as I know, SCAL (FED 735) will be removed from DAQ in Run3 and the luminosity info will be provided by FED 1022. I am not sure though if the code is transparent to this change. Can you confirm?

@fioriNTU, if you haven't noticed, that's precisely the purpose of this PR :D

> Also, for the issue of black plots, could you introduce a switch to turn off such plots in MC?

In MC they are not useful indeed, though they are not black in Run3: https://tinyurl.com/yayfa8vl
I am wondering if this is an hint of some other bug.

mtosi:
I agree they should not be ON in MC, though

fioriNTU:
> @fioriNTU, if you haven't noticed, that's precisely the purpose of this PR :D

good, thank you :-) ... so I would also suggest to change the plot name to avoid confusion SCALLUMI->ONLINELUMI ?

mmusich:
@mtosi

> I agree they should not be ON in MC, though

You authored these plots (https://github.com/cms-sw/cmssw/pull/14857), I leave to you the decision about them. 

>  so I would also suggest to change the plot name to avoid confusion SCALLUMI->ONLINELUMI ?

@fioriNTU 
Run2 data will be still defaulted to SCAL though. For sake of simplicity we can move to the more modern data source, though it doesn't seem a big deal to me :)

arossi83:
@jfernan2 with the last commit I fix the bug which create the black plots on phase2 wf but I also remove that plots from MC.

jfernan2:
please test

jfernan2:
thanks a lot @arossi83

sroychow:
Can this be tested on top of PR #29826 ?

jfernan2:
please test with 29826

jfernan2:
please test with #29826

mmusich:
@jfernan2 despite the bot saying tests are still running, I suspect they are stuck, if one looks at the log of jenkins at https://cmssdt.cern.ch/jenkins/job/ib-run-pr-tests/6292/console it says aborted.
Can this round of tests be aborted and restarted?

jfernan2:
please abort

jfernan2:
please test with #29826

sroychow:
@jfernan2 the tests seem to be aborted again. Since #29826 is already merged, could you restart the test for this PR without asking it to be tested with 29826?

jfernan2:
please test

jfernan2:
+1

- SCAL disabled for RUN3

- Black plots vs Lumi disabled for MC

silviodonato:
+1

PR #29824, opened 2020-05-13
L1Trigger/TrackerDTC code restructuring

tschuh:
#### PR description:

I produce the class serving configuration parameter now with an ESProducer, that makes it easier for the downstream producers and analyzers to use it. I moved some functions from the DTC ed product to this es product.

#### PR validation:

Just building and running L1Trigger/TrackerDTC/test/test_cfg.py


kpedro88:
please test

kpedro88:
@tschuh thanks, this is a very useful improvement.

However, there is one problem: this PR introduces a dependency on L1Trigger/TrackerDTC in DataFormats/L1TrackTrigger. This dependence is not allowed. In this case, it's a circular dependence, which is also an issue. This needs to be disentangled, and potentially the Setup class (and associated EventSetup classes) should be moved to a different package.

tschuh:
Regarding the movement of Setup and co. to another package: do you have a particular package in mind?

tschuh:
Regarding the Analyzer: we do in this Analyzer more as one does typically for DQM. Even if one extend the current Validation subsystem with some trackerDTC related stuff, we doubt it that they want everything we have right now. I saw that L1Trigger/TrackTrigger/ puts Analyzer in the test folder. Shall I move  my Analyzer to the test folder too?

kpedro88:
* since the circular dependence is now removed, the Setup class can stay where it is
* if the analyzer is not intended to be used in production, it's better to put it in test/. if it is intended to run in standard workflows (e.g. relvals), it should ultimately be put in a Validation or DQM subsystem.

silviodonato:
please test

L1Trigger/TrackerDTC/src/Setup.cc
```diff
@@ -0,0 +1,730 @@
...
+    }
+  }
+
+  // create sensor modules
+  void Setup::produceSensorModules() {
+    sensorModules_.reserve(numModules_);
+    dtcModules_ = vector<vector<SensorModule*>>(numDTCs_);
+    for (vector<SensorModule*>& dtcModules : dtcModules_)
+      dtcModules.reserve(numModulesPerDTC_);
+    enum SubDetId { pixelBarrel = 1, pixelDisks = 2 };
```
kpedro88, line 442: why not use the existing enum at https://github.com/cms-sw/cmssw/blob/master/DataFormats/SiPixelDetId/interface/PixelSubdetector.h?

kpedro88:
@tomalin @skinnari this PR provides an example of migrating a shared Settings class to an EventSetup product. This approach should eventually be followed for the TMTT and Tracklet code, also. ESProducts are the framework-supported way to share information between different modules/classes/etc.

kpedro88:
please test

kpedro88:
+upgrade

silviodonato:
+operations
- update of ` Configuration/StandardSequences/python/L1TrackTrigger_cff.py` to use the new file `L1Trigger/TrackerDTC/python/ProducerED_cff.py` instead of the old `Producer_cff`

silviodonato:

rekovic:
+1

silviodonato:
+1

PR #29826, opened 2020-05-13
UL remini updates: fix running PUPPI, add egamma scale and smearing

slava77:
- fix running PUPPI, add egamma scale and smearing : 
    - this resolves #29823 (a catchup to #29769 merged in 05-10-0000)
- add egamma scale and smearing, as expected to be used in reminiAOD
    - this is a catch-up to #29526  (merged 05-10-0000)

tested on 136.88811

PR #29832, opened 2020-05-13
Move EffectiveAreas and ThreadSafeFunctor classes out of EgammaTools

guitargeek:
#### Updated PR description after sign-off:

1. Move [EffectiveAreas](https://github.com/cms-sw/cmssw/blob/master/RecoEgamma/EgammaTools/interface/EffectiveAreas.h) and [ConversionTools](https://github.com/cms-sw/cmssw/blob/master/RecoEgamma/EgammaTools/interface/ConversionTools.h) from RecoEgamma/EgammaTools into new package **CommonTools/Egamma**. This new package should have only minimal dependencies (DataFormats and FWCore) and the idea is that non-reco subsystems like NanoAOD or DQM can use these common EgammaTools without inheriting many Reco dependencies.
2. Moved the former `ThreadSafeStringCut` class from **RecoEgamma/EgammaTools** to **CommonTools/Utils**, generalized it and renamed it to `ThreadSafeFunctor`
3.  [`MVAVariableManager`](https://github.com/cms-sw/cmssw/blob/master/RecoEgamma/EgammaTools/interface/MVAVariableManager.h) was refactored a tiny bit to make it theoretically independent from the rest of **RecoEgamma/EgammaTools**

#### Original PR description:

The only **RecoEgamma** library that is used outside the **RecoEgamma** subsystem itself is **RecoEgamma/EgammaTools**, mostly for the [`ConversionTools`](https://github.com/cms-sw/cmssw/blob/master/RecoEgamma/EgammaTools/interface/ConversionTools.h) and the [`EffectiveAreas`](https://github.com/cms-sw/cmssw/blob/master/RecoEgamma/EgammaTools/interface/EffectiveAreas.h) helper class. Unfortunately, this comes with a big [party of reco-dependencies](https://github.com/cms-sw/cmssw/blob/master/RecoEgamma/EgammaTools/BuildFile.xml), which is not desirable. The `EffectiveAreas` helper class for example only depends on `FWCore/Utilites`, so all these reco dependencies from local calo and ecal reconstruction are not needed. What bothers me most is that **NanoAOD** inherits all these low-level reco dependencies like **RecoLocalCalo/HGCalRecAlgos** just for using the `EffectiveAreas`.

To solve this issue, I suggest to move the `EffectiveAreas` class to **CommonTools/Utils**. That package also [only depends on FWCore](https://github.com/cms-sw/cmssw/blob/master/CommonTools/Utils/BuildFile.xml), and all the users of `EffectiveAreas` already depend on it anyway. While at it, I moved the former `ThreadSafeStringCut` class from **RecoEgamma/EgammaTools** to **CommonTools/Utils** as well, generalized it and renamed it to `ThreadSafeFunctor`. Finally, the [`MVAVariableManager`](https://github.com/cms-sw/cmssw/blob/master/RecoEgamma/EgammaTools/interface/MVAVariableManager.h) was also refactored a tiny bit to make it theoretically independent from the rest of **RecoEgamma/EgammaTools** too.

Let me know what you think, and if you agree that **CommonTools/Utils** is the right packaged to move these classes into.

#### PR validation:

No backport intended.


perrotta:
please test

perrotta:
please test with #29826

jfernan2:
+1

slava77:
I think that, if EgammaTools is used outside of the RecoEgamma subsystem *but* is still egamma-specific, I think that it should be moved to a common area with a relevant name.
CommonTools/Utils should not become a collector of everything possibly common.

Is the ```EffectiveAreas``` here really egamma-specific, or is it more relevant as a common tool to quantify response to pileup?
Ignoring a concern on dependence of Reco\* subsystem on PhysicsTools, I'd suggest to move this to PhysicsTools/IsolationUtils, which has a small upstream dependency list and also fits the purpose in physics terms.

If there is a stronger desire to use a CommonTools subsystem, 
Among existing packages CommonTools/PileupAlgos is remotely matching in purpose, but it is perhaps not particularly egamma-specific.

How about a new CommonTools/Egamma ?
Are there any other good candidates (no/small upstream dependence chain) for a move from RecoEgamma/EgammaTools ? If there are at least two, it could be a good reason for a new package.

guitargeek:
Hi Slava, thanks for your assessment! I Didn't see PhysicsTools/IsolationUtils, that might be indeed a good place for the effective area class. But as you point out, it's better if the reco subsystems does not depend too much on PhysicsTools.

I really like the idea of a CommonTools/Egamma that has only light dependencies (FWCore and DataFormats) and is meant to contain classes which are not only used by Egamma but also by other subsystems (or usercode). This would allow for RecoEgamma to be factored out from everything else in the sense that there are not dependencies on RecoEgamma in other subsystems*. And indeed there are two candidates for this package:

1. [EffectiveAreas](https://github.com/cms-sw/cmssw/blob/master/RecoEgamma/EgammaTools/interface/EffectiveAreas.h)
2. [ConversionTools](https://github.com/cms-sw/cmssw/blob/master/RecoEgamma/EgammaTools/interface/ConversionTools.h)

I know that the ConversionTools class is used in a lot of usercode, so if we were to move it, we should maybe leave a "poisoned" header behind that has some verbose static assert which explains where the class went.

What do you think about that?

*Except for Valildation/RecoEgamma depending on RecoEgamma/EgammaMCTools. But that seems fine considering the purpose of the Validation subsystem.

slava77:
> I know that the ConversionTools class is used in a lot of usercode, so if we were to move it, we should maybe leave a "poisoned" header behind that has some verbose static assert which explains where the class went.
> 
> What do you think about that?

sounds good.
Either a replacement include with an ```#error ``` or a still functioning with ```#warning "Obsolete header, include xyz instead"``` should work, IMO.

perrotta:
please test

guitargeek:
> Either a replacement include with an `#error ` or a still functioning with `#warning "Obsolete header, include xyz instead"` should work, IMO.

Okay, the version with the `#error` does not work because of the header consistency checks. I have to change this to `#warning` then.

slava77:
> > Either a replacement include with an `#error ` or a still functioning with `#warning "Obsolete header, include xyz instead"` should work, IMO.
> 
> Okay, the version with the `#error` does not work because of the header consistency checks. I have to change this to `#warning` then.

it sounds like our setup is more OK with removing a header file than obsoleting it by having an ```#error``` in the body.

@smuzaffar @makortel 
is it the desired behavior?

smuzaffar:
@slava77 , looks like a feature :-) It is better to delete the header instead of adding `#error` . If one wants that developers of other package start moving to a new header or do a cleanup then better to use `#warning`

guitargeek:
Thanks for the clarification! And sorry for the force push, I had to change previous commits to avoid files in invalid states.

slava77:
@cmsbuild please test

slava77:
+1

for https://github.com/cms-sw/cmssw/pull/29832	9d189b5 
- code changes are in line with the review, while the PR description may need an update to match what's in the latest version. 
    - This is a rather technical refactoring/relocation of the code.
    - This PR may become an example/reference on how to decouple some more of the POG simple tools used across subsystems

santocch:
+1

jfernan2:
+1

peruzzim:
+xpog
for the technical changes in PhysicsTools/NanoAOD

guitargeek:
Hi @slava77, thanks for the sign-off, I updated the PR description.

slava77:
we are still missing PDMV and L1 signatures.

chayanit:
+1

silviodonato:

rekovic:
+1

silviodonato:
+1

PR #29835, opened 2020-05-14
ecalDigis: do not explicitly set default value

fwyzard:
#### PR description:

`ecalDigis.DoRegional = False` is the default value, no need to set it explicitly in the cff files.


fwyzard:
@cmsbuild, please test

fwyzard:
Wrong branch.

PR #29852, opened 2020-05-15
Linear hbhe channel lookup

igv4321:
#### PR description:

The calibration information for HB/HE channels is now cached locally
(some per run and some per lumi section). Multiple O(log(N)) lookups
of calibration constants (pedestals, gains, channel quality) are
replaced by a single O(1) lookup into the cache using dense channel
indexing. This results in about 20% increase in the throughput if
"Method 0" alone is used for HCAL reco (the speedup is less noticeable
with MAHI). See

https://indico.cern.ch/event/918334/contributions/3861974/attachments/2037521/3412046/Igor_DenseIndex.pdf

This pull request corresponds to the "lazy" version of calibration
information lookup (on the first occurrence of the channel in the
run/lumi section). This type of lookup is slightly more efficient for
processing small numbers of events per lumi section, when there is
a chance that some channels are never read out.

This pull request is not supposed to change any relval results.

#### PR validation:

The usual runTheMatrix tests were run. Some private modifications
were also used to print calibration info and to make sure that it is
the same as before.

### DIFF
--- a/RecoLocalCalo/HcalRecAlgos/interface/HBHEChannelProperties.h
+++ b/RecoLocalCalo/HcalRecAlgos/interface/HBHEChannelProperties.h
@@ -1,0 +1,54 @@
+#ifndef RecoLocalCalo_HcalRecAlgos_HBHEChannelProperties_h_
+#define RecoLocalCalo_HcalRecAlgos_HBHEChannelProperties_h_
+
+#include <array>
+#include <cassert>
+
+#include "RecoLocalCalo/HcalRecAlgos/interface/HBHEPipelinePedestalAndGain.h"
+
+class HcalCalibrations;
+class HcalRecoParam;
+class HcalQIECoder;
+class HcalQIEShape;
+class HcalSiPMParameter;
+
+// Collection of HCAL HB/HE channel information, for faster lookup
+// of this information in reco. This struct does not own any pointers.
+struct HBHEChannelProperties {
+  inline HBHEChannelProperties()
+    : calib(nullptr), paramTs(nullptr), channelCoder(nullptr),
+      shape(nullptr), siPMParameter(nullptr), pedestalsUpdated(false),
+      taggedBadByDb(false), qualityUpdated(false) {}
+
+  inline HBHEChannelProperties(
+    const HcalCalibrations* i_calib,
+    const HcalRecoParam* i_paramTs,
+    const HcalQIECoder* i_channelCoder,
+    const HcalQIEShape* i_shape,
+    const HcalSiPMParameter* i_siPMParameter,
+    const std::array<HBHEPipelinePedestalAndGain, 4>& i_pedsAndGains,
+    const bool i_taggedBadByDb)
+    : calib(i_calib), paramTs(i_paramTs), channelCoder(i_channelCoder),
+      shape(i_shape), siPMParameter(i_siPMParameter),
+      pedsAndGains(i_pedsAndGains), pedestalsUpdated(true),
+      taggedBadByDb(i_taggedBadByDb), qualityUpdated(true) {
+    assert(calib);
+    assert(paramTs);
+    assert(channelCoder);
+    assert(shape);
+    // siPMParameter is allowed to be nullptr (for QIE8)
+  }
+
+  const HcalCalibrations* calib;
+  const HcalRecoParam* paramTs;
+  const HcalQIECoder* channelCoder;
+  const HcalQIEShape* shape;
+  const HcalSiPMParameter* siPMParameter;
+  std::array<HBHEPipelinePedestalAndGain, 4> pedsAndGains;
+  bool pedestalsUpdated;
+
+  bool taggedBadByDb;
+  bool qualityUpdated;
+};
+
+#endif // RecoLocalCalo_HcalRecAlgos_HBHEChannelProperties_h_
--- a/RecoLocalCalo/HcalRecAlgos/interface/HBHEChannelProperties.h
+++ b/RecoLocalCalo/HcalRecAlgos/interface/HBHEChannelProperties.h
@@ -16,22 +16,31 @@ class HcalSiPMParameter;
 // of this information in reco. This struct does not own any pointers.
 struct HBHEChannelProperties {
   inline HBHEChannelProperties()
-    : calib(nullptr), paramTs(nullptr), channelCoder(nullptr),
-      shape(nullptr), siPMParameter(nullptr), pedestalsUpdated(false),
-      taggedBadByDb(false), qualityUpdated(false) {}
+      : calib(nullptr),
+        paramTs(nullptr),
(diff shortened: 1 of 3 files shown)


slava77:
@cmsbuild please test

abdoulline:
@slava77 
Slava, in the light of what has been just discussed at the Release planning meeting (imminent opening of 11_2_X very soon ) may we get this rather technical PR in pre8 to avoid unnecessary backporting?

perrotta:
Measuring with IgProf the memory allocated by the HBHEPhase1Reconstructor, I notice (as expected) an increase in the live memory footage of beginRun from 345 kb to 4.43 Mb:
```
$fgrep HBHEPhase1Reco $OLD | fgrep begin
   0.02        340'632              5  HBHEPhase1Reconstructor::beginRun(edm::Run const&, edm::EventSetup const&) [4159]
            0.0  .............        344'724 / 344'724                    6 / 6                HBHEPhase1Reconstructor::beginRun(edm::Run const&, edm::EventSetup const&) [4159]
[4159]      0.0        344'724        340'632 / 4'092                      6                  HBHEPhase1Reconstructor::beginRun(edm::Run const&, edm::EventSetup const&)
            0.0  .............          4'092 / 344'724                    1 / 6                HBHEPhase1Reconstructor::beginRun(edm::Run const&, edm::EventSetup const&) [4159]

$ fgrep HBHEPhase1Reco $NEW | fgrep begin
   0.02        340'632              5  HBHEPhase1Reconstructor::beginRun(edm::Run const&, edm::EventSetup const&) [1380]
            0.2  .............      4'429'716 / 4'429'716                  7 / 7                HBHEPhase1Reconstructor::beginRun(edm::Run const&, edm::EventSetup const&) [1380]
[1380]      0.2      4'429'716        340'632 / 4'089'084                  7                  HBHEPhase1Reconstructor::beginRun(edm::Run const&, edm::EventSetup const&)
            0.2  .............      4'084'992 / 4'429'716                  1 / 7                HBHEPhase1Reconstructor::beginRun(edm::Run const&, edm::EventSetup const&) [1380]
            0.0  .............          4'092 / 4'429'716                  1 / 7                HBHEPhase1Reconstructor::beginRun(edm::Run const&, edm::EventSetup const&) [1380]

```

In the same workflow (136.874: RunEGamma2018C) I don't notice any sizeable effect on the reco CPU timing: it is true that probably for just 100 events the advantage of caching is not evident, and as you say the effect is less noticeable in MAHI: probably at HLT there is a more evident advantage that in offline reco.

Could you please explain how the results you attached to the PR description were produced, which workflow, etc?
Can you provide an estimate of the gain in some typical use case?

mariadalfonso:
@perrotta 
The test were done on the run321177 (I read the files run321177_ls0142_index000000.raw to remove the problem of reading from disk). Test done on fu-c2a02-37-03.
Test are done averaging over ~4000 events. I think the first few hundred are removed from the average.
Test done with this https://github.com/cms-patatrack/patatrack-scripts/blob/master/scan

For my test to emulate the HLT I put the: noiseFlags off and the M3 are off (those are calculated offline but not at HLT, )
Mahi is also off, since we are trying to slim the pre-processing part.

slava77:
> Measuring with IgProf the memory allocated by the HBHEPhase1Reconstructor, I notice (as expected) an increase in the live memory footage of beginRun from 345 kb to 4.43 Mb:

Looking at the code, this extra memory is apparently going to scale per module instance (per stream/thread). 4MB extra per thread seems excessive. This extra cached data should be moved to some global space (event setup or globalCache). This likely would not allow lazy initialization; there will be some increased cost for somewhat empty events from doing it, but it seems to me that this is not an issue for most applications.

igv4321:
@slava77 If we go into all the trouble of coding this cache in the global space, perhaps we should also make it available to other Hcal modules (HFPreReconstructor in particular). Do I understand correctly that the global cache can not be shared between different modules and this need to be done using event setup?

slava77:
> Do I understand correctly that the global cache can not be shared between different modules and this need to be done using event setup?

I think that the event setup is the most straightforward.
The globalCache is indeed hard to share.

igv4321:
@slava77 In order to build something like this inside the event setup system, I will need to create a new event setup record which depends on CaloGeometryRecord from the package Geometry/Records and on HcalDbRecord from the package CalibFormats/HcalObjects. Which package should be used for this new record?

igv4321:
@slava77 Please also let me know if you want the relevant ESProducer to go somewhere other than CalibCalorimetry/HcalPlugins

slava77:
CaloGeometryRecord seems to depend on many more things than just HCAL, and I did not find it directly in HBHEPhase1Reconstructor.cc. Please clarify what triggers the dependence on CaloGeometryRecord.

I see that creation of HBHEChannelProperties depends on 
- HcalRecoParamsRcd from CondFormats/DataRecord (depends on Geometry/Records)
    - already depends on used HcalRecNumberingRecord (HcalTopology)
- HcalChannelQualityRcd from CondFormats/DataRecord
- HcalSeverityLevelComputerRcd from RecoLocalCalo/HcalRecAlgos
- HcalDbRecord (HcalDbService) from CalibFormats/HcalObjects 

Most of these depend on IdealGeometryRecord. Did I miss CaloGeometry somehow?
If the dependence on HcalSeverityLevelComputerRcd is real, then we may be better of just staying in RecoLocalCalo/HcalRecAlgos, otherwise some clean(er) dependency tree may trigger some moves of existing records/objects.

FWIW, if the only customer for this record and the objects in it are in RecoLocalCalo/HcalRec\*, we could have this in RecoLocalCalo/HcalRecAlgos without introducing dependency problems.

igv4321:
@slava77 What triggers dependence on the CaloGeometryRecord is the necessity to preemptively obtain the list of active channels in various parts of Hcal. Can't rely on what appears in the data steam anymore as the lazy method did. Unfortunately, the dense Ids are not truly dense, there are holes there, and they change from one run period to another.

It is fine with me to stay inside RecoLocalCalo/HcalRecAlgos. I don't expect other modules to use this record. Then the ESProducer goes into RecoLocalCalo/HcalRecProducers?

slava77:
> @slava77 What triggers dependence on the CaloGeometryRecord is the necessity to preemptively obtain the list of active channels in various parts of Hcal. Can't rely on what appears in the data steam anymore as the lazy method did. Unfortunately, the dense Ids are not truly dense, there are holes there, and they change from one run period to another.
> 

I see, it's the part of the code that's not in the PR yet.

> It is fine with me to stay inside RecoLocalCalo/HcalRecAlgos. I don't expect modules other than Hcal to use this record. Then the ESProducer goes into RecoLocalCalo/HcalRecProducers?

Either that or in RecoLocalCalo/HcalRecAlgos/plugins, as HcalRecAlgoESProducer.cc

igv4321:
Replaced by PR #30060

PR #29856, opened 2020-05-15
Revise densities & atomic weights for chemical elements to match Geant4

cvuosalo:
`materials.xml` has contained for many years some non-standard or nonsense values for the densities and atomic weights of certain chemical elements. To avoid any confusion, we are standardizing these values to match those used by Geant4. References for the Geant4 values are as follows.

Atomic weights from NIST: https://www.nist.gov/pml/periodic-table-elements

Densities: http://geant4-userdoc.web.cern.ch/geant4-userdoc/UsersGuides/ForApplicationDeveloper/html/Appendix/materialNames.html

2021 DDD and DD4hep scenarios are updated to use the new version of `materials.xml`. Going forward, all new Run 3 and Phase 2 scenarios should use this new version.

#### PR validation:

Test DB payloads were created with the new materials file and compared with existing DB payloads to check for errors.

Workflow 11634.0_TTbar_14TeV+TTbar_14TeV_TuneCP5_2021 was run using XML geometry with and without this PR. The DQM results show no significant differences.

No backport is needed.

### DIFF
--- a/Geometry/CMSCommonData/data/dd4hep/cmsExtendedGeometry2021.xml
+++ b/Geometry/CMSCommonData/data/dd4hep/cmsExtendedGeometry2021.xml
@@ -4,7 +4,7 @@
   <close_geometry/>
 
   <IncludeSection>
-    <Include ref="Geometry/CMSCommonData/data/materials.xml"/>
+    <Include ref="Geometry/CMSCommonData/data/materials/2021/v1/materials.xml"/>
     <Include ref="Geometry/CMSCommonData/data/rotations.xml"/>
     <Include ref="Geometry/CMSCommonData/data/extend/v2/cmsextent.xml"/>
     <Include ref="Geometry/CMSCommonData/data/cms/2021/v2/cms.xml"/>
--- a/Geometry/CMSCommonData/data/materials/2021/v1/materials.xml
+++ b/Geometry/CMSCommonData/data/materials/2021/v1/materials.xml
@@ -1,77 +1,76 @@
 <?xml version="1.0"?>
 <DDDefinition xmlns="http://www.cern.ch/cms/DDL" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.cern.ch/cms/DDL ../../../DetectorDescription/Schema/DDLSchema.xsd">
  <MaterialSection label="materials.xml">
-  <ElementaryMaterial name="Aluminium" density="2.7*g/cm3" symbol=" " atomicWeight="26.98*g/mole" atomicNumber="13"/>
-  <ElementaryMaterial name="TD_Aluminium" density="8.1*g/cm3" symbol=" " atomicWeight="26.98*g/mole" atomicNumber="13"/>
-  <ElementaryMaterial name="Antimony" density="6.679*g/cm3" symbol=" " atomicWeight="121.75*g/mole" atomicNumber="51"/>
-  <ElementaryMaterial name="Argon" density="1.639*mg/cm3" symbol=" " atomicWeight="39.948*g/mole" atomicNumber="18"/>
-  <ElementaryMaterial name="Arsenic" density="5.72*g/cm3" symbol=" " atomicWeight="74.922*g/mole" atomicNumber="33"/>
+  <ElementaryMaterial name="Aluminium" density="2.699*g/cm3" symbol=" " atomicWeight="26.982*g/mole" atomicNumber="13"/>
+  <ElementaryMaterial name="Antimony" density="6.691*g/cm3" symbol=" " atomicWeight="121.76*g/mole" atomicNumber="51"/>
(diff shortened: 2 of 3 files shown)


cvuosalo:
@cmsbuild please test

civanch:
@cvuosalo , I would agree with fixes in many of materials, however, there are questions:

1) Carbon is the most difficult material, because it may be in different forms from graphite to brilliant with different densities. What is used in CMS is a problem, which need to be evaluated.
2) Bor10 and Bor11 are defined as a separate materials, what is used in CMS?
3) Deuterium - is it used anywhere?

fabiocos:
@cvuosalo @civanch thanks for attacking this issue. I believe that an important practical check is to verify the impact of any change on our material budget evaluation. Was this already tested?

cvuosalo:
@fabiocos  No, there has been no material budget test. Who could perform this test? I would need instructions.

davidlange6:
its not obvious to me why standard densities should be always used in the simulation. (or are these densities not actually used by g4?)

cvuosalo:
@civanch 
1. I'm not sure if the elemental Carbon density is ever used. Composite materials seem to specify their own densities. Various graphite materials also specify their own densities.
2. Bor 10 and Bor 11 are used in materials like "AQBB borated concre", "Heavy Boron concrete", and "B_2 O_3". Heavy boron concrete is used in the muon yoke, and Bor 11 is used in the "TEC_PitchAdapter" in the Tracker.
3. Deuterium is used as a target in `​SimG4Core/​CheckSecondary/​data/​DTarget.xml`.

cvuosalo:
@davidlange6 I think the density of elements that exist in pure form in the detector are used (e.g. copper). But if an element only occurs in compounds, then the elemental density is probably never used.

civanch:
@cvuosalo , because there is no change of results I suspect, that all what is changed is not used in baseline simulation. 

These material descriptions are historical. Even if many of that materials are not used in CMS directly better to keep new values coherent with Geant4.

civanch:
These updated materials are in v1 sub-directory, so do not break anything.

civanch:
+1

cvuosalo:
urgent

cvuosalo:
We would like this PR to go into 11_1_pre8.

davidlange6:
I guess what you mean is that you are updating only the xml geometry and not the GT one so that there is no effect. That doesn’t seem like a good approach as it leaves surprises for later.

On May 19, 2020, at 5:35 PM, Vladimir Ivantchenko <notifications@github.com<mailto:notifications@github.com>> wrote:



These updated materials are in v1 sub-directory, so do not break anything.

civanch:
@davidlange6 , it may be used or not for run-3 - old material.xml is not removed and we need to discuss what is optimal. This new file may be useful for Phase-2, why use very old values for Phase-2? At the same time, the modification is really small, there is no one big fix, so no significant change of results are expected.

cvuosalo:
@davidlange6 This PR should be the last step of revising Run 3 geometry. There have been several geometry revisions in the last few weeks. We decided to bundle them into a single Global Tag update. With the merging of this PR and the closing of 11_1 to new development, I will create new 2021 geometry DB payloads and a new Global Tag.

I have already done preliminary tests that show that there are no major differences between 2021 DB and 11_1 XML geometry.

davidlange6:
Its instead a question of showing that the new values are better.  Should these not get added to the run 3 GT (as thats the geometry that got updated) to see their effect?

On May 19, 2020, at 5:54 PM, Vladimir Ivantchenko <notifications@github.com<mailto:notifications@github.com>> wrote:



@davidlange6<https://github.com/davidlange6> , it may be used or not for run-3 - old material.xml is not removed and we need to discuss what is optimal. This new file may be useful for Phase-2, why use very old values for Phase-2? At the same time, the modification is really small, there is no one big fix, so no significant change of results are expected.

cvuosalo:
@davidlange6 I did test this PR checking the DQM and physics values from workflow 11634.0_TTbar_14TeV+TTbar_14TeV_TuneCP5_2021 using geometry XML and found no significant differences. After I create new geometry DB payloads for 2021, they will be tested, but those will be overall tests of all geometry changes since the previous geometry DB payloads were created, and not just this PR.

davidlange6:
Thanks - this approach makes sense.. Sounds like the PR for the new GT will come soon.

On May 19, 2020, at 6:59 PM, Carl Vuosalo <notifications@github.com<mailto:notifications@github.com>> wrote:



@davidlange6<https://github.com/davidlange6> I did test this PR checking the DQM and physics values from workflow 11634.0_TTbar_14TeV+TTbar_14TeV_TuneCP5_2021 using geometry XML and found no significant differences. After I create new geometry DB payloads for 2021, they will be tested, but those will be overall tests of all geometry changes since the previous geometry DB payloads were created, and not just this PR.

kpedro88:
+upgrade

silviodonato:
+1

PR #29861, opened 2020-05-17
Release memory in HGCGraph

rovere:
#### PR description:

Recent memory profiles done using igprof revealed an important usage of
memory by TICL, in particular during the pattern-recognition stage by
the HGCGraph. After inspection, it has been noticed that the dynamic data
structured internally used, mainly vectors, were cleared at every event
but that a request to free the underlying memory was never
triggered. A call to `shrink_to_fit` has been added to the clearing
step. This solved the issue. The memory profile while processing
events is far less demanding. The swap with an empty vector has also
been tested: since the results were identical, the former solution has
been implemented.

This PR does not reduce the overall memory footprint of TICL. Memory
allocations will still be present while processing data. The possible
absolute improvement of the memory footprint can be addressed at a later
stage. Moreover, the underlying memory re-allocations could be reduced, to a
partial extent, by implementing some smart pre-allocation. In all cases
in which the dimension of the dynamic data structures is known upfront,
the memory allocation is done in one single step calling `reserve`.  In
all other cases, the memory is increased and re-allocated following the
usual exponential growth.

#### PR validation:

`runTheMatrix.py -l limited`

### DIFF
--- a/RecoHGCal/TICL/plugins/HGCGraph.h
+++ b/RecoHGCal/TICL/plugins/HGCGraph.h
@@ -43,8 +43,10 @@ public:
   void clear() {
     allDoublets_.clear();
     theRootDoublets_.clear();
-    theNtuplets_.clear();
     isOuterClusterOfDoublets_.clear();
+    allDoublets_.shrink_to_fit();
+    theRootDoublets_.shrink_to_fit();
+    isOuterClusterOfDoublets_.shrink_to_fit();
   }
   void setVerbosity(int level) { verbosity_ = level; }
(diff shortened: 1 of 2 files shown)


rovere:
please test

perrotta:
+1
- The large memory usage during the pattern-recognition stage by HGCGraph gets corrected by this PR 
- It could be interesting to measure the increase in CPU time when processing high PU samples, but it is not expected to be significant even for PU200, and it can be deferred to the performance monitoring of pre8 (unless @rovere has already a number at hand that can be posted here)
- Jenkins tests pass

kpedro88:
+upgrade

silviodonato:
+1

PR #29872, opened 2020-05-18
PhysicsTools: apply clang-format formatting

fwyzard:
Apply existing `clang-format` rules.


fwyzard:
@cmsbuild, please test

PR #29907, opened 2020-05-19
DQM: Make DQMProvInfo not use lumi transitions.

schneiml:
#### PR description:
As pointed out in #25090 , `DQMProvInfo` is still a `DQMOneLumiEDAnalyzer`.

Rather than giving it the 1:1 port to lumi caches, I restructured it in a way that it does not need lumi transitions at all: In part, to have a nice example for how to do this; in part, because I think this is easier to understand; and in part, because it gives more correct output.

This is done by using a `TProfile` (edit: not any more) to aggregate data within a lumisection, and keeping state in local variables rather than instance variables. The code could easily run as `DQMEDAnalyzer` or `DQMGlobalEDAnalyzer` apart from one logical race condition around `setupLumiSection`, which is needed to treat "never seen" different from "empty". There would be other ways to do this, but `DQMOneEDAnalyzer` should be ok for such a small/fast analyzer.

However, this means the semantics *have* slightly changed, and a more thorough validation is needed.

~Themain difference is that now `reportSummaryMap` is a `TProfile2D`, where each bin is the _percentage (0..1)  of events where this flag was set_. This matches the old definition in the corner cases, but provides more information on the transitions. It also handles merging more correctly (though we should never merge *inside* lumis, so this is not really relevant).~ (Edit: switched back to TH2F for compatibility with the render plugin).

A side effect of the design change is that _empty_ LS now show up as _not valid_, while they were _valid_ before. In case of missing information, bins will be white, instead of red. I think this is acceptable.

Finally, the _valid_ flag for each lumi is set as soon as _at least one_ event of it is processed, while before it was set when _the lumisection is completely processed_. The latter semantics are hard to achieve in modern CMSSW, since we never really know if we have seen the full lumi or not. (It also does not really mean much in online, since this plot is created in a different client compared to most others, and its indistinguishable in the offline output).

#### PR validation:

Not much so far. PR tests should provide a good check on whether things work correctly.

### DIFF
--- a/DQMServices/Components/plugins/DQMProvInfo.cc
+++ b/DQMServices/Components/plugins/DQMProvInfo.cc
@@ -367,7 +367,16 @@ void DQMProvInfo::fillDcsBitsFromDcsStatusCollection(const edm::Handle<DcsStatus
                                                      bool* dcsBits) {
   // Loop over the DCSStatus entries in the DcsStatusCollection
   // (Typically there is only one)
+  bool first = true;
   for (auto const& dcsStatusItr : *dcsStatusCollection) {
+    // By default all the bits are false. We put all the bits on true only
+    // for the first DCSStatus that we encounter:
+    if (first) {
+      for (int vbin = 1; vbin <= MAX_DCS_VBINS; vbin++) {
+        dcsBits[vbin] = true;
+      }
+      first = false;
+    }
     dcsBits[VBIN_CSC_P] &= dcsStatusItr.ready(DcsStatus::CSCp);
     dcsBits[VBIN_CSC_M] &= dcsStatusItr.ready(DcsStatus::CSCm);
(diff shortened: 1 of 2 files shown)


schneiml:
please test

schneiml:
Ok, 700kB added is more than I expected, but exactly consistent with the reportSummarMap now being a profile. The comparison seems to be somewhere between perfectly clean and confused, I'll look at it in more detail tomorrow.

schneiml:
Ah, the GUI seems to have trouble rendering the TProfile based plot, and it is blacklisted now. Good to have checked...

schneiml:
please test

schneiml:
I missed part of the logic used for the legacy (SCAL) DCS status extraction. Should be fixed now.

schneiml:
please test

now things should be fine.

schneiml:
+1

looks indeed good.

silviodonato:
+1

PR #29909, opened 2020-05-19
Run3-ft20 Modify eras and cff's to accommodate steps using dd4hep

bsunanda:
#### PR description:

Modify eras and cff's to accommodate steps using dd4hep

#### PR validation:

Use runTheMatrix test workflows

#### if this PR is a backport please specify the original PR and why you need to backport that PR:

Nothing special

PR #29923, opened 2020-05-20
Run3-ft22 Add 3 scripts to test 3 steps with geometry done through dd4hep

bsunanda:
#### PR description:

Add scripts to test the 3 steps for SIM-DIGI-RECO with Run3 geometry defined using dd4hep 

#### PR validation:

The 3 scripts can work only after integrating PR #29909

Nothing special

### DIFF
--- a/SimG4Core/Configuration/test/dd4hep_ZMM_Run3_Step1_cfg.py
+++ b/SimG4Core/Configuration/test/dd4hep_ZMM_Run3_Step1_cfg.py
@@ -1,0 +1,203 @@
+import FWCore.ParameterSet.Config as cms
+
+from Configuration.Eras.Era_Run3_dd4hep_cff import Run3_dd4hep
+
+process = cms.Process('SIM',Run3_dd4hep)
+
+# import of standard configurations
+process.load('Configuration.StandardSequences.Services_cff')
+process.load('SimGeneral.HepPDTESSource.pythiapdt_cfi')
+process.load('FWCore.MessageService.MessageLogger_cfi')
+process.load('Configuration.EventContent.EventContent_cff')
+process.load('SimGeneral.MixingModule.mixNoPU_cfi')
+process.load('Configuration.StandardSequences.MagneticField_cff')
+process.load('Configuration.StandardSequences.Generator_cff')
+process.load('IOMC.EventVertexGenerators.VtxSmearedRun3RoundOptics25ns13TeVLowSigmaZ_cfi')
+process.load('GeneratorInterface.Core.genFilterSummary_cff')
+process.load('Configuration.StandardSequences.SimIdeal_cff')
+process.load('Configuration.StandardSequences.EndOfProcess_cff')
+process.load('Configuration.StandardSequences.FrontierConditions_GlobalTag_cff')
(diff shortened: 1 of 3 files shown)


bsunanda:
@cmsbuild Please test

SimG4Core/Configuration/test/dd4hep_ZMM_Run3_Step1_cfg.py
```diff
@@ -0,0 +1,203 @@
...
+
+process.maxEvents = cms.untracked.PSet(
+    input = cms.untracked.int32(10),
+    output = cms.optional.untracked.allowed(cms.int32,cms.PSet)
+)
+
+# --------------------------------start of new part----------------------------
+#Other statements
+
+process.RandomNumberGeneratorService = cms.Service("RandomNumberGeneratorService",
```
civanch, line 30: @bsunanda , our default random number generator MixMaxRng and not HepJamesRandom, also initial integer better to be different for different producers. Make all = 69 may be acceptable but better having it different.

fabiocos:
@bsunanda is this supposed to be a runTheMatrix-like standard workflow? I had started to prepare for that in https://github.com/cms-sw/cmssw/blob/master/Geometry/CMSCommonData/test/dd4hep/runDD4hepTestWF.sh where only the step1 was functional at that time. What is different in these configurations compared to what you might get there by extending up to step3? This of course is just a test in view of a full integration in PyRelVal, but in practice it goes along the spirit of this PR, just using the standard tools instead of an explicit coding of each step.

slomeo:
@fabiocos , just a comment: I need to test, for the Muon System, the DDD Geometry vs DD4hep Geometry (scenario 2021) by using MC Samples. I wrote 4 configuration files (from step1 to step4) in order to do it, by using Z->MuMu fragment at 14 TeV (in CMSSW_11_1_0_pre7). Step 4 is necessary in order to check quickly the differences among histos. When I added Configuration.Geometry.GeometryDD4hepExtended2021_cff or Configuration.Geometry.GeometryDD4hepExtended2021Reco_cff, suggested by @cvuosalo, in my configuration files I got some problems for step3, so I wrote to @bsunanda @ianna @cvuosalo @civanch to get help.

SimG4Core/Configuration/test/dd4hep_ZMM_Run3_Step1_cfg.py
bsunanda: I am removing that part. I took the script from Sergio and made minimal change to make it work - I am removing that.

bsunanda:
@fabiocos @slomeo @civanch As Sergio mentioned he initiated this which triggered me to define a new ERA as in PR #29909. But the steps did not work and that was because issues in some of the detectors which are not usually looked after by me. So I thought it is better to have the scripts available to all so that the DPG contacts can at their corner can solve the respective problems. I would like to discuss this in tomorrow's meeting

bsunanda:
@cmsbuild Please test

civanch:
+1

civanch:
urgent

silviodonato:
+1

PR #29924, opened 2020-05-20
Fixing online behaviour of DQMFileSaverPB

Sam-Harper:
#### PR description:

This PR is a bug fix to the recent DQM revamp which solely effects the HLT when run by the DAQ.   For hltd to properly collect and merge the dqm histogram stream produced by the HLT, it needs a init file to be written.  Otherwise the stream is ignored and at best the histograms are lost. 

This PR adds that code.  It should have no effect on any offline test, this only effects P5 running.

#### PR validation:

The hilton tests successfully produced the DQM Histogram stream.


jfernan2:
please test

jfernan2:
+1

silviodonato:
+1

PR #29933, opened 2020-05-21
[RFC] Test clang-tidy --checks performance-unnecessary-value-param

fwyzard:
See the description at https://clang.llvm.org/extra/clang-tidy/checks/performance-unnecessary-value-param.html .


fwyzard:
please test

CalibTracker/SiStripCommon/interface/Book.h
```diff
@@ -24,7 +24,7 @@ class Book {
   TDirectory* directory;
 
   struct match_name {
-    match_name(string_t re) : expression(re) {}
+    match_name(string_t& re) : expression(re) {}
```
kpedro88, line 27: why didn't it use `const` for any references in this file?

CommonTools/Utils/src/FormulaEvaluator.cc
```diff
@@ -236,9 +236,10 @@ namespace {
       return false;
     }
 
-    EvaluatorInfo createEvaluator(std::string::const_iterator iBegin,
-                                  std::string::const_iterator iEnd,
-                                  std::shared_ptr<reco::formula::BinaryOperatorEvaluatorBase> iPreviousBinary) const {
+    EvaluatorInfo createEvaluator(
+        std::string::const_iterator iBegin,
+        std::string::const_iterator iEnd,
+        const std::shared_ptr<reco::formula::BinaryOperatorEvaluatorBase>& iPreviousBinary) const {
```
kpedro88, line 242: not sure this is desirable - maybe it's better to modify this function to take the underlying pointer directly, if it's not supposed to have ownership?

Alignment/APEEstimation/interface/TrackerSectorStruct.h
```diff
@@ -54,14 +56,29 @@ class TrackerSectorStruct {
...
+                                        const TString &,
+                                        const TString &,
+                                        const TString &,
+                                        int,
+                                        int,
+                                        double,
+                                        double,
+                                        const std::string & = "nphtr");
   inline CorrelationHists bookCorrHistsX(
       TString, TString, TString, TString, int, int, double, double, std::string = "nphtr");
```
slava77, line 70: why would this method not be updated in the same way to ```const &```

slava77:
-1

please resubmit a smaller version if possible; my browser is failing to show diffs in 2K files and it took about a minute to highlight the right line to make a comment

slava77:
On 5/26/20 5:53 PM, Slava Krutelyov wrote:
> -1

The "-1" was mostly to what the tool does.

I'm fine with the update conceptually,
 but the tools is apparently not doing it in a clearly consistent way;
also, inserting moves to everything that compiler would optimize away
seems excessive

mrodozov:
-note
needed some fixes until get it in

mrodozov:
@fwyzard I only need help with this one, the rest I got in. It needed more input before get it in, correct ?

fwyzard:
Yes, we should review what this check is doing (and not doing).

Something that we did not discuss yesterday, maybe for the core software meeting: what is people's feeling about enabling some checks as "warning only" during the compilation step, without generating the automatic fixes ?

mrodozov:
and so we put ourselves a reminder to ask that next Tuesday.
In the past we had warnings for which we didn't break the PR checks (code-checks or others)
I don't remember if there were some consequences following from that

PR #29935, opened 2020-05-21
[RFC] Test clang-tidy --checks modernize-loop-convert

fwyzard:
See the description at https://clang.llvm.org/extra/clang-tidy/checks/modernize-loop-convert.html .


kpedro88:
It seems like this one could also interact with modernize-use-auto

kpedro88:
I'm curious why this misses some seemingly trivial cases, e.g. https://github.com/cms-sw/cmssw/blob/a004c2908ca3206fad482b61bd60e00b84de883d/PhysicsTools/PatUtils/interface/ShiftedJetProducerT.h#L113-L114

PR #29945, opened 2020-05-21
Replaced boost::unordered_map for std::unordered_map at HcalCommon

camolezi:
#### PR description:
std::unordered_map and boost::unordered_map have very similar performance. So, we can reduce boost dependency by replacing boost::unordered_map for std::unordered_map.

#### PR validation:

I've run some benchmarks to compare boost vs std unordered_maps performance:

Map search benchmark.
https://gist.github.com/camolezi/f14800fd4c60af7d2ef501b6b92f9ddd

Adding and removing elements.
https://gist.github.com/camolezi/a5a39539a8c7252d74ec8e6f17f6aa37

@vgvassilev @davidlange6

### DIFF
--- a/DQM/HcalCommon/interface/Container1D.h
+++ b/DQM/HcalCommon/interface/Container1D.h
@@ -17,8 +17,8 @@
 #include "DQM/HcalCommon/interface/Utilities.h"
 #include "DQM/HcalCommon/interface/ValueQuantity.h"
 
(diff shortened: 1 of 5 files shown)


jfernan2:
please test

jfernan2:
+1

silviodonato:
+1

PR #29926, opened 2020-05-21
[RFC] Test clang-tidy --checks performance-faster-string-find

fwyzard:
See the description at https://clang.llvm.org/extra/clang-tidy/checks/performance-faster-string-find.html .

PR #29947, opened 2020-05-21
Fix DetectorDescription Parser TE unit test so it runs

cvuosalo:
In PR #29926, the DetectorDescription Parser TE unit test fails because a test program cannot be found in the PATH. The test program may be located in any of three locations, depending on how the unit tests are being run. I added a search of those directories to find the program.

#### PR validation:

The unit test for the DetectorDescription/Parser package that was failing can now run successfully.

Backport to 11_1 will be done after this PR is merged.

### DIFF
--- a/DetectorDescription/Parser/test/run_testDoc.sh
+++ b/DetectorDescription/Parser/test/run_testDoc.sh
@@ -7,6 +7,7 @@ pushd ${LOCAL_TMP_DIR}
   cd ${LOCAL_TEST_DIR}
 #  cp ${LOCAL_TOP_DIR}/test/${SCRAM_ARCH}/testDoc .
   echo ${test}testDoc ------------------------------------------------------------
+  env
   testDoc testConfiguration.xml || die "testDoc" $?
   export PATH=${mecpath}
   popd
--- a/DetectorDescription/Parser/test/run_testDoc.sh
+++ b/DetectorDescription/Parser/test/run_testDoc.sh
@@ -1,13 +1,19 @@
 #!/bin/bash
+
 test=testtestDoc
 function die { echo Failure $1: status $2 ; exit $2 ; }
 pushd ${LOCAL_TMP_DIR}
-  export mecpath=${PATH}
-  test -z "${CMSSW_RELEASE_BASE}" && echo CMSSW_RELEASE_BASE is not set.
-  export PATH=${CMSSW_RELEASE_BASE}/test/${SCRAM_ARCH}/:${LOCAL_TOP_DIR}/test/${SCRAM_ARCH}/:${PATH}
+  export PATH=${LOCAL_TOP_DIR}/test/${SCRAM_ARCH}/:${PATH}
   cd ${LOCAL_TEST_DIR}
   echo ${test} testDoc ------------------------------------------------------------
-  testDoc testConfiguration.xml || die "testDoc" $?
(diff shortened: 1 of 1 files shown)


cvuosalo:
type bug-fix

cvuosalo:
@cmsbuild please test

cvuosalo:
The Jenkins environment is different from the cmsenv environment. My fix worked in my test directory with `scram b runtests`, but it doesn't work in Jenkins.

cvuosalo:
@cmsbuild please abort test

cvuosalo:
@cmsbuild please test

cvuosalo:
@cmsbuild please test

cvuosalo:
@cmsbuild please test

DetectorDescription/Parser/test/run_testDoc.sh
```diff
@@ -3,11 +3,11 @@ test=testtestDoc
...
   export mecpath=${PATH}
-  export PATH=${LOCAL_TOP_DIR}/test/${SCRAM_ARCH}/:${PATH}
+  test -z "${CMSSW_RELEASE_BASE}" && echo CMSSW_RELEASE_BASE is not set.
+  export PATH=${CMSSW_RELEASE_BASE}/test/${SCRAM_ARCH}/:${LOCAL_TOP_DIR}/test/${SCRAM_ARCH}/:${PATH}
   cd ${LOCAL_TEST_DIR}
-#  cp ${LOCAL_TOP_DIR}/test/${SCRAM_ARCH}/testDoc .
-  echo ${test}testDoc ------------------------------------------------------------
-  testDoc testConfiguration.xml|| die "testDoc" $?
+  echo ${test} testDoc ------------------------------------------------------------
+  testDoc testConfiguration.xml || die "testDoc" $?
```
smuzaffar, line 10: @cvuosalo , the `testDoc` is not part of this Package. So calling it here is not going to work. I would suggest to move this unit tst in to `DetectorDescription/RegressionTest` package (which has `testDoc`). If you do not want to move this then you need to do something like [a]. Note that one has to lookin local directory (i.e. CMSSW_BASE) , release directory (i.e. CMSSW_RELEASE_BASE) or full release path e.g. in case of patch release (CMSSW_FULL_RELEASE_BASE). I would still suggest to move this test to `DetectorDescription/RegressionTest` . In that case `scram n runtests` will find it.

```
testDoc=testDoc
for d in $CMSSW_BASE $CMSSW_RELEASE_BASE $CMSSW_FULL_RELEASE_BASE ; do
  if [ -x $d/test/${SCRAM_ARCH}/testDoc ] ; then 
    testDoc=$d/test/${SCRAM_ARCH}/testDoc
    break
  fi
done
$testDoc testConfiguration.xml
```

cvuosalo:
@cmsbuild please test

cvuosalo:
+1

DetectorDescription/Parser/test/run_testDoc.sh
cvuosalo: @smuzaffar Thanks for the code snippet. That solves the problem. I don't want to move the test. This whole package may go away when the old DD is deleted next year. Also, this test is for this package, so it does belong here, not in `DetectorDescription/RegressionTest`.

silviodonato:
+1

PR #29950, opened 2020-05-22
Run3-ft23 Correct run3 scenario for dd4hep and add the other Phase2 scenarios

bsunanda:
#### PR description:

Correct run3 scenario for dd4hep and add the other Phase2 scenarios

#### PR validation:

Use standard runTheMatrix test workflows 

Nothing special


bsunanda:
@cmsbuild Please test

kpedro88:
@bsunanda eventually we should duplicate/extend/modify the phase 2 geometry scripts so this can be maintained more automatically

cvuosalo:
The Run 3 change, at least, should be backported into 11_1.

cvuosalo:
+1

silviodonato:
+1

PR #29952, opened 2020-05-22
[DD4HEP] common and muon alignment DD4HEP migration

hyunyong:
**PR description:**
DD4HEP migration targets

- /Alignment/CommonAlignmentMonitor/plugins/AlignmentMonitorAsAnalyzer.cc
- /Alignment/CommonAlignmentProducer/src/AlignmentProducerBase.cc
- /Alignment/MuonAlignment/plugins/MisalignedMuonESProducer.cc
- /Alignment/MuonAlignment/src/MuonAlignmentInputMethod.cc
- /Alignment/MuonAlignment/src/MuonAlignmentOutputXML.cc
- /Alignment/MuonAlignment/test/TestReader.cpp

Migration codes get the moun detector geometries from the event setup, so no more geometry builds on the alignment codes. 
Tracker alignment parts on common alignment are modified by @adewit


mmusich:
@hyunyong can you apply the required code-format patch? otherwise the PR can't be tested.

hyunyong:
> @hyunyong can you apply the required code-format patch? otherwise the PR can't be tested.

Yes, thank you!

cvuosalo:
@cmsbuild please test

cvuosalo:
Looks good. Removing the dependence on DD is a helpful improvement.

pohsun:
please test

cvuosalo:
+1

Alignment/MuonAlignment/plugins/MuonGeometryArrange.cc
```diff
@@ -44,7 +44,10 @@ MuonGeometryArrange::MuonGeometryArrange(const edm::ParameterSet& cfg)
       _levelStrings(cfg.getUntrackedParameter<std::vector<std::string> >("levels")),
       _writeToDB(false),
       _commonMuonLevel(align::invalid),
-      firstEvent_(true) {
+      firstEvent_(true),
+      idealLabel1("MuonGeometryArrangeLabel1"),
+      idealLabel2("MuonGeometryArrangeLabel2"),
+      idealLabel3("MuonGeometryArrangeLabel3") {
```
pohsun, line 50: Given the treatments with the three labels are not identical, I suggest using more intention revealing names than `idealLabel`.

hyunyong: They are identical ideal geometries. This code needs three ideal geometries, so the labels were ```idealLabel```. 
I tried to use a bit detailed label names but basically the label name comes from ```inputMethod1()```, ```inputMethod2()```, and ```inputMethod3()```.

christopheralanwest:
please test

Alignment/MuonAlignment/plugins/MuonGeometryArrange.cc
pohsun: My mistake for not clearly expressing my concern. What I want to emphasis is related to the function `void MuonGeometryArrange::compareGeometries(Alignable* refAli, Alignable* curAli, Alignable* curCopy)`, which clearly shows the meaning of each `Alignable*`s via the argument name. As the three ideal geometry labels has their own role in geometry comparison, could you use names like `refMuonGeometryArrangeLabel` to make the label names less ambiguous?

cvuosalo:
+1

silviodonato:
Do you have any comments? alca:  @pohsun @tlampen @tocheng @christopheralanwest

Alignment/MuonAlignment/plugins/MuonGeometryArrange.cc
hyunyong: Since all of the input geometries are the same ideal geometry, so I think if I put the label as reference and current, it looks like they are different geometries. So, I would like to keep ```idealLable```.

pohsun: Ok, thank you for the explaination.

pohsun:
+1

silviodonato:
+1

PR #29956, opened 2020-05-22
Allow DDNames to be created concurrently

Dr15Jones:
#### PR description:

In order to support concurrent running of ES modules we need to be able to concurrently create DDNames. This change replaces the global containers being used with tbb concurrent containers.

#### PR validation:
The code compiles and all unit tests in DetectorDescription/Core pass.

### DIFF
--- a/DetectorDescription/Core/BuildFile.xml
+++ b/DetectorDescription/Core/BuildFile.xml
@@ -4,6 +4,7 @@
 <use name="boost"/>
 <use name="clhep"/>
 <use name="rootmath"/>
+<use name="tbb"/>
 <export>
   <lib name="1"/>
 </export>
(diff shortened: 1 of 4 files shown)


Dr15Jones:
please test

cvuosalo:
@Dr15Jones How urgent is this change? We are in the midst of DD4hep migration, and we would like to keep the old code like this unchanged to avoid validation problems. If both the old DD and DD4hep code are changing at the same time, it becomes hard to understand differing results between them.

Dr15Jones:
This change should have absolutely no difference. And it is fairly urgent since I have the code to allow concurrent ES modules to run.

cvuosalo:
@ianna Could you please review this change? I would like to be sure it doesn't cause any inadvertent changes.

ianna:
+1

no adverse effect of this change

silviodonato:
+1

PR #29965, opened 2020-05-23
ECAL : Tools to include EcalPFRecHitThresholds in Cond DB

depasse:
#### PR description:

To allow the test of multiple scenarii with Ecal Particle Flow RecHit Thresholds, some simple txt files are asked to be included in DataBase in tag form.  These pgms are the tools for this.  a bug was found in  CalibCalorimetry/EcalTrivialCondModules/src/EcalTrivialConditionRetriever.cc  to perform this operation.  It is corrected.

#### PR validation:

Creation of sqlite files with the pgms and dumping the payloads gives the same results as txt files.

Done : 
scram build code-checks
scram build code-format

### DIFF
--- a/CalibCalorimetry/EcalTrivialCondModules/src/EcalTrivialConditionRetriever.cc
+++ b/CalibCalorimetry/EcalTrivialCondModules/src/EcalTrivialConditionRetriever.cc
@@ -27,10 +27,11 @@
 using namespace edm;
 
 EcalTrivialConditionRetriever::EcalTrivialConditionRetriever(const edm::ParameterSet& ps) {
-  //  std::string path = "CalibCalorimetry/EcalTrivialCondModules/data/";
-  std::string path = "CalibCalorimetry/EcalTrivialCondModules/data_test/";
+  std::string dataPath_ =
+      ps.getUntrackedParameter<std::string>("dataPath", "CalibCalorimetry/EcalTrivialCondModules/data/");
+  //  std::string dataPath_ = "CalibCalorimetry/EcalTrivialCondModules/data_test/";
 
-  /*since CMSSW_10_6_0, this path points to https://github.com/cms-data/CalibCalorimetry-EcalTrivialCondModules  (extra package). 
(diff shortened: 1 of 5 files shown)


CalibCalorimetry/EcalTrivialCondModules/src/EcalTrivialConditionRetriever.cc
```diff
@@ -27,7 +27,8 @@
 using namespace edm;
 
 EcalTrivialConditionRetriever::EcalTrivialConditionRetriever(const edm::ParameterSet& ps) {
-  std::string path = "CalibCalorimetry/EcalTrivialCondModules/data/";
+  //  std::string path = "CalibCalorimetry/EcalTrivialCondModules/data/";
+  std::string path = "CalibCalorimetry/EcalTrivialCondModules/data_test/";
```
pohsun, line 31: If I have a correct understanding.
There could be some effect when other user tries to use this package.
Will it be better to change the path with an optional configuration?

depasse: Could you suggest stgh with this optional argument ?  Thanks.

pohsun: Unless there are some special concern, `ps.getUntrackedParameter<std::string>("dataPath", "CalibCalorimetry/EcalTrivialCondModules/data/");` should be fine, with which the default path will be taken when `dataPath` is not explicitly defined in the corresponding Python configuration file. 

Ref: The path of this format `CalibCalorimetry/EcalTrivialCondModules/data/` is special in CMSSW, as described in [twiki](https://twiki.cern.ch/twiki/bin/view/CMSPublic/SWGuideBuildFile#CmsswData).

pohsun:
please test

pohsun:
+1

silviodonato:
Do you have any comments? @ggovi

ggovi:
+1

silviodonato:
+1

PR #25591, opened 2019-01-07
New TrackingRecHitRange and use it in RecoParticleFlow

guitargeek:
This PR suggests to add a member function to TrackExtraBase which returns an instance of the new TrackingRecHitRange (template instantiation of new edm::ConstRange<T>) that can be used to express range-based loops over the RecHits associated to a track. I think this would be beneficial for the modernization of reconstruction code, as I show at the example of RecoParticleFlow, where I replaced all occurrences of recHitsBegin and recHitsEnd.

Local matrix tests pass.

PR #29971, opened 2020-05-24
Better TrajectorySeed interface to recHits

guitargeek:
#### PR description:

The current interface of the TrajectorySeed class to its recHits bothered me for quite some time, especially since the ElectronSeed inherits it. I'm speaking about this member function [1]:

```C++
std::pair<const_iterator, const_iterator> recHits() const { return std::make_pair(hits_.begin(), hits_.end()); }
```
~~which I think should better just return a const reference to the `edm::OwnVector<TrackingRecHit>` of seeds (proposed in this PR):~~ *original idea that got discarded in review*
which should better return a `edm::Range<RecHitContainer::const_iterator>` that implements `begin()` and `end()` such that `recHits()` can be used in range-based loops:
```C++
edm::Range<RecHitContainer::const_iterator> recHits() const { return {hits_.begin(), hits_.end()}; }
```
this also avoids the very misleading expression `recHits().second`: one might think that it points to the second rechit, but actually it is the `hits_.end()` iterator. This caused possible bugs here [2] and here [3], I think. No attempt is made in this PR to fix these potential bugs, it only adapts the code to the newly proposed range-loop friendly interface without changing any functionality (hopefully).

[1] https://github.com/cms-sw/cmssw/blob/master/DataFormats/TrajectorySeed/interface/TrajectorySeed.h#L52
[2] https://github.com/cms-sw/cmssw/blob/master/RecoTracker/MeasurementDet/src/StartingLayerFinder.cc#L79
[3] https://github.com/cms-sw/cmssw/blob/master/Validation/RecoMuon/src/MuonSeedTrack.cc#L208

#### PR validation:

No backport intended.

### DIFF
--- a/DataFormats/MuonSeed/src/L2MuonTrajectorySeed.cc
+++ b/DataFormats/MuonSeed/src/L2MuonTrajectorySeed.cc
@@ -12,7 +12,7 @@ L2MuonTrajectorySeed::L2MuonTrajectorySeed() : TrajectorySeed() {}
 
 // Constructor
 L2MuonTrajectorySeed::L2MuonTrajectorySeed(PTrajectoryStateOnDet const& ptsos,
-                                           recHitContainer const& rh,
+                                           RecHitContainer const& rh,
                                            PropagationDirection dir,
                                            l1extra::L1MuonParticleRef l1Ref)
     : TrajectorySeed(ptsos, rh, dir) {
@@ -20,7 +20,7 @@ L2MuonTrajectorySeed::L2MuonTrajectorySeed(PTrajectoryStateOnDet const& ptsos,
 }
 
 L2MuonTrajectorySeed::L2MuonTrajectorySeed(PTrajectoryStateOnDet const& ptsos,
-                                           recHitContainer const& rh,
+                                           RecHitContainer const& rh,
                                            PropagationDirection dir,
                                            l1t::MuonRef l1Ref)
     : TrajectorySeed(ptsos, rh, dir) {
--- a/RecoEgamma/EgammaElectronAlgos/src/ElectronSeedGenerator.cc
+++ b/RecoEgamma/EgammaElectronAlgos/src/ElectronSeedGenerator.cc
@@ -37,8 +37,8 @@ namespace {
     if (s1.nHits() != s2.nHits())
       return false;
 
-    TrajectorySeed::RecHitRange r1 = s1.recHits();
-    TrajectorySeed::RecHitRange r2 = s2.recHits();
+    const TrajectorySeed::RecHitRange r1 = s1.recHits();
+    const TrajectorySeed::RecHitRange r2 = s2.recHits();
     for (auto i1 = r1.begin(), i2 = r2.begin(); i1 != r1.end(); ++i1, ++i2) {
       if (!i1->isValid() || !i2->isValid())
         return false;
--- a/RecoEgamma/EgammaElectronAlgos/src/ElectronSeedGenerator.cc
+++ b/RecoEgamma/EgammaElectronAlgos/src/ElectronSeedGenerator.cc
@@ -37,7 +37,8 @@ namespace {
     if (s1.nHits() != s2.nHits())
       return false;
 
-    auto const &r1 = s1.recHits(), r2 = s2.recHits();
+    TrajectorySeed::RecHitRange r1 = s1.recHits();
+    TrajectorySeed::RecHitRange r2 = s2.recHits();
     for (auto i1 = r1.begin(), i2 = r2.begin(); i1 != r1.end(); ++i1, ++i2) {
       if (!i1->isValid() || !i2->isValid())
         return false;
--- a/RecoEgamma/EgammaElectronAlgos/src/ElectronSeedGenerator.cc
+++ b/RecoEgamma/EgammaElectronAlgos/src/ElectronSeedGenerator.cc
@@ -37,10 +37,8 @@ namespace {
     if (s1.nHits() != s2.nHits())
       return false;
 
-    unsigned int nHits;
-    TrajectorySeed::range r1 = s1.recHits(), r2 = s2.recHits();
-    TrajectorySeed::const_iterator i1, i2;
-    for (i1 = r1.first, i2 = r2.first, nHits = 0; i1 != r1.second; ++i1, ++i2, ++nHits) {
+    auto const &r1 = s1.recHits(), r2 = s2.recHits();
+    for (auto i1 = r1.begin(), i2 = r2.begin(); i1 != r1.end(); ++i1, ++i2) {
       if (!i1->isValid() || !i2->isValid())
         return false;
       if (i1->geographicalId() != i2->geographicalId())
--- a/RecoTracker/CkfPattern/plugins/GroupedCkfTrajectoryBuilder.cc
+++ b/RecoTracker/CkfPattern/plugins/GroupedCkfTrajectoryBuilder.cc
@@ -836,10 +836,6 @@ void GroupedCkfTrajectoryBuilder::rebuildSeedingRegion(const TrajectorySeed& see
   KFTrajectoryFitter fitter(backwardPropagator(seed), &updator(), &estimator(), 3, nullptr, &hitCloner);
   //
   std::vector<const TrackingRecHit*> seedHits;
-  //seedHits.insert(seedHits.end(), rseedHits.first, rseedHits.second);
-  //for (TrajectorySeed::recHitContainer::const_iterator iter = rseedHits.first; iter != rseedHits.second; iter++){
-  //	seedHits.push_back(&*iter);
-  //}
 
   unsigned int nSeed = seed.nHits();
   //seedHits.reserve(nSeed);
--- a/RecoTracker/CkfPattern/plugins/GroupedCkfTrajectoryBuilder.cc
+++ b/RecoTracker/CkfPattern/plugins/GroupedCkfTrajectoryBuilder.cc
@@ -836,15 +836,13 @@ void GroupedCkfTrajectoryBuilder::rebuildSeedingRegion(const TrajectorySeed& see
   auto hitCloner = static_cast<TkTransientTrackingRecHitBuilder const*>(hitBuilder())->cloner();
   KFTrajectoryFitter fitter(backwardPropagator(seed), &updator(), &estimator(), 3, nullptr, &hitCloner);
   //
-  TrajectorySeed::range rseedHits = seed.recHits();
   std::vector<const TrackingRecHit*> seedHits;
   //seedHits.insert(seedHits.end(), rseedHits.first, rseedHits.second);
   //for (TrajectorySeed::recHitContainer::const_iterator iter = rseedHits.first; iter != rseedHits.second; iter++){
   //	seedHits.push_back(&*iter);
   //}
 
-  //unsigned int nSeed(seedHits.size());
-  unsigned int nSeed(rseedHits.second - rseedHits.first);
+  unsigned int nSeed = seed.nHits();
   //seedHits.reserve(nSeed);
   TempTrajectoryContainer rebuiltTrajectories;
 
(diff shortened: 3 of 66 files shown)


fwyzard:
Does the current code already guarantee that the rechits are always stored in a vector ?

guitargeek:
Pardon, I should have said are stored in an `edm::OwnVector<TrackingRecHit>` not a normal vector. I clarified that above. The `OwnVector`s interface fortunately comes with it's own iterator class and implements `begin()`, `end()` and `empty()`, which is why the range-based loop can be used:
https://github.com/cms-sw/cmssw/blob/master/DataFormats/Common/interface/OwnVector.h#L153
Unfortunately I'm not that familiar with the  `edm::OwnVector` class, so if you have any further concerns I might not be able to comment them immediately.

slava77:
@cmsbuild please test

fwyzard:
> Pardon, I should have said that they are stored in an `edm::OwnVector<TrackingRecHit>` not a normal vector. I clarified that above. The `OwnVector`s interface fortunately comes with it's own iterator class and implements `begin()`, `end()` and `empty()`, which is why the range-based loop can be used:
> https://github.com/cms-sw/cmssw/blob/master/DataFormats/Common/interface/OwnVector.h#L153
> Unfortunately I'm not that familiar with the `edm::OwnVector` class, so if you have any further concerns I might not be able to comment them immediately.

Sorry, I wasn't very clear: I was wondering if there were instances where `range recHits() const (...)` would not return the full `OwnVector`, but a subset of it: the current interface would allow for it, while the proposed one would not.

RecoEgamma/EgammaElectronAlgos/src/ElectronSeedGenerator.cc
```diff
@@ -37,10 +37,8 @@ namespace {
     if (s1.nHits() != s2.nHits())
       return false;
 
-    unsigned int nHits;
-    TrajectorySeed::range r1 = s1.recHits(), r2 = s2.recHits();
-    TrajectorySeed::const_iterator i1, i2;
-    for (i1 = r1.first, i2 = r2.first, nHits = 0; i1 != r1.second; ++i1, ++i2, ++nHits) {
+    auto const &r1 = s1.recHits(), r2 = s2.recHits();
```
fwyzard, line 40:
```suggestion
    auto const &r1 = s1.recHits(), &r2 = s2.recHits();
```
otherwise `r2` will be a copy of `s2.recHits()`.

RecoTracker/CkfPattern/plugins/GroupedCkfTrajectoryBuilder.cc
```diff
@@ -1077,11 +1075,8 @@ TempTrajectory GroupedCkfTrajectoryBuilder::backwardFit(TempTrajectory& candidat
   //cout << "firstTsos "<< firstTsos << endl;
   firstTsos.rescaleError(10.);
   //TrajectoryContainer bwdFitted(fitter.fit(fwdTraj.seed(),fwdTraj.recHits(),firstTsos));
-  Trajectory&& bwdFitted = fitter.fitOne(
-      TrajectorySeed(
-          PTrajectoryStateOnDet(), TrajectorySeed::recHitContainer(), oppositeDirection(candidate.direction())),
-      fwdTraj.recHits(),
-      firstTsos);
+  Trajectory&& bwdFitted =
```
fwyzard, line 1078: why `Trajectory&&` ?

(I understand it was like that in the original code, but I'm curious, so if anybody has any comments about it...)

RecoTracker/TrackProducer/plugins/FakeTrackProducers.cc
```diff
@@ -33,6 +34,15 @@
 #include "Geometry/Records/interface/TrackerDigiGeometryRecord.h"
 #include "Geometry/Records/interface/TrackerTopologyRcd.h"
 #include "MagneticField/Records/interface/IdealMagneticFieldRecord.h"
+#include "DataFormats/TrackingRecHit/interface/TrackingRecHit.h"
+
+namespace {
+  auto getHits(const TrajectorySeed &seed) { return seed.recHits(); }
```
fwyzard, line 40: does this make a copy ?

RecoTracker/TrackProducer/plugins/FakeTrackProducers.cc
```diff
@@ -33,6 +34,15 @@
...
+  auto getHits(const TrackCandidate &seed) {
```
fwyzard, line 41: call me old fashioned, but could you spell the return type explicitly ?

fwyzard (commented):
Could you double check that the changes do not introduce any spurious copies of the rechits ?

VinInn:
-1
Guess WHY it was decided NOT to expose the Abominable old-fashion OwnVector (there should be a good reason why such a thing does not exist in std)
It is just for storage (make old root happy in particular), not to be used by humans.

Maybe you want to reimplement it (well, just substitute it in the enclosing class) in terms of a plain vector of unique_ptr (convince root to store it) and try to forbid the implicit "cloning"  (and removing all those tricky iterators).
The clients shall get a plain const-reference to  a unique_ptr: better they know what they really want to do with it (if not, it will not compile, no hidden copy/cloning)

guitargeek:
Hi @fwyzard and @VinInn, I understand your concerns very well. What do you think about returning a `edm::Range<>`? It does not contain more information than the oiginal pair (hence not exposing the OwnVector more than before) and can also easily changed when the recHits would be a subset of the OwnVector. Maybe I can address both your concerns with this? I have already done that before in https://github.com/cms-sw/cmssw/pull/25591 for TrackExtraBase.h:

https://github.com/cms-sw/cmssw/pull/25591/files#diff-bf20a87d4f4072f2395ba69e6591cf9bR48

Another option might be to return `hits_.data()`, which is just a `std::vector<TrackingRecHit*>`, but this seems dangerous to me because it's not a vector of `TrackingRecHit const*`.

In any case I would prefer not to change the class members in order to keep incompatibilities with old versions at a minimum.

VinInn:
ok for the TrackingRecHitRange as in Track. At least it is consistent.

RecoTracker/CkfPattern/plugins/GroupedCkfTrajectoryBuilder.cc
guitargeek: I don't know, to me it seems unnecessary. I didn't change it in my upcoming commit that address the other comments though, because maybe it has a purpose I'm not aware of.

fwyzard:
> What do you think about returning a `edm::Range<edm::OwnVector<TrackingRecHit>::const_iterator>`?

Sounds like a good compromise.

slava77:
@cmsbuild please test

guitargeek:
Unfortunately I just noticed a mistake I made in the last commit that had to be fixed, but by now I went over all the changes twice and couldn't spot more mistakes.

slava77:
@cmsbuild please test

alja:
+1

jfernan2:
@guitargeek  this PR is creating a lot modfications in the Tracking directories
https://cmssdt.cern.ch/SDT/jenkins-artifacts/baseLineComparisons/CMSSW_11_2_X_2020-05-25-1100+ad7c88/36619/dqm-histo-comparison-summary.html
In many cases moving the plots to unknown tags, I am not a tracker/tracking expert, but I wonder if this is expected or understood
FYI @mmusich  @mtosi

fwyzard:
This should have no impact at all on the reconstruction and DQM results.

mmusich:
-1
this is not expected. Please fix

mmusich:
@jfernan2 thanks for the pointer. Can you please hold this PR?

jfernan2:
hold

guitargeek:
Hi @jfernan2, thanks for the nice catch! I got to make it a habit to also check the DQM output and not only the reco comparisons... Fortunately it was very clear from the histograms where I made the mistake, fix will follow.

jfernan2:
unhold

jfernan2:
please test

jfernan2:
+1
DQM changes are gone now

guitargeek:
Unfortunately, this had to be rebased. There is also an additional code-format commit in the end which has not been there before, because apparently the way the code is automatically formatted  has been changed in the meantime.

RecoTracker/TrackProducer/plugins/FakeTrackProducers.cc
```diff
@@ -33,6 +34,17 @@
 #include "Geometry/Records/interface/TrackerDigiGeometryRecord.h"
 #include "Geometry/Records/interface/TrackerTopologyRcd.h"
 #include "MagneticField/Records/interface/IdealMagneticFieldRecord.h"
+#include "DataFormats/TrackingRecHit/interface/TrackingRecHit.h"
+
+namespace {
+  edm::Range<edm::OwnVector<TrackingRecHit>::const_iterator> getHits(const TrajectorySeed &seed) {
```
perrotta, line 40: Why not ```TrajectorySeed::RecHitRange```?

RecoTracker/TrackProducer/plugins/FakeTrackProducers.cc
```diff
@@ -33,6 +34,17 @@
...
+    return seed.recHits();
+  }
+  edm::Range<edm::OwnVector<TrackingRecHit>::const_iterator> getHits(const TrackCandidate &seed) {
```
perrotta, line 43: (same here)

DataFormats/MuonSeed/src/L2MuonTrajectorySeed.cc
```diff
@@ -12,15 +12,15 @@ L2MuonTrajectorySeed::L2MuonTrajectorySeed() : TrajectorySeed() {}
 
 // Constructor
 L2MuonTrajectorySeed::L2MuonTrajectorySeed(PTrajectoryStateOnDet const& ptsos,
-                                           recHitContainer const& rh,
+                                           RecHitContainer const& rh,
```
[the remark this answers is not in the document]
perrotta: Then perhaps ```TrajectorySeed::RecHitContainer``` explicitely

RecoTracker/CkfPattern/plugins/GroupedCkfTrajectoryBuilder.cc
```diff
@@ -836,15 +835,13 @@ void GroupedCkfTrajectoryBuilder::rebuildSeedingRegion(const TrajectorySeed& see
   auto hitCloner = static_cast<TkTransientTrackingRecHitBuilder const*>(hitBuilder())->cloner();
   KFTrajectoryFitter fitter(backwardPropagator(seed), &updator(), &estimator(), 3, nullptr, &hitCloner);
   //
-  TrajectorySeed::range rseedHits = seed.recHits();
   std::vector<const TrackingRecHit*> seedHits;
   //seedHits.insert(seedHits.end(), rseedHits.first, rseedHits.second);
```
perrotta, line 839: These commented out lines become now obsolete: better to remove.
If you really want to keep them, then please resurrect (commented out) also 
```C++
  TrajectorySeed::range rseedHits = seed.recHits();
```

perrotta (commented):
(By the way, it looks like it has to be rebased once again..) -> NO, now it merges, no need to rebase!

DataFormats/MuonSeed/src/L2MuonTrajectorySeed.cc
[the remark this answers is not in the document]
guitargeek: Yes, that was my intention. Thanks for pointing out these other typedefs, I'll remove them as well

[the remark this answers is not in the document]
guitargeek: Hi, I just realized that `TrackCandidate` does not derive from `TrajectorySeed`. In that case I think it makes sense that it keeps it's own typedef for the rechit collection. But for the TrajectorySeed derived classes I remove it for sure!

RecoTracker/CkfPattern/plugins/GroupedCkfTrajectoryBuilder.cc
guitargeek: I think the commented out code can be removed. Just a loop that fills a vector from each element of the original one is not very interesting to keep IMO

RecoTracker/TrackProducer/plugins/FakeTrackProducers.cc
guitargeek: Sure, it's better!

RecoTracker/TrackProducer/plugins/FakeTrackProducers.cc
guitargeek: Sure, it's better indeed!

perrotta:
please test

perrotta:
+1
- Code changes are meant to simplify the TrajectorySeed interface to recHits; they are made according to the lines proposed in the PR description, and modified in the following PR review (also/mostly from @fwyzard and @VinInn )
- Jenkins tests pass and show no differences inoutput wrt baseline

jfernan2:
+1

Martin-Grunewald:
+1

silviodonato:
Any objections? 
analysis:  @santocch
visualization:  @Dr15Jones @alja @makortel
fastsim:  @civanch @lveldere @mdhildreth @sbein @ssekmen
simulation:  @civanch @mdhildreth

santocch:
+1

alja:
+1

civanch:
+1

silviodonato:
+1

PR #29975, opened 2020-05-25
DD4hep geometry test: update dd4hep tentative workflow

fabiocos:
#### PR description:

Add step2 and step3 to the test chain of standard workflows using DD4hep as input. Please note that at this point the proposed steps are nothing but the output of ```runTheMatrix.py --what upgrade -l 11607.0 --command "--geometry DD4hepExtended2021 --era Run3_dd4hep" ```. 

#### PR validation:

The steps fail, as known, but they can be routinely used to test progresses and issues. A dryRun of the suggested matrix command above shows that the cmsDriver commands generated are those in the script, as desired.

### DIFF
--- a/Geometry/CMSCommonData/test/dd4hep/runDD4hepTestWF.sh
+++ b/Geometry/CMSCommonData/test/dd4hep/runDD4hepTestWF.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/sh -e
 
 set -x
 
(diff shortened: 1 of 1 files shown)


fabiocos:
please test

fabiocos:
@bsunanda @slomeo @cvuosalo @ianna @civanch as we were discussing last Friday

kpedro88:
@fabiocos maybe it's more maintainable if the test script just directly calls the matrix command, rather than copy/pasting the resulting cmsDriver commands?

fabiocos:
@kpedro88 my idea is that this script disappears as soon as we are able to run a real matrix workflow, the idea was more to have steps separately available for tests in the intermediate period, as discussed at the simulation meeting. As this was explicitly asked, and I saw ad hoc test configurations appearing, I considered useful to have a "standard" set of step workflows available in parallel.

Now that we have geometry files and Eras defined, I could even just remove the script. In principle even the recent DD4hep eras are not needed, one may just invoke the ```dd4hep``` process modifier separately.

kpedro88:
@fabiocos another idea: a "special" workflow could be added in the upgrade matrix, using the existing infrastructure. If that workflow isn't forwarded to the regular matrix, it won't run in IB tests, so it's okay if it's broken at first. This would give developers an easy test to run privately.

cvuosalo:
+1

fabiocos:
hold

fabiocos:
superseded by #30003

PR #29976, opened 2020-05-25
bug fix to runTauIdMVA.py tool

mbluj:
#### PR description:

This PR fixes bugs introduced to `runTauIdMVA.py` tool during implementation of the new tauID dataformat #28417.

#### PR validation:

This tool is not used in standard workflows except miniAOD (to include deepTauID) which uses its non-affected part. Therefore, the fix was tested with [RecoTauTag/RecoTau/test/runDeepTauIDsOnMiniAOD.py](https://github.com/cms-sw/cmssw/blob/master/RecoTauTag/RecoTau/test/runDeepTauIDsOnMiniAOD.py) test configuration. (Actaully, the bugs have been discovered when running this configuration). 
=> This test configuration is added to units tests in this PR.

Q: It will be good to have this fix already in 11_1_X series - does it require a backport, i.e. if master is already switched to 11_2_X?

### DIFF
--- a/RecoTauTag/RecoTau/python/tools/runTauIdMVA.py
+++ b/RecoTauTag/RecoTau/python/tools/runTauIdMVA.py
@@ -90,8 +90,8 @@ class TauIDEmbedder(object):
             if debug: print ("is_above_cmssw_version:", True)
             return True
 
-    def tauIDMVAinputs(module, wp):
-        return cms.PSet(inputTag = self.cms.InputTag(module), workingPointIndex = self.cms.int32(-1 if wp=="raw" else -2 if wp=="category" else getattr(self.process, module).workingPoints.index(wp)))
+    def tauIDMVAinputs(self, module, wp):
+        return self.cms.PSet(inputTag = self.cms.InputTag(module), workingPointIndex = self.cms.int32(-1 if wp=="raw" else -2 if wp=="category" else getattr(self.process, module).workingPoints.index(wp)))
 
     def loadMVA_WPs_run2_2017(self):
         if self.debug: print ("loadMVA_WPs_run2_2017: performed")
@@ -172,7 +172,7 @@ class TauIDEmbedder(object):
                         variable = self.cms.string("pt"),
                     )
                 ),
-                workingPoints = cms.vstring(
+                workingPoints = self.cms.vstring(
                     "_WPEff95",
                     "_WPEff90",
                     "_WPEff80",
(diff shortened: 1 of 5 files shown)


RecoTauTag/RecoTau/python/tools/runTauIdMVA.py
```diff
@@ -172,7 +172,7 @@ def runTauID(self):
                         variable = self.cms.string("pt"),
                     )
                 ),
-                workingPoints = cms.vstring(
+                workingPoints = self.cms.vstring(
```
slava77, line 175: was it broken or is the "self." superfluous in this case?
(same question for many other cases).

slava77:
> Q: It will be good to have this fix already in 11_1_X series - does it require a backport, i.e. if master is already switched to 10_2_X?

there are no automated backports or forwardports. So, if it's needed in 11_1_X, there should be a PR for 11_1_X.

slava77:
> This tool is not used in standard workflows except miniAOD (to include deepTauID) which uses its non-affected part. Therefore, the fix was tested with [RecoTauTag/RecoTau/test/runDeepTauIDsOnMiniAOD.py](https://github.com/cms-sw/cmssw/blob/master/RecoTauTag/RecoTau/test/runDeepTauIDsOnMiniAOD.py) test configuration. (Actaully, the bugs have been discovered when running this configuration).

It sounds like runTauIdMVA needs a unit test (some script that runs this tool mainly for integrity checks). Please add or enable one in RecoTauTag/RecoTau/test/runtests.sh

RecoTauTag/RecoTau/python/tools/runTauIdMVA.py
mbluj: The "self." is mandatory here and other places as the `TauIDEmbedder` python class is designed such that the global `cms` is a member of the class. The "self." was not added/was removed when tool have been updated to new tau discriminant dataformat with its configuration.

mbluj:
> 
> 
> > Q: It will be good to have this fix already in 11_1_X series - does it require a backport, i.e. if master is already switched to 10_2_X?
> 
> there are no automated backports or forwardports. So, if it's needed in 11_1_X, there should be a PR for 11_1_X.

OK, thank you. 
I assume that the PR to 11_1_X can be based on the same development branch as one to master (11_2_X) as the two does not differ much. Is it correct?

mbluj:
> 
> 
> > This tool is not used in standard workflows except miniAOD (to include deepTauID) which uses its non-affected part. Therefore, the fix was tested with [RecoTauTag/RecoTau/test/runDeepTauIDsOnMiniAOD.py](https://github.com/cms-sw/cmssw/blob/master/RecoTauTag/RecoTau/test/runDeepTauIDsOnMiniAOD.py) test configuration. (Actaully, the bugs have been discovered when running this configuration).
> 
> It sounds like runTauIdMVA needs a unit test (some script that runs this tool mainly for integrity checks). Please add or enable one in RecoTauTag/RecoTau/test/runtests.sh

I agree. Could you advice which input file (MiniAOD, eg. ttbar) can be safety used for the tests, please? Currently, one file from Fall17 production is used, but probably it is better to use a relval one.

RecoTauTag/RecoTau/python/tools/runTauIdMVA.py
slava77: what was the rationale for this?
it seems rather burdensome to have to prefix almost everything with "self"

slava77:
> > It sounds like runTauIdMVA needs a unit test (some script that runs this tool mainly for integrity checks). Please add or enable one in RecoTauTag/RecoTau/test/runtests.sh
> 
> I agree. Could you advice which input file (MiniAOD, eg. ttbar) can be safety used for the tests, please? Currently, one file from Fall17 production is used, but probably it is better to use a relval one.

something that's already used in other unit tests should be fine.
Minimally, the test could be even without any actual running so that at least the parsing is checked

RecoTauTag/RecoTau/python/tools/runTauIdMVA.py
mbluj: @kandrosov, do you remember why the global `cms` (holding definitions of classes used for the python CMSSW configuration) is a member of the `TauIDEmbedder` class?

Regardless, the main intention of the PR is to quickly fix the bug rather than improving the tool.

slava77: well, if the missing addition of ```self.``` was the leading origin for the bug
I'd rather we do not keep the error-prone implementation.
It's perhaps OK to postpone the update that does not require the ```self.``` for a later PR; however, I think that a unit test is needed now with this PR.

mbluj: It is now added to unit tests. 
I hope we will hear soon about a rationale of usage "self." from Konstantin who (as far as I remember) introduced it.

slava77:
@cmsbuild please test

RecoTauTag/RecoTau/python/tools/runTauIdMVA.py
mbluj: After some thinking, I do not see any rationale of the current design and I agree it is error probe. Therefore, I will commit changes with direct use of "cms".
@kandrosov, please let me know if if there is a real need to keep current design - if it is present I am fine with reverting it.

slava77:
@cmsbuild please test

RecoTauTag/RecoTau/python/tools/runTauIdMVA.py
kandrosov: @mbluj This design choice has been made before my time (for example, it is already present in the code referenced in quite old recommendations for 2017 https://github.com/greyxray/TauAnalysisTools/blob/CMSSW_9_4_X_tau-pog_RunIIFall17/TauAnalysisTools/python/runTauIdMVA.py#L29). As far as I see, it can safely be replaced by `import FWCore.ParameterSet.Config as cms`

mbluj: OK, thanks for comment and check. 
It has just been replaced.

santocch:
+1

slava77:
+1

for https://github.com/cms-sw/cmssw/pull/29976	7aaf1ab 
- jenkins tests pass and comparisons with the baseline show show no (relevant) differences; the added unit test passes OK and takes about 30 sec to run, which seems acceptable

slava77:
@peruzzim 
xpog signature is missing for this PR.
Please check and comment or sign.
Thank you.

silviodonato:
merge
cc: @peruzzim

PR #29979, opened 2020-05-25
drop type specs in RecoLocalCalo/{CaloTowersCreator,Configuration,EcalRecProducers}

slava77:
Instead of modifying parameters with full type specs, which can be interpreted as an insertion of a new parameter,
it is a safer way to protect from parameter name mistakes and will also help in possible parameter migrations.

Additionally, in a few places this also moves parameter modifications in the ```clone``` method calls to the argument list instead of a separate line.

This PR is an example for the planned migration/update task.


slava77:
@cmsbuild please test

slava77:
+1

for https://github.com/cms-sw/cmssw/pull/29979	e9b40ab
- jenkins tests confirmed no (relevant) differences

silviodonato:
+1

PR #29994, opened 2020-05-26
Add safety check to spot negative jet energy correction factors

ahinzmann:
#### PR description:

Add a safety check when evaluating jet energy correction factors to spot cases where the functions yield unphysical negative corrections.
A warning is printed and the value is truncated to be small positive.
This should help spotting cases when calibration functions in the conditions are not fully positive defined (as they should).

#### PR validation:

runTheMatrix -l limited

To all release down to 10_6 to protect UL/Run3/Phase-2 studies.


santocch:
+1

silviodonato:
Do you have any comments? @ggovi

silviodonato:
@ggovi Do you have any comments?

silviodonato:
merge
cc: @ggovi

PR #30010, opened 2020-05-28
moved hardcoded correction boundary values to configuration file in EcalLaserCondTools

grasph:
Two parameters (correction boudaries) were hardcoded in the module EcalLaserCondTools. This PR makes them configurable thanks to the work of @fcouderc . Application of correction boundaries has also been simplified.

The code is not used in the reco and simulation workflows.

### DIFF
--- a/CalibCalorimetry/EcalTrivialCondModules/src/EcalLaserCondTools.cc
+++ b/CalibCalorimetry/EcalTrivialCondModules/src/EcalLaserCondTools.cc
@@ -33,8 +33,7 @@ EcalLaserCondTools::EcalLaserCondTools(const edm::ParameterSet& ps)
       fromTime_(ps.getParameter<int>("fromTime")),
       toTime_(ps.getParameter<int>("toTime")),
       minP_(ps.getParameter<double>("transparencyMin")),
-      maxP_(ps.getParameter<double>("transparencyMax"))
-{
+      maxP_(ps.getParameter<double>("transparencyMax")) {
   ferr_ = fopen("corr_errors.txt", "w");
   fprintf(ferr_, "#t1\tdetid\tp1\tp2\tp3");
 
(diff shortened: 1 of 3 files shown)


christopheralanwest:
please test

pohsun:
+1

CalibCalorimetry/EcalTrivialCondModules/src/EcalLaserCondTools.cc
```diff
@@ -233,14 +228,7 @@ void EcalLaserCondTools::processIov(CorrReader& r, int t1, int t2[EcalLaserCondT
...
            << "p3 = " << corr.p3 << "\n";
     }
-    if (detid.subdetId() == EcalBarrel) {
-      if (corr.p1 < 0.55)
-        corr.p1 = .985;
-      if (corr.p2 < 0.55)
-        corr.p2 = .985;
-      if (corr.p3 < 0.55)
-        corr.p3 = .985;
-    }
```
silviodonato, line 243: Are you removing these lines on purpose? If yes, please add few words about this change in the PR description.

CalibCalorimetry/EcalTrivialCondModules/src/EcalLaserCondTools.cc
```diff
@@ -31,7 +31,9 @@ EcalLaserCondTools::EcalLaserCondTools(const edm::ParameterSet& ps)
       skipIov_(ps.getParameter<int>("skipIov")),
       nIovs_(ps.getParameter<int>("nIovs")),
       fromTime_(ps.getParameter<int>("fromTime")),
-      toTime_(ps.getParameter<int>("toTime")) {
+      toTime_(ps.getParameter<int>("toTime")),
+      minP_(ps.getParameter<double>("transparencyMin")),
+      maxP_(ps.getParameter<double>("transparencyMax")) {
```
silviodonato, line 36: Adding new parameters means that current py files without the new parameters will crash in future CMSSW release and that the newer py files with the new parameters will crash in the older CMSSW release. I would suggest to introduce the `fillDescriptions` to set default values in order to preserve the compatibility between the releases (https://twiki.cern.ch/twiki/bin/view/CMSPublic/SWGuideConfigurationValidationAndHelp#An_Example)

silviodonato (commented):
Looking at "6 – Packaging Rules" https://cms-sw.github.io/cms_coding_rules.html it seems `EcalLaserCondTools` should go in a single `.cc` file in `/plugins` directory.

silviodonato:
The main comment is https://github.com/cms-sw/cmssw/pull/30010#discussion_r432307028. You can consider the other comments as "nice to have" features going beyond the purpose of this PR.

CalibCalorimetry/EcalTrivialCondModules/src/EcalLaserCondTools.cc
grasph: I don't understand. I expect that by "crash" you meant an exception, not a core dump.

The parameters are tracked parameters, so an exception will already be raised if they are missing in the configuration file.

CalibCalorimetry/EcalTrivialCondModules/src/EcalLaserCondTools.cc
grasph: It's on purpose. PR description updated.

CalibCalorimetry/EcalTrivialCondModules/src/EcalLaserCondTools.cc
grasph: In addition this "fillDescription" is a source of inconsistency in the code. The new of the parameters will appear twice with the risk to update at one place and forget in another place or using a different spelling.  I have missed the rationally of this new method.

silviodonato:
+1

PR #30016, opened 2020-05-28
Introduce DeepDoubleX V2 

andrzejnovak:
This PR introduces V2 of DeepDoubleX. Required changes w.r.t. V1 of DeepDoubleX producer include adding a new group of variables, add more variables to existing groups and reordering some of them. 

Input x-check in training FW and CMSSW - :white_check_mark:.
image

Evaluation x-check - :white_check_mark:

image

### DIFF
--- a/DataFormats/BTauReco/src/classes_def.xml
+++ b/DataFormats/BTauReco/src/classes_def.xml
@@ -409,16 +409,19 @@
     <version ClassVersion="3" checksum="386672900"/>
     <version ClassVersion="4" checksum="113477759"/>
   </class>
-  <class name="btagbtvdeep::SecondaryVertexFeatures" ClassVersion="3">
+  <class name="btagbtvdeep::SecondaryVertexFeatures" ClassVersion="4">
+   <version ClassVersion="4" checksum="3987593676"/>
     <version ClassVersion="3" checksum="3610780843"/>
   </class>
   <class name="btagbtvdeep::ShallowTagInfoFeatures" ClassVersion="3">
     <version ClassVersion="3" checksum="273372007"/>
   </class>
-  <class name="btagbtvdeep::NeutralCandidateFeatures" ClassVersion="3">
+  <class name="btagbtvdeep::NeutralCandidateFeatures" ClassVersion="4">
+   <version ClassVersion="4" checksum="3657612864"/>
     <version ClassVersion="3" checksum="938442516"/>
   </class>
-  <class name="btagbtvdeep::ChargedCandidateFeatures" ClassVersion="4">
+  <class name="btagbtvdeep::ChargedCandidateFeatures" ClassVersion="5">
+   <version ClassVersion="5" checksum="3385708923"/>
     <version ClassVersion="4" checksum="2173116472"/>
     <version ClassVersion="3" checksum="1302021141"/>
   </class>
@@ -432,7 +435,8 @@
     <version ClassVersion="3" checksum="2487956635"/>
     <version ClassVersion="4" checksum="3827695028"/>
   </class>
-    <class name="btagbtvdeep::DeepDoubleXFeatures" ClassVersion="3">
+    <class name="btagbtvdeep::DeepDoubleXFeatures" ClassVersion="4">
+     <version ClassVersion="4" checksum="3057931654"/>
     <version ClassVersion="3" checksum="1764960743"/>
   </class>
   <class name="btagbtvdeep::BoostedDoubleSVTagInfoFeatures" ClassVersion="3">
--- a/RecoBTag/FeatureTools/interface/ChargedCandidateConverter.h
+++ b/RecoBTag/FeatureTools/interface/ChargedCandidateConverter.h
@@ -30,7 +30,12 @@ namespace btagbtvdeep {
       trackSip3dVal = -trackSip3dVal;
     }
 
+    c_pf_features.deltaR = reco::deltaR(*c_pf, jet);
     c_pf_features.ptrel = catch_infs_and_bound(c_pf->pt() / jet.pt(), 0, -1, 0, -1);
+    c_pf_features.ptrel_noclip = c_pf->pt() / jet.pt();
+    c_pf_features.erel = c_pf->energy() / jet.energy();
+    const float etasign = jet.eta() > 0 ? 1 : -1;
+    c_pf_features.etarel = etasign * (c_pf->eta() - jet.eta());
 
     c_pf_features.btagPf_trackEtaRel = catch_infs_and_bound(track_info.getTrackEtaRel(), 0, -5, 15);
     c_pf_features.btagPf_trackPtRel = catch_infs_and_bound(track_info.getTrackPtRel(), 0, -1, 4);
@@ -45,6 +50,26 @@ namespace btagbtvdeep {
     c_pf_features.btagPf_trackJetDistVal = catch_infs_and_bound(track_info.getTrackJetDistVal(), 0, -20, 1);
 
     c_pf_features.drminsv = catch_infs_and_bound(drminpfcandsv, 0, -1. * jetR, 0, -1. * jetR);
+
+    // subjet related
+    const auto* patJet = dynamic_cast<const pat::Jet*>(&jet);
+    if (!patJet) {
+      throw edm::Exception(edm::errors::InvalidReference) << "Input is not a pat::Jet.";
+    }
+
+    if (patJet->nSubjetCollections() > 0) {
+      auto subjets = patJet->subjets();
+      // sort by pt
+      std::sort(subjets.begin(), subjets.end(), [](const edm::Ptr<pat::Jet>& p1, const edm::Ptr<pat::Jet>& p2) {
+        return p1->pt() > p2->pt();
+      });
+      c_pf_features.drsubjet1 = !subjets.empty() ? reco::deltaR(*c_pf, *subjets.at(0)) : -1;
+      c_pf_features.drsubjet2 = subjets.size() > 1 ? reco::deltaR(*c_pf, *subjets.at(1)) : -1;
+    } else {
+      // AK4 jets don't have subjets
+      c_pf_features.drsubjet1 = -1;
+      c_pf_features.drsubjet2 = -1;
+    }
   }
 
   void packedCandidateToFeatures(const pat::PackedCandidate* c_pf,
--- a/RecoBTag/FeatureTools/interface/NeutralCandidateConverter.h
+++ b/RecoBTag/FeatureTools/interface/NeutralCandidateConverter.h
@@ -30,8 +30,31 @@ namespace btagbtvdeep {
                                         const float& drminpfcandsv,
                                         const float& jetR,
                                         NeutralCandidateFeatures& n_pf_features) {
+    const auto* patJet = dynamic_cast<const pat::Jet*>(&jet);
+
+    if (!patJet) {
+      throw edm::Exception(edm::errors::InvalidReference) << "Input is not a pat::Jet.";
+    }
+    // Do Subjets
+    if (patJet->nSubjetCollections() > 0) {
+      auto subjets = patJet->subjets();
+      // sort by pt
+      std::sort(subjets.begin(), subjets.end(), [](const edm::Ptr<pat::Jet>& p1, const edm::Ptr<pat::Jet>& p2) {
+        return p1->pt() > p2->pt();
+      });
+      n_pf_features.drsubjet1 = !subjets.empty() ? reco::deltaR(*n_pf, *subjets.at(0)) : -1;
+      n_pf_features.drsubjet2 = subjets.size() > 1 ? reco::deltaR(*n_pf, *subjets.at(1)) : -1;
+    } else {
+      n_pf_features.drsubjet1 = -1;
+      n_pf_features.drsubjet2 = -1;
+    }
+
+    // Jet relative vars
     n_pf_features.ptrel = catch_infs_and_bound(n_pf->pt() / jet.pt(), 0, -1, 0, -1);
+    n_pf_features.ptrel_noclip = n_pf->pt() / jet.pt();
     n_pf_features.deltaR = catch_infs_and_bound(reco::deltaR(*n_pf, jet), 0, -0.6, 0, -0.6);
+    n_pf_features.deltaR_noclip = reco::deltaR(*n_pf, jet);
+    n_pf_features.erel = n_pf->energy() / jet.energy();
     n_pf_features.isGamma = 0;
     if (std::abs(n_pf->pdgId()) == 22)
       n_pf_features.isGamma = 1;
--- a/RecoBTag/FeatureTools/plugins/DeepDoubleXTagInfoProducer.cc
+++ b/RecoBTag/FeatureTools/plugins/DeepDoubleXTagInfoProducer.cc
@@ -22,6 +22,7 @@
 #include "DataFormats/BTauReco/interface/DeepDoubleXTagInfo.h"
 
 #include "RecoBTag/FeatureTools/interface/BoostedDoubleSVTagInfoConverter.h"
+#include "RecoBTag/FeatureTools/interface/NeutralCandidateConverter.h"
 #include "RecoBTag/FeatureTools/interface/ChargedCandidateConverter.h"
 #include "RecoBTag/FeatureTools/interface/JetConverter.h"
 #include "RecoBTag/FeatureTools/interface/SecondaryVertexConverter.h"
@@ -119,6 +120,8 @@ void DeepDoubleXTagInfoProducer::produce(edm::Event& iEvent, const edm::EventSet
 
     // reco jet reference (use as much as possible)
     const auto& jet = jets->at(jet_n);
+    const auto* pf_jet = dynamic_cast<const reco::PFJet*>(&jet);
+    const auto* pat_jet = dynamic_cast<const pat::Jet*>(&jet);
 
     edm::RefToBase<reco::Jet> jet_ref(jets, jet_n);
     if (jet.pt() > min_jet_pt_) {
@@ -175,7 +178,8 @@ void DeepDoubleXTagInfoProducer::produce(edm::Event& iEvent, const edm::EventSet
       math::XYZVector jet_dir = jet.momentum().Unit();
       GlobalVector jet_ref_track_dir(jet.px(), jet.py(), jet.pz());
 
-      std::vector<btagbtvdeep::SortingClass<size_t>> c_sorted;
+      std::vector<btagbtvdeep::SortingClass<size_t>> c_sorted, n_sorted;
+      std::vector<int> n_indexes;
 
       // to cache the TrackInfo
       std::map<unsigned int, btagbtvdeep::TrackInfoBuilder> trackinfos;
@@ -205,9 +209,9 @@ void DeepDoubleXTagInfoProducer::produce(edm::Event& iEvent, const edm::EventSet
         }
       }
 
+      std::sort(daughters.begin(), daughters.end(), [](const auto& a, const auto& b) { return a->pt() > b->pt(); });
       for (unsigned int i = 0; i < daughters.size(); i++) {
         auto const* cand = daughters.at(i);
-
         if (cand) {
           // candidates under 950MeV (configurable) are not considered
           // might change if we use also white-listing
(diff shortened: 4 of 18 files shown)


andrzejnovak:
@slava77

RecoBTag/FeatureTools/interface/ChargedCandidateConverter.h
```diff
@@ -45,6 +51,18 @@ namespace btagbtvdeep {
...
+
+    // subjet related
+    const auto *patJet = dynamic_cast<const pat::Jet *>(&jet);
+    if (!patJet) {
+      throw edm::Exception(edm::errors::InvalidReference) << "Input is not a pat::Jet.";
+    }
+    auto subjets = patJet->subjets();
+    std::sort(subjets.begin(), subjets.end(), [](const edm::Ptr<pat::Jet> &p1, const edm::Ptr<pat::Jet> &p2) {
+      return p1->pt() > p2->pt();
+    });  // sort by pt
```
perrotta, line 63: If you move the comment above, quite likely the line won't get split so oddly

RecoBTag/FeatureTools/interface/NeutralCandidateConverter.h
```diff
@@ -30,8 +30,23 @@ namespace btagbtvdeep {
...
                                         const float& jetR,
                                         NeutralCandidateFeatures& n_pf_features) {
+
+    const auto *patJet = dynamic_cast<const pat::Jet *>(&jet);
+    
+    if (!patJet) {
+      throw edm::Exception(edm::errors::InvalidReference) << "Input is not a pat::Jet.";
+    }
+    // Do Subjets
+    auto subjets = patJet-> subjets();;   // sort by pt
```
perrotta, line 40: remove one semicolon

RecoBTag/FeatureTools/plugins/DeepDoubleXTagInfoProducer.cc
```diff
@@ -220,48 +224,66 @@ void DeepDoubleXTagInfoProducer::produce(edm::Event& iEvent, const edm::EventSet
...
+            reco_ptr = pf_jet->getPFConstituent(i);
+          } else if (pat_jet && reco_cand) {
+            reco_ptr = pat_jet->getPFConstituent(i);
+          }
           // get PUPPI weight from value map
           float puppiw = 1.0;  // fallback value
-
-          float drminpfcandsv = btagbtvdeep::mindrsvpfcand(svs_unsorted, cand, jet_radius_);
-
+          // if (reco_cand && use_puppi_value_map_) {
```
perrotta, line 269: Commented out code can be remove if not needed (otherwise add one more comment line which specifies why it is needed commented out)

RecoBTag/FeatureTools/plugins/DeepDoubleXTagInfoProducer.cc
```diff
@@ -288,9 +310,27 @@ void DeepDoubleXTagInfoProducer::produce(edm::Event& iEvent, const edm::EventSet
...
+                  packed_cand, *pat_jet, drminpfcandsv, static_cast<float>(jet_radius_), n_pf_features);
+            } else if (reco_cand) {
+              btagbtvdeep::recoCandidateToFeatures(
+                  reco_cand, jet, drminpfcandsv, static_cast<float>(jet_radius_), puppiw, n_pf_features);
+                  
+            }
           }
         }
       }
+      // DEBUG: Inspect features in order supplied to tagger
```
perrotta, line 330: Could this be replaced by a LogDebug, or simply removed if not needed?

RecoBTag/ONNXRuntime/plugins/DeepDoubleXONNXJetTagsProducer.cc
```diff
@@ -141,12 +146,33 @@ void DeepDoubleXONNXJetTagsProducer::produce(edm::Event& iEvent, const edm::Even
       for (const auto& taginfo : *tag_infos) {
         if (!taginfo.features().empty()) {
           make_inputs(idx, taginfo);
+          etas_.push_back(taginfo.jet()->eta()); //DEBUG
```
perrotta, line 149: Trailing underscore are normally reserved to class members

RecoBTag/ONNXRuntime/plugins/DeepDoubleXONNXJetTagsProducer.cc
```diff
@@ -141,12 +146,33 @@ void DeepDoubleXONNXJetTagsProducer::produce(edm::Event& iEvent, const edm::Even
...
           ++idx;
         }
       }
 
       // run prediction
       outputs = globalCache()->run(input_names_, data_, output_names_, batch_size)[0];
+
+      // DEBUG: Dump inputs to file
```
perrotta, line 157: Has this debug to remain in the producer?

RecoBTag/ONNXRuntime/plugins/DeepDoubleXONNXJetTagsProducer.cc
```diff
@@ -189,34 +215,35 @@ void DeepDoubleXONNXJetTagsProducer::make_inputs(unsigned i_jet, const reco::Dee
   start = ptr;
   const auto& tag_info_features = features.tag_info_features;
   *ptr = tag_info_features.jetNTracks;
-  *(++ptr) = tag_info_features.jetNSecondaryVertices;
+  //*(++ptr) = tag_info_features.jetNSecondaryVertices;
```
perrotta, line 218: Do all these removed features need to stay here? If so, please add another comment line which explains why they must stay. Otherwise just remove

perrotta (commented):
Not a real review, just pointing out a few thing that jumped on my eyes while scrolling the differences

perrotta:
> This is a **draft** PR for introducing V2 of DeepDoubleX. Required changes w.r.t. V1 of DeepDoubleX producer include adding a new group of variables, add more variables to existing groups and reordering some of them.
> 
> I would like to ask for feedback on how backcomp to V1 should be treated? Should there be a new producer? It seems like a possibly simpler choice than making all the differences configurable.
> 
> (model file will be factored out before this PR is ready)

What is the purpose of back compatibility with V1? Do you plan to use both V1 and V2 at the same time, or just allow finishing some analysis which already started with V1? That decision has to be taken in the the POG: please summarize here the result of the discussion in the POG.

If the two version must be kept available at the same time, at the first glance it does not seem to me too much complicated to factorize out the differences and make them configurable inside the same producer: code duplication doesn't seem neededd here, and normally better to be avoided.

andrzejnovak:
@perrotta Thanks for getting back to me. This is supposed to supersede V1, I am asking from maintenance and backcomp point of view. From POG POV V2 could replace V1.

RecoBTag/ONNXRuntime/plugins/DeepDoubleXONNXJetTagsProducer.cc
```diff
@@ -128,6 +132,7 @@ void DeepDoubleXONNXJetTagsProducer::produce(edm::Event& iEvent, const edm::Even
     auto batch_size = std::count_if(
         tag_infos->begin(), tag_infos->end(), [](const auto& taginfo) { return !taginfo.features().empty(); });
 
+    std::vector<float> etas_; //DEBUG
```
slava77, line 135: locals should not have a trailing underscore to avoid confusion with data members

RecoBTag/ONNXRuntime/plugins/DeepDoubleXONNXJetTagsProducer.cc
andrzejnovak: I can gate it by a debug_ bool, but I would prefer to keep it here as it helps a lot with catching problems when syncing with the outside framework.

DataFormats/BTauReco/src/classes_def.xml
```diff
@@ -409,16 +409,20 @@
...
+   <version ClassVersion="4" checksum="3987593676"/>
     <version ClassVersion="3" checksum="3610780843"/>
   </class>
   <class name="btagbtvdeep::ShallowTagInfoFeatures" ClassVersion="3">
     <version ClassVersion="3" checksum="273372007"/>
   </class>
-  <class name="btagbtvdeep::NeutralCandidateFeatures" ClassVersion="3">
+  <class name="btagbtvdeep::NeutralCandidateFeatures" ClassVersion="4">
+   <version ClassVersion="4" checksum="3657612864"/>
+   <version ClassVersion="4" checksum="2615117930"/>
```
perrotta, line 421: One of those lines (same ClassVersion, different cheksum) is too much

mrodozov:
test parameters:
- pull_request =https://github.com/cms-data/RecoBTag-Combined/pull/31

mrodozov:
please test

perrotta:
@cmsbuild please test

andrzejnovak:
Ok, maybe it's not necessary to remove the non mass decorrelated versions in this PR

perrotta:
The implementation here adds DeepDoubleX V2 in PAT bTags (PhysicsTools/PatAlgos/python/recoLayer0/bTagging_cff.py) together with the current V1 ones.

My understanding, also after the reco meeting of June 12, was that V2 should have completely replaced V1, instead, and therefore there was no need to save both them. The V1 version should then only remain for back compatibility, but not saved in the official miniAOD.

To do so, one should :
- remove RecoBTag/ONNXRuntime/python/pfMassIndependentDeepDoubleXV2JetTags_cff.py
- replace the V2 parameters with the V1 ones in RecoBTag/ONNXRuntime/python/pfMassIndependentDeepDoubleXJetTags_cff.py

Then, for the possible backport in 10_6_X (for the UL re-miniAOD) one could keep the V1 version as the default one in pfMassIndependentDeepDoubleXJetTags_cff.py, and switch to the V2 one in correspondence of a ```run2_miniAOD_devel``` modifier.

Therefore, please specify here if my understanding was correct, or if you really need to save the tags for  both versions in miniAOD outputs instead.
If you don't need to save them both, please apply the modifications suggested above. (If you do so, you won't need to apply modifications for nanoAOD, while otherwise I'd expect they should also get updated to save both versions at the same time)
versions

andrzejnovak:
@perrotta we would only like to replace the default after doing performance measurements, so after this PR the default should still be unchanged. I suppose it would be too much of a mess to have temporarily both V1 and V2 and then replacing V1 with V2.

perrotta:
> @perrotta we would only like to replace the default after doing performance measurements, so after this PR the default should still be unchanged. I suppose it would be too much of a mess to have temporarily both V1 and V2 and then replacing V1 with V2.

But having (temporarily) both V1 and V2 is exactly what you are implementing in this PR!
Puzzled...

andrzejnovak:
Yes, that's what it does in the current state. You're advocating against that right?

perrotta:
> Yes, that's what it does in the current state. You're advocating against that right?

It depends about your plans for "performance measurements": are you expecting some central production with the master release in which you can compare event by event the V1 wrt V2 outputs? If so, it can make sense to have (temporarily) both them at the same time in miniAOD and then revert to a single one in a future follow up PR.

But if you instead plan to make some private production yourself, there is no need to add six extra taggers in the release: just having the code and the configurations for doing it can be sufficient for your needs.

Please let us know about the validation and deployment plans.

In any case, for the possible 10_6 backport (for which time is getting short...) the performance results must be available, and a decision based on them must be taken. Please let us also know about possible ETA for validation and SF derivation.

andrzejnovak:
Is there a way to make V2 conditional on era, as in the whole `V2JetTag`, instead of modifying V1? 

We will be making private production for validation, but wouldn't want to run it twice with the era on/off.
.
SF timescale is maybe around a month or two. I agree that a **replacement** of V1 should be contingent of SF measurement, but **adding** V2 shouldn't. Maybe then keeping both as is currently implemented in the PR is best. The PR is adding 3 new taggers (DDX V2), but with the understanding that on a timescale of several months we would remove 6 (nominal and decorrelated DDX V1).

perrotta:
If I am not wrong, the addition of the new taggers to miniAOD output applies in
PhysicsTools/PatAlgos/python/slimming/applyDeepBtagging_cff.py

Therefore, either a modifier can be added in that file, or even simpler the new lines in that file are only added for your private production, in any case leaving the default unchanged until you are ready to switch,

In any case, I think that a simple check that all your gymn did not disrupt the current V1 could be removing the V2 results, let the test run in this PR, and verify that there are no differences at all, as it should. Then we can decide whether to re-implement the code for having V1 and V2 at the same time (i.e. as it is now), or let the pull request as it will be and you'll add V2 yourself for your private productions. Can you do so? I'll launch the tests at once.

andrzejnovak:
@perrotta Is this what you meant?

perrotta:
please test
(Thank you @andrzejnovak : then we'll decide how to put in the event, in master and in the possible backports)

andrzejnovak:
@perrotta So is this missing anything apart from me updating the models in the external data repo?

perrotta:
> @perrotta So is this missing anything apart from me updating the models in the external data repo?

Thank you @andrzejnovak 
Please go ahead with updating the models in the externals, if those updates are already available

The last check certified that the V1 version is not affected by this code: it was expected, but worh being checked.

We understand that for your purpose the best thing would be having both V1 and V2 available in the same miniAOD output. Would you be able to compare the timing and the additional memory footage for having only  V1 run, or both V1 and V2, and post the results here? If nothing is going to explode, the last commit can be reverted and this PR merged with your original scheme.

perrotta:
@andrzejnovak please let us know as soon as you have news about the update of the model in the external, and about the performance comparisons (timing and memory) between running V1 alone and V1+V2 in the miniAOD

andrzejnovak:
@perrotta Sorry, for the delay, running some last minute checks so I don't have to update the files again in a week.

andrzejnovak:
@perrotta Could you point me to what to run for the mem and timing checks? The models have been updated.

andrzejnovak:
```
TimeReport   0.000177     0.000177     0.000177  pfBoostedDoubleSVAK8TagInfos
TimeReport   0.000117     0.000117     0.000117  pfBoostedDoubleSecondaryVertexAK8BJetTags
TimeReport   0.000017     0.000017     0.000017  pfDeepDoubleXTagInfos
TimeReport   0.006630     0.006630     0.006630  pfImpactParameterAK8TagInfos
TimeReport   0.000110     0.000110     0.000110  pfInclusiveSecondaryVertexFinderAK8TagInfos
TimeReport   0.000022     0.000022     0.000022  pfMassIndependentDeepDoubleBvLJetTags
TimeReport   0.000030     0.000030     0.000030  pfMassIndependentDeepDoubleBvLV2JetTags
TimeReport   0.000022     0.000022     0.000022  pfMassIndependentDeepDoubleCvBJetTags
TimeReport   0.000044     0.000044     0.000044  pfMassIndependentDeepDoubleCvBV2JetTags
TimeReport   0.000022     0.000022     0.000022  pfMassIndependentDeepDoubleCvLJetTags
TimeReport   0.000031     0.000031     0.000031  pfMassIndependentDeepDoubleCvLV2JetTags
```

perrotta:
Thank you Andrzej, and apologize for the non-prompt answer: I will be in vacation till the end of the week, with some limited connection.

For timing and memory measurements we normally rely on what reported in the reconstruction twiki, eg.
https://twiki.cern.ch/twiki/bin/viewauth/CMS/RecoIntegration#Compare_timing
and IgProf for the memory profile:

slava77:
@cmsbuild please test

andrzejnovak:
@perrotta igprof results

Perf:
http://test-anovak.web.cern.ch/test-anovak/cgi-bin/igprof-navigator/new/igreport_perf_ticks
http://test-anovak.web.cern.ch/test-anovak/cgi-bin/igprof-navigator/old/igreport_perf_ticks

Mem:
http://test-anovak.web.cern.ch/test-anovak/cgi-bin/igprof-navigator/new/igreport_mem_live
http://test-anovak.web.cern.ch/test-anovak/cgi-bin/igprof-navigator/old/igreport_mem_live

andrzejnovak:
@slava77 @perrotta Could you take a look? I squashed everything into one commit so it's easier to backport, no changes to the actual code. Let me know if anything else is needed.

slava77:
@cmsbuild please test

andrzejnovak:
@slava77 @perrotta Is anything missing?

slava77:
> @slava77 @perrotta Is anything missing?

https://github.com/cms-sw/cmssw/pull/30016#issuecomment-651120675

andrzejnovak:
> @perrotta igprof results
> 
> Perf:
> http://test-anovak.web.cern.ch/test-anovak/cgi-bin/igprof-navigator/new/igreport_perf_ticks
> http://test-anovak.web.cern.ch/test-anovak/cgi-bin/igprof-navigator/old/igreport_perf_ticks
> 
> Mem:
> http://test-anovak.web.cern.ch/test-anovak/cgi-bin/igprof-navigator/new/igreport_mem_live
> http://test-anovak.web.cern.ch/test-anovak/cgi-bin/igprof-navigator/old/igreport_mem_live

@slava77 not this?

slava77:
> @slava77 not this?

sorry, I meant that Andrea is not available this week.

perrotta:
@andrzejnovak thank you for providing your performance results in https://github.com/cms-sw/cmssw/pull/30016#issuecomment-651309477 and https://github.com/cms-sw/cmssw/pull/30016#issuecomment-651049307. With which workflow were they obtained?

Looking at them, I see that both V1 and V2 behaves rather similarly for what timing and memory are concerned: timing seems well manageable, and also for the memory usage some 9 Mb MEM_LIVE footprint for each version seems affordable.

@slava77 please comment if you disagree; but given the measured performances I believe that both V1 and V2 could be run simultaneously, so that full studies can be performed on V2 before finally switching to it as default suggested version for analyses. This would require reverting the effect of the commit in which you removed V2 from the list of versions saved in miniAOD output. 
It must be understood that such a contemporary presence of two versions/tunes of the same algo must be intended as temporary, and as soon as all studies and scale factors have been concluded and computed only the recommended version will remin available in the release.

slava77:
related to https://github.com/cms-sw/cmssw/pull/30016#issuecomment-652991388
and https://github.com/cms-sw/cmssw/pull/30016#issuecomment-651049307 (taking just DeepDouble module names)
```
0.000017  pfDeepDoubleXTagInfos
0.000022  pfMassIndependentDeepDoubleBvLJetTags
0.000030  pfMassIndependentDeepDoubleBvLV2JetTags
0.000022  pfMassIndependentDeepDoubleCvBJetTags
0.000044  pfMassIndependentDeepDoubleCvBV2JetTags
0.000022  pfMassIndependentDeepDoubleCvLJetTags
0.000031  pfMassIndependentDeepDoubleCvLV2JetTags
```
I see in my recent (IB, not this PR) that in 136.888, using JetHT inputs, the times per event are typically around 1 ms/ev per JetTags modules, which is about a factor of 40 larger than the values above.
Even with a very fast machine possibly doing work a factor of 3-4 faster, I conclude that the timing posted above is more likely to be for events with no jets passing preselections.
Anyways, an addition of 3 ms from 3 extra modules seems acceptable, if I extrapolate from my reference.

andrzejnovak:
Hmm, fair point. The sample processed certainly had some passing jets, but it's not majority. If the times get averaged it could be biased. Though both tests were on the same sample.

@perrotta The igprof results are from running the test_deep_doublex file.

I think actually think it's fine as is - not running v2 by default in mini. The values can be calculated when producing the validation tuples/nano. I think that will make for a cleaner switch even if it wouldn't be too expensive to keep both temporarily.

slava77:
> @perrotta The igprof results are from running the test_deep_doublex file.

 'file:72164088-CB67-E811-9D0D-008CFA197AC4.root', ?
If it's a local copy of what's a row above from GluGluHToCC_M125_13TeV_powheg_pythia8,
then at least from the name it sounds like there are to first order no AK8 jets with pt>120 GeV (do I recall the cut value correctly?) in these events.

andrzejnovak:
@slava77  There are certainly some, but not many. I've rerun both on a dedicated higher pt H tagger training samples, which improves the stats a bit, though it is still far off having a passing jet in every event.

http://test-anovak.web.cern.ch/test-anovak/cgi-bin/igprof-navigator/DDX/new/igreport_perf_ticks
http://test-anovak.web.cern.ch/test-anovak/cgi-bin/igprof-navigator/DDX/old/igreport_perf_ticks

http://test-anovak.web.cern.ch/test-anovak/cgi-bin/igprof-navigator/DDX/new/igreport_mem_live
http://test-anovak.web.cern.ch/test-anovak/cgi-bin/igprof-navigator/DDX/old/igreport_mem_live

perrotta:
+1
- DeepDoubleX version 2 added as a configurable possibility in the code, but not switched on in the settings of the default workflows
- The following external must also be merged to allow running V2: cms-data/RecoBTag-Combined#31
- The timing and memory performance of V2 are comparable with the ones of current V1, therefore no issues are expected when the switch will be decided and implemented by BTV (final checks and corresponding scale factors still to be computed)
- Since the developers agree on having only by now V1 in the release (see https://github.com/cms-sw/cmssw/pull/30016#issuecomment-653064942), no changes are expected in reco outputs: jenkins tests pass and actually show no differences

silviodonato:
merge

santocch:
+1

PR #30032, opened 2020-05-28
Flow cs subtraction for heavy ions

stepobr:
Hybrid CS-PU jet reconstruction with flow correction and better background eta control. 
Used as primary algorithm in HI group, migrating to central repo now.

@mandrenguyen @bi-ran


perrotta:
please test

Issue #14728, opened 2016-06-01
mcdb and libhepml dependency


smuzaffar:
Looks like mcdb and libhepml are not actually used in cmssw. Only reference to these tools are 
https://github.com/cms-sw/cmssw/blob/CMSSW_8_1_X/GeneratorInterface/LHEInterface/BuildFile.xml
https://github.com/cms-sw/cmssw/blob/CMSSW_8_1_X/GeneratorInterface/LHEInterface/plugins/BuildFile.xml
https://github.com/cms-sw/cmssw/blob/CMSSW_8_1_X/GeneratorInterface/MCatNLOInterface/plugins/BuildFile.xml

but these can be removed without breaking the build.

smuzaffar:
assign generators
can you please check this? If not needed then we can drop these tools from cmssw externals

smuzaffar:
@thuer,@covarell,@perrozzi,@govoni any issues removing this tool for cmssw?

smuzaffar:
ping?

govoni:
I think it's ok

smuzaffar:
assign externals

smuzaffar:
https://github.com/cms-sw/cmsdist/pull/5886 and https://github.com/cms-sw/cmssw/pull/30042 remove these dependencies.

smuzaffar:
+externals

PR #30042, opened 2020-05-29
Remove mcdb and libhepml dependencies

smuzaffar:
As proposed here https://github.com/cms-sw/cmssw/issues/14728 , we think there is nothing in cmssw which is using mcdb libhepml.


smuzaffar:
please test with cms-sw/cmsdist#5886

qliphy:
+1

silviodonato:
+1

PR #30046, opened 2020-05-29
Strip clusterizer speedup: technical changes from #28304

pieterdavid:
#### PR description:

This PR contains the technical changes from https://github.com/cms-sw/cmssw/pull/28304 by @VinInn , i.e. those other than the new unpacker+clusterizer module and the use of the averaged noise per APV (rather than per strip), which need some more study and validation still (I did keep the changes to make the code that uses SiStripCluster independent of the amplitudes vector type because it makes future changes easier).

I also reorganised the `StripClusterizer` class a bit: the version currently in CMSSW caches indices of the quality, noise, and gain ranges etc. in a few vectors, and creates a helper object with all the information for one module when needed. #28304 added caching of all those per-module objects and pre-calculation of 1/gain (such that the gain calibration is a multiplication instead of a division) and the average noise per module (not included here) - I kept this, but then removed the vectors from `StripClusterizer` since they are not needed anymore, and moved all of this to the EventSetup to avoid having a copy for every stream.
This is where I have a few questions: is `SiStripClusterizerConditions` a reasonable name for that object? I kept the header in `RecoLocalTracker/SiStripClusterizer` and put the record in `RecoLocalTracker/Records` by analogy with the CPE, but I'm not sure these are the recommended places. Another option is to move the whole clusterizer tool to the EventSetup (like the CPE).

On the performance gain: in local tests in a rather unrealistic setup (only running strip local reco, so the limitation is IO and (de)compression) the full clusterizer seems a bit faster (that's also expected), but the numbers are not very reproducible (if there are instructions or tools to do proper benchmarking please let me know). Allocation and deallocation per event are a bit higher, but the difference is the same, so I suppose this is because of the `reserve` calls (the number of allocations should be smaller).

#### PR validation:

I dumped all reconstructed clusters in 5 data events (279076 in total) and found three digis that were different by 1 (probably due to numerical precision and truncation/rounding after applying the gain), charge/CM and noise squared (cuts in the clusterizer) are identical for all.

CC: @VinInn @vischia @robervalwalsh


slava77:
>  This is where I have a few questions: is `SiStripClusterizerConditions` a reasonable name for that object? I kept the header in `RecoLocalTracker/SiStripClusterizer` and put the record in `RecoLocalTracker/Records` by analogy with the CPE, but I'm not sure these are the recommended places. Another option is to move the whole clusterizer tool to the EventSetup (like the CPE).

this apparently depends only on values in Cond or Calib Formats (SiStripGain, SiStripNoises, SiStripQuality). I'm thinking that it's more practical to place the object in CalibFormats/SiStripObjects.

slava77:
@cmsbuild please test

pieterdavid:
The new commits move the object to CalibFormats/SiStripObjects, as suggested by @slava77 , and add the HLT customise function to add the ESProducer, which I forgot before (the failing relvals run without errors locally with this).

Martin-Grunewald:
please test

pieterdavid:
The remaining failures in the addOn tests https://github.com/cms-sw/cmssw/pull/30046#issuecomment-637518183 are because with the `OnLine_HLT_Fake{,1,2}` configurations no `SiStripGain` (product of two conditions from the DB) is provided (in the others it is, by `SiStripGainESProducer`). What would be the correct way to add it or work around that? Would a check `hasattr(process, "SiStripGainESProducer")` inside the customise method work, or is there a better / other way?

Martin-Grunewald:
The Fake* HLT menus do not run anything, so just skipping in that case via hasattr would indeed be fine. Please note it in the file that this is the reason for using this hasattr, so I know when this customisation gets moved inside the HLT menus themselves!

pieterdavid:
Thanks @Martin-Grunewald , not adding the ESProducer if the one it depends on isn't there fixed the addOn tests for the Fake* HLT menus indeed.

slava77:
@cmsbuild please test

pieterdavid:
Sorry for the extra iterations, I forgot to check that the previous solution was fine with the matrix tests - I changed the condition in the customisation to `if not menuType.startswith("Fake")` now, that seems fine with both the addOnTest and the failing workflows from https://github.com/cms-sw/cmssw/pull/30046#issuecomment-638510100

Martin-Grunewald:
please test

pieterdavid:
Thanks @jfernan2 , only small changes (due to rounding/truncation of the ADCs back to integers) in the clusters are expected. Some of those you linked to seem a bit large for that, I will investigate.

mmusich:
>  don't know if they are expected from the technical changes involved in this PR.
I let the experts to comment
FYI @mmusich

Thanks @jfernan2 I am actually aware of this PR and its a foreseen development in the Tk DPG. As for the changes you spotted as Pieter said it might be due to truncation of ADCs, though it seems that some digis in the clusters (even clusters?) are disappearing:

image

pieterdavid:
The disappearing clusters were because the different `SiStripQuality` label ("unbiased" instead of the default) used in the ALCARECO configs was not correctly integrated, the last two commits should fix that.

slava77:
@cmsbuild please test

jfernan2:
@pieterdavid  I still some differences and they do NOT look like due to truncation like the ones I spotted in my previous comment, I am not sure if relevants though:
https://tinyurl.com/y9soxokn

I don't find now the plot quoted by @mmusich  with these last commits

RecoLocalTracker/SubCollectionProducers/src/HITrackClusterRemover.cc
```diff
@@ -319,8 +319,7 @@ void HITrackClusterRemover::process(OmniClusterRef const &ocluster, SiStripDetId
     return;
   if (!fromTrack) {
     int clusCharge = 0;
-    for (std::vector<uint8_t>::const_iterator iAmp = cluster->amplitudes().begin(); iAmp != cluster->amplitudes().end();
-         ++iAmp) {
+    for (auto iAmp = cluster->amplitudes().begin(); iAmp != cluster->amplitudes().end(); ++iAmp) {
       clusCharge += *iAmp;
```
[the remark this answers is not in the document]
pieterdavid: Thanks, there's even `SiStripCluster::charge` which does exactly this, so I used that instead.

pieterdavid:
Thanks @jfernan2 ! The feature from the plot in https://github.com/cms-sw/cmssw/pull/30046#issuecomment-638952502 is understood and seems fixed indeed. Looking through the other differences, those in strip amplitudes or cluster charge distributions seem compatible with expectations, but the new one in PV quantities (packed vs. reco) is quite puzzling.

slava77:
@cmsbuild please test

Martin-Grunewald:
The conflict in HLTrigger/Configuration/python/customizeHLTforCMSSW.py appearing now is due to the integration of #30060 modifying the same file locations.
Please update!

jfernan2:
> Thanks @jfernan2 ! The feature from the plot in [#30046 (comment)](https://github.com/cms-sw/cmssw/pull/30046#issuecomment-638952502) is understood and seems fixed indeed. Looking through the other differences, those in strip amplitudes or cluster charge distributions seem compatible with expectations, but the new one in PV quantities (packed vs. reco) is quite puzzling.

And it seems to be there still after your last commit, sorry:
https://tinyurl.com/y99vqoys
Thanks

mmusich:
> And it seems to be there still after your last commit, sorry:

Am I wrong or it's at the sub-permil level effect? For my education, are we sure we control floating point aritmetics at that level, when performing PR tests (e.g. on different CPU architectures)?

davidlange6:
Yes (partly by not running on different CPU architectures)

On Jun 11, 2020, at 1:56 PM, Marco Musich <notifications@github.com<mailto:notifications@github.com>> wrote:



And it seems to be there still after your last commit, sorry:

Am I wrong or it's at the sub-permil level effect? For my education, are we sure we control floating point aritmetics at that level, when performing PR tests (e.g. on different CPU architectures)?

slava77:
> > Thanks @jfernan2 ! The feature from the plot in [#30046 (comment)](https://github.com/cms-sw/cmssw/pull/30046#issuecomment-638952502) is understood and seems fixed indeed. Looking through the other differences, those in strip amplitudes or cluster charge distributions seem compatible with expectations, but the new one in PV quantities (packed vs. reco) is quite puzzling.
> 
> And it seems to be there still after your last commit, sorry:
> https://tinyurl.com/y99vqoys
> Thanks

after looking a bit upstream I see changes in onTrack plots 
https://tinyurl.com/ybrp9rvu
If hits on tracks change, then track parameters can change

mmusich:
> Yes (partly by not running on different CPU architectures)

in that case we can conclude the effect is real. Given small changes (due to rounding/truncation of the ADCs back to integers) in the clusters are expected, I think we are observing small differences in the track parameters and covariance matrices that are being captured in these particular plots because the are essentially tuned to show rounding effects, while in the standard tracking DQM are not visible because of binning. @mtosi

jfernan2:
> > > Thanks @jfernan2 ! The feature from the plot in [#30046 (comment)](https://github.com/cms-sw/cmssw/pull/30046#issuecomment-638952502) is understood and seems fixed indeed. Looking through the other differences, those in strip amplitudes or cluster charge distributions seem compatible with expectations, but the new one in PV quantities (packed vs. reco) is quite puzzling.
> > 
> > 
> > And it seems to be there still after your last commit, sorry:
> > https://tinyurl.com/y99vqoys
> > Thanks
> 
> after looking a bit upstream I see changes in onTrack plots
> https://tinyurl.com/ybrp9rvu
> If hits on tracks change, then track parameters can change

Those are actually a ROOT overlay articaft in the GUI, you can see it by disabling the Overlay mode and putting plots on side: 
https://tinyurl.com/y8c37n56

jfernan2:
> > Yes (partly by not running on different CPU architectures)
> 
> in that case we can conclude the effect is real. Given small changes (due to rounding/truncation of the ADCs back to integers) in the clusters are expected, I think we are observing small differences in the track parameters and covariance matrices that are being captured in these particular plots because the are essentially tuned to show rounding effects, while in the standard tracking DQM are not visible because of binning. @mtosi

@mmusich if you are fine with the changes, I am fine too. I have just pointed them out to not let them pass unnoticed

slava77:
> Those are actually a ROOT overlay artifact in the GUI, you can see it by disabling the Overlay mode and putting plots on side:

I'm not sure what you mean,
there are bin-to-bin differences as more clearly visible in 
https://cmsweb.cern.ch/dqm/dev/plotfairy/overlay?session=KuSEWM;v=1591868355093537582;xmin=18;xmax=19;obj=archive%2F301998%2Fwf136_793_pr%2FCMSSW_11_2_X-PRcmssw_30046-6949%2FDQMIO%2FSiStrip%2FMechanicalView%2FTEC%2FPLUS%2FSummary_ClusterStoNCorrThin_OnTrack__TEC__PLUS;obj=archive%2F301998%2Fwf136_793_base%2FCMSSW_11_2_X-PRcmssw_30046-6949%2FDQMIO%2FSiStrip%2FMechanicalView%2FTEC%2FPLUS%2FSummary_ClusterStoNCorrThin_OnTrack__TEC__PLUS;reflabel=301998;norm=False;w=1115;h=762

jfernan2:
Sorry @slava77 I was refering to the apparently change in mean and rms that is shown if you overlay the plots dut to different roundings/truncations, which disappears when you show the two plots side by side.
The bin-to-bin difference you point out is indeed clearly there, and probably it is the cause of the difference in mean/rms shown by the overlay.

pieterdavid:
Thanks for the pointers, I digged a bit more into the expected numerical differences (adding back the gain range, such that I could do cluster-by-cluster comparisons of the charges after truncation, and S/N).
They show the absolute and relative difference in cluster charge between doing (`/gain` and `*weight`), and the 2D plot of S/N (new vs old) - this one is for all clusters (also off-track), and the whole detector (not just TEC+ thin sensors), but even then they don't explain the bin-by-bin differences, I will try to understand where these come from.
cluster_deltaCharge
cluster_deltaChargerel
SNmigration

pieterdavid:
Rebased to solve the merge conflicts from https://github.com/cms-sw/cmssw/pull/30046#issuecomment-642556135 , and squashed some commits to fix the invalid states for files that were added in one and moved in a subsequent commit https://github.com/cms-sw/cmssw/pull/30046#issuecomment-642443408.

At a closer look I have to correct https://github.com/cms-sw/cmssw/pull/30046#issuecomment-643131561 : the bin-by-bin difference in S/N in https://github.com/cms-sw/cmssw/pull/30046#issuecomment-642690668 is a single entry moving across a bin boundary, that's compatible with expectations.

perrotta:
@cmsbuild please test

jfernan2:
Sorry @pieterdavid  now there seems to be actual differences in tracking diffDz/Dxy for wf 10224.0  and 250202.181
https://tinyurl.com/yb2zkof6
https://tinyurl.com/y98rpjjl

DataFormats/SiStripCluster/src/SiStripCluster.cc
```diff
@@ -2,28 +2,30 @@
 #include "DataFormats/SiStripCluster/interface/SiStripCluster.h"
 
 SiStripCluster::SiStripCluster(const SiStripDigiRange& range) : firstStrip_(range.first->strip()), error_x(-99999.9) {
-  amplitudes_.reserve(range.second - range.first);
+  std::vector<uint8_t> v;
```
dpiparo, line 5: I may have misunderstood the code, but here a naive question: why not filling the *amplitudes_* vector directly without the temporary *v* vector?

VinInn: because makes no difference? (and avoid a dangerous interface...)

RecoLocalTracker/SubCollectionProducers/src/HITrackClusterRemover.cc
```diff
@@ -318,13 +318,8 @@ void HITrackClusterRemover::process(OmniClusterRef const &ocluster, SiStripDetId
...
   if (!fromTrack) {
-    int clusCharge = 0;
-    for (std::vector<uint8_t>::const_iterator iAmp = cluster->amplitudes().begin(); iAmp != cluster->amplitudes().end();
-         ++iAmp) {
-      clusCharge += *iAmp;
-    }
-    if (pblocks_[subdet - 1].cutOnStripCharge_ &&
-        (clusCharge > (pblocks_[subdet - 1].minGoodStripCharge_ * sensorThickness(detid))))
+    if (pblocks_[subdet - 0].cutOnStripCharge_ &&
+        (cluster->charge() > (pblocks_[subdet - 1].minGoodStripCharge_ * sensorThickness(detid))))
```
slava77, line 322: I'm confused why the first conditional is on ```subdet -0```, while the other is on ```subdet - 1```.

slava77: @pieterdavid 
ping
[in case the last comment got lost]

slava77:
> @pieterdavid
> ping
> [in case the last comment got lost]

that's about https://github.com/cms-sw/cmssw/pull/30046#pullrequestreview-430142970

DataFormats/SiStripCluster/src/SiStripCluster.cc
pieterdavid: I think this changes anticipates changing the underlying storage to a `std:array`, but it's not necessary indeed. Should I revert this for now? I don't know if it would be optimised out by the compiler.

RecoLocalTracker/SubCollectionProducers/src/HITrackClusterRemover.cc
pieterdavid: That was an unintentional change indeed - reverted now. Thanks for spotting it!

pieterdavid:
Sorry for the delay; https://github.com/cms-sw/cmssw/pull/30046#pullrequestreview-430142970 is fixed, but I'm still looking into the small differences in the comparisons.

DataFormats/SiStripCluster/src/SiStripCluster.cc
VinInn: this constructor is absolutely irrelevant for any production wf
https://cmssdt.cern.ch/dxr/CMSSW/search?q=%2Bcallers%3A%22SiStripCluster%3A%3ASiStripCluster%28const+SiStripCluster%3A%3ASiStripDigiRange+%26%29%22

slava77:
@cmsbuild please test

slava77:
+1

for https://github.com/cms-sw/cmssw/pull/30046	fe6b06f 
- jenkins tests pass and comparisons with the baseline show only small differences consistent with numerical changes (checked/reviewed earlier in the thread)
- using 100 events from wf 136.888, running only step3 with ```RECO:striptrackerlocalreco```, I see about 12% speedup in the full unpacking ```SiStripClusterizer/siStripClusters``` module (the timing seemed rather reproducible, within 1-2% on two reruns)

jfernan2:
> Sorry for the delay; [#30046 (review)](https://github.com/cms-sw/cmssw/pull/30046#pullrequestreview-430142970) is fixed, but I'm still looking into the small differences in the comparisons.

Sorry @pieterdavid  I understand you are still working on the DQM differences, aren't you?

pieterdavid:
> > Sorry for the delay; [#30046 (review)](https://github.com/cms-sw/cmssw/pull/30046#pullrequestreview-430142970) is fixed, but I'm still looking into the small differences in the comparisons.
> 
> Sorry @pieterdavid I understand you are still working on the DQM differences, aren't you?

In the mean time I looked through all the histograms with differences, and they all seem compatible with the expected small changes in the adcs or cluster charges (and, less frequently, in hit and track parameters).

tlampen:
+1

Martin-Grunewald:
+1

ggovi:
+1

silviodonato:
@pieterdavid  could you solve the conflict in ` HLTrigger/Configuration/python/customizeHLTforCMSSW.py `?

perrotta:
please test

pieterdavid:
I noticed that I forgot to add one line in resolving the merge conflicts... could you stop the tests and give me a few minutes to fix it?

perrotta:
please abort

slava77:
@cmsbuild please test

slava77:
+1

for https://github.com/cms-sw/cmssw/pull/30046	24e1b8c 
- rebased after the last reco signoff as of fe6b06f, the update has only context line changes in HLTrigger/Configuration/python/customizeHLTforCMSSW.py compared to the previous version

jfernan2:
+1

Martin-Grunewald:
+1

tlampen:
+1

ggovi:
+1

silviodonato:
+1

PR #30052, opened 2020-05-29
Validate GsfElectron before running MVA to avoid reading wrong DataFormat

guitargeek:
#### PR description:

You might remember that the GsfElectron DataFormat was changed in the beginning of last year: https://github.com/cms-sw/cmssw/pull/26485. One good thing about the new format was that all ID variables are stored in it and the MVA producer got adapted accordingly.

This broke the possibility to run the MVA ID directly on the old format with the new CMSSW code. One variable (conversion rejection probability) will just assume a default value, and the MVA gets wrong input variables but will not complain. It is something that I noticed back then but didn't say anything, because EGM PostRecoTools convert the old format into the new one if you follow the instructions from the EGM Twiki. 

However, not all production workflows convert the GsffElectron format where needed. Workflows that produce miniAOD from pre-106X AOD get it wrong. Two of these workflows are also in the bot-tests: `136.8311` and `136.7611` (see first tests below). There is also some pedagogic EGammma code like my [ElectronMVATutorial](https://github.com/guitargeek/ElectronMVATutorial) (presented on the last Physics Object School and still getting many views and clones) which doesn't do any DataFormat conversion. Maybe there is also other code floating around that doesn't do it. After all, samples in the old DataFormat are still recommended for analysis until the ultra-legacy samples are out, If I get that correctly.

I suggest to add a simple check which makes sure that the affected variables in the GsfElectron are not their default  value, so it's sure that they are filled and we are dealing with the new format. If this not the case, an error will be thrown with a reference to NanoAOD as an example where the conversion is implemented correctly.

The workflows that don't do the data format updating before computing the MVA on the electrons from older AOD files are also fixed.

[1] https://github.com/cms-sw/cmssw/pull/26485/files#diff-857e86e09a96cefe6edac6eb206e1d22R611

#### PR validation:

CMSSW compiles, matrix tests pass.

No backport intended (so far).

### DIFF
--- a/RecoEgamma/EgammaTools/interface/MVAValueMapProducer.h
+++ b/RecoEgamma/EgammaTools/interface/MVAValueMapProducer.h
@@ -14,6 +14,7 @@
 #include "DataFormats/Common/interface/Handle.h"
 #include "DataFormats/Common/interface/ValueMap.h"
 #include "FWCore/Framework/interface/Event.h"
+#include "RecoEgamma/EgammaTools/interface/validateEgammaCandidate.h"
 
 #include <cmath>
 #include <memory>
@@ -121,6 +122,7 @@ void MVAValueMapProducer<ParticleType>::produce(edm::StreamID,
 
     // Loop over particles
     for (auto const& cand : src->ptrs()) {
+      egammaTools::validateEgammaCandidate(*cand);
       int cat = -1;  // Passed by reference to the mvaValue function to store the category
       const float response = mvaEstimators_[iEstimator]->mvaValue(cand.get(), auxVariables, cat);
       mvaRawValues.push_back(response);                             // The MVA score
--- a/RecoEgamma/EgammaTools/interface/MVAValueMapProducer.h
+++ b/RecoEgamma/EgammaTools/interface/MVAValueMapProducer.h
@@ -15,6 +15,7 @@
 #include "DataFormats/Common/interface/ValueMap.h"
 #include "FWCore/Framework/interface/Event.h"
 #include "RecoEgamma/EgammaTools/interface/validateEgammaCandidate.h"
+#include "FWCore/Framework/interface/ConsumesCollector.h"
 
 #include <cmath>
 #include <memory>
@@ -31,8 +32,8 @@ public:
 private:
   void produce(edm::StreamID, edm::Event&, const edm::EventSetup&) const override;
 
-  // for AOD and MiniAOD case
-  const edm::EDGetTokenT<edm::View<ParticleType>> src_;
+  const edm::EDGetTokenT<edm::View<ParticleType>> srcToken_;
+  const edm::EDGetTokenT<edm::View<ParticleType>> keysForValueMapsToken_;
 
   // MVA estimators
   const std::vector<std::unique_ptr<AnyMVAEstimatorRun2Base>> mvaEstimators_;
@@ -88,11 +89,21 @@ namespace {
 
     return names;
   }
+
+  template <class ParticleType>
+  auto getKeysForValueMapsToken(edm::InputTag const& keysForValueMapsTag, edm::ConsumesCollector&& cc) {
+    const bool tagGiven = !keysForValueMapsTag.label().empty();
+    return tagGiven ? cc.consumes<edm::View<ParticleType>>(keysForValueMapsTag)
+                    : edm::EDGetTokenT<edm::View<ParticleType>>{};
+  }
+
 }  // namespace
 
 template <class ParticleType>
 MVAValueMapProducer<ParticleType>::MVAValueMapProducer(const edm::ParameterSet& iConfig)
-    : src_(consumes<edm::View<ParticleType>>(iConfig.getParameter<edm::InputTag>("src"))),
+    : srcToken_(consumes<edm::View<ParticleType>>(iConfig.getParameter<edm::InputTag>("src"))),
+      keysForValueMapsToken_(getKeysForValueMapsToken<ParticleType>(
+          iConfig.getParameter<edm::InputTag>("keysForValueMaps"), consumesCollector())),
       mvaEstimators_(getMVAEstimators(iConfig.getParameterSetVector("mvaConfigurations"))),
       mvaValueMapNames_(getValueMapNames(iConfig.getParameterSetVector("mvaConfigurations"), "Values")),
       mvaRawValueMapNames_(getValueMapNames(iConfig.getParameterSetVector("mvaConfigurations"), "RawValues")),
@@ -112,7 +123,9 @@ void MVAValueMapProducer<ParticleType>::produce(edm::StreamID,
                                                 const edm::EventSetup& iSetup) const {
   std::vector<float> auxVariables = variableHelper_.getAuxVariables(iEvent);
 
-  auto src = iEvent.getHandle(src_);
+  auto srcHandle = iEvent.getHandle(srcToken_);
+  auto keysForValueMapsHandle =
+      keysForValueMapsToken_.isUninitialized() ? srcHandle : iEvent.getHandle(keysForValueMapsToken_);
 
   // Loop over MVA estimators
   for (unsigned iEstimator = 0; iEstimator < mvaEstimators_.size(); iEstimator++) {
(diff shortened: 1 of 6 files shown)


slava77:
@cmsbuild please test

RecoEgamma/EgammaTools/interface/MVAValueMapProducer.h
```diff
@@ -121,6 +122,7 @@ void MVAValueMapProducer<ParticleType>::produce(edm::StreamID,
 
     // Loop over particles
     for (auto const& cand : src->ptrs()) {
+      egammaTools::validateEgammaCandidate(*cand);
```
perrotta, line 125: Does it have to be run in every event, and for every candidate?

perrotta:
In any case, before the tests can finish and the review can proceed a fix for the throwing cases must be implemented.

guitargeek:
Hi @perrotta, I had some time this weekend to understand what's going on. What I said about these modifiers before was in the end not very relevant in the end. The modifiers which update the data formats are used in the `slimmedElectrons` module, but our problem starts earlier: the MVA is computed on the `reducedGedGsfElectrons` before the slimming.

Since the object updators are written for `pat::Electrons`, I had to adapt them a bit and added a new task to apply them to the `gedGsfElectrons`. This is documented a bit more in detail in the [corresponding `_cff.py` file](https://github.com/cms-sw/cmssw/pull/30052/files#diff-042764519e976e632d6ad0de7a0023bcR2).

The `MVAValueMapProducer` was changed such that it can key the value maps to a different collection than it's `src` collection. In this case, the MVA will be computed for the updated `gedGsfElectronsFrom80XTo106XTask` collection but still keyed to the `gedGsfElectrons`. In this way, I don't need to change anything in the configuration of other producers or other value map producers.

Sorry for the added complexity of the configuration. I hope you understand that it's not good for me if the electron MVA gets garbage in some cases. If this might cause problems eventually, I'm probably the one who will be asked to deal with it :)

All **tests should succeed now**. Maybe we can discuss after the bot tests how this "validation" should best look like? I know doing it for each electron is a waste, but maybe there are also better ways than checking the values explicitly? Is it better to use the provenance?

perrotta:
@cmsbuild please test

guitargeek:
I missed the equivalent problem in 2017... Now I manually checked that also the 2017 workflow succeed.

perrotta:
@cmsbuild please test

guitargeek:
Changes are expected, as we discovered a bug in this PR and fixed it. They also go in the right direction of increased MVA scores: the wrong conversion probabilities of infinity (before this PR) caused smaller MVA values, because conversions are a right for background (electron fakes from photons).

guitargeek:
By the way, should we do something about this bug in general? I'm a bit worried because it might very likely exist also in CMSSW_10_6_X which is used for a lot of things. If "UL re-miniAOD" means that the miniAOD is reproduced from pre-10_6_X AOD we might have a problem here?

perrotta:
- Is there any possibility that it does not get validated in one cand/event rather that in another one? If there is not such a possibility, shouldn't then we better validate it once per job, or at most per thread? Still, having to run the validation on every single candidate looks like an overkill to me
- It is hard to appreciated the effect of this "bug fix" from the comparisons of the miniAOD outputs: could you quickly summarize the extent of it, possibly by attaching a few slides able to point out the size of differences in physics outputs?
- UL re-miniAOD will be based on the AOD already produced with 10_6, not on older ones

guitargeek:
### Quantifying the effect of un-filled conversion probability on electron MVA

I investigated the differences with a fraction of the 2017 DY+Jets dataset produced in 94X [1], checking tens of thousands of reconstructed electrons. Signal are (as in the MVA training) electrons matched to true prompt electrons, and background are non-prompt matches and unmatched reco electrons. Electrons from tau decays are excluded.

**Efficiencies at all Fall17 V2 working points for pT > 10 GeV:**

<table border="0" class="dataframe">  <thead>    <tr style="text-align: right;">      <th></th>      <th>sig [%]</th>      <th>sig (bugged) [%]</th>      <th>bkg [%]</th>      <th>bkg (bugged) [%]</th>    </tr>  </thead>  <tbody>    <tr>      <th>Fall17noIsoV2wp80</th>      <td>49.07</td>      <td>2.07</td>      <td>2.28</td>      <td>0.02</td>    </tr>    <tr>      <th>Fall17noIsoV2wpLoose</th>      <td>78.86</td>      <td>19.64</td>      <td>5.76</td>      <td>0.73</td>    </tr>    <tr>      <th>Fall17noIsoV2wp90</th>      <td>73.04</td>      <td>13.90</td>      <td>5.62</td>      <td>0.54</td>    </tr>    <tr>      <th>Fall17isoV2wpHZZ</th>      <td>77.74</td>      <td>24.13</td>      <td>4.19</td>      <td>0.63</td>    </tr>    <tr>      <th>Fall17isoV2wp80</th>      <td>48.72</td>      <td>4.75</td>      <td>1.60</td>      <td>0.07</td>    </tr>    <tr>      <th>Fall17isoV2wpLoose</th>      <td>81.07</td>      <td>26.65</td>      <td>4.86</td>      <td>0.81</td>    </tr>    <tr>      <th>Fall17isoV2wp90</th>      <td>72.09</td>      <td>18.88</td>      <td>3.99</td>      <td>0.49</td>    </tr>  </tbody></table>

**Efficiencies at all Fall17 V2 working points for 5 < pT < 10 GeV:**

<table border="0" class="dataframe">  <thead>    <tr style="text-align: right;">      <th></th>      <th>sig [%]</th>      <th>sig (bugged) [%]</th>      <th>bkg [%]</th>      <th>bkg (bugged) [%]</th>    </tr>  </thead>  <tbody>    <tr>      <th>Fall17noIsoV2wp80</th>      <td>78.35</td>      <td>7.45</td>      <td>1.70</td>      <td>0.10</td>    </tr>    <tr>      <th>Fall17noIsoV2wpLoose</th>      <td>98.98</td>      <td>93.25</td>      <td>13.45</td>      <td>5.32</td>    </tr>    <tr>      <th>Fall17noIsoV2wp90</th>      <td>89.29</td>      <td>23.72</td>      <td>3.53</td>      <td>0.53</td>    </tr>    <tr>      <th>Fall17isoV2wpHZZ</th>      <td>96.89</td>      <td>77.09</td>      <td>4.81</td>      <td>1.51</td>    </tr>    <tr>      <th>Fall17isoV2wp80</th>      <td>78.15</td>      <td>7.83</td>      <td>1.03</td>      <td>0.05</td>    </tr>    <tr>      <th>Fall17isoV2wpLoose</th>      <td>98.96</td>      <td>93.93</td>      <td>10.18</td>      <td>4.20</td>    </tr>    <tr>      <th>Fall17isoV2wp90</th>      <td>89.18</td>      <td>23.58</td>      <td>2.20</td>      <td>0.35</td>    </tr>  </tbody></table>

**BDT output for all reconstructed electrons:**

<div align="center">
electron_mva_conversion_bug
</div>

I can confirm that an infinite conversion probability indeed decreases the score. A high conversion probability is a sign for background.

**Conclusion:**

* Effect of forgetting to update the data format has been investigated for reconstructed electrons in 2017 DY+Jets simulation
* It causes the BDT responds to shift down, strongly reducing the efficiencies at the EGM working points
* The difference is so large that if anyone affected by this problem would probably have noticed the drastic efficiency loss and reported to some EGM hypernews

[1] `/DYJetsToLL_M-50_TuneCP5_13TeV-amcatnloFXFX-pythia8/RunIIFall17MiniAODv2-PU2017_12Apr2018_new_pmx_94X_mc2017_realistic_v14_ext1-v1/MINIAODSIM`

### On the overkill of validating all electrons

> Is there any possibility that it does not get validated in one cand/event rather that in another one?

The result of the validation for one electron can be generalized to all others, so just checking one single electron is fine.

The `MVAValueMapProducer<>` is a global producer. Would you be fine with adding a `CMS_THREAD_SAFE mutable std::atomic<bool>` as a class member that is flagged when the validation has been done once so it's not done again?

I would prefer to not have some more complicated optimization, because the MVA code should not get more complicated as it already is with all this VID stuff.

### UL re-miniAOD

Thanks, it's very reassuring to  hear from you that this bug will have no consequences on upcoming campaigns.

perrotta:
Thank you @guitargeek for the comprehansive explanation of the comparisons.
Your proposal for not having to check every single electron in every single event looks reasonable to me, and I don't have anything better in mind by now: please proceed as such.

perrotta:
please test

RecoEgamma/EgammaTools/interface/MVAValueMapProducer.h
```diff
@@ -111,7 +127,15 @@ void MVAValueMapProducer<ParticleType>::produce(edm::StreamID,
...
   std::vector<float> auxVariables = variableHelper_.getAuxVariables(iEvent);
 
-  auto src = iEvent.getHandle(src_);
+  auto srcHandle = iEvent.getHandle(srcToken_);
+  auto keysForValueMapsHandle =
+      keysForValueMapsToken_.isUninitialized() ? srcHandle : iEvent.getHandle(keysForValueMapsToken_);
+
+  // check if nothing is wrong with the data format of the candidates
+  if (!validated_ && !srcHandle->empty()) {
+    egammaTools::validateEgammaCandidate((*srcHandle)[0]);
```
perrotta, line 136: Is it possible that the first Canditate to get validated is not a reco::GsfElectron? If so, the validation wouldn't be applied to the correct collection.

guitargeek: The `ParticleType` template argument is either specialized to [`reco::Photon`](https://github.com/cms-sw/cmssw/blob/b3864aef5f76bde11db583070526ac87607aea33/RecoEgamma/PhotonIdentification/plugins/plugins.cc#L8) or [`reco::GsfElectron`](https://github.com/cms-sw/cmssw/blob/5c3f56ee0db7642e67bdab5530e0781a1fbcec80/RecoEgamma/ElectronIdentification/plugins/plugins.cc#L34), not something higher up in the inheritance hierarchy like `reco::Candidate`.

So if any of the candidates passed to the producer is a `reco::GsfElectron`, the first must be one too. If they are photons, the validation function will do nothing thanks to [this check](https://github.com/cms-sw/cmssw/pull/30052/files#diff-68efcb949db29068bd93c757901dceb3R12). I hope I  understood your concerns correctly!

perrotta: My concern is that if the first candidate is a Photon, then the validation function will do nothing but the ```validated_``` flag is set to true anyhow
Perhaps, ```validateEgammaCandidate(...)``` should return a bool saying whether he has actually validated a reco::GsfElectron

guitargeek: I thought this is the meaningful behaviour: if it's a photon, the validation is trivial so the validated_ flag can be set to true. What you are saying is basically for `validateEgammaCandidate` to return whether the validation was non-trivial, and then the producer doesn't set the `validated_` flag if it wasn't. I have to admit that I don't see the advantage of that, but maybe I'm missing something?

perrotta: Ok, thank you Jonas. Indeed I misunderstood the behaviour at first (too much time passed since I inspected this code...).

The validation will run for every instantiation of the MVAValueMapProducer separately, any of those instantiations will get validated independently, and the source for those instantiation will remain the same, so no electron and photons having to get validated in the same module. Therefore, it is correct that the validated_ flag is switched to true at the first check, which will be on the only ParticleType to be inspected in that module.

perrotta:
+1
- The validation of GSF electrons allows throwing in case MVA input variables are not converted from the old GsfElectron DataFormat, as they would give meaningless outputs in that case
- Adaptations (i.e. a fix) for the standard miniAOD workflows which do not  yet apply such format change are provided
  - this implies modifications in the MVA outputs of two workflows that run in the automated tests, 136.8311 and 136.7611: their outputs show now the fixed MVA outputs
- Jenkins tests pass and show no other differences besides those showing up for the electron IDs in the fixed workflows

guitargeek:
Hi @santocch, do you have any objections? The file ` PhysicsTools/PatAlgos/python/slimming/miniAOD_tools.py` gets modified quite frequently, so I'd be happy if this PR is not kept open for too long.

santocch:
+1

silviodonato:
+1

PR #30057, opened 2020-05-30
Flow modulated CS subtraction for heavy ions

stepobr:
Flow-modulated constituent subtraction for heavy ions. An option to the existing CSJetProducer. Takes into account azimuthal anisotropy of the underlying event.

@mandrenguyen @bi-ran


slava77:
@cmsbuild please test

RecoHI/HiJetAlgos/plugins/HiPFCandCleaner.h
```diff
@@ -0,0 +1,22 @@
+#include "DataFormats/ParticleFlowCandidate/interface/PFCandidateFwd.h"
+#include "FWCore/Framework/interface/EDProducer.h"
+#include "FWCore/Utilities/interface/EDGetToken.h"
+
+class HiPFCandCleaner : public edm::EDProducer {
```
slava77, line 5: this should be changed to a thread-aware producer
https://twiki.cern.ch/twiki/bin/view/CMSPublic/FWMultithreadedFrameworkModuleTypes
It looks like this can be edm::global

RecoHI/HiJetAlgos/plugins/HiPuRhoProducer.h
```diff
@@ -0,0 +1,125 @@
...
+#include "fastjet/PseudoJet.hh"
+
+#include <map>
+#include <vector>
+
+//
+// class declaration
+//
+
+class HiPuRhoProducer : public edm::EDProducer {
```
slava77, line 24: make this thread-aware
https://twiki.cern.ch/twiki/bin/view/CMSPublic/FWMultithreadedFrameworkModuleTypes
at a glance, edm::stream should be the easiest

RecoHI/HiJetAlgos/plugins/HiFJRhoFlowModulationProducer.h
```diff
@@ -0,0 +1,34 @@
+#ifndef RecoHI_HiJetAlgos_HiFJRhoFlowModulationProducer_h
+#define RecoHI_HiJetAlgos_HiFJRhoFlowModulationProducer_h
```
slava77, line 2: here and in other cases of .h files declaring EDProducers, please merge the .h file content into .cc; there is no need for a header file

slava77 (commented):
I have somewhat cursory comments to start with.
more to come later, hopefully tomorrow.

slava77:
@cmsbuild please test

RecoHI/HiJetAlgos/plugins/HiPuRhoProducer.cc
```diff
@@ -0,0 +1,645 @@
...
+      continue;
+
+    for (auto const& im : allgeomid_) {
+      double dr = reco::deltaR(geo_->getPosition((DetId)im), pseudojetTMP);
+      std::vector<std::pair<int, int>>::const_iterator exclude =
+          std::find(excludedTowers.begin(), excludedTowers.end(), std::pair<int, int>(im.ieta(), im.iphi()));
+      if (dr < radiusPU_ && exclude == excludedTowers.end() &&
+          (geomtowers_[im.ieta()] - ntowersWithJets_[im.ieta()]) > minimumTowersFraction_ * (geomtowers_[im.ieta()])) {
+        ntowersWithJets_[im.ieta()]++;
+        excludedTowers.emplace_back(std::pair<int, int>(im.ieta(), im.iphi()));
```
slava77, line 506:
```suggestion
        excludedTowers.emplace_back(im.ieta(), im.iphi());
```

RecoHI/HiJetAlgos/plugins/HiPuRhoProducer.cc
```diff
@@ -0,0 +1,645 @@
...
+      }
+    }
+
+    for (auto const& input : fjInputs_) {
+      int index = input.user_index();
+      const reco::CandidatePtr& originalTower = inputs_[index];
+
+      int ie = ieta(originalTower);
+      int ip = iphi(originalTower);
+      auto exclude = std::find(excludedTowers.begin(), excludedTowers.end(), std::pair<int, int>(ie, ip));
```
slava77, line 519:
```suggestion
      auto exclude = std::find(excludedTowers.begin(), excludedTowers.end(), std::pair(ie, ip));
```

RecoHI/HiJetAlgos/plugins/HiPuRhoProducer.cc
```diff
@@ -0,0 +1,645 @@
...
+      if (exclude != excludedTowers.end()) {
+        jettowers.push_back(index);
+      }
+
+      double dr = reco::deltaR(input, pseudojetTMP);
```
slava77, line 524: deltaR2 and use ```EtaPhi``` for pseudojetTMP as suggested above

RecoHI/HiJetAlgos/plugins/HiPuRhoProducer.cc
```diff
@@ -0,0 +1,645 @@
...
+      if (dr < radiusPU_ && nref < nMaxJets_) {
+        jtexntow[nref]++;
+        jtexpt[nref] += originalTower->pt();
+      }
+    }
+
+    if (nref < nMaxJets_)
+      nref++;
+  }
```
slava77, line 533:
```suggestion
  } // for (auto const& pseudojetTMP : fjJets_)
```
for readability

RecoHI/HiJetAlgos/plugins/HiPuRhoProducer.cc
```diff
@@ -0,0 +1,645 @@
...
+
+  postOrphan_ = true;
+}
+
+void HiPuRhoProducer::putRho(edm::Event& iEvent, const edm::EventSetup& iSetup) {
+  std::size_t size = etaEdgeLow_.size();
+
+  std::vector<std::pair<std::size_t, double>> order;
+  for (std::size_t i = 0; i < size; ++i) {
+    order.emplace_back(std::make_pair(i, etaEdgeLow_[i]));
```
slava77, line 558:
```suggestion
    order.emplace_back(i, etaEdgeLow_[i]);
```
no need for details with emplace

RecoHI/HiJetAlgos/plugins/HiPuRhoProducer.cc
```diff
@@ -0,0 +1,645 @@
...
+  iEvent.put(std::move(mapToTowExcludeEtaOut), "mapToTowExcludeEta");
+}
+
+// ------------ method fills 'descriptions' with the allowed parameters for the module  ------------
+void HiPuRhoProducer::fillDescriptions(edm::ConfigurationDescriptions& descriptions) {
+  //The following says we do not know what parameters are allowed so do no validation
+  // Please change this to state exactly what you do use, even if it is no parameters
+  edm::ParameterSetDescription desc;
+  desc.setUnknown();
+  descriptions.addDefault(desc);
```
slava77, line 619: this should implement all parameters properly, after that the explicit RecoHI/HiJetAlgos/python/hiPuRhoProducer_cfi.py file can be removed from the source code

RecoHI/HiJetAlgos/plugins/HiPuRhoProducer.cc
```diff
@@ -0,0 +1,645 @@
...
+int HiPuRhoProducer::iphi(const reco::CandidatePtr& in) const {
+  const CaloTower* ctc = dynamic_cast<const CaloTower*>(in.get());
+  if (ctc) {
+    return ctc->id().iphi();
+  }
+
+  throw cms::Exception("Invalid Constituent") << "CaloJet constituent is not of CaloTower type";
+
+  return 0;
+}
```
slava77, line 642: not needed if CaloTower is used directly (it seems like there is no other alternative)

RecoHI/HiJetAlgos/python/HiRecoPFJets_cff.py
```diff
@@ -45,6 +45,17 @@
...
+ak4PFJetsForFlow = akPu5PFJets.clone(
+    Ghost_EtaMax = cms.double(5.0),
+    Rho_EtaMax = cms.double(4.4),
+    doRhoFastjet = cms.bool(False),
+    jetPtMin = cms.double(15.0),
+    nSigmaPU = cms.double(1.0),
+    rParam = cms.double(0.4),
+    radiusPU = cms.double(0.5),
+    src = cms.InputTag("pfcandCleaner", "particleFlowCleaned"),
+)
```
slava77, line 57: please drop type specifications for all parameters which already exist.
```suggestion
ak4PFJetsForFlow = akPu5PFJets.clone(
    Ghost_EtaMax = 5.0,
    Rho_EtaMax = 4.4,
    doRhoFastjet = False,
    jetPtMin = 15.0,
    nSigmaPU = 1.0,
    rParam = 0.4,
    radiusPU = 0.5,
    src = ("pfcandCleaner", "particleFlowCleaned"),
)
```

RecoHI/HiJetAlgos/python/hiFJRhoFlowModulationProducer_cfi.py
```diff
@@ -0,0 +1,13 @@
+import FWCore.ParameterSet.Config as cms
+
+hiFJRhoFlowModulationProducer = cms.EDProducer(
```
slava77, line 3: this file is to be removed after ```HiFJRhoFlowModulationProducer::fillDescriptions``` is properly implemented

RecoHI/HiJetAlgos/python/hiPuRhoProducer_cfi.py
```diff
@@ -0,0 +1,14 @@
+import FWCore.ParameterSet.Config as cms
+
+hiPuRhoProducer = cms.EDProducer(
```
slava77, line 3: this file is to be removed after ```HiPuRhoProducer::fillDescriptions``` is properly implemented

RecoJets/JetProducers/plugins/CSJetProducer.cc
```diff
@@ -6,20 +6,28 @@
...
+    : VirtualJetProducer(conf), csRParam_(-1.0), csAlpha_(0.), useModulatedRho_(false) {
   //get eta range, rho and rhom map
   etaToken_ = consumes<std::vector<double>>(conf.getParameter<edm::InputTag>("etaMap"));
   rhoToken_ = consumes<std::vector<double>>(conf.getParameter<edm::InputTag>("rho"));
   rhomToken_ = consumes<std::vector<double>>(conf.getParameter<edm::InputTag>("rhom"));
   csRParam_ = conf.getParameter<double>("csRParam");
   csAlpha_ = conf.getParameter<double>("csAlpha");
+  useModulatedRho_ = conf.getParameter<bool>("useModulatedRho");
+  minFlowChi2Prob_ = conf.getParameter<double>("minFlowChi2Prob");
+  maxFlowChi2Prob_ = conf.getParameter<double>("maxFlowChi2Prob");
```
slava77, line 28: move to the initializer list (before ```{``` of the constructor)

RecoJets/JetProducers/plugins/CSJetProducer.cc
```diff
@@ -78,30 +89,44 @@ void CSJetProducer::runAlgorithm(edm::Event& iEvent, edm::EventSetup const& iSet
...
-    double rho = 1e-6;
-    double rhom = 1e-6;
+    //assign rho and rhom to ghosts according to local eta-dependent map + modulation as function of phi
     for (fastjet::PseudoJet& ighost : ghosts) {
+      double rhoModulationFactor = 1.;
+      double ghostPhi = ighost.phi_std();
+
+      if (useModulatedRho_) {
+        if (!rhoFlowFitParams->empty()) {
+          double val = ROOT::Math::chisquared_cdf_c(rhoFlowFitParams->at(5), rhoFlowFitParams->at(6));
```
slava77, line 99: compute this outside of the loop over jets/ghosts

RecoJets/JetProducers/plugins/CSJetProducer.cc
```diff
@@ -143,10 +168,20 @@ void CSJetProducer::fillDescriptions(edm::ConfigurationDescriptions& description
...
   desc.add<double>("csRParam", -1.);
   desc.add<double>("csAlpha", 2.);
-
+  desc.add<bool>("useModulatedRho", false);
+  desc.add<double>("minFlowChi2Prob", false);
+  desc.add<double>("maxFlowChi2Prob", false);
   desc.add<edm::InputTag>("etaMap", edm::InputTag("hiFJRhoProducer", "mapEtaEdges"));
   desc.add<edm::InputTag>("rho", edm::InputTag("hiFJRhoProducer", "mapToRho"));
   desc.add<edm::InputTag>("rhom", edm::InputTag("hiFJRhoProducer", "mapToRhoM"));
+  desc.add<edm::InputTag>("rhoFlowFitParams", edm::InputTag("hiFJRhoFlowModulationProducer", "rhoFlowFitParams"));
```
slava77, line 177:
```suggestion
  desc.add<edm::InputTag>("rhoFlowFitParams", {"hiFJRhoFlowModulationProducer", "rhoFlowFitParams"});
```
a bit shorter; but update the above as well, if apply this one

RecoJets/JetProducers/plugins/CSJetProducer.h
```diff
@@ -32,13 +34,21 @@ namespace cms {
   protected:
     void runAlgorithm(edm::Event& iEvent, const edm::EventSetup& iSetup) override;
 
+    double getModulatedRhoFactor(
+        const double phi, const double eventPlane2, const double eventPlane3, const double par1, const double par2);
```
slava77, line 38:
```suggestion
    double getModulatedRhoFactor(
        double phi, double eventPlane2, double eventPlane3, double par1, double par2) const;
```
the method itself is simple enough that const modifiers to the arguments are not adding clarity

slava77 (commented):
Please provide timing performance on some central events

RecoHI/HiJetAlgos/plugins/HiPuRhoProducer.cc
```diff
@@ -0,0 +1,645 @@
...
+  inputs_.clear();
+  fjInputs_.clear();
+  fjJets_.clear();
+
+  // Get the particletowers
+  edm::Handle<reco::CandidateView> inputsHandle;
+  iEvent.getByToken(input_candidateview_token_, inputsHandle);
+
+  for (size_t i = 0; i < inputsHandle->size(); ++i)
+    inputs_.push_back(inputsHandle->ptrAt(i));
```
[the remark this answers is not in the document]
slava77: apparently this should be ```auto const&``` and  then ```(&input)```

RecoHI/HiJetAlgos/plugins/HiFJRhoFlowModulationProducer.cc
```diff
@@ -0,0 +1,251 @@
...
+  if (doJettyExclusion_)
+    iEvent.getByToken(jetTag_, jets);
+
+  int nFill = 0;
+
+  double eventPlane2 = -100;
+  double eventPlane3 = -100;
+
+  constexpr int nParamVals = 9;
+  auto rhoFlowFitParamsOut = std::make_unique<std::vector<double>>(nParamVals, 1e-6);
```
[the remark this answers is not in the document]
stepobr: Sorry can't figure that out, `auto rhoFlowFitParamsOut = std::make_unique<std::array<double, nParamVals>>` not seem to work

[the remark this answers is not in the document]
slava77: what was the error?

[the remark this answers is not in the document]
stepobr: Probably just doing something wrong
` error: pointer to a function used in arithmetic [-Werror=pointer-arith]
   rhoFlowFitParamsOut[0] = 0;
   error: assignment of read-only location '**rhoFlowFitParamsOut'
   rhoFlowFitParamsOut[0] = 0;
`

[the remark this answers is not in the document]
slava77: but this looks more like about ```(*rhoFlowFitParamsOut)[0]``` vs ```rhoFlowFitParamsOut[0]```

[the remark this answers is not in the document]
stepobr: Same error if I change to `(*rhoFlowFitParamsOut)[0]`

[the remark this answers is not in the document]
slava77: I may be missing what's being tried on your side.
The following compiles for me:
```c++
  auto arrUP = std::make_unique<std::array<double, 9>>();
  (*arrUP)[2] = 3;
  std::cout<<(*arrUP)[2]<<std::endl;
```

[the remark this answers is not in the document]
stepobr: I was missing the `()` at the end when the array is defined, thank you

RecoHI/HiJetAlgos/plugins/HiFJRhoFlowModulationProducer.cc
```diff
@@ -0,0 +1,251 @@
...
+        if (!isGood)
+          continue;
+      }
+
+      phi_h->Fill(pfCandidate.phi());
+    }
+
+    std::string flowFitForm = "[0]*(1.+2.*([1]*TMath::Cos(2.*(x-[2]))+[3]*TMath::Cos(3.*(x-[4]))))";
+
+    TF1* flowFit_p = new TF1("flowFit", flowFitForm.c_str(), -TMath::Pi(), TMath::Pi());
```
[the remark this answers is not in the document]
stepobr: I've managed to redo the formulas as user functions and move the fits to data members, but I can't make them `auto`:
`non-static data member declared with placeholder 'auto'`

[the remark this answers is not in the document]
slava77: right, a type detail needed for the case of a data member is needed.

RecoHI/HiJetAlgos/plugins/HiFJRhoFlowModulationProducer.cc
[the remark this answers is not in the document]
stepobr: Also I was wondering what is the right way to `iEvent.put` this array if it expects a vector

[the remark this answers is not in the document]
slava77: I may have not considered the fact that this is to be saved in the event.
@makortel @Dr15Jones do we have support for std::arrays saved directly in edm::Event (even if transient)?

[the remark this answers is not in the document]
Dr15Jones: `ticl::Trackster` contains `std::array` s and it has a ROOT dictionary.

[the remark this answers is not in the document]
slava77: indeed, array as a part of a class works (also used in PFCandidate).
However, there are no StdDictionaries .
@stepobr please update the produces call to write an array.
You may need to add ```std::array<double,9>``` (or what's the size) in  DataFormats/StdDictionaries/src/classes_def_others.xml
and also add an include for ```<array>```  in DataFormats/StdDictionaries/src/classes_others.h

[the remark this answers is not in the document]
Dr15Jones: I think ROOT has built in direct support for std::array which is why we do not have any explicit dictionary definitions for them even though they are used as member data.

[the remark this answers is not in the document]
slava77: even better, no need to fiddle with DataFormats/StdDictionaries/src/

RecoHI/HiJetAlgos/plugins/HiPuRhoProducer.cc
```diff
@@ -0,0 +1,645 @@
...
+  int medianWindowWidth_;
+  double minimumTowersFraction_;
+  double nSigmaPU_;  // number of sigma for pileup
+  double puPtMin_;
+  double radiusPU_;    // pileup radius
+  double rParam_;      // the R parameter to use
+  edm::InputTag src_;  // input constituent source
+  double towSigmaCut_;
+
+  std::vector<edm::Ptr<reco::Candidate>> inputs_;  // input candidates
```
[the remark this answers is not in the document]
stepobr: If I change that, I get problems later:
`
error: no matching function for call to 'std::vector<const reco::Candidate*>::push_back(edm::Ptr<reco::Candidate>)'
     inputs_.push_back(inputsHandle->ptrAt(i));
`
So `inputs_.push_back(inputsHandle->ptrAt(i));` should be modified somehow

RecoHI/HiJetAlgos/plugins/HiFJRhoFlowModulationProducer.cc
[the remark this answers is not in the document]
stepobr: So should I directly `iEvent.put(rhoFlowFitParamsOut);` ?

RecoHI/HiJetAlgos/plugins/HiPuRhoProducer.cc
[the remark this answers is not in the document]
slava77: right, there is another comment (below) https://github.com/cms-sw/cmssw/pull/30057#discussion_r441607647 which suggests to change to regular pointer values

RecoHI/HiJetAlgos/plugins/HiFJRhoFlowModulationProducer.cc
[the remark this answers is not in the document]
slava77: move is expected to be better; IIUC, the issue was that the matching ```produces``` call should have a std::array instead of a vector

RecoHI/HiJetAlgos/plugins/HiPuRhoProducer.cc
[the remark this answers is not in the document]
stepobr: Ok, almost there. Together with the comment below, I guess I need to change every `reco::CandidatePtr&` with `reco::Candidate*`? If not, I get these kind of errors:
`
error: invalid initialization of reference of type 'const CandidatePtr&' {aka 'const edm::Ptr<reco::Candidate>&'} from expression of type '__gnu_cxx::__alloc_traits<std::allocator<const reco::Candidate*>, const reco::Candidate*>::value_type' {aka 'const reco::Candidate*'}
     const reco::CandidatePtr& originalTower = inputs_[input_object.user_index()];

`

RecoHI/HiJetAlgos/plugins/HiFJRhoFlowModulationProducer.cc
```diff
@@ -0,0 +1,251 @@
...
+
+  float* hiEvtPlane;
+};
+
+HiFJRhoFlowModulationProducer::HiFJRhoFlowModulationProducer(const edm::ParameterSet& iConfig)
+    : doEvtPlane_(iConfig.getParameter<bool>("doEvtPlane")),
+      doFreePlaneFit_(iConfig.getParameter<bool>("doFreePlaneFit")),
+      doJettyExclusion_(iConfig.getParameter<bool>("doJettyExclusion")),
+      evtPlaneLevel_(iConfig.getParameter<int>("evtPlaneLevel")),
+      jetTag_(consumes<reco::JetView>(iConfig.getParameter<edm::InputTag>("jetTag"))),
```
makortel, line 64: If the event product pointed by `getTag_` is read only if `doJettyExclusion_ == true`, please consume it also only if `doJettyExclusion_ == true`. Otherwise a dependence is declared unnecessarily when `doJettyExclusion_ == false`.

RecoHI/HiJetAlgos/plugins/HiFJRhoFlowModulationProducer.cc
```diff
@@ -0,0 +1,251 @@
...
+    rhoFlowFitParamsOut->at(7) = lineFit_p->GetChisquare();
+    rhoFlowFitParamsOut->at(8) = lineFit_p->GetNDF();
+
+    delete phi_h;
+
+    delete flowFit_p;
+    delete lineFit_p;
+  }
+
+  iEvent.put(std::move(rhoFlowFitParamsOut), "rhoFlowFitParams");
```
makortel, line 225: The instance name needs to be removed also here
```suggestion
  iEvent.put(std::move(rhoFlowFitParamsOut));
```

RecoHI/HiJetAlgos/plugins/HiPuRhoProducer.cc
```diff
@@ -0,0 +1,645 @@
...
+    fjInputs_.back().set_user_index(i - inBegin);
+  }
+}
+
+void HiPuRhoProducer::setupGeometryMap(edm::Event& iEvent, const edm::EventSetup& iSetup) {
+  LogDebug("PileUpSubtractor") << "The subtractor setting up geometry...\n";
+
+  if (geo_ == nullptr) {
+    edm::ESHandle<CaloGeometry> pG;
+    iSetup.get<CaloGeometryRecord>().get(pG);
```
makortel, line 252: Please use `esConsumes()` and `ESGetToken` to access EventSetup products https://twiki.cern.ch/twiki/bin/view/CMSPublic/SWGuideHowToGetDataFromES#In_ED_module

RecoHI/HiJetAlgos/plugins/HiPuRhoProducer.cc
```diff
@@ -0,0 +1,645 @@
...
+  produces<std::vector<double>>("mapToTowExcludeEta");
+}
+
+// ------------ method called to produce the data  ------------
+void HiPuRhoProducer::produce(edm::Event& iEvent, const edm::EventSetup& iSetup) {
+  setupGeometryMap(iEvent, iSetup);
+
+  inputs_.clear();
+  fjInputs_.clear();
+  fjJets_.clear();
```
makortel, line 192: How bit can these containers become? In general we should not keep unnecessary stuff in memory from one event to the next.

RecoHI/HiJetAlgos/plugins/HiPuRhoProducer.cc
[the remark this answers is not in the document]
slava77: this was the original comment

> there is no need to convert to Ptr, use simple pointers

So, it did mean to suggest to switch to using simple pointers, as written originally.

I didn't see any case where the Ptr part was used, it was always dereferenced to get the underlying object.

RecoHI/HiJetAlgos/plugins/HiPuRhoProducer.cc
```diff
@@ -0,0 +1,645 @@
...
+  }
+
+  postOrphan_ = false;
+}
+
+void HiPuRhoProducer::subtractPedestal(std::vector<fastjet::PseudoJet>& coll) {
+  LogDebug("PileUpSubtractor") << "The subtractor subtracting pedestals...\n";
+
+  std::vector<fastjet::PseudoJet> newcoll;
+  for (auto& input_object : coll) {
```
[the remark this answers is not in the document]
stepobr: `auto const&` doesn't work with `input_object.reset_momentum(towP4.px(), towP4.py(), towP4.pz(), towP4.energy())`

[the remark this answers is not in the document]
slava77: ah, right, the inputs are indeed modified.
OK.

RecoHI/HiJetAlgos/plugins/HiPuRhoProducer.cc
```diff
@@ -0,0 +1,645 @@
...
+
+    int ietaold = -10000;
+    ietamax_ = -10000;
+    ietamin_ = 10000;
+    for (auto const& did : alldid) {
+      if (did.det() == DetId::Hcal) {
+        HcalDetId hid = HcalDetId(did);
+        allgeomid_.push_back(did);
+
+        if (hid.ieta() != ietaold) {
```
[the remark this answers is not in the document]
stepobr: It works as expected. DetIds are ordered like subdet->depth->eta,phi. So same ieta indeed appears again when the next depth comes in. So in order not to double-count, the geomtowers are reset to 1 for every new depth.

RecoHI/HiJetAlgos/plugins/HiPFCandCleaner.cc
```diff
@@ -0,0 +1,63 @@
...
+  // ----------member data ---------------------------
+  edm::EDGetTokenT<reco::PFCandidateCollection> candidatesToken_;
+
+  double ptMin_;
+  double absEtaMax_;
+};
+//
+// constructors and destructor
+//
+HiPFCandCleaner::HiPFCandCleaner(const edm::ParameterSet& iConfig)
```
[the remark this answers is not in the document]
stepobr: hiPFCandCleaner has to be used by `ak4PFJetsForFlow`. They are used only if `doJettyExclusion_=True`. Currently it's not used but we are planning to do that in the future. But it's good to know that such filter exists, I'll drop that from the PR.

RecoHI/HiJetAlgos/plugins/HiFJRhoFlowModulationProducer.cc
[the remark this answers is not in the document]
stepobr: @slava77 I'd like to get back to this. So when I change the matching `produces` to `std::array', as well as the `consumes` in `CSJetProducer` I get this:
`
An exception of category 'DictionaryNotFound' occurred while
   [1] Constructing module: class=HiFJRhoFlowModulationProducer label='hiFJRhoFlowModulationProducer'
   [2] Calling ProductRegistryHelper::addToRegistry, checking dictionaries for produced types
Exception Message:

  edm::Wrapper<std::array<double,9> >
  std::array<double,9>

`

I tried to add the array in `StdDictionaries` but that didn't help.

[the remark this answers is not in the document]
slava77: > I tried to add the array in `StdDictionaries` but that didn't help.

this seems to disagree with 
https://github.com/cms-sw/cmssw/pull/30057#discussion_r445126814
@Dr15Jones it looks like the support to stream array types directly is not that simple (if even present)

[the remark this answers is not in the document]
stepobr: Adding the array to `StdDictionaries` gives an error during compilation:
`Error: It is not necessary to explicitly select class array<double,9>. I/O is supported for it transparently.`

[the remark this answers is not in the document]
makortel: The symptoms (ROOT saying it has a dictionary, framework that it has not) smell similar to https://github.com/cms-sw/cmssw/pull/27861. I'll take a look.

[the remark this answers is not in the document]
makortel: The cause is different from #27861, and may take some time to resolve. I opened an issue for it https://github.com/cms-sw/cmssw/issues/30493.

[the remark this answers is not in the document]
slava77: In case the resolution will take longer than ~today, I'm willing to accept the current vector solution for this product.
With an assumption/note that it can/should be updated once the framework has support for array.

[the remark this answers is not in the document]
makortel: Almost certainly will take longer than today.

RecoHI/HiJetAlgos/plugins/HiFJRhoFlowModulationProducer.cc
```diff
@@ -0,0 +1,251 @@
...
+    eventPlane3 = hiEvtPlane[15];
+  }
+
+  if (nFill >= 100 && eventPlane2 > -99) {
+    const int nPhiBins = std::max(10, nFill / 30);
+
+    std::string name = "phiTestIEta4_" + std::to_string(iEvent.id().event()) + "_h";
+    std::string nameFlat = "phiTestIEta4_Flat_" + std::to_string(iEvent.id().event()) + "_h";
+
+    TH1F* phi_h = new TH1F(name.data(), "", nPhiBins, -TMath::Pi(), TMath::Pi());
```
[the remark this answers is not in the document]
stepobr: > variants with 3 to 10 bins
it's not like that actually, it's `std::max(10, nFill / 30);` so not sure that we can avoid histogram re-creation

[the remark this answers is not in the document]
slava77: Let's try to be more specific/practical then.
what's the current cost per event?
please get a timing measurement per event (TimeReport is good enough) and a profiler report for CPU https://twiki.cern.ch/twiki/bin/viewauth/CMS/RecoIntegration#Run_profiler_igprof 
this should help to understand if something other than TH1 should be used or if TH1 is acceptable.
Thank you.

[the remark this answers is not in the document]
stepobr: That's time in seconds averaged over 10 events 

`hiPuRhoProducer 0.3736367
hiFJRhoFlowModulationProducer 0.001974513
CSJetProducer 0.630398`

For profiler report, `hiFJRhoFlowModulationProducer` is mentioned a couple of times like:
`
0.0  .........       0.00 / 0.00         HiFJRhoFlowModulationProducer::produce(edm::Event&, edm::EventSetup const&) [11411]
`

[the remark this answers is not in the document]
slava77: thank you.
please share the full report.
Also, please check/clarify what input data is used; it is important to have most events to pass ```   if (nFill >= 100 && eventPlane2 > -99) {``` for this check to make sense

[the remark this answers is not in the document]
stepobr: profiler report [link](https://cernbox.cern.ch/index.php/s/ol4vChdgUKTzzgL)
time report [link](https://cernbox.cern.ch/index.php/s/pEPHZf0jpkUPyUD)

10 out of 10 events have passed this check, but I'll take a look at a larger sample

RecoHI/HiJetAlgos/plugins/HiPuRhoProducer.cc
stepobr: For some reason quite a few jet producers clears this kind of containers at the beginning of an event. I don't see why they can't be cleared at the end instead.

stepobr:
I've pushed all the changes based on the review, please let me know if something isn't right or just badly implemented, hopefully I haven't missed anything.

slava77:
@cmsbuild please test

RecoHI/HiJetAlgos/plugins/HiPuRhoProducer.cc
makortel: Thanks. Maybe this hints for an opportunity for a later cleanup in all affected jet producers (@slava77 @perrotta @jpata). In principle there should be no reason to store such data in member variables in the first place (especially given that `.clear()` does not free the memory).

mandrenguyen:
Just a kindly ping.  I'm waiting on this PR to update the jet sequences in HI miniAOD.   Is there something still blocking  this one?

slava77:
> Just a kindly ping. I'm waiting on this PR to update the jet sequences in HI miniAOD. Is there something still blocking this one?

I expect to get to this in a day or two, perhaps sooner.

RecoHI/HiJetAlgos/plugins/HiPuRhoProducer.cc
```diff
@@ -0,0 +1,615 @@
...
+  CaloGeometry const* geo_ = nullptr;  // geometry
+  std::vector<HcalDetId> allgeomid_;   // all det ids in the geometry
+
+  int ietamax_;                               // maximum eta in geometry
+  int ietamin_;                               // minimum eta in geometry
+  std::map<int, int> ntowersWithJets_;        // number of towers with jets
+  std::map<int, int> geomtowers_;             // map of geometry towers to det id
+  std::map<int, double> esigma_;              // energy sigma
+  std::map<int, double> emean_;               // energy mean
+  std::map<int, std::vector<double>> eTop4_;  // energy mean
```
slava77, line 153:
```suggestion
  std::map<int, std::array<double, 4>> eTop4_;  // energy mean
```

RecoHI/HiJetAlgos/plugins/HiPuRhoProducer.cc
```diff
@@ -0,0 +1,615 @@
...
+
+      emean_[it] = e1 / (double)nt;
+      double eee = e2 / nt - e1 * e1 / (nt * nt);
+      if (eee < 0.)
+        eee = 0.;
+      esigma_[it] = nSigmaPU_ * sqrt(eee);
+
+      double etaWidth = etaedge[abs(it)] - etaedge[abs(it) - 1];
+      if (etaWidth < 0)
+        etaWidth *= -1.;
```
slava77, line 396: isn't this the same as 
```suggestion
      double etaWidth = std::abs(etaedge[abs(it)] - etaedge[abs(it) - 1]);
```
?

RecoHI/HiJetAlgos/plugins/HiPuRhoProducer.cc
```diff
@@ -0,0 +1,615 @@
...
+
+      int sign = (it < 0) ? -1 : 1;
+
+      if (sign * etaedge[abs(it)] < sign * etaedge[abs(it) - 1]) {
+        etaEdgeLow_.push_back(sign * etaedge[abs(it)]);
+        etaEdgeHi_.push_back(sign * etaedge[abs(it) - 1]);
+      } else {
+        etaEdgeHi_.push_back(sign * etaedge[abs(it)]);
+        etaEdgeLow_.push_back(sign * etaedge[abs(it) - 1]);
+      }
```
slava77, line 406:
```suggestion
      auto absIt = std::abs(it);
      if (it < 0) {
        etaEdgeLow_.push_back(sign * etaedge[absIt]);
        etaEdgeHi_.push_back(sign * etaedge[absIt - 1]);
      } else {
        etaEdgeHi_.push_back(sign * etaedge[absIt]);
        etaEdgeLow_.push_back(sign * etaedge[absIt - 1]);
      }
```
- temporary instead of function call for abs(it)
- std::abs
- etaedge is ordered by construction; so it's known that ```etaedge[i] > etaedge[i-1]``` for positive i, so the condition reduces to it < 0 (or sign < 0)

RecoHI/HiJetAlgos/plugins/HiPuRhoProducer.cc
```diff
@@ -0,0 +1,615 @@
...
+
+    if (it < 0)
+      vi = nEtaTow_ + it;
+
+    double e1 = emean_[it];
+    double e2 = emean2[it];
+    int nt = gt.second - ntowersWithJets_[it];
+
+    if (vi < nEtaTow_) {
+      vngeom[vi] = gt.second;
```
slava77, line 357: it looks like vngeom depends only on geomtowers_.
move this computation to ```setupGeometryMap```

RecoHI/HiJetAlgos/plugins/HiPuRhoProducer.cc
```diff
@@ -0,0 +1,615 @@
...
+
+void HiPuRhoProducer::subtractPedestal(std::vector<fastjet::PseudoJet>& coll) {
+  LogDebug("PileUpSubtractor") << "The subtractor subtracting pedestals...\n";
+
+  std::vector<fastjet::PseudoJet> newcoll;
+  for (auto& input_object : coll) {
+    int index = input_object.user_index();
+    reco::Candidate const* itow = inputs_[index];
+
+    const CaloTower* ctc = dynamic_cast<const CaloTower*>(itow);
```
slava77, line 440: another place that leads to an unnecessary cast

RecoHI/HiJetAlgos/plugins/HiPuRhoProducer.cc
```diff
@@ -0,0 +1,615 @@
...
+    int it = ctc->id().ieta();
+
+    double original_Et = itow->et();
+    double etnew = original_Et - (emean_.find(it))->second - (esigma_.find(it))->second;
```
slava77, line 444:
```suggestion
    double etnew = original_Et - emean_.at(it).second - esigma_.at(it).second;
```

RecoHI/HiJetAlgos/python/HiRecoPFJets_cff.py
```diff
@@ -45,6 +45,33 @@
...
+# ak4PFJetsForFlow = akPu5PFJets.clone(
+#     Ghost_EtaMax = cms.double(5.0),
+#     Rho_EtaMax = cms.double(4.4),
+#     doRhoFastjet = cms.bool(False),
+#     jetPtMin = cms.double(15.0),
+#     nSigmaPU = cms.double(1.0),
+#     rParam = cms.double(0.4),
+#     radiusPU = cms.double(0.5),
+#     src = cms.InputTag("hiPFCandCleanerforJets"),
+# )
```
slava77, line 73: is this commented out code needed?

RecoHI/HiJetAlgos/python/HiRecoPFJets_cff.py
```diff
@@ -59,20 +86,52 @@
 kt4PFJetsForRho.GhostArea     = cms.double(0.005)
 
 from RecoHI.HiJetAlgos.hiFJRhoProducer import hiFJRhoProducer
+#from RecoHI.HiJetAlgos.hiPuRhoProducer_cfi import hiPuRhoProducer
+#from RecoHI.HiJetAlgos.hiFJRhoFlowModulationProducer_cfi import hiFJRhoFlowModulationProducer
+#from RecoHI.HiJetAlgos.hiPuRhoProducer import hiPuRhoProducer
+#from RecoHI.HiJetAlgos.hiFJRhoFlowModulationProducer import hiFJRhoFlowModulationProducer
+#from RecoHI.HiJetAlgos.hiPFCandCleaner_cfi import hiPFCandCleaner
```
slava77, line 93: is this commented out code needed?

RecoHI/HiJetAlgos/python/HiRecoPFJets_cff.py
```diff
@@ -59,20 +86,52 @@
...
+    'HiFJRhoFlowModulationProducer',
+    pfCandSource = cms.InputTag('particleFlow'),
+    doJettyExclusion = cms.bool(False),
+    doFreePlaneFit = cms.bool(False),
+    doEvtPlane = cms.bool(False),
+    doFlatTest = cms.bool(False),
+    jetTag = cms.InputTag("ak4PFJets"),
+    EvtPlane = cms.InputTag("hiEvtPlane"),
+    evtPlaneLevel = cms.int32(0)
+    )
```
slava77, line 105: replace with what's generated by the fillDescriptions.
actually, I'd suggest to drop the "Producer" from the product name, which is a more common for persisted data

```suggestion
import RecoHI.HiJetAlgos.hiFJRhoFlowModulationProducer_cfi as _mod
hiFJRhoFlowModulation = _mod.hiFJRhoFlowModulationProducer.clone()
```

RecoHI/HiJetAlgos/python/HiRecoPFJets_cff.py
```diff
@@ -59,20 +86,52 @@
...
+    dropZeroTowers = cms.bool(True),
+    medianWindowWidth = cms.int32(2),
+    minimumTowersFraction = cms.double(0.7),
+    nSigmaPU = cms.double(1.0),
+    puPtMin = cms.double(15.0),
+    rParam = cms.double(.3),
+    radiusPU = cms.double(.5),
+    src = cms.InputTag('PFTowers'),
+    towSigmaCut = cms.double(5.), 
+    )
```
slava77, line 118: replace with what's generated by the fillDescriptions
 actually, I'd suggest to drop the "Producer" from the product name, which is a more common for persisted data; this would mean also a few edits in the InputTags
```suggestion
import RecoHI.HiJetAlgos.hiPuRhoProducer_cfi as _mod
hiPuRho = _mod.hiPuRhoProducer.clone()
```

RecoHI/HiJetAlgos/plugins/HiPuRhoProducer.cc
```diff
@@ -0,0 +1,615 @@
...
+      auto exclude = std::find(excludedTowers.begin(), excludedTowers.end(), std::pair(im.ieta(), im.iphi()));
+      if (dr2 < radiusPU_ * radiusPU_ && exclude == excludedTowers.end() &&
+          (geomtowers_[im.ieta()] - ntowersWithJets_[im.ieta()]) > minimumTowersFraction_ * (geomtowers_[im.ieta()])) {
+        ntowersWithJets_[im.ieta()]++;
+        excludedTowers.emplace_back(im.ieta(), im.iphi());
+
+        if (nref < nMaxJets_)
+          jtexngeom[nref]++;
+      }
+    }
```
slava77, line 498: looking at igprof results (sent earlier), it looks like this loop is rather heavy.
It looks more practical/efficient to :
- loop over ```towermap``` elements instead of ```allgeomid_```, as only ieta and iphi are used
- compute exclude only after ```dr2 <``` check is done
- use an std::map or std::unordered_map to speedup lookup inside the excludedTowers from linear (vector)

RecoHI/HiJetAlgos/plugins/HiPuRhoProducer.cc
stepobr: Not sure how to fix `error: request for member 'second' in '((HiPuRhoProducer*)this)->HiPuRhoProducer::emean_.std::map<int, double>::at(it)', which is of non-class type 'std::map<int, double>::mapped_type' {aka 'double'}`

slava77: ah, sorry, I forgot to drop .second

RecoHI/HiJetAlgos/plugins/HiPuRhoProducer.cc
```diff
@@ -0,0 +1,615 @@
...
+  const double puPtMin_;
+  const double radiusPU_;  // pileup radius
+  const double rParam_;    // the R parameter to use
+  const double towSigmaCut_;
+  const edm::EDGetTokenT<reco::CandidateView> input_candidateview_token_;
+  edm::ESGetToken<CaloGeometry, CaloGeometryRecord> caloToken_;
+
+  constexpr static int nMaxJets_ = 200;
+
+  float jteta[nMaxJets_];
```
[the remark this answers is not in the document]
stepobr: Sure. Also, looks like all the jet arrays `jtpt, jteta, jtphi, jtpu, jtexpt, jtexngeom, jtexntow` are not needed. Probably they were used for testing.

RecoHI/HiJetAlgos/plugins/HiPuRhoProducer.cc
stepobr: But `towermap` gives eta and phi coordinates, not ieta and iphi

slava77: ah, right; looking more carefully, I see that the towermap (a std::map) is used only in conjunction with looping over allgeomid_ (a vector). So, the map functionality of it is not really needed.
For this case it's more appropriate to make it a vector, perhaps even a vector of {eta, phi, ieta, iphi} and use it here instead of allgeomid_.

stepobr: So `towermap` should be `std::vector<std::array<double,4>>`?

slava77: I meant a struct, like ```struct EtaPhiTower{ int ieta, iphi; float eta, phi;}```
so that access to the values is readable (much better than e.g.  ```tower[i][0]```)

slava77:
@cmsbuild please test

mandrenguyen:
@slava77 Are there still outstanding items to be addressed on this PR?

slava77 (commented):
- https://github.com/cms-sw/cmssw/pull/30057#discussion_r451865078 implied that the newly added constant etaedge array is also used in RecoHI/HiJetAlgos/plugins/ParticleTowerProducer.cc; the current variant still represents a duplicate

slava77:
@cmsbuild please test

RecoHI/HiJetAlgos/plugins/HiPuRhoProducer.cc
```diff
@@ -0,0 +1,584 @@
...
+  std::map<int, double> emean2;
+  std::map<int, int> ntowers;
+
+  int ietaold = -10000;
+  // Initial values for emean_, emean2, esigma_, ntowers
+
+  for (int vi = 0; vi < nEtaTow_; ++vi) {
+    int it = vi + 1;
+    if (it > nEtaTow_ / 2)
+      it = vi - nEtaTow_;
```
slava77, line 288:
```suggestion
```
to resolve the issue reported in https://cmssdt.cern.ch/SDT/jenkins-artifacts/pull-request-integration/PR-561667/8025/llvm-analysis/report-a044d2.html#EndPath

that's unless the change made in https://github.com/cms-sw/cmssw/pull/30057/files/a0e15386a5d2675bc8eeba5558d4bc905cfd65d0..e8657815f4596a3349308006bf8981f9d004ca6b#diff-2093a69ad726653693a30f6c4415e430L308 (click "Load diff" for RecoHI/HiJetAlgos/plugins/HiPuRhoProducer.cc to get to the change), which had dependence on ```it``` was made incorrectly.

This appears to be the last comment for this PR. Sorry for missing it in the last round(s).

slava77:
@cmsbuild please test

slava77:
Looking at comparisons, e.g. in 140.56 workflow, all changes are in akCs[34]PFJets collections

it looks like the most significant change is at high eta
all_OldVSNew_RunHI2018wf140p56c_recoPFCandidates_akCs4PFJets_pfParticlesCs_reRECO_obj_eta
wf140 56_ak4CsPF_candDr

wf140 56_ak4CsPF_eta

please clarify if these kind of changes are expected and if so, please add a slink to some summary of changes in the physics outputs (from an HI group meeting) in the PR description to improve self-documentation

stepobr:
> please clarify if these kind of changes are expected and if so, please add a slink to some summary of changes in the physics outputs (from an HI group meeting) in the PR description to improve self-documentation

Most of the changes are outside of the usual range for jets in heavy ions. We are usually limited to |eta|<3.0 due to many reasons. Unfortunately I'm not aware of any comparison in the |eta|>3.0 region. Maybe @mandrenguyen can add something.

mandrenguyen:
Not much to add wrt @stepobr  We are usually limited to |eta|<2, so even more restricted than |eta|<3.  Even in the unlikely case we mange to do something with very forward jets, we can always recluster them from the PF candidates.

slava77:
> > please clarify if these kind of changes are expected and if so, please add a slink to some summary of changes in the physics outputs (from an HI group meeting) in the PR description to improve self-documentation
> 
> Most of the changes are outside of the usual range for jets in heavy ions. We are usually limited to |eta|<3.0 due to many reasons. Unfortunately I'm not aware of any comparison in the |eta|>3.0 region. Maybe @mandrenguyen can add something.

this is a puzzling response (including a clarification in https://github.com/cms-sw/cmssw/pull/30057#issuecomment-660286277; it somewhat sounds like "the origins are unclear, as we don't care about this region, and even if there is a bug, we can rerun this code", but perhaps I misunderstood). 
Is it clear which part of the code is responsible for such an increase in the number of passing objects?

If all these jets are not needed beyond eta 3, why not just place a cut at around 3.5 and save a few kCHF of tape/disk cost?

slava77:
in 158.0 the changes in akCs4PFJets from this PR lead to 2.3% increase in the AOD size
all_OldVSNew_HydjetQB12ppRECOin2018wf158p0c_recoPFCandidates_akCs4PFJets_pfParticlesCs_RECO_obj_eta

stepobr:
Sorry, we'd better figure this out properly. My guess is the following:
Reference (old style) CsJets uses `kt4PFJetsForRho` to later calculate rho inside `hiFJRhoProducer`. The new CsJets calculates rho inside `HiPuRhoProducer` using `PFTowers` as input. The difference is that for `PFTowers` we set `useHF = False`, while for `kt4PFJetsForRho` there are no such limitations. Rho is the pt density that is subtracted from PFCandidates to mitigate pileup before clustering jets. If the resulting PFCandidate is zero it is deleted from the collection. So what we get is that we basically don't subtract anything for HF.

mandrenguyen:
@stepobr  yes, that sounds right.  Was also looking from my side.  Can we try to set useHF to true and see if it removes the HF excess?
@slava77 you make a good point, i had reversed the colors, and thought we had no jets in the HF with the new code.

slava77:
@cmsbuild please test

stepobr:
@slava77 I struggle to get the same type of plots as the ones you provided. I'm familiar with DQM GUI but it doesn't help. Or may be it's done in some other way... Are there any instructions on that by any chance? To be more precise, for the latest tests I can't find in the DQM the datasets with `30057-8096` in it.

slava77:
> @slava77 I struggle to get the same type of plots as the ones you provided. I'm familiar with DQM GUI but it doesn't help. Or may be it's done in some other way... Are there any instructions on that by any chance? To be more precise, for the latest tests I can't find in the DQM the datasets with `30057-8096` in it.

I tried to follow the DQM bin-to-bin and the overlay variant showed up empty for e.g. 140.56
@jfernan2  please check
Screen Shot 2020-07-18 at 8 58 09 PM

The RelMon in "Comparison failed" https://cmssdt.cern.ch/SDT/jenkins-artifacts/baseLineComparisons/CMSSW_11_2_X_2020-07-17-2300+561667/37828/validateJR.html
 DQM  part has some relevant plots
https://cmssdt.cern.ch/SDT/jenkins-artifacts/baseLineComparisons/CMSSW_11_2_X_2020-07-17-2300+561667/37828/140.56_RunHI2018+RunHI2018+RECOHID18+HARVESTDHI18/JetMET_HIJetValidation_akCs4PFJets.html

reco comparisons , linked from the same "Comparisons failed" page https://cmssdt.cern.ch/SDT/jenkins-artifacts/baseLineComparisons/CMSSW_11_2_X_2020-07-17-2300+561667/37828/validateJR/all_OldVSNew_RunHI2018wf140p56


Plots similar to https://github.com/cms-sw/cmssw/pull/30057#issuecomment-660223881 are below (red is with this PR compared to the baseline in CMSSW_11_2_X_2020-07-17-2300 in black)
all_OldVSNew_RunHI2018wf140p56c_recoPFCandidates_akCs4PFJets_pfParticlesCs_reRECO_obj_eta
wf140 56_ak4CsPF_candDr
wf140 56_ak4CsPF_eta

It looks like the last update went in the right direction.
Please check/confirm.

stepobr:
@slava77 thanks. Great, that is what we expected!

RecoJets/JetProducers/plugins/CSJetProducer.cc
```diff
@@ -78,30 +99,40 @@ void CSJetProducer::runAlgorithm(edm::Event& iEvent, edm::EventSetup const& iSet
...
+                                                        rhoFlowFitParams->at(2),
+                                                        rhoFlowFitParams->at(4),
+                                                        rhoFlowFitParams->at(1),
+                                                        rhoFlowFitParams->at(3));
+          }
+        }
+      }
+
+      int32_t ghostPos = -1;
       if (ighost.eta() <= etaRanges->at(0) || bkgVecSize == 1) {
```
slava77, line 120:
```suggestion
      auto const ighostEta = ighost.eta();
      if (ighostEta <= etaRanges->at(0) || bkgVecSize == 1) {
```
apparently I'm (re)discovering for myself that the compiler wouldn't necessarily optimize away repeated calls

This minor optimization of fastjet::PseudoJet::pseudorapidity() calls will save about 3% of RECO time (ignoring startup and IO cost) in wf 158.0 
- before https://slava77sk.web.cern.ch/slava77sk/reco/cgi-bin/igprof-navigator/CMSSW_11_2_X_2020-07-18-1100-test30057.158.0.step3.prod.100.pp/102
- after https://slava77sk.web.cern.ch/slava77sk/reco/cgi-bin/igprof-navigator/CMSSW_11_2_X_2020-07-18-1100-test30057.19f86335b13.158.0.step3.prod.100.pp/164

slava77 (commented):
this would really be the last one.
I couldn't pass on the cost increase triggered by this PR introducing more etaRanges in the CSJetProducer

RecoJets/JetProducers/plugins/CSJetProducer.cc
slava77: > This minor optimization

the ```s/ighost.eta()/ighostEta/```
to apply to the full scope here

slava77:
@cmsbuild please test

slava77:
+1

for https://github.com/cms-sw/cmssw/pull/30057	968daf9 
- jenkins tests pass and comparisons with the baseline show differences in HI workflows in CS jet variables
- local test with 200 events in wf 158.0 and 140.56 confirm the expected behavior with more stats
    - CPU use is somewhat under control (based on wf 140.56; with rather central events): the new modules hiPuRho (36 ms/ev; a factor of about 100 faster than the earliest version as of e865781) and hiFJRhoFlowModulation (1.3 ms/ev) contribute somewhat negligibly to the total (of about 300 s/ev ); the last update in 968daf9 fully mitigated the compute performance in akCs4PFJets (based on a test done in 158.0, though). 

plots posted in https://github.com/cms-sw/cmssw/pull/30057#issuecomment-660583427 were confirmed to be OK/expected

silviodonato:
+1

silviodonato:
merge

stepobr:
Many thanks for all the help!!!

PR #30060, opened 2020-05-30
Hcal calibration cache

igv4321:
#### PR description:

This pull request replaces PR #29852 with a similar calibration cache implementation based on an ESProducer. In addition to HBHEPhase1Reconstructor, HFPreReconstructor has been switched to this cache as well.

One additional push to follow, for the HLT customization.

#### PR validation:

The usual runTheMatrix tests were run. Some private modifications were also used to print calibration info and to make sure that it is the same as before.

### DIFF
--- a/RecoLocalCalo/HcalRecAlgos/interface/HcalPipelinePedestalAndGain.h
+++ b/RecoLocalCalo/HcalRecAlgos/interface/HcalPipelinePedestalAndGain.h
@@ -20,9 +20,7 @@ public:
         gain_(i_gain),
         gainWidth_(i_gainWidth) {}
 
-  inline float pedestal(const bool useEffectivePeds) const {
-    return useEffectivePeds ? effPedestal_ : pedestal_;
-  }
+  inline float pedestal(const bool useEffectivePeds) const { return useEffectivePeds ? effPedestal_ : pedestal_; }
 
   inline float pedestalWidth(const bool useEffectivePeds) const {
     return useEffectivePeds ? effPedestalWidth_ : pedestalWidth_;
@@ -40,4 +38,4 @@ private:
   float gainWidth_;
 };
 
-#endif // RecoLocalCalo_HcalRecAlgos_HcalPipelinePedestalAndGain_h_
+#endif  // RecoLocalCalo_HcalRecAlgos_HcalPipelinePedestalAndGain_h_
--- a/RecoLocalCalo/HcalRecAlgos/interface/HcalPipelinePedestalAndGain.h
+++ b/RecoLocalCalo/HcalRecAlgos/interface/HcalPipelinePedestalAndGain.h
@@ -1,0 +1,43 @@
+#ifndef RecoLocalCalo_HcalRecAlgos_HcalPipelinePedestalAndGain_h_
+#define RecoLocalCalo_HcalRecAlgos_HcalPipelinePedestalAndGain_h_
+
+// HB/HE channel information stored for each pipeline "capacitor id"
+class HcalPipelinePedestalAndGain {
+public:
+  inline HcalPipelinePedestalAndGain()
+      : pedestal_(0.f), pedestalWidth_(0.f), effPedestal_(0.f), effPedestalWidth_(0.f), gain_(0.f), gainWidth_(0.f) {}
+
+  inline HcalPipelinePedestalAndGain(const float i_pedestal,
+                                     const float i_pedestalWidth,
+                                     const float i_effPedestal,
+                                     const float i_effPedestalWidth,
+                                     const float i_gain,
+                                     const float i_gainWidth)
+      : pedestal_(i_pedestal),
+        pedestalWidth_(i_pedestalWidth),
+        effPedestal_(i_effPedestal),
+        effPedestalWidth_(i_effPedestalWidth),
+        gain_(i_gain),
+        gainWidth_(i_gainWidth) {}
+
+  inline float pedestal(const bool useEffectivePeds) const {
+    return useEffectivePeds ? effPedestal_ : pedestal_;
+  }
+
+  inline float pedestalWidth(const bool useEffectivePeds) const {
+    return useEffectivePeds ? effPedestalWidth_ : pedestalWidth_;
+  }
+
+  inline float gain() const { return gain_; }
+  inline float gainWidth() const { return gainWidth_; }
+
+private:
+  float pedestal_;
+  float pedestalWidth_;
+  float effPedestal_;
+  float effPedestalWidth_;
+  float gain_;
+  float gainWidth_;
+};
+
+#endif // RecoLocalCalo_HcalRecAlgos_HcalPipelinePedestalAndGain_h_
--- a/RecoLocalCalo/HcalRecAlgos/plugins/HcalChannelPropertiesEP.cc
+++ b/RecoLocalCalo/HcalRecAlgos/plugins/HcalChannelPropertiesEP.cc
@@ -45,7 +45,7 @@ public:
     const HcalTopology& htopo = rcd.getRecord<HcalRecNumberingRecord>().get(topoToken_);
     const HcalRecoParams& params = rcd.getRecord<HcalRecoParamsRcd>().get(paramsToken_);
 
-    ReturnType1 prod(new HcalRecoParams(params));
+    ReturnType1 prod = std::make_unique<HcalRecoParams>(params);
     prod->setTopo(&htopo);
     return prod;
   }
@@ -71,7 +71,7 @@ public:
     const HcalTopology& htopo(*params.topo());
 
     // Build the product
-    ReturnType2 prod(new HcalChannelPropertiesVec(htopo.ncells()));
+    ReturnType2 prod = std::make_unique<HcalChannelPropertiesVec>(htopo.ncells());
     std::array<HcalPipelinePedestalAndGain, 4> pedsAndGains;
     const HcalSubdetector subdetectors[3] = {HcalBarrel, HcalEndcap, HcalForward};
 
--- a/RecoLocalCalo/HcalRecAlgos/plugins/HcalChannelPropertiesEP.cc
+++ b/RecoLocalCalo/HcalRecAlgos/plugins/HcalChannelPropertiesEP.cc
@@ -1,0 +1,119 @@
+#include "FWCore/Framework/interface/ESProducer.h"
+#include "FWCore/Framework/interface/ESHandle.h"
+#include "FWCore/ParameterSet/interface/ParameterSet.h"
+
+#include "CalibFormats/HcalObjects/interface/HcalDbService.h"
+#include "CalibFormats/HcalObjects/interface/HcalDbRecord.h"
+
+#include "DataFormats/HcalDetId/interface/HcalGenericDetId.h"
+#include "DataFormats/HcalDetId/interface/HcalSubdetector.h"
+
+#include "CondFormats/HcalObjects/interface/HcalRecoParams.h"
+#include "CondFormats/DataRecord/interface/HcalRecoParamsRcd.h"
+
+#include "Geometry/CaloGeometry/interface/CaloGeometry.h"
+#include "Geometry/Records/interface/CaloGeometryRecord.h"
+#include "Geometry/CaloTopology/interface/HcalTopology.h"
+#include "Geometry/HcalTowerAlgo/interface/HcalGeometry.h"
+
+#include "RecoLocalCalo/HcalRecAlgos/interface/HcalSeverityLevelComputer.h"
+#include "RecoLocalCalo/HcalRecAlgos/interface/HcalSeverityLevelComputerRcd.h"
+#include "RecoLocalCalo/HcalRecAlgos/interface/HcalChannelProperties.h"
+#include "RecoLocalCalo/HcalRecAlgos/interface/HcalChannelPropertiesRecord.h"
+
+class HcalChannelPropertiesEP : public edm::ESProducer {
+public:
+    typedef std::unique_ptr<HcalChannelPropertiesVec> ReturnType;
+
+    inline HcalChannelPropertiesEP(const edm::ParameterSet&) 
+    {
+        auto cc = setWhatProduced(this);
+        edm::ESInputTag qTag("", "withTopo");
+        cc.setConsumes(condToken_).setConsumes(topoToken_).setConsumes(paramsToken_);
+        cc.setConsumes(sevToken_).setConsumes(qualToken_, qTag).setConsumes(geomToken_);
+    }
+
+    inline virtual ~HcalChannelPropertiesEP() override {}
+
+    ReturnType produce(const HcalChannelPropertiesRecord& rcd)
+    {
+        // There appears to be no easy way to trace the internal
+        // dependencies of HcalDbService. So, rebuild the product
+        // every time anything changes in the parent records.
+        // This means that we are sometimes going to rebuild the
+        // whole table on the lumi block boundaries instead of
+        // just updating the list of bad channels.
+        using namespace edm;
+
+        // Retrieve various event setup records and data products
+        const HcalDbRecord& dbRecord = rcd.getRecord<HcalDbRecord>();
+        const HcalDbService& cond = dbRecord.get(condToken_);
+        const HcalTopology& htopo = dbRecord.getRecord<HcalRecNumberingRecord>().get(topoToken_);
+        const HcalRecoParams& params = dbRecord.getRecord<HcalRecoParamsRcd>().get(paramsToken_);
+        const HcalSeverityLevelComputer& severity = rcd.getRecord<HcalSeverityLevelComputerRcd>().get(sevToken_);
+        const HcalChannelQuality& qual = dbRecord.getRecord<HcalChannelQualityRcd>().get(qualToken_);
+        const CaloGeometry& geom = rcd.getRecord<CaloGeometryRecord>().get(geomToken_);
+
+        paramTS_ = std::make_unique<HcalRecoParams>(params);
+        paramTS_->setTopo(&htopo);
+
+        // Build the product
+        ReturnType prod(new HcalChannelPropertiesVec(htopo.ncells()));
+        std::array<HcalPipelinePedestalAndGain, 4> pedsAndGains;
+        const HcalSubdetector subdetectors[3] = {HcalBarrel, HcalEndcap, HcalForward};
+
+        for (HcalSubdetector subd : subdetectors) {
+            const HcalGeometry* hcalGeom = static_cast<const HcalGeometry*>(
+                geom.getSubdetectorGeometry(DetId::Hcal, subd));
+            const std::vector<DetId>& ids = hcalGeom->getValidDetIds(DetId::Hcal, subd);
+
+            for (const auto cell : ids) {
+                const auto rawId = cell.rawId();
+
+                // ADC decoding tools, etc
+                const HcalRecoParam* param_ts = paramTS_->getValues(rawId);
+                const HcalQIECoder* channelCoder = cond.getHcalCoder(cell);
+                const HcalQIEShape* shape = cond.getHcalShape(channelCoder);
+                const HcalSiPMParameter* siPMParameter = cond.getHcalSiPMParameter(cell);
+
+                // Pedestals and gains
+                const HcalCalibrations& calib = cond.getHcalCalibrations(cell);
+                const HcalCalibrationWidths& calibWidth = cond.getHcalCalibrationWidths(cell);
+                for (int capid=0; capid<4; ++capid) {
+                    pedsAndGains[capid] = HcalPipelinePedestalAndGain(
+                        calib.pedestal(capid), calibWidth.pedestal(capid),
+                        calib.effpedestal(capid), calibWidth.effpedestal(capid),
+                        calib.respcorrgain(capid), calibWidth.gain(capid));
+                }
+
+                // Channel quality
+                const HcalChannelStatus* digistatus = qual.getValues(rawId);
+                const bool taggedBadByDb = severity.dropChannel(digistatus->getValue());
+
+                // Fill the table entry
+                const unsigned linearId = htopo.detId2denseId(cell);
+                prod->at(linearId) = HcalChannelProperties(
+                    &calib, param_ts, channelCoder, shape,
+                    siPMParameter, pedsAndGains, taggedBadByDb);
+            }
+        }
+
+        return prod;
+    }
+
+private:
+    HcalChannelPropertiesEP() = delete;
+    HcalChannelPropertiesEP(const HcalChannelPropertiesEP&) = delete;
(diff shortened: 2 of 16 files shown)


igv4321:
HLT customization pushed

slava77:
@cmsbuild please test

slava77:
@cmsbuild please test

igv4321:
No idea why HeaderConsistency was failing. Everything compiled fine on lxplus anyway.

perrotta:
please test

mariadalfonso:
@slava77 @perrotta @igv4321 
Some reference plot for the gain in the hbhereco
https://indico.cern.ch/event/918334/contributions/3861974/attachments/2037521/3432569/PR30060.pdf
(I have not tested HF)

perrotta:
> @slava77 @perrotta @igv4321
> Some reference plot for the gain in the hbhereco
> https://indico.cern.ch/event/918334/contributions/3861974/attachments/2037521/3432569/PR30060.pdf
> (I have not tested HF)

Thank you @mariadalfonso 
Which workflow did you use in your comparisons?
Did you test the official 11_1_0_pre8 with and without this PR?
Do you have similar comparisons also for offline reco?

makortel:
> No idea why HeaderConsistency was failing. Everything compiled fine on lxplus anyway.

Header consistency (i.e. a header file `#include`ng all necessary headers) is checked in PR tests but not checked with `scram b`.

RecoLocalCalo/HcalRecAlgos/plugins/HcalChannelPropertiesEP.cc
```diff
@@ -0,0 +1,117 @@
...
+
+  inline ~HcalChannelPropertiesEP() override {}
+
+  ReturnType produce(const HcalChannelPropertiesRecord& rcd) {
+    // There appears to be no easy way to trace the internal
+    // dependencies of HcalDbService. So, rebuild the product
+    // every time anything changes in the parent records.
+    // This means that we are sometimes going to rebuild the
+    // whole table on the lumi block boundaries instead of
+    // just updating the list of bad channels.
```
makortel, line 43: There is a pattern/mechanism for partial updates, but do I understand correctly that the challenge is not in using that but somewhere else?

mariadalfonso:
> > @slava77 @perrotta @igv4321
> > Some reference plot for the gain in the hbhereco
> > https://indico.cern.ch/event/918334/contributions/3861974/attachments/2037521/3432569/PR30060.pdf
> > (I have not tested HF)
> 
> Thank you @mariadalfonso
> Which workflow did you use in your comparisons?

it's an HCALonly sequence (digi + hbheprereco).
There is no such workflow (yet, but I will add also because bocci requested it).

> Did you test the official 11_1_0_pre8 with and without this PR?

yes, fresh checkout go the 11_1_0_pre8 + cherry pick those commits from Igor
runned on the run2 data.

> Do you have similar comparisons also for offline reco?

In the offline Reco we will have: calculation of the flags (will not be there in Run3), method3 energy, arrivalTime in Mahi.
But what do we learn from adding these ? The point is that this PR reduce of 25% the overhead (clearly is only < 5% later). In Run3 we have a factor 1.7 more channels to build, so this effect will be larger.

RecoLocalCalo/HcalRecAlgos/plugins/HcalChannelPropertiesEP.cc
```diff
@@ -0,0 +1,117 @@
...
+
+private:
+  edm::ESGetToken<HcalDbService, HcalDbRecord> condToken_;
+  edm::ESGetToken<HcalTopology, HcalRecNumberingRecord> topoToken_;
+  edm::ESGetToken<HcalRecoParams, HcalRecoParamsRcd> paramsToken_;
+  edm::ESGetToken<HcalSeverityLevelComputer, HcalSeverityLevelComputerRcd> sevToken_;
+  edm::ESGetToken<HcalChannelQuality, HcalChannelQualityRcd> qualToken_;
+  edm::ESGetToken<CaloGeometry, CaloGeometryRecord> geomToken_;
+
+  std::unique_ptr<HcalRecoParams> paramTS_;
```
makortel, line 114: An ESProducer member data that gets changed for each IOV and is referenced by the produced ESProducts is incompatible with concurrent IOVs.

If I understood correctly, a pointer to the `HcalRecoParams` is passed to each `HcalChannelProperties` that need to keep the pointer as member data. Therefore concurrent IOVs could be supported if the ownership of the `HcalRecoParams` object would be given to the EventSetup system. I can think of three options
* modify `HcalChannelPropertiesVec` to hold the `HcalRecoParams` as well
  * would imply that `HcalChannelPropertiesVec` could not be simply `std::vector<HcalChannelProperties>` but a custom class with enough `vector`-like interface
* make `paramTS_` a `std::shared_ptr` and give each `HcalChannelProperties` a shared ownership to `HcalRecoParams`
  * sounds a bit overkill given that all `HcalChannelProperties` have the same life time
* split the `HcalChannelPropertiesEP::produce()` into two
  * first one produces the `paramTS_` into the `HcalChannelPropertiesRecord`
  * second one reads the `paramTS_` from `HcalChannelPropertiesRecord` and produces the `HcalChannelPropertiesVec`
     * `HcalChannelProperties` can continue use a raw pointer to refer to the `HcalRecoParams`, proper object life times are guaranteed by the EventSetup system

To me the last option sounds the cleanest one.

RecoLocalCalo/HcalRecProducers/src/HFPreReconstructor.cc
```diff
@@ -138,16 +142,14 @@ void HFPreReconstructor::fillInfos(const edm::Event& e, const edm::EventSetup& e
...
   // Get the calibrations
-  ESHandle<HcalDbService> conditions;
-  eventSetup.get<HcalDbRecord>().get(conditions);
+  ESHandle<HcalTopology> htopoHandle;
+  eventSetup.get<HcalRecNumberingRecord>().get(htopoHandle);
+  const HcalTopology& htopo(*htopoHandle);
+
+  ESHandle<HcalChannelPropertiesVec> propHandle;
+  eventSetup.get<HcalChannelPropertiesRecord>().get(propHandle);
+  const HcalChannelPropertiesVec& prop(*propHandle);
```
makortel, line 152: Could these be changed to use the `ESGetToken`?
https://twiki.cern.ch/twiki/bin/view/CMSPublic/SWGuideHowToGetDataFromES#In_ED_module
https://twiki.cern.ch/twiki/bin/view/CMSPublic/SWGuideHowToGetDataFromES#Getting_data_from_EventSetup_wit

RecoLocalCalo/HcalRecAlgos/plugins/HcalChannelPropertiesEP.cc
igv4321: The problem is that HcalDbService product is an aggregate associated with the record HcalDbRecord which has 25 parent records. Among these parent records, there are records whose IOVs are expected to change on the run boundaries and there records whose IOVs are expected to change on the lumi section boundaries. I don't want to dig into the internals of HcalDbService in order to figure out which pieces of information provided by this service depend on which record, when they get invalidated, etc.

RecoLocalCalo/HcalRecAlgos/plugins/HcalChannelPropertiesEP.cc
igv4321: The first option looks simplest to me. But could you explain why "an ESProducer member data that gets changed for each IOV and is referenced by the produced ESProducts is incompatible with concurrent IOVs"? The product is allocated on the heap, same way as a new version of HcalRecoParams. So whatever memory model is used to synchronize product making should work in this case as well.

RecoLocalCalo/HcalRecProducers/src/HFPreReconstructor.cc
igv4321: Fixes of this kind are outside of the purpose and scope of this PR. They need to be done for many modules, and it would be best if these changes are performed separately.

RecoLocalCalo/HcalRecAlgos/plugins/HcalChannelPropertiesEP.cc
igv4321: OK, never mind. I thought about it a little more and understood what the problem is.

RecoLocalCalo/HcalRecProducers/src/HFPreReconstructor.cc
makortel: Separate PR is perfectly fine, thanks.

RecoLocalCalo/HcalRecAlgos/plugins/HcalChannelPropertiesEP.cc
igv4321: In your third option, do I have to create another record? If I don't, I get a compilation error which says "Trying to get a product with an ESGetToken specifying a Record from another Record where the second Record is not dependent on the first Record.
         static_assert(!std::is_same<FoundItrT, EndItrT>::value,"

Can I have HcalChannelPropertiesRecord depend on itself?

makortel: > In your third option, do I have to create another record?

Same record should be fine.

Could you share the `ESGetToken` type and the line that gets the `HcalRecoParams` from  `HcalChannelPropertiesRecord`?

igv4321: OK, I resolved the dependency issue by introducing another record

igv4321: I tried what you suggested before, but I was getting the compilation error above. The system thinks that I can not create that token because HcalChannelPropertiesRecord does not depend on HcalChannelPropertiesRecord

makortel: I took a look out of curiosity, and
https://github.com/makortel/cmssw/commit/536c28537f6ac388cbf2951dd7f4b950dc09678b
should be sufficient to get rid of the `HcalChannelPropertiesAuxRecord`. The issue is that if we have dependent records like `A -> B -> C` (`A` depends on `B` that depends on `C`), an `ESProducer::produce(A const& rcd)` can
* get another product from `A` with `rcd.get(tokenA)`
* get a product from `B` also with `rcd.get(tokenB)`
* get a product from `C` with `rcd.getRecord<B>.get(tokenC)`

i.e. the `rcd.get(token)` extends only to the immediate parents of `rcd`. It's a known deficiency, but so far has not felt to be a big-enough problem to justify the development cost.

Some of the other `get()`s can also be simplified
https://github.com/makortel/cmssw/commit/090f84ef2ac5735dee2381fe3a9a96d62e40da13

I leave it to your and reviewer's discretion whether to simplify or not.

igv4321: Well, it seems to work as it is. Another record solves the dependency issue in the most transparent manner (at least to me).

RecoLocalCalo/HcalRecAlgos/interface/HcalPipelinePedestalAndGain.h
```diff
@@ -0,0 +1,41 @@
+#ifndef RecoLocalCalo_HcalRecAlgos_HcalPipelinePedestalAndGain_h_
+#define RecoLocalCalo_HcalRecAlgos_HcalPipelinePedestalAndGain_h_
+
+// HB/HE channel information stored for each pipeline "capacitor id"
+class HcalPipelinePedestalAndGain {
```
fwyzard, line 5: This seems to always (often, at least) be used as an `std::array<HcalPipelinePedestalAndGain, 4>`.
You could add at the end of this header file something like
```c++
constexpr int HcalPipelinePedestalAndGainSize = 4;
using HcalPipelinePedestalAndGainArray = std::array<HcalPipelinePedestalAndGain, 4>;
```
and use those in the rest of the code ?

RecoLocalCalo/HcalRecAlgos/plugins/HcalChannelPropertiesEP.cc
```diff
@@ -0,0 +1,131 @@
...
+
+  inline ~HcalChannelPropertiesEP() override {}
+
+  ReturnType1 produce1(const HcalChannelPropertiesAuxRecord& rcd) {
+    using namespace edm;
+
+    const HcalTopology& htopo = rcd.getRecord<HcalRecNumberingRecord>().get(topoToken_);
+    const HcalRecoParams& params = rcd.getRecord<HcalRecoParamsRcd>().get(paramsToken_);
+
+    ReturnType1 prod(new HcalRecoParams(params));
```
fwyzard, line 48:
```suggestion
    auto prod = std::make_unique<HcalRecoParams>(params);
```

RecoLocalCalo/HcalRecAlgos/plugins/HcalChannelPropertiesEP.cc
```diff
@@ -0,0 +1,131 @@
...
+    const HcalRecoParams& params = rcd.getRecord<HcalChannelPropertiesAuxRecord>().get(myParamsToken_);
+    const HcalSeverityLevelComputer& severity = rcd.getRecord<HcalSeverityLevelComputerRcd>().get(sevToken_);
+    const HcalChannelQuality& qual = dbRecord.getRecord<HcalChannelQualityRcd>().get(qualToken_);
+    const CaloGeometry& geom = rcd.getRecord<CaloGeometryRecord>().get(geomToken_);
+
+    // HcalTopology is taken from "params" created by the "produce1" method
+    const HcalTopology& htopo(*params.topo());
+
+    // Build the product
+    ReturnType2 prod(new HcalChannelPropertiesVec(htopo.ncells()));
```
fwyzard, line 74:
```suggestion
    auto prod = std::make_unique<HcalChannelPropertiesVec>(htopo.ncells());
```

fwyzard (commented):
Hi @igv4321,
I've added a couple of comments to make the code more readable for somebody less familiar with it.
Feel free to ignore them if you prefer the current naming.

Using `make_unique` "hides" the raw `new`, which tends to ring an alarm bell about the use of raw pointers.

igv4321:
Sorry, I am going to ignore all of these suggestions. They do not add anything to the validity of the code and boil down to personal taste. Since I am going to be the person maintaining this code for the foreseeable future, I prefer it as is.

fwyzard:
The use of `make_unique` is not a matter of personal taste, but coming as part of the CMS code rules.
Feel free to follow up in a subsequent PR.

The use of meaningful names is indeed a matter of personal taste.

fwyzard:
> Also, every Hcal person familiar with its electronics knows the pipeline depth of its readout chips, so the use of "4" will not be a surprise, etc.

Every person with a minimum of mathematical background should not find the number `3.14159265...` surprising. Yet we prefer to use named constants for it.

igv4321:
Looking at the cms coding rules at http://cms-sw.github.io/cms_coding_rules.html, it seems that the people who wrote them do not realize that "std::make_unique" is not a constructor but a utility function. Preferring utility functions to constructors is the strangest rule I've ever seen

fwyzard:
> it seems that the people who wrote them do not realize that "std::make_unique" is not a constructor but a utility function.

Please see https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#r11-avoid-calling-new-and-delete-explicitly .

igv4321:
We prefer named constants for 3.14159265... because writing M_PI is shorter. Writing "4" instead of "HcalPipelineDepth" suffices for me.

igv4321:
Looking at the reason quoted at https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#r11-avoid-calling-new-and-delete-explicitly, it is clear that this rule should not be applied to the constructors of smart pointers

fwyzard:
I find you conclusion both confusing and amusing.
In your opinion, which part of
> Warn on any explicit use of `new` and `delete`. Suggest using `make_unique` instead.

makes it
> clear that this rule should not be applied to the constructors of smart pointers

?

igv4321:
The rationale for using "make_unique" is given in 
https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#r23-use-make_unique-to-make-unique_ptrs
It is stated simply as "for convenience and consistency with shared_ptr". I do not find this reason to be convenient or compelling. 
See also https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#c150-use-make_unique-to-construct-objects-owned-by-unique_ptrs
which recommends using make_unique on the basis that without it the code becomes repetitive. Well, this looks to me again like a pure personal preference. And, strangely enough, if one does not use "auto" (which I prefer not using if I know the return type), the "repetitive" code becomes shorter than the "non repetitive" one.

fwyzard:
OK.

If you believe that the CMS Coding Rules should be updated, please bring the issue up at one of the upcoming Core Software meetings. 

In the meantime, while it's unfortunate that your personal preference differs in this case, please update this code to follow the CMS Coding Rules.

igv4321:
OK, the code has been made slightly longer and slightly less efficient by using utility functions for making unique pointers

fwyzard:
Thanks for the fix (what makes you think it would be less efficient ?)

igv4321:
Because both the utility function and the smart pointer constructor are now called instead of just the constructor. But, of course, the difference in miniscule.

fwyzard:
Actually I think, if you look at the generated code with the usual optimisations, that it is exactly the same.

igv4321:
This time you are probably right. I've just tried it with some simple code, and the generated binary was exactly the same with the -O3 flag.

igv4321:
@perrotta, @slava77 It seems like a cycle of comments/responses is over. Could we please move forward with this PR?

slava77:
@cmsbuild please test

igv4321:
The "Failed tests: AddOn" problem looks to me like something is wrong with the HLT test system and not with my PR. The HLT customization is there, and there are no problems in, for example, hlt_data_2018, hlt_data_GRun, etc.

Martin-Grunewald:
@igv4321 
Ah, perhaps you could bother looking at the failing tests? The addOn tests cover different data taking periods, so it is quite weird to blame the HLT infrastructure without commenting on the actual errors in the log files.

```
----- Begin Fatal Exception 04-Jun-2020 00:32:18 CEST-----------------------
An exception of category 'NoRecord' occurred while
   [0] Processing  stream begin Run run: 191718 stream: 1
   [1] Calling method for module EventSetupRecordDataGetter/'hltGetConditions'
   [2] Using EventSetup component HcalChannelPropertiesEP/'hcalChannelPropertiesESProd' to make data HcalChannelPropertiesVec/'' in record HcalChannelPropertiesRecord
   [3] While getting dependent Record from Record HcalChannelPropertiesRecord
Exception Message:
No "HcalDbRecord" record found in the EventSetup.n
 Please add an ESSource or ESProducer that delivers such a record.
```
This looks like caused by your changes, no? So then you get to fix it!

The 'Fake*' menus use a stripped down HLT menu for older data taking period data/MC workflows.
Maybe the customisation would not need to be or even should not be applied in that case?

Martin-Grunewald:
It may be the same symptom/possible solution as here: https://github.com/cms-sw/cmssw/pull/30046#issuecomment-637694458

igv4321:
All failing "AddOn" tests correspond to the python files in the directory HLTrigger/Configuration/test. The configuration files in the directory HLTrigger/Configuration/python work just fine. This tells me that the the code in HLTrigger/Configuration/test is not properly configured to pick up customizations.

Martin-Grunewald:
-1
I am NOT fixing your problems

fwyzard:
@igv4321 
> The "Failed tests: AddOn" problem looks to me like something is wrong with the HLT test system and not with my PR. The HLT customization is there, and there are no problems in, for example, hlt_data_2018, hlt_data_GRun, etc.

The problem is that the customisation:
```python
# Hcal calibration lookup using linear indexing
def customiseFor30060(process):
    process.hcalChannelPropertiesESProd = cms.ESProducer('HcalChannelPropertiesEP')
    return process
```
adds the `hcalChannelPropertiesESProd` unconditionally, assuming that the `HcalDbRecord` is always available.
That, however, is not the case for the "Fake", "Fake1" and "Fake2" menus.

Some different ways in which you could fix this are:
  - apply the customisation only to the menus that need it (e.g. add a check on the `menuType` before calling `process = customiseFor30060(process)`)

or
  - apply the change only if the `HcalDbRecord` is present (I think it is produced by the `HcalDbProducer`, but I'm not familiar with the HCAL DB modules structure)

or
  - always apply the change, but add the `HcalDbRecord` if it is not present

Martin-Grunewald:
Again, you are not even looking. There are cff file tests and cfg file tests. The cff file tests run. The cfg file tests fail. Both need to run and were running before your changes. So do nit blame others before you take a more detailed look!

igv4321:
@fwyzard Thanks for the constructive response!

fwyzard:
@cmsbuild, please test

Martin-Grunewald:
Hmm, this switches it off for both cff and cfg tests, while cff tests with it did run and only cfg tests failed. Maybe checking the presence of the producer in the py config is better.

igv4321:
As far as I can see, the HLT "fake" menus are present only in cfg but not in cff.

fwyzard:
They are definitely there, see https://github.com/cms-sw/cmssw/tree/master/HLTrigger/Configuration/python/ :
image

The previous customisation works for the Fake menus in cff format, because those take most of the EventSetup records and conditions from the configuration provided by cmsDriver.py, which I assume includes the `HcalDbProducer` for those scenarios.

igv4321:
OK, I stand corrected. But it is a strange arrangement.

fwyzard:
To give you a little background information:
  - the cff files under HLTrigger/Configuration/python/ are picked up by the workflows generated with cmsDriver, which are used for MC production and potentially for data reprocessing; as the HLT step runs as part of the DIGI-RAW step, the EventSetup modules that would conflict among the two parts have to be removed from the dumps of HLT menus in cff format;
  - the cfg files under HLTrigger/Configuration/test/ are stand alone tests that try to mimic as much as possible the configurations that are actually used for data taking; as such, they include all the necessary EventSetup modules.

igv4321:
OK, thanks. It is confusing that the menus work differently in "test" and "python" -- this is where my original reaction came from. Is there a standard way to diagnose HLT customization problems? I usually just run something like "runTheMatrix.py -j 5 -s --useInput all", but it does not pick up problems of this kind.

Martin-Grunewald:
Using ```addOnTests.py``` runs the add-on tests.

igv4321:
OK, thanks. Will keep that in mind.

silviodonato:
+1

PR #30061, opened 2020-05-30
Use HGCalRecHitMapProducer

lecriste:
#### PR description:

Replace all occurrences of `fillHitMap` with `HGCalRecHitMapProducer` and `std::map` with `std::unordered_map` in `HGCalRecHitMapProducer`. No changes are expected in the output.

#### PR validation:
 
`runTheMatrix -l limited`


RecoEgamma/EgammaTools/src/HGCalEgammaIDHelper.cc
```diff
@@ -25,11 +27,13 @@ void HGCalEgammaIDHelper::eventInit(const edm::Event& iEvent, const edm::EventSe
...
   edm::Handle<HGCRecHitCollection> recHitHandleBH;
   iEvent.getByToken(recHitsBH_, recHitHandleBH);
+  edm::Handle<std::unordered_map<DetId, const HGCRecHit*>> hitMapHandle;
+  iEvent.getByToken(hitMap_, hitMapHandle);
 
   recHitTools_.getEventSetup(iSetup);
   pcaHelper_.setRecHitTools(&recHitTools_);
   isoHelper_.setRecHitTools(&recHitTools_);
-  pcaHelper_.fillHitMap(*recHitHandleEE, *recHitHandleFH, *recHitHandleBH);
+  pcaHelper_.setHitMap(const_cast<std::unordered_map<DetId, const HGCRecHit*>*>(hitMapHandle.product()));
```
makortel, line 36: Casting away `const` of  input event products is not allowed.

SimCalorimetry/HGCalAssociatorProducers/plugins/LayerClusterAssociatorByEnergyScoreProducer.cc
```diff
@@ -42,9 +42,9 @@ void LayerClusterAssociatorByEnergyScoreProducer::produce(edm::StreamID,
                                                           edm::Event &iEvent,
                                                           const edm::EventSetup &es) const {
   rhtools_->getEventSetup(es);
-  edm::Handle<std::map<DetId, const HGCRecHit *>> hitMapHandle;
-  iEvent.getByToken(hitMapToken_, hitMapHandle);
-  const std::map<DetId, const HGCRecHit *> *hitMap = &*hitMapHandle;
+  edm::Handle<std::unordered_map<DetId, const HGCRecHit *>> hitMapHandle;
+  iEvent.getByToken(hitMap_, hitMapHandle);
+  const std::unordered_map<DetId, const HGCRecHit *> *hitMap = &*hitMapHandle;
```
makortel, line 47: This could also be expressed as
```suggestion
  const std::unordered_map<DetId, const HGCRecHit *> *hitMap = &iEvent.get(hitMap_);
```

RecoEgamma/EgammaTools/src/HGCalEgammaIDHelper.cc
lecriste: @makortel do you mean that I should move the `const_cast` like [here](https://github.com/cms-sw/cmssw/commit/cfcd7da245570e0c73fae960574848f183340fd5) or that I should never use it?
If the latter, [this](https://github.com/cms-sw/cmssw/pull/30061/files/f5f59f96fe6b23499fc120293838ed9e6f99f77f..cfcd7da245570e0c73fae960574848f183340fd5#diff-aab1c6f6feeda84bbf94d90c54b3a408R95) `hitMap_` can not be `const` because [this method](https://github.com/cms-sw/cmssw/pull/30061/files/f5f59f96fe6b23499fc120293838ed9e6f99f77f..cfcd7da245570e0c73fae960574848f183340fd5?file-filters%5B%5D=.cc#diff-087a9dbe46a3a27810ba50a90ef2dc28R43) modifies it. How can I work it out?

Dr15Jones: If it modifies it, you must make a copy. Under no circumstances are any modules allowed to change data in the Event, LuminosityBlock, Run or EventSetup.

lecriste: @Dr15Jones is [this](https://github.com/cms-sw/cmssw/commit/cfcd7da245570e0c73fae960574848f183340fd5#diff-087a9dbe46a3a27810ba50a90ef2dc28R34) copy OK or shall I define a `const *hitMap_`, assign it the event product and then copy the `hitMap_` to a non-const instance?

RecoEgamma/EgammaTools/src/EgammaPCAHelper.cc
```diff
@@ -29,9 +29,9 @@ EGammaPCAHelper::~EGammaPCAHelper() {
     delete hitMap_;
 }
 
-void EGammaPCAHelper::setHitMap(std::unordered_map<DetId, const HGCRecHit*>* hitMap) {
+void EGammaPCAHelper::setHitMap(const std::unordered_map<DetId, const HGCRecHit*>* hitMap) {
   hitMapOrigin_ = 1;
-  hitMap_ = hitMap;
+  hitMap_ = const_cast<std::unordered_map<DetId, const HGCRecHit*>*>(hitMap);
```
Dr15Jones, line 34: This is not allowed

RecoEgamma/EgammaTools/src/HGCalEgammaIDHelper.cc
Dr15Jones: So the problem appears to be `fillHitMap` since that appears to be the only function in `EGammaPCAHelper` that actually modifies the content of the map. (in addition `setHitMap` leaks memory). So is `fillHitMap` needed by this code?

Dr15Jones: It looks like you removed the use of `fillHitMap` with the call to `setHitMap`. In addition, only `HGCalEgammaIDHelper` uses `EGammaPCAHelper` which means now no piece of code calls `fillHitMap`. Therefore you can safely remove that function and change the internal variable from
```C++
std::unordered_map<DetId, const HGCRecHit *> *hitMap_;
```
to
```C++
std::unordered_map<DetId, const HGCRecHit *> const* hitMap_;
```

perrotta:
please test

RecoEgamma/EgammaTools/src/EgammaPCAHelper.cc
```diff
@@ -19,18 +19,13 @@ EGammaPCAHelper::EGammaPCAHelper()
       // See RecoLocalCalo.HGCalRecProducers.HGCalRecHit_cfi
       invThicknessCorrection_({1. / 1.132, 1. / 1.092, 1. / 1.084}),
       pca_(new TPrincipal(3, "D")) {
-  hitMapOrigin_ = 0;
-  hitMap_ = new std::map<DetId, const HGCRecHit*>();
+  hitMap_ = new std::unordered_map<DetId, const HGCRecHit*>();
```
Dr15Jones, line 22: This will cause a memory leak. This instance of an object is never supposed to be used so it seems you should just be able to do 
```suggestion
  hitMap_ = nullptr;
```

By the way, is `pca_` also going to cause a memory leak? What code is responsible for deleting it?

Dr15Jones: I should have checked further, `pca_` is a `std::unique_ptr` so the memory is properly managed.

Fireworks/Calo/interface/FWHeatmapProxyBuilderTemplate.h
```diff
@@ -93,40 +93,12 @@ class FWHeatmapProxyBuilderTemplate : public FWSimpleProxyBuilder {
...
-        const auto& rechitsBH = *recHitHandleBH;
+      hitmap.clear();
 
-        for (unsigned int i = 0; i < rechitsBH.size(); ++i) {
-          hitmap[rechitsBH[i].detid().rawId()] = &rechitsBH[i];
-        }
-      }
+      edm::Handle<std::unordered_map<DetId, const HGCRecHit*>> hitMapHandle;
+      event->getByLabel(edm::InputTag("hgcRecHitMapProducer"), hitMapHandle);
+      hitmap = *hitMapHandle;
```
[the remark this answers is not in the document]
alja: I'm not the author of the code and I don't know why map is copied.  The map does not use a lot of memory because it contains references to objects.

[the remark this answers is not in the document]
rovere: The map is used to colour the hits (or any other HGCAL object supported by Fireworks) according to its energy, with a variable cut-off. 
I agree with @kpedro88: a const ref is a better choice.

alja:
+1

slava77:
@cmsbuild please test

jfernan2:
+1

perrotta:
please test

alja:
+1

perrotta:
While measuring the CPU time performance for this PR (by the way: no significant changes in time due to it when measured on wf 21234.0 TTbar_14TeV 2026D44) I notice that the step3 crashes if I don't run the DQM/Validation modules.

In fact, the reco process hgcElectronID depends on  ```hitMapTag = cms.InputTag("hgcRecHitMapProducer")```, which is run in the validation Task
 ```
process.hgcalAssociators = cms.Task(process.hgcRecHitMapProducer, process.lcAssocByEnergyScoreProducer)

kpedro88:
please test workflow 20434.21
(prodlike workflows don't include validation/dqm)

rovere:
@lecriste this is the error message of workflow 20434.21:

```
----- Begin Fatal Exception 04-Jun-2020 21:25:12 CEST-----------------------
   [1] Running path 'MINIAODSIMoutput_step'
   [2] Prefetching for module PoolOutputModule/'MINIAODSIMoutput'
   [3] Prefetching for module PATElectronSlimmer/'slimmedElectronsFromMultiCl'
   [4] Prefetching for module PATElectronSelector/'selectedPatElectronsFromMultiCl'
   [5] Prefetching for module PATElectronProducer/'patElectronsFromMultiCl'
   [6] Calling method for module HGCalElectronIDValueMapProducer/'hgcElectronID'
Exception Message:
Looking for type: std::unordered_map<DetId,const HGCRecHit*,std::hash<DetId>,std::equal_to<DetId> >
Looking for module label: hgcRecHitMapProducer

   Additional Info:

```

kpedro88:
@rovere it's the same problem noted by @perrotta in https://github.com/cms-sw/cmssw/pull/30061#issuecomment-638997218. I just wanted to make sure it shows up in the tests here (since it will show up in the IB if the PR is merged as is).

rovere:
Ciao,
I'd definitely revert the dependency arrow: from reconstruction down to validation since we will likely run the former without the latter but not vice-versa.
I guess the only remaining issue is to identify which is the best place to put this.
@lecriste maybe we can embed this task into the `HGCAL` rechits reconstruction sequence.

rovere:
@lecriste sorry to bother you, again, but I do not particularly like the fact that a map that is constructed from RecHits has its definition in a package from `SimCalorimetry`.
I'd, therefore, move the configuration of this piece of code into a package coming from `RecoLocalCalo`, possibly `HGCAL` related.

lecriste:
please test workflow 20434.21

kpedro88:
please test workflow 20434.21

jfernan2:
+dqm

kpedro88:
+upgrade

civanch:
+1

perrotta:
@alja can you please evaluate and sign for visualization (if you think so)?

alja:
+1

silviodonato:
+1

PR #30063, opened 2020-05-31
MTD reco geometry: revision of DetLayers verbosity and test

fabiocos:
#### PR description:

In preparation of the update of ```RecoMTD/DetLayers``` needed by the addition of the new two-layers ETL design, this PR refreshes the verbosity of the package (for a more flexible management during debugging) and update its test part.

As a side note, also the verbosity of the ```TrackingTools/DetLayers``` base classes could be refreshed in case.

#### PR validation:

The test runs smoothly, with no change in its output on D50 compared to the previous state of the code.


fabiocos:
please test

cvuosalo:
+1

RecoMTD/DetLayers/src/MTDDetRing.cc
```diff
@@ -12,7 +12,10 @@
 #include <iostream>
 #include <vector>
 
+//#define EDM_ML_DEBUG
+
 using namespace std;
+const std::string MTDDetRing::metname = "MTD|RecoMTD|RecoMTDDetLayers|MTDDetRing";
```
slava77, line 18: wouldn't this trigger a quadruplicate printout (one for each category)?

RecoMTD/DetLayers/plugins/BTLDetLayerGeometryBuilder.cc
```diff
@@ -33,26 +35,33 @@ vector<DetLayer*> BTLDetLayerGeometryBuilder::buildLayers(const MTDGeometry& geo
         const GeomDet* geomDet = geo.idToDet(BTLDetId(side, tray, module, 0, 1));
         if (geomDet != nullptr) {
           geomDets.push_back(geomDet);
-          LogTrace(metname) << "get BTL module " << std::hex << BTLDetId(side, tray, module, 0, 1).rawId() << std::dec
-                            << " at R=" << geomDet->position().perp() << ", phi=" << geomDet->position().phi();
+#ifdef EDM_ML_DEBUG
+          edm::LogVerbatim(metname) << "get BTL module " << std::hex << BTLDetId(side, tray, module, 0, 1).rawId()
```
slava77, line 39: LogVerbatim with EDM_ML_DEBUG has the same behavior as LogTrace.
Is there a substantial problem in enabling the debug/trace level via MessageLogger? 
Info is not on by default either; so, I'd guess that ability to enable is there.

slava77 (commented):
I stopped after a few files.
The new pattern of LogInfo/LogVerbatim behind EDM_ML_DEBUG seems "novel"; what is the practical use compared to the existing functionality?

I see also that a few Log categories were extended to multiple variants: are these really all useful?(IIRC, if all categories are enabled, the printout will appear once for each case, which would seem excessive.)

fabiocos:
@slava77 thanks for the comments. 

- concerning the use of LogInfo/LogVerbatim + EDM_ML_DEBUG vs LogDebug/LogTrace it is likely a novelty in the reconstruction area, it is used everywhere in geometry and simulation.  To me (and not only) it looks easier to configure and use with minimal modifications of the default MessageLogger configuration compared to the LogDebug management, which I find heavier and impractical. Of course this is nothing compelling, one can spend some time and write down an ad-hoc MessageLogger configuration for LogTrace and thats it. With this approach it is easier to selectively activate or deactivate the verbosity in some specific file to me.

- as far as ```metname``` and its patterns is concerned: this is something funny in the reco code that you likely may explain better than me, I am curious, because it is largely spread in the Reco area: ```RecoMuon```,```RecoEgamma```,```TrackingTools```,```TrackPropagation```. What I did was just to avoid replicas of the declaration in each function, sometimes not done, making the handling of the message homogeneous throughout each class. And fix some clear cut-and-paste remnant of the end-of-2018 review (MTD is not Muon). Why ```metname```? It looks like unrelated to MET code, may be "name of MessageLogger T*?" Why a complex pattern (not practical fo the intended purpose of MessageLogger categories)? Likely to provide the full pattern of a given class, as far as I see. Calling the message ```theMetname``` it's ok with me, I do not care at all, I was more interested in a homogeneous and not redundant management of the message. Otherwise better to just drop this common pattern and to something more usual IMO.

RecoMTD/DetLayers/interface/MTDDetRing.h
```diff
@@ -42,6 +42,7 @@ class MTDDetRing : public ForwardDetRingOneZ {
 private:
   typedef PeriodicBinFinderInPhi<float> BinFinderType;
   BinFinderType theBinFinder;
+  static const std::string metname;
```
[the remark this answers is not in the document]
fabiocos: see here and for next instances the general comment below

RecoMTD/DetLayers/src/MTDDetRing.cc
fabiocos: @slava77 yes, you are correct, it does, but this is what it already was present, and it is used in several other places in the RECO code. Are you suggesting to simplify it? It might make sense to me.

RecoMTD/DetLayers/plugins/BTLDetLayerGeometryBuilder.cc
fabiocos: it just requires less manipulations of the default MessageLogger configuration to be selected in my view.

slava77:
> @slava77 thanks for the comments.
> 
>     * concerning the use of LogInfo/LogVerbatim + EDM_ML_DEBUG vs LogDebug/LogTrace it is likely unusual in the reconstruction area, it is used everywhere in geometry and simulation.  To me (and apparently not only) it looks easier to configure and use with minimal modifications of the default MessageLogger configuration compared to the LogDebug management, which I find heavier and impractical. Of course this is nothing compelling, one can spend some time and write down an ad-hoc MessageLogger configuration for LogTrace and thats it. With this approach it is easier to selectively activate or deactivate the verbosity in some specific file to me.

If the point is about compiling with EDM_ML_DEBUG enabled file by file, then it can be done with LogTrace just as well, you only need to define EDM_ML_DEBUG before including MessageLogger (and perhaps also include the MessageLogger header last out of all CMSSW includes in a given file).

If the issue is with using LogVerbatim itself after EDM_ML_DEBUG is defined, please elaborate a bit, perhaps with some pseudo-code example.
I thought that the only difference in the MessageLogger setup is INFO vs DEBUG in the threshold.

> 
>     * as far as `metname` and its patterns is concerned: this is something funny in the reco code that you likely may explain better than me, I am curious, because it is largely spread in the Reco area: `RecoMuon`,`RecoEgamma`,`TrackingTools`,`TrackPropagation`. What I did was just to avoid replicas of the declaration in each function, sometimes not done, making the handling of the message homogeneous throughout each class. 

I see, so it sounds like this variety of categories is simply copy-pasted from the muon code and there is no actual need for multiple categories.
I suggest to simplify and pick just one, perhaps even geometry-specific named category (MTDGeometry?).

> And fix some clear cut-and-paste remnant of the end-of-2018 review (MTD is not Muon, all this was clearly copied in MTD code from somewhere else). Why `metname`? It looks like unrelated to MET code, may be "name of MessageLogger T*?" Why a complex pattern (~not practical fo the intended purpose of MessageLogger categories~)? Likely to provide the full pattern of a given class, as far as I see. Calling the message `theMetname` it's ok with me, but according to https://cms-sw.github.io/cms_coding_rules.html#2----naming-rules-1 point 9 should better be `metname_`, or, it could even be `theMetname_`. 

"the", "m_" or any common decorator, following existing style would work, as long as it's consistent for the class.
`theMetname_` by itself will still break consistency with most of the existing data members that use "the"

> Code aesthetics apart, I was more interested in a homogeneous and not redundant management of the message. Otherwise better to just drop this common pattern and to something more usual IMO.

RecoMTD/DetLayers/interface/MTDDetRing.h
```diff
@@ -42,7 +42,7 @@ class MTDDetRing : public ForwardDetRingOneZ {
 private:
   typedef PeriodicBinFinderInPhi<float> BinFinderType;
   BinFinderType theBinFinder;
-  static const std::string metname;
+  static const std::string theMetname_;
```
slava77, line 45: as I mentioned in the main thread the trailing underscore breaks consistency with naming used for existing data members.
Please drop the underscore (the other alternative is to add it to all other existing data members)

RecoMTD/DetLayers/src/MTDDetRing.cc
slava77: yes, please keep it simple (or at least do not increase complexity compared to what it was when using a single category of "MTDDetRing")

fabiocos:
@slava77 following our discussion: leaving the ```#define EDM_ML_DEBUG``` statement commented is not helping, uncommenting it alone does not work to activate LogTrace.

I have also turned a few LogInfo into LogError as it looked to me that the treatment of not implemented methods was inconsistent across the code.

fabiocos:
please test

slava77:
> @slava77 following our discussion: leaving the `#define EDM_ML_DEBUG` statement commented is not helping, uncommenting it alone does not work to activate LogTrace.
> 

do you have an example? (maybe in a separate commit or branch to check)

RecoMTD/DetLayers/src/MTDDetRing.cc
```diff
@@ -58,20 +60,22 @@ pair<bool, TrajectoryStateOnSurface> MTDDetRing::compatible(const TrajectoryStat
...
 
   // Propagate and check that the result is within bounds
   pair<bool, TrajectoryStateOnSurface> compat = compatible(startingState, prop, est);
   if (!compat.first) {
-    LogTrace(metname) << "    MTDDetRing::compatibleDets: not compatible"
-                      << "    (should not have been selected!)";
+#ifdef EDM_ML_DEBUG
+    LogTrace("MTDDetLayers") << "    MTDDetRing::compatibleDets: not compatible"
+                             << "    (should not have been selected!)";
+#endif
```
slava77, line 78: having EDM_ML_DEBUG around LogTrace looks completely redundant.
this looks like an incomplete cleanup (I see a few more in other places of this PR).
Or did I miss something?

fabiocos:
please test

fabiocos:
this does not work, I already know. But let me try one more thing...

fabiocos:
@slava77 ok, this latest version works, the trick is to declare ```EDM_ML_DEBUG``` before *any* include that might contain ```MessageLogger.h```, where the macro defining ```LogDebug``` is defined. Of course I comment the statement for regular use, but I have tested it works as expected if uncommented.

@cvuosalo @makortel this may be a follow up of the recent discussion about``` LogVerbatim +EDM_ML_DEBUG``` vs ```LogTrace```

fabiocos:
please test

cvuosalo:
+1

kpedro88:
+upgrade

slava77:
+1

for https://github.com/cms-sw/cmssw/pull/30063	7ba026f 

silviodonato:
+1

PR #17087, opened 2016-12-21
Further performance improvement in BoostedTauSeeds

lgray:
As discussed at the end of: https://github.com/cms-sw/cmssw/pull/17040

This PR uses a cache to collapse a double loop.

PR #30064, opened 2020-06-01
Fix bug affecting boosted tau isolation

mbluj:
#### PR description:

This PR fixes a bug affecting indexing jet-constituent-maps. In the result of the bug constituents of a subjet nearby a considered tau - a seed of a nearby tau candidate - are not excluded from particles used to compute isolation of the tau, which breaks the boosted tau algorithm. The bug has been introduced in #17087 (for 90X) speeding up boosted tau seed producer code. Unfortunately the bug was discovered only recently by users (*).
In addition to the bug fix, in this PR we remove remnants of previous implementation changed by #17087.

Expected changes: boosted taus (in MiniAOD), especially those with other tau candidate within R<0.5, are more isolated. This will recover states from before #17087 as shown in the attached figure.

BoostedH

We intend to backport the fix to 11_1_X (using current dev branch) and 10_6_X for analyses based on those releases, i.e. respectively preparation to Run3 and Run2 legacy analyses.

(*) Unfortunately, developers and users of boosted taus prepared their analyses with 80X samples and then changed interests or even left CMS. In addition the code resides beyond tau packages and the change has not been consulted with Tau POG... This suggest need of more focus on monitoring boosted Taus, but this requires DQM based on miniAOD objects not present for taus, yet.

#### PR validation:

Tested with the standard miniAOD workflow.

### DIFF
--- a/RecoJets/JetProducers/plugins/BoostedTauSeedsProducer.cc
+++ b/RecoJets/JetProducers/plugins/BoostedTauSeedsProducer.cc
@@ -157,10 +157,7 @@ namespace {
       const reco::Jet& jet,
       const edm::Handle<reco::PFCandidateCollection>& pfCandidates,
       const JetToConstitMap::value_type& constitmap,
-      const reco::Jet::Constituents& jetConstituents,
-      double /*dRmatch*/,
       bool invert) {
-    //const double dRmatch2 = dRmatch*dRmatch; // comment out for now in case someone needs a dR-based search again
     auto const& collection_cand = (*pfCandidates);
     std::vector<reco::PFCandidateRef> pfCandidates_exclJetConstituents;
     size_t numPFCandidates = pfCandidates->size();
@@ -257,9 +254,9 @@ void BoostedTauSeedsProducer::produce(edm::Event& evt, const edm::EventSetup& es
 
     // find all PFCandidates that are not constituents of the **other** subjet
     std::vector<reco::PFCandidateRef> pfCandidatesNotInSubjet1 = getPFCandidates_exclJetConstituents(
-        *subjet1, pfCandidates, constitmap[2 * idx], subjetConstituents2, 1.e-4, false);
+        *subjet1, pfCandidates, constitmap[2 * idx + 1], false);
     std::vector<reco::PFCandidateRef> pfCandidatesNotInSubjet2 = getPFCandidates_exclJetConstituents(
-        *subjet2, pfCandidates, constitmap[2 * idx + 1], subjetConstituents1, 1.e-4, false);
+        *subjet2, pfCandidates, constitmap[2 * idx], false);
     if (verbosity_ >= 1) {
       std::cout << "#pfCandidatesNotInSubjet1 = " << pfCandidatesNotInSubjet1.size() << std::endl;
       std::cout << "#pfCandidatesNotInSubjet2 = " << pfCandidatesNotInSubjet2.size() << std::endl;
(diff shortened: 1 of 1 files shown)


perrotta:
please test

slava77:
> The bug has been introduced in #17087 (for 90X) speeding up boosted tau seed producer code.

what is the impact of this fix on the CPU time?
Please check and update the PR description.

> but this requires DQM based on miniAOD objects not present for taus, yet.

What is missing in the miniAOD jobs to allow for the DQM to run?

mbluj:
> 
> 
> > The bug has been introduced in #17087 (for 90X) speeding up boosted tau seed producer code.
> 
> what is the impact of this fix on the CPU time?
> Please check and update the PR description.
> 
> > but this requires DQM based on miniAOD objects not present for taus, yet.
> 
> What is missing in the miniAOD jobs to allow for the DQM to run?



> 
> 
> > The bug has been introduced in #17087 (for 90X) speeding up boosted tau seed producer code.
> 
> what is the impact of this fix on the CPU time?
> Please check and update the PR description.

No, should not affect CPU time. It is as it does not change algorithmic improvements introduced by #17087 (replacement of looping over jet constituents by search of a key stored in unordered_set), but just fixes indexes of maps used (2nd map is used in 1st call of affected function and 2nd in 1st rather than 1st index in 1st call and 2nd in 2nd). The following tau modules are run in the same way as before the fix and should not be affected by reverted indexes or even work a bit faster thanks to a smaller number of isolation candidates.  
> 
> > but this requires DQM based on miniAOD objects not present for taus, yet.
> 
> What is missing in the miniAOD jobs to allow for the DQM to run?

I think that nothing is missing, it is just a lack of standard monitoring workflow for taus in miniAOD. As far as I know it is foreseen, but not present, yet. One should also find a good set of samples to monitor boosted taus as I am affraid that standard relvals used for tau monitoring (ZTT, qqHTT) do not cover relevant phase-space. Regardless, the comment is meant to inform that we note the lack of boosted taus monitoring and need to fix it.

mbluj:
Adding @abdollah110 who found the bug.

perrotta:
@mbluj 
The jenkins tests outputs show differences only for boosted taus in miniAOD, as expected.
Where can I find a map of the TauID content in miniAOD, just to get a quick idea about the differences shown, e.g., in https://cmssdt.cern.ch/SDT/jenkins-artifacts/baseLineComparisons/CMSSW_11_2_X_2020-06-01-1100+51d92b/36739/validateJR/all_OldVSNew_RunJetHT2017FreMINIAODwf136p8311/ ?

mbluj:
TauID list as in recent 11_2_X IB attached
[tauID_list_11_2_X.txt](https://github.com/cms-sw/cmssw/files/4722404/tauID_list_11_2_X.txt)
I hope it helps. In general, I expect some changes in a number of tauIDs as many of them examine activity around tau ("isolation").

slava77:
a manual check can be done with 
compare_patTaus_IDs from https://github.com/slava77/cms-reco-tools/blob/master/printFWLite.C#L219
with smth like
```
.L ~/tools/printFWLite.C++
std::string fNameRef = "ref.root"
std::string fNameNew = "new.root"
compare_patTaus_IDs(fNameRef, fNameNew, "slimmedTaus");
```
it should print the ID re-index from the first event and show if there are differences in existing IDs

perrotta:
+1
- The fix acts as it should, following the PR description and the intention behind the original code
- Jenkins tests pass and only show differences in the boosted tau id objects in miniAOD, as expected

silviodonato:
+1

PR #30066, opened 2020-06-01
Fix bug affecting boosted tau isolation (10_6_X)

mbluj:
#### PR description:

This PR fixes a bug affecting indexing jet-constituent-maps. In the result of the bug constituents of a subjet nearby a considered tau - a seed of a nearby tau candidate - are not excluded from particles used to compute isolation of the tau, which breaks the boosted tau algorithm. The bug has been introduced in #17087 (for 90X) speeding up boosted tau seed producer code. Unfortunately the bug was discovered only recently by users (*).
In addition to the bug fix, in this PR we remove remnants of previous implementation changed by #17087 and improve code formatting. 

Expected changes: boosted taus (in MiniAOD), especially those with other tau candidate within R<0.5, are more isolated. This will recover state from before #17087 as shown in the figure attached to #30064.

Backport of #30064 to 10_6_X. Some additional details therein.

#### PR validation:

Tested with the standard miniAOD workflow.

Backport of  #30064

### DIFF
--- a/PhysicsTools/PatAlgos/python/slimming/miniAOD_tools.py
+++ b/PhysicsTools/PatAlgos/python/slimming/miniAOD_tools.py
@@ -332,10 +332,6 @@ def miniAOD_customizeCommon(process):
     addBoostedTaus(process)
     process.load("RecoTauTag.Configuration.RecoPFTauTag_cff")
     process.load("RecoTauTag.Configuration.HPSPFTaus_cff")
-    #enable correct behaviour of overlap removal in boosted tau seeding
-    from Configuration.Eras.Modifier_run2_miniAOD_devel_cff import run2_miniAOD_devel
-    run2_miniAOD_devel.toModify(process.boostedTauSeeds,
-                                correctlyExcludeOverlap = True)
     #-- Adding customization for 94X 2017 legacy reMniAOD
     from Configuration.Eras.Modifier_run2_miniAOD_94XFall17_cff import run2_miniAOD_94XFall17
     _makePatTausTaskWithRetrainedMVATauID = process.makePatTausTask.copy()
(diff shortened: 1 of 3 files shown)


slava77:
@mbluj 
please update the PR title ("ix" -> "fix").

Do I understand correctly that this update is violating the no-change policy?
(or is the modified code enabled only on analysis jobs running on miniAOD)
If the regular miniAOD production is affected, please update to use run_miniAOD_devel

mbluj:
> 
> 
> @mbluj
> please update the PR title ("ix" -> "fix").
> 
> Do I understand correctly that this update is violating the no-change policy?
> (or is the modified code enabled only on analysis jobs running on miniAOD)
> If the regular miniAOD production is affected, please update to use run_miniAOD_devel

Hmm, yes, it changes isolationCandidates of boosted taus and their isolation-related tauIDs, but not taus itself. In this sense this breaks compatibility with already produced 106X miniAOD samples. It is be a bit strange to introduce a configuration parameter with switches on/off a bug, but of course I will introduce it.

perrotta:
backport of #30064

mbluj:
Adding @abdollah110 who found the bug.

mbluj:
@slava: a flag enabling/disabling fixed behaviour of the boosted tau seeding is added in the most recent commit. Please consider to add this fix to the list of UL reMiniAOD features.

silviodonato:
please test

PhysicsTools/PatAlgos/python/slimming/miniAOD_tools.py
```diff
@@ -332,6 +332,10 @@ def miniAOD_customizeCommon(process):
     addBoostedTaus(process)
     process.load("RecoTauTag.Configuration.RecoPFTauTag_cff")
     process.load("RecoTauTag.Configuration.HPSPFTaus_cff")
+    #enable correct behaviour of overlap removal in boosted tau seeding
```
perrotta, line 335: Please move into RecoTauTag/Configuration/python/boostedHPSPFTaus_cff.py, where ```boostedTauSeed``` is defined

mbluj: Moved as suggested.
I am only a bit afraid that this modifier can be overlooked when preparing UL reMiniAOD one as it is now outside PAT packages...

perrotta: Thank you @mbluj 
The reason to allow modifier-driven changes only in the config which defines the module configuration is exactly because otherwise it could be difficult to track the modifications applied in non related packages.

perrotta:
please test

silviodonato:
+1

PR #30054, opened 2020-05-30
optimize: rely on Candidate pt,eta,phi without casting in MuonPFIsolationWithConeVeto

slava77:
visual optimization based on https://jiwoong.web.cern.ch/jiwoong/cgi-bin/igprof-navigator/igprofCPU_CMSSW_11_1_0_pre8PAT/28

PR #30071, opened 2020-06-01
skip complicated checks for PFCandidate-on-PFCandidate TopProjector

slava77:
another case of CPU optimization/minimization from PU200 hotspots
https://jiwoong.web.cern.ch/jiwoong/cgi-bin/igprof-navigator/igprofCPU_CMSSW_11_1_0_pre8PAT/16

TopProjector matching collections of PFCandidate-on-PFCandidate by Ptr can now check for just the available pointer value, without navigating deep to the parent pointers. This is enabled for pfNoPileUp\*. Other cases can possibly be made from different collections where the deeper matching may be useful.

Time spent in pfNoPileUpJMENoHF and pfNoPileUpIsoPFBRECO goes down by about a factor of 10, or 19% of the miniAOD step in wf 23434.21 (PU200 ttbar) relative to a separate improvement in #30054 . 
The impact is small for Run2/3 (perhaps up to 4% for "nominal" pileup cases).

tests were limited to the miniAOD wf with PU.

### DIFF
--- a/CommonTools/ParticleFlow/plugins/TopProjector.h
+++ b/CommonTools/ParticleFlow/plugins/TopProjector.h
@@ -1,5 +1,5 @@
-#ifndef PhysicsTools_PFCandProducer_TopProjector_
-#define PhysicsTools_PFCandProducer_TopProjector_
+#ifndef CommonTools_ParticleFlow_TopProjector_
+#define CommonTools_ParticleFlow_TopProjector_
 
 // system include files
 #include <iostream>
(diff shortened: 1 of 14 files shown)


slava77:
@cmsbuild please test

slava77:
@smuzaffar 
I was trying to navigate in the relmon to the differences
e.g. in the following
https://cmssdt.cern.ch/SDT/jenkins-artifacts/baseLineComparisons/CMSSW_11_2_X_2020-06-01-2300+87c83b/36744/10224.0_TTbar_13+TTbar_13TeV_TuneCUETP8M1_2017PU_GenSimFull+DigiFullPU_2017PU+RecoFullPU_2017PU+HARVESTFullPU_2017PU+NanoFull_2017PU/Physics_Top.html
the summary shows 3 failures in comparisons in SingleTopElectronMediumDQM_miniAOD subdirectory,
but the link shown comes up with "There are no changes in this directory, nothing to be shown or navigated to. "
https://cmssdt.cern.ch/SDT/jenkins-artifacts/baseLineComparisons/CMSSW_11_2_X_2020-06-01-2300+87c83b/36744/10224.0_TTbar_13+TTbar_13TeV_TuneCUETP8M1_2017PU_GenSimFull+DigiFullPU_2017PU+RecoFullPU_2017PU+HARVESTFullPU_2017PU+NanoFull_2017PU/Physics_Top_SingleTopElectronMediumDQM_miniAOD.html
Did the bot scripts end up pruning the directories too much?

DQM bin-to-bin confirms that there are actual differences
https://tinyurl.com/yczuod7o

smuzaffar:
@slava77 , this looks strange. I am re-running the comparison job to see if it fails in the same way.

smuzaffar:
@slava77 , yes looks like bot is deleting too much. This is due to 'SingleTopMuonMediumDQM' which is all green , so bot decided to delete all 'SingleTopMuonMediumDQM_*' we resulted in deleting 'SingleTopMuonMediumDQM_miniAOD'  . The script we run for clesnup is https://github.com/cms-sw/cms-bot/blob/master/comparisons/filter-failed-comparison.py . Any idea how to avodi such situations ?

slava77:
> The script we run for clesnup is https://github.com/cms-sw/cms-bot/blob/master/comparisons/filter-failed-comparison.py . Any idea how to avodi such situations ?

```python
for h in sorted(htmls):
  for s in all_ok[::-1]:
    if h.startswith(s):
```
is it a matter of comparing the full h? I'm not sure of my python reading capability here; does the h either contain / or ends with s for a more precise match?

smuzaffar:
h is e.g `['Physics', 'Physics_Top', 'Physics_Top_SingleTopElectronMediumDQM',  'Physics_Top_SingleTopElectronMediumDQM_minAOD']` . So h=`Physics_Top_SingleTopElectronMediumDQM` (which is all green) will be added to `all_ok` and then s=`Physics_Top_SingleTopElectronMediumDQM` will match start of h and ignore it (assuming that all `Physics_Top_SingleTopElectronMediumDQM_*` should be green too.

slava77:
uhm, so we are confusing ```_``` as a dir separator and an actual character in the dir name.
Would it make sense to change the dir separator to ```__``` ? although this may still find a conflict with a dir name if someone is creative enough

smuzaffar:
yes directory seperator is the issue here. Any change for this ned to go in all IBs :-( as this comes from some cmssw code. 
Let me see if I can improve the logic by going through the header files and look for the parent path. For example if`SingleTopElectronMediumDQM` is all green then all html files where parent path is `SingleTopElectronMediumDQM.html` can be dropped.

smuzaffar:
@slava77 , https://github.com/cms-sw/cms-bot/pull/1316 should fix the filtering issue. The change makes sure that we only drop those files which are all green. This means we have to read all html files but it least this will generate correct results. 

I have started the comparison of this PR with https://github.com/cms-sw/cms-bot/pull/1316

smuzaffar:
@slava77 , comparison results look good now.

slava77:
@cmsbuild please test with cms-sw/cms-bot#1319

slava77:
@cmsbuild please test with cms-sw/cms-bot#1319

slava77:
> ---> test runtestPhysicsToolsPatAlgos had ERRORS

I'm slowly becoming an example of how not to update PRs

slava77:
@cmsbuild please test with cms-sw/cms-bot#1319

jfernan2:
+1

slava77:
+1

for https://github.com/cms-sw/cmssw/pull/30071	3804249
- jenkins tests confirm no differences
- local test with 136.88811 (UL reminiAOD) shows about 2.5% speedup; miniAOD step of PU200 in 23434.21 has 18% speedup compared to CMSSW_11_2_X_2020-05-30-1100 as a baseline

slava77:
@makortel @Dr15Jones 
please check this PR and comment or sign
Thank you.

FWCore/ParameterSet/python/MassReplace.py
```diff
@@ -1,5 +1,19 @@
 from __future__ import print_function
 import FWCore.ParameterSet.Config as cms
+from FWCore.ParameterSet.Types import _ProxyParameter
+import copy
+
+def isinstanceWproxy(value, typeId):
```
Dr15Jones, line 6: I think this is asking the wrong question. Checking for `isinstance` in python is usually a poor choice because the language uses _duck typing_ . (Yes, we have `isinstance` in many places in our code for historic reasons and because I'm a poor python programmer :) ). I don't think we want to as an `isinstance` question, instead we want to extend our parameter type classes with a `canAssignUsing` method. Then anything that can be assigned from  a `cms.InputTag` would work properly.

slava77: OK, please let me know if ```canAssignUsing``` can be added to our python Types soon.
Will it be something that can be backported to 11_1_X?

Dr15Jones:
See #30164

slava77:
@cmsbuild please test with #30164

jfernan2:
+1

slava77:
+1

for https://github.com/cms-sw/cmssw/pull/30071	4f3068e 
- updated to remove MassReplace.py changes in lieu of  #30164  (now merged)
- jenkins tests confirmed there are no differences in comparisons

slava77:
@santocch 
please check or sign this PR for analysis
Thank you.

@silviodonato 
please check if this can be merged without the analysis signature if it takes some time.
Thank you.

santocch:
+1

silviodonato:
+1

silviodonato:
+1

PR #30088, opened 2020-06-03
Adding new R9 plots to ECAL

alejands:
#### PR description:

Since some sections of the current ECAL DQM code were written originally, there have been some minor changes in how variables important to the monitoring are commonly understood by ECAL experts. Specifically, there are a few ways of defining R9, all of which might give slightly different values for the variable -- whether to use the zero-suppressed energy, whether to use the raw values for the energy, and so on. And so the current plots are sometimes a little difficult to interpret -- for example, there was an inconsistency noted in the R9 1D distributions between UL and EOY rereco for some samples, and we are investigating whether this has to do with the old definition.

#### PR validation:

Validation of added SC energy and R9 definitions have been done by processing some sample runs with ECAL DQM code and uploading them to a private DQM GUI. Plots appear to have no issues.

### DIFF
--- a/DQM/EcalMonitorTasks/python/ClusterTask_cfi.py
+++ b/DQM/EcalMonitorTasks/python/ClusterTask_cfi.py
@@ -184,6 +184,42 @@ ecalClusterTask = cms.untracked.PSet(
             btype = cms.untracked.string('User'),
             description = cms.untracked.string('Distribution of E_seed / E_3x3 of the super clusters.')
         ),
+        SCR9Raw = cms.untracked.PSet(
+            path = cms.untracked.string('%(subdet)s/%(prefix)sClusterTask/%(prefix)sCLT SC R9 Raw'),
+            kind = cms.untracked.string('TH1F'),
+            otype = cms.untracked.string('Ecal2P'),
+            xaxis = cms.untracked.PSet(
+                high = cms.untracked.double(1.2),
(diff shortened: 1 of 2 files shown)


DQM/EcalMonitorTasks/python/ClusterTask_cfi.py
```diff
@@ -236,6 +272,19 @@
...
         ),
+        SCRawE = cms.untracked.PSet(
+            path = cms.untracked.string('%(subdet)s/%(prefix)sClusterTask/%(prefix)sCLT SC raw energy'),
+            kind = cms.untracked.string('TH1F'),
+            otype = cms.untracked.string('Ecal2P'),
+            xaxis = cms.untracked.PSet(
+                high = cms.untracked.double(150.0),
+                nbins = cms.untracked.int32(50),
+                low = cms.untracked.double(0.0),
+                title = cms.untracked.string('energy (GeV)')
```
jfernan2, line 283: May I ask why the 3 definitions above do not have a title too?

jfernan2: Answering myself: I understand because they are adimensional

jfernan2:
please test

jfernan2:
+1
10 histograms added and filled

silviodonato:
+1

PR #30092, opened 2020-06-03
fix cmssw patch version for non-standard IBs e.g ROOT6, DEVEL

smuzaffar:
#### PR description:

This fixes the unit test failure (added  by #29976 ) in non-standard IBs (e.g DEVEL, ROOT6 etc). Code here https://github.com/cms-tau-pog/cmssw/blob/7aaf1ab52d1f044b47e43ce446576a726fb2970e/RecoTauTag/RecoTau/python/tools/runTauIdMVA.py#L70 Assumes that that patch version of cmssw is either `X` or a number but for non-standard IBs it could be `ROOT6`, `DEVEL` etc. The change proposed here should fix this.

#### PR validation:

Build and successfully run unit test locally for both standard and non-standard IBs.


smuzaffar:
please test

slava77:
+1

for https://github.com/cms-sw/cmssw/pull/30092	ab45056
- technical

silviodonato:
+1

PR #30106, opened 2020-06-04
fix direct access to tau ID container

swozniewski:
#### PR description:

This PR fixes the index mapping for direct access to tau ID containers. The functionality is only for private use and test purposes and does not concern existing workflows.
The fixed lines can be compared to the regular workflows some lines above, which are using provenance, e.g. https://github.com/cms-sw/cmssw/blob/master/PhysicsTools/PatAlgos/plugins/PATTauProducer.cc#L405-L409 and are supposed to follow the same logic of the internal index mapping.

#### PR validation:

compiles, no further checks run due to simplicity of changes

### DIFF
--- a/PhysicsTools/PatAlgos/plugins/PATTauProducer.cc
+++ b/PhysicsTools/PatAlgos/plugins/PATTauProducer.cc
@@ -414,9 +414,10 @@ void PATTauProducer::produce(edm::Event& iEvent, const edm::EventSetup& iSetup) 
               try {
                 int i = std::stoi(prov_ID_label);
                 if (prov_cfg_label == "direct_rawValues")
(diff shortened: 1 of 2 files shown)


jfernan2:
please test

jfernan2:
please test
Triggering the test agian to decouple from external merge commits included above

jfernan2:
+1
No changes

slava77:
type bugfix

slava77:
+1

for https://github.com/cms-sw/cmssw/pull/30106	723af99 	

silviodonato:
+1

silviodonato:
merge
cc: @santocch

PR #30113, opened 2020-06-04
Changed legacy modules in L1Trigger/TrackTrigger

Dr15Jones:
#### PR description:

Changed modules to be thread friendly types. These modules are used in workflow 20034.0.

#### PR validation:

The code compiles.


kpedro88:
please test

Dr15Jones:
@kpedro88 we should probably make an issue that this code should be looked at in more detail. E.g. the module is a template but it appears only 1 type can work with it.

kpedro88:
thanks @Dr15Jones 
FYI: @mdhildreth @skinnari @delaere

kpedro88:
+upgrade

silviodonato:
merge
cc: @rekovic

Issue #30112, opened 2020-06-04
Invalid pointers in reco::GenJets in MINIAOD


rappoccio:
There is a problem with invalid pointers somewhere in the MINIAOD code. This was at least present in 10.6.x. It's a very rare corner case (1 event in 10M or so). To reproduce you can execute this `pyfwlite` file. You can see that there is an invalid pointer where it says "not valid" in the following code. 

@ahinzmann @laurenhay @alefisico 

```
import sys
oldargv = sys.argv[:]
sys.argv = [ '-b-' ]
import ROOT
ROOT.gROOT.SetBatch(True)
sys.argv = oldargv

# load FWLite C++ libraries
ROOT.gSystem.Load("libFWCoreFWLite.so");
ROOT.gSystem.Load("libDataFormatsFWLite.so");
ROOT.FWLiteEnabler.enable()


# load FWlite python libraries
from DataFormats.FWLite import Handle, Events

genjets, genjetLabel = Handle("std::vector<reco::GenJet>"), "slimmedGenJets"
genjetsAK8, genjetAK8Label = Handle("std::vector<reco::GenJet>"), "slimmedGenJetsAK8"
gencands, gencandsLabel = Handle("std::vector<pat::PackedGenParticle>"), "packedGenParticles"

# open file (you can use 'edmFileUtil -d /store/whatever.root' to get the physical file name)
#events = Events("step3_inMINIAODSIM.root")
#events = Events("root://eoscms.cern.ch//eos/cms/store/mc/RunIIFall17MiniAODv2/TTJets_SingleLeptFromT_TuneCP5_13TeV-madgraphMLM-pythia8/MINIAODSIM/PU2017_12Apr2018_94X_mc2017_rea
listic_v14-v1/910000/D842F996-2149-E811-AF09-0242AC1C0502.root")

events = Events("file:PhysicsTools/NanoAODJMAR/test/nano106X_on_mini106X_2017_mc_NANO.root")


for iev,event in enumerate(events):
    if iev != 13109:
        continue


    #if iev >= 10: break
    event.getByLabel(genjetLabel, genjets)
    event.getByLabel(genjetAK8Label, genjetsAK8)
    event.getByLabel(gencandsLabel, gencands)

    print ("\nEvent %d: run %6d, lumi %4d, event %12d" % (iev,event.eventAuxiliary().run(), event.eventAuxiliary().luminosityBlock(),event.eventAuxiliary().event()))


[...]

PR #30114, opened 2020-06-04
Adding protection against garbage pointers

rappoccio:
#### PR description:

This adds a trivial protection against the issue reported [here](https://github.com/cms-sw/cmssw/issues/30112). 

#### PR validation:

There is a test script in the issue report [here](https://github.com/cms-sw/cmssw/issues/30112)

This not a backport, but the original. 

Also tagging @laurenhay @ahinzmann @alefisico

### DIFF
--- a/PhysicsTools/NanoAOD/interface/SimpleFlatTableProducer.h
+++ b/PhysicsTools/NanoAOD/interface/SimpleFlatTableProducer.h
@@ -163,7 +163,7 @@ public:
     } else {
       for (unsigned int i = 0, n = prod->size(); i < n; ++i) {
         const auto &obj = (*prod)[i];
-        if (cut_(obj)) {
+        if (prod->ptrAt(i).isNonnull() && prod->ptrAt(i).isAvailable() && cut_(obj)) {
           selobjs.push_back(&obj);
           if (!extvars_.empty())
             selptrs.emplace_back(prod->ptrAt(i));


silviodonato:
please test

PhysicsTools/NanoAOD/interface/SimpleFlatTableProducer.h
```diff
@@ -163,7 +163,7 @@ class SimpleFlatTableProducer : public SimpleFlatTableProducerBase<T, edm::View<
     } else {
       for (unsigned int i = 0, n = prod->size(); i < n; ++i) {
         const auto &obj = (*prod)[i];
-        if (cut_(obj)) {
+        if (prod->ptrAt(i).isNonnull() && prod->ptrAt(i).isAvailable() && cut_(obj)) {
```
gpetruc, line 166: I would remove the isAvailable, otherwise this would cause NanoAOD to fail silently if it's misconfigured to read from a collection that has been dropped (e.g. if one skims some collection by reference, then drops the original, and tries to make a table from the skim)

rappoccio: I tried that, but it doesn't catch the dangling pointers because they aren't null (they are garbage).

gpetruc: Pointers shouldn't be garbage in CMSSW, they should either be null or invalid because they point to a dropped collection => I'm assuming it's the latter case.
I don't understand why this fix is related to that:  in #30112 the problem is daughters of a GenJet (and it's a feature, not a bug), here prod is a collection read from the event, so one level less deep.
If in your workflow you're first taking the daughters of the jet and putting them in a collection and then running the flat table on top of it, then IMO you should put the fix in the piece of code that is dong the first task, not here.

rappoccio: OK, that's also fine. We can adjust the `UniqueMerger` class instead. We can close this PR and start a new one with the updates.

rappoccio:
Per review we will close this PR and implement a check in the `UniqueMerger` class.

PR #30115, opened 2020-06-04
Update valid condition for GEM pads and clusters

dildick:
#### PR description:

Update valid condition for GEM pads and clusters according to [DN-20-016, page 28](https://gitlab.cern.ch/tdr/notes/DN-20-016/blob/master/temp/DN-20-016_temp.pdf). In short: ME0 pad 255, GE1/1 pad 255 and GE2/1 pad 511 are considered invalid. This PR also fixes a minor bug in the GEM-CSC trigger that excluded GE1/1 or GE2/1 zero pads.

#### PR validation:

Tested with WF 27434.0

N/A

@jshlee @hyunyong

### DIFF
--- a/DataFormats/GEMDigi/src/GEMPadDigi.cc
+++ b/DataFormats/GEMDigi/src/GEMPadDigi.cc
@@ -1,12 +1,15 @@
 #include "DataFormats/GEMDigi/interface/GEMPadDigi.h"
 #include <iostream>
 
-GEMPadDigi::GEMPadDigi(int pad, int bx) : pad_(pad), bx_(bx) {}
+GEMPadDigi::GEMPadDigi(uint16_t pad, int16_t bx, enum GEMSubDetId::Station station)
+    : pad_(pad), bx_(bx), station_(station) {}
 
-GEMPadDigi::GEMPadDigi() : pad_(65535), bx_(-99) {}
+GEMPadDigi::GEMPadDigi() : pad_(GE11InValid), bx_(-99), station_(GEMSubDetId::Station::GE11) {}
 
 // Comparison
-bool GEMPadDigi::operator==(const GEMPadDigi& digi) const { return pad_ == digi.pad() and bx_ == digi.bx(); }
+bool GEMPadDigi::operator==(const GEMPadDigi& digi) const {
+  return pad_ == digi.pad() and bx_ == digi.bx() and station_ == digi.station();
+}
 
 // Comparison
 bool GEMPadDigi::operator!=(const GEMPadDigi& digi) const { return pad_ != digi.pad() or bx_ != digi.bx(); }
@@ -19,7 +22,15 @@ bool GEMPadDigi::operator<(const GEMPadDigi& digi) const {
     return digi.bx() < bx_;
 }
 
-bool GEMPadDigi::isValid() const { return bx_ != -99 and pad_ != 65535; }
+bool GEMPadDigi::isValid() const {
+  uint16_t invalid = GE11InValid;
+  if (station_ == GEMSubDetId::Station::ME0) {
+    invalid = ME0InValid;
+  } else if (station_ == GEMSubDetId::Station::GE21) {
+    invalid = GE21InValid;
+  }
+  return pad_ != invalid;
+}
 
 std::ostream& operator<<(std::ostream& o, const GEMPadDigi& digi) {
   return o << " GEMPadDigi Pad = " << digi.pad() << " bx = " << digi.bx();
--- a/DataFormats/GEMDigi/src/GEMPadDigiCluster.cc
+++ b/DataFormats/GEMDigi/src/GEMPadDigiCluster.cc
@@ -1,13 +1,14 @@
 #include "DataFormats/GEMDigi/interface/GEMPadDigiCluster.h"
 #include <iostream>
 
-GEMPadDigiCluster::GEMPadDigiCluster(std::vector<uint16_t> pads, int bx) : v_(pads), bx_(bx) {}
+GEMPadDigiCluster::GEMPadDigiCluster(std::vector<uint16_t> pads, int16_t bx, enum GEMSubDetId::Station station)
+    : v_(pads), bx_(bx), station_(station) {}
 
-GEMPadDigiCluster::GEMPadDigiCluster() : v_(std::vector<uint16_t>()), bx_(-99) {}
+GEMPadDigiCluster::GEMPadDigiCluster() : v_(std::vector<uint16_t>()), bx_(-99), station_(GEMSubDetId::Station::GE11) {}
 
 // Comparison
 bool GEMPadDigiCluster::operator==(const GEMPadDigiCluster& digi) const {
-  return v_ == digi.pads() and bx_ == digi.bx();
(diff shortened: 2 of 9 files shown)


DataFormats/GEMDigi/src/GEMPadDigi.cc
```diff
@@ -1,9 +1,16 @@
 #include "DataFormats/GEMDigi/interface/GEMPadDigi.h"
 #include <iostream>
 
-GEMPadDigi::GEMPadDigi(int pad, int bx) : pad_(pad), bx_(bx) {}
+GEMPadDigi::GEMPadDigi(uint16_t pad, int16_t bx, int station) : pad_(pad), bx_(bx) {
```
kpedro88, line 4: can this just be initialized with a Station instead of an int?

DataFormats/GEMDigi/src/GEMPadDigi.cc
```diff
@@ -19,7 +26,14 @@ bool GEMPadDigi::operator<(const GEMPadDigi& digi) const {
     return digi.bx() < bx_;
 }
 
-bool GEMPadDigi::isValid() const { return bx_ != -99 and pad_ != 65535; }
+bool GEMPadDigi::isValid() const {
+  // ME0 and GE11 have currently the same invalid pad
```
kpedro88, line 30: since there are separate enums for them, maybe better to check explicitly (in case it changes in the future)

DataFormats/GEMDigi/src/GEMPadDigiCluster.cc
```diff
@@ -1,9 +1,14 @@
 #include "DataFormats/GEMDigi/interface/GEMPadDigiCluster.h"
 #include <iostream>
 
-GEMPadDigiCluster::GEMPadDigiCluster(std::vector<uint16_t> pads, int bx) : v_(pads), bx_(bx) {}
+GEMPadDigiCluster::GEMPadDigiCluster(std::vector<uint16_t> pads, int16_t bx, int station) : v_(pads), bx_(bx) {
```
kpedro88, line 4: initialize directly with Station?

kpedro88:
please test

dildick:
I dropped the enum, since the GEM station number is not particularly magic (0, 1 or 2). Class members, the constructors and the isValid function were updated in GEMPadDigi and GEMPadDigiCluster accordingly.

civanch:
please test

civanch:
+1

dildick:
@kpedro88 Let me know if this solution is OK. @rekovic Do you have any comments?

> I dropped the enum, since the GEM station number is not particularly magic (0, 1 or 2). Class members, the constructors and the isValid function were updated in GEMPadDigi and GEMPadDigiCluster accordingly.

dildick:
Hi Kevin, I understand your concern. My impression is that the station number mapping is trivial. See class headers below. If it is confusing, I'll make an enum.

```
class GEMPadDigi {
  /*
    station number:
    0 -> ME0
    1 -> GE1/1
    2 -> GE2/1
   */
};

class GEMPadDigiCluster {
  /*
    station number:
    1 -> GE1/1
    2 -> GE2/1
   */
};
```

kpedro88:
It's trivial in some sense, but five years from now, it may not be clear if one comparison to the number 0 is the same as another comparison to the number 0.

L1Trigger/L1TGEM/plugins/GEMPadDigiProducer.cc
```diff
@@ -96,7 +104,7 @@ void GEMPadDigiProducer::buildPads(const GEMDigiCollection& det_digis, GEMPadDig
 
     // fill the output collections
     for (const auto& d : proto_pads) {
-      GEMPadDigi pad_digi(d.first, d.second);
+      GEMPadDigi pad_digi(d.first, d.second, enumStation(p->id().station()));
```
kpedro88, line 107: as a future development item, it could be nice to use the station enum everywhere, including in the DetId

kpedro88:
please test

L1Trigger/L1TGEM/plugins/GEMPadDigiProducer.cc
dildick: Do I understand correctly that you suggest to move the `Station` class enum to `GEMDetId`? I suppose it wouldn't be a big deal to include `GEMDetId` in the `GEMPadDigi` and similar objects.

kpedro88: alternatively, a separate header like https://github.com/cms-sw/cmssw/blob/master/DataFormats/MuonDetId/interface/MuonSubdetId.h could be introduced

dildick: E.g. something like this?

```
#ifndef DataFormats_MuonDetId_GEMSubDetId_h
#define DataFormats_MuonDetId_GEMSubDetId_h

/** \class GEMSubDetId
 *
 */

class  GEMSubDetId {
public:
   enum class Station { ME0 = 0, GE11 = 1, GE21 = 2 };
   static Station station(uint16_t st) const { 
     Station returnValue = Station::GE11;
     if (st == 0) {
       returnValue = Station::ME0;
     }
     else if (st == 2) {
       returnValue = Station::GE21;
     }
     return returnValue;
    };
};

#endif
```

kpedro88: I was just thinking of a short header file to host the enum class definition, but having the converter function there as well could be a nice touch

dildick: Cool. Once the results from the tests are in I can make the change in this PR to clean it up nice.

dildick: I have proposal here: https://github.com/cms-sw/cmssw/compare/master...dildick:CMSSW_11_2_X-GEM-valid-clusters?expand=1, but this is causing memory issues in WF 24834.0. I'm looking into it

```
----- Begin Fatal Exception 17-Jun-2020 15:40:36 CDT-----------------------
An exception of category 'BadAlloc' occurred while
   [0] Processing  Event run: 1 lumi: 1 event: 2 stream: 0
   [1] Running path 'HLTAnalyzerEndpath'
   [2] Prefetching for module L1TRawToDigi/'hltGtStage2Digis'
   [3] Prefetching for module RawDataCollectorByLabel/'rawDataCollector'
   [4] Prefetching for module CSCDigiToRawModule/'cscpacker'
   [5] Calling method for module CSCTriggerPrimitivesProducer/'simCscTriggerPrimitiveDigis'
Exception Message:
A std::bad_alloc exception was thrown.
The job has probably exhausted the virtual memory available to the process.
```

dildick:
Alright, this should be good enough. I tested it with WF 20434.0, which passed.

kpedro88:
please test workflow 23234.1001,23434.1001

civanch:
+1

rekovic:
+1

kpedro88:
+upgrade

silviodonato:
+1

PR #30117, opened 2020-06-04
Add new ProcessBlock feature to the Framework

wddgit:
This introduces the new transitions beginProcessBlock, endProcessBlock, beginProcessBlockProduce, and endProcessBlockProduce. These can be used by producers, filters, and analyzers in a way that is analogous to beginRun, endRun, beginRunProduce, and endRunProduce. These abilities become available when the types ProcessBlockCache, BeginProcessBlockProducer, or EndProcessBlockProducer are passed to the base class of the module as template parameters. These abilities are not available to legacy module types.

The order modules are executed respects product dependencies established with calls to consumes and produces. One module can put products into a ProcessBlock and another module get them.

beginProcessBlock occurs after beginJob and after beginStream, but before any run, lumi or event transitions. endProcessBlock occurs after all run, lumi, and event transitions, but before endStream and endJob. Each occurs only once per job, except in the very unusual case where all output files are closed and new output files opened. In that case only, there will be a pair of end and begin transitions in between the output file close and open.

This is only the first of a series of PRs implementing this feature. Future PRs will add persistency and the ability to access this information from Events, Runs, and Lumis through a cache. With this PR, these abilities cannot be used by OutputModules. That will be added later. (In this PR, the OutputModule part of this is partially implemented. In the code from the EventProcessor to the functions in the OutputModuleBase classes, the implementation is complete. The functions in OutputModuleBase in this PR do nothing. The PoolOutputModule only has the minimal changes required to prevent this PR from breaking it. The next PR will complete the implementation for output.)

Similarly, the implementation is not done in the InputSource. The future transition accessInputProcessBlock is also only partially implemented. (The part in the InputSource is not implemented although the part of the code from EventProcessor down to the modules is complete. Those functions will never get called with the code in this PR.)

There should be no changes in output. After this PR, these new abilities are only used by modules in Core tests. Nothing extra should be running in production jobs and nothing should be modified. You can see the new transitions in the output from the Tracer. There are new signals ActivityRegistry. The MessageLogger is modified to properly handle these new contexts. I expect any performance impact to be negligible until modules start doing nontrivial work in these transitions.

In comments to this PR requesting changes, it would be useful to identify issues that can be handled in future PRs. This PR is already large.

#### PR validation:

There are several new Framework unit tests and extensions to existing Framework unit tests to cover the new abilities. Output from tests in runTheMatrix and relVals should be identical.

I expect there will be conflicts between this PR and #29553 in at least EDConsumerBase. They should not be difficult to resolve though.

### DIFF
--- a/DataFormats/Provenance/src/BranchType.cc
+++ b/DataFormats/Provenance/src/BranchType.cc
@@ -1,5 +1,7 @@
 #include "DataFormats/Provenance/interface/BranchType.h"
 #include "FWCore/Utilities/interface/EDMException.h"
+#include <array>
+#include <cstddef>
 #include <iostream>
 
 namespace edm {
@@ -17,44 +19,40 @@ namespace edm {
     std::string const branchEntryInfo = "BranchEntryInfo";  // backward compatibility
     std::string const productProvenance = "ProductProvenance";
 
-    // Prefixes
-    std::string const process = "ProcessBlock";
-    std::string const run = "Run";
-    std::string const lumi = "LuminosityBlock";
-    std::string const event = "Event";
+    std::array<std::string, NumBranchTypes> const branchTypeNames{{"Event", "LuminosityBlock", "Run", "ProcessBlock"}};
+    std::size_t constexpr eventLumiRunSize = 3;
+    using NameArray = const std::array<std::string, eventLumiRunSize>;
 
-    // Trees, branches, indices
-    std::string const runs = run + 's';
-    std::string const lumis = lumi + 's';
-    std::string const events = event + 's';
+    NameArray treeNames{{branchTypeNames[InEvent] + 's', branchTypeNames[InLumi] + 's', branchTypeNames[InRun] + 's'}};
 
-    std::string const runMeta = run + metaData;
-    std::string const lumiMeta = lumi + metaData;
-    std::string const eventMeta = event + metaData;
+    NameArray metaTreeNames{
+        {branchTypeNames[InEvent] + metaData, branchTypeNames[InLumi] + metaData, branchTypeNames[InRun] + metaData}};
 
-    std::string const runInfo = run + "StatusInformation";      // backward compatibility
-    std::string const lumiInfo = lumi + "StatusInformation";    // backward compatibility
-    std::string const eventInfo = event + "StatusInformation";  // backward compatibility
+    // backward compatibility
+    NameArray infoNames{{branchTypeNames[InEvent] + "StatusInformation",
+                         branchTypeNames[InLumi] + "StatusInformation",
+                         branchTypeNames[InRun] + "StatusInformation"}};
 
-    std::string const runAuxiliary = run + auxiliary;
-    std::string const lumiAuxiliary = lumi + auxiliary;
-    std::string const eventAuxiliary = event + auxiliary;
+    NameArray auxiliaryNames{{branchTypeNames[InEvent] + auxiliary,
+                              branchTypeNames[InLumi] + auxiliary,
+                              branchTypeNames[InRun] + auxiliary}};
 
-    std::string const runProductStatus = run + productStatus;      // backward compatibility
-    std::string const lumiProductStatus = lumi + productStatus;    // backward compatibility
-    std::string const eventProductStatus = event + productStatus;  // backward compatibility
+    // backward compatibility
+    NameArray productStatusNames{{branchTypeNames[InEvent] + productStatus,
+                                  branchTypeNames[InLumi] + productStatus,
+                                  branchTypeNames[InRun] + productStatus}};
 
-    std::string const runEventEntryInfo = run + branchEntryInfo;      // backward compatibility
-    std::string const lumiEventEntryInfo = lumi + branchEntryInfo;    // backward compatibility
-    std::string const eventEventEntryInfo = event + branchEntryInfo;  // backward compatibility
+    // backward compatibility
+    NameArray eventEntryInfoNames{{branchTypeNames[InEvent] + branchEntryInfo,
+                                   branchTypeNames[InLumi] + branchEntryInfo,
+                                   branchTypeNames[InRun] + branchEntryInfo}};
 
-    std::string const runProductProvenance = run + productProvenance;
-    std::string const lumiProductProvenance = lumi + productProvenance;
-    std::string const eventProductProvenance = event + productProvenance;
+    NameArray productProvenanceNames{{branchTypeNames[InEvent] + productProvenance,
+                                      branchTypeNames[InLumi] + productProvenance,
+                                      branchTypeNames[InRun] + productProvenance}};
 
-    std::string const runAux = run + aux;
-    std::string const lumiAux = lumi + aux;
-    std::string const eventAux = event + aux;
+    // backward compatibility
+    NameArray auxNames{{branchTypeNames[InEvent] + aux, branchTypeNames[InLumi] + aux, branchTypeNames[InRun] + aux}};
 
     std::string const entryDescriptionTree = "EntryDescription";
     std::string const entryDescriptionIDBranch = "Hash";
@@ -88,17 +86,14 @@ namespace edm {
     std::string const idToParameterSetBlobsBranch = "IdToParameterSetsBlobs";
   }  // namespace
 
-  std::string const& BranchTypeToString(BranchType const& branchType) {
-    return ((branchType == InEvent) ? event
-                                    : ((branchType == InRun) ? run : ((branchType == InLumi) ? lumi : process)));
-  }
+  std::string const& BranchTypeToString(BranchType const& branchType) { return branchTypeNames[branchType]; }
 
   std::string const& BranchTypeToProductTreeName(BranchType const& branchType) {
     if (branchType >= InProcess) {
       throw Exception(errors::LogicError,
                       "branchType argument out of legal range for the function BranchTypeToProductTreeName");
     }
-    return ((branchType == InEvent) ? events : ((branchType == InRun) ? runs : lumis));
+    return treeNames[branchType];
   }
 
   std::string const& BranchTypeToMetaDataTreeName(BranchType const& branchType) {
@@ -106,51 +101,51 @@ namespace edm {
       throw Exception(errors::LogicError,
                       "branchType argument out of legal range for the function BranchTypeToMetaDataTreeName");
     }
-    return ((branchType == InEvent) ? eventMeta : ((branchType == InRun) ? runMeta : lumiMeta));
+    return metaTreeNames[branchType];
   }
 
-  std::string const& BranchTypeToInfoTreeName(BranchType const& branchType) {  // backward compatibility
+  // backward compatibility
+  std::string const& BranchTypeToInfoTreeName(BranchType const& branchType) {
     if (branchType >= InProcess) {
       throw Exception(errors::LogicError,
                       "branchType argument out of legal range for the function BranchTypeToInfoTreeName");
     }
-    return ((branchType == InEvent) ? eventInfo
-                                    : ((branchType == InRun) ? runInfo : lumiInfo));  // backward compatibility
-  }                                                                                   // backward compatibility
+    return infoNames[branchType];
+  }
 
   std::string const& BranchTypeToAuxiliaryBranchName(BranchType const& branchType) {
     if (branchType >= InProcess) {
       throw Exception(errors::LogicError,
                       "branchType argument out of legal range for the function BranchTypeToAuxiliaryBranchName");
     }
-    return ((branchType == InEvent) ? eventAuxiliary : ((branchType == InRun) ? runAuxiliary : lumiAuxiliary));
+    return auxiliaryNames[branchType];
   }
 
-  std::string const& BranchTypeToAuxBranchName(BranchType const& branchType) {  // backward compatibility
+  // backward compatibility
+  std::string const& BranchTypeToAuxBranchName(BranchType const& branchType) {
     if (branchType >= InProcess) {
       throw Exception(errors::LogicError,
                       "branchType argument out of legal range for the function BranchTypeToAuxBranchName");
     }
-    return ((branchType == InEvent) ? eventAux : ((branchType == InRun) ? runAux : lumiAux));  // backward compatibility
-  }                                                                                            // backward compatibility
+    return auxNames[branchType];
+  }
 
-  std::string const& BranchTypeToProductStatusBranchName(BranchType const& branchType) {  // backward compatibility
+  // backward compatibility
+  std::string const& BranchTypeToProductStatusBranchName(BranchType const& branchType) {
     if (branchType >= InProcess) {
       throw Exception(errors::LogicError,
                       "branchType argument out of legal range for the function BranchTypeToProductStatusBranchName");
     }
-    return ((branchType == InEvent)
-                ? eventProductStatus
-                : ((branchType == InRun) ? runProductStatus : lumiProductStatus));  // backward compatibility
-  }                                                                                 // backward compatibility
+    return productStatusNames[branchType];
+  }
 
+  // backward compatibility
   std::string const& BranchTypeToBranchEntryInfoBranchName(BranchType const& branchType) {
     if (branchType >= InProcess) {
       throw Exception(errors::LogicError,
                       "branchType argument out of legal range for the function BranchTypeToBranchEntryInfoBranchName");
     }
-    return ((branchType == InEvent) ? eventEventEntryInfo
-                                    : ((branchType == InRun) ? runEventEntryInfo : lumiEventEntryInfo));
+    return eventEntryInfoNames[branchType];
   }
 
   std::string const& BranchTypeToProductProvenanceBranchName(BranchType const& branchType) {
@@ -159,8 +154,7 @@ namespace edm {
           errors::LogicError,
           "branchType argument out of legal range for the function BranchTypeToProductProvenanceBranchName");
     }
-    return ((branchType == InEvent) ? eventProductProvenance
-                                    : ((branchType == InRun) ? runProductProvenance : lumiProductProvenance));
+    return productProvenanceNames[branchType];
   }
 
   namespace poolNames {
@@ -234,12 +228,12 @@ namespace edm {
     // Branch on ParameterSets Tree
     std::string const& idToParameterSetBlobsBranchName() { return idToParameterSetBlobsBranch; }
 
-    std::string const& eventTreeName() { return events; }
+    std::string const& eventTreeName() { return treeNames[InEvent]; }
 
-    std::string const& eventMetaDataTreeName() { return eventMeta; }
+    std::string const& eventMetaDataTreeName() { return metaTreeNames[InEvent]; }
 
     std::string const& eventHistoryTreeName() { return eventHistory; }
-    std::string const& luminosityBlockTreeName() { return lumis; }
-    std::string const& runTreeName() { return runs; }
+    std::string const& luminosityBlockTreeName() { return treeNames[InLumi]; }
+    std::string const& runTreeName() { return treeNames[InRun]; }
   }  // namespace poolNames
 }  // namespace edm
--- a/DataFormats/Provenance/src/BranchType.cc
+++ b/DataFormats/Provenance/src/BranchType.cc
@@ -1,4 +1,5 @@
 #include "DataFormats/Provenance/interface/BranchType.h"
+#include "FWCore/Utilities/interface/EDMException.h"
 #include <iostream>
 
 namespace edm {
@@ -17,6 +18,7 @@ namespace edm {
     std::string const productProvenance = "ProductProvenance";
 
     // Prefixes
+    std::string const process = "ProcessBlock";
     std::string const run = "Run";
     std::string const lumi = "LuminosityBlock";
     std::string const event = "Event";
@@ -49,15 +51,6 @@ namespace edm {
     std::string const runProductProvenance = run + productProvenance;
     std::string const lumiProductProvenance = lumi + productProvenance;
     std::string const eventProductProvenance = event + productProvenance;
-
-    std::string const majorIndex = ".id_.run_";
-    std::string const runMajorIndex = runAuxiliary + majorIndex;
-    std::string const lumiMajorIndex = lumiAuxiliary + majorIndex;
-    std::string const eventMajorIndex = eventAuxiliary + majorIndex;
-
-    std::string const runMinorIndex;  // empty
-    std::string const lumiMinorIndex = lumiAuxiliary + ".id_.luminosityBlock_";
-    std::string const eventMinorIndex = eventAuxiliary + ".id_.event_";
 
     std::string const runAux = run + aux;
     std::string const lumiAux = lumi + aux;
@@ -96,52 +89,78 @@ namespace edm {
   }  // namespace
 
   std::string const& BranchTypeToString(BranchType const& branchType) {
-    return ((branchType == InEvent) ? event : ((branchType == InRun) ? run : lumi));
+    return ((branchType == InEvent) ? event
+                                    : ((branchType == InRun) ? run : ((branchType == InLumi) ? lumi : process)));
   }
 
   std::string const& BranchTypeToProductTreeName(BranchType const& branchType) {
+    if (branchType >= InProcess) {
+      throw Exception(errors::LogicError,
+                      "branchType argument out of legal range for the function BranchTypeToProductTreeName");
+    }
     return ((branchType == InEvent) ? events : ((branchType == InRun) ? runs : lumis));
   }
 
   std::string const& BranchTypeToMetaDataTreeName(BranchType const& branchType) {
+    if (branchType >= InProcess) {
+      throw Exception(errors::LogicError,
+                      "branchType argument out of legal range for the function BranchTypeToMetaDataTreeName");
+    }
     return ((branchType == InEvent) ? eventMeta : ((branchType == InRun) ? runMeta : lumiMeta));
   }
 
   std::string const& BranchTypeToInfoTreeName(BranchType const& branchType) {  // backward compatibility
+    if (branchType >= InProcess) {
+      throw Exception(errors::LogicError,
+                      "branchType argument out of legal range for the function BranchTypeToInfoTreeName");
+    }
     return ((branchType == InEvent) ? eventInfo
                                     : ((branchType == InRun) ? runInfo : lumiInfo));  // backward compatibility
   }                                                                                   // backward compatibility
 
   std::string const& BranchTypeToAuxiliaryBranchName(BranchType const& branchType) {
+    if (branchType >= InProcess) {
+      throw Exception(errors::LogicError,
+                      "branchType argument out of legal range for the function BranchTypeToAuxiliaryBranchName");
+    }
     return ((branchType == InEvent) ? eventAuxiliary : ((branchType == InRun) ? runAuxiliary : lumiAuxiliary));
   }
 
-  std::string const& BranchTypeToAuxBranchName(BranchType const& branchType) {                 // backward compatibility
+  std::string const& BranchTypeToAuxBranchName(BranchType const& branchType) {  // backward compatibility
+    if (branchType >= InProcess) {
+      throw Exception(errors::LogicError,
+                      "branchType argument out of legal range for the function BranchTypeToAuxBranchName");
+    }
     return ((branchType == InEvent) ? eventAux : ((branchType == InRun) ? runAux : lumiAux));  // backward compatibility
   }                                                                                            // backward compatibility
 
   std::string const& BranchTypeToProductStatusBranchName(BranchType const& branchType) {  // backward compatibility
+    if (branchType >= InProcess) {
+      throw Exception(errors::LogicError,
+                      "branchType argument out of legal range for the function BranchTypeToProductStatusBranchName");
+    }
     return ((branchType == InEvent)
                 ? eventProductStatus
                 : ((branchType == InRun) ? runProductStatus : lumiProductStatus));  // backward compatibility
   }                                                                                 // backward compatibility
 
   std::string const& BranchTypeToBranchEntryInfoBranchName(BranchType const& branchType) {
+    if (branchType >= InProcess) {
+      throw Exception(errors::LogicError,
+                      "branchType argument out of legal range for the function BranchTypeToBranchEntryInfoBranchName");
+    }
     return ((branchType == InEvent) ? eventEventEntryInfo
                                     : ((branchType == InRun) ? runEventEntryInfo : lumiEventEntryInfo));
   }
 
   std::string const& BranchTypeToProductProvenanceBranchName(BranchType const& branchType) {
+    if (branchType >= InProcess) {
+      throw Exception(
+          errors::LogicError,
+          "branchType argument out of legal range for the function BranchTypeToProductProvenanceBranchName");
+    }
     return ((branchType == InEvent) ? eventProductProvenance
                                     : ((branchType == InRun) ? runProductProvenance : lumiProductProvenance));
-  }
-
-  std::string const& BranchTypeToMajorIndexName(BranchType const& branchType) {
-    return ((branchType == InEvent) ? eventMajorIndex : ((branchType == InRun) ? runMajorIndex : lumiMajorIndex));
-  }
-
-  std::string const& BranchTypeToMinorIndexName(BranchType const& branchType) {
-    return ((branchType == InEvent) ? eventMinorIndex : ((branchType == InRun) ? runMinorIndex : lumiMinorIndex));
   }
 
   namespace poolNames {
--- a/DataFormats/Provenance/src/BranchType.cc
+++ b/DataFormats/Provenance/src/BranchType.cc
@@ -1,6 +1,7 @@
 #include "DataFormats/Provenance/interface/BranchType.h"
-#include "FWCore/Utilities/interface/EDMException.h"
+
 #include <array>
+#include <cassert>
 #include <cstddef>
 #include <iostream>
 
@@ -89,71 +90,46 @@ namespace edm {
   std::string const& BranchTypeToString(BranchType const& branchType) { return branchTypeNames[branchType]; }
 
   std::string const& BranchTypeToProductTreeName(BranchType const& branchType) {
-    if (branchType >= InProcess) {
-      throw Exception(errors::LogicError,
-                      "branchType argument out of legal range for the function BranchTypeToProductTreeName");
-    }
+    assert(branchType < eventLumiRunSize);
     return treeNames[branchType];
   }
 
   std::string const& BranchTypeToMetaDataTreeName(BranchType const& branchType) {
-    if (branchType >= InProcess) {
-      throw Exception(errors::LogicError,
-                      "branchType argument out of legal range for the function BranchTypeToMetaDataTreeName");
-    }
+    assert(branchType < eventLumiRunSize);
     return metaTreeNames[branchType];
   }
 
   // backward compatibility
   std::string const& BranchTypeToInfoTreeName(BranchType const& branchType) {
-    if (branchType >= InProcess) {
-      throw Exception(errors::LogicError,
-                      "branchType argument out of legal range for the function BranchTypeToInfoTreeName");
-    }
+    assert(branchType < eventLumiRunSize);
     return infoNames[branchType];
   }
 
   std::string const& BranchTypeToAuxiliaryBranchName(BranchType const& branchType) {
-    if (branchType >= InProcess) {
-      throw Exception(errors::LogicError,
-                      "branchType argument out of legal range for the function BranchTypeToAuxiliaryBranchName");
-    }
+    assert(branchType < eventLumiRunSize);
     return auxiliaryNames[branchType];
   }
 
   // backward compatibility
   std::string const& BranchTypeToAuxBranchName(BranchType const& branchType) {
-    if (branchType >= InProcess) {
-      throw Exception(errors::LogicError,
-                      "branchType argument out of legal range for the function BranchTypeToAuxBranchName");
-    }
+    assert(branchType < eventLumiRunSize);
     return auxNames[branchType];
   }
 
   // backward compatibility
   std::string const& BranchTypeToProductStatusBranchName(BranchType const& branchType) {
-    if (branchType >= InProcess) {
-      throw Exception(errors::LogicError,
-                      "branchType argument out of legal range for the function BranchTypeToProductStatusBranchName");
-    }
+    assert(branchType < eventLumiRunSize);
     return productStatusNames[branchType];
   }
 
   // backward compatibility
   std::string const& BranchTypeToBranchEntryInfoBranchName(BranchType const& branchType) {
-    if (branchType >= InProcess) {
-      throw Exception(errors::LogicError,
-                      "branchType argument out of legal range for the function BranchTypeToBranchEntryInfoBranchName");
-    }
+    assert(branchType < eventLumiRunSize);
     return eventEntryInfoNames[branchType];
   }
 
   std::string const& BranchTypeToProductProvenanceBranchName(BranchType const& branchType) {
-    if (branchType >= InProcess) {
-      throw Exception(
-          errors::LogicError,
-          "branchType argument out of legal range for the function BranchTypeToProductProvenanceBranchName");
-    }
+    assert(branchType < eventLumiRunSize);
     return productProvenanceNames[branchType];
   }
 
--- a/DataFormats/Provenance/src/BranchType.cc
+++ b/DataFormats/Provenance/src/BranchType.cc
@@ -24,36 +24,33 @@ namespace edm {
     std::size_t constexpr eventLumiRunSize = 3;
     using NameArray = const std::array<std::string, eventLumiRunSize>;
 
-    NameArray treeNames{{branchTypeNames[InEvent] + 's', branchTypeNames[InLumi] + 's', branchTypeNames[InRun] + 's'}};
+    NameArray makeNameArray(std::string const& postfix) {
+      static_assert(InEvent == 0);
+      static_assert(InLumi == 1);
+      static_assert(InRun == 2);
+      return NameArray{
+          {branchTypeNames[InEvent] + postfix, branchTypeNames[InLumi] + postfix, branchTypeNames[InRun] + postfix}};
+    }
 
-    NameArray metaTreeNames{
-        {branchTypeNames[InEvent] + metaData, branchTypeNames[InLumi] + metaData, branchTypeNames[InRun] + metaData}};
+    NameArray treeNames{makeNameArray(std::string("s"))};
 
-    // backward compatibility
-    NameArray infoNames{{branchTypeNames[InEvent] + "StatusInformation",
-                         branchTypeNames[InLumi] + "StatusInformation",
-                         branchTypeNames[InRun] + "StatusInformation"}};
-
-    NameArray auxiliaryNames{{branchTypeNames[InEvent] + auxiliary,
-                              branchTypeNames[InLumi] + auxiliary,
-                              branchTypeNames[InRun] + auxiliary}};
+    NameArray metaTreeNames{makeNameArray(metaData)};
 
     // backward compatibility
-    NameArray productStatusNames{{branchTypeNames[InEvent] + productStatus,
-                                  branchTypeNames[InLumi] + productStatus,
-                                  branchTypeNames[InRun] + productStatus}};
+    NameArray infoNames{makeNameArray(std::string("StatusInformation"))};
+
+    NameArray auxiliaryNames{makeNameArray(auxiliary)};
 
     // backward compatibility
-    NameArray eventEntryInfoNames{{branchTypeNames[InEvent] + branchEntryInfo,
-                                   branchTypeNames[InLumi] + branchEntryInfo,
-                                   branchTypeNames[InRun] + branchEntryInfo}};
-
-    NameArray productProvenanceNames{{branchTypeNames[InEvent] + productProvenance,
-                                      branchTypeNames[InLumi] + productProvenance,
-                                      branchTypeNames[InRun] + productProvenance}};
+    NameArray productStatusNames{makeNameArray(productStatus)};
 
     // backward compatibility
-    NameArray auxNames{{branchTypeNames[InEvent] + aux, branchTypeNames[InLumi] + aux, branchTypeNames[InRun] + aux}};
+    NameArray eventEntryInfoNames{makeNameArray(branchEntryInfo)};
+
+    NameArray productProvenanceNames{makeNameArray(productProvenance)};
+
+    // backward compatibility
+    NameArray auxNames{makeNameArray(aux)};
 
     std::string const entryDescriptionTree = "EntryDescription";
     std::string const entryDescriptionIDBranch = "Hash";
--- a/DataFormats/Provenance/src/BranchType.cc
+++ b/DataFormats/Provenance/src/BranchType.cc
@@ -22,35 +22,39 @@ namespace edm {
 
     std::array<std::string, NumBranchTypes> const branchTypeNames{{"Event", "LuminosityBlock", "Run", "ProcessBlock"}};
     std::size_t constexpr eventLumiRunSize = 3;
-    using NameArray = const std::array<std::string, eventLumiRunSize>;
+    using NameArray = std::array<std::string, eventLumiRunSize>;
 
     NameArray makeNameArray(std::string const& postfix) {
       static_assert(InEvent == 0);
       static_assert(InLumi == 1);
       static_assert(InRun == 2);
-      return NameArray{
-          {branchTypeNames[InEvent] + postfix, branchTypeNames[InLumi] + postfix, branchTypeNames[InRun] + postfix}};
+      static_assert(InProcess == 3);
+      NameArray ret;
+      for (auto i = 0U; i != eventLumiRunSize; ++i) {
+        ret[i] = branchTypeNames[i] + postfix;
+      }
+      return ret;
     }
 
-    NameArray treeNames{makeNameArray(std::string("s"))};
+    const NameArray treeNames{makeNameArray(std::string("s"))};
 
-    NameArray metaTreeNames{makeNameArray(metaData)};
-
-    // backward compatibility
-    NameArray infoNames{makeNameArray(std::string("StatusInformation"))};
-
-    NameArray auxiliaryNames{makeNameArray(auxiliary)};
+    const NameArray metaTreeNames{makeNameArray(metaData)};
 
     // backward compatibility
-    NameArray productStatusNames{makeNameArray(productStatus)};
+    const NameArray infoNames{makeNameArray(std::string("StatusInformation"))};
+
+    const NameArray auxiliaryNames{makeNameArray(auxiliary)};
 
     // backward compatibility
-    NameArray eventEntryInfoNames{makeNameArray(branchEntryInfo)};
-
-    NameArray productProvenanceNames{makeNameArray(productProvenance)};
+    const NameArray productStatusNames{makeNameArray(productStatus)};
 
     // backward compatibility
-    NameArray auxNames{makeNameArray(aux)};
+    const NameArray eventEntryInfoNames{makeNameArray(branchEntryInfo)};
+
+    const NameArray productProvenanceNames{makeNameArray(productProvenance)};
+
+    // backward compatibility
+    const NameArray auxNames{makeNameArray(aux)};
 
     std::string const entryDescriptionTree = "EntryDescription";
     std::string const entryDescriptionIDBranch = "Hash";
--- a/DataFormats/Provenance/src/ProductRegistry.cc
+++ b/DataFormats/Provenance/src/ProductRegistry.cc
@@ -35,15 +35,11 @@ namespace edm {
       : frozen_(false),
         productProduced_(),
         anyProductProduced_(false),
-        eventProductLookup_(new ProductResolverIndexHelper),
-        lumiProductLookup_(new ProductResolverIndexHelper),
-        runProductLookup_(new ProductResolverIndexHelper),
-        processProductLookup_(new ProductResolverIndexHelper),
-        eventNextIndexValue_(0),
-        lumiNextIndexValue_(0),
-        runNextIndexValue_(0),
-        processNextIndexValue_(0),
-
+        productLookups_{{new ProductResolverIndexHelper,
+                         new ProductResolverIndexHelper,
+                         new ProductResolverIndexHelper,
+                         new ProductResolverIndexHelper}},
+        nextIndexValues_(),
         branchIDToIndex_() {
     for (bool& isProduced : productProduced_)
       isProduced = false;
@@ -56,15 +52,10 @@ namespace edm {
     anyProductProduced_ = false;
 
     // propagate_const<T> has no reset() function
-    eventProductLookup_ = std::make_unique<ProductResolverIndexHelper>();
-    lumiProductLookup_ = std::make_unique<ProductResolverIndexHelper>();
-    runProductLookup_ = std::make_unique<ProductResolverIndexHelper>();
-    processProductLookup_ = std::make_unique<ProductResolverIndexHelper>();
-
-    eventNextIndexValue_ = 0;
-    lumiNextIndexValue_ = 0;
-    runNextIndexValue_ = 0;
-    processNextIndexValue_ = 0;
+    for (int i = InEvent; i < NumBranchTypes; ++i) {
+      productLookups_[i] = std::make_unique<ProductResolverIndexHelper>();
+    }
+    nextIndexValues_.fill(0);
 
     branchIDToIndex_.clear();
   }
@@ -151,23 +142,11 @@ namespace edm {
   }
 
   std::shared_ptr<ProductResolverIndexHelper const> ProductRegistry::productLookup(BranchType branchType) const {
-    if (branchType == InEvent)
-      return transient_.eventProductLookup();
-    if (branchType == InLumi)
-      return transient_.lumiProductLookup();
-    if (branchType == InRun)
-      return transient_.runProductLookup();
-    return transient_.processProductLookup();
+    return get_underlying_safe(transient_.productLookups_[branchType]);
   }
 
   std::shared_ptr<ProductResolverIndexHelper> ProductRegistry::productLookup(BranchType branchType) {
-    if (branchType == InEvent)
-      return transient_.eventProductLookup();
-    if (branchType == InLumi)
-      return transient_.lumiProductLookup();
-    if (branchType == InRun)
-      return transient_.runProductLookup();
-    return transient_.processProductLookup();
+    return get_underlying_safe(transient_.productLookups_[branchType]);
   }
 
   void ProductRegistry::setFrozen(bool initializeLookupInfo) {
@@ -464,10 +443,10 @@ namespace edm {
     productLookup(InRun)->setFrozen();
     productLookup(InProcess)->setFrozen();
 
-    transient_.eventNextIndexValue_ = productLookup(InEvent)->nextIndexValue();
-    transient_.lumiNextIndexValue_ = productLookup(InLumi)->nextIndexValue();
-    transient_.runNextIndexValue_ = productLookup(InRun)->nextIndexValue();
-    transient_.processNextIndexValue_ = productLookup(InProcess)->nextIndexValue();
+    for (unsigned int i = InEvent; i < NumBranchTypes; ++i) {
+      BranchType branchType = static_cast<BranchType>(i);
+      transient_.nextIndexValues_[branchType] = productLookup(branchType)->nextIndexValue();
+    }
 
     for (auto const& product : productList_) {
       auto const& desc = product.second;
@@ -586,22 +565,10 @@ namespace edm {
   }
 
   ProductResolverIndex const& ProductRegistry::getNextIndexValue(BranchType branchType) const {
-    if (branchType == InEvent)
-      return transient_.eventNextIndexValue_;
-    if (branchType == InLumi)
-      return transient_.lumiNextIndexValue_;
-    if (branchType == InRun)
-      return transient_.runNextIndexValue_;
-    return transient_.processNextIndexValue_;
+    return transient_.nextIndexValues_[branchType];
   }
 
   ProductResolverIndex& ProductRegistry::nextIndexValue(BranchType branchType) {
-    if (branchType == InEvent)
-      return transient_.eventNextIndexValue_;
-    if (branchType == InLumi)
-      return transient_.lumiNextIndexValue_;
-    if (branchType == InRun)
-      return transient_.runNextIndexValue_;
-    return transient_.processNextIndexValue_;
+    return transient_.nextIndexValues_[branchType];
   }
 }  // namespace edm
--- a/DataFormats/Provenance/src/ProductRegistry.cc
+++ b/DataFormats/Provenance/src/ProductRegistry.cc
@@ -38,9 +38,11 @@ namespace edm {
         eventProductLookup_(new ProductResolverIndexHelper),
         lumiProductLookup_(new ProductResolverIndexHelper),
         runProductLookup_(new ProductResolverIndexHelper),
+        processProductLookup_(new ProductResolverIndexHelper),
         eventNextIndexValue_(0),
         lumiNextIndexValue_(0),
         runNextIndexValue_(0),
+        processNextIndexValue_(0),
 
         branchIDToIndex_() {
     for (bool& isProduced : productProduced_)
@@ -57,10 +59,12 @@ namespace edm {
     eventProductLookup_ = std::make_unique<ProductResolverIndexHelper>();
     lumiProductLookup_ = std::make_unique<ProductResolverIndexHelper>();
     runProductLookup_ = std::make_unique<ProductResolverIndexHelper>();
+    processProductLookup_ = std::make_unique<ProductResolverIndexHelper>();
 
     eventNextIndexValue_ = 0;
     lumiNextIndexValue_ = 0;
     runNextIndexValue_ = 0;
+    processNextIndexValue_ = 0;
 
     branchIDToIndex_.clear();
   }
@@ -151,7 +155,9 @@ namespace edm {
       return transient_.eventProductLookup();
     if (branchType == InLumi)
       return transient_.lumiProductLookup();
-    return transient_.runProductLookup();
+    if (branchType == InRun)
+      return transient_.runProductLookup();
+    return transient_.processProductLookup();
   }
 
   std::shared_ptr<ProductResolverIndexHelper> ProductRegistry::productLookup(BranchType branchType) {
@@ -159,7 +165,9 @@ namespace edm {
       return transient_.eventProductLookup();
     if (branchType == InLumi)
       return transient_.lumiProductLookup();
-    return transient_.runProductLookup();
+    if (branchType == InRun)
+      return transient_.runProductLookup();
+    return transient_.processProductLookup();
   }
 
   void ProductRegistry::setFrozen(bool initializeLookupInfo) {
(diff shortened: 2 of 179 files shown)


wddgit:
please test

fwyzard:
@wddgit could you elaborate on what is the purpose of a ProcessBlock ?

wddgit:
Hi Andrea, Issue #28354 contains the primary discussion that I am aware of that lead to this pull request. The primary purpose is the DQM use case. The feature is general enough that it might be useful for other use cases as well. For example there are some generator products where it might be useful to use this instead of Run products. Possibly there are uses we have not thought of as well. Chris and Matti might want to add to my reply as they requested I start working on this. FYI @schneiml

makortel:
The most immediate motivation comes indeed from DQM (multi-run) harvesting (https://github.com/cms-sw/cmssw/issues/28354), which this PR should be sufficient to solve. The full ProcessBlock feature will be more generic to accommodate other use cases as well. The ProcessBlock` is likely the closest thing to a "file-level storage" that we can get.

wddgit:
The unit test failure is not related to this PR. It looks unrelated from the content of the log. I can see the same error in other PRs. Probably I see the same unit test failure in the IBs as well (there is one in the same package, although I can't see to see any error at all in the IB log file, not sure why)

fwyzard:
@wddgit @makortel thanks for the pointer to #28354 .

So, a "ProcessBlock" spans all the work done within a cmsRun job, or at least all the work done for a set of output files (this last part is not very clear).

It has the usual begin/end transitions - which IIUC are global, not per-stream.

Can it have data attached to it ?
Can such data be produced at the beginning, at the end, or both, of the ProcessBlock ?
Does that mean that one can also save per-ProcessBlock data in an output file ?

wddgit:
Yes. A "ProcessBlock" spans all the work done within a cmsRun job.

There is an exceptional case though. There is a maxSize parameter to OutputModule's. When the output file size exceeds this limit it tells the Framework to look for an opportunity to close all output files of all output modules and open new ones. There are a lot of details related to this that I am not going to try to describe in this comment (ask if you are interested). As far as I know, no one actually uses this capability. Although, maybe someone does that I do not know about. I was just looking through our unit tests and so far I have not even found any Core unit tests of this. Maybe we should add one. Anyway, in this special case the Framework will add additional "ProcessBlock" transitions as if each set of output files was from a different process.

ProcessBlock transitions are all global. There are no stream ProcessBlock transitions.

Yes, data can be attached to it. One can put data products into a ProcessBlock and get them with getByToken, just like for a Run, LuminosityBlock, or Event. There are unit tests in the PR that test this ability.

Data can be produced at the beginning, end,  or both.

This PR does not implement persistence. With only this PR you cannot save per-ProcessBlock to a an output file. But the plan is to implement persistence in a future PR. Then you will be able to save per-ProcessBlock data to an output file. (Although just with the features in this PR, you can do useful things. One alternative to our current plan would be to never implement persistence.)

fwyzard:
Thanks for all the clarifications !

wddgit:
please test

try again. looks like the failing unit test is working in the IBs now. (the failure unrelated to the changes in this PR)

schneiml:
Thanks @wddgit and everybody involved for getting this in!

Regarding the question about persistence: Without persistence, as this PR is today (as I understand), this helps to get rid of the hidden (from EDM) dependencies in HARVESTING, without destroying the ability to do multi-run harvesting. We will need to stick to "legacy" output though, just writing a file outside the control of EDM in an EDAnalyzer (we can transfer the data using products, or at least token products, though).

For the "full thing", we'd need persistence, so we can make the DQMIO output module (`DQMRootOutputModule`) save harvesting output at these new transitions. But that is more effort on our side as well as EDM side, and also on the Computing/Infrastructure side. So it is quite a lot of effort for a simplification with little practical benefits, and we might well choose not to do it.

FWCore/Framework/interface/ProcessBlock.h
```diff
@@ -0,0 +1,232 @@
...
+
+  class ProcessBlock {
+  public:
+    ProcessBlock(ProcessBlockPrincipal const&, ModuleDescription const&, ModuleCallingContext const*, bool isAtEnd);
+
+    template <typename PROD>
+    bool getByToken(EDGetToken token, Handle<PROD>& result) const;
+
+    template <typename PROD>
+    bool getByToken(EDGetTokenT<PROD> token, Handle<PROD>& result) const;
```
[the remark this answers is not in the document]
Dr15Jones: I'm a strong believer in 'learn once, use everywhere' so even if a change might be 'more modern' I don't think the increased learning curve for users is worth the change.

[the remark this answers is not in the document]
makortel: Let's keep the `getByToken()` then.

FWCore/Framework/interface/ProcessBlock.h
```diff
@@ -0,0 +1,232 @@
...
+  template <typename PROD>
+  Handle<PROD> ProcessBlock::getHandle(EDGetTokenT<PROD> token) const {
+    BasicHandle bh = provRecorder_.getByToken_(TypeID(typeid(PROD)), PRODUCT_TYPE, token, moduleCallingContext_);
+    return convert_handle<PROD>(std::move(bh));
+  }
+
+  template <typename PROD>
+  PROD const& ProcessBlock::get(EDGetTokenT<PROD> token) const noexcept(false) {
+    BasicHandle bh = provRecorder_.getByToken_(TypeID(typeid(PROD)), PRODUCT_TYPE, token, moduleCallingContext_);
+    return *convert_handle<PROD>(std::move(bh));
```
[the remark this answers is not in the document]
wddgit: We do not need this code for ProcessBlock. checkIfComplete is related to the automatic product merging that we do for Runs and Lumis. At least with the design I am currently working towards, there will be no merging for ProcessBlock.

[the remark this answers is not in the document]
makortel: Ok, thanks for the clarification.

FWCore/Framework/src/EventProcessor.cc
```diff
@@ -483,6 +484,14 @@ namespace edm {
       principalCache_.insert(std::move(lp));
     }
 
+    {
+      auto pb = std::make_unique<ProcessBlockPrincipal>(preg(), *processConfiguration_);
+      principalCache_.insert(std::move(pb));
+
+      auto pbForInput = std::make_unique<ProcessBlockPrincipal>(preg(), *processConfiguration_);
+      principalCache_.insertForInput(std::move(pb));
```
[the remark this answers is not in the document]
wddgit: A bug. I will push a commit fixing it. This bug does not really affect this PR, but would have broken the persistence planned for the next PR. Thanks for noticing it.

FWCore/Framework/src/SubProcess.cc
```diff
@@ -189,6 +190,14 @@ namespace edm {
       principalCache_.insert(std::move(lbpp));
     }
 
+    {
+      auto pb = std::make_unique<ProcessBlockPrincipal>(preg_, *processConfiguration_, false);
+      principalCache_.insert(std::move(pb));
+
+      auto pbForInput = std::make_unique<ProcessBlockPrincipal>(preg_, *processConfiguration_, false);
+      principalCache_.insertForInput(std::move(pb));
```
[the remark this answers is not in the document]
wddgit: Same bug as noticed above in EventProcessor. I will push a commit with a fix. Thanks.

FWCore/Framework/src/EventProcessor.cc
```diff
@@ -862,6 +871,98 @@ namespace edm {
                              << "This likely indicates a bug in an input module or corrupted input or both\n";
   }
 
+  void EventProcessor::beginProcessBlock(bool& beginProcessBlockSucceeded) {
+    ProcessBlockPrincipal& processBlockPrincipal = principalCache_.processBlockPrincipal();
+    processBlockPrincipal.fillProcessBlockPrincipal(processConfiguration_->processName());
+
+    typedef OccurrenceTraits<ProcessBlockPrincipal, BranchActionGlobalBegin> Traits;
```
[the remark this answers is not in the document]
wddgit: I will push a commit soon that replaces all the typedefs in EventProcessor.cc and SubProcess.cc with using statements. I was just copying the style already there. They look more modern and readable with this change. Thanks.

FWCore/Framework/src/EventProcessor.cc
```diff
@@ -862,6 +871,98 @@ namespace edm {
...
+    // For now the input source always returns false from readProcessBlock,
+    // so this does nothing at all.
+    // Eventually the ProcessBlockPrincipal needs to be properly filled
+    // and cleared. The delayed reader needs to be set. The correct process name
+    // needs to be supplied.
+    while (input_->readProcessBlock()) {
+      DelayedReader* reader = nullptr;
+      processBlockPrincipal.fillProcessBlockPrincipal(processConfiguration_->processName(), reader);
+
+      typedef OccurrenceTraits<ProcessBlockPrincipal, BranchActionGlobalOther> Traits;
```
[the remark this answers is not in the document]
wddgit: Done.

FWCore/Framework/src/EventProcessor.cc
```diff
@@ -862,6 +871,98 @@ namespace edm {
...
+      for (auto& s : subProcesses_) {
+        s.clearProcessBlockPrincipal(isInputProcessBlock);
+      }
+    }
+  }
+
+  void EventProcessor::endProcessBlock(bool cleaningUpAfterException, bool beginProcessBlockSucceeded) {
+    ProcessBlockPrincipal& processBlockPrincipal = principalCache_.processBlockPrincipal();
+
+    typedef OccurrenceTraits<ProcessBlockPrincipal, BranchActionGlobalEnd> Traits;
```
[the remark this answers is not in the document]
wddgit: Done

FWCore/Framework/src/SubProcess.cc
```diff
@@ -403,6 +412,53 @@ namespace edm {
...
+      ProcessBlockPrincipal const& principal,
+      IOVSyncValue const&,
+      std::vector<std::shared_ptr<const EventSetupImpl>> const*) {
+    ServiceRegistry::Operate operate(serviceToken_);
+
+    ProcessBlockPrincipal& processBlockPrincipal = principalCache_.inputProcessBlockPrincipal();
+    processBlockPrincipal.fillProcessBlockPrincipal(principal.processName(), principal.reader());
+    propagateProducts(InProcess, principal, processBlockPrincipal);
+
+    typedef OccurrenceTraits<ProcessBlockPrincipal, BranchActionGlobalOther> Traits;
```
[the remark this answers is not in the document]
wddgit: Done

FWCore/Framework/src/SubProcess.cc
```diff
@@ -403,6 +412,53 @@ namespace edm {
...
+      ProcessBlockPrincipal const& principal,
+      IOVSyncValue const&,
+      std::vector<std::shared_ptr<const EventSetupImpl>> const*) {
+    ServiceRegistry::Operate operate(serviceToken_);
+
+    ProcessBlockPrincipal& processBlockPrincipal = principalCache_.processBlockPrincipal();
+    processBlockPrincipal.fillProcessBlockPrincipal(processConfiguration_->processName());
+    propagateProducts(InProcess, principal, processBlockPrincipal);
+
+    typedef OccurrenceTraits<ProcessBlockPrincipal, BranchActionGlobalBegin> Traits;
```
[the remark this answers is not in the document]
wddgit: Done

FWCore/Framework/src/SubProcess.cc
```diff
@@ -403,6 +412,53 @@ namespace edm {
...
+
+  void SubProcess::doEndProcessBlockAsync(WaitingTaskHolder iHolder,
+                                          ProcessBlockPrincipal const& principal,
+                                          IOVSyncValue const&,
+                                          std::vector<std::shared_ptr<const EventSetupImpl>> const*,
+                                          bool cleaningUpAfterException) {
+    ProcessBlockPrincipal& processBlockPrincipal = principalCache_.processBlockPrincipal();
+    propagateProducts(InProcess, principal, processBlockPrincipal);
+
+    typedef OccurrenceTraits<ProcessBlockPrincipal, BranchActionGlobalEnd> Traits;
```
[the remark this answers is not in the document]
wddgit: Done

FWCore/Framework/src/Principal.cc
```diff
@@ -471,6 +475,20 @@ namespace edm {
...
 
+  // Set the principal for the ProcessBlock
+  void Principal::fillPrincipal(std::string const& processNameOfBlock, DelayedReader* reader) {
+    fillPrincipal(reader);
+
+    std::vector<std::string> const& lookupProcessNames = productLookup_->lookupProcessNames();
+    auto iter = std::find(lookupProcessNames.begin(), lookupProcessNames.end(), processNameOfBlock);
+    lookupProcessOrder_.assign(lookupProcessNames.size(), 0);
+    if (!lookupProcessOrder_.empty()) {
+      if (iter != lookupProcessNames.end()) {
```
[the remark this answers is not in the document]
wddgit: Done.

FWCore/Framework/src/PrincipalCache.h
```diff
@@ -50,6 +51,12 @@ namespace edm {
     ~PrincipalCache();
     PrincipalCache(PrincipalCache&&) = default;
 
+    ProcessBlockPrincipal& processBlockPrincipal() const { return *processBlockPrincipal_; }
+    ProcessBlockPrincipal& inputProcessBlockPrincipal() const { return *inputProcessBlockPrincipal_; }
+    ProcessBlockPrincipal& processBlockPrincipal(bool input) const {
```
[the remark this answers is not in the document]
wddgit: Done. I replaced the bool with enum and will push the commit soon.

FWCore/Framework/src/PrincipalCache.h
```diff
@@ -81,8 +90,10 @@ namespace edm {
     void throwRunMissing() const;
     void throwLumiMissing() const;
 
-    // These are explicitly cleared when finished with the run,
+    // These are explicitly cleared when finished with the processblock, run,
     // lumi, or event
+    std::unique_ptr<ProcessBlockPrincipal> processBlockPrincipal_;
+    std::unique_ptr<ProcessBlockPrincipal> inputProcessBlockPrincipal_;
```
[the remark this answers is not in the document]
wddgit: The new ProcessBlockPrincipal is created at beginProcessBlock (shortly after beginJob) and stays alive until endProcessBlock (a little before endJob). The products it contains stay alive the entire time. You can access a product put during beginProcessBlock in the endProcessBlock transition. It is not cleared. In a future PR when persistency is implemented, the design includes a loop over ProcessBlocks from the input file immediately after opening each input file. On each iteration of the loop the ProcessBlockPrincipal use for input will get cleared and refilled (or at least setup for delayed read).

FWCore/Framework/src/Worker.h
```diff
@@ -311,35 +315,7 @@ namespace edm {
 
     static void exceptionContext(cms::Exception& ex, ModuleCallingContext const* mcc);
 
-    /*This base class is used to hide the differences between the ID used
-     for Event, LuminosityBlock and Run. Using the base class allows us
-     to only convert the ID to string form if it is actually needed in
-     the call to shouldRethrowException.
-     */
-    class TransitionIDValueBase {
```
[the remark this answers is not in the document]
wddgit: Yes. It was not used. In addition, I would have needed to modify it and extend it to work in a reasonable way with ProcessBlock. Much easier to just delete it.

FWCore/Framework/src/global/OutputModuleBase.cc
```diff
@@ -157,6 +158,13 @@ namespace edm {
       ProductSelector::checkForDuplicateKeptBranch(desc, trueBranchIDToKeptBranchDesc);
 
       EDGetToken token;
+
+      std::vector<std::string> missingDictionaries;
+      if (!checkDictionary(missingDictionaries, desc.className(), desc.unwrappedType())) {
+        std::string context("Calling OutputModuleBase::keepThisBranch, checking dictionaries for kept types");
+        throwMissingDictionariesException(missingDictionaries, context);
+      }
```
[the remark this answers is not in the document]
wddgit: This is not related to ProcessBlock. I had to modify some other unrelated lines of code in this same OutputModuleBase function that occur later. I noticed that for "one" OutputModules we make this check, but for "global" and "limited" we didn't. It seemed like they ought to be consistent. Possibly I am wrong. Is there a reason for them to be different? It makes sense for PoolOutputModule to check this for products it will need to write. PoolOutputModule is a "one". At this point, all global or limited output modules are in Framework tests except for AsciiOutputModule. Let me know if you want me remove this check for global and limited.

FWCore/Framework/src/globalTransitionAsync.h
```diff
@@ -33,11 +34,12 @@
...
 
 namespace edm {
-  class EventSetupImpl;
   class LuminosityBlockPrincipal;
+  class ProcessBlockPrincipal;
   class RunPrincipal;
 
   //This is code in common between beginStreamRun and beginGlobalLuminosityBlock
+  template <typename T>
   inline void subProcessDoGlobalBeginTransitionAsync(
```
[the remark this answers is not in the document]
wddgit: The following line of code works for lumis, runs, and process blocks.
```
subProcessDoGlobalBeginTransitionAsync<Traits>(iWait, subProcess, iPrincipal, iTS, iEventSetupImpls);
```
It chooses from 3 functions using the overload mechanism. In the ProcessBlock case the function needs to be a template. To keep the function signature the same I made the run and lumi cases have a template parameter also. But you are correct, the run and lumi cases do not use the template parameter. The ProcessBlock function needs the template parameter to support the beginProcessBlock transition and the accessInputProcessBlock transitions using the same function. One of the advantages of this approach is that it minimized changes to the existing code.

FWCore/Framework/src/globalTransitionAsync.h
```diff
@@ -47,6 +49,7 @@ namespace edm {
     iSubProcess.doBeginLuminosityBlockAsync(std::move(iHolder), iPrincipal, iTS, iEventSetupImpls);
   }
 
+  template <typename T>
   inline void subProcessDoGlobalBeginTransitionAsync(
```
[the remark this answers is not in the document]
wddgit: Same answer as above

FWCore/Framework/test/stubs/TestGlobalAnalyzers.cc
```diff
@@ -300,6 +301,86 @@ namespace edmtest {
       }
     };
 
+    class ProcessBlockIntAnalyzer : public edm::global::EDAnalyzer<edm::ProcessBlockCache<UnsafeCache>> {
+    public:
+      explicit ProcessBlockIntAnalyzer(edm::ParameterSet const& p) : trans_(p.getParameter<int>("transitions")) {
+        {
+          auto tag = p.getParameter<edm::InputTag>("consumesBeginProcessBlock");
+          if (not tag.label().empty()) {
```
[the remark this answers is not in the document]
wddgit: I copied this pattern from some other existing test module. My first thought is just to leave it as it is. This is repeated many dozens of places in tests in this PR. It would take a while to change and it is probably good enough. I never really liked complicated  conditional expressions much. But maybe I just need to get used to them. But your suggestion would work and I don't feel strongly either. I'll do it if you like. We could also write a new utility function, put it somewhere, and replace all these little blocks with one line function calls.
```
template <typename T>
void ifTagCallConsumes(ParameterSet&, EDConsumerBase*, const char* label, EDGetTokenT<T>&);
```

FWCore/Framework/test/stubs/ToyIntProducers.cc
```diff
@@ -552,51 +580,138 @@ namespace edmtest {
...
   };
 
   void NonEventIntProducer::accumulate(edm::StreamID iID, edm::Event const& e, edm::EventSetup const&) const {}
+  void NonEventIntProducer::beginProcessBlockProduce(edm::ProcessBlock& processBlock) const {
+    if (not bpbGet_.isUninitialized()) {
+      check(processBlock.get(bpbGet_), bpbExpect_);
+    }
+    if (sleepTime_ > 0) {
+      usleep(sleepTime_);
+    }
```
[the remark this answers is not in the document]
wddgit: The purpose was to force them to run concurrently. There is a unit test trying to that with a short sleep. I also was running them manually, setting the sleep time large, and watching the Tracer output as I did different tests (including tests with and without product dependencies). Without a sleep they are so fast that I wondered if one would not finish before the next starts even when the Framework was trying to run them asynchronously.

FWCore/Framework/src/global/OutputModuleBase.cc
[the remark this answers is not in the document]
makortel: I was mostly curious, your explanation makes sense (thanks!). @Dr15Jones, do you have anything to add?

FWCore/Framework/src/globalTransitionAsync.h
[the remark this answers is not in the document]
makortel: Right, the function name is the same in all cases, and the type of `iPrincipal` dictates which overload gets set. And indeed the caller is templated on the `iPrincipal` type, so this solution makes sense. Thanks for the explanation.

FWCore/Framework/test/stubs/TestGlobalAnalyzers.cc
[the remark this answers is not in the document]
makortel: Given that it is a copied pattern, let's keep it for now.

FWCore/Framework/test/stubs/ToyIntProducers.cc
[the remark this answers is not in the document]
makortel: Ok, thanks for the explanation.

FWCore/Framework/test/stubs/TestGlobalAnalyzers.cc
```diff
@@ -300,6 +301,86 @@ namespace edmtest {
...
+        if (not getTokenBegin_.isUninitialized()) {
+          if (processBlock.get(getTokenBegin_) != valueToGet) {
+            throw cms::Exception("BadValue")
+                << "expected " << valueToGet << " but got " << processBlock.get(getTokenBegin_);
+          }
+        }
+      }
+
+      void analyze(edm::StreamID iID, edm::Event const&, edm::EventSetup const&) const override {
+        if (m_count < 1u) {
```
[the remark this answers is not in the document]
wddgit: Done. I will soon push a commit containing your suggestion. The test that compares a log file with Tracer output to a reference file covered this already, but this is sufficiently important it is worth testing two ways.

FWCore/Framework/test/test_non_event_ordering.sh
```diff
@@ -9,3 +9,9 @@ cmsRun $TEST_DIR/test_non_event_ordering_beginLumi_cfg.py || die "begin Lumi tes
 cmsRun $TEST_DIR/test_non_event_ordering_beginRun_cfg.py || die "begin Run test failed" $?
 cmsRun $TEST_DIR/test_non_event_ordering_endLumi_cfg.py  || die "end Lumi test failed" $?
 cmsRun $TEST_DIR/test_non_event_ordering_endRun_cfg.py || die "end Run test failed" $?
+echo test_non_event_ordering_beginProcessBlock_cfg.py
```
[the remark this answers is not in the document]
wddgit: Done. I will push commits soon.

wddgit:
please test

@makortel I believe I've done everything I need to do in response to the review comments. Let me know if there is anything else I need to do or you have more questions. Thanks for the careful review.

DataFormats/Provenance/src/BranchType.cc
```diff
@@ -95,53 +86,75 @@ namespace edm {
...
   }  // namespace
 
-  std::string const& BranchTypeToString(BranchType const& branchType) {
-    return ((branchType == InEvent) ? event : ((branchType == InRun) ? run : lumi));
-  }
+  std::string const& BranchTypeToString(BranchType const& branchType) { return branchTypeNames[branchType]; }
 
   std::string const& BranchTypeToProductTreeName(BranchType const& branchType) {
-    return ((branchType == InEvent) ? events : ((branchType == InRun) ? runs : lumis));
+    if (branchType >= InProcess) {
```
Dr15Jones, line 92: Given this is a programming error, why not just use `assert`?

DataFormats/Provenance/src/ProductRegistry.cc
```diff
@@ -35,13 +35,11 @@ namespace edm {
...
         productProduced_(),
         anyProductProduced_(false),
-        eventProductLookup_(new ProductResolverIndexHelper),
-        lumiProductLookup_(new ProductResolverIndexHelper),
-        runProductLookup_(new ProductResolverIndexHelper),
-        eventNextIndexValue_(0),
-        lumiNextIndexValue_(0),
-        runNextIndexValue_(0),
-
+        productLookups_{{new ProductResolverIndexHelper,
```
Dr15Jones, line 38: would `std::make_unique<ProductResolverIndexHelper>()` work here?

DataFormats/Provenance/src/ProductRegistry.cc
```diff
@@ -54,13 +52,10 @@ namespace edm {
...
 
     // propagate_const<T> has no reset() function
-    eventProductLookup_ = std::make_unique<ProductResolverIndexHelper>();
-    lumiProductLookup_ = std::make_unique<ProductResolverIndexHelper>();
-    runProductLookup_ = std::make_unique<ProductResolverIndexHelper>();
-
-    eventNextIndexValue_ = 0;
-    lumiNextIndexValue_ = 0;
-    runNextIndexValue_ = 0;
+    for (int i = InEvent; i < NumBranchTypes; ++i) {
```
Dr15Jones, line 55:
````
for(auto &p : productLookups_) { p = std::make_unique<ProductResolverIndexHelper>(); }
```
would be better since you have made the tacit assumption that `InEvent==0` but the point of the symbols is to not have to know that.

DataFormats/Provenance/src/BranchType.cc
```diff
@@ -16,52 +19,40 @@ namespace edm {
...
-    std::string const event = "Event";
+    std::array<std::string, NumBranchTypes> const branchTypeNames{{"Event", "LuminosityBlock", "Run", "ProcessBlock"}};
+    std::size_t constexpr eventLumiRunSize = 3;
+    using NameArray = const std::array<std::string, eventLumiRunSize>;
 
-    // Trees, branches, indices
-    std::string const runs = run + 's';
-    std::string const lumis = lumi + 's';
-    std::string const events = event + 's';
+    NameArray treeNames{{branchTypeNames[InEvent] + 's', branchTypeNames[InLumi] + 's', branchTypeNames[InRun] + 's'}};
```
Dr15Jones, line 26: This assumes `InEvent==0`, `InLumi == 1` and `InRun` ==2 else the arrays get out of order. I think it would be better to just stick with the numbers and not the symbols.

DataFormats/Provenance/src/ProductRegistry.cc
```diff
@@ -454,10 +441,12 @@ namespace edm {
     productLookup(InEvent)->setFrozen();
     productLookup(InLumi)->setFrozen();
     productLookup(InRun)->setFrozen();
+    productLookup(InProcess)->setFrozen();
 
-    transient_.eventNextIndexValue_ = productLookup(InEvent)->nextIndexValue();
-    transient_.lumiNextIndexValue_ = productLookup(InLumi)->nextIndexValue();
-    transient_.runNextIndexValue_ = productLookup(InRun)->nextIndexValue();
+    for (unsigned int i = InEvent; i < NumBranchTypes; ++i) {
```
Dr15Jones, line 446: Again, this assumes `InEvent==0` for all time.

FWCore/Framework/interface/BranchActionType.h
```diff
@@ -12,7 +12,8 @@ namespace edm {
     BranchActionGlobalBegin = 0,
     BranchActionStreamBegin = 1,
     BranchActionStreamEnd = 2,
-    BranchActionGlobalEnd = 3
+    BranchActionGlobalEnd = 3,
+    BranchActionGlobalOther = 4
```
Dr15Jones, line 16: This name doesn't appear to be very self-explanatory. Would another one be better?

FWCore/Framework/interface/ProcessBlockPrincipal.h
```diff
@@ -0,0 +1,44 @@
...
+
+    std::string const& processName() const { return processName_; }
+
+    void put(ProductResolverIndex index, std::unique_ptr<WrapperBase> edp) const;
+
+    // Should only be 1 ProcessBlock needed at a time (no concurrent ProcessBlocks)
+    unsigned int index() { return 0; }
+
+  private:
+    unsigned int transitionIndex_() const override;
```
Dr15Jones, line 38: That could probably be marked `final`

FWCore/Framework/interface/BranchActionType.h
Dr15Jones: So for events which do not have a begin or end we just use `BranchActionStreamBegin`. So for input process block, it is like event (no begin or end) so I suggest using the same convention and just use `BranchActionGlobalBegin`.  Otherwise we should add two new types which deal with the 'instantaneous' nature of event and input process block.

Dr15Jones: I see that `BranchActionGlobalBegin` can't be used for this case as it is already used for 'begin' process block. So instead how about being more descriptive `BranchActionGlobalInput` or very descriptive `BranchActionProcessBlockInput`?

FWCore/Framework/interface/Principal.h
```diff
@@ -80,7 +80,9 @@ namespace edm {
 
     void adjustIndexesAfterProductRegistryAddition();
 
+    void fillPrincipal(DelayedReader* reader);
```
Dr15Jones, line 83: Are these different `fillPrincipal` needed for different inheriting types? If so, then having them in the base class feels wrong.

FWCore/Framework/interface/global/analyzerAbilityToImplementor.h
```diff
@@ -37,6 +37,11 @@ namespace edm {
         typedef edm::global::impl::StreamCacheHolder<edm::global::EDAnalyzerBase, C> Type;
       };
 
+      template <typename C>
+      struct AbilityToImplementor<edm::ProcessBlockCache<C>> {
```
Dr15Jones, line 41: So the only difference between a ProcessBlockCache and a GlobalCache is another ProcessBlock can be activated if we close an output file during running?

wddgit: There will be a ProcessBlockCache for each ProcessBlock in the input that has any products in it. Probably we would clear them all when output files are closed, but even in the normal case with only one output file per output module there could be many ProcessBlockCache objects. The idea is input ProcessBlocks cannot be saved in memory throughout the process and only one input ProcessBlock is allowed in memory at a time. The ProcessBlockCache is a way to communicate some part of the ProcessBlock data to Events, Lumis, and Runs with the Framework keeping track of which ProcessBlocks (could be more than one) correspond to a particular Event, Lumi, or Run.

DataFormats/Provenance/src/BranchType.cc
wddgit: I will make this change. An assert is better here. Thanks.

DataFormats/Provenance/src/BranchType.cc
wddgit: Hi Chris, I disagree with you on this one, because BranchType is in the interface of the functions that appear later in the file. It is the argument type to those functions. Those functions use BranchType to access the element in the std::array's this code builds. The dependence exists even if we build the arrays without the enum values.  Also, I don't like putting in magic numbers (0,1,2) here.

I was trying to think of something that could be done better here. Note that in the initial pull request, I simply extended the pre-existing coding style to include ProcessBlock. Then I switched to using std::array because Matti asked (and it's better that way). I could move branchTypeNames to the FWCore/Utilities/interface/BranchType.h file. Or I could add a stronger comment in BranchType.h about the fact that multiple things depend on the enum values (there already is such a comment).  Note that it's not only here, there are also places in IOPool that depend on the enum values.

Another alternative is to rewrite this file entirely in some different style in a different PR.

There are a lot of ways this could be done, but what is in the PR now will work.

makortel: How about something along the following?
```cpp
std::array<std::string, eventLumiRunSize> makeNameArray(std::string_view postfix) {
  std::array<std::string, eventLumiRunSize> ret;
  for (auto i=0U; i != eventLumiRunSize; ++i) {
    ret[i] = branchTypeNames[i] + postfix;
  }
  return ret;
}
NameArray treeNames = makeNameArray("s");
NameArray metaTreeNames = makeNameArray(metaData);
```

Dr15Jones: I'm with @makortel on this (as we require a runtime behavior anyway). The problem with using `InEvent`, `InLumi` and `InRun` is it makes the initialization appear invariant to the changes of value for the `In*` variables. However, that is not the case. If `InEvent` and `InLumi` values were swapped, the order of elements in the array `treeNames` would remain the same (i.e. `"Events"` would be in the 0th element) but using `treeNames[InEvent]` would return the string`"LuminosityBlocks"`.

DataFormats/Provenance/src/ProductRegistry.cc
wddgit: Done.

DataFormats/Provenance/src/ProductRegistry.cc
wddgit: Done

DataFormats/Provenance/src/BranchType.cc
wddgit: I added static_assert's requiring the enum values have the values that the design in this file requires. I also put in a makeNameArray function to reduce the code duplication. I pushed the commit so you can look and see if you like it. I'll not start the tests yet because a few more commits are coming tomorrow.

DataFormats/Provenance/src/ProductRegistry.cc
wddgit: Done

DataFormats/Provenance/src/BranchType.cc
```diff
@@ -16,52 +20,37 @@ namespace edm {
...
+      static_assert(InLumi == 1);
+      static_assert(InRun == 2);
+      return NameArray{
+          {branchTypeNames[InEvent] + postfix, branchTypeNames[InLumi] + postfix, branchTypeNames[InRun] + postfix}};
+    }
 
-    std::string const runInfo = run + "StatusInformation";      // backward compatibility
-    std::string const lumiInfo = lumi + "StatusInformation";    // backward compatibility
-    std::string const eventInfo = event + "StatusInformation";  // backward compatibility
+    NameArray treeNames{makeNameArray(std::string("s"))};
```
makortel, line 35: A rather minor point, the construction of `std::string` could be avoided if `makeNameArray` would take `std::string_view`.

DataFormats/Provenance/src/BranchType.cc
```diff
@@ -16,52 +20,37 @@ namespace edm {
...
+    std::size_t constexpr eventLumiRunSize = 3;
+    using NameArray = const std::array<std::string, eventLumiRunSize>;
 
-    std::string const runMeta = run + metaData;
-    std::string const lumiMeta = lumi + metaData;
-    std::string const eventMeta = event + metaData;
+    NameArray makeNameArray(std::string const& postfix) {
+      static_assert(InEvent == 0);
+      static_assert(InLumi == 1);
+      static_assert(InRun == 2);
```
makortel, line 30: Now at least it won't compile if the values are changed. On the other hand, a loop would work automatically with value changes.

(this discussion makes me wonder if we have implicit assumptions on their numerical values somewhere else)

DataFormats/Provenance/src/BranchType.cc
wddgit: I tried and had difficulties because operator+ does not support adding a string and string_view. This seemed the simplest way around that restriction.

DataFormats/Provenance/src/BranchType.cc
wddgit: In the version of this file in the current master release, there is no assumption about the enum values. In other files, especially in the IOPool subsystem, such assumptions already exist in the code (and one might find them other places if one looked).

When converting to std::array and using the BranchType as an index, I created the dependence. First here:
```
std::array<std::string, NumBranchTypes> const branchTypeNames{{"Event", "LuminosityBlock", "Run", "ProcessBlock"}};
```
Personally, I think the current version in the PR is good enough. It is simple and easy to read. I doubt the enum values will ever be reordered. If you want I could change initialization of the above array so that the enum values could  be reordered but still restricted to 0, 1, 2. If we want to support arbitrary values like InEvent = 100, InLumi = 101, and InRun = 102, then I would need to use something more like maps or add an additional layer of indirection. Note that there is also a presumption in the file that InProcess is 3. Most of the arrays have size 3 (run/lumi/event and no processBlock). Another alternative would be to remove the std::array's and go back to old way of implementing this.

FWCore/Framework/interface/BranchActionType.h
wddgit: I changed the name to BranchActionProcessBlockInput. Thanks, that is a more descriptive name and will help when someone tries to read this code. I'll push a commit soon.

FWCore/Framework/interface/ProcessBlockPrincipal.h
wddgit: Done. I will push the commit soon.

FWCore/Framework/interface/Principal.h
wddgit: The first function is the common part that is the same for all four types of Principals. There is no problem with that one.

Yes. The second and third functions are needed for different inheriting types. The reason is that there is not a unique ProcessHistory associated with a ProcessBlock. This is a real and fundamental difference.

I agree this "feels" wrong. I was not completely happy with what I did here. I can see three choices.

1. Leave it as it is. Just live with it. It is probably the simplest solution.

2. I could combine the two functions into one function. When called from ProcessBlockPrincipal I could pass in dummy arguments and with an if-else block or two I could probably make it work. To be honest that still kind of feels "wrong", but it would be less obvious and hidden inside the function (not in the interface).

3. I could add a new class to the inheritance hierarchy. RunPrincipal inherits from PrincipalUsingProcessHistory which inherits from Principal. Same for EventPrincipal and LuminosityBlockPrincipal. Then move the content of one of the functions up into PrincipalUsingProcessHistory.  In addition, move the functionality in the other function up into ProcessBlockPrincipal since that is the only type not using the ProcessHistory. It would takes some time to work out the details, but it should work.

Which would you prefer? Or do you have a better idea?

wddgit:
please test

I pushed all the commits. As far as I know this is done now. I responded to all Chris's comments. There are 3 where I am half expecting you will reply to my response and ask for more changes (related to BranchType.cc, fillPrincipal, and string_view). The rest were straightforward and are resolved.

In the last commit I changed makeNameArray to use a loop. I liked the other way better, but it's not worth discussing more. The static_assert's are still there and I added another one for InProcess == 3.

Dr15Jones:
I confess that I've only been able to get through a review for the first 1/3rd of the files in this pull request.

wddgit:
It is big, isn't it. OK, I will be expecting more comments. I appreciate the careful review. It reduces the chance of errors.

FWCore/Framework/src/TransitionProcessors.icc
```diff
@@ -19,9 +19,14 @@ struct FileResources {
   ~FileResources() {
     // See the message in catch clause
     CMS_SA_ALLOW try {
-      ep_.respondToCloseInputFile();
-      ep_.closeInputFile(cleaningUpAfterException_);
-      ep_.closeOutputFiles();
+      // Don't try to execute the following sequence of functions twice.
+      // If the sequence was already attempted and failed, then do nothing.
+      if (!closingSequenceAlreadyFailed_) {
```
Dr15Jones, line 24: What prompted this change?

FWCore/Framework/src/TransitionProcessors.icc
```diff
@@ -242,18 +252,33 @@ private:
...
     iEP.readFile();
     iEP.respondToOpenInputFile();
+
+    iEP.inputProcessBlocks();
   }
 
   void gotoNewInputAndOutputFiles(EventProcessor& iEP) {
-    iEP.respondToCloseInputFile();
-    iEP.closeInputFile(false);
+    // If this is still true when we hit the destructor for
```
Dr15Jones, line 260: For clarity, how about putting the block of code where `closingSequenceAlreadyFailed_` is being used in a new block i.e. `{...}`

FWCore/Framework/src/TransitionProcessors.icc
```diff
@@ -242,18 +252,33 @@ private:
...
 
+    iEP.respondToCloseInputFile();
+    bool cleaningUpAfterException = false;
+    iEP.closeInputFile(cleaningUpAfterException);
+    iEP.endProcessBlock(cleaningUpAfterException, filesOpen_->beginProcessBlockSucceeded_);
     iEP.closeOutputFiles();
 
+    // Initialize these for the next output file
+    filesOpen_->closingSequenceAlreadyFailed_ = false;
+    filesOpen_->beginProcessBlockSucceeded_ = false;
```
Dr15Jones, line 273: Similarly, put the lines of code dealing with `beginProcessBlockSucceeded_` into its own block

Dr15Jones (changes requested):

FWCore/Framework/src/TransitionProcessors.icc
wddgit: I was testing by intentionally throwing exceptions in different transitions to see that the behavior after an exception and the printed messages were OK. When I tried this in endProcessBlock under certain circumstances it would throw an exception and then later try running the endProcessBlock transition again in the FileResources destructor and terminate immediately instead of cleanly dealing with the exception and printing the intended message.

FWCore/Framework/src/TransitionProcessors.icc
wddgit: Done

FWCore/Framework/src/TransitionProcessors.icc
wddgit: Done

wddgit:
please test

FWCore/Framework/src/global/OutputModuleBase.cc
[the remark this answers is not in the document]
Dr15Jones: Being consistent is good.

FWCore/Framework/src/globalTransitionAsync.h
```diff
@@ -113,6 +136,27 @@ namespace edm {
...
 
+  // The only purpose of this function is to create dummy values for the
+  // EventSetup arguments needed by the generic function beginGlobalTransitionsAsync
+  template <typename Traits, typename P, typename SC>
+  void beginGlobalTransitionAsync(
+      WaitingTaskHolder iWait, Schedule& iSchedule, P& iPrincipal, ServiceToken const& token, SC& iSubProcesses) {
+    // There is no EventSetup for ProcessBlock transitions
+    // so we just pass in dummy values that are not used.
+    IOVSyncValue dummyIOVSyncValue;
+    EventSetupImpl dummyEventSetupImpl;
```
Dr15Jones, line 147: I'm really not a fan of this. Should we redo the `beginGlobalTransitionAsync` to take as argument something form the `Traits` which can be used to deal with EventSetup related information?

FWCore/Framework/src/limited/EDProducerBase.cc
```diff
@@ -75,6 +76,33 @@ namespace edm {
...
+      ProcessBlock const& constProcessBlock = processBlock;
+      this->doBeginProcessBlock_(constProcessBlock);
+      processBlock.setProducer(this);
+      this->doBeginProcessBlockProduce_(processBlock);
+      commit_(processBlock);
+    }
+
+    void EDProducerBase::doAccessInputProcessBlock(ProcessBlockPrincipal const& pbp, ModuleCallingContext const* mcc) {
+      ProcessBlock processBlock(pbp, moduleDescription_, mcc, false);
+      processBlock.setConsumer(this);
```
Dr15Jones, line 91: Just an observation. It really seems like the Principal wrappers should have their constructors changed so that the consumer is passed to the constructor. Calling a constructor followed immediately by a `set` function smells bad.

FWCore/Framework/test/stubs/TestOneFilters.cc
```diff
@@ -228,6 +229,191 @@ namespace edmtest {
       }
     };
 
+    class ProcessBlockIntFilter : public edm::one::EDFilter<edm::ProcessBlockCache<fltr::Cache>> {
```
Dr15Jones, line 232: How does the ProcessBlock cache get created? There doesn't appear to be a mechanism in this module. If a ::one module has a RunCache the beginRun function returns a `std::shared_ptr` to the cache object.

Dr15Jones: I just noticed that this question applies to every single module type which can declare a `edm::ProcessBlockCache`.

FWCore/Framework/src/limited/EDProducerBase.cc
wddgit: From the way you worded the comment, I am assuming this is something we'll deal with in a separate PR later. This pattern is simply copied from what we already do for Runs, Lumis, and Events. It is not new with ProcessBlock.

FWCore/Framework/test/stubs/TestOneFilters.cc
wddgit: I planned to address this in one of the future PR's implementing persistence. There are no tests of this yet and the function creating the caches will never actually be called by any code in this PR. But the interface is in this PR, for example for global modules in FWCore/Framework/interface/global/implementors.h:
```
virtual std::shared_ptr<C> accessInputProcessBlock(ProcessBlock const&) const { return std::shared_ptr<C>(); }
```

Dr15Jones:
Do we have any tests using the ProcessBlock transistions but running cmsRum in the 'Run' and 'RunAndLumi' modes?

FWCore/Integration/test/TestFindProduct.cc
```diff
@@ -29,7 +30,12 @@
 #include <vector>
 
 namespace edmtest {
-  class TestFindProduct : public edm::one::EDAnalyzer<edm::one::WatchRuns, edm::one::WatchLuminosityBlocks> {
+
+  struct DummyTestFindProduct {};
+
+  class TestFindProduct : public edm::one::EDAnalyzer<edm::one::WatchRuns,
+                                                      edm::one::WatchLuminosityBlocks,
+                                                      edm::ProcessBlockCache<DummyTestFindProduct>> {
```
Dr15Jones, line 38: Does this illustrate wanting a `edm::one::WatchProcessBlocks` instead of/in addition to `edm::ProcessBlockCache<>`?

FWCore/ServiceRegistry/src/GlobalContext.cc
```diff
@@ -74,9 +92,15 @@ namespace edm {
...
         os << "global end Run " << RunID(gc.luminosityBlockID().run());
         break;
+      case GlobalContext::Transition::kEndProcessBlock:
+        os << "end ProcessBlock";
+        break;
       case GlobalContext::Transition::kEndJob:
         os << "endJob";
         break;
+      case GlobalContext::Transition::kWriteProcessBlock:
+        os << "write ProcessBlock ";
```
Dr15Jones, line 102: There seems to be an extra space at the end of the string.

makortel:
(those tests are failing in the IB)

FWCore/Framework/src/limited/EDProducerBase.cc
makortel: Let's look into it in a separate PR, I opened an issue to remind us https://github.com/cms-sw/cmssw/issues/30317.

FWCore/Utilities/interface/ProcessBlockIndex.h
```diff
@@ -0,0 +1,47 @@
...
+  private:
+    ///Only the ProcessBlockPrincipal is allowed to make one of these
+    friend class ProcessBlockPrincipal;
+    explicit ProcessBlockIndex(unsigned int iIndex) : value_(iIndex) {}
+
+    ProcessBlockIndex() = delete;
+
+    unsigned int value_;
+
+    static const unsigned int invalidValue_;
```
Dr15Jones, line 43: Could this be `constexpr`?

FWCore/Utilities/interface/Transition.h
```diff
@@ -1,29 +1,41 @@
...
+  enum class Transition {
+    Event,
+    BeginLuminosityBlock,
+    EndLuminosityBlock,
+    BeginRun,
+    EndRun,
+    BeginProcessBlock,
+    EndProcessBlock,
+    NumberOfTransitions,
+    NumberOfEventSetupTransitions = 5
```
Dr15Jones, line 21: How about 
```suggestion
    NumberOfEventSetupTransitions = BeginProcessBlock
```

Dr15Jones:
I have reviewed all files.

Dr15Jones:
Given that the cache deals with the input ProcessBlock behavior, perhaps changing the class name to `edm::InputProcessBlockCache`?

FWCore/Utilities/interface/ProcessBlockIndex.h
wddgit: If I do, then I have to initialize it in the header instead of the cc file, and that forces me to move  the #include for the limits header into ProcessBlockIndex.h. It is also inconsistent with RunIndex.h. I mildly prefer to leave it as it is. But I don't feel strongly. I'll move it if you ask.

FWCore/Utilities/interface/Transition.h
wddgit: Done.

FWCore/ServiceRegistry/src/GlobalContext.cc
wddgit: Done.

wddgit:
I'm not restarting tests because there are some comments I need to think about more before I respond, probably Monday before I get to it.

wddgit:
Two of Chris's comments together suggest the following change
for all the module types (global, stream, limited, one).

**Current PR**

Ability ```ProcessBlockCache<T>``` enables:
```
  void beginProcessBlock(ProcessBlock const&)
  std::shared_ptr<T> accessInputProcessBlock(ProcessBlock const&)
  void endProcessBlock(ProcessBlock const&)
```

**Instead have 2 abilities**

1. Ability ```WatchProcessBlock``` enables:
```
  void beginProcessBlock(ProcessBlock const&)
  void endProcessBlock(ProcessBlock const&)
```
2. Ability ```InputProcessBlockCache<T>``` enables:
```
  std::shared_ptr<T> accessInputProcessBlock(ProcessBlock const&)
```
I like this better than what is currently in the PR. Before
I start making the change, do both of you agree with this?
This will require a bit of work. It's not something I can
change in 5 minutes. 

Either way will work. From a users perspective, there are
two abilities to understand instead of one. But they are
simpler and more clearly named.

Internally, it will take more code to implement the two abilities,
but things will be better named.

FWCore/Framework/interface/Principal.h
wddgit: I responded to the comment on Principal.h related to the fillPrincipal functions. I just wanted to highlight this discussion because I could use some more feedback on what you want me to do. I am leaning towards leaving the PR as it is, but I could add this extra class in the inheritance hierarchy or do something else.

It feels a little bit like I pounded a square block in a round whole here and in one or two other places, because there are some differences between ProcessBlocks and Runs/Lumis/Events. But the alternatives introduce so much complexity that I think just living with this little imperfection is best. We could change this later if we find it is actually causing problems.

makortel:
> I like this better than what is currently in the PR. Before
> I start making the change, do both of you agree with this?

I like it, I think the `WatchProcessBlock` and `InputProcessBlockCache` are clearer and more consistent with other `Watch*` and `*Cache` extensions. I guess non-input `ProcessBlockCache` would not make much sense because there can be exactly one `ProcessBlock` being processed at a time by construction (or even by definition?).

@Dr15Jones, do you have further comments on this subject?

Dr15Jones:
I could see that any module type that can use a GlobalCache could also want a non-input `ProcessBlockCache`.

wddgit:
OK. So we want two abilities:
```
InputProcessBlockCache<T>
```
```
ProcessBlockCache<T>
```
With a separate cache for the new ProcessBlock as well. And we want this for all types: global, limited, stream and one. (Without any WatchProcessBlock ability)

It is easy to change this plan now, so if I am misunderstanding please let me know.

makortel:
After a chat with @Dr15Jones I take back my earlier "non-input ProcessBlockCache does not make sense". Such a concept would make sense to share a cache between stream module instances. On the other hand, a `GlobalCache` could be easily used as such a cache as well, and is probably simpler.

Using the `GlobalCache` for a "ProcessBlockCache" would mean that the `beginProcessBlock()`/`endProcessBlock()` for stream modules would take a pointer to non-const `GlobalCache` as an argument (but only if `GlobalCache` extension is used; if I read correctly in this PR the `GlobalCache` is delivered as a pointe-to-const). Maybe they should also be renamed to `globalBeginProcessBlock()`/`globalEndProcessBlock()`. In this case the ability would be called `WatchProcessBlock` (and it would be the first `Watch*` for stream modules).

makortel:
@wddgit (our messages crossed). To me, in principle, the `ProcessBlockCache<T>` feels unnecessary for `global` and `one` (one can just modify the object members because the module object itself is a global one). For `stream` such a concept would be useful (to be shared across streams), but the `GlobalCache<T>` could also be used for that. What to do with `limited` is not currently clear to me.

Apparently the question boils down to whether it is better to allow using `GlobalCache<T>` for "process block cache" in `stream` modules, or provide a separate cache type. I would be leaning a bit towards using `GlobalCache<T>` to be simpler, but could be missing some detail.

What do you think?

wddgit:
I like using the GlobalCache for stream modules.

I will implement ```WatchProcessBlock``` and ```InputProcessBlockCache<T>``` unless there are further comments here.

limited should be the same as global.

I don't see a reason to name them globalBeginProcessBlock, because there is no streamBeginProcessBlock. Simply beginProcessBlock is shorter and then the name will be the same everywhere.

I think it is a mistake that I made the GlobalCache pointer be **const** in the stream methods. The methods should not be called concurrently, so the pointer does not need to be const. I will change that.

makortel:
@wddgit Sounds good. One (maybe last) suggestion from my earlier chat with @Dr15Jones, in case a stream module does not use `GlobalCache`, could the `GlobalCache` argument be left out? The `EndRunProducer` behaves that way towards `RunSummaryCache`
https://twiki.cern.ch/twiki/bin/view/CMSPublic/FWMultithreadedFrameworkStreamModuleInterface#edm_EndRunProducer
(even though it would be inconsistent with e.g. `RunCache` which always passes `GlobalCache` or void
https://twiki.cern.ch/twiki/bin/view/CMSPublic/FWMultithreadedFrameworkStreamModuleInterface#edm_RunCacheT
)

wddgit:
Yes. I will implement it so there is no GlobalCache* argument when there is no GlobalCache ability. That makes sense. Thanks.

FWCore/Framework/interface/Principal.h
makortel: Let's leave it as it is now, and see later if we get a better idea for the code organization.

wddgit:
I'm not running tests yet intentionally. 3 more commits coming soon. I expect tests might fail until I rebase on top of Chris's PR making modules run concurrently. There are some logical conflicts in there. The commit I just pushed implements the two new abilities we agreed to.

wddgit:
please test

There are 2 more commits related to Chris's comments coming soon so no need to approve this PR yet.

I rebased on top of the EventSetup concurrent module PR recently merged plus the last commit I pushed resolves some logical conflicts with that PR.

FWCore/Framework/src/globalTransitionAsync.h
wddgit: I spent a while improving the interface in globalTransitionAsync.h. I think it is better now. There were many ways this could be done. Let me know if you have comments and want me to do this differently.  The changes are in a separate commit (currently the last one).

After this there is one more comment I still need to address, so I am not starting the tests yet.

wddgit:
I added tests for the modes for Runs and RunsAndLumis. It is a little hard to see how the Framework code related to those modes could interact badly with the ProcessBlock  code, but now the tests will check for that. Note that I had already modified TestProcessor to include the ProcessBlock transitions. So testing a specific module using TestProcessor was already supported and should be able to test a module in those modes.

wddgit:
Please test

I believe I've responded to all the comments now. Let me know if there is anything further I need to do or if there are any more questions.

makortel:
Looks good to me, thanks. @Dr15Jones do you have any further comments?

FWCore/Framework/interface/stream/callAbilities.h
```diff
@@ -65,33 +65,23 @@ namespace edm {
...
     //********************************
     template <typename T, bool>
-    struct CallGlobalProcessBlockImpl {
-      static void beginProcessBlock(edm::ProcessBlock const& iProcessBlock, typename T::GlobalCache const* iGC) {
-        T::beginProcessBlock(iProcessBlock, iGC);
-      }
-
-      static void accessInputProcessBlock(edm::ProcessBlock const& iProcessBlock, typename T::GlobalCache const* iGC) {
+    struct CallInputProcessBlockImpl {
+      static void accessInputProcessBlock(edm::ProcessBlock const& iProcessBlock, typename T::GlobalCache* iGC) {
```
Dr15Jones, line 72: I believe using non-const `T::GlobalCache*` was what we agreed on, correct? My memory is fuzzy.

FWCore/Framework/test/stubs/TestStreamProducers.cc
```diff
@@ -500,7 +500,7 @@ namespace edmtest {
         }
       }
 
-      static std::shared_ptr<UnsafeCache> accessInputProcessBlock(edm::ProcessBlock const&, TestGlobalCache const*) {
+      static std::shared_ptr<UnsafeCache> accessInputProcessBlock(edm::ProcessBlock const&, TestGlobalCache*) {
```
Dr15Jones, line 503: Does this get called? I thought it required `edm::InputProcessBlockCache<>` to be used?

FWCore/Framework/interface/stream/callAbilities.h
wddgit: Yes. And the reason was that this function fills the GlobalCache so it is expected that the GlobalCache will change. And also it will be called serially so there is no need for the implementation to make it thread safe.

FWCore/Framework/test/stubs/TestStreamAnalyzers.cc
```diff
@@ -492,7 +492,7 @@ namespace edmtest {
         }
       }
 
-      static std::shared_ptr<Cache> accessInputProcessBlock(edm::ProcessBlock const&, TestGlobalCacheAn const*) {
+      static std::shared_ptr<Cache> accessInputProcessBlock(edm::ProcessBlock const&, TestGlobalCacheAn*) {
```
Dr15Jones, line 495: Will this actually be called?

FWCore/Framework/interface/OccurrenceTraits.h
```diff
@@ -424,6 +425,7 @@ namespace edm {
     using MyPrincipal = ProcessBlockPrincipal;
     using Context = GlobalContext;
     static bool constexpr isEvent_ = false;
+    static Transition constexpr transition_ = Transition::BeginProcessBlock;
```
Dr15Jones, line 435: Is this BeginProcessBlock or is it InputProcessBlock?

FWCore/Framework/interface/OccurrenceTraits.h
```diff
@@ -423,6 +429,7 @@ namespace edm {
   class OccurrenceTraits<ProcessBlockPrincipal, BranchActionProcessBlockInput> {
   public:
     using MyPrincipal = ProcessBlockPrincipal;
+    using TransitionInfoType = ProcessBlockTransitionInfo;
     using Context = GlobalContext;
     static bool constexpr isEvent_ = false;
     static Transition constexpr transition_ = Transition::BeginProcessBlock;
```
Dr15Jones, line 435: We don't have a separate ProcessBlockInput transition?

FWCore/Framework/src/EventProcessor.cc
```diff
@@ -1275,15 +1272,10 @@ namespace edm {
 
           EventSetupImpl const& es = status->eventSetupImpl(esp_->subProcessIndex());
           {
+            LumiTransitionInfo transitionInfo(lumiPrincipal, es, &status->eventSetupImpls());
```
Dr15Jones, line 1275: I like how the TransitionInfos make the global work more uniform. I'd suggest we also do the same for the Stream functions.

wddgit: Do you want this in this PR? Or as a separate PR? ProcessBlock does not even use the stream transitions.

FWCore/Framework/src/TransitionInfoTypes.h
```diff
@@ -0,0 +1,73 @@
...
+    EventSetupImpl const& eventSetupImpl_;
+  };
+
+  class ProcessBlockTransitionInfo {
+  public:
+    ProcessBlockTransitionInfo(ProcessBlockPrincipal& iPrincipal) : processBlockPrincipal_(iPrincipal) {}
+
+    ProcessBlockPrincipal& principal() { return processBlockPrincipal_; }
+    ProcessBlockPrincipal const& principal() const { return processBlockPrincipal_; }
+    EventSetupImpl const& eventSetupImpl() const;
```
Dr15Jones, line 65: We could probably find a way to avoid this.

FWCore/Framework/interface/OccurrenceTraits.h
wddgit: I'm thinking about this. Before I recently merged in the concurrent EventSetup module pull request, there was no reason to add this transition into the Transition enum. It wasn't needed anywhere.  Still thinking, I'm not sure.

FWCore/Framework/src/globalTransitionAsync.h
```diff
@@ -18,161 +18,105 @@
...
-    auto subs = make_waiting_task(
-        tbb::task::allocate_root(),
-        [&iSubProcesses, iWait, &iPrincipal, iTS, iEventSetupImpls](std::exception_ptr const* iPtr) mutable {
-          if (iPtr) {
-            auto excpt = *iPtr;
-            auto delayError =
-                make_waiting_task(tbb::task::allocate_root(),
+    auto subs =
+        make_waiting_task(tbb::task::allocate_root(),
+                          [&iSubProcesses, iWait, info = transitionInfo](std::exception_ptr const* iPtr) mutable {
```
Dr15Jones, line 86: OK. Copying of the TransitionInfo is done and is necessary.

FWCore/Framework/src/TransitionInfoTypes.h
wddgit: Schedule::processOneGlobalAsync requires and EventSetupImpl. So we have to create a dummy object somewhere unless I change the Schedule class interface also.

FWCore/Framework/src/globalTransitionAsync.h
```diff
@@ -181,46 +125,20 @@ namespace edm {
...
-                                    iWait, subProcess, iPrincipal, iTS, iEventSetupImpls, cleaningUpAfterException);
+                                subProcessDoGlobalEndTransitionAsync(iWait, subProcess, info, cleaningUpAfterException);
                               }
                             }
                           });
 
     WaitingTaskHolder h(subs);
-    iSchedule.processOneGlobalAsync<Traits>(std::move(h), iPrincipal, iES, token, cleaningUpAfterException);
+    iSchedule.processOneGlobalAsync<Traits>(
+        std::move(h), transitionInfo.principal(), transitionInfo.eventSetupImpl(), token, cleaningUpAfterException);
```
Dr15Jones, line 139: So the reason why the ProcessBlockTransitionInfo has a `eventSetupImpl()` method is because of the Schedule class? So we change it to also take a `TransitionInfo` as an argument?

Dr15Jones:
I don't think I saw any test modules using the `InputProcessBlockCache` ability? Did I just miss them or is the plan to add them later when that ability is more fleshed out?

FWCore/Framework/src/globalTransitionAsync.h
wddgit: We could try that. I would have to look closer, but I think this goes all the way down the stack into the Worker methods like doWorkAsync. And then the other things that call doWorkAsync would be affected if we changed that interface. In the end this might be an improvement, but it would affect a lot of code.

FWCore/Framework/interface/OccurrenceTraits.h
wddgit: I added a new value to the edm::Transition enum. Since OccurrenceTraits now wants this value, it makes since to add it.

wddgit: I meant "it makes sense to add it"

FWCore/Framework/test/stubs/TestStreamProducers.cc
wddgit: In this PR, it does not get called. When the next PR which implements persistence is submitted, it will get called and there will be meaningful tests.

FWCore/Framework/test/stubs/TestStreamAnalyzers.cc
wddgit: Same answer as previous comment.

FWCore/Framework/interface/OccurrenceTraits.h
wddgit: Fixed now with the new enum value.

wddgit:
please test

The plan is to implement the tests for the InputProcessBlockCache in the future PR that implements persistence.

I would propose implementing the changes in streamTransitionAsync in a separate PR, but if you want I can add it to this PR. Let me know and I will start working on that.

I would also propose changing Schedule::processOneGlobalAsync (and all the other things related to that) in a separate PR. Again, let me know and if you want I will add it to this PR. It was not clear to me that this was worth the effort to change at all.

I think I am back to the point that I've replied to all the comments. Let me know if there is something more you want me to do or there are more questions.

makortel:
> I would propose implementing the changes in streamTransitionAsync in a separate PR, but if you want I can add it to this PR. Let me know and I will start working on that.
> 
> I would also propose changing Schedule::processOneGlobalAsync (and all the other things related to that) in a separate PR. Again, let me know and if you want I will add it to this PR. It was not clear to me that this was worth the effort to change at all.

I agree on doing both improvements in later PR(s). This was is already huge, and I feel neither is really necessary for the ProcessBlock interface or implementation as in this PR.

wddgit:
please test

Try again. I think I have seen this zombie killer unit test failure before in other PRs. Also, it does not look related to this PR.

makortel:
+1

makortel:
@cmsbuild, please test

Let's refresh the tests though

makortel:
> Comparison is ready
> https://cmssdt.cern.ch/SDT/jenkins-artifacts/pull-request-integration/PR-63ad0a/7783/summary.html

@smuzaffar What happened to the summary of the comparisons?

The `validateJR` says no differences for both DQM and reco comparisons.

smuzaffar:
@mrodozov , can you please look in to this?

mrodozov:
@smuzaffar 

unfortunately the file is empty.
https://cmssdt.cern.ch/SDT/jenkins-artifacts/baseLineComparisons/CMSSW_11_2_X_2020-07-07-1100+63ad0a/37565/validateJR/qaResultsSummary.log

there was some failures in the compare-root-files-short-matrix job. I restarted the job (I don't remember if this is possible, to run it standalone)

```
10:06:19 Traceback (most recent call last):
10:06:19   File "/data/cmsbld/jenkins/workspace/compare-root-files-short-matrix/cms-bot/report-pull-request-results.py", line 6, in <module>
10:06:19     from github import Github
10:06:19 ImportError: cannot import name Github
10:09:07 Traceback (most recent call last):
10:09:07   File "/data/cmsbld/jenkins/workspace/compare-root-files-short-matrix/cms-bot/report-pull-request-results.py", line 6, in <module>
10:09:07     from github import Github
10:09:07 ImportError: cannot import name Github
10:10:30 you chose the action COMPARISON_READY
10:10:30 API Rate Limit: 4796/5000, Reset in 758 sec i.e. at 2020-07-09 09:23:01
10:10:30 Alt comparison directory:  /data/cmsbld/jenkins/workspace/compare-root-files-short-matrix/upload/alternative-comparisons
10:10:30 
10:10:30 JR comparison Summary:  /data/cmsbld/jenkins/workspace/compare-root-files-short-matrix/upload/validateJR/qaResultsSummary.log
```

silviodonato:
+1

PR #30120, opened 2020-06-05
TBLayer computeCrossings doesn't allow for some displaced starting points

aminnj:
When trying to compute expected barrel pixel layer hits for tracks with a displaced starting point, the logic in `TBLayer::computeCrossings` fails to return crossings when the starting point is between the inner and outer cylinders of a pixel layer. In the following plot of barrel pixel layers projected into the transverse plane, the current logic will return crossing points in blue for a dimuon pair originating at the red point, but fails to provide crossing points in the layer just above the red point.

image

This PR modifies the logic to allow for such crossings to be returned by not requiring that both `innerCrossing` and `outerCrossing` have solutions.

In my tests, "prompt" tracks are unaffected.

### DIFF
--- a/RecoTracker/TkDetLayers/src/TBLayer.cc
+++ b/RecoTracker/TkDetLayers/src/TBLayer.cc
@@ -55,7 +55,8 @@ SubLayerCrossings TBLayer::computeCrossings(const TrajectoryStateOnSurface& star
   HelixBarrelCylinderCrossing outerCrossing(
       startPos, startDir, rho, propDir, *theOuterCylinder, HelixBarrelCylinderCrossing::onlyPos);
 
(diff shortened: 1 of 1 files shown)


perrotta:
@aminnj what is the purpose to submit this PR in 11_0_X?
This has to start being submitted into the master (by now 11_2_X): then, do you really need a backport in 11_0_X, which is a basically closed cycle?

slava77:
This indeed should go to the master branch first.
Some backports can be considered, since this was actually used for analysis; 11_0 is perhaps the least useful for this. Perhaps 10_6_X.
Let's see if it introduces any changes in the production setup.

aminnj:
Yes, sorry, I was not paying close attention. There's no reason to use 11_0_X and no need for a backport in it. Let me prepare another PR wrt master.

PR #30127, opened 2020-06-05
TBLayer computeCrossings doesn't allow for some displaced starting points

aminnj:
When trying to compute expected barrel pixel layer hits for tracks with a displaced starting point, the logic in TBLayer::computeCrossings fails to return crossings when the starting point is between the inner and outer cylinders of a pixel layer. In the following plot of barrel pixel layers projected into the transverse plane, the current logic will return crossing points in blue for a dimuon pair originating at the red point, but fails to provide crossing points in the layer just above the red point.

image

This PR modifies the logic to allow for such crossings to be returned by not requiring that both innerCrossing and outerCrossing have solutions.

In my tests, "prompt" tracks are unaffected.

### DIFF
--- a/RecoTracker/TkDetLayers/src/TBLayer.cc
+++ b/RecoTracker/TkDetLayers/src/TBLayer.cc
@@ -49,14 +49,16 @@ SubLayerCrossings TBLayer::computeCrossings(const TrajectoryStateOnSurface& star
   GlobalVector startDir(startingState.globalMomentum());
   double rho(startingState.transverseCurvature());
 
+  bool inBetween = ((theOuterCylinder->position() - startPos).perp() < theOuterCylinder->radius()) &&
+                   ((theInnerCylinder->position() - startPos).perp() > theInnerCylinder->radius());
+
   HelixBarrelCylinderCrossing innerCrossing(
       startPos, startDir, rho, propDir, *theInnerCylinder, HelixBarrelCylinderCrossing::onlyPos);
+  if (!inBetween && !innerCrossing.hasSolution())
+    return SubLayerCrossings();
 
   HelixBarrelCylinderCrossing outerCrossing(
       startPos, startDir, rho, propDir, *theOuterCylinder, HelixBarrelCylinderCrossing::onlyPos);
-
-  if (!innerCrossing.hasSolution() && !outerCrossing.hasSolution())
-    return SubLayerCrossings();
 
   if (!innerCrossing.hasSolution() && outerCrossing.hasSolution()) {
     innerCrossing = outerCrossing;
--- a/RecoTracker/TkDetLayers/src/TBLayer.cc
+++ b/RecoTracker/TkDetLayers/src/TBLayer.cc
@@ -51,13 +51,18 @@ SubLayerCrossings TBLayer::computeCrossings(const TrajectoryStateOnSurface& star
 
   HelixBarrelCylinderCrossing innerCrossing(
       startPos, startDir, rho, propDir, *theInnerCylinder, HelixBarrelCylinderCrossing::onlyPos);
-  if (!innerCrossing.hasSolution())
-    return SubLayerCrossings();
 
   HelixBarrelCylinderCrossing outerCrossing(
       startPos, startDir, rho, propDir, *theOuterCylinder, HelixBarrelCylinderCrossing::onlyPos);
-  if (!outerCrossing.hasSolution())
+
+  if (!innerCrossing.hasSolution() && !outerCrossing.hasSolution())
     return SubLayerCrossings();
+
+  if (!innerCrossing.hasSolution() && outerCrossing.hasSolution()) {
+    innerCrossing = outerCrossing;
+  } else if (!outerCrossing.hasSolution() && innerCrossing.hasSolution()) {
+    outerCrossing = innerCrossing;
+  }
 
   GlobalPoint gInnerPoint(innerCrossing.position());
   GlobalPoint gOuterPoint(outerCrossing.position());


slava77:
@cmsbuild please test

RecoTracker/TkDetLayers/src/TBLayer.cc
```diff
@@ -51,14 +51,19 @@ SubLayerCrossings TBLayer::computeCrossings(const TrajectoryStateOnSurface& star
...
   HelixBarrelCylinderCrossing innerCrossing(
       startPos, startDir, rho, propDir, *theInnerCylinder, HelixBarrelCylinderCrossing::onlyPos);
-  if (!innerCrossing.hasSolution())
-    return SubLayerCrossings();
 
   HelixBarrelCylinderCrossing outerCrossing(
       startPos, startDir, rho, propDir, *theOuterCylinder, HelixBarrelCylinderCrossing::onlyPos);
-  if (!outerCrossing.hasSolution())
+
+  if (!innerCrossing.hasSolution() && !outerCrossing.hasSolution())
```
slava77, line 58: the main difference by this point is the extra cost of computing the outerCrossing.
Looking at out monitoring for PU200 : track building takes up ~900 units vs about 2 units in ```TBLayer::computeCrossings```.
I think that this is a somewhat acceptable increase in cost.
Still, it would be better to add a  check that  startPos is close enough to care. 
[pseudo-coding, mind the typos]
One can use the following test to see if the starting point is in the gap between two cylinders 
```
(theOuterCylinder->position() - startPos).perp() < theOuterCylinder->radius() && (theInnerCylinder->position() - startPos).perp() > theInnerCylinder->radius() 
```
(this assumes the cylinders are aligned, perhaps it's good enough?)

aminnj: I verified I get the same exact results without this short circuit on my 100k event signal sample.
```cpp
  bool inBetween = ((theOuterCylinder->position() - startPos).perp() < theOuterCylinder->radius())
    && ((theInnerCylinder->position() - startPos).perp() > theInnerCylinder->radius());

  HelixBarrelCylinderCrossing innerCrossing(
      startPos, startDir, rho, propDir,*theInnerCylinder, HelixBarrelCylinderCrossing::onlyPos);
  if (!inBetween && !innerCrossing.hasSolution()) return SubLayerCrossings();

  HelixBarrelCylinderCrossing outerCrossing(
      startPos, startDir, rho, propDir,*theOuterCylinder, HelixBarrelCylinderCrossing::onlyPos);

  if (!innerCrossing.hasSolution() && outerCrossing.hasSolution()) {
    innerCrossing = outerCrossing;
  } else if (!outerCrossing.hasSolution() && innerCrossing.hasSolution()) {
    outerCrossing = innerCrossing;
  }
```
If this is what you had in mind, I'll update the PR

slava77: yes, it looks like what I wanted (in the sense of having a possible/likely return before the ```outerCrossing``` is created).

slava77:
> Reco comparison results: 14328 differences found in the comparisons

it looks like some differences show up in essentially all tested workflows.
Apparently this means that some logic has changed and will be less trivial to integrate.
At least this prevents a backport.
To still be useful for analysis needs in older releases I'm thinking that some internal flag will be needed to control the checks.

It could be that the behavior changes in the strip tracker (based on some larger differences in https://cmssdt.cern.ch/SDT/jenkins-artifacts/baseLineComparisons/CMSSW_11_2_X_2020-06-05-1100+6f7bf9/36844/validateJR/all_OldVSNew_RunEGamma2018Cwf136p874/all_OldVSNew_RunEGamma2018Cwf136p874c_recoTracks_generalTracks__reRECO_obj_originalAlgo.png in iter9 (pixel-less).
If that's the case, the solution may be to enable the check only for the pixel layers (e.g. TBLayer would have a "false" flag, which can be reset to "true" for TBPLayers).

slava77:
@cmsbuild please test

slava77:
@mtosi @JanFSchulte @VinInn @rovere @cerati @makortel 
do you recall if this compatible det situation (getting a compatible from between the layers) came up in the past? Was there a deliberate decision to return no compatible dets for points between two (inner and outer) cylinders in the (barrel) layer?

slava77:
> @mtosi @JanFSchulte @VinInn @rovere @cerati @makortel
> do you recall if this compatible det situation (getting a compatible from between the layers) came up in the past? Was there a deliberate decision to return no compatible dets for points between two (inner and outer) cylinders in the (barrel) layer?

@mtosi @JanFSchulte 
Please let me know if there are any concerns from TRK POG to merge this.
Thank you.

mtosi:
thanks @slava77 
this is the first time I read about this development
and I find it really interesting, actually
but, I think that as expected / design this change is not transparent at all
did you discuss it at one of the RECO meeting ?
shall we discuss about this at a TRK POG meeting ? (the next one will be at the end of June, though)

may I ask which is the goal of this update ?
@cerati @VinInn @rovere @makortel do you have any comments ?

VinInn:
@mtosi all coordinators should be subscribed to issues and pr affecting their packages.
Indeed none of the current POG coord are notified...
https://github.com/cms-sw/cmssw/pull/30127#issuecomment-639618926

mtosi:
can you please send me the link to the file which needs to be updated in order to be added to the list of packages ?

slava77:
> can you please send me the link to the file which needs to be updated in order to be added to the list of packages ?

https://github.com/cms-sw/cms-bot/blob/master/watchers.yaml


Note that the message from earlier today was quoting a message already sent to conveners in this thread 8 days ago.

mtosi:
thanks for the link
yes, sorry @slava77 I probably overlooked it :(

mtosi:
@aminnj will you be available for discussing this at the next TRK POG on Jun 29th at 2pm ?

in the meantime, looking at the code and considering the logic, I find it reasonable ....
@VinInn @rovere @cerati @venturia @stensokm do you find this update reasonable ?

aminnj:
@mtosi Yes, I am available to present this info at that time.

slava77:
@cmsbuild please test
to refresh the tests

mtosi:
@pgunnell @chayanit 
for the validation of the release in which this PR will be integrated, it would be interesting to check the impact on the conversion
and therefore SinglePhoton relval sample should be added to the list

thanks

mtosi:
btw, do we have a measurement of the timing ?

chayanit:
> @pgunnell @chayanit @wajidalikhan
> for the validation of the release in which this PR will be integrated, it would be interesting to check the impact on the conversion
> and therefore SinglePhoton relval sample should be added to the list
> 
> thanks

@mtosi sure but which SinglePhoton you referred to? which condition/ detector geometry is needed here?

slava77:
> @mtosi sure but which SinglePhoton you referred to? which condition/ detector geometry is needed here?

from the discussion in the POG meeting earlier today I understood that this was about the regular run3 realistic (corresponding to the standard set usually done for Run3/run2 relvals).

mtosi:
@slava77 thanks !
@chayanit yes, indeed, run2/3 should be the benchmark

slava77:
+1

for https://github.com/cms-sw/cmssw/pull/30127	20461e0 
- jenkins tests pass and comparisons with the baseline show small differences in tracking-related variables
- local tests on larger stat samples show negligible differences in the regular tracking distributions and large impact for InOut conversion tracks
    - IO conversion radius looks more appropriately placed at the inner pixel layer
    - electron gun events show very large decrease in duplicates in IO conversions as well as a good improvement in missing hits
    - general tracks are not visibly affected (ttbar PU50 test wf11024 had about 0.1% more reconstructed tracks (without an obvious source being fakes/dups or eff due to small size of the change)
    - CPU reco time on a reference 11024 workflow for 2018 is up by about 1% scattered among many modules, roughly equally in tracking and downstream (a significant part of it is likely within measurement uncertainties; although the nature of the change in this PR implies that more hits can be considered in tracker module navigation)
- signing based on the agreement in the POG meeting of Jun 29 that this change is conceptually correct

ttbar PU50 2018 wf 11024
wf11024_conv_io_r
here the IO conversions are more visibly increased on the outer edge of BPIX1, where the impact from the fix in this PR is expected

in the electron gun pt35 events (wf 11802) the duplicate rate is reduced and it now does not have the phi dependence corresponding to the outer BPIX ladders in a layer
wf10802_inOutConv_dup_phi
tracks are more complete with fewer missing inner
wf10802_inOutConv_misInner
and fewer missing outer hits
wf10802_inOutConv_misOuter

The single photon sample (wf 10805) also shows a shift/addition of the inner hit on the BPIX1 in IO conversions
wf10805_conv_rInnermost
with some more moderate improvement in the inner missing layers
wf10805_ckfIOconv_missing_inner

silviodonato:
+1

mtosi:
@slava77 thanks a lot for this detailed checks !
and thanks to @aminnj for this PR, of course

@bendavid from this preliminary checks it seems that your point is indeed confirmed

chayanit:
@mtosi since this PR is merged for 11_2_0_pre2, we will produce SinglePhoton sample for you specially to check this. Would you mind to open RelVal JIRA with information? I'm not sure which SinglePhoton would be good for this check. Thanks.

JanFSchulte:
@chayanit @mtosi I have opened a ticket here: https://its.cern.ch/jira/browse/PDMVRELVALS-89

chayanit:
@JanFSchulte noticed thanks

PR #30130, opened 2020-06-06
Replaced boost use in CalibFormats/SiStripObjects

camolezi:
#### PR description:
Replaced boost library use for C++ STL alternatives. The code should have similar behavior. 

#### PR validation:

@vgvassilev @davidlange6

### DIFF
--- a/CalibFormats/SiStripObjects/BuildFile.xml
+++ b/CalibFormats/SiStripObjects/BuildFile.xml
@@ -3,7 +3,6 @@
 <use name="CondFormats/SiStripObjects"/>
 <use name="CalibTracker/SiStripCommon"/>
 <use name="DataFormats/TrackerCommon"/>
-<use name="boost"/>
 <export>
   <lib name="1"/>
 </export>
--- a/CalibFormats/SiStripObjects/src/SiStripQuality.cc
+++ b/CalibFormats/SiStripObjects/src/SiStripQuality.cc
@@ -14,9 +14,6 @@
 // Needed only for output
 #include "DataFormats/DetId/interface/DetId.h"
 
-#include <boost/bind.hpp>
-#include <boost/function.hpp>
-
 SiStripQuality::SiStripQuality()
     : toCleanUp(false),
       FileInPath_("CalibTracker/SiStripCommon/data/SiStripDetInfo.dat"),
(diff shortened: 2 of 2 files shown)


mmusich:
Hello, 
out of curiosity - as I see this is one of a series - has the usage of `boost` become totally discouraged in cmssw?

camolezi:
I think @vgvassilev @davidlange6 would be better suited to answer that.

guitargeek:
Hi, this looks like a good opportunity to make the code more readable.

What do you think about using lambda functions  instead? From my experience, it's worth to at least consider lambda functions everywhere `boost::bind` was used previously. they are often more compact and readable.  
```C++
    std::remove_copy_if(
        runInfo->m_fed_in.begin(),
        runInfo->m_fed_in.end(),
        std::back_inserter(activeFedsFromRunInfo),
        [&](int x){ return !(x >= int(FEDNumbering::MINSiStripFEDID) && x <= int(FEDNumbering::MAXSiStripFEDID)); });
```

vgvassilev:
> Hello,
> out of curiosity - as I see this is one of a series - has the usage of `boost` become totally discouraged in cmssw?

I cannot speak for cmssw but from general engineering perspective having less library dependencies reduces the technical debt. I would say that the usage of `boost` should be encouraged in cases where facilities in the standard library fall short ;)

camolezi:
> Hi, this looks like a good opportunity to make the code more readable.
> 
> What do you think about using lambda functions instead? From my experience, it's worth to at least consider lambda functions everywhere `boost::bind` was used previously. Lambdas are often more compact and readable.
> 
> ```c++
>     std::remove_copy_if(
>         runInfo->m_fed_in.begin(),
>         runInfo->m_fed_in.end(),
>         std::back_inserter(activeFedsFromRunInfo),
>         [&](int x){ return !(x >= int(FEDNumbering::MINSiStripFEDID)
>                           && x <= int(FEDNumbering::MAXSiStripFEDID)); });
> ```
> 
> You also don't need the `<functional>` header in this case.

Yeah, thanks. I agree that using a lambda function, in this case, is much more cleaner. I will update the PR.

Dr15Jones:
please test

pohsun:
+1

silviodonato:
+1

PR #29906, opened 2020-05-19
Clean BuildFiles in Cond subsystems

guitargeek:
#### PR description:

Another quick partially automatic BuildFile cleaning PR in the style of many before (for example #29805), this time covering all Cond subsystems.

I checked that all the BuildFiles are still well formatted, to not undo any work done by the formatting PRs a few days ago.

#### PR validation:

CMSSW compiles. It was checked that all newly added dependencies are actually required by the package with `git grep`.

PR #30139, opened 2020-06-07
Clean BuildFiles before DataFormats

guitargeek:
#### PR description:

Another quick partially automatic BuildFile cleaning PR in the style of many before (for example https://github.com/cms-sw/cmssw/pull/29906), this time covering all subsystems before **DataFormats**.

Some plugins which were not merged with their header files actually had their header files in the `interface` directory, even though they are not needed by the library (only the plugins). This confuses my BuildFile/includes checker, which does not expect this entanglement between the library and the plugins. For these cases, I merged the plugin header with the `cc` file.

#### PR validation:


slava77:
@cmsbuild please test

jfernan2:
+1

guitargeek:
Thanks for your comment, @pieterdavid! Of course your developments have priority and the BuildFile cleaning can be easily rebased. I don't know if this PR could make it into 11_2_0_pre1, maybe @silviodonato can comment more. Otherwise there is no problem to leave the changes to the plugin code out.

slava77:
+1

for https://github.com/cms-sw/cmssw/pull/30139	59fea5f
- this is quite technical for reco, only some cleanups in BuildFiles

pohsun:
+1

guitargeek:
Hi @pieterdavid, I'm a bit confused now because this PR got the `alca-approved` a few days ago even thought you were asking for it to be delayed. Does this mean that there is actually no problem and it should be merged, of was the signature just a misunderstanding? Thanks for any clarification.

pieterdavid:
Hi @guitargeek thanks for checking, but there's no need to delay this (I brought it up because having it in last week's pre-release would have been convenient, but we will manage).

ggovi:
+1

silviodonato:
code-checks

silviodonato:
+1

PR #30169, opened 2020-06-10
GE2/1 pads with 16 eta partitions & option to send overflow GEM clusters

dildick:
#### PR description:

The GE2/1 will soon have 16 eta partitions in CMSSW to reflect the updated design. This PR defines a new pad builder for this particular geometry (not enabled yet). See also PR https://github.com/cms-sw/cmssw/pull/29440.

Also, the GE1/1 and GE2/1 on-chamber optohybrid to off-chamber backend (CTP7/ATCA) communication protocol will have an option to send overflow clusters in the next bunch crossing. This effectively increases the bandwidth by up to a factor 2. See https://gitlab.cern.ch/tdr/notes/DN-20-016/blob/master/temp/DN-20-016_temp.pdf, page 29. This option is also not enabled yet. 

EDIT: while checking the performance on Run-3 samples I noticed that a GEM copad vector was not cleared for each new event.

#### PR validation:

Code compiles.

N/A

FYI @watson-ij @jshlee

### DIFF
--- a/L1Trigger/CSCTriggerPrimitives/src/GEMCoPadProcessor.cc
+++ b/L1Trigger/CSCTriggerPrimitives/src/GEMCoPadProcessor.cc
@@ -27,6 +27,8 @@ GEMCoPadProcessor::GEMCoPadProcessor() : theRegion(1), theStation(1), theChamber
 void GEMCoPadProcessor::clear() { gemCoPadV.clear(); }
 
 std::vector<GEMCoPadDigi> GEMCoPadProcessor::run(const GEMPadDigiCollection* in_pads) {
+  clear();  
+    
   // Build coincidences
   for (auto det_range = in_pads->begin(); det_range != in_pads->end(); ++det_range) {
     const GEMDetId& id = (*det_range).first;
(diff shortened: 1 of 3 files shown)


silviodonato:
please test

silviodonato:

silviodonato:
please test

dildick:
A few failures in the comparisons. They are expected. 

Previously, a container in the GEMCoPadProcessor was not cleared in each event, so coincidence pads would incorrectly get duplicated in the sample. In WF 20434.0 one can see the result of the fix. In 10 ttbar events without pileup, there shouldn't too many coincidence pads. In this test, there were just 2, but previously they would get copied 7 times over (first coincidence pads appearing in the second event).

Screen Shot 2020-06-18 at 3 40 22 PM
Screen Shot 2020-06-18 at 3 40 43 PM

dildick:
@rekovic Do you have any further comments?

dildick:
@rekovic Can you sign this PR? There is  bug fix needs to be incorporated. Thanks.

silviodonato:
@rekovic, can you please check and eventually sign?

rekovic:
+1

silviodonato:
+1

rekovic:
@dildick 
Please make a backport of this PR and of https://github.com/cms-sw/cmssw/pull/30564 to 11_1_X.   These known development/fix should be included in the 11_1_0_patch to be used to process L1T for HLT TDR.  Thanks.

PR #30179, opened 2020-06-10
Move creation of vector outside a loop to avoid useless (de)allocations

dpiparo:
This PR is a simple performance optimisation, which moves outside of a hot double loop the creation of a std::vector. The result is more than 1% speedup of the full phase-2 reco. No change in physics performance expected.
See:

 - before: https://dpiparo.web.cern.ch/dpiparo/cgi-bin/igprof-navigator/step3_RAW2DIGI_L1Reco_RECO_RECOSIM_200.pp/63
 - after: https://dpiparo.web.cern.ch/dpiparo/cgi-bin/igprof-navigator/step3_RAW2DIGI_L1Reco_RECO_RECOSIM_200_opt.pp/74


perrotta:
+1
- Technical improvement
- Jenkinst tests pass

silviodonato:
+1

PR #30183, opened 2020-06-10
Reduce boost dependency in IOPool/Streamer

camolezi:
#### PR description:
Replaced boost library use for C++ STL alternatives. A shared_ptr to a std::vector should have similar behavior as a boost::shared_vector.
I also used unique_ptr instead of shared_ptr, since it seems like there's no need to use a shared_ptr.
note:
I've not found references to ELmapDump (so maybe it could be deleted)

#### PR validation:

@vgvassilev @davidlange6

### DIFF
--- a/IOPool/Streamer/bin/CalcAdler32.cpp
+++ b/IOPool/Streamer/bin/CalcAdler32.cpp
@@ -1,7 +1,7 @@
 #include "FWCore/Utilities/interface/Adler32Calculator.h"
 #include "IOPool/Streamer/interface/MsgTools.h"
 
-#include "boost/shared_array.hpp"
+#include <memory>
 
 #include <fstream>
 #include <iostream>
@@ -23,7 +23,7 @@ int main(int argc, char* argv[]) {
   std::ifstream::pos_type size = file.tellg();
   file.seekg(0, std::ios::beg);
 
-  boost::shared_array<char> ptr(new char[1024 * 1024]);
+  auto ptr = std::make_unique<char[]>(1024 * 1024);
   uint32_t a = 1, b = 0;
 
   std::ifstream::pos_type rsize = 0;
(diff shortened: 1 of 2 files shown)


davidlange6:
please test

davidlange6:
@camolezi - I think you can also remove boost from the BuildFile.xml (in bin directory)

davidlange6:
please test


On Jun 11, 2020, at 2:33 PM, cmsbuild <notifications@github.com<mailto:notifications@github.com>> wrote:



Pull request #30183<https://github.com/cms-sw/cmssw/pull/30183> was updated. @makortel<https://github.com/makortel>, @smuzaffar<https://github.com/smuzaffar>, @cmsbuild<https://github.com/cmsbuild>, @Dr15Jones<https://github.com/Dr15Jones> can you please check and sign again.

IOPool/Streamer/bin/CalcAdler32.cpp
```diff
@@ -23,16 +24,16 @@ int main(int argc, char* argv[]) {
   std::ifstream::pos_type size = file.tellg();
   file.seekg(0, std::ios::beg);
 
-  boost::shared_array<char> ptr(new char[1024 * 1024]);
+  std::unique_ptr<std::vector<char>> ptr(std::make_unique<std::vector<char>>(1024 * 1024));
```
makortel, line 27: Based on the usage, an array looks more appropriate (`std::vector::data()` leaves a bit icky feeling). Could you replace it with
```suggestion
  auto ptr = std::make_unique<char[]>(1024 * 1024);
```
?

camolezi: Yeah, it looks cleaner this way. We could also use std::array, but I think we would still need to use the data() method.

makortel:
@cmsbuild, please test

makortel:
+1

silviodonato:
+1

PR #30093, opened 2020-06-03
[DD4hep] Tests Cleanup and 2021 Production Scenario Update

ianna:
#### PR description:

* 2021 production scenario updated
* Use it in the tests
* Simplify comparison for identifying a regular expression
* Anchor a regular expression
* Tests moved to appropriate subsystems
* Integration tests added to run them

#### PR validation:

unit tests and integration tests run by default

#### if this PR is a backport please specify the original PR and why you need to backport that PR:

Before submitting your pull requests, make sure you followed this checklist:
- verify that the PR is really intended for the chosen branch
- verify that changes follow [CMS Naming, Coding, And Style Rules](http://cms-sw.github.io/cms_coding_rules.html)
- verify that the PR passes the basic test procedure suggested in the [CMSSW PR instructions](https://cms-sw.github.io/PRWorkflow.html)

PR #30202, opened 2020-06-11
Have Geometry/DTGeometryBuilder test ignore irrelevant diffs

Dr15Jones:
#### PR description:

When doing the diff, ignore any framework or irrelevant module outputs.

#### PR validation:

The changed unit test passes and no longer reports any differences.

### DIFF
--- a/Geometry/DTGeometryBuilder/test/dtGeometry.log.org
+++ b/Geometry/DTGeometryBuilder/test/dtGeometry.log.org
@@ -1,4 +1,12 @@
+Label '' IGNORING alignment labels ''.
+%MSG
 DTGeometryTest::analyze: 
+DDDetectorESProducer::Produce MUON
+
+
+
+
+
  Geometry node for DTGeom is valid
  I have 0 detTypes
  I have 2720 detUnits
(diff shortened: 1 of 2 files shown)


Dr15Jones:
please test

Dr15Jones:
This unit test is causing a false positive for #30144

ianna:
@Dr15Jones - I think, it was fixed in https://github.com/cms-sw/cmssw/pull/30093

smuzaffar:
@ianna , this is still failing in DEVEL IBs where we have https://github.com/cms-sw/cmssw/pull/30145 from @Dr15Jones

ianna:
@smuzaffar - does DEVEL IB include #30093?

Dr15Jones:
> @smuzaffar - does DEVEL IB include #30093?

Yes, it is. If I understand the change in that pull request, it just upped the number of differences that were being tolerated. My change actually drops the differences to 0. With my change, we could require that no changes are seen and then catch all problems.

ianna:
+1

silviodonato:
+1

fabiocos:
@Dr15Jones @ianna sorry, the solution I had adopted for MTD was to simply drop the MessageLogger statistics printout in the comparison used by the unit test. Do you see any drawback in this approach?

PR #30206, opened 2020-06-12
DAQ: support for FRD file header with file locking

smorovic:
#### PR description:

- raw (FRD) file format for DAQ3 has recently been modified to include metadata in the header, so json files can be removed (due to 2 MB block constraint in huge-page ramdisk). Initially support was added with file broker mode. Now it is added to the file-locking mode which is used in testing and in Hilton.
- Support for the file header is also added to the single-buffer mode.
- Error handling in file locking mode is improved. It also allows unit test to catch problem in case JSON (raw) files can not be found. Since throwing exception is an issue in separate file discovery thread used by input source, error condition is propagated to the main thread instead of throwing directly.

PR includes also two bug fixes:
 - unique_ptr release() calls do not execute destructor of allocated object wrapped with unique_ptr. This created a (small) memory leak and also caused processed files to not be deleted in single-threaded input source mode (which is available, but not used by default in production setup).

 - in the check of non-opened fd value comparision was with 0, it should be compared with -1.

#### PR validation:

Verified to work correctly in DAQ test system in single-buffered/multi-buffered running modes, with file broker/locking, and with/without file header (and combinations).

### DIFF
--- a/EventFilter/Utilities/interface/EvFDaqDirector.h
+++ b/EventFilter/Utilities/interface/EvFDaqDirector.h
@@ -108,7 +108,12 @@ namespace evf {
     void removeFile(unsigned int ls, unsigned int index);
     void removeFile(std::string);
 
-    FileStatus updateFuLock(unsigned int& ls, std::string& nextFile, uint32_t& fsize, uint16_t& rawHeaderSize, uint64_t& lockWaitTime, bool& setExceptionState);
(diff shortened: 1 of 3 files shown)


perrotta:
please test

mommsen:
+1

perrotta:
+1
- Updates relevant for DAQ
- Jenkins tests pass

silviodonato:
+1

PR #30209, opened 2020-06-12
Preselect tracks for electron track isolation computation in reco

guitargeek:
#### PR description:

I had a closer look at the profiling reports linked in https://github.com/cms-sw/cmssw/pull/30179 for Phase II reconstruction with PU 200. One thing I noticed was that the GsfElectronAlgo [spends a lot of time](https://dpiparo.web.cern.ch/dpiparo/cgi-bin/igprof-navigator/step3_RAW2DIGI_L1Reco_RECO_RECOSIM_200_opt.pp/67) computing track isolation variables, amounting to 1.4 % of the total run time (look for `EleTkIsolFromCands::calIsol`):

It turns out that it takes so much time because  `.eta()` is called for each track, and this for each electron. This results in a bit too many calls to `atan2` and `log`. The obvious solution to this (proposed in this PR) would be to cache the track parameters relevant for electron matching in some simple data structure. Furthermore, one can already preselect the tracks which are stored in this cache because many of the selection criteria tor including a track in the isolation don't depend on the electron.

I checked if this works by profiling step3 workflow 20634.21 (Phase II TTBar with PU). The absolute percentages don't exactly match the reports in https://github.com/cms-sw/cmssw/pull/30179, because the workflow I used didn't produce the LowPtGsfElectrons. Still, one can clearly see that `EleTkIsolFromCands::calIsol` goes from dominating the GsfElectronAlgo to a negligible position:

* **Phase II** representative workflow **20634.21 step3** [before](https://rembserj.web.cern.ch/rembserj/cgi-bin/igprof-navigator/wf_20634p21_igreport_perf/174) and [after](https://rembserj.web.cern.ch/rembserj/cgi-bin/igprof-navigator/wf_20634p21_preselect_electron_track_iso_igreport_perf/265) this PR
* **Run 2** representative workflow **10224.0 step3** [before](https://rembserj.web.cern.ch/rembserj/cgi-bin/igprof-navigator/wf_10224p0_igreport_perf/1693) and [after](https://rembserj.web.cern.ch/rembserj/cgi-bin/igprof-navigator/wf_10224p0_preselect_electron_track_iso_igreport_perf/2084) this PR
* For each profiling 100 events were processed
* HLT has its own track isolation producers and therefore no steps with HLT were profiled

Note that with this PR  `EleTkIsolFromCands::calIsol` got renamed to `EleTkIsolFromCands::operator()`.

#### PR validation:

No backport intended.


slava77:
@cmsbuild please test

RecoEgamma/ElectronIdentification/plugins/ElectronHEEPIDValueMapProducer.cc
```diff
@@ -234,21 +232,42 @@ void ElectronHEEPIDValueMapProducer::produce(edm::Event& iEvent, const edm::Even
...
   const auto& candVetos = isAOD ? candVetosAOD_ : candVetosMiniAOD_;
 
+  std::vector<EleTkIsolFromCands::PreselectedTracks> preselectedTracksForIso03;
+  std::vector<EleTkIsolFromCands::PreselectedTracks> preselectedTracksForIso04;
+
+  {
+    int handleNr = 0;
+    for (auto const& handle : getHandles(iEvent, candTokens_)) {
+      if (handle.isValid()) {
+        preselectedTracksForIso03.push_back(trkIsoCalc_.preselectTracks(*handle, candVetos[handleNr]));
```
slava77, line 246: shouldn't there be a check somewhere that there are any electrons to consider?
This seems too heavy for somewhat empty events

RecoEgamma/EgammaElectronAlgos/src/GsfElectronAlgo.cc
```diff
@@ -489,6 +497,10 @@ GsfElectronAlgo::EventData GsfElectronAlgo::beginEvent(edm::Event const& event,
                                                 *endcapRecHits,
                                                 &ecalSeveretyLevelAlgo,
                                                 DetId::Ecal),
+      .preselectedTracksForIso03 = tkIsol03Calc_.preselectTracks(*ctfTracks),
+      .preselectedTracksForIso04 = tkIsol04Calc_.preselectTracks(*ctfTracks),
+      .preselectedTracksForIsoHEEP03 = tkIsolHEEP03Calc_.preselectTracks(*ctfTracks),
+      .preselectedTracksForIsoHEEP04 = tkIsolHEEP04Calc_.preselectTracks(*ctfTracks),
```
slava77, line 503: can this be done conditioned on available electrons?
It seems rather heavy to me to process the full track collection 4 times if there are no electrons in the event

slava77 (commented):
I think that it would be better to run preselectTracks with some knowledge about available electrons.
The current solution is slower in events without electrons

Please check this code execution time also in the HLT context on something run-2 or run-3 like

RecoEgamma/EgammaElectronAlgos/src/GsfElectronAlgo.cc
slava77: passing a list of vz of electrons, which is a required cut (if I got the selections right) maybe already good enough

guitargeek:
Hi @slava77, thanks for your valuable feedback! You are right, the preselection should not have to be done when there are no electrons in the event. Let's see from the bot-checks if I have not done any mistakes so far, and if not I'll tweak a l little more such that there is no obvious overhead for 0 (or few) electrons.

I think probably the good way to deal with this is to pass a track collection reference to the `EleTkIsolFromCands` at construction, and it will then cache the track info if needed "under the hood". Of course this means that I'd have to change `EleTkIsolFromCands` such that it is constructed once per event and not once per producer instance to avoid mandatory `clear()` methods.

guitargeek:
Hello @slava77, I implemented exactly what I said it the previous comment. The interface to `EleTkIsolFromCands` is much cleaner now, an the tracks will only be read when you want to compute the isolation for a given electron.

After the bot tests pass again and you generally fine with this solution I'll do the profiling for Run 2 and Phase 2.

perrotta:
@cmsbuild please test

guitargeek:
Unfortunately I forgot to initialize a member variable which caused trouble with the track isolation in the HEEP ID. I fixed this in the most recent commit.

On the bright side, everything went fine for the track isolation in the GsfElectronProducer which we are more interested in. You can have a look at the igprof reports for a typical Phase I and a typical Phase II workflow (both with PU) in the PR description. Fortunately, the caching does not slow down the Run 2 reco. It even seems to speed it up a tiny bit, but not enough that I would call it an "improvement". The improvement is more for Phase II, in particular when the LowPtGsfElectrons are produced.

slava77:
@cmsbuild please test

slava77:
> Fortunately, the caching does not slow down the Run 2 reco. It even seems to speed it up a tiny bit, but not enough that I would call it an "improvement". The improvement is more for Phase II, in particular when the LowPtGsfElectrons are produced.

thank you for the detailed profiling/investigation.
Do you happen to know if the original submission of this PR was noticeably slower in run2?

guitargeek:
I don't know right now, but I'm also interested to know at which cost this "naive track information caching" comes. The fact that  `.eta()` comes at a cost might eventually mean that we have to do this in other places too, I think, so it's probably good to know!

guitargeek:
For the TTBar Run 2 workflow I chose, the original solution that iterates over all tracks 8 times would have been roughly twice as expensive as the original `EleTkIsolFromCands::calIsol` was (**warning!** link to profiling of outdated commit):
```
https://rembserj.web.cern.ch/rembserj/cgi-bin/igprof-navigator/wf_20634p21_preselect_electron_track_iso_igreport_perf_initial_solution/2653
```
Honestly, I thought that since the cost for ```preselectTracks``` was basically negligible in Phase 2, I thought that the cost of iterating over all tracks would even more negligible so in Run 2 because there are less tracks. Good that  we have a solution now that works for both conditions!

RecoEgamma/EgammaIsolationAlgos/interface/EleTkIsolFromCands.h
```diff
@@ -65,50 +58,84 @@ class EleTkIsolFromCands {
...
+  //
+  class SimpleTrack {
+  public:
+    SimpleTrack(reco::TrackBase const& trk) : pt{trk.pt()}, eta{trk.eta()}, phi{trk.phi()}, vz{trk.vz()} {}
+    const double pt;
+    const double eta;
+    const double phi;
+    const double vz;
+  };
+
```
dpiparo, line 112: Since this is a pr aiming to performance, a comment which maybe is naive: isn't this an opportunity to introduce a simple SOA and a way to access it through some kind of view (e.g. a struct with references as data members)? But I might be reading the code incorrectly..

RecoEgamma/EgammaIsolationAlgos/src/EleTkIsolFromCands.cc
```diff
@@ -135,21 +133,20 @@ EleTkIsolFromCands::PIDVeto EleTkIsolFromCands::pidVetoFromStr(const std::string
...
-
-  return dR2 >= cuts.minDR2 && dR2 <= cuts.maxDR2 && std::abs(dEta) >= cuts.minDEta && std::abs(dZ) < cuts.maxDZ &&
-         trk.hitPattern().numberOfValidHits() >= cuts.minHits &&
+bool EleTkIsolFromCands::passTrackPreselection(const reco::TrackBase& trk, const TrkCuts& cuts) {
+  return trk.hitPattern().numberOfValidHits() >= cuts.minHits &&
          trk.hitPattern().numberOfValidPixelHits() >= cuts.minPixelHits &&
-         (trk.ptError() / trkPt < cuts.maxDPtPt || cuts.maxDPtPt < 0) && passQual(trk, cuts.allowedQualities) &&
-         passAlgo(trk, cuts.algosToReject) && trkPt > cuts.minPt;
+         (trk.ptError() / trk.pt() < cuts.maxDPtPt || cuts.maxDPtPt < 0) && passQual(trk, cuts.allowedQualities) &&
+         passAlgo(trk, cuts.algosToReject) && trk.pt() > cuts.minPt;
```
dpiparo, line 140: I think ::pt() invokes a sqrt which is called 2 times in this formula.

guitargeek: Thanks for spotting this! I'll make sure `::pt()` is only called once.

RecoEgamma/EgammaIsolationAlgos/interface/EleTkIsolFromCands.h
guitargeek: This is definitely an opportunity to introduce this. Such a simple track-SOA could then also be reused when we optimize the other worst offenders in RecoEgamma at a later point (KF track matching and conversion finding), since the underlying problem is basically the same there.

Just to avoid reinventing the wheel here: do you know of existing CMSSW code that does what you describe, so I could use it as inspiration or in the best case even reuse it? I also found the [FWCore/SOA](https://github.com/cms-sw/cmssw/tree/master/FWCore/SOA) package quite fitting. I think it could be nicely applied here.

RecoEgamma/EgammaIsolationAlgos/src/EleTkIsolFromCands.cc
slava77: I expect that repeated calls to an object function are optimized away by the compiler, at least for small enough scope of the code and where it's clear that the object is not changing between the calls

guitargeek: Right, that might happen. But there is also third call outside the function when the SimpleTracks are filled. So to call `pt()` only once, I'll just do it in `preselectTracks` and then pass the value to `passTrackPreselection` and we don't have to rely on the compiler optimization anyway.

guitargeek:
Local matrix tests pass. Please wait a few minutes for profiling reports before starting bot tests, so we know that there is no unexpected surprise when using `FWCore/SOA`. I think it's not used in CMSSW so far, so this is a bit experimental.

guitargeek:
Hi @dpiparo, @slava77, with the new commit I tried out the approach with the structure of arrays, which seems to not change the profiling results:
https://rembserj.web.cern.ch/rembserj/cgi-bin/igprof-navigator/wf_20634p21_preselect_electron_track_iso_igreport_perf_soa/228
Unfortunately the absolute cost fluctuates a bit in the different reports, but the relative position of `EleTkIsolFromCands::operator()` in `createElectron` is the same.

I guess it's sort of expected that SoA does not change much here, because ell the members of the simple track have to be read at each iteration anyway and the actual matching with the electron is not vectorized. Still, I think the move to SoA is valuable in future optimizations, where we might reuse the same track information object also for the other Egamma hotspots.

And it's of course nice to see that FWCore/SOA works so nicely!

RecoEgamma/EgammaIsolationAlgos/interface/EleTkIsolFromCands.h
dpiparo: I ping @Dr15Jones, since he is the person which invented the FWCore/SOA. I think it makes sense in this context.

Dr15Jones: `edm::soa::Table` and `edm::SoATuple` are both generic implementations of SoA. `Table` was designed to work well as an Event data product (in theory could handle being made persistent with a bit of work). `SoATuple` is really meant for internal use and does better memory packing.

RecoEgamma/EgammaIsolationAlgos/interface/EleTkIsolFromCands.h
```diff
@@ -100,40 +103,37 @@ class EleTkIsolFromCands {
...
-  class SimpleTrack {
-  public:
-    SimpleTrack(reco::TrackBase const& trk) : pt{trk.pt()}, eta{trk.eta()}, phi{trk.phi()}, vz{trk.vz()} {}
-    const double pt;
-    const double eta;
-    const double phi;
-    const double vz;
-  };
+
+  SOA_DECLARE_COLUMN(Pt, double, "pt");
```
Dr15Jones, line 107: I would suggest putting this within a namespace.

RecoEgamma/EgammaIsolationAlgos/interface/EleTkIsolFromCands.h
guitargeek: Hi @Dr15Jones thanks for your answer! So with "internal use" you mean use inside the framework implementation details? Does that mean I made the right pick with `edm::soa::Table`?

RecoEgamma/EgammaIsolationAlgos/interface/EleTkIsolFromCands.h
guitargeek: Okay! Then I can't put it in the class, but that's fine because I plan to reuse it in other egamma code anyway. If there are no objections, I'd put it in the namespace `egamma::tracktable`.

RecoEgamma/EgammaIsolationAlgos/interface/EleTkIsolFromCands.h
Dr15Jones: I actually meant internal details of a module or the framework but not really meant to be used as an data product. `edm::soa::Table` will work either as an internal detail or as something to pass between modules via the Event.

guitargeek:
Hello, I mentioned already that there are also similar track-looping hotspots in the RecoEgamma and how the solution will be similar. Now, especially with this recent SoA discussion in the review, the next step in this PR would have been to move the egamma SoA code to a separate namespace and file, which does not make much sense without the context of the other optimizations. Therefore, I also pushed now the optimization for `egamma::getClosestCtfToGsf` such that you have the full picture in the review.

To address concerns by @slava77 that the intermediate data structures are computed unnecessarily when there are no electrons, I introduced a `LazyResult` class to abstract this out of the producers.

Profiling looks good, `egamma::getClosestCtfToGsf` goes from [0.4 %](https://rembserj.web.cern.ch/rembserj/cgi-bin/igprof-navigator/wf_20634p21_preselect_electron_track_iso_igreport_perf_soa/298) to [0.03 %](https://rembserj.web.cern.ch/rembserj/cgi-bin/igprof-navigator/wf_20634p21_preselect_electron_track_iso_igreport_perf_soa_v2/2089). In @dpiparo's reports made with a "fatter" reconstruction that also included the LowPtGsfElectrons, it was at [0.86 %](https://dpiparo.web.cern.ch/dpiparo/cgi-bin/igprof-navigator/step3_RAW2DIGI_L1Reco_RECO_RECOSIM_200_opt.pp/224). So if we assume this to be the correct baseline, this PR would already save **2.5 % runtime**.

If the bot-tests turn out to be fine...

perrotta:
@cmsbuild please test

RecoEgamma/EgammaIsolationAlgos/src/EleTkIsolFromCands.cc
VinInn: and since you are in the process 
```(trk.ptError()  < cuts.maxDPtPt *  trk.pt() ```
and in any case all that code is dominated by this
https://cmssdt.cern.ch/dxr/CMSSW/source/DataFormats/TrackReco/interface/TrackBase.h#697

(and NO the compiler will NOT optimize pt()*pt()* p() * p() / charge() / charge() )
and will recompute again in the line below  pt() * p() / charge()  given that the order is different
(maybe with Ofast, do we like Ofast these days???)

So, if anything optimize the computation of ptError or even better of ptError/pt that is equivalent to deltaP/(1/p) that is what is stored and the only meaningful quantity for tracks

File: RecoEgamma/EgammaIsolationAlgos/interface/EleTkIsolFromCands.h [hunk 6]
```diff
@@ -100,40 +103,37 @@ class EleTkIsolFromCands {
...
+  SOA_DECLARE_COLUMN(Eta, double, "eta");
```
VinInn: in my opinion float will suffice

RecoEgamma/EgammaIsolationAlgos/src/EleTkIsolFromCands.cc
```diff
@@ -91,9 +111,13 @@ EleTkIsolFromCands::Output EleTkIsolFromCands::operator()(const reco::TrackBase&
   auto const& preselectedTracks = getPreselectedTracks(isBarrelElectron);
   auto const& cuts = isBarrelElectron ? cfg_.barrelCuts : cfg_.endcapCuts;
 
-  for (auto& trk : preselectedTracks) {
-    if (passMatchingToElectron(trk, cuts, eleEta, elePhi, eleVz)) {
-      ptSum += trk.pt;
+  for (auto const& trk : preselectedTracks) {
+    const float dR2 = reco::deltaR2(eleEta, elePhi, trk.get<Eta>(), trk.get<Phi>());
+    const float dEta = trk.get<Eta>() - eleEta;
```
VinInn, line 119: typical mess of mixed precision

RecoEgamma/EgammaIsolationAlgos/src/EleTkIsolFromCands.cc
```diff
@@ -133,20 +157,11 @@ EleTkIsolFromCands::PIDVeto EleTkIsolFromCands::pidVetoFromStr(const std::string
...
-}
-
-bool EleTkIsolFromCands::passMatchingToElectron(
-    SimpleTrack const& trk, const TrkCuts& cuts, double eleEta, double elePhi, double eleVZ) {
-  const float dR2 = reco::deltaR2(eleEta, elePhi, trk.eta, trk.phi);
-  const float dEta = trk.eta - eleEta;
-  const float dZ = eleVZ - trk.vz;
-
-  return dR2 >= cuts.minDR2 && dR2 <= cuts.maxDR2 && std::abs(dEta) >= cuts.minDEta && std::abs(dZ) < cuts.maxDZ;
+         (trk.ptError() / trackPt < cuts.maxDPtPt || cuts.maxDPtPt < 0) && passQual(trk, cuts.allowedQualities) &&
```
VinInn, line 166: resort alll those comparison in order of decreasing filtering power balanced by increasing computational complexity
(in practice move ```(trk.ptError() / trackPt < cuts.maxDPtPt || cuts.maxDPtPt < 0)``` to the end AND even this one please ```cuts.maxDPtPt < 0 || trk.ptError() / trackPt < cuts.maxDPtPt`` and then consider just to cut on
trk.qoverpError()*trk.p()  (is not 100% backward compatible, but equivalent and even superior)
(and I never understood why reco::Track does not store what actually fitted....)

RecoEgamma/EgammaIsolationAlgos/src/EleTkIsolFromCands.cc
guitargeek: Thanks for the reminder! I didn't change this so far, because I didn't want to risk even minor precision differences in the output. Would be hard to disentangle from possible mistakes I made in my code. Now that the comparisons showed no reco differences, I'm happy to address this.

RecoEgamma/EgammaIsolationAlgos/interface/EleTkIsolFromCands.h
guitargeek: Thanks for the comment, I'll move all computations to float.

RecoEgamma/EgammaIsolationAlgos/src/EleTkIsolFromCands.cc
guitargeek: Thanks for the comment and sharing your opinion about `ptError()`. Absolutely, I should reorder the checks!

I think for the moment we better stick with ```trk.ptError() / trackPt < cuts.maxDPtPt``` to not risk reconstruction differences in this already large PR, if this is OK for you. But I will definitely keep your suggestion in mind!

Just as a reminder: this loop we are commenting on here is not the hot nNtrack^2 loop that had to be optimized. While I understand that obvious optimizations should be done anyway, like the reordering, I don't necessarily think that the cost of this loop justifies an optimization that alters the reco output at this point.

slava77:
@cmsbuild please test

guitargeek:
This came just to my mind randomly. If we are already optimizing `getClosestCtfTrack`, we should probably not call it redundantly...

slava77:
@cmsbuild please test

RecoEgamma/EgammaElectronAlgos/src/GsfElectronTools.cc
```diff
@@ -15,24 +13,27 @@ namespace egamma {
...
     for (auto ctfTkIter = ctfTrackCollection->begin(); ctfTkIter != ctfTrackCollection->end(); ctfTkIter++, counter++) {
-      double dEta = gsfTrackRef->eta() - ctfTkIter->eta();
-      double dPhi = gsfTrackRef->phi() - ctfTkIter->phi();
-      double pi = acos(-1.);
-      if (std::abs(dPhi) > pi)
-        dPhi = 2 * pi - std::abs(dPhi);
+      float dEta = gsfEta - trackTable.get<egamma::soa::col::Eta>(counter);
+      float dPhi = gsfPhi - trackTable.get<egamma::soa::col::Phi>(counter);
+      if (std::abs(dPhi) > M_PI)
+        dPhi = 2 * M_PI - std::abs(dPhi);
```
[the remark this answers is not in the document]
guitargeek: Of course, thanks

RecoEgamma/EgammaIsolationAlgos/interface/EleTkIsolFromCands.h
```diff
@@ -65,50 +59,77 @@ class EleTkIsolFromCands {
...
+  // tracks (Phase II conditions). In particular, calling
+  // reco::TrackBase::eta() many times is costy because eta is not precomputed.
+  // To solve this, we first cache the tracks in a simpler data structure in
+  // which eta is already computed (TrackTable). Furthermore, the tracks are
+  // preselected by the cuts that can already be applied without considering
+  // the electron. Note that this has to be done twice, because the required
+  // preselection is different for barrel and endcap electrons.
+
+  using TrackTable =
+      edm::soa::Table<egamma::soa::col::Pt, egamma::soa::col::Eta, egamma::soa::col::Phi, egamma::soa::col::Vz>;
```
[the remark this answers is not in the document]
guitargeek: `using` statements within `class` are unfortunately not possible, and using it directly in the header file should not be done. But I'll try to find another way to make this less verbose.

RecoEgamma/EgammaTools/interface/LazyResult.h
```diff
@@ -0,0 +1,76 @@
...
+*/
+//
+// Original Author:  Jonas Rembser
+//         Created:  Mon, 14 Jun 2020 00:00:00 GMT
+//
+
+#include <tuple>
+#include <type_traits>
+
+template <class Func, class... Args>
```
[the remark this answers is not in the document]
guitargeek: Right, this should be done

RecoEgamma/EgammaTools/interface/tables.h
```diff
@@ -0,0 +1,41 @@
+#ifndef RecoEgamma_EgammaTools_tables_h
+#define RecoEgamma_EgammaTools_tables_h
+
+#include "FWCore/SOA/interface/Column.h"
+#include "FWCore/SOA/interface/Table.h"
+#include "RecoEgamma/EgammaTools/interface/LazyResult.h"
+
+#include <vector>
+
+namespace egamma {
```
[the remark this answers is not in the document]
guitargeek: Okay, I agree about the `cms` part, but what about the `soa` part of the namespace? I feel that this is something that might also be decided to call differently, like `table` or `tables` or even `df` for "data frame" like in pandas, which is probably the most famous SoA example that everybody knows (okay, two letters might be a bit short however). My main argument for not using `soa` is that it's a very technical term and people might not immediately understand what it meant when reading the code.

My personal preference would be `cms::table::col` (all in singular). Let me know how you feel about this, if you don't have a strong preference I'll change `soa` to `table` in the next commit besides your recommendation of changing `egm` to `cms`.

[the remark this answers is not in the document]
slava77: ```soa``` part is fine.
looking up in the discussion, ```edm::soa``` may even be better than ```cms::soa```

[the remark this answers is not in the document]
guitargeek: Thanks for your answer, it will be `edm::soa::col` if I understood you correctly. I'm also fine with that.

RecoEgamma/EgammaTools/interface/tables.h
```diff
@@ -0,0 +1,41 @@
...
+  namespace soa {
+
+    namespace col {
+
+      SOA_DECLARE_COLUMN(Pt, float, "pt");
+      SOA_DECLARE_COLUMN(Eta, float, "eta");
+      SOA_DECLARE_COLUMN(Phi, float, "phi");
+      SOA_DECLARE_COLUMN(Vz, float, "vz");
+
+    }  // namespace col
```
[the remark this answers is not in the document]
guitargeek: Couldn't address all comments in one go, I suddenly had connection problems.

Yeah I was also thinking about this. I think you are right about splitting the columns from the generic tables, after all we already have `EleTkIsolFromCands` which is using the columns but not the table defined here.

However, I think anticipating different column categories is a bit premature at this point, and even then I'm probably against splitting it up, mainly to avoid collisions in `egm::soa::col` (current name). Similar for the tables: we might have tables building on each other with for example the `AddColumns_t` mechanism [1]. Having different tables in different files that are connected by added and subtracted columns would be also complicated to oversee.

Would you  be fine with one file for the columns, and one for the tables in `CommonTools/Utils`, while keeping more generic names? I understand and would share the feeling that simply `tables.h` and `columns.h` sounds too generic, would you prefer `SoATables.h` and `SoAColumns.h`?

[1] https://github.com/cms-sw/cmssw/blob/master/FWCore/SOA/test/table_t.cppunit.cpp#L69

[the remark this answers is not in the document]
slava77: I'd like to avoid a setup where a single header is expected to describe everything for a concept of SOA.
So, some categorization from the start (like ```Kinematic*```) would make more sense to me.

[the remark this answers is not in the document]
guitargeek: Okay!

RecoEgamma/EgammaTools/interface/tables.h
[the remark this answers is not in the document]
guitargeek: By the way, when is the `cms` namespace preferred over `edm`?

RecoEgamma/EgammaIsolationAlgos/interface/EleTkIsolFromCands.h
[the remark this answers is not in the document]
guitargeek: Well, since `edm::soa::col` is a bit shorter than `egamma::soa::col` this is now formatted on one line. Is this acceptable?

guitargeek:
Hi, I had to force push because the IB on which I based my developments on had expired. The only thing that has actually changed is the addition of the latest commit, which addresses the comments made by @slava77.

perrotta:
@cmsbuild please test

RecoEgamma/EgammaTools/interface/tables.h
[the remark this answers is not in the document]
slava77: to me this was by precedence; there is already edm::soa

Dr15Jones:
One point to make about Columns. You really want to be able to share Columns across classes as you can define functions like `momentum(edm::soa::View<Px,Py,Pz>)` and then any `edm::soa::Table` that has those three Columns (in any order) can be passed to that function to do the calculation.

guitargeek:
> One point to make about Columns. You really want to be able to share Columns across classes as you can define functions like `momentum(edm::soa::View<Px,Py,Pz>)` and then any `edm::soa::Table` that has those three Columns (in any order) can be passed to that function to do the calculation.

Thanks a lot @Dr15Jones for chiming in! To my understanding, your comment implied that `edm::soa::TableView` should be preferred over `edm::soa::Table` in function arguments, so that's what I did in the latest commit to set a good precedent. And now this PR also covers functions of `edm::soa::TableView` which are not covered by its unit tests, so I had to make a small fix in `TableView.h`.

slava77:
@cmsbuild please test

makortel:
+core

slava77:
+1

for https://github.com/cms-sw/cmssw/pull/30209	ec0a7ac 
- a local test for timing purpose using a run2 workflow 136.888 shows some decrease in \*ElectronCoreProducer calls (by a factor of 3-5) and in \*ElectronProducer (by about 15-20%), which adds up to about 1% of the reco time. The benefit should be much larger in the PU200 case.

santocch:
+1

silviodonato:
+1

PR #30213, opened 2020-06-12
HCAL - HB TDC digi default to non-packed

lwang046:
This is HCAL DPG request.

The previous default TDC value was packed/compressed, forcing people to make separate MC production requests with customized commands to unpack the value for hcal TDC studies. By changing this default behavior all centrally produced samples using this release and above will be available for hcal TDC studies without having to ask for separate production. 

Since currently TDC value is not used anywhere else in the cmssw, this change won't affect anything else.

Would also like to request to back port this setting to 11_0_X for the same reason.

### DIFF
--- a/EventFilter/HcalRawToDigi/plugins/HcalDigiToRawuHTR.cc
+++ b/EventFilter/HcalRawToDigi/plugins/HcalDigiToRawuHTR.cc
@@ -285,7 +285,7 @@ void HcalDigiToRawuHTR::fillDescriptions(edm::ConfigurationDescriptions& descrip
   desc.addUntracked<int>("Verbosity", 0);
   desc.add<int>("tdc1", 4);
   desc.add<int>("tdc2", 20);
-  desc.add<bool>("packHBTDC", true);
+  desc.add<bool>("packHBTDC", false);
   desc.add<std::string>("ElectronicsMap", "");
   desc.add<edm::InputTag>("QIE10", edm::InputTag("simHcalDigis", "HFQIE10DigiCollection"));
   desc.add<edm::InputTag>("QIE11", edm::InputTag("simHcalDigis", "HBHEQIE11DigiCollection"));


perrotta:
please test

perrotta:
A few DQM entries for HCAL show up as empty now in the comparisons, e.g.

HcalDqm

Should DQM get adapted to read non packed inputs?
Besides that, what else should one expect to find in the ouputs? I would naively expect that root outputs when non packed values are stored should end up with some small but not vanishing difference in the data size, but they don't...

lwang046:
This nonpacked feature is only available in MC, the real data will still have packed tdc values due to hardware limitations. I think we don't need to modify DQM unless there is a way to differentiate MC and data in DQM? 

The unpacked tdc reuses 4 bits that was used to store capid and error bits in real data taken, where in MC they are not needed anyway. That's why no extra size is introduced in the output file size.

perrotta:
Thank you @lwang046 for your explanations

Let add @jfernan2 to make DQM aware of the implications of this PR (and maybe also provide possible suggestions for making those HCAL DQM plots also available for MC)

jfernan2:
There is a way to diferenciate in DQM between Data and MC, but not the necessity if HCAL is happy with the monitoring done at present.

Current DQM differences in MC for HCAL are the following w.r.t. Data:
https://github.com/cms-sw/cmssw/blob/master/DQMOffline/Configuration/python/DQMOfflineMC_cff.py#L6-L7

So, some flagging could be set in order to run the DQM module in question if you want to distinguish between data and MC

lwang046:
Thank you @jfernan2 for the nice example, I can see how it could be done. Indeed as for the necessity I'm not sure what hcal can gain from monitoring tdc in MC since up to now it's still used in very specific study. Maybe @abdoulline can tell more about this.

abdoulline:
Gentlemen,  
in principle, for MC we're permanently doing TDC monitoring by analyzing (privately) the samples.
But right now I'd propose (to move on with this PR):   
I'll ask (or I'll do it myself) to add several TDC-related histos to Validation/HcalDigis (existing MC-specific Validation package) next week.

perrotta:
Thank you @lwang046 @jfernan2 @abdoulline 

My intention was making aware the interested people of this lack of DQM monitoring for MC as a xonsequence to this PR. If both HCAL and DQM can stay as such, it, it will not be RECO that insisits for having it.
 
I think that the proposal from Salavat of adjusting the validation package is worth pursuing: just let me know if you plan to do it in the context of this PR, or if you'll submit a separate PR once ready.

abdoulline:
> Thank you @lwang046 @jfernan2 @abdoulline
> 
> My intention was making aware the interested people of this lack of DQM monitoring for MC as a xonsequence to this PR. If both HCAL and DQM can stay as such, it, it will not be RECO that insisits for having it.
> 
> I think that the proposal from Salavat of adjusting the validation package is worth pursuing: just let me know if you plan to do it in the context of this PR, or if you'll submit a separate PR once ready.


Andrea, we'll submit a separate PR (adding MC TDC plots)  next week.

perrotta:
+1
- Non packed values cannot be read with current DQM, which is not considered a big issue for HCAL validation, but it will be fixed anyhow with a  dedicated PR

silviodonato:
+1

PR #30214, opened 2020-06-12
Generate 2021 Scenario Automatically

ianna:
#### PR description:

Similar to 2026 scenarios, 2021 scenario is autogenerated by a script. 

```
python ./Configuration/Geometry/scripts/generate2021Geometry.py
```

@fabiocos , could you, please, check if eras and modifies are correctly defined

- [x] produce a dd4hep 2021 scenario from the same script

- [x] add PPS in the CTPPS section

- [x] add automated validation to check if the 2021 scenario was modified by hand


kpedro88:
@ianna this is great. Since there are only some minor changes in generate2021Geometry.py compared to generate2026Geometry.py, maybe it's worth trying to unify these scripts? I can look at this if you want.

Also, it might be nice to update Configuration/Geometry/README.md with info about the 2021 geometry.

ianna:
> @ianna this is great. Since there are only some minor changes in generate2021Geometry.py compared to generate2026Geometry.py, maybe it's worth trying to unify these scripts? I can look at this if you want.
> 
> Also, it might be nice to update Configuration/Geometry/README.md with info about the 2021 geometry.

I'd appreciate if you take it from my hands. Thanks! 
There is a catch though - 2021 dd4hep scenario is slightly different :-)

ianna:
@kpedro88 - also, if you'd like to keep "D" in the file name, there will be wide-spread changes in configurations across the whole CMSSW :-(

kpedro88:
We can make the "D" depend on the 2021 vs. 2026 settings.

kpedro88:
Indeed, part of my motivation is to be able to reuse the 2021 dd4hep setup for 2026

cvuosalo:
@ianna @mundim How does this PR affect the work Luiz is doing to include the PPS simulation geometry?

cvuosalo:
@cmsbuild please test

mundim:
> @ianna @mundim How does this PR affect the work Luiz is doing to include the PPS simulation geometry?

Hi.. I updated the PR  (#29916) with the new cmsExtendedGeometry2021XML_cfi. As i understood from our discussion, I should prepare the file with the PPS geometry. I hope I have not caused problem when changing the file above.
cheers

ianna:
> @ianna @mundim How does this PR affect the work Luiz is doing to include the PPS simulation geometry?

@cvuosalo - This PR makes it safer to modify the 2021 scenario. After it is integrated, Luiz can modify the ctpps section in the 2021 dictionary and re-generate the scenario. It will also assure that the 2021 scenario for DD is identical to DD4hep - I understand, that @kpedro88 is working on it.

Configuration/Geometry/python/dict2021Geometry.py
```diff
@@ -0,0 +1,415 @@
...
+    }
+}
+
+ctppsDict = {
+    "abbrev" : "P",
+    "name" : "ctpps",
+    "default" : 1,
+    "P1" : {
+        1 : [
+            'Geometry/VeryForwardData/data/pps/2021/v1/pps.xml',
```
ianna, line 395: @cvuosalo - this is the section for Luis to fill in.

ianna (commented):
@cvuosalo - please, not that before this PR the 2021 scenarios differ between DD and DD4hep.

civanch:
@ianna , thanks for that PR! I would propose merge this asap. @mundim , I would propose you to make a small PR only for PPS geometry, in which fill the place proposed by Ianna.  It is also recommended where possible use "pps" not "ctpps".

Configuration/Geometry/python/dict2021Geometry.py
mundim: Hi.. I 'm a bit confused, where should I put the pps geometry files? I thoough I would work on the Geometry/CMSCommon/python/cmsExtendedGeometry2021XML_cfi.py
So, where should I put the XML files? Or should I put in both files?
Another question is that as I told in the other PR, I could not make the single pps.xml file to work, only when including all of them in the cmsExtendedGeometry2021XML_cfi.py. Should I test again using the dictionary above?
Thanks.

fabiocos:
@ianna nice step forward, I have a question ( @kpedro88 ): for 2021 we are regularly using just ```Run3``` era, do you have a special reason to have the breakdown you have committed? I believe this is not very relevant in practice for the xml list, but it might affect the Reco configuration

fabiocos:
@ianna concerning your second question, do you meand adding an option to the script to dump optionally the same configuration for DD4hep?

fabiocos:
In general I agree with @kpedro88 that ideally it would be preferable to avoid duplications, although at the end of the day will likely mean to add a number of ```if``` statements here and there in the original script (according to the experience with ```ConfigBuilder```)

Configuration/Geometry/python/dict2021Geometry.py
ianna: @mundim - as soon as this PR is merged you will not be able to modify the `cmsExtendedGeometry2021XML_cfi.py` by hand. Instead, you'd have to put all the xml files in this dictionary replacing the line 395, then generating the `cmsExtendedGeometry2021XML_cfi.py` by
```
python ./Configuration/Geometry/scripts/generate2021Geometry.py
```

ianna:
> @ianna concerning your second question, do you meand adding an option to the script to dump optionally the same configuration for DD4hep?

Yes, that's the idea - we have one dictionary and generate both the DD and the DD4hep configurations from it. Also, it would be useful to extend it to cover the legacy scenarios as well as the material variations scenarios. The latter would cover the 2021 zero material scenario by simply configuring current 2021 configuration with the 2015 Tracker zero material configuration, I think. There is a pending request from an E-gamma group.

civanch:
@ianna , it seems to me for each occasion we should use "pps" instead "ctpps" in comments, names and new file names. It is not possible to replace all old files but we need introduce this in new files.

ianna:
> @ianna , it seems to me for each occasion we should use "pps" instead "ctpps" in comments, names and new file names. It is not possible to replace all old files but we need introduce this in new files.

no problem, I'll change it

ianna:
> @ianna , thanks for that PR! I would propose merge this asap. @mundim , I would propose you to make a small PR only for PPS geometry, in which fill the place proposed by Ianna. It is also recommended where possible use "pps" not "ctpps".

done

ianna:
@cvuosalo, @kpedro88 and @civanch - I've updated this PR as suggested. The script generates both DD and DD4hep 2021 scenarios. Please, integrate it in the IB asap so that @mundim could update the PPS section. I think, the dictionary eras can be updated at a later stage.

ianna:
please test

ianna:
please test

fabiocos:
@ianna if the purpose is to generate a DD4hep full configuration, the ```dd4hep``` process modifier should also be activated

cvuosalo:
+1

kpedro88:
+upgrade
I'll submit a followup PR to unify the scripts, haven't had time yet

silviodonato:
+1

PR #30217, opened 2020-06-12
HCAL - HB TDC digi default to non-packed backport to 11_0_X

lwang046:
Backporting to 11_0_X:

This is HCAL DPG request. The previous default TDC value was packed/compressed, forcing people to make separate MC production requests with customized commands to unpack the value for hcal TDC studies. By changing this default behavior all centrally produced samples using this release and above will be available for hcal TDC studies without having to ask for separate production.

Since currently TDC value is not used anywhere else in the cmssw, this change won't affect anything else.

### DIFF
--- a/EventFilter/HcalRawToDigi/plugins/HcalDigiToRawuHTR.cc
+++ b/EventFilter/HcalRawToDigi/plugins/HcalDigiToRawuHTR.cc
@@ -285,7 +285,7 @@ void HcalDigiToRawuHTR::fillDescriptions(edm::ConfigurationDescriptions& descrip
   desc.addUntracked<int>("Verbosity", 0);
   desc.add<int>("tdc1", 4);
   desc.add<int>("tdc2", 20);
-  desc.add<bool>("packHBTDC", true);
+  desc.add<bool>("packHBTDC", false);
   desc.add<std::string>("ElectronicsMap", "");
   desc.add<edm::InputTag>("QIE10", edm::InputTag("simHcalDigis", "HFQIE10DigiCollection"));
   desc.add<edm::InputTag>("QIE11", edm::InputTag("simHcalDigis", "HBHEQIE11DigiCollection"));


perrotta:
@lwang046 @abdoulline
What is the purpose for having this PR backported in 11_1?

perrotta:
backport of #30213

lwang046:
> @lwang046 @abdoulline
> What is the purpose for having this PR backported in 11_1?

Currently the centrally produced MC samples we would like to use are produced in this backported release, it would be great if we could take advantage of these samples with this feature

perrotta:
please test

perrotta:
+1
- Non packed values cannot be read with current DQM, which is not considered a big issue for HCAL validation

perrotta:
hold

perrotta:
I'm realizing only after having signed that this backport is for 11_0, not 11_1.
Please confirm that you really want to submit it in that basically closed cycle, and not in 11_1

lwang046:
Hi @perrotta , yeah that's the decision from hcal dpg group, to be back ported into 11_0_X, not 11_1_X

perrotta:
unhold

perrotta:
@silviodonato please notice that this PR will not satisfy the "no change" requirement for closed cycles.
However, HCAL TDC data are only relevant for HCAL local developments and validations: then such a requirement can probably get relaxed here.

silviodonato:
+1

slava77:
> Hi @perrotta , yeah that's the decision from hcal dpg group, to be back ported into 11_0_X, not 11_1_X

I was naively thinking that the next time we may do a global run, in a month or so we will be using 11_1_X

slava77:
> > Hi @perrotta , yeah that's the decision from hcal dpg group, to be back ported into 11_0_X, not 11_1_X
> 
> I was naively thinking that the next time we may do a global run, in a month or so we will be using 11_1_X

OH, but then this is MC, but even there the new Run3 largish samples will likely be in 11_1_X.
So, still, a decision not to backport to 11_1 is rather odd

abdoulline:
@perrotta,  @slava77

At DPG meeting last Friday the idea was about modifying 11_0_X because many LLP-physics GEN-SIM samples (to be recycled) were produced there.
But Slava is right, we may need to try yet another chance with 11_1_X, given long turnover of non-mainstream (our special LLP samples) MC production...

PR #30232, opened 2020-06-14
[GEOMETRY] Fixes needed for UBSAN

smuzaffar:
#### PR description:

Use a smaller data type to represent these small numbers. This should fix the UBSAN build errors

```
Geometry/EcalAlgo/test/EcalPreshowerParameterDump.cc:49:23: error: '%d' directive writing between 1 and 10 bytes into a region of size between 6 and 15 [-Werror=format-overflow=]
          sprintf(name, "hist%d%d", iz, lay);
```

#### PR validation:

Successfully built for UBSAN and normal IBs

### DIFF
--- a/Geometry/EcalAlgo/test/EcalPreshowerParameterDump.cc
+++ b/Geometry/EcalAlgo/test/EcalPreshowerParameterDump.cc
@@ -17,6 +17,7 @@
 
 #include <iomanip>
 #include <iostream>
+#include <sstream> // for ostringstream
 
 typedef EZArrayFL<GlobalPoint> CornersVec;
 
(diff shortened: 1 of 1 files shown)


smuzaffar:
please test

Geometry/EcalAlgo/test/EcalPreshowerParameterDump.cc
```diff
@@ -42,9 +42,9 @@ EcalPreshowerCellParameterDump::EcalPreshowerCellParameterDump(const edm::Parame
 
   if (debug_) {
     edm::Service<TFileService> fs;
-    for (int iz = 0; iz < 2; ++iz) {
-      int zside = 2 * iz - 1;
-      for (int lay = 1; lay <= 2; ++lay) {
+    for (short iz = 0; iz < 2; ++iz) {
+      short zside = 2 * iz - 1;
+      for (short lay = 1; lay <= 2; ++lay) {
```
ianna, line 47: @smuzaffar - `sprintf` is not thread safe - it should be replaced with` std::ostringstream`:

```c++
#include <sstream> // for ostringstream

std::ostringstream out;

out << ”hist” << iz << lay;
```

ianna (changes requested):

Geometry/EcalAlgo/test/EcalPreshowerParameterDump.cc
ianna: Or even better to `std::format` when we switch to C++20:
```c++
std::string message = std::format("The answer is {}.", 42);
```
https://en.cppreference.com/w/cpp/utility/format

smuzaffar: done , used `ostringstream`

smuzaffar:
please test

smuzaffar:
please test

cvuosalo:
+1

silviodonato:
+1

PR #30253, opened 2020-06-15
Add muon ID selectors for packed candidate collections

stahlleiton:
#### PR description:

This PR adds muon ID selectors (isTrackerMuon and TMOneStationTight), as PackedCandidateRefVectors,  associated the packedPFCandidates and lostTracks collections.

These are part of the HIN developments for MiniAOD event content.

Running over ~1000 events from the HIDoubleMuon PD of 2018 PbPb data, the average compressed size of the output collections (PackedCandidateRefVectors) were ~9 bytes/event for packedPFCandidates and ~6 bytes/event for lostTracks.

#### PR validation:

Tested with workflow 159

@mandrenguyen

### DIFF
--- a/PhysicsTools/PatAlgos/plugins/PackedCandidateMuonSelectorProducer.cc
+++ b/PhysicsTools/PatAlgos/plugins/PackedCandidateMuonSelectorProducer.cc
@@ -11,29 +11,27 @@
 namespace pat {
 
   class PackedCandidateMuonSelectorProducer : public edm::stream::EDProducer<> {
-
   public:
     explicit PackedCandidateMuonSelectorProducer(const edm::ParameterSet& iConfig)
         : muonToken_(consumes<reco::MuonCollection>(iConfig.getParameter<edm::InputTag>("muons"))),
-          candidateToken_(
-              consumes<pat::PackedCandidateCollection>(iConfig.getParameter<edm::InputTag>("candidates"))),
-          candidate2PFToken_(
-              consumes<edm::Association<reco::PFCandidateCollection> >(iConfig.getParameter<edm::InputTag>("candidates"))),
-          track2LostTrackToken_(
-              consumes<edm::Association<pat::PackedCandidateCollection> >(iConfig.getParameter<edm::InputTag>("lostTracks"))),
+          candidateToken_(consumes<pat::PackedCandidateCollection>(iConfig.getParameter<edm::InputTag>("candidates"))),
+          candidate2PFToken_(consumes<edm::Association<reco::PFCandidateCollection> >(
+              iConfig.getParameter<edm::InputTag>("candidates"))),
+          track2LostTrackToken_(consumes<edm::Association<pat::PackedCandidateCollection> >(
+              iConfig.getParameter<edm::InputTag>("lostTracks"))),
           muonSelectors_(iConfig.getParameter<std::vector<std::string> >("muonSelectors")),
           muonIDs_(iConfig.getParameter<std::vector<std::string> >("muonIDs")) {
       for (const auto& sel : muonSelectors_) {
-        produces<pat::PackedCandidateRefVector>("lostTracks"+sel);
-        produces<pat::PackedCandidateRefVector>("pfCandidates"+sel);
+        produces<pat::PackedCandidateRefVector>("lostTracks" + sel);
+        produces<pat::PackedCandidateRefVector>("pfCandidates" + sel);
       }
       for (const auto& sel : muonIDs_) {
         muonIDMap_[sel].reset(new StringCutObjectSelector<reco::Muon>("passed('" + sel + "')"));
-        produces<pat::PackedCandidateRefVector>("lostTracks"+sel);
-        produces<pat::PackedCandidateRefVector>("pfCandidates"+sel);
+        produces<pat::PackedCandidateRefVector>("lostTracks" + sel);
+        produces<pat::PackedCandidateRefVector>("pfCandidates" + sel);
       }
     }
-    ~PackedCandidateMuonSelectorProducer() = default;
+    ~PackedCandidateMuonSelectorProducer() override = default;
 
     void produce(edm::Event&, const edm::EventSetup&) override;
 
@@ -79,10 +77,12 @@ void pat::PackedCandidateMuonSelectorProducer::produce(edm::Event& iEvent, const
     const auto& lostTrack = track2LostTrack[muonTrack];
     if (lostTrack.isNonnull()) {
       for (const auto& sel : muonSelectors_) {
-        if (muon::isGoodMuon(muon, muon::selectionTypeFromString(sel))) lostTrackMap[sel]->push_back(lostTrack);
+        if (muon::isGoodMuon(muon, muon::selectionTypeFromString(sel)))
+          lostTrackMap[sel]->push_back(lostTrack);
       }
       for (const auto& sel : muonIDs_) {
-        if ((*muonIDMap_.at(sel))(muon)) lostTrackMap[sel]->push_back(lostTrack);
+        if ((*muonIDMap_.at(sel))(muon))
+          lostTrackMap[sel]->push_back(lostTrack);
       }
       continue;
     }
@@ -94,10 +94,12 @@ void pat::PackedCandidateMuonSelectorProducer::produce(edm::Event& iEvent, const
       // check if candidate and muon are compatible
       if (candTrack.isNonnull() && muonTrack.id() == candTrack.id()) {
         for (const auto& sel : muonSelectors_) {
-          if (muon::isGoodMuon(muon, muon::selectionTypeFromString(sel))) candMap[sel]->push_back(cand);
+          if (muon::isGoodMuon(muon, muon::selectionTypeFromString(sel)))
+            candMap[sel]->push_back(cand);
         }
         for (const auto& sel : muonIDs_) {
-          if ((*muonIDMap_.at(sel))(muon)) candMap[sel]->push_back(cand);
+          if ((*muonIDMap_.at(sel))(muon))
+            candMap[sel]->push_back(cand);
         }
         break;
       }
@@ -105,10 +107,10 @@ void pat::PackedCandidateMuonSelectorProducer::produce(edm::Event& iEvent, const
   }
 
   for (auto& s : lostTrackMap) {
-    iEvent.put(std::move(s.second), "lostTracks"+s.first);
+    iEvent.put(std::move(s.second), "lostTracks" + s.first);
   }
   for (auto& s : candMap) {
-    iEvent.put(std::move(s.second), "pfCandidates"+s.first);
+    iEvent.put(std::move(s.second), "pfCandidates" + s.first);
   }
 }
 
@@ -118,9 +120,9 @@ void pat::PackedCandidateMuonSelectorProducer::fillDescriptions(edm::Configurati
   desc.add<edm::InputTag>("muons", edm::InputTag("muons"))->setComment("muon input collection");
   desc.add<edm::InputTag>("candidates", edm::InputTag("packedPFCandidates"))
       ->setComment("packed PF candidate input collection");
-  desc.add<edm::InputTag>("lostTracks", edm::InputTag("lostTracks"))
-      ->setComment("lost track input collection");
-  desc.add<std::vector<std::string> >("muonSelectors", {"AllTrackerMuons", "TMOneStationTight"})->setComment("muon selectors");
+  desc.add<edm::InputTag>("lostTracks", edm::InputTag("lostTracks"))->setComment("lost track input collection");
+  desc.add<std::vector<std::string> >("muonSelectors", {"AllTrackerMuons", "TMOneStationTight"})
+      ->setComment("muon selectors");
   desc.add<std::vector<std::string> >("muonIDs", {})->setComment("muon IDs");
   descriptions.add("packedCandidateMuonID", desc);
 }
--- a/PhysicsTools/PatAlgos/plugins/PackedCandidateMuonSelectorProducer.cc
+++ b/PhysicsTools/PatAlgos/plugins/PackedCandidateMuonSelectorProducer.cc
@@ -3,7 +3,6 @@
 #include "FWCore/ParameterSet/interface/ParameterSet.h"
 #include "FWCore/ParameterSet/interface/ConfigurationDescriptions.h"
 #include "FWCore/Utilities/interface/InputTag.h"
-#include "DataFormats/Common/interface/ValueMap.h"
 #include "DataFormats/PatCandidates/interface/Muon.h"
 #include "DataFormats/PatCandidates/interface/PackedCandidate.h"
 #include "DataFormats/MuonReco/interface/MuonSelectors.h"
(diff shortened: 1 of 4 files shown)


PhysicsTools/PatAlgos/plugins/PackedCandidateMuonSelectorProducer.cc
```diff
@@ -0,0 +1,119 @@
...
+    for (size_t i = 0; i < nCand; i++) {
+      const auto& cand = candidates->refAt(i);
+      // ignore neutral candidates or without track
+      if (cand->charge() == 0 || !cand->hasTrackDetails())
+        continue;
+
+      // check if candidate and muon are compatible
+      const auto& candTrack = cand->pseudoTrack();
+      if (muonTrack.charge() == candTrack.charge() && muonTrack.numberOfValidHits() == candTrack.numberOfValidHits() &&
+          std::abs(muonTrack.eta() - candTrack.eta()) < 1E-3 && std::abs(muonTrack.phi() - candTrack.phi()) < 1E-3 &&
```
jpata, line 82: Is this phi subtraction safe in case of wraparound? I think one might also use a `deltaR` method from the framework: https://github.com/cms-sw/cmssw/blob/master/DataFormats/Math/interface/deltaR.h
In addition, perhaps it can be considered to put these cutoffs 1E-3, 1E-2 at the top of the file as a static const, or otherwise as a configurable parameter in a visible spot?

santocch:
please test

PhysicsTools/PatAlgos/plugins/PackedCandidateMuonSelectorProducer.cc
stahlleiton: Thanks for the suggestions. I replaced the deltaPhi and deltaEta for a deltaR, and changed the cutoffs to be configurable parameters.

perrotta:
@cmsbuild please test

PhysicsTools/PatAlgos/python/slimming/slimming_cff.py
```diff
@@ -62,6 +62,12 @@
 from Configuration.Eras.Modifier_pp_on_AA_2018_cff import pp_on_AA_2018
 pp_on_AA_2018.toReplaceWith(slimmingTask, slimmingTask.copyAndExclude([slimmedOOTPhotons]))
 
+from Configuration.Eras.Modifier_pp_on_PbPb_run3_cff import pp_on_PbPb_run3
+from PhysicsTools.PatAlgos.packedPFCandidateMuonID_cfi import packedPFCandidateMuonID
+(pp_on_AA_2018 | pp_on_PbPb_run3).toModify(packedPFCandidateMuonID, muonSelectors = cms.vstring(["AllTrackerMuons", "TMOneStationTight"]))
+lostTrackMuonID = packedPFCandidateMuonID.clone(candidates = cms.InputTag("lostTracks"))
```
slava77, line 68:
```suggestion
lostTrackMuonID = packedPFCandidateMuonID.clone(candidates = "lostTracks")
```

PhysicsTools/PatAlgos/python/slimming/MicroEventContent_cff.py
```diff
@@ -171,3 +171,11 @@
...
 from Configuration.Eras.Modifier_phase2_timing_cff import phase2_timing
 phase2_timing.toModify(MicroEventContentMC, outputCommands = MicroEventContentMC.outputCommands + _phase2_timing_extraCommands)
+
+_pp_on_AA_extraCommands = [
+    'keep booledmValueMap_*MuonID_*_*',
+]
+from Configuration.Eras.Modifier_pp_on_AA_2018_cff import pp_on_AA_2018
+from Configuration.Eras.Modifier_pp_on_PbPb_run3_cff import pp_on_PbPb_run3
+(pp_on_AA_2018 | pp_on_PbPb_run3).toModify(MicroEventContent, outputCommands = MicroEventContent.outputCommands + _pp_on_AA_extraCommands)
+(pp_on_AA_2018 | pp_on_PbPb_run3).toModify(MicroEventContentMC, outputCommands = MicroEventContentMC.outputCommands + _pp_on_AA_extraCommands + ['keep *_heavyIon_*_*'])
```
slava77, line 181: - what's heavyIons ? this seems unrelated to this PR
- the common modifications of MicroEventContent should be done before MicroEventContentMC is made out of it on L136
- as commented in the review, it seems like these valuMaps should not be saved (they can not be used as saved)

PhysicsTools/PatAlgos/python/slimming/slimming_cff.py
slava77, line 68: this is a bit too low-level and not related to slimming directly.
PhysicsTools/PatAlgos/python/slimming/miniAOD_tools.py is apparently a more appropriate place to add these ID producers

slava77 (commented):
It looks like the muon ID selectors are not saved in a way usually expected in miniAOD.

The kind of matching implemented in the PR is rather expensive and is not obviously the best possible.
Both lostTracks and packedPFCandidates have associations saved to the original tracks.
The underlying reco::Muons ```muons``` have the original track refs and at that level the track matching can be done directly.
Do the selectors computed here have to be done on pat::Muons?

The current proposal is apparently to save the valueMaps computed per patMuons directly, while the patMuons themselves are not saved. This will likely render these maps useless.
The standard way of saving extra ID data in PAT is to embed the values in the objects, as ```userInt``` here should be appropriate. These extras can be added either directly in the PATMuonProducer or later in PATMuonSlimmer via modifierConfig.

stahlleiton:
> It looks like the muon ID selectors are not saved in a way usually expected in miniAOD.
> 
> The kind of matching implemented in the PR is rather expensive and is not obviously the best possible.
> Both lostTracks and packedPFCandidates have associations saved to the original tracks.
> The underlying reco::Muons `muons` have the original track refs and at that level the track matching can be done directly.
> Do the selectors computed here have to be done on pat::Muons?
> 
> The current proposal is apparently to save the valueMaps computed per patMuons directly, while the patMuons themselves are not saved. This will likely render these maps useless.
> The standard way of saving extra ID data in PAT is to embed the values in the objects, as `userInt` here should be appropriate. These extras can be added either directly in the PATMuonProducer or later in PATMuonSlimmer via modifierConfig.

Thanks Slava for your comments. I will address them as soon as possible. However I wanted to first clarify that the value maps produced in this PR are associated to the packed candidates and not to the pat::Muons. The muon collection is only used to determine which of the packed candidates are also a muon and if yes, extract the outcome of the muon selectors of interest from the muon object matched to the packed candidate. In other words, this producer extends the information of the packed candidates by tagging the outcome of the muon selectors for those that have a track also present in the muon collection. This is done in order to keep the basic information needed to identify all the tracker muons used in HIN analyses (pass the Muon POG Soft ID) without having to increase the size of the slimmed muon collection (as low pT tracker muons are not kept currently in it).

slava77:
> However I wanted to first clarify that the value maps produced in this PR are associated to the packed candidates and not to the pat::Muons.

ah, I should have been more careful to read what's saved on the keying side.
This voids the comments about saving IDs in the slimmedMuons.
However, the part about using available associations to (general) tracks (saved as TrackRefs or equivalent) to match directly to TrackRef of the reco::Muons ```muons``` collection still applies.

PhysicsTools/PatAlgos/plugins/PackedCandidateMuonSelectorProducer.cc
```diff
@@ -0,0 +1,122 @@
...
+        }
+        break;
+      }
+    }
+  }
+
+  // fill the value maps
+  for (const auto& s : muonSelMap) {
+    std::unique_ptr<BoolMap> valueMap(new BoolMap());
+    BoolMap::Filler filler(*valueMap);
```
slava77, line 102: are the ValueMaps compressing well enough?
IIUC, the ratio is perhaps 100:1 of false:true.
Saving a RefVector for passing candidates will be much smaller.
It sounds like the more likely case is to read from these selected muon-like packed cands, rather than reading the full packed cand collection to check/reject the case with muons.
A RefVector will be also easier to use in that case.

PhysicsTools/PatAlgos/python/slimming/MicroEventContent_cff.py
```diff
@@ -171,3 +171,11 @@
 _phase2_timing_extraCommands = ["keep *_offlineSlimmedPrimaryVertices4D_*_*"]
 from Configuration.Eras.Modifier_phase2_timing_cff import phase2_timing
 phase2_timing.toModify(MicroEventContentMC, outputCommands = MicroEventContentMC.outputCommands + _phase2_timing_extraCommands)
+
+_pp_on_AA_extraCommands = [
+    'keep booledmValueMap_*MuonID_*_*',
```
slava77, line 176: wildcards in module names should not be used; the goal is to select everything that's needed explicitly.
'keep booledmValueMap_packedPFCandidateMuonID_*_*'

PhysicsTools/PatAlgos/python/slimming/slimming_cff.py
```diff
@@ -62,6 +62,12 @@
 from Configuration.Eras.Modifier_pp_on_AA_2018_cff import pp_on_AA_2018
 pp_on_AA_2018.toReplaceWith(slimmingTask, slimmingTask.copyAndExclude([slimmedOOTPhotons]))
 
+from Configuration.Eras.Modifier_pp_on_PbPb_run3_cff import pp_on_PbPb_run3
+from PhysicsTools.PatAlgos.packedPFCandidateMuonID_cfi import packedPFCandidateMuonID
+(pp_on_AA_2018 | pp_on_PbPb_run3).toModify(packedPFCandidateMuonID, muonSelectors = cms.vstring(["AllTrackerMuons", "TMOneStationTight"]))
```
slava77, line 67: this is not particularly HI-specific, it should be OK to just have these two selectors in the fillDescriptions directly.

If we still keep it with an era, please drop the type specs:
```suggestion
(pp_on_AA_2018 | pp_on_PbPb_run3).toModify(packedPFCandidateMuonID, muonSelectors = ["AllTrackerMuons", "TMOneStationTight"])
```

slava77 (commented):
this is a catch up after the clarification/correction to my original review error of interpreting the IDs to be mapped to the muons

stahlleiton:
> > However I wanted to first clarify that the value maps produced in this PR are associated to the packed candidates and not to the pat::Muons.
> 
> ah, I should have been more careful to read what's saved on the keying side.
> This voids the comments about saving IDs in the slimmedMuons.
> However, the part about using available associations to (general) tracks (saved as TrackRefs or equivalent) to match directly to TrackRef of the reco::Muons `muons` collection still applies.

How can I access in the packed candidate class the TrackRef associated to the general track?
Checking the class in http://cmsdoxygen.web.cern.ch/cmsdoxygen/CMSSW_11_2_0_pre1/doc/html/d8/d79/classpat_1_1PackedCandidate.html , I see that bestTrack returns a pointer to the unpacked track and pseudoTrack returns the unpacked track directly. However, it is not clear to me which attribute returns the reco::TrackRef.

slava77:
> How can I access in the packed candidate class the TrackRef associated to the general track?
> Checking the class in http://cmsdoxygen.web.cern.ch/cmsdoxygen/CMSSW_11_2_0_pre1/doc/html/d8/d79/classpat_1_1PackedCandidate.html , I see that bestTrack returns a pointer to the unpacked track and pseudoTrack returns the unpacked track directly. However, it is not clear to me which attribute returns the reco::TrackRef.

- PATPackedCandidateProducer (packedPFCandidates) produces an association between the packed and the original PFCandidates, which in turn have the trackRef. So, this one would need a full loop over this association to check trackRef match with the muon's innerTrack.
- PATLostTracks (lostTracks) produces an association from used tracks (generalTracks) to the produced collection (lostTracks). So, in this case a check is simply to see if the muon's inner track corresponds to a lost track via this association (just check if the associated lost track ref is valid for a given trackRef).

perrotta:
@cmsbuild please test with #30354  workflow 140.5611

PhysicsTools/PatAlgos/python/slimming/slimming_cff.py
stahlleiton: Moved the two selectors to the fillDescriptions.

PhysicsTools/PatAlgos/python/slimming/MicroEventContent_cff.py
stahlleiton: Removed the wildcard on the module name

PhysicsTools/PatAlgos/plugins/PackedCandidateMuonSelectorProducer.cc
stahlleiton: Changed the output to a RefVector as suggested. Running over ~1000 events from the HIDoubleMuon PD of 2018 PbPb data, the size of the output went in the case of the packedPFCandidates from an average compressed size of ~160 bytes/event for valueMaps down to 9 bytes/event for the PackedCandidateRefVectors, and for the case of lostTracks from ~38 bytes/event for valueMaps down to 6 bytes/event for the PackedCandidateRefVectors.

PhysicsTools/PatAlgos/python/slimming/MicroEventContent_cff.py
stahlleiton: Removed the + ['keep *_heavyIon_*_*'] part as this was introduced by mistake, and moved the common modifier before the MicroEventContentMC is made out.

PhysicsTools/PatAlgos/plugins/PackedCandidateMuonSelectorProducer.cc
```diff
@@ -0,0 +1,122 @@
...
+      const auto& cand = candidates->refAt(i);
+      // ignore neutral candidates or without track
+      if (cand->charge() == 0 || !cand->hasTrackDetails())
+        continue;
+
+      // check if candidate and muon are compatible
+      const auto& candTrack = cand->pseudoTrack();
+      if (muonTrack.charge() == candTrack.charge() && muonTrack.numberOfValidHits() == candTrack.numberOfValidHits() &&
+          reco::deltaR(muonTrack.eta(), muonTrack.phi(), candTrack.eta(), candTrack.phi()) < maxDeltaR_ &&
+          std::abs((muonTrack.pt() - candTrack.pt()) / muonTrack.pt()) < maxDeltaRelPt_) {
```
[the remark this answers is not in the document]
stahlleiton: Changed to exact matching by reco::TrackRef.id() , by using the associations between the pat::PackedCandidate objects and the general tracks.

PhysicsTools/PatAlgos/plugins/PackedCandidateMuonSelectorProducer.cc
```diff
@@ -0,0 +1,122 @@
...
+  // loop over muons
+  for (const auto& muon : *muons) {
+    // ignore muons without high purity inner track
+    if (muon.innerTrack().isNull() || !muon.innerTrack()->quality(reco::TrackBase::qualityByName("highPurity")))
+      continue;
+
+    // find candidate associated to muon
+    const auto& muonTrack = *muon.innerTrack();
+    for (size_t i = 0; i < nCand; i++) {
+      const auto& cand = candidates->refAt(i);
```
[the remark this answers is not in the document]
stahlleiton: Did not change this part, as now the packedCandidate ref is used in the association

PhysicsTools/PatAlgos/plugins/PackedCandidateMuonSelectorProducer.cc
```diff
@@ -0,0 +1,122 @@
...
+  };
+
+}  // namespace pat
+
+void pat::PackedCandidateMuonSelectorProducer::produce(edm::Event& iEvent, const edm::EventSetup& iSetup) {
+  edm::Handle<pat::MuonCollection> muons;
+  edm::Handle<edm::View<pat::PackedCandidate> > candidates;
+
+  iEvent.getByToken(muonToken_, muons);
+  iEvent.getByToken(candidateToken_, candidates);
```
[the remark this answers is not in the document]
stahlleiton: Changed this part as suggested. However, getHandler() is used for the case of packedPFCandidates, to get their reference.

PhysicsTools/PatAlgos/plugins/PackedCandidateMuonSelectorProducer.cc
```diff
@@ -0,0 +1,122 @@
...
+
+    static void fillDescriptions(edm::ConfigurationDescriptions&);
+
+  private:
+    edm::EDGetTokenT<pat::MuonCollection> muonToken_;
+    edm::EDGetTokenT<edm::View<pat::PackedCandidate> > candidateToken_;
+    std::vector<std::string> muonSelectors_;
+    std::vector<std::string> muonIDs_;
+    std::map<std::string, std::unique_ptr<StringCutObjectSelector<pat::Muon> > > muonIDMap_;
+    double maxDeltaR_, maxDeltaRelPt_;
```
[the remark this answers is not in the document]
stahlleiton: These parameters were removed as now the matching is done by reco::TrackRef id.

PhysicsTools/PatAlgos/plugins/PackedCandidateMuonSelectorProducer.cc
```diff
@@ -0,0 +1,122 @@
...
+
+    void produce(edm::Event&, const edm::EventSetup&) override;
+
+    static void fillDescriptions(edm::ConfigurationDescriptions&);
+
+  private:
+    edm::EDGetTokenT<pat::MuonCollection> muonToken_;
+    edm::EDGetTokenT<edm::View<pat::PackedCandidate> > candidateToken_;
+    std::vector<std::string> muonSelectors_;
+    std::vector<std::string> muonIDs_;
```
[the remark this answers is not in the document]
stahlleiton: changed to const

PhysicsTools/PatAlgos/plugins/PackedCandidateMuonSelectorProducer.cc
```diff
@@ -0,0 +1,122 @@
...
+          maxDeltaRelPt_(iConfig.getUntrackedParameter<double>("maxDeltaRelPt", 1.E-2)) {
+      for (const auto& sel : muonSelectors_) {
+        produces<BoolMap>(sel);
+      }
+      for (const auto& sel : muonIDs_) {
+        muonIDMap_[sel].reset(new StringCutObjectSelector<pat::Muon>("passed('" + sel + "')"));
+        produces<BoolMap>(sel);
+      }
+    }
+    ~PackedCandidateMuonSelectorProducer() override{};
```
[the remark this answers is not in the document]
stahlleiton: Changed this to = default. However, the code style format script adds back the "override" keyword.

PhysicsTools/PatAlgos/plugins/PackedCandidateMuonSelectorProducer.cc
```diff
@@ -0,0 +1,122 @@
...
+
+  public:
+    explicit PackedCandidateMuonSelectorProducer(const edm::ParameterSet& iConfig)
+        : muonToken_(consumes<pat::MuonCollection>(iConfig.getParameter<edm::InputTag>("muons"))),
+          candidateToken_(
+              consumes<edm::View<pat::PackedCandidate> >(iConfig.getParameter<edm::InputTag>("candidates"))),
+          muonSelectors_(iConfig.getParameter<std::vector<std::string> >("muonSelectors")),
+          muonIDs_(iConfig.getParameter<std::vector<std::string> >("muonIDs")),
+          maxDeltaR_(iConfig.getUntrackedParameter<double>("maxDeltaR", 1.E-3)),
+          maxDeltaRelPt_(iConfig.getUntrackedParameter<double>("maxDeltaRelPt", 1.E-2)) {
```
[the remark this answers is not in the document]
stahlleiton: Removed these parameters as the matching is done now directly using reco::TrackRef id.

PhysicsTools/PatAlgos/plugins/PackedCandidateMuonSelectorProducer.cc
```diff
@@ -0,0 +1,122 @@
...
+namespace pat {
+
+  class PackedCandidateMuonSelectorProducer : public edm::stream::EDProducer<> {
+    typedef edm::ValueMap<bool> BoolMap;
+
+  public:
+    explicit PackedCandidateMuonSelectorProducer(const edm::ParameterSet& iConfig)
+        : muonToken_(consumes<pat::MuonCollection>(iConfig.getParameter<edm::InputTag>("muons"))),
+          candidateToken_(
+              consumes<edm::View<pat::PackedCandidate> >(iConfig.getParameter<edm::InputTag>("candidates"))),
```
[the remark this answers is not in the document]
stahlleiton: Changed to PackedCandidateCollection

perrotta:
@cmsbuild please test with #30354 workflow 140.5611

slava77:
@cmsbuild please test workflow 140.5611 with #30354

trying again in a different order

slava77:
@cmsbuild please test workflow 140.5611 with #30354

hopefully this can finish this time

slava77:
@cmsbuild please test workflow 140.5611 with #30354

slava77:
@cmsbuild please test workflow 140.5611

now the baseline has the MC remini included

PhysicsTools/PatAlgos/plugins/PackedCandidateMuonSelectorProducer.cc
slava77: please copy/add the summary about the event sizes to the PR description.
Thank you.

slava77:
+1

for https://github.com/cms-sw/cmssw/pull/30253	68a6b85
- jenkins tests pass and comparisons with the baseline show differences in the new products of the HI reminiAOD workflow, as expected

santocch:
+1

slava77:
@silviodonato 
please check this PR.
Merging it would help to disentangle from the overlap in #30313

silviodonato:
+1

PR #30255, opened 2020-06-15
embed hi photon isolation in pat::photon userdata

bi-ran:
this PR takes the HI-style photon isolation information, which are stored in RECO as valuemaps, and adds it to the userdata of the PAT photons. the change is applied to the pp_on_AA_2018 and pp_on_PbPb_run3 eras.

the existing code is modified slightly to allow the reco::Photon collection to also be used as input, because the isolation valuemaps are keyed by that collection. the changes should not affect usage in pp, since edm::View allows both reco:: and pat:: collections to be read.

tested with the 158/159 workflows that the information can be stored and read out.

@mandrenguyen

### DIFF
--- a/RecoEgamma/EgammaTools/interface/EGExtraInfoModifierFromValueMaps.h
+++ b/RecoEgamma/EgammaTools/interface/EGExtraInfoModifierFromValueMaps.h
@@ -83,12 +83,12 @@ public:
   using ValMapToken = edm::EDGetTokenT<edm::ValueMap<MapType>>;
   using ValueMaps = std::unordered_map<std::string, ValMapToken>;
   struct electron_config {
-    edm::EDGetTokenT<edm::View<pat::Electron>> tok_electron_src;
+    edm::EDGetTokenT<edm::View<reco::GsfElectron>> tok_electron_src;
     ValueMaps tok_valuemaps;
   };
 
   struct photon_config {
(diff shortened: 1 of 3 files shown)


perrotta:
please test

perrotta:
please test
(to refresh the tests)

perrotta:
@cmsbuild please test with #30354 workflow 140.5611

perrotta:
please test workflow 140.5611 with #30354
(order matters..)

slava77:
@cmsbuild please test workflow 140.5611 with #30354

slava77:
@cmsbuild please test workflow 140.5611 with #30354

perrotta:
+1
- The new isolation selection for PAT photons in Run3 HI has been compared with the current one thanks to the new HI re-miniAOD workflows defined with #30354: the size of slimmed photons is significantly reduced with this selection with respect to the one that would have been obtained with the standard (pp) selection
- Changes in the reco code are instrumental to apply them to reco objects, instead of only candidates, but they have no effect on the other standard workflows 
- Jenkins tests pass and show no differences anywhere else than in the HI re-miniAOD

silviodonato:
+1

perrotta:
@cmsbuild please test workflow 140.5611
(this will hopefully unblock bot)

PR #30257, opened 2020-06-15
Keep GEMDigiSimLinks after the DIGI step

dildick:
#### PR description:

RAW and RECO simulated data currently keep the `StripDigiSimLink` objects for GEMs. They are actually not that useful. However, the `GEMDigiSimLink` is much more useful as it properly links the GEM `SimHit` with the `GEMDigi` This PR adds it to the event content after the DIGI step. The size is about the same as the `StripDigiSimLink`. After running WF 23234.0:

<PRE>
-bash-4.2$ edmEventSize -a -v step2.root | grep GEM
GEMDigiSimLinkedmDetSetVector_simMuonME0Digis_ME0_HLT. 15540 5387.4
GEMDigiSimLinkedmDetSetVector_simMuonGEMDigis_GEM_HLT. 14341 4546.3
StripDigiSimLinkedmDetSetVector_simMuonGEMDigis_GEM_HLT. 14279.2 4508.1
</PRE>

If the `GEMDigiSimLink` takes up too much space, I would propose to drop `StripDigiSimLink`, which is not used in the simulation AFAIK.

#### PR validation:

Tested with WF 23234.0

@jshlee

### DIFF
--- a/SimMuon/Configuration/python/SimMuon_EventContent_cff.py
+++ b/SimMuon/Configuration/python/SimMuon_EventContent_cff.py
@@ -55,23 +55,23 @@ SimMuonPREMIX = cms.PSet(
 
 from Configuration.Eras.Modifier_run2_GEM_2017_cff import run2_GEM_2017
 run2_GEM_2017.toModify( SimMuonFEVTDEBUG, outputCommands = SimMuonFEVTDEBUG.outputCommands + ['keep *_simMuonGEMDigis_*_*'] )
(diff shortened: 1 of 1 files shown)


civanch:
please test

civanch:
+1

@dildick , if StripDigiSimLink is not used, then drop it with the next occasion independently on size

silviodonato:
+1
fyi @perrotta @slava77

PR #30216, opened 2020-06-12
added esConsumes to modules in SimCalorimetry/ EcalEBTrigPrimProducers

JamminJones:
#### PR description:

added esConsumes to 1 module
#### PR validation:

the code compiles

PR #30264, opened 2020-06-16
change EcalTPGScale to use esConsumes

JamminJones:
#### PR description:

changed all modules that used EcalTPGScale to also use esConsumes.
#### PR validation:

the code compiles


Dr15Jones:
This change was made under my supervision

Dr15Jones:
please test

Dr15Jones:
please test

RecoLocalCalo/EcalRecProducers/plugins/EcalRecHitWorkerRecover.cc
```diff
@@ -175,9 +177,10 @@ bool EcalRecHitWorkerRecover::run(const edm::Event& evt,
     EcalTrigPrimDigiCollection::const_iterator tp = tpDigis->find(ttDetId);
     // recover the whole trigger tower
     if (tp != tpDigis->end()) {
+      EcalTPGScale ecalScale{ecalScaleTokens_, *eventSetup_};
```
slava77, line 180: is this really fast? The implementation suggests 3 ```EventSetup::getData``` calls.
This piece is inside of a call that can repeat many times per event.
Better keep these cached in the ```::set``` method (same comment applies to the other case in this file)

SimCalorimetry/EcalEBTrigPrimProducers/plugins/EcalEBTrigPrimProducer.cc
```diff
@@ -85,46 +80,37 @@ void EcalEBTrigPrimProducer::beginRun(edm::Run const& run, edm::EventSetup const
 
 unsigned long long EcalEBTrigPrimProducer::getRecords(edm::EventSetup const& setup) {
   // get parameter records for xtals
-  edm::ESHandle<EcalTPGLinearizationConst> theEcalTPGLinearization_handle;
-  setup.get<EcalTPGLinearizationConstRcd>().get(theEcalTPGLinearization_handle);
+  edm::ESHandle<EcalTPGLinearizationConst> theEcalTPGLinearization_handle =
+      setup.getHandle(theEcalTPGLinearization_Token_);
```
slava77, line 84: wouldn't auto work here?
```suggestion
  auto theEcalTPGLinearization_handle = setup.getHandle(theEcalTPGLinearization_Token_);
```
for a more compact impl

RecoLocalCalo/EcalRecProducers/plugins/EcalRecHitWorkerRecover.cc
Dr15Jones: @slava77 this is likely faster than the old code and at worst only just as fast. In the old code, each call to functions of `EcalTPGScale` were making the 3 calls to `EventSetup::getData`. Now we only do it in the constructor and cache the results so they can be reused.

SimCalorimetry/EcalEBTrigPrimProducers/plugins/EcalEBTrigPrimProducer.cc
Dr15Jones: We kept the type in order to minimize the amount of change done. Using `auto` would definitely work.

RecoLocalCalo/EcalRecProducers/plugins/EcalRecHitWorkerRecover.cc
slava77: I'm confused, ```ecalScale``` is clearly not cached in this case/loc.
Wouldn't it be faster/more effective to cache ecalScale in the ::set method (where a few other ES data is cached already)?

Dr15Jones: In the old code, caching `ecalScale` was only caching a pointer to the `EventSetup`, which the new code also does. In the old code, each call to a function of `EcalTPGScale` would call multiple `EventSetup::getData` calls. Now this is only done in the constructor. So the changes we made are better.

Caching ecalScale would be more difficult since `EcalTPGScale` now requires the `EventSetup` to be passed to its constructor (since all other uses of `EcalTPGScale` other than in `EcalRecHitWorkerRecover` were setting it right after calling the constructor) which means `EcalRecHitWorkerRecover` could not hold an `EalTPGScale` but would have to hold a `std::unique_ptr`. As I'm dealing with an undergraduate and our only purpose is to get the `esConsumes` done and the fact that the changes we made here are already a performance improvement over the old code, I'd prefer not to do any further changes in this pull request. If further changes are wished, I'd prefer a GitHub issue be made and the actual developers of this code do futher changes.

Dr15Jones:
please test

RecoLocalCalo/EcalRecProducers/plugins/EcalRecHitWorkerRecover.cc
slava77: > Caching ecalScale would be more difficult since `EcalTPGScale` now requires the `EventSetup` to be passed to its constructor

It's unclear to me why adding ```ecalScale_ =  EcalTPGScale ecalScale(ecalScaleTokens_, es)``` in ```EcalRecHitWorkerRecover::set(const edm::EventSetup& es)``` is complicated.
Perhaps I'm missing something (was e.g. a default constructor or an assignment operator removed).

Dr15Jones: There is no default constructor

kpedro88:
Looks like conflict is caused by #30216
@Dr15Jones now you can teach @JamminJones how to rebase: http://cms-sw.github.io/tutorial-resolve-conflicts.html

Dr15Jones:
please test

Dr15Jones:
The unit test failures are unrelated to this pull request because the same failures are happening in the IBs.

slava77:
+1

for https://github.com/cms-sw/cmssw/pull/30264	6724f8c
- technical; after some discussion, the suboptimal update in ``` EcalRecHitWorkerRecover::run``` was considered acceptable
- jenkins tests pass for what's related to this PR updates

kpedro88:
+upgrade

Dr15Jones:
ping

silviodonato:
merge
l1:  @benkrikler @rekovic
analysis:  @santocch
please have a look

santocch:
+1

tlampen:
+1

PR #30273, opened 2020-06-17
Reduce boost dependency in PhysicsTools/NanoAOD

camolezi:
#### PR description:
Changed boost::ptr_vector for std::vector<std::unique_ptr<>>
The code should have the same behavior.

#### PR validation:

@vgvassilev @davidlange6

### DIFF
--- a/PhysicsTools/NanoAOD/interface/SimpleFlatTableProducer.h
+++ b/PhysicsTools/NanoAOD/interface/SimpleFlatTableProducer.h
@@ -8,8 +8,8 @@
 #include "CommonTools/Utils/interface/StringCutObjectSelector.h"
 #include "CommonTools/Utils/interface/StringObjectFunction.h"
 
+#include <memory>
 #include <vector>
-#include <boost/ptr_container/ptr_vector.hpp>
 
 template <typename T, typename TProd>
 class SimpleFlatTableProducerBase : public edm::stream::EDProducer<> {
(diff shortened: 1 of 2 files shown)


gpetruc:
isn't this worse performance-wise?
here all pointers are owned by the owner of the vector, so there's no need of reference counting or carrying around a custom deleter. 
A plain `vector<Variable *>` could have been sufficient for all purposes except that it would have needed extra code in the destructor and the copy & move operators (and require a few more derefercing in the access), so I had used boost.
If boost is disfavoured, shouldn't be `vector<unique_ptr>`  the right thing instead, to avoid the added overheads?

camolezi:
> isn't this worse performance-wise?
> here all pointers are owned by the owner of the vector, so there's no need of reference counting or carrying around a custom deleter.
> A plain `vector<Variable *>` could have been sufficient for all purposes except that it would have needed extra code in the destructor and the copy & move operators (and require a few more derefercing in the access), so I had used boost.
> If boost is disfavoured, shouldn't be `vector<unique_ptr>` the right thing instead, to avoid the added overheads?

Yeah, you're right. Probably a unique_ptr is better in this situation.

silviodonato:
Any objections @peruzzim @santocch ?

silviodonato:
merge
 @santocch @peruzzim

santocch:
+1

Issue #28914, opened 2020-02-11
Migration to DD4hep: Geometric(Timing)Det dependence on DDD


fabiocos:
In the process of migrating the MTD reco geometry to DD4hep, I notice that the ```GeometricTimingDet``` object in ```MTDNumberingBuilder```, as well as its tracker counterpart ```GeometricDet```, depend explicitly on DDD through ```DDSolidShape``` and the use of the ```ExpandedNode```. This means that even the present tracker migration is not fully complete.

Before inventing ad-hoc solutions, I would like to understand from @ianna and @cvuosalo whether they already envisaged a strategy to manage this issue.

@vargasa FYI

fabiocos:
assign geometry

ianna:
@cvuosalo and @fabiocos - following the discussion with Adinda here.
Given that we want to support the legacy payloads and due to a recorded `int` as a shape identifier in a `PGeometricDet` we need to make sure that the old and the new `DDSolidShape` enums are identical. It's not the case now:
https://cmssdt.cern.ch/lxr/source/DetectorDescription/Core/interface/DDSolidShapes.h#0006
https://cmssdt.cern.ch/lxr/source/DetectorDescription/DDCMS/interface/DDSolidShapes.h#0062

fabiocos:
@ianna thanks for looking into this. Silly question: what was the reason for this choice at the beginning of the porting? I mean I see that the two lists mostly overlaps, and codes could be made correspondent,  with 2 o3 noticeable differences. Do we have loops that are preventing discontinuities in the lists? The question is of course about the correspondence of polycones and plyhedras...

ianna:
> @ianna thanks for looking into this. Silly question: what was the reason for this choice at the beginning of the porting? I mean I see that the two lists mostly overlaps, and codes could be made correspondent, with 2 o3 noticeable differences. Do we have loops that are preventing discontinuities in the lists? The question is of course about the correspondence of polycones and polyhedras...

Changing `enum` was intentional to weed out reliance on `int`.

fabiocos:
@ianna sorry, perhaps I was not clear. I am not questioning the ```enum``` choice, just wondering about the reason for the two different structures.

cvuosalo:
@fabiocos Adinda is fixing the `DDSolidShape` issue in these files, but the `DDExpandedView` part still needs to be migrated. What are your plans for migrating these files?

fabiocos:
@cvuosalo are you speaking of 
[...]

PR #30276, opened 2020-06-17
Tracker/MTD geometry: clean Geometric(Timing)Det dependence on DDD

fabiocos:
#### PR description:

Following the discussion in #28914 , I propose a cleaning of ```GeometricDet``` and ```GeometricTimingDet``` classes to remove the dependence on ```DDExpandedNode```, which as far as I can see is just fake. In order to make things transparent, I add in ```DDCMS``` a few ```using``` statements included in ```DDD``` which may help in making transparent the migration. 

A non trivial dependence remains in ```GeometricDetExtras``` and into its timing counterpart. As far as I can see:

- the use of ```DDExpandedNode``` is real, and a fully independent implementation of similar functionalities should be made using DD4hep if needed, so far there is none;

- this is anyway not impacting the persistent ```PGeometricDet```;

- the use of this object is for pure debugging purpose only, as far as I can see, so in principle its removal should not break production workflows. For sure this is true for ```GeometricTimingDetExtras``` that appears to be a cut-and-paste minimal adjustment of the tracker code, never used and not providing meaningful results at present as far as I can see.

For this reason I propose here a simple removal of this part for MTD, to be possibly replaced with something else in future. But there are already debugging classes that can be possibly extended in case.

As far as Tracker geometry is concerned, it is up to the DPG experts say whether they still need these functionalities, and in case provide a DD4hep-based implementation (that could replace asymptotically the DDD implementation).

#### PR validation:

MTD geometry unit tests and wf 23234.0 run.


fabiocos:
please test

fabiocos:
hold

I apparently missed this issue in my second commit, checking it again

fabiocos:
please test

Alignment/OfflineValidation/plugins/TrackerGeometryCompare.cc
```diff
@@ -53,6 +53,8 @@
 #include "DataFormats/TrackerCommon/interface/TrackerTopology.h"
 #include "Geometry/Records/interface/TrackerTopologyRcd.h"
 
+#include "DetectorDescription/Core/interface/DDCompactView.h"
```
cvuosalo, line 56: Can you please explain this file? It is not migrated. Is it part of `GeometricTimingDetExtras`?

fabiocos: @cvuosalo in the case of this file, as well as the ```RecoTracker/TkDetLayers/src/Phase2EndcapRingBuilder.cc``` it is a matter of nested include dependencies, that now are exposed by the removal of the ```DDExpandedView``` from ```GeometricDet```. The compilation fails as soon as you update that class.

fabiocos:
unhold

Alignment/OfflineValidation/plugins/TrackerGeometryCompare.cc
cvuosalo: My confusion results from two different versions of this file being included in #29905 and this PR. #29905 has the migrated version. If #29905 is merged first, this PR will probably need to be rebased.

RecoTracker/TkDetLayers/src/Phase2EndcapRingBuilder.cc
```diff
@@ -1,5 +1,6 @@
 #include "Phase2EndcapRingBuilder.h"
 #include "TrackingTools/DetLayers/interface/DetLayerException.h"
+#include "FWCore/MessageLogger/interface/MessageLogger.h"
```
cvuosalo, line 3: Does this file even belong in this PR? It doesn't seem related, and it brings in the requirement for a Reco signature.

cvuosalo:
+1

cvuosalo:
The decision in this PR is to leave the Tracker version of `GeometricDetExtras` unmigrated. If the Tracker DPG wants to use `GeometricDetExtras` tool in the future, they will have to migrate it.

Alignment/OfflineValidation/plugins/TrackerGeometryCompare.cc
fabiocos: @cvuosalo up to the release manager decide which PR is mature for integration first. In case I will rebase mine on that one, not a big issue. BTW, I see that @ianna has already made this addition in #29905

RecoTracker/TkDetLayers/src/Phase2EndcapRingBuilder.cc
fabiocos: @cvuosalo unfortunately yes, without it the code does not compile, you may easily try yourself.

fabiocos:
@cvuosalo as I said, I believe that it is up to the Tracker DPG decide what they want to do with this object. For MTD it is not used so far, and its implementation is not really usable, just a cut-and-paste adjustment of the tracker code that is unusable in its present form. This is not true for the tracker original object, but a DD4hep based implementation of that requires some non completely trivial work, which goes beyond my main task related to MTD. I wonder to which extent this debugging code is really used nowadays, it looks useless to invest work in updating dead code (if that is the case).

fabiocos:
to be even more explicit: this object is used in just 3 places in test modules as far as I can see, either through 

https://cmssdt.cern.ch/lxr/source/Geometry/TrackerNumberingBuilder/src/CmsTrackerDebugNavigator.cc#0017

or to print equivalent information, and the name of some higher level volumes in the stack. One should replace the equivalent information with something from DD4hep filtered view.  The information coming from DDD logical parts should be appropriately replaced, the copy number is easily accessible, as far as the GeoHistory is concerned, for what I can see most probably the information needed in https://cmssdt.cern.ch/lxr/source/SLHCUpgradeSimulations/Geometry/test/ModuleInfo_Phase2.cc#0433 can be retrieved through the DD4hep geoHistory() as done in https://cmssdt.cern.ch/lxr/source/Geometry/MTDNumberingBuilder/test/DD4hep_MTDTopologyAnalyzer.cc#0210

perrotta:
+1
- Just a technical fix for reco
- Jenkins tests pass

pohsun:
+1

kpedro88:
+upgrade

silviodonato:
+1

PR #30295, opened 2020-06-18
HCAL - HB TDC digi default to non-packed backport to 11_1_X

lwang046:
backport of #30213


perrotta:
please test

perrotta:
+1
- Non packed values cannot be read with current DQM, which is not considered a big issue for HCAL validation

silviodonato:
+1

PR #30302, opened 2020-06-19
Faster mahi pulse shape initialization

igv4321:
#### PR description:

MAHI now memoizes PulseShapeFunctor objects instead of recreating them every time a different pulse shape is encountered. Only a few different pulse shapes exist (corresponding to different SiPM/HPD types), so std::vector is used to hold the collection instead of std::map.

No changes in the relval results expected.

#### PR validation:

The usual runTheMatrix tests were run

### DIFF
--- a/RecoLocalCalo/HcalRecAlgos/interface/MahiFit.h
+++ b/RecoLocalCalo/HcalRecAlgos/interface/MahiFit.h
@@ -1,6 +1,10 @@
 #ifndef RecoLocalCalo_HcalRecAlgos_MahiFit_HH
 #define RecoLocalCalo_HcalRecAlgos_MahiFit_HH
 
+#include <climits>
+#include <utility>
+#include <memory>
+
 #include "DataFormats/HcalDetId/interface/HcalSubdetector.h"
 #include "DataFormats/HcalDetId/interface/HcalDetId.h"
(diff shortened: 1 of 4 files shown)


kpedro88:
please test

perrotta:
@igv4321 : do you have any measurement of the speedup expected because of this PR?
I'm comparing step3 of wf 136.888 and I can't see any relevant improvement in timing...

igv4321:
Yeah, the difference for wf 136.888 would be that the old MahiFit would call PulseShapeFunctor constructor 200 times while the new one calls it just two times. At the same time, hcal local reco processes about half a million HB/HE rechits. While the time difference should be there, it would not be easy to see it.

I did not try to measure the speed improvement. I simply do not have access to a machine with a sufficiently low load to make precise measurements.

perrotta:
+1
- Reduction in number of calls of the PulseShapeFunctor is expected to speed up the Mahi initialization time: such an effect wasn't clearly observed in the limited sample workflow run possibly with some concurrent job in the machine, but it is expected to be there
- IgProf shows that the increased memory consumption (in  MahiFit::setPulseShapeTemplate) is quite acceptable: from 4.2 kb to 8.6 kb
- Jenkins tests pass and show no differences in outputs

silviodonato:
+1

PR #28857, opened 2020-02-03
New relvals workflows for JME customized nanoaod

srimanob:
#### PR description:
This adds workflows need to test JME customised NanoAOD.

Workflows: 25202.01 

More to come in this PR, this is just a quick version so JME can test.

#### PR validation:
Local test with https://github.com/cms-sw/cmssw/pull/28854 works well.

#### if this PR is a backport please specify the original PR and why you need to backport that PR:

Before submitting your pull requests, make sure you followed this checklist:
- verify that the PR is really intended for the chosen branch
- verify that changes follow [CMS Naming, Coding, And Style Rules](http://cms-sw.github.io/cms_coding_rules.html)
- verify that the PR passes the basic test procedure suggested in the [CMSSW PR instructions](https://cms-sw.github.io/PRWorkflow.html)

PR #30306, opened 2020-06-19
GEN-only NanoAOD configs (NanoGEN) and cmsDriver support

kdlong:
#### PR description:

Configuration files to produce only the GEN branches of NanoAOD for a convenient and standardized ntuple format for gen analysis. cmsDriver support to produce directly from a gridpack, or from a MiniAOD file, or anything in between.

#### PR validation:

Has been tested by generating quite a few samples privately. Some configs and cmsDriver commands tested here: https://github.com/kdlong/WMassNanoGen

Additionally configurability is still a work in progress. Small amount of restructuring of configs should be approved by NanoAOD experts

### DIFF
--- a/Configuration/Applications/python/ConfigBuilder.py
+++ b/Configuration/Applications/python/ConfigBuilder.py
@@ -937,6 +937,7 @@ class ConfigBuilder(object):
         self.RECOSIMDefaultCFF="Configuration/StandardSequences/RecoSim_cff"
         self.PATDefaultCFF="Configuration/StandardSequences/PAT_cff"
         self.NANODefaultCFF="PhysicsTools/NanoAOD/nano_cff"
+        self.NANOGENDefaultCFF="PhysicsTools/NanoAOD/nanogen_cff"
         self.EIDefaultCFF=None
         self.SKIMDefaultCFF="Configuration/StandardSequences/Skims_cff"
         self.POSTRECODefaultCFF="Configuration/StandardSequences/PostRecoGenerator_cff"
@@ -986,6 +987,8 @@ class ConfigBuilder(object):
         self.REPACKDefaultSeq='DigiToRawRepack'
         self.PATDefaultSeq='miniAOD'
         self.PATGENDefaultSeq='miniGEN'
+        #TODO: Check based of file input
+        self.NANOGENDefaultSeq='nanogenSequence' if any([x in self.stepMap for x in ['LHE', 'GEN', 'AOD']]) else 'nanogenMiniSequence' 
         self.NANODefaultSeq='nanoSequence'
 
         self.EVTCONTDefaultCFF="Configuration/EventContent/EventContent_cff"
@@ -1689,6 +1692,17 @@ class ConfigBuilder(object):
                 self._options.customise_commands = self._options.customise_commands + " \n"
             self._options.customise_commands = self._options.customise_commands + "process.unpackedPatTrigger.triggerResults= cms.InputTag( 'TriggerResults::"+self._options.hltProcess+"' )\n"
 
+    def prepare_NANOGEN(self, sequence = "nanoAOD"):
+        ''' Enrich the schedule with NANOGEN '''
+        # TODO: Need to modify this based on the input file type
+        fromGen = any([x in self.stepMap for x in ['LHE', 'GEN', 'AOD']])
+        self.loadDefaultOrSpecifiedCFF(sequence,self.NANOGENDefaultCFF)
+        self.scheduleSequence(sequence.split('.')[-1],'nanoAOD_step')
+        custom = "customizeNanoGEN" if fromGen else "customizeNanoGENFromMini"
+        if self._options.runUnscheduled:
+            self._options.customisation_file_unsch.insert(0, '.'.join([self.NANOGENDefaultCFF, custom]))
+        else:
+            self._options.customisation_file.insert(0, '.'.join([self.NANOGENDefaultCFF, custom]))
 
     def prepare_EI(self, sequence = None):
         ''' Enrich the schedule with event interpretation '''
--- a/Configuration/Applications/python/ConfigBuilder.py
+++ b/Configuration/Applications/python/ConfigBuilder.py
@@ -988,7 +988,7 @@ class ConfigBuilder(object):
         self.PATDefaultSeq='miniAOD'
         self.PATGENDefaultSeq='miniGEN'
         #TODO: Check based of file input
-        self.NANOGENDefaultSeq='nanogenSequence' if any([x in self.stepMap for x in ['LHE', 'GEN', 'AOD']]) else 'nanogenMiniSequence' 
+        self.NANOGENDefaultSeq='nanogenSequence'
         self.NANODefaultSeq='nanoSequence'
 
         self.EVTCONTDefaultCFF="Configuration/EventContent/EventContent_cff"
--- a/PhysicsTools/NanoAOD/python/genWeightsTable_cfi.py
+++ b/PhysicsTools/NanoAOD/python/genWeightsTable_cfi.py
@@ -1,0 +1,20 @@
+import FWCore.ParameterSet.Config as cms
+
+genWeightsTable = cms.EDProducer("GenWeightsTableProducer",
+    genEvent = cms.InputTag("generator"),
+    genLumiInfoHeader = cms.InputTag("generator"),
+    lheInfo = cms.VInputTag(cms.InputTag("externalLHEProducer"), cms.InputTag("source")),
+    preferredPDFs = cms.VPSet( # see https://lhapdf.hepforge.org/pdfsets.html
+        cms.PSet( name = cms.string("PDF4LHC15_nnlo_30_pdfas"), lhaid = cms.uint32(91400) ),
+        cms.PSet( name = cms.string("NNPDF31_nnlo_hessian_pdfas"), lhaid = cms.uint32(306000) ),
+        cms.PSet( name = cms.string("NNPDF30_nlo_as_0118"), lhaid = cms.uint32(260000) ), # for some 92X samples. Note that the nominal weight, 260000, is not included in the LHE ...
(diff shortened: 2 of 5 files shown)


kdlong:
@agrohsje @peruzzim @menglu21

kdlong:
If the ultra legacy GEN is going to run in CMSSW 10_6, then we could have a consistent setup to do central production of NanoGEN through a similar campaign. How hard would this be to support in McM and DAS?

silviodonato:
please test

silviodonato:
@kdlong let me forward your question to  @chayanit @pgunnell @wajidalikhan

wajidalikhan:
> If the ultra legacy GEN is going to run in CMSSW 10_6, then we could have a consistent setup to do central production of NanoGEN through a similar campaign. How hard would this be to support in McM and DAS?

At the moment we have UL 16, 17, 18 all running in CMSSW 10_6. We can open a new campaign for NanoGEN campaign and not sure by if I understood your point regarding the support on DAS. If we want to change the data tier from GEN to NanoGEN that would need some fix.

silviodonato:
@kdlong What is the plan of this PR?

kdlong:
Sorry I overlooked @wajidalikhan 's comment. I don't think you need a new data tier. It produces NANOAODSIM with a subset of the usual branches.

As for my plan for the PR: Broadly, I would like it to be possible for people to make central requests for NANOGEN samples. The current version of this PR does exactly that, it supports cmsDriver commands to build configs that produce a NANOAODSIM output from gridpack, LHE EDM, MINIAOD etc. 

I firmly believe that there would be wide applicability with this. Note that we've used this setup extensively to validate the large single V requests for the mW analysis, which are by far the most carefully validated MC samples we've ever had in CMS. The vast majority of samples in CMS are produced with no careful study of the gen content before production, and I think this data tier would help that (even if just done privately on crab)

Specifically to why I labled it a WIP: I can imagine some use cases that aren't fully covered, for example, it could be more configurable (one might want to configure the selection on some gen collections, for example). For this I'd like more feedback from NANOAOD and GEN experts.

silviodonato:
What's the status/plan of this PR? If I remember correctly, it was supposed to be discussed this week at GEN meeting.

kdlong:
It was discussed at GEN last week and also in PdmV coordination. I think there is support from both GEN and pdmv for this idea.

On the PR itself, I was hoping to add a bit more customization. Is it possible to schedule a module using --customize in cmsDriver? For example, it would be nice to schedule the gen particle pruner if someone wants a more slimmed down format. If not, I suppose I could schedule it and just modify whether or not it does any actual filtering.

silviodonato:
@kdlong yes, we can call customization function using `--customize` in `cmsDriver`

kdlong:
Hi @silviodonato but is it really possible to modify the sequence and schedule new modules with --customize? I had decided the easiest thing to do was always schedule the GenParticlePruner and just modify the arguments passed into it (so sometimes running it but not really doing any skimming).

silviodonato:
@kdlong the customization function takes as input `process`, including its schedule and sequence, so you can basically do whatever you want. You can find a simple example at https://github.com/cms-sw/cmssw/pull/30484/files#diff-e47f1178d5ccdc2b052574fd67ba44f3R1-R4

santocch:
+1

silviodonato:
@kdlong do you have any news about this PR?

kdlong:
Sorry for the delay. I added a few customize functions that I think could be useful. From my end I think it's ready to merge, some additional features (like the ability to store more of the LHE file content) can come in separate PRs.

Configuration/Applications/python/ConfigBuilder.py
```diff
@@ -1689,6 +1692,17 @@ def prepare_NANO(self, sequence = "nanoAOD"):
                 self._options.customise_commands = self._options.customise_commands + " \n"
             self._options.customise_commands = self._options.customise_commands + "process.unpackedPatTrigger.triggerResults= cms.InputTag( 'TriggerResults::"+self._options.hltProcess+"' )\n"
 
+    def prepare_NANOGEN(self, sequence = "nanoAOD"):
+        ''' Enrich the schedule with NANOGEN '''
+        # TODO: Need to modify this based on the input file type
```
kdlong, line 1697: Is there a way I can detect the input data tier? I think this would fail if someone wanted to run only the NanoGen sequence from an input GEN file, for example

mariadalfonso:
please test

mariadalfonso:
@kdlong 
is there a workflow to automatically test this 'NanoGen'?

kdlong:
@mariadalfonso good question... Nothing has been set up. Most important is that none of the small refactoring breaks Nano. It should be really easy to set up a test by just asking for a subset of the Nano checks to be done (since it is 100% a subset of the Nano). If there are examples of what is there for Nano and how to define new tests, I could probably put something together.

mariadalfonso:
@kdlong 
you can check how JME wf nano were added here
https://github.com/cms-sw/cmssw/pull/28857
https://github.com/cms-sw/cmssw/pull/29279

I suggest to do this in a separatePR

mariadalfonso:
@kdlong 
one last general question: you introduced now the nanogenSequence and nanogenMiniSequence; 
what are the use cases for those ? I'm trying to figure out when and who runs what

kdlong:
The sequences are used or cmsDriver, since there are a few extra things to schedule when the input is MiniAOD. It would be possible to have only one sequence and modify it to add the mergedGenParticles, genParticles2HepMC, and  genParticles2HepMCHiggsVtx in the customizeNanoGENFromMini function if you prefer. It would probably be more clear, and less duplicate code to maintain.

> @kdlong
> you can check how JME wf nano were added here
> #28857
> #29279
> 
> I suggest to do this in a separatePR

Thanks! Will take a look at this in a week (we'll be mostly offline this coming week)

mariadalfonso:
> The sequences are used or cmsDriver, since there are a few extra things to schedule when the input is MiniAOD. It would be possible to have only one sequence and modify it to add the mergedGenParticles, genParticles2HepMC, and genParticles2HepMCHiggsVtx in the customizeNanoGENFromMini function if you prefer. It would probably be more clear, and less duplicate code to maintain.

yes, let's find a compact way to handle the common part of the sequences. Will be easier in the future.

mariadalfonso:
please test

kdlong:
Can anyone suggest a way to detect the input data tier in cmsDriver? Right now it just checks to see if steps before MINIAOD are run to decide whether to expect MINIAOD or GEN-like event content.

silviodonato:
@kdlong `cmsDriver.py` creates the python files without opening the input files. Therefore, there is no way to check whether a file is a MINIAOD or not (unless you simply look at the file name or something like that). (@cms-sw/pdmv-l2 please correct me if I'm wrong)

silviodonato:
I would also suggest you to add this new step (NanoGEN) in the unit test https://github.com/cms-sw/cmssw/blob/d5ef6105923adc43ea260f69b2ba569c5f291078/PhysicsTools/NanoAOD/test/runtests.sh#L12

mariadalfonso:
+xpog

This PR add a new nanoGEN format, sign off at commit d5ef610
It will be followed up by the creation of a dedicated workflow to test in the runTheMatrix and unit test

silviodonato:
assign generators

silviodonato:
@cms-sw/generators-l2 could you tell us what is the status of this PR? Is it ready to go?

agrohsje:
i tested this pr a while ago and it looks quite good to me. i have one question about the preferred pdfs. for ul the last list is this one for 5 FS:
https://github.com/cms-sw/genproductions/blob/master/MetaData/pdflist_5f_2017.dat
but 325300 doesn't appear in the preferred list. @kdlong should we extend it/adjust?

mariadalfonso:
@agrohsje 
let's keep this long standing PR limited to the first set of changes for the GENnano.

Based on the discussion on the xPOG 25aug, we indeed need to update the PDF also for central nano UL.
Among other things I would like to understand which precision we need for these weights and understand the size change, update the nanoDQM for pdf ...

agrohsje:
Thanks @mariadalfonso. Doing this in another PR is also fine with me. I just spotted and wanted to make sure we don't forget. 
+1

agrohsje:
+1

santocch:
+1

silviodonato:
+operations

silviodonato:
+1

PhysicsTools/NanoAOD/python/genWeightsTable_cfi.py
```diff
@@ -0,0 +1,20 @@
...
+        cms.PSet( name = cms.string("NNPDF31_nnlo_hessian_pdfas"), lhaid = cms.uint32(306000) ),
+        cms.PSet( name = cms.string("NNPDF30_nlo_as_0118"), lhaid = cms.uint32(260000) ), # for some 92X samples. Note that the nominal weight, 260000, is not included in the LHE ...
+        cms.PSet( name = cms.string("NNPDF30_lo_as_0130"), lhaid = cms.uint32(262000) ), # some MLM 80X samples have only this (e.g. /store/mc/RunIISummer16MiniAODv2/DYJetsToLL_M-50_TuneCUETP8M1_13TeV-madgraphMLM-pythia8/MINIAODSIM/PUMoriond17_80X_mcRun2_asymptotic_2016_TrancheIV_v6_ext1-v2/120000/02A210D6-F5C3-E611-B570-008CFA197BD4.root )
+        cms.PSet( name = cms.string("NNPDF30_nlo_nf_4_pdfas"), lhaid = cms.uint32(292000) ), # some FXFX 80X samples have only this (e.g. WWTo1L1Nu2Q, WWTo4Q)
+        cms.PSet( name = cms.string("NNPDF30_nlo_nf_5_pdfas"), lhaid = cms.uint32(292200) ), # some FXFX 80X samples have only this (e.g. DYJetsToLL_Pt, WJetsToLNu_Pt, DYJetsToNuNu_Pt)
+    ),
+    namedWeightIDs = cms.vstring(),
+    namedWeightLabels = cms.vstring(),
+    lheWeightPrecision = cms.int32(14),
+    maxPdfWeights = cms.uint32(150),
```
mariadalfonso, line 18: this miss the
keepAllPSWeights = cms.bool(False)

note we need to set these to False for all the central-nan

PR #29592, opened 2020-04-29
[LLVM10] EventFilter fix warnings for clang 10

mrodozov:
#### PR description:

Fixes warnings in clang 10

#### PR validation:

Builds without warnings with CLANG IB
tokenArraySeparator and tokenArrayEnd belong to the same enum so thats probably or not and

PR #30315, opened 2020-06-19
Revert change causing regression in DAQ/HLT JSON input parsing

smorovic:
#### PR description:
Reverts json_reader.cpp modification which was part of this commit:
https://github.com/cms-sw/cmssw/commit/d030f663df51846c06f626488b8dfa2490d187bd

According to description, this commit fixes CLANG-10 warnings, however it seems to change behavior as well. Running with it creates this error (see below for test script description):
```
%MSG-e EvFDaqDirector:  (NoModuleName) 19-Jun-2020 20:37:32 CEST pre-events
grabNextFile Failed to deserialize JSON file -: "./data/run000100/run000100_ls0001_index000000_pid28154.jsn"
ERROR:
* Line 2, Column 20
  Missing ',' or ']' in array declaration
CONTENT:
{
   "data" : [ "20" ],
   "definition" : "./ramdisk//run000100/jsd/rawData.jsd",
   "source" : "lxplus709.cern.ch_28075"
}
.
%MSG
%MSG-e EvFDaqDirector:  (NoModuleName) 19-Jun-2020 20:37:32 CEST pre-events
grabNextFile runtime Exception -: Cannot deserialize input JSON file
%MSG
cmsRun: /tmp/smorovic/CMSSW_11_1_0_pre8/src/EventFilter/Utilities/src/FedRawDataInputSource.cc:1013: void FedRawDataInputSource::readSupervisor(): Assertion `eventsInNewFile >= 0' failed.
```
Assertion is consequence of the previous failure.

#### PR validation:
Tested manually and traced to be introduced between CMSSW_11_1_0_pre7 and CMSSW_11_1_0_pre8.
Test scripts in EventFilter/Utilities/test:
 - startBU.py (terminate manually after 10 seconds) and startFU.py, or
 - LocalRunBUFU.sh (with a log file in /tmp)

Failure mode is currently not covered with any automatic unit tests, as there have been issues implementing and running those reliably with DAQ file locking mechanism (also, for Run3 use of JSON files at HLT input stage is being phased out).

Equivalent PR will be created also for 11_1_X branch because currently this breaks release use in DAQ/HLT.

### DIFF
--- a/EventFilter/Utilities/src/json_reader.cpp
+++ b/EventFilter/Utilities/src/json_reader.cpp
@@ -430,7 +430,7 @@ namespace Json {
       while (token.type_ == tokenComment && ok) {
         ok = readToken(token);
       }
-      bool badTokenType = (token.type_ == tokenArraySeparator || token.type_ == tokenArrayEnd);
+      bool badTokenType = (token.type_ == tokenArraySeparator && token.type_ == tokenArrayEnd);
       if (!ok || badTokenType) {
         return addErrorAndRecover("Missing ',' or ']' in array declaration", token, tokenArrayEnd);
       }
--- a/EventFilter/Utilities/src/json_reader.cpp
+++ b/EventFilter/Utilities/src/json_reader.cpp
@@ -430,7 +430,8 @@ namespace Json {
       while (token.type_ == tokenComment && ok) {
         ok = readToken(token);
       }
-      if (!ok) {
+      bool badTokenType = (token.type_ != tokenArraySeparator && token.type_ != tokenArrayEnd);
+      if (!ok || badTokenType) {
         return addErrorAndRecover("Missing ',' or ']' in array declaration", token, tokenArrayEnd);
       }
       if (token.type_ == tokenArrayEnd)
--- a/EventFilter/Utilities/src/json_reader.cpp
+++ b/EventFilter/Utilities/src/json_reader.cpp
@@ -430,8 +430,7 @@ namespace Json {
       while (token.type_ == tokenComment && ok) {
         ok = readToken(token);
       }
-      bool badTokenType = (token.type_ == tokenArraySeparator && token.type_ == tokenArrayEnd);
-      if (!ok || badTokenType) {
+      if (!ok) {
         return addErrorAndRecover("Missing ',' or ']' in array declaration", token, tokenArrayEnd);
       }
       if (token.type_ == tokenArrayEnd)


EventFilter/Utilities/src/json_reader.cpp
```diff
@@ -430,7 +430,7 @@ namespace Json {
       while (token.type_ == tokenComment && ok) {
         ok = readToken(token);
       }
-      bool badTokenType = (token.type_ == tokenArraySeparator || token.type_ == tokenArrayEnd);
+      bool badTokenType = (token.type_ == tokenArraySeparator && token.type_ == tokenArrayEnd);
       if (!ok || badTokenType) {
```
smuzaffar, line 434: This does not make sense, how can this condition be true. As it will always be false so better to remove it

smorovic: You're right. I bisected and found what broke the code, but didn't spot that original condition is always false.

smorovic: jsoncpp seems to support comment at the end of array, which is non-standard. The check is part of this (and buggy). That feature is not relevant for what it is used in Event Filter code, so I dropped the condition with another commit.

perrotta:
@cmsbuild please test

perrotta:
So, this PR reverts to the situation that was there before #29592, when ```badTokenType``` was always identically ```false```.
As long as DAQ confirms that such a ```badTokenType``` was really useless (i.e. there was not a typo in it, as #29592 implicitly assumed) this can be considered to be ok also by reco

smorovic:
@perrotta this code is from the external project included in the tree long time ago. It has also diverged, getting e.g. thread safety and various other fixes from CMSSW devs.

I had removed the condition because I couldn't figure out the original intention. But now I checked the original project and found out that the bug was fixes by original authors (problem was, '==' operators are supposed be '!='). Commit in their project:
https://github.com/open-source-parsers/jsoncpp/commit/9c98f2277b2972ddee93c97bb5273e86d6045195

I will test it first and then update both PRs. It is more 'proper' although either way would work fine for DAQ.

smorovic:
I applied the described bugfix and tested prior to pushing it to both PRs.

perrotta:
@cmsbuild please test

emeschi:
+1

perrotta:
+1
- Fixed a wrong expression implemented with #29592: original intention behind the code is now restored
- Jenkinns tests pass

silviodonato:
+1

PR #30321, opened 2020-06-21
Reduce boost in CalibTracker/SiStripLorentzAngle

camolezi:
#### PR description:
Replaced boost::regex for std::regex, they should have similar behaviour.

#### PR validation:

@vgvassilev @davidlange6


christopheralanwest:
please test

pohsun:
+1

silviodonato:
+1

camolezi:
Can you review this again?
I've made some benchmarks on std::regex vs boost::regex.
They show that std::regex is at least 10x slower than boost::regex.

https://gist.github.com/camolezi/87283aaf7cdeb584ba294588a5d77abd
https://gist.github.com/camolezi/ae83ce140a5ade6e9c4aee158f9f95e2

With this new information, we may need to revert this, if performance is needed in this case.

PR #30336, opened 2020-06-22
added esConsumes to modules in L1Trigger /L1ExtraFromDigis

JamminJones:
#### PR description:

added esConsumes to 1 module
#### PR validation:

the code compiles

### DIFF
--- a/L1Trigger/L1ExtraFromDigis/interface/L1ExtraParticlesProd.h
+++ b/L1Trigger/L1ExtraFromDigis/interface/L1ExtraParticlesProd.h
@@ -25,6 +25,7 @@
 #include "FWCore/Framework/interface/stream/EDProducer.h"
 #include "FWCore/ParameterSet/interface/ParameterSet.h"
 #include "FWCore/Utilities/interface/InputTag.h"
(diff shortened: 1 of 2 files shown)


Dr15Jones:
This code change was made under my supervision.

Dr15Jones:
hold

the EventSetup gets are done conditionally so we should set the tokens conditionally as well.

Dr15Jones:
please test

Dr15Jones:
unhold

Dr15Jones:
NOTE: The new code conditionally calls esConsumes based on the conditional getting of the data from the EventSetup. This was done to avoid prefetching of EventSetup data that will not be needed.

Looking at the existing `consumes` calls, many of them should also be conditional as the configuration parameters set which Event data will actually be requested. Not calling `consumes` conditionally is requiring the framework to prefetch Event data which is not needed.

I suggest that the maintainers of this code modify the `consumes` calls to only be called if needed.

Dr15Jones:
ping

Dr15Jones:
ping (it has been  2 weeks now without a review)

rekovic:
+1

silviodonato:
+1

PR #30292, opened 2020-06-18
Adding eval=true in ConstantsSection in all PPS xml files

mundim:
#### PR description:
Adding eval="true" in the PPS xml files, as requested in

https://github.com/cms-sw/cmsdist/pull/5954#issuecomment-645925272

#### PR validation:
Tests passed  and the new files were tested again in the PPS simulation, producing properly the hits as expected.

PR #30340, opened 2020-06-23
Adding geometry scenarios D59 and D60 implementing ETL version 4 and 5

icosivi:
#### PR description:

This PR implements geometry scenarios D59 and D60: 
- D59 modifies scenario D58 introducing ETL version 4
- D60 modifies scenario D58 introducing ETL version 5

ETL version 4 was already implemented in scenario D53 (https://indico.cern.ch/event/868736/contributions/3709315/attachments/1978993/3295064/MTDdays_20200131.pdf) and fits in the new Endcap Calorimeter envelope defined in D58.
ETL version 5 implements the latest version of ETL disks, as proposed by N. Koss (https://indico.cern.ch/event/868736/contributions/3706894/attachments/1978848/3294508/20200130_ETL_mechanics_timing_days.pdf).

ETL v5 introduces D-shaped sectors, instead of the quarters used in ETL v4, comprising more than 512 modules, which is the maximum number of modules that could be accommodated using the allocated bits in ETLDetId. Therefore ETLDetId.h has been modified to fix this issue, by adding 2 bits. The new ETLDetId is backward compatible with the older versions, thanks to a dedicated compatibility code implemented in ETLDetId.cc.

Both ETL versions and the backward compatibility issue were presented at the last MTD DPG meeting (https://indico.cern.ch/event/919571/contributions/3864046/attachments/2039610/3415525/MTD_DPG_20200515.pdf )

#### PR validation:
- Both scenarios have been tested using testMTDinDDD.py and testMTDinDD4hep.py: the dump geometries have been checked and both volumes positions and logical paths are correct.
- The overlaps have been checked too using  g4OverlapCheck_cfg.py.
- The dump geometries for scenario D50 using the old ETLDetId format and the new one described above have been compared: the logical volumes paths are the same, while the rawID differ only for the positions of the 2 bits related to the sensor modules, as expected. The new ETLDetId format is therefore correct and backward compatible.


fabiocos:
please test

fabiocos:
@icosivi as you update ```ETLDetId``` in a backward incompatible way, the unit test comparison fails because the reference files are old (at least the MTD tests). To be understood for the others

fabiocos:
other unit tests failures seem to be in the release itself (@silviodonato please confirm, is this going to be fixed?)

silviodonato:
Yes, there are 6 unit tests failing in IB. They are going to be fixed see https://github.com/cms-sw/cmssw/issues/30298 and https://github.com/cms-sw/cmssw/pull/30292

fabiocos:
@icosivi te failures in the relval tests seem to come from https://cmssdt.cern.ch/lxr/source/RecoLocalFastTime/FTLCommonAlgos/src/MTDTimeCalib.cc#0021 
To be understood why

silviodonato:
@fabiocos https://github.com/cms-sw/cmssw/pull/30292 has been merged right now. The unit test failures should be fixed now.

fabiocos:
please test

knowing that part of the unit tests and relvals will anyway fail

fabiocos:
please test

kpedro88:
@icosivi does the latest commit address the test failure? if not, please make an explicit comment when the PR is expected to work. (As always, please test the PR locally first.)

icosivi:
@kpedro88 not yet, I will make an explicit comment when it will be expected to work, thanks

fabiocos:
@kpedro88 in my understanding from @icosivi this PR was locally tested with the standalone geometry checks, and no obvious issue was found. The failure of the unit test in itself is totally trivial, as ETL DetIds change a new reference needs to be used. What is clearly unexpected is the failure in the old relvals, which is of course something that needs to be verified to work. The test of reading old files with new code and verify that the translation was properly done was done (@icosivi please confirm). But apparently there is some unexpected issue in the chain. The Geant4 step looks ok from the tests I have suggested, while a closer check to DIGIs is needed, and the update of the dumper I asked @icosivi is meant for this.

I believe this PR could be put at least on hold. But probably we could have a different strategy: split it into a PR to update ETLDetId, once all checks are completed and problems identified and solved, and another one with the remaining update of new scenarios, that looks less prone to create issues.

icosivi:
@fabiocos I confirm that the test of reading old files with new code was done

fabiocos:
hold

this may be eventually split into a PR for ETLDetId, and another one with the new scenarios (updated for the new work by @bsunanda )

DataFormats/ForwardDetId/src/classes_def.xml
```diff
@@ -30,8 +30,15 @@
   <class name="BTLDetId" ClassVersion="3">
    <version ClassVersion="3" checksum="1515591716"/>
   </class>
-  <class name="ETLDetId" ClassVersion="3">
+  <class name="ETLDetId" ClassVersion="4">
+   <version ClassVersion="4" checksum="1644731879"/>
    <version ClassVersion="3" checksum="1644731879"/>
+   <ioread sourceClass = "ETLDetId" version="[-4]" targetClass="ETLDetId" source="" target="">
```
fabiocos, line 36: @icosivi @casarsa most likely the version here should be [-3], we want the conversion on older ClassVersion data. This would explain why everything looks ok while writing, but then it miserably fails after reading back data from file.

fabiocos: while it looks ok reading data from previous versions...

